PR #336, opened 2018-02-09
Modifies the user ID cookie to have a configurable TTL

### LINKED ISSUE #326: Make Cookie TTLs configurable
Right now, the TTL on synced User IDs is hardcoded: https://github.com/prebid/prebid-server/blob/master/pbs/usersync.go#L146

For some reason, we need to set a different TTL (I didn't get the details, but @mkendall07 has them)... so we plan to make this configurable.

hhhjort:
Fixes #326

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,6 +35,8 @@ type HostCookie struct {
 	OptOutURL    string `mapstructure:"opt_out_url"`
 	OptInURL     string `mapstructure:"opt_in_url"`
 	OptOutCookie Cookie `mapstructure:"optout_cookie"`
+	// Cookie timeout in days
+	TTL int64 `mapstructure:"ttl"`
 }
 
 type Adapter struct {
--- a/config/config.go
+++ b/config/config.go
@@ -36,7 +36,7 @@ type HostCookie struct {
 	OptInURL     string `mapstructure:"opt_in_url"`
 	OptOutCookie Cookie `mapstructure:"optout_cookie"`
 	// Cookie timeout in days
-	TTL int64 `mapstructure:"ttl"`
+	TTL int64 `mapstructure:"ttl_days"`
 }
 
 type Adapter struct {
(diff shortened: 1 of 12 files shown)


config/config.go
```diff
@@ -35,6 +35,8 @@ type HostCookie struct {
 	OptOutURL    string `mapstructure:"opt_out_url"`
 	OptInURL     string `mapstructure:"opt_in_url"`
 	OptOutCookie Cookie `mapstructure:"optout_cookie"`
+	// Cookie timeout in days
+	TTL int64 `mapstructure:"ttl"`
```
dbemiller: Can we call this `ttl_days`? Less chance of screwing it up while editing the config.

pbs_light.go
```diff
@@ -885,6 +885,7 @@ func serve(cfg *config.Configuration) error {
 		OptOutURL:    cfg.HostCookie.OptOutURL,
 		OptInURL:     cfg.HostCookie.OptInURL,
 		OptOutCookie: cfg.HostCookie.OptOutCookie,
+		TTL:          time.Duration(cfg.HostCookie.TTL) * 24 * time.Hour,
```
dbemiller: Are you sure this is right?

It looks like you're multiplying by `* 24 * time.Hour` both here and in `ToHTTPCookie` in `usersync.go`. I'm not certain, but... seems like a bug?

config/config.go
hhhjort: Done

pbs_light.go
hhhjort: Good catch. go is fine implicitly converting a static `int` to `time.Duration`, but not converting a variable of type `int64`. So I found I had to cascade back a `time.Duration`declaration, and at the end (beginning) I again did a day-time conversion. I think its cleaner to do the conversion at the front, and let `ttl` be a `time.duration` throughout the code than do it as the last step.

dbemiller: I had almost exactly the same bug recently. I don't like how Go implements these type aliases at all.

I think your strategy is the right one, though. Convert once ASAP, and pass the value through from there.

hhhjort:
I added a default for the TTL. Did notice that we don't have defaults for any of the other cookie parameters, but on second look I don't think we can pick sensible defaults for the remainder. This will keep that app backwards compatible with old configs, which is plenty of reason to do it still.

mkendall07:
@bretg fyi on this - we added a config for pbs cookie ttl but it should be backwards compatible.

PR #339, opened 2018-02-12
Discard data more liberally on database errors.

### LINKED ISSUE #338: Stored Request Cache Bug?
The other day, I saw some weird behavior.

@anwzhang sent me an HTTP payload with Stored Request IDs to reproduce an issue she was seeing. I didn't see the same behavior locally. I retried the request several times, and always got back the same result. After a meeting, I tried it again and got a different error. Again, the result was consistent. Later that day, I tried it again and got the result she expected.

Judging by the symptoms, I think this is a bug in the Stored Request cache. Unfortunately I don't remember exactly what the response was on the _first_ call for each of these cases... but with local dev & remote database, it's very common that the first request to the DB times out, and subsequent ones work fine.

I think what's happening is that the call to the database returns some corrupt data, which we then save to the cache. Depending on exactly when the request times out, we can see different behavior for the same request.

dbemiller:
Fixes #338 ...hopefully. It's tough to say for sure, because it's not really reproducible... but it seems like this should help.

### DIFF
--- a/stored_requests/backends/db_fetcher/fetcher.go
+++ b/stored_requests/backends/db_fetcher/fetcher.go
@@ -49,23 +49,31 @@ func (fetcher *dbFetcher) FetchRequests(ctx context.Context, ids []string) (map[
 		}
 		return nil, []error{err}
 	}
-	defer rows.Close()
+	defer func() {
+		if err := rows.Close(); err != nil {
+			glog.Errorf("error closing DB connection: %v", err)
+		}
+	}()
 
 	reqData := make(map[string]json.RawMessage, len(ids))
-	var errs []error = nil
 	for rows.Next() {
 		var id string
 		var thisReqData []byte
+
+		// Fixes #338?
 		if err := rows.Scan(&id, &thisReqData); err != nil {
-			errs = append(errs, err)
+			return nil, []error{err}
 		}
 
 		reqData[id] = thisReqData
 	}
+
+	// Fixes #338?
 	if rows.Err() != nil {
-		errs = append(errs, rows.Err())
+		return nil, []error{rows.Err()}
 	}
 
+	var errs []error
 	for _, id := range ids {
 		if _, ok := reqData[id]; !ok {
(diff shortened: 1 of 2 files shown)


stored_requests/backends/db_fetcher/fetcher.go
```diff
@@ -49,23 +49,31 @@ func (fetcher *dbFetcher) FetchRequests(ctx context.Context, ids []string) (map[
...
 		var id string
 		var thisReqData []byte
+
+		// Fixes #338?
 		if err := rows.Scan(&id, &thisReqData); err != nil {
-			errs = append(errs, err)
+			return nil, []error{err}
 		}
 
 		reqData[id] = thisReqData
```
hhhjort: Wouldn't we fix the problem if we simply put this line into an `else`, that is only populate `reqData[]` if there were no errors? Or is there some logic downstream that depends on all IDs being populated if we don't return a nil?

dbemiller: That was actually my first thought too, but... on closer investigation, I don't think so.

I tried to write a failed test to verify it, but `sqlmock` doesn't let you fake "scan errors". There's a good explanation of why here: https://github.com/DATA-DOG/go-sqlmock/issues/47

I tried to save other types as mock row data, but... it seems like Go is pretty liberal about what it coerces into `[]byte`. Both `true` and `-1` came back as real data, with no `Scan` error. I'm not sure what if there are other types of data which we could use to force it, either.

Judging by the race condition tickets I linked in #338, though, I think this behavior is caused by something much less straightforward.

hhhjort: Ahh, yes, dug into them now. If the buffer is being corrupted, that makes subsequent calls suspect. May be able to get away with exiting returning the rows collected prior to the row with the error. But then at that point just tossing everything out may not be a bad response either.

Sounds like we should be able to revisit after go 1.10 is out, if we remember this issue.

dbemiller:
Merging this per the new process. No complaints from the second reviewer in > 5 days.

PR #342, opened 2018-02-13
Added Adform adapter

vstatkevich:


adapters/adform/adform.go
```diff
@@ -0,0 +1,380 @@
...
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/pbs"
+	"golang.org/x/net/context/ctxhttp"
+)
+
+type AdformAdapter struct {
+	http         *adapters.HTTPAdapter
+	URI          string
+	usersyncInfo *pbs.UsersyncInfo
```
hhhjort: Now that we have a usersync module, does the adapter still need usersync info?

adapters/adform/adform.go
```diff
@@ -0,0 +1,380 @@
...
+
+	return requests, errors
+}
+
+func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []error) {
+	adUnits := make([]*adformAdUnit, 0, len(request.Imp))
+	errors := make([]error, 0, len(request.Imp))
+	for _, imp := range request.Imp {
+		// adapter supports only banner impressions for now.
+		if imp.Native != nil || imp.Audio != nil || imp.Video != nil {
```
hhhjort: Technically the spec allows for requesting multiple types of ads for one Imp. It would then be up to the publisher to sort out how to display the ad depending on what type of ad was returned. But still a valid strategy to abstain from bidding if competing against the other types, as they are likely to be bidding higher anyway.

More importantly, some future addition to the spec may allow for yet another ad type. You would be more future proof if you checked for `banner` as a requested type, rather than simply assuming if its not one of the others it must be `banner`.

adapters/adform/adform.go
vstatkevich: No need, have removed it

adapters/adform/adform.go
vstatkevich: Fixed

sandraleon:
Haven't had a chance to review yet. I should have some time tomorrow.

adapters/adform/adform.go
```diff
@@ -0,0 +1,364 @@
...
+		referer = request.Site.Page
+	}
+
+	tid := ""
+	if request.Source != nil {
+		tid = request.Source.TID
+	}
+
+	secure := false
+	for _, imp := range request.Imp {
```
sandraleon: just a suggestion, maybe you could check if it's secure in the for loop above (line 263) since you're already looping through the imps.

vstatkevich: updated

usersync/usersync.go
```diff
@@ -30,6 +30,7 @@ func NewSyncerMap(cfg *config.Configuration) map[openrtb_ext.BidderName]Usersync
 		openrtb_ext.BidderPubmatic:   NewPubmaticSyncer(cfg.ExternalURL),
 		openrtb_ext.BidderPulsepoint: NewPulsepointSyncer(cfg.ExternalURL),
 		openrtb_ext.BidderRubicon:    NewRubiconSyncer(cfg.Adapters["rubicon"].UserSyncURL),
+		openrtb_ext.BidderAdform:     NewAdformSyncer(cfg.Adapters["adform"].UserSyncURL, cfg.ExternalURL),
```
dbemiller: Going to merge this PR because it's got the approvals, but... one open issue you should be aware of.

The way this is written, a PBS host company won't know what to set their config value to be in order to make usersyncs work with your bidder... so you'll be at a disadvantage compared to the other bidders in the project.

To level the playing field, you may want to add a `viper.SetDefault()` line to `init()` in `pbs_light.go`. Or, if there's no reason why one PBS deploy would use a different sync URL than any other, just eliminate the config altogether and hardcode the URL as a `const` in your Usersyncer.

This can be addressed in a separate PR, though, and is not required for the project.

PR #346, opened 2018-02-14
Allow explicit userids

### LINKED ISSUE #320: Allow User IDs to be passed explicitly
## Motivation

Right now, the only way to pass different User IDs for each Bidder is through the Cookie.

Since cookies are sent by the browser automatically, this can't be used to store them manually somewhere (e.g. `localStorage` in `prebid.js`) so they can be sent explicitly in the request.

## Proposal

Allow `bidrequest.user.ext.prebid` to support the following:

```
{
  "buyeruids": {
    "appnexus": "abc",
    "rubicon": "def"
  }
}
```

For any defined `buyeruids` values, the core auction logic would use these values to populate the `request.user.buyeruid` field before forwarding them to the Bidder. It would also delete the `request.user.ext.prebid.buyeruids` field during preprocessing, so that info intended for one Bidder doesn't leak into the others.

If the user's request defines both `request.user.buyeruid` _and_ `request.user.ext.prebid.buyeruids`, we'll return a 400. If defined, `request.user.ext.prebid.buyeruids` will take precedence over the Cookie values. This will be consistent with how we treat other "implicitly populated" fields (user-agent, IP, etc.)

dbemiller:
Fixes #320

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -391,24 +391,31 @@ func (deps *endpointDeps) validateSite(site *openrtb.Site) error {
 	return nil
 }
 
-func (deps *endpointDeps) validateUser(user *openrtb.User, aliases map[string]string) error {
+func validateUser(user *openrtb.User, aliases map[string]string) error {
 	// DigiTrust support
 	if user != nil && user.Ext != nil {
 		// Creating ExtUser object to check if DigiTrust is valid
 		var userExt openrtb_ext.ExtUser
 		if err := json.Unmarshal(user.Ext, &userExt); err == nil {
 			// Checking if DigiTrust is valid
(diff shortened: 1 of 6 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -391,17 +391,35 @@ func (deps *endpointDeps) validateSite(site *openrtb.Site) error {
...
+					if _, ok := openrtb_ext.BidderMap[bidderName]; !ok {
+						if _, ok := aliases[bidderName]; !ok {
+							return fmt.Errorf("request.user.ext.%s is neither a known bidder name nor an alias in request.ext.prebid.aliases.", bidderName)
+						}
+					}
+				}
+			}
+
+			// Make sure that user.ext is not empty.
+			if userExt.DigiTrust == nil && userExt.Prebid == nil {
```
cindyhere33: Maybe this can be moved to line 400?

PR #347, opened 2018-02-15
Adding bid cache to the docs

dbemiller:
@sandraleon was asking questions about this, so... adding docs for it.

Let me know if this is unclear, or doesn't cover what you need to know.


docs/endpoints/openrtb2/auction.md
```diff
@@ -226,6 +226,28 @@ A typical `storedrequest` value looks like this:
...
+
+```
+{
+  "bids": {}
+}
+```
+
+If `request.ext.prebid.targeting` was also in the request, then Prebid Server will make a _best effort_ to
+include these extra `bid.ext.prebid.targeting` keys:
+
```
hhhjort: Should probably put in something to indicate how to find the cache keys if targeting parameters are not set, or explicitly call out that caching is not supported in the case that targeting is not requested.

dbemiller: That sounds good. For now, caching is not supported when targeting is not requested. I definitely expect someone to add this feature in the future though.

Issue #295, opened 2018-01-25
Split auction_test.go tests into files


dbemiller:
The [endpoints/openrtb2/auction_test.go](https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/auction_test.go) file embeds lots of JSON into strings, which has a few problems:

1. It's easy to typo a comma inside a string and have a valid request be rejected as "invalid" because of the typo.
2. When a test fails, it's difficult to tell which test case failed, or why.

These test cases should be migrated into `.json` files, like the [adapters](https://github.com/prebid/prebid-server/tree/master/adapters/appnexus/appnexustest/supplemental) do.

cirla:
@dbemiller is this closed by https://github.com/prebid/prebid-server/pull/435?

dbemiller:
No... different layer of the application, sadly.

There are still some tests in `auction_test.go` which handle Stored Request stuff: https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/auction_test.go#L497

It would be nice to clear those out into JSON files at some point too.

Issue #218, opened 2017-11-27
OpenRTB Support: Native request validation


dbemiller:
Currently, the `/auction/openrtb2` endpoint does not validate `request.imp[i].native.request`, even though it should be a JSON-encoded string following [the OpenRTB Native Extension](https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-Native-Ads-Specification-1-1_2016.pdf).

We should do this, and return 400s on bad input. The openRTB library we use [recently added](https://github.com/mxmCherry/openrtb/pull/19) contract classes for Native objects, which should help.

dbemiller:
This was implemented in #353. I'll close it and open a new issue for Native 1.2 support.

PR #353, opened 2018-02-20
Validate native inputs

dbemiller:
This fixes _most_ of #218, and _some_ of #295.

OpenRTB Native v1.2 support was added in https://github.com/mxmCherry/openrtb/issues/20, but the 10.0.0 release of that library included some other breaking changes... so in the interest of moving this along, I'll save those changes for another PR. 

Most of these changes are just refactoring so that my native test payloads don't add contribute to the tech debt of #295. I pulled many of the existing test payloads out into their own `*.json` files.


endpoints/openrtb2/amp_auction_test.go
```diff
@@ -4,24 +4,66 @@ import (
...
 // const maxSize = 1024 * 256
 
 // TestGoodRequests makes sure that the auction runs properly-formatted stored bids correctly.
 func TestGoodAmpRequests(t *testing.T) {
+	testAmpStoredRequestData := map[string]json.RawMessage{
+		"10":  json.RawMessage(validRequest(t, "site.json")),
+		"11":  json.RawMessage(validRequest(t, "app.json")),
+		"12":  json.RawMessage(validRequest(t, "timeout.json")),
+		"100": json.RawMessage("5"),
+		"101": json.RawMessage("5"),
```
hhhjort: Ok, that was a bug in the original AMP tests, should have been testing all the invalid requests. Not terribly important though, since this shares code with openrtb, and the invalid requests should be handled there anyway. I had failed to advance the index in `invalidRequests[]` on each line.

endpoints/openrtb2/auction.go
```diff
@@ -300,6 +299,183 @@ func validateBanner(banner *openrtb.Banner, impIndex int) error {
...
+	}
+
+	for i := 0; i < len(assets); i++ {
+		// Per the OpenRTB spec docs, this is a "unique asset ID, assigned by exchange. Typically a counter for the array"
+		// To avoid conflict with the Request, we'll return a 400 if the Request _did_ define this ID,
+		// and then populate it as the spec suggests.
+		if err := validateNativeAsset(assets[i], impIndex, i); err != nil {
+			return err
+		}
+		assets[i].ID = int64(i)
```
hhhjort: From the spec: "Typically a counter for the array." This enforces that it will be a counter. I don't know of any use cases for ID being anything else, so its probably good. But potentially will break something that stretches the spec here.

endpoints/openrtb2/auction.go
```diff
@@ -300,6 +299,183 @@ func validateBanner(banner *openrtb.Banner, impIndex int) error {
...
+	}
+	if err := validateNativeVideoProtocols(video.Protocols, impIndex, assetIndex); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateNativeAssetData(data *nativeRequests.Data, impIndex int, assetIndex int) error {
+	if data.Type < 1 || data.Type > 11 {
```
hhhjort: 12 is a valid data type.

dbemiller: Good catch! haha 12 was on a separate page in the spec, so I must have missed it.

endpoints/openrtb2/auction.go
dbemiller: yeah... to be honest I'm not 100% sure how to handle this. I saw the following:

> Unique asset ID, assigned by exchange.

Since Prebid is the exchange, in this context, I don't _think_ we're breaking the spec here. We just decided that the "typical" implementation is the one we'll be using.

I decided to return 4xx on IDs sent by the Publisher just to prevent all sorts of nasty logic/confusion. As a general rule, we never _overwrite_ anything which the Publisher gives us. So if we allowed IDs, then we'd have to make sure that the ones we added weren't duplicates, which gets... messy.

Decided to just reject those requests to keep things simple.

endpoints/openrtb2/amp_auction_test.go
dbemiller: ah... I see. Yeah, I didn't pay much attention to the semantics of these tests, but I'll clean them up so they make sense again.

endpoints/openrtb2/auction.go
hhhjort: I am fine with that then.

dbemiller:
been way longer than 5 days, so... merging.

Issue #341, opened 2018-02-12
Need to do cookie sync for all bidders when passing empty bidders


jaiminpanchal27:
In Amp RTC config example we will be using Amp-iframe to do the pixel syncs. But we don't know the bidders so we would like to have cookie sync status for all the bidders.

Request like this should return status for all bidders
```
{"uuid":"<uuid>","bidders":[]}
```

dbemiller:
We chatted offline about this... I think it makes sense, with a few caveats.

## Motivation

AMP pages include one element for Auctions, and another totally separate element for Usersyncs.

The `/openrtb2/amp` endpoint takes a Stored Request ID, and resolves it as an _entire_ OpenRTB Request. Because of this, the AMP page has no idea which Bidders are involved.

## Options

There are a few implementation options and tradeoffs we can consider.

### Sync all bidders

As @jaiminpanchal27 suggested, we could expose an API which returns Usersyncs for _every_ bidder.
This would be fine now, but will become inefficient as the number of Bidders grows in PBS.

### Require AMP pages to list Bidders

The existing endpoint supports this already. A major drawback is that publishers need to remember to update the bidder list on their pages whenever they add or remove one from their Stored Request.

### Resolve usersyncs from Stored Request ID(s)

We could allow Pubs to send in their Stored Request ID when generating usersyncs. We would need to support multiple IDs, since the `openrtb2/amp` endpoint only supports a single `Imp`, and any given page may have multiple Imps.

If we added this, we could parse the Stored Requests to figure out which bidders are  currently needed, and return the syncs for just those Bidders.

This has the minor drawback that pubs need to put the Stored Request ID on their AMP pages in two places (auction & usersync), which is vulnerable to typos. @mkendall07 also points out that Pubs may want to "seed" their page with syncs from a new Bidder before they add that bidder to the auctions.
 
## Use case summary

In the long term, the cookie sync endpoint should support:

1. Specifying Bidders through Stored Request IDs
2. Specifying Bidders manually
3. Specifying all bidders

## Next steps

I think we should first work out an API which satisfies all of the known use-cases, and work backwards from there to make a feature we can implement now. It's the best way (that I know of) to minimize the chance of breaking changes later.

hhhjort:
My thoughts on the use case summary above:
2. is already handled
3. is trivial if we say if no `bidders` field plus no stored request supplied equals sync all bidders.
So 1. is where all the action is.

[...]

PR #354, opened 2018-02-20
Cookie sync 341

hhhjort:
Partially implements the solution to #341. Takes care of the original request, but none of the extra features discussed in the thread.

### DIFF
--- a/docs/endpoints/cookieSync.md
+++ b/docs/endpoints/cookieSync.md
@@ -7,12 +7,10 @@ must supply a JSON object to define the list of bidders that may need to be sync
 
 ```
 {
-    "uuid": "blahblah",
     "bidders": ["appnexus", "rubicon"]
 }
 ```
 
-The `uuid` field is due to be deprecated, as it is no longer needed. It just gets passed through unused into the response currently.
 If the `bidders` field is an empty list, it will not supply any syncs. If the `bidders` field is omitted completely, it will attempt
 to sync all bidders.
 
@@ -20,7 +18,6 @@ This will return a JSON object that will allow the client to request cookie sync
 
 ```
 {
-    "uuid": "blahblah",
     "status": "ok",
     "bidder_status": [
         {
--- a/docs/endpoints/cookieSync.md
+++ b/docs/endpoints/cookieSync.md
@@ -1,0 +1,37 @@
+# Handling cookie syncs.
+
+## POST /cookie_sync
+
+This returns a set of URLs to enable cookie syncs across bidders. (See Prebid.js documentation?) The request
+must supply a JSON object to define the list of bidders that may need to be synced.
+
+```
+{
+    "uuid": "blahblah",
+    "bidders": ["appnexus", "rubicon"]
+}
+```
+
+The `uuid` field is due to be deprecated, as it is no longer needed. It just gets passed through unused into the response currently.
+If the `bidders` field is an empty list, it will not supply any syncs. If the `bidders` field is omitted completely, it will attempt
+to sync all bidders.
+
+This will return a JSON object that will allow the client to request cookie syncs with bidders that still need to be synced:
+
+```
+{
+    "uuid": "blahblah",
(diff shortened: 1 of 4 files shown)


docs/endpoints/cookieSync.md
```diff
@@ -0,0 +1,37 @@
...
+must supply a JSON object to define the list of bidders that may need to be synced.
+
+```
+{
+    "uuid": "blahblah",
+    "bidders": ["appnexus", "rubicon"]
+}
+```
+
+The `uuid` field is due to be deprecated, as it is no longer needed. It just gets passed through unused into the response currently.
```
dbemiller: If the endpoint works without it, don't document it here. With no documentation, it's unlikely anyone will start using it. If it's mentioned, then some people will. Also: it complicates the docs & wastes time for anyone who just wants to know how to use it properly.

Honestly... Kendall said prebid.js wasn't using it, and Prebid Mobile doesn't use Cookies. If we're going to make breaking changes, sooner is better than later. I think we should just delete it from the endpoint now.

pbs_light.go
```diff
@@ -211,14 +211,37 @@ func (deps *cookieSyncDeps) CookieSync(w http.ResponseWriter, r *http.Request, _
...
+	csReqRaw := map[string]json.RawMessage{}
+	err := json.NewDecoder(r.Body).Decode(&csReqRaw)
 	if err != nil {
 		if glog.V(2) {
 			glog.Infof("Failed to parse /cookie_sync request body: %v", err)
 		}
 		http.Error(w, "JSON parse failed", http.StatusBadRequest)
 		return
 	}
+	if UUID, ok := csReqRaw["uuid"]; ok {
```
dbemiller: I'm pretty confused... you're parsing this, but the value isn't used anywhere. What's the point here?

hhhjort: I need to parse it of I am going to pass it through. If we a deprecating it then I can not do it. :) Just grabbing it raw leaves in the quotes around the string.

docs/endpoints/cookieSync.md
hhhjort: It is now deprecated completely. :) Bandaid yanked!

pbs_light.go
hhhjort: Oh, you are correct that I never put it in `csReq`. But now I have removed `UUID` from the struct and it is now a one element list. Should I keep the struct (balances against the response struct, and eases expanding the API later) or should I just pull the struct and just work with a bidders list?

pbs_light_test.go
```diff
@@ -113,10 +103,70 @@ func TestCookieSyncHasCookies(t *testing.T) {
...
+
+	if csresp.Status != "no_cookie" {
+		t.Errorf("Expected status = no_cookie; got %s", csresp.Status)
+	}
+
+	if len(csresp.BidderStatus) != 0 {
+		t.Errorf("Expected 0 bidder status rows; got %d", len(csresp.BidderStatus))
+	}
+
+	// Now test a missing bidders returns all syncs
```
dbemiller: You might want to split this into two tests.

Extreme "test-minded" engineers will tell you that each test should be tiny--ideally a single assertion. I've tried that, and think it's a bit overkill... but they're definitely right that short tests which check "a single thing" are easier to debug.

It basically comes down to "when a test fails, how much code do I have to read to understand to figure out what went wrong?" Methods like this double that effort.

hhhjort: I had pondered splitting that myself ... so now done.

pbs_light.go
```diff
@@ -184,12 +184,10 @@ func getAccountMetrics(id string) *AccountMetrics {
 }
 
 type cookieSyncRequest struct {
-	UUID    string   `json:"uuid"`
```
mkendall07: extra fields are ignored right? in other words, legacy requests with UUID continue to work

hhhjort: That is correct

PR #355, opened 2018-02-21
Fixing some race conditions

dbemiller:
High priority changes to fix some race conditions which are crashing us in prod.

Don't pay much attention to `race_test.go`; I'm going to be rewriting it very shortly, building it such that it won't be overlooked by new adapter submissions. I'm including it because it proves that this fixes some real race conditions.

You can verify with `go test -race github.com/prebid/prebid-server/endpoints/openrtb2 -run ^TestRequestConcurrently$ -count 100`. On `master`, it shows a bunch of race conditions. On this branch, it passes cleanly.

### DIFF
--- a/endpoints/openrtb2/race_test.go
+++ b/endpoints/openrtb2/race_test.go
@@ -1,0 +1,106 @@
+package openrtb2
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/exchange"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/pbsmetrics"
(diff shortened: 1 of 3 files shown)


endpoints/openrtb2/race_test.go
```diff
@@ -0,0 +1,106 @@
...
+	metrics "github.com/rcrowley/go-metrics"
+)
+
+func TestRequestConcurrently(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(nobidServer))
+	defer server.Close()
+
+	cfg := &config.Configuration{}
+
+	theMetrics := pbsmetrics.NewMetrics(metrics.NewRegistry(), exchange.AdapterList())
```
hhhjort: You could also use `pbsmetrics.NewBlankMetrics()` to streamline the metrics code. Without code to actually pipe the metrics to influxdb or somewhere, it shouldn't make a real difference though. And perhaps the load of handling the full metrics code helps bring out the race conditions.

Issue #141, opened 2017-10-03
Mitigate race conditions in Adapters


dbemiller:
In this project's short lifetime, we've already started to see how easy it is to introduce race conditions (#122, #140).

This is a problem, because:
1. The project has no automated way to catch them. Only code review & comments.
2. Adapters are run in their own goroutine, so segfaults crash the entire process.

Programmers make mistakes, and anything which isn't caught automatically will happen in the future. As this project becomes more active, this will become a bigger issue.

There are a few options for how we can mitigate these issues.
1. Copy every object we send into the Adapters' goroutines, so that data races aren't possible.
2. Add "good adapter citizen" tests which run on _all_ adapters, and make sure they don't mutate shared state.
3. Change the Types we send into `Adapter.Call()` so that adapters _can't_ mutate things they're not allowed to. For example, make the PBSRequest state private and expose methods which only return immutable data.

Anyone have preferences, or other ideas?

mkendall07:
I like option 3. Although 2 is probably also a good idea. 1 seems a bit heavy handed. That's all I got. no new ideas.

dbemiller:
There's also some overlap here with #150.

If the PBS external API changes to accept openRTB requests, then the `pbsrequest` type will go away, and the adapter API will _have_ to change.

dbemiller:
#356 should (I hope) make Travis catch many of these. If we still find that we're running into problems, we can look into ways to improve the design farther.

PR #356, opened 2018-02-21
More maintainable race condition tests

dbemiller:
This is a followup from #355, and another partial (but slightly more thorough) fix for #141.

The `{bidder}test/params/race/{mediaType}.json` files may also be useful as "sample bidder params" someday, but... the real goal here is just to make it easy for new adapters to become part of the race condition test.

The reason I split those files by `mediaType` is because I saw that Rubicon's params had a `video` property with a bunch more fields in it. That seemed reasonable enough to me... so I assumed some other bidders might be in a similar situation.

`-count 2` is pretty arbitrary. It's the smallest number which, when I test, reproduced the error fixed in #355 reliably.

The `config.yaml` refactor is because the `config_test.yaml` tests themselves had race conditions, which caused validation failures, and this was the fastest way to fix them. Global, mutable variables are evil, we should work to purge them from the project, yada, yada...


dbemiller:
Actually... let's hold off on this for a bit. I thought of a fairly significant improvement, which I think is worth making sooner rather than later.

dbemiller:
on third thought, never mind. I chatted with @hhhjort a bit and we decided against that alternate design--at least for now.

This can get a second review whenever.

validate.sh
```diff
@@ -40,10 +40,15 @@ else
...
+# Run the actual tests. Make sure there's enough coverage too, if the flags call for it.
 if $COVERAGE; then
   ./scripts/check_coverage.sh
+
+  # Tests take a *really* long time with both the race detector and code coverage, so...
+  # just re-run the tests designed for concurrency again, without code coverage.
+  go test -race $(go list ./... | grep -v /vendor/) -run ^TestRace.*$ -count 2
 else
-  go test $(go list ./... | grep -v /vendor/)
+  go test -race $(go list ./... | grep -v /vendor/) -count 2
```
cindyhere33: As per our discussion offline, let's see if we can increase the `-- count` value as long as it does not introduce unwanted issues and workarounds. Your idea of using a flag to allow/disallow a race command that has a huge `-- count` number (to enable quick execution) is - superb! Trust you to do what's best!

dbemiller: I added a `--race ###` flag so the caller can configure it, and set `travis.yml` to run it 10 times.

I also limited the race detector to only run on tests which target it because I think that filesystem issue I ran into at high values of `--race` is more likely to cause confusion than anything else.

validate.sh
```diff
@@ -40,10 +42,18 @@ else
   test $GOFMT_LINES -eq 0 || die "gofmt needs to be run, ${GOFMT_LINES} files have issues.  Below is a list of files to review:\n`gofmt -l *.go pbs adapters`"
 fi
 
+# Run the actual tests. Make sure there's enough coverage too, if the flags call for it.
 if $COVERAGE; then
   ./scripts/check_coverage.sh
+
+  if [ "$RACE" -ne "0" ]; then
```
cindyhere33: Does the $RACE if condition need to be inside the $COVERAGE if body? It is executed either way.

dbemiller: derp.... no. no it doesn't. Just sloppy of me.

Cleaning up now.

PR #358, opened 2018-02-22
Added site to the sample payload

dbemiller:
This is a required param... will fix the sample payload.


hhhjort:
OpenRTB 2.5 seems to claim it is just recommended for web pages (and not applicable for apps). But it is probably practically required for web bids (who wants to bid on an unknown website ... may be possible that a bidder knows a user and doesn't care where he is.

dbemiller:
It is optional in OpenRTB... but since it was recommended, I made it required in PBS.

I'm open to making it optional, but... it's better to be strict rather than loose, until enough people complain.

hhhjort:
Sounds good to me.

PR #361, opened 2018-02-23
Delete Adapter.Name()

dbemiller:
I just noticed that no code actually calls these functions, or reads these values, so... they're dead code and shouldn't exist.

The BidderName and `exchange` map keys are the real source of truth... so I added some comments to make that clear.

### DIFF
--- a/adapters/adform/adform.go
+++ b/adapters/adform/adform.go
@@ -58,7 +58,7 @@ func NewAdformAdapter(config *adapters.HTTPAdapterConfig, endpointURL string) *A
 }
 
 // used for cookies and such
-func (a *AdformAdapter) FamilyName() string {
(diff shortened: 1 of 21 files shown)


hhhjort:
Its a good direction for cleanup. I just wonder if perhaps we should remove `FamilyName` and replace it with `Name` wherever it is used. I think the original intention of a separate `FamilyName` has been replaced with aliases, so the "Family" part is now redundant, and may lead to confusion.

dbemiller:
akh... sorry. I got confused about which PR this was, and merged it prematurely. @hhhjort let me know if there were any other changes you wanted and I can make a new one.

hhhjort:
No, that was the only change I found to recommend, I think we are good.

PR #367, opened 2018-02-26
Adapter metadata endpoints

### LINKED ISSUE #173: Document current status of demand adapters
All of the demand adapters support different media-types (banner, video, native) as well as supply types (web, mobile). This should definitely be tracked somewhere so that it is easy for developers to know without digging through the code what an adapter supports.

@dbemiller is thinking to do this through APIs.

Might need to add a GET /bidders endpoint which returns something simpler, like:

[
  "appnexus",
  "rubicon",
  ...
]
And maybe a /bidders/media-types which returns something like:

{
  "appnexus": ["banner", "video"],
  "index": ["banner"]
}
Basic motivation is: docs fall out of date all the time. Since we're running a server anyway, we should be able to generate these things from the code, so they're guaranteed to be up to date.

It'll also help with the fact that different companies might deploy different versions of PBS... so the features might be different depending on whose deploy you're using.

dbemiller:
This fixes #173.

I had to move `info.yaml` files into the `static/` directory because of the `Dockerfile` mechanics. Only files inside `static/` end up inside the container, so this the only way to make them accessible at runtime.

Please pay special attention to the API endpoints themselves. If you see a possible use-case in the future that can't be added without breaking backwards compatibly, now's a really good time to bring it up.


endpoints/info/bidders.go
```diff
@@ -0,0 +1,92 @@
...
+		if _, err := w.Write(jsonData); err != nil {
+			glog.Errorf("error writing response to /info/bidders: %v", err)
+		}
+	})
+}
+
+// NewBiddersEndpoint implements /info/bidders/*
+func NewBidderDetailsEndpoint(infoDir string) httprouter.Handle {
+	// Build all the responses up front, since there are a finite number and it won't use much memory.
+	files, err := ioutil.ReadDir(infoDir)
```
hhhjort: Would it be better to use the same list of bidders generated above, and panic if the file for a bidder is missing? This assures that we will have info for all supported bidders, and not bother reading random files that somehow got into the directory.

endpoints/info/bidders.go
```diff
@@ -0,0 +1,92 @@
...
+				glog.Errorf("error writing response to /info/bidders/%s: %v", forBidder, err)
+			}
+		} else {
+			w.WriteHeader(http.StatusNotFound)
+		}
+	})
+}
+
+type infoFile struct {
+	Maintainer   *maintainerInfo   `yaml:"maintainer" json:"maintainer"`
```
hhhjort: I suppose translating from `yaml` to `json` insures that what we have internally as the info matches what we respond with. Is there any other reason that we are storing the files as `yaml`?

dbemiller: hmm... that's a good question.

The real reason I used `yaml` is just because it's easier on the eyes. I've seen our services people ask things like "which bidders are supported?" and "does XYZ support video?" over and over again... so I was gonna point them to these files until we can generate some real user-facing documentation.

endpoints/info/bidders.go
dbemiller: This is a good idea.... I'll change this shortly.

dbemiller: Actually... one major problem with this is that our unit tests can't load the yaml files from the `sample` directory anymore. That makes the tests a bit less stable, since they're relying on prod code which could change at any time.

Maybe a better idea is to just update `TestInfoFiles` to make sure that that directory _only_ has files for known bidders in it?

hhhjort: Probably best to fix up `TestInfoFiles`. Either add the production `yaml` file to the `sample` directory, or just have a default info file for all bidders you don't have specific tests for (you can then bulk test the rest if you want).

dbemiller: oh... I misread this, and implemented something a bit different. I think this way is better, though.

I just added an assertion to check that `numFiles == len(bidderMap)`. If any extra files get added into the prod folder, that check will fail.

I like this better because it has a lower maintenance cost. We won't have to add a dummyfile to the test directory every time there's a new Bidder this way.

dbemiller: Refactored some things as we discussed, trying to get the best balance of all the benefits.

The `BidderList` is now passed into the endpoint, so the tests can use the mock directory. The implementation now relies on the `BidderMap`, rather than the directory file contents, which should make it a bit more robust if people add or remove files between test time and deployment time.

endpoints/info/bidders.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// NewBiddersEndpoint implements /info/bidders
+func NewBiddersEndpoint() httprouter.Handle {
+	bidderNames := make([]string, 0, len(openrtb_ext.BidderMap))
+	for bidderName := range openrtb_ext.BidderMap {
+		bidderNames = append(bidderNames, bidderName)
+	}
+
+	jsonData, err := json.Marshal(bidderNames)
```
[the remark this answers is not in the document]
dbemiller: Fine by me ^^

PR #368, opened 2018-02-28
Fixed #315

### LINKED ISSUE #315: Travis builds don't show error messages
The `./validate --nofmt --cov` command doesn't print out the reason for failure--at least in some cases. This makes failed CI hard to debug if it doesn't reproduce locally.

To reproduce: Add `glog.Fatalf("Fail!")` to the start of `init()` in `pbs_light.go`, and then run `./validate --nofmt --cov`.

`echo $?` shows that it exits with a failure status code, but no information about the test failure is printed to the terminal.

dbemiller:
This fixes #315

Looks like this was caused by two things:

- `set -e`: makes the script exit immediately if any line errors
- `OUTPUT=./scripts/coverage.sh`: Saves `stdout` of `coverage.sh` to `OUTPUT` (rather than printing it to the terminal). If `coverage.sh` fails, then this line fails too... which triggers `set -e` so that the script exits without printing anything.

I basically just removed `set -e`, and checked the return from `coverage.sh` explicitly.

### DIFF
--- a/scripts/check_coverage.sh
+++ b/scripts/check_coverage.sh
@@ -25,8 +25,6 @@ if [ "$?" = "0" ]; then
 
 # Fixes #315. If it has errors, print those and exit with an error code
 else
-  while read -r LINE; do
-    echo -e "$LINE"
-  done <<< "$OUTPUT"
+  echo -e "$OUTPUT"
   exit 1
 fi
--- a/scripts/check_coverage.sh
+++ b/scripts/check_coverage.sh
@@ -25,6 +25,8 @@ if [ "$?" = "0" ]; then
 
 # Fixes #315. If it has errors, print those and exit with an error code
 else
-  echo $OUTPUT
+  while read -r LINE; do
+    echo -e "$LINE"
+  done <<< "$OUTPUT"
   exit 1
 fi
--- a/scripts/check_coverage.sh
+++ b/scripts/check_coverage.sh
@@ -1,12 +1,7 @@
 #!/bin/bash
 
-set -e
-
-CHECKCOV=false
-
 #cleanup
 finish() {
-
   if [ -d ".cover" ]; then
     rm -rf .cover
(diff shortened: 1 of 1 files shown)


dbemiller:
I rewrote this based on some suggestions from @jsnellbaker. Much more elegant this way.

I also noticed that the code coverage min message had fallen out of sync with the actual required value, so... I moved that into a variable.

scripts/check_coverage.sh
```diff
@@ -17,17 +16,13 @@ trap finish EXIT ERR INT TERM
...
 
-while read -r LINE; do
+while IFS= read -r LINE; do
   echo -e "$LINE"
   if [[ $LINE =~ "%" ]]; then
     PERCENT=$(echo "$LINE"|cut -d: -f2-|cut -d% -f1|cut -d. -f1|tr -d ' ')
-    if [[ $PERCENT -lt 30 ]]; then
-      CHECKCOV=true
+    if [[ $PERCENT -lt $COV_MIN ]]; then
+      echo "Package has less than $COV_MIN% code coverage. Run ./scripts/coverage.sh --html to see a detailed coverage report, and add tests to improve your coverage"
```
jsnellbaker: If you move this message+exit inside the loop, it'll only report on the first package that falls below the coverage threshold.  It won't initially indicate that other packages are below, you'd only see those once you fix the coverage for this first package.

Is that acceptable?  

Also, I think that the `$COV_MIN%` would render the text as you want (eg 30%), but normally the variables are encased like this `${COV_MIN}%` to mesh with the rest of the string.

dbemiller: yeah. I think this is at least better than what it was. Probably the ideal state is to store all the packages which need more coverage, and summarize them at the end, but... I think that's a little out of scope for this PR.

I'll fix `${COV_MIN}%`... that's a good point.

PR #373, opened 2018-03-01
Adding GDPR extensions

### LINKED ISSUE #371: GDPR - attempting to pass data into user.ext.consent returns error
**Context/Problem** 

As part of the GDPR spec for prebid.js, I need to pass an encoded user string into the pbs request object through the field `user.ext.consent`.  Currently, when I'm attempting to pass a value in this field, pbs is returning the following error:
`Invalid request format: request.user.ext should not be an empty object.`

When I looked into the code, I traced this error back to `auction.go:400-403`:
```
if userExt.DigiTrust == nil {
				// Make sure that user.ext is not empty.
				if userExt.Prebid == nil {
					return errors.New("request.user.ext should not be an empty object.")
```

It seems like only two potential values are accepted in the `user.ext` field; `Prebid` and `DigiTrust`.  

**Request**
Can we alter the validation logic and/or add support for the `user.ext.consent` field as well?

**Additional info**
For reference, below is a copy of my request object that erred out:
```
{"id":"660b2341-50d2-45bb-9594-82abd8ae1b50","site":{"publisher":{"id":"a3de7af2-a86a-4043-a77b-c7e86744155e"}},"source":{"tid":"660b2341-50d2-45bb-9594-82abd8ae1b50"},"tmax":1000,"imp":[{"id":"/19968336/header-bid-tag-0","ext":{"appnexus":{"placementId":10433394}},"banner":{"format":[{"w":300,"h":250},{"w":300,"h":300}]}},{"id":"/19968336/header-bid-tag1","ext":{"rubicon":{"accountId":1001,"siteId":113932,"zoneId":535510}},"banner":{"format":[{"w":300,"h":250}]}}],"test":1,"user":{"ext":{"consent":"BOJ5g/KOJ5g/KABAB8AAAAAZ+A=="}}}
```

dbemiller:
Fixes #371.

I tested this manually... but I'm going lighter than usual on the automated tests because they'll cause merge conflict with #353. When this looks approvable, let me know and I'll add the test cases over in that PR.

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -420,7 +420,7 @@ func validateUser(user *openrtb.User, aliases map[string]string) error {
 			}
 		} else {
 			// Return error.
-			return errors.New("request.user.ext object is not valid.")
+			return fmt.Errorf("request.user.ext object is not valid: %v", err)
 		}
 	}
 
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -397,12 +397,7 @@ func validateUser(user *openrtb.User, aliases map[string]string) error {
 		// Creating ExtUser object to check if DigiTrust is valid
 		var userExt openrtb_ext.ExtUser
 		if err := json.Unmarshal(user.Ext, &userExt); err == nil {
-			if userExt.DigiTrust == nil {
-				// Make sure that user.ext is not empty.
-				if userExt.Prebid == nil {
-					return errors.New("request.user.ext should not be an empty object.")
-				}
-			} else if userExt.DigiTrust.Pref != 0 {
+			if userExt.DigiTrust != nil && userExt.DigiTrust.Pref != 0 {
 				// DigiTrust is not valid. Return error.
 				return errors.New("request.user contains a digitrust object that is not valid.")
 			}
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -223,6 +223,10 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) error {
 		return err
 	}
 
+	if err := validateRegs(req.Regs); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -421,13 +420,26 @@ func validateUser(user *openrtb.User, aliases map[string]string) error {
...
 }
 
+func validateRegs(regs *openrtb.Regs) error {
+	if regs != nil && len(regs.Ext) > 0 {
+		var regsExt openrtb_ext.ExtRegs
+		if err := json.Unmarshal(regs.Ext, &regsExt); err != nil {
+			return fmt.Errorf("request.regs.ext is invalid: %v", err)
+		}
+		if regsExt.GDPR != nil && (*regsExt.GDPR < 0 || *regsExt.GDPR > 1) {
+			return errors.New("request.regs.ext.gdpr must be either 0 or 1.")
```
jsnellbaker: Should this be `errors.New`?  I saw you change other error messages to `fmt.Errorf`.

openrtb_ext/user.go
```diff
@@ -2,6 +2,11 @@ package openrtb_ext
 
 // ExtUser defines the contract for bidrequest.user.ext
 type ExtUser struct {
+
+	// Consent is a GDPR consent string. See "Advised Extensions" of
+	// https://iabtechlab.com/wp-content/uploads/2018/02/OpenRTB_Advisory_GDPR_2018-02.pdf
+	Consent string `json:"consent"`
```
jsnellbaker: Just to confirm, there is a use-case where the lookup process to grab a person's GDPR consent ID string could fail and the publisher would still want the auction to proceed without the ID.

In this case, we're going to set the consent ID to `undefined` and pass it along.  Would the code as is support this type of value?

endpoints/openrtb2/auction.go
dbemiller: Yeah. `fmt.Errorf` lets you use formatting strings like `request.imp[%d], 3`. `errors.New` just accepts raw strings.

No formatting strings in this one, so.... `errors.New`

openrtb_ext/user.go
dbemiller: Let's chat about this tomorrow. I need to get a better feel for the use-cases.

As written, Go will fill this out with an empty string if you pass it undefined. It'll pass this onto each Bidder, and they can do whatever they want with it.

I haven't done the GDPR research... so I need to understand what "failed lookup" means, legally & semantically.

hhhjort: Won't it pass the original `user.ext` to the bidder, which will leave it undefined as submitted? I don't see where we are rewriting it here, and can't think of why we would rewrite it elsewhere.

dbemiller: I think he means the javascript `undefined`, and not the string `"undefined"`.

JSON doesn't have an `undefined`, so I _think_ the browser would omit that key in the object it sends. Go would `Unmarshal` that as the zero object, which is an empty string.

Not sure, though... let's iron out the details tomorrow.

Issue #376, opened 2018-03-02
Prebid Server OpenRTB bid requests for AMP


Xoben:
Anyone know how to setup an amp config on prebid server?
When I test the placement via this link https://prebid.adnxs.com/pbs/v1/openrtb2/amp?tag_id=11111

I get the following error message

Invalid request format: AMP requests require Targeting and Caching to be set
`{ "imp": [ { "ext": { "appnexus": { "placementId": 11111 } }, "banner": { "format": [ { "h": 250, "w": 300 }, { "h": 600, "w": 300 } ] }, "id": "imp-div-gpt-300x250_0" } ], "id": "div-gpt-300x250_0" }`

I can't seem to find any documentation on how to setup OpenRTB bid requests for AMP

https://prebid.adnxs.com/amp/

mkendall07:
are you hosting your own prebid server?
The error I get is:
`Invalid request format: Stored Request with ID="11111" not found.`

Xoben:
@mkendall07 sorry the correct link is https://prebid.adnxs.com/pbs/v1/openrtb2/amp?tag_id=9114174

dbemiller:
ahh, yes. It's exactly as the error message says: `Targeting` and `Cache` need to be set for AMP to work at all. Docs on Targeting are [here](https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/auction.md#targeting). It looks like docs on `cache` don't exist.

It also looks like you're missing a `site` argument, which OpenRTB requests require.

Try something like this instead:

```
{
  "id": "div-gpt-300x250_0",
  "site": {
	"page": "your-page.com"
  },
  "imp": [
    {
      "ext": {
        "appnexus": {
          "placementId": 9114174
        }
      },
      "banner": {
        "format": [
          {
            "h": 250,
            "w": 300
          },
          {
            "h": 600,
            "w": 300
          }
        ]
      },
      "id": "imp-div-gpt-300x250_0"
    }
  ],
  "ext": {
    "prebid": {
      "targeting": {
        "pricegranularity": "you must define this as {low, med, high, ...} -- depending on your DFP Ad Units"
      },
      "cache": {
        "bids": {}
      }
    }
  }
}
```

dbemiller:
One more caveat: `prebid.adnxs.com` caches these payloads... so if you change it, it might take up to an hour before the cache clears.

[...]

PR #377, opened 2018-03-02
Better AMP error message

dbemiller:
Because of the confusion in #376.


cindyhere33:
+1 for this! would be great if this goes into the docs/endpoints/openrtb2/amp.md as well

dbemiller:
@cindyhere33 I agree, but... wouldn't delay this change because of it. I'm also knee deep in fixing another bug I found recently with Targeting logic, which (since it's used in AMP) is more important than the docs overall.

cindyhere33:
@dbemiller no problemo, I'll put it on my list for when I get a little free time

PR #380, opened 2018-03-05
Fixes #378

### LINKED ISSUE #378: Targeting params not being set properly
I stumbled on a bug with the Targeting params. Basically the `hb_param_{bidder}` keys are being applied to _all_ bids. Instead they should only be applied to the top Bid for each Bidder in each Imp.

Better unit tests to demonstrate the bug coming soon. Fix will be coming shortly after.

dbemiller:
The core goal here is to fix #378. It's pretty high priority, because it breaks AMP in many cases.

The bulk of these changes refactor the targeting so that all the keys are set inside the exchange (sorry @hhhjort... Nicole & my advice on that turned out to be really bad).

The main issue here is the `hb_cache_id_{bidder}` key, which (to our credit, I suppose) got added after that initial PR. Since we want to save all the necessary bids in a single HTTP call, it's not actually possible for the `Bidder` to populate it on its own. This made the targeting logic scattered and convoluted without any clear distinction, so... not surprising that some bugs crept in.

This consolidates all the targeting keys back into a single spot in the auction logic.


cirla:
This really simplifies a lot of things! I like not passing the `auction` and `targetData` structs around and handling that logic **after** the call to `getAllBids` on the `exchange`, though I don't think it's necessary to kill auction.go and move that logic all into targeting.go, but maybe do something like:
```go
// in `exchange.HoldAuction`

...

adapterBids, adapterExtra := e.getAllBids(auctionCtx, cleanRequests, aliases)

// this populates `winningBids`, etc. and rounds prices inside the `auction` struct
auction := auctionFromBids(adapterBids) 

// targeting now doesn't have to worry about auction logic, cache clients, etc. and can just use 
// the fields in the `auction` struct and an `auction.cacheIds` method)
targData.setTargeting(ctx, e.cache, len(bidRequest.Imp), auction, adapterExtra)
```

dbemiller:
Yeah... I think that makes sense. Changes incoming

PR #381, opened 2018-03-05
Checked response code of test in --cov mode.

dbemiller:
Must have introduced this during that last-minute rewrite in #315. Without `set -e`, we need to check this return code explicitly.


dbemiller:
nvm @jsnellbaker :)... cirla beat you to it.

Tiny change, so... probably doesn't need a second review.

PR #384, opened 2018-03-06
Abstract the metrics support

### LINKED ISSUE #363: Abstract metrics support
Currently PBS supports recording metrics to influx with `github.com/rcrowley/go-metrics` and `github.com/vrischmann/go-metrics-influxdb`. However `go-metrics` does not support tags, limiting the details we can report on. The legacy endpoint works around this by defining a new metric for every possible tag we might want to report on. This is an ugly solution. The OpenRTB endpoint doesn't even try to report on detailed tags.

In addition to this, influx may not be the desired metrics reporting solution for all users. Being able to use something like Prometheus which does support tags would solve the first problem above. But again Prometheus is not the correct solution for everyone.

So to solve these issues, I would like to add an abstraction layer to metrics reporting, which can read the metrics backend from the config, and log to influx or prometheus (or any other backend that support is eventually written for) as desired for that particular instance of prebid server. This would be an interface attached to an object that would be backend-dependent. It would look something like:

```func (*metricsEngine) RecordMetrics(name string, data map[string]string, tags map[string][string])```

For standard meters, `data["value"]` will default to `"1"` if not supplied, and be used for setting the metric. For timers `data["startTime"]` would be the start of the interval, and `data["endTime"]` would be assumed to be "now" if not provided. The CPM histogram can pass `data["float"]` or something.

The `RecordMetrics` function would be a wrapper function for `recordMetricsXxxxx()` functions, where we would have a `meterMetrics` map that would map the name into known meter type metrics, and so on for the other metrics type. Or alternatively we would just expose the `RecordMetricsXxxx()` function directly and do away with the wrapper, as well as replacing `data[]` with the concrete parameters that that metric type needs. I am undecided which method is best.

There would be a `metricsEngine` type for each backend, `pbs_light.serve()` would call `NewMetricsEngine(cfg)` to create the interface and pass it on to all users.

In the case of the influxdb metrics engine, it may ignore the `tags[]` map (current OpenRTB endpoint), or concatenate it with `name` to fully identify the metric (legacy behavior). `RecordMetrics` should accept the full amount of detail any engine might use, and it would be up to the engine (and possibly configuration) how much detail will actually be used.

### PR DESCRIPTION
**Pull request by hhhjort**
This abstracts and standardizes the metrics support. The abstraction layer will make it easier to add new metrics databases in the future. The abstraction layer was designed with tag support in mind.

Fixes #363

### CODE REVIEW

**Comment by hhhjort** [comment]
I am working on better test coverage.

**Comment by dbemiller** [comment]
and merge conflicts, i hope :)

**Comment by hhhjort** [comment]
There were no merge conflicts when I submitted the PR ... Keeping up with all the PRs ....

**Comment by hhhjort** [comment]
Merge conflicts fixed and coverage up to 57%. I don't think we need to be as worried about coverage here, I have a dummy adapter that just does an instant `return` for all the interface methods. Need the dummy adapter in here rather than test for use when all metrics options are disabled. Then there is the multi object that simply doles out the request to a list of metrics engines, so the logic is super simple and identical except for the method name. But I am open to requests for more coverage.

File: config/config.go [hunk 1]
```diff
@@ -51,12 +51,26 @@ type Adapter struct {
...
 	Username string `mapstructure:"username"`
 	Password string `mapstructure:"password"`
 }
 
+type PrometheusMetrics struct {
+	Enabled string `mapstructure:"enabled"`
+}
+
+type GoMetrics struct {
+	Enabled string `mapstructure:"enabled"`
```
dbemiller: I'm very confused by this... it seems like `go-metrics` should be an implementation detail. It helps for _some_ backends (like `influx`), but hurts for others (like `prometheus`). Why has this bubbled all the way out to the `config`?

The only use I see of this variable is in `NewMetricsEngine`... and it looks very suspicious:

> // Seperate check in case we find another metrics library we want to hook to InfluxDB
> // If we want to support tagging, then we will need something other than the legacy goMetrics

Seems much simpler, more flexible, and less error-prone if the `config` just offers backends, and the code gets to decide when `go-metrics` is appropriate.

pbsmetrics/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+// RecordUserIDSet across all engines
+func (me *MultiMetricsEngine) RecordUserIDSet(userLabels UserLabels) {
+	for _, thisME := range me.engineList {
+		thisME.RecordUserIDSet(userLabels)
+	}
+}
+
+// DummyMetricsEngine is a Noop metrics engine in case no metrics are configured. (may also be useful for tests)
+type DummyMetricsEngine struct {
+	dummy int
```
dbemiller: You don't need to do this... I don't think? Empty structs should be allowed.

pbsmetrics/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+// UserLabels : Labels for /setuid endpoint
+type UserLabels struct {
+	Action RequestAction
+	Bidder openrtb_ext.BidderName
+}
+
+// RequestAction : The setuid request result
+type RequestAction string
```
dbemiller: Maybe an opinion, but... do you think these work well as labels, or separate methods on the interface?

With the exception of this one cluster of constants, it seems to me like all the `MetricsEngine` methods are different Events, and all the `Labels` are little packages of cross-cutting concerns.

I don't think it's a big deal either way... just wanted to point it out cause I find it a _bit_ more intuitive that way, so I'm fishing to see if you shared the uncertainty.

pbsmetrics/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+
+// Browser flag; at this point we only care about identifying Safari
+const (
+	BrowserSafari Browser = "safari"
+	BrowserOther  Browser = "other"
+)
+
+// Cookie flag
+const (
+	CookieFlagYes     CookieFlag = "2"
```
dbemiller: Is there any reason these are numeric strings? If not, let's name them something real & readable.

`yes`, `no`, and `unknown` would do just fine... but since cookies have expiration dates, I might recommend renaming `yes` to `exists`. Just so that someday, if we find it useful, we can scale to `exists`, `expired`, `no`, and `unknown`.

(There may also be a better name for "no", but... I can't think of one offhand)

pbsmetrics/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+	CookieFlagNo      CookieFlag = "1"
+	CookieFlagUnknown CookieFlag = "0"
+)
+
+// Request/return status
+const (
+	RequestStatusOK      RequestStatus = "ok"
+	RequestStatusErr     RequestStatus = "err"
+	RequestStatusNoBid   RequestStatus = "nobid"   // Only for adapters
+	RequestStatusTimeout RequestStatus = "timeout" // Only for adapters
```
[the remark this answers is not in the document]
hhhjort: I was a little bit split on this. Seems it would be good to limit the number of label structs I create, it reduces clutter for one. In addition, this allows taking the general request struct and simply set the adapter to create the adapter label struct, instantly inheriting all the settings such as Safari and Cookie without any work.

pbsmetrics/metrics.go
hhhjort: I do believe these are all the same incoming event (`/setuid` request), but just enumerate how PBS responds to the request. We can refuse to do anything because of opt out, we can error on an invalid request, or we can execute a valid request.

pbsmetrics/metrics.go
hhhjort: Agreed.

pbsmetrics/metrics.go
```diff
@@ -0,0 +1,257 @@
...
+
+// MetricsEngine is a generic interface to record PBS metrics into the desired backend
+// The first three metrics function fire off once per incoming request, so total metrics
+// will equal the total numer of incoming requests. The remaining 5 fire off per outgoing
+// request to a bidder adapter, so will record a number of hits per incoming request. The
+// two groups should be consistent within themselves, but comparing numbers between groups
+// is generally not useful.
+type MetricsEngine interface {
+	RecordRequest(labels Labels)                    // ignores adapter. only statusOk and statusErr fom status
+	RecordTime(labels Labels, length time.Duration) // ignores adapter. only statusOk and statusErr fom status
```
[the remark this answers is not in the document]
hhhjort: Probably a better name ... think it came from the legacy name being as simple and uninformative.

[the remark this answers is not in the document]
hhhjort: Bonus plus, in changing the name, I discover that the OpenRTB endpoints are not recording request times. Fixing now. :)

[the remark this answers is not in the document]
hhhjort: And on review, I cannot defer the `RecordRequest(labels)` as `labels` is evaluated at time of the defer statement, not at the time the function is actually executed at the end of the function.

exchange/exchange.go
```diff
@@ -142,34 +145,34 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
...
+			e.me.RecordAdapterTime(*blabels[coreBidder], time.Since(start))
 			serr := make([]string, len(err))
 			for i := 0; i < len(err); i++ {
 				serr[i] = err[i].Error()
 				// TODO: #142: for a bidder that return multiple errors, we will log multiple errors for that request
 				// in the metrics. Need to remember that in analyzing the data.
 				switch err[i] {
 				case context.DeadlineExceeded:
-					e.m.AdapterMetrics[coreBidder].TimeoutMeter.Mark(1)
+					blabels[aName].RequestStatus = pbsmetrics.RequestStatusTimeout
```
dbemiller: `aName` may be an alias. I think these should log under `coreBidder` so we don't explode the number of metrics we're storing.

glide.lock
```diff
@@ -54,12 +54,14 @@ imports:
...
   version: 3f56f5d84c99cf28c575a90ffddbf9f0337c39bd
+  subpackages:
+  - native/request
 - name: github.com/pelletier/go-toml
   version: 1d6b12b7cb290426e27e6b4e38b89fcda3aeef03
 - name: github.com/rcrowley/go-metrics
   version: 1f30fe9094a513ce4c700b9a54458bbb0c96996c
 - name: github.com/rs/cors
-  version: 7af7a1e09ba336d2ea14b1ce73bf693c6837dbf6
+  version: feef513b9575b32f84bafa580aad89b011259019
```
dbemiller: Did you intend to change this? It looks a little out of place in this PR.

pbsmetrics/metrics.go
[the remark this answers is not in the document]
dbemiller: Ok. I'm not super convinced, but... I'm not sure of myself to be too stubborn about it either.

In general each it's better if each "thing" in the code only means one thing.

exchange/exchange.go
hhhjort: `blabels[aName]` is a map to a label, the bidder in the lable doesn't have to be `aName` I tired to be sure I used `coreBidder` when setting the bidder name inside of the label itself, but I will review all that code and make sure I only send corebidders as bidder flags.

hhhjort: `blabels[aName]` is a bug, it should be `blabels[coreBidder]` I was correct that I always set the bidder name inside the labels as the core bidder, not sure how I ended up with some map references as `aName` instead.

config/config.go
```diff
@@ -51,12 +51,21 @@ type Adapter struct {
...
+
+type InfluxMetrics struct {
 	Host     string `mapstructure:"host"`
 	Database string `mapstructure:"database"`
 	Username string `mapstructure:"username"`
 	Password string `mapstructure:"password"`
 }
 
+type PrometheusMetrics struct {
+	Enabled string `mapstructure:"enabled"`
```
dbemiller: What is this `enabled` string? Should it be a `bool`?

Why does it even exist? Wouldn't you just omit this propertyif you didn't want to use prometheus?

glide.lock
```diff
@@ -0,0 +1,131 @@
+hash: df48123f9cd400234264795207b4dca1f90b46ee3b919e5162c1cbc057e93da4
```
dbemiller: Looks like this slipped in accidentally. `glide.lock` should have been removed during the upgrade to `dep`. You'll have to remove it here.

pbs_light.go
```diff
@@ -793,24 +797,6 @@ func main() {
 	}
 }
 
-func setupExchanges(cfg *config.Configuration) {
-	exchanges = newExchangeMap(cfg)
```
dbemiller: Now that these have been removed, there's a _lot_ of dead code which can be deleted from this file.

Start out with `makeExchangeMetrics` and try out the "find all references" feature in Visual Studio Code. You'll find a bunch of stuff which you can get rid of.

pbsmetrics/go_metrics.go
```diff
@@ -121,6 +151,186 @@ func registerAdapterMetrics(registry metrics.Registry, adapterOrAccount string,
...
+
+// RecordAdapterRequest implements a part of the MetricsEngine interface
+func (me *Metrics) RecordAdapterRequest(labels Labels) {
+	_, ok := me.AdapterMetrics[labels.Adapter]
+	if !ok {
+		glog.Errorf("Trying to run adapter metrics on %s: adapter metrics not found", string(labels.Adapter))
+		return
+	}
+	// Adapter metrics
+	am := me.AdapterMetrics[labels.Adapter]
```
dbemiller: Tiny simplification: you could move this to `am, ok := me.AdapterMetrics[labels.Adapter]` and save a map lookup.

Same goes for some other functions in this file.

pbsmetrics/go_metrics.go
```diff
@@ -121,6 +151,186 @@ func registerAdapterMetrics(registry metrics.Registry, adapterOrAccount string,
...
+	aam := me.getAccountMetrics(labels.PubID).adapterMetrics[labels.Adapter]
+	aam.RequestMeter.Mark(1)
+
+	switch labels.RequestStatus {
+	case RequestStatusErr:
+		am.ErrorMeter.Mark(1)
+	case RequestStatusNoBid:
+		am.NoBidMeter.Mark(1)
+	case RequestStatusTimeout:
+		am.TimeoutMeter.Mark(1)
```
dbemiller: Should we log an error in the `default` here? (and elsewhere)

pbsmetrics/metrics.go
dbemiller: oh! Gotcha. For some reason I thought it was logging an event each time a user chose to opt out.

This makes sense though.

config/config.go
hhhjort: Because this is the only element of prometheus. Prometheus uses a pull structure, so we don't need to configure for a particular Promethus instance. But bool is a better choice, yes.

pbs_light.go
hhhjort: Also ripped out the old metrics globals which uncovered a couple of other things. `makeExchangeMetrics` was only used twice however.

pbsmetrics/go_metrics.go
hhhjort: Good catch, thanks.

pbsmetrics/go_metrics.go
hhhjort: Given the previous behavior that we only log an error if we explicitly found an error, I felt this was the best match to the legacy metrics behavior. If there were no bugs, the two cases should be identical. Also if we add a default, then we need a case for `RequestStatusOk`

pbsmetrics/metrics.go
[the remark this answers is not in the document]
hhhjort: Not super convinced of your side either, but decided to go ahead and flip.

config/config.go
dbemiller: Actually... I just realized, this isn't even used anywhere. It's just dead code.

Let's remove it until the PR with Prometheus support.

pbsmetrics/go_metrics.go
dbemiller: ah, ok. yeah don't worry about it... not a big deal either way.

exchange/utils.go
```diff
@@ -40,9 +40,20 @@ func splitBidRequest(req *openrtb.BidRequest, impsByBidder map[string][]openrtb.
 	for bidder, imps := range impsByBidder {
 		reqCopy := *req
 		coreBidder := resolveBidder(bidder, aliases)
-		met.AdapterMetrics[coreBidder].RequestMeter.Mark(1)
+		newLabel := pbsmetrics.AdapterLabels{
```
dbemiller: This might be a good use-case for embedded structs: https://travix.io/type-embedding-in-go-ba40dd4264df

I've never used these before, so I don't know how well they work in practice... but I _think_ it would let you write this in such a way that `AdapterLabels` picks up all the changes from `Labels` automatically.

Might make this a bit easier to maintain, if you always expect `AdapterLabels` to be a superset of `Labels`.

Not something I'd insist on... or even something that I'm confident would work out better. But I wanted to throw out the possibility in case you hadn't heard of it yet, so you could play around with it.

hhhjort: Actually I don't have any use cases in mind to expand the labels. If we want to expand them later, them we can look into type embedding.

dbemiller: ok, sounds good.

PR #386, opened 2018-03-07
Extend debug capabilities of OpenRTB2 endpoints

cirla:
1. Accept `debug` query param in AMP endpoint which sets `"test": 1` on the OpenRTP bid request
2. Show "resolved request" (after resolving stored requests and debug overrides) in debug output for `/openrtb2/auction` and `/openrtb2/amp` endpoints


endpoints/openrtb2/amp_auction.go
```diff
@@ -182,9 +182,13 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
 		return
 	}
 
+	debugParam, ok := httpRequest.URL.Query()["debug"]
+	debug := ok && len(debugParam) > 0 && debugParam[0] == "1"
```
cirla: This could be extracted into common code to check for boolean presence of a query param. We may also want to treat an empty value (e.g. `?debug`) as true.

dbemiller: I think it's ok as is.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -50,6 +50,8 @@ func TestGoodAmpRequests(t *testing.T) {
 	}
 }
 
+// TODO: (@cirla) test `debug=1` query param
```
dbemiller: Did you plan to do this one? Same with the other tests.

If not, I think you can just remove the TODO unless you open issues for them... in my experience, untracked TODOs generally do more harm than good.

exchange/exchange.go
```diff
@@ -208,27 +208,28 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
...
 	bidResponseExt := &openrtb_ext.ExtBidResponse{
 		Errors:             make(map[openrtb_ext.BidderName][]string, len(adapterBids)),
 		ResponseTimeMillis: make(map[openrtb_ext.BidderName]int, len(adapterBids)),
 	}
-	if test == 1 {
+	if req.Test == 1 {
 		bidResponseExt.Debug = &openrtb_ext.ExtResponseDebug{
-			HttpCalls: make(map[openrtb_ext.BidderName][]*openrtb_ext.ExtHttpCall),
+			HttpCalls:       make(map[openrtb_ext.BidderName][]*openrtb_ext.ExtHttpCall),
+			ResolvedRequest: req,
```
dbemiller: hmmm...... the only thing I'm a bit worried about, here, is that the same `BidRequest` instance was passed around as a pointer through a _lot_ of code before it got here. I think you've definitely done the most elegant thing, here... but I'm a little bit worried that it'll get mutated somewhere during the auction logic and end up returning a bad result here.

Although it's uglier, it'd probably be safer to `json.Marshal` the object as soon as you get it, and pass it through here as a `json.RawMessage`. You could also make a deep copy of the request struct, but... that's a lot of performance overhead, and you'd need to add a new project dependency (since deep clone isn't in Go's core libs), so... not sure it's worth all that trouble.

docs/endpoints/openrtb2/auction.md
```diff
@@ -243,6 +243,10 @@ However, the publisher can improve performance by only offering impressions whic
 This contains info about every request and response sent by the bidder to its server.
 It is only returned on `test` bids for performance reasons, but may be useful during debugging.
 
+`response.ext.debug.resolvedrequest` will be populated **only if** `request.test` **was set to 1**.
+
+This contains the request after the resolution of stored requests and debug overrides.
```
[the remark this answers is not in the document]
cirla: I meant overriding `test` to `1` via the `debug` parameter, which only applies to the AMP endpoint, so I should just leave it at
>This contains the request after the resolution of stored requests

to avoid confusion.

endpoints/openrtb2/amp_auction.go
```diff
@@ -182,9 +182,13 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
...
+
 	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(storedRequestTimeoutMillis)*time.Millisecond)
 	defer cancel()
 
+	// TODO: (@cirla) if debug is set to true, invalidate cache before fetching stored requests
```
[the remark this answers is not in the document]
cirla: I was using these TODO's mainly as placeholders while the PR is tagged "in progress" and plan on resolving them before final review/merge. Any left unresolved or determined out of scope of the PR I'll create an issue and follow your convention.

exchange/exchange.go
cirla: I think marshaling it to JSON (if `req.Test == 1`) at the beginning of `HoldAuction` (or somewhere before `getAllBids`) should be sufficient to avoid risk of mutation by adapters, and then passing that down through `buildBidResponse` and `makeExtBidResponse` to here. It seems a bit awkward having the debug value extracted so far from where it's placed in the response, though I don't see a better way at the moment.

dbemiller:
One more general question... once the auction is being run in test mode, how would someone actually see the debugging output?

Later on in the AMP handler, the ORTB response is filtered so that only the targeting params are returned: https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/amp.md#sample-response

but... in debug mode, a user would need to actually see the full response. As of now, I don't see that happening anywhere.

endpoints/openrtb2/amp_auction.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 type AmpResponse struct {
-	Targeting map[string]string `json:"targeting"`
+	Targeting map[string]string             `json:"targeting"`
+	Debug     *openrtb_ext.ExtResponseDebug `json:"debug,omitempty"`
```
cirla: This could just be of type `openrtb.RawJSON` and avoid the extraneous unmarshall->remarshall, though that only happens on a debug call and this ensures the interface.

exchange/exchange.go
```diff
@@ -57,6 +59,16 @@ func NewExchange(client *http.Client, cache prebid_cache_client.Client, cfg *con
 }
 
 func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, usersyncs IdFetcher) (*openrtb.BidResponse, error) {
+	// Snapshot of resolved bid request for debug if test request
+	var resolvedRequest openrtb.RawJSON
```
cirla: This could just be:
```go
if resolvedRequest, err := json.Marshal(bidRequest); err != nil {
    glog.Errorf("Error marshalling bid request for debug: %v", err)
}
```
but I wanted to only spend the time doing this in test requests.

exchange/exchange.go
```diff
@@ -198,27 +210,30 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
...
 	bidResponseExt := &openrtb_ext.ExtBidResponse{
 		Errors:             make(map[openrtb_ext.BidderName][]string, len(adapterBids)),
 		ResponseTimeMillis: make(map[openrtb_ext.BidderName]int, len(adapterBids)),
 	}
-	if test == 1 {
+	if req.Test == 1 {
 		bidResponseExt.Debug = &openrtb_ext.ExtResponseDebug{
 			HttpCalls: make(map[openrtb_ext.BidderName][]*openrtb_ext.ExtHttpCall),
 		}
+		if err := json.Unmarshal(resolvedRequest, &bidResponseExt.Debug.ResolvedRequest); err != nil {
```
cirla: Again, `ResolvedRequest` could just be of type `openrtb.RawJSON` (`byte[]`) and avoid the extraneous unmarshall->remarshall, though that only happens on a debug call and this ensures the interface.

exchange/exchange.go
hhhjort: I believe we can use `json.RawMessage` in place of `openrtb.RawJSON`. I have also heard that the openrtb library is planning on switching to this to have a more consistent handling of raw JSON blobs.

cirla: Sounds good. They both are just aliases for `byte[]` I believe but better to use the one in the go stdlib.

exchange/exchange.go
hhhjort: I lean towards having the `ResolvedRequest` converted to raw JSON as soon as possible, less chance of accidentally modifying something inside. But that isn't terribly important.

exchange/exchange_test.go
```diff
@@ -293,6 +293,7 @@ func TestBuildBidResponse(t *testing.T) {
 		Test: 0,
 		Ext:  bidReqExtRaw,
 	}
+	var resolvedRequest openrtb.RawJSON
```
hhhjort: One more openrtb.RawJSON we can convert.

exchange/exchange.go
dbemiller: I think there's a small bug in here.

If the `json.Marshal` call fails, then `resolvedRequest` becomes nil. If you try to `json.Unmarshal` on a `nil` slice, I'm pretty sure that Go panics.

May be worth testing, but... if I'm right about all that, then you might need a `len(resolvedRequest) > 0` check here.

dbemiller:
One other note... testing locally, I made a request for the following AMP config:

```
{
  "id": "test-imp-id",
  "site": {
    "page": "prebid.org"
  },
  "imp": [
    {
      "id": "some-imp-id",
      "banner": {
        "format": [
          {
            "w": 300,
            "h": 250
          },
          {
            "w": 300,
            "h": 600
          }
        ]
      },
      "ext": {
        "appnexus": {
          "placementId": 10433394
        }
      }
    }
  ]
}
```

and I got the following response:

```
{"targeting":{},"debug":{"resolvedrequest":{"id":"test-imp-id","imp":[{"id":"some-imp-id","banner":{"format":[{"w":300,"h":250},{"w":300,"h":600}]},"secure":1,"ext":{"appnexus":{"placementId":10433394}}}],"site":{"page":"prebid.org"},"device":{"ua":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.186 Safari/537.36","ip":"::1"},"test":1,"ext":{"prebid":{"aliases":null,"cache":{"bids":{}},"storedrequest":null,"targeting":{"pricegranularity":"medium"}}}}}}
```

For the most part that's right... but a few things are a bit ugly:

```
"aliases":null,
"storedrequest":null
```

Could you add `omitempty` to the contract classes so that those don't show up when they don't need to?

cirla:
Added `omitempty`s and a test to make sure the debug is propagated when requested. Re https://github.com/prebid/prebid-server/pull/386#discussion_r173798921, I tested a `json.Unmarshal(nil, ...)` and it doesn't panic, just returns an error value of `"unexpected end of JSON input"`.

PR #391, opened 2018-03-08
Viewability support

cindyhere33:
This PR allows bid requests to use `imp[i].metric` to send viewability scores.  

An example `bidrequest.imp` would look like: 

```
{
  "imp": [
    {
      "id": "example-imp",
       ..., 
      "metric": [
        {
          "type": "viewability",  
          "value": 0.87,        
          "vendor": "some-value" 
        }
      ],
     "ext": {
         "prebid": {
             "viewabilityvendors": ["vendor1", "vendor2"]    
          }
      }
    }
  ]
}

```

Every adapter can implement its own way of passing along viewability vendors in its exchange. This PR has it implemented for rubicon adapters.


djsu:
**Use cases**
1. Publishers can self-declare what Viewability vendors are available for a particular impression. Demand partners will be able to ingest this information and send back an ad response with the corresponding vendor tag knowing that they will be able to measure Viewability for that particular impression.
2. Publishers can self-declare some Viewability score and pass this information to demand partners.

docs/endpoints/openrtb2/auction.md
```diff
@@ -290,6 +290,12 @@ This adds two optional properties:
 
 These fields will be forwarded to each Bidder, so they can decide how to process them.
 
+
+#### Viewability
+
+`bidrequest.imp[i].metric` is now supported. `type` and `score` are mandatory fields. The viewability vendors can be sent as a string array through `bidrequest.imp[i].ext.prebid.viewabilityvendors`
```
dianhua1560: Sorry I don't see the **score** field. Did you mean **value**?

cindyhere33: oops sorry! yes, it is value.

cindyhere33: Fixed it. Thanks! 👍

dbemiller:
I had one question on this, though I'll follow up with the Product folks in the chat. According to the ORTB spec, the `type` and `vendor` strings are described as:

> Type of metric being presented using exchange curated string names which should be published to bidders _a priori_.

> Source of the value using exchange curated string names which should be published to bidders _a priori_.

So... why would we add `ext` to the Request to declare the viewability vendors? It sounds like those values should be intrinsic to PBS (in the same way that bidderNames are)

endpoints/openrtb2/auction.go
```diff
@@ -275,6 +275,22 @@ func (deps *endpointDeps) validateImp(imp *openrtb.Imp, aliases map[string]strin
 	return nil
 }
 
+func validateMetric(metrics []openrtb.Metric, impIndex int) error {
+	if len(metrics) == 0 {
+		return nil
+	}
+
+	for i, metric := range metrics {
+		if len(metric.Type) == 0 {
```
dianhua1560: is Value also mandatory?

dbemiller:
hey @cindyhere33 

> bidderNames is equivalent to viewabilityVendorMap inside the adapter. The viewabilityvendors in imp[i].ext.prebid lets us know which of those vendors are involved. This map with viewability vendors - we have it inside the rubicon adapter because I was unsure if this vendors list is global across all exchanges.

Ok... that's what it seemed like. Is there any reason you all _don't_ think the vendors list would be constant across adapters? To me, it looks like the OpenRTB spec expects it to be.

Perhaps more importantly: this API doesn't look like it would scale very well at all if different adapters recognized different viewability vendors. For example, suppose `rubicon` acknowledged the mapping:

```
"vendor-a": "some-vendor.com"
```

And the `conversant` independently acknowledged the mapping:

```
"vendor-a": "other-vendor.com"
```

Now a Publisher makes a request with some `Imp` to both `rubicon` and `conversant` with `vendor-a` inside `ext.prebid.viewabilityvendors `. Since two different adapters interpret it in different ways, which one does the viewability metric mean?

If we plan to make sure that adapters coordinate so that they don't conflict and this situation never happens, then we're committing to making a global list for the project anyway.
----------------------------------------------------------------------------
As a second, and unrelated question: why does `ext.prebid.viewabilityvendors` exist? The `metric` object has a `vendor` value, and `imp.metric` is an array. Is this just duplicate information?

It seems to me like `imp.metric` is sufficient, and we can derive the list of vendors by looping through the array and reading all the `vendor` properties. Even though the OpenRTB spec marks this field as `recommended`, I'm totally OK with requiring it.

--------------------------------------------------------------------------------
Let me know if I'm missing something here. Just by looking at this PR, and given that there's no Issue with design discussions, it _seems_ to me like we should:

1. Make the list of viewability vendors global, and stick it inside `openrtb_ext` somewhere.
2. Delete `imp.ext.prebid.viewabilityvendors`.
3. Return 400 on any request which has an `request.imp.metric.vendor` which is not in the global list of recognized ones.

Let me know what you think.

dbemiller:
One last thought: we spoke about how each adapter might want to "whitelist" vendors that their servers will understand, if we have a global vendors list.

If so, it may be a good idea to put those whitelists into the `info` files? https://github.com/prebid/prebid-server/blob/master/static/bidder-info/rubicon.yaml

That would set us up for much more maintainable docs in the future.

Just an idea, in case you like it... not something I'd insist on or anything ^^.

adapters/rubicon/rubicon.go
```diff
@@ -16,6 +16,7 @@ import (
 
 	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/endpoints/info"
```
dbemiller: This is a bad dependency... adapters shouldn't depend on endpoints.

For example, it's easy to imagine a CLI tool which can be used to test adapters, but has nothing to do with the HTTP endpoints at all. I could also imagine an Endpoint which lets you call a specific adapter, which would create a circular dependency here.

A better way would be to pull the "info file reading" code into a different module, which both `adapters` and `endpoints` could depend on. Perhaps something like `adapter_metadata`?

endpoints/info/bidders.go
```diff
@@ -68,9 +70,23 @@ func NewBidderDetailsEndpoint(infoDir string, bidders []openrtb_ext.BidderName)
 	})
 }
 
-type infoFile struct {
-	Maintainer   *maintainerInfo   `yaml:"maintainer" json:"maintainer"`
-	Capabilities *capabilitiesInfo `yaml:"capabilities" json:"capabilities"`
+//Get information from /static/bidder-info/{bidder}.yaml
+func GetBidderInfo(bidder string) *InfoFile {
```
dbemiller: Should this be an `openrtb_ext.BidderName`?

endpoints/openrtb2/auction.go
dbemiller: According to the OpenRTB spec:

> Type of metric being presented using exchange curated string names which should be published to bidders _a priori_

So... these should be a global (and documented) list too.

We should also return an error if the publisher's request uses a Type or Vendor that we don't recognize.

openrtb_ext/viewability_vendors.go
```diff
@@ -0,0 +1,22 @@
+package openrtb_ext
+
+import "fmt"
+
+var viewabilityVendorMap = map[string]string{
```
dbemiller: I think these would be better off in `static/metrics/*.yaml` files, like the adapters have. There are a bunch of advantages to this, which we can talk about if you're curious.

It seems to me like we should be able to answer the following questions using just data stored in the `yaml` files:

1. Which metric types does PBS recognize?
2. Which metric vendors does PBS recognize?
3. Which URLs correspond to each vendor?
4. Which vendors are supported by each adapter? (You're already doing this one ^^)

It seems like there should also be unit tests which ensure that:

1. Every `vendor` which an adapter says it supports is also one recognized by Prebid Server.
2. No two `vendors` have the same URL.

endpoints/info/bidders.go
```diff
@@ -8,9 +8,11 @@ import (
 	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
 	"github.com/prebid/prebid-server/openrtb_ext"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
+var responses map[string]json.RawMessage
```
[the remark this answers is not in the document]
cindyhere33: Agreed. The thought hit me about the  `NewBidderDetailsEnpdoint()` but still went for it because it is definitely called on startup and definitely throws an error on startup if anything was wrong. There is no chance that `NewBidderDetailsEndpoint` is executed and `responses` is nil. So don't think it should break.

But frankly, that was a half-hearted implementation. Appreciate your suggestion. I'll try it out.

hhhjort:
Any activity here, or should we consider closing this PR?

dbemiller:
I think it's closeable... they can open another one if they want

PR #393, opened 2018-03-08
Amp usability improvements

dbemiller:
This is part of my attempt to address mass confusion surrounding AMP.

Functionally, this sets defaults for `request.ext.prebid.cache` and `request.ext.prebid.targeting`, which should make AMP easier to use.

I also updated the docs to make it clearer what's expected and what's not, remove AMP implementation details, and link to more thorough documentation.

I will also be looking for ways to improve the `/openrtb2/auction.md` file... but I think that's better left for another PR.


dbemiller:
I know you did, and I did :). It _does_ help project maintenance to be restrictive... but the PBS/AMP community are (understandably) in a bit of a crisis at the moment.

If the AMP timeline hadn't been so short, we could have improve the UI so that these params got added automatically, but... unfortunately the timeline _is_ very short, and people are stuck with the UI we have for now.

Issue #349, opened 2018-02-15
Switch from Glide to dep


dbemiller:
According to [the Glide docs](https://github.com/Masterminds/glide#golang-dep):

> The Go community now has the dep project to manage dependencies. Please consider trying to migrate from Glide to dep. If there is an issue preventing you from migrating please file an issue with dep so the problem can be corrected. Glide will continue to be supported for some time but is considered to be in a state of support rather than active feature development.

Prebid Server should also migrate to use [dep](https://github.com/golang/dep) instead.

dbemiller:
akh.... literally _days_ after opening this, there's a blog post about an official tool: https://research.swtch.com/vgo

Community doesn't seem happy about it, though.... maybe we should see where this goes.

dbemiller:
This was fixed by #397

PR #397, opened 2018-03-08
Convert from glide to dep

cirla:
Go community is converging on dep as the dependency management tool of choice.


Gopkg.toml
```diff
@@ -0,0 +1,82 @@
...
+[[constraint]]
+  name = "github.com/coocood/freecache"
+  version = "1.0.0"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/erikstmartin/go-testdb"
+
+[[constraint]]
+  branch = "master"
```
mkendall07: should we consider locking down versions of the some of the dependencies that we are using from master?

cirla: Yes, even if just freezing them to the commit hash for whatever master currently is. We'll also need to set up Jenkins autodeploy to use dep so it doesn't break with removing glide. There's no urgency on all of this, I more just wanted to see how much work it would be to switch over and it was super simple.

dbemiller:
This looks good to me. Just so I'm aware: did you generate your `Gopkg.lock` file to use the same commits as the `glide.lock` file, or are there hidden updates in this PR too?

Either way, it's probably not a huge deal. We're probably due for some dependency version upgrades anyway.

dbemiller:
Oh, nice... ok, don't fret too much over it then. I trust their migration tool (better than human eyes at least, anyway).

Makefile
```diff
@@ -2,23 +2,23 @@
...
 	@echo "  image: build docker image"
 	@echo ""
 
 .PHONY: install deps test build image
 
-# install glide https://github.com/Masterminds/glide (assumes go is already installed)
+# install dep https://golang.github.io/dep/ (assumes go is already installed)
 install:
-	curl https://glide.sh/get | sh
+	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
dbemiller: Should we hardcode this to use whatever version the CI does?

README.md
```diff
@@ -13,7 +13,7 @@ For more information, see:
 
 ## Installation
 
-First install [Go 1.9.1](https://golang.org/doc/install) and [Glide](https://github.com/Masterminds/glide#install). Note that Glide requires an explicit GOPATH to be set.
+First install [Go 1.9.1](https://golang.org/doc/install) or later and [dep](https://golang.github.io/dep/docs/installation.html). Note that dep requires an explicit GOPATH to be set.
```
dbemiller: Git won't let me make a comment down there, but... about 10 lines below this one, there's a `glide install` command in the "download and prepare" section of the docs.

dbemiller:
Looks awesome... thanks!

dbemiller:
This fixes #349

PR #400, opened 2018-03-09
Quick fix for some segfaults in legacy code

dbemiller:
Stack trace from the problem is here:

```
panic: runtime error: index out of range
 goroutine 119248 [running]:
github.com/prebid/prebid-server/adapters/appnexus.(*AppNexusAdapter).Call(0xc45a04a2c0, 0xd1c520, 0xc4a6395860, 0xc4a6634700, 0xc4a6642780, 0x0, 0x0, 0x0, 0x0, 0x0)
    /some/path/prebid-server/adapters/appnexus/appnexus.go:102 +0x1e1d
main.(*auctionDeps).auction.func1(0xd1b020, 0xc45a04a2c0, 0xd1c520, 0xc4a6395860, 0xc4a6634700, 0xc420124700, 0xc4a68b6c80, 0xc4a6577d00, 0xc4a68d7620, 0xc4a6642780)
    /some/path/prebid-server/pbs_light.go:370 +0xa2
created by main.(*auctionDeps).auction
    /some/path/prebid-server/pbs_light.go:368 +0x62a
```

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -88,6 +88,10 @@ func (a *AppNexusAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder 
 			return nil, errors.New("No placement or member+invcode provided")
 		}
 
+		// Fixes some segfaults. Since this is legacy code, I'm not looking into it too deeply
+		if len(anReq.Imp) <= i {
+			break
+		}
 		if params.InvCode != "" {
 			anReq.Imp[i].TagID = params.InvCode
 			if params.Member != "" {


cirla:
Digging into this a little:

Using [this index `i`](https://github.com/prebid/prebid-server/blob/c85df5b8bbe4f682e09041bb1191f930d02142a4/adapters/appnexus/appnexus.go#L80) from `bidder.AdUnits` as the index for `anReq.Imp` assumes a one-to-one mapping from ad units to imps.
However, it appears that this may not be true in two cases:

1.  [The call to `MakeOpenRTBGeneric`](https://github.com/prebid/prebid-server/blob/c85df5b8bbe4f682e09041bb1191f930d02142a4/adapters/appnexus/appnexus.go#L74) has `singleMediaTypeImp` set to `true`, which could result in [multiple imps per ad unit](https://github.com/prebid/prebid-server/blob/c85df5b8bbe4f682e09041bb1191f930d02142a4/adapters/openrtb_util.go#L110)  depending on the number of media types.
2. Regardless of the value of `singleMediaTypeImp`, there are many `continue` paths [in this for loop](https://github.com/prebid/prebid-server/blob/c85df5b8bbe4f682e09041bb1191f930d02142a4/adapters/openrtb_util.go#L81) that would result in skipping an ad unit without creating any imps.

PR #403, opened 2018-03-09
Set secure=1 on AMP requests

### LINKED ISSUE #402: Need to add `secure:1` default for AMP ORTB configs
In case the ORTB config for the AMP endpoint doesn't have a `secure` field, we should default to `secure:1` since all AMP traffic must be secure. 

Alternatively we could detect with the incoming request protocol the same as the legacy endpoint does.

hhhjort:
fixes #402 , AMP spec requires secure content URLs, so no use cases for insecure ads.

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -211,6 +211,15 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req 
 		errs = []error{fmt.Errorf("data for tag_id '%s' includes %d imp elements. Only one is allowed", ampId, len(req.Imp))}
 		return
 	}
+
+	// Force HTTPS as AMP requires it, but pubs can forget to set it.
+	if req.Imp[0].Secure == nil {
+		secure := int8(1)
+		req.Imp[0].Secure = &secure
+	} else {
+		*req.Imp[0].Secure = 1
+	}
+
 	return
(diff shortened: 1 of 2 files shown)


endpoints/openrtb2/amp_auction.go
```diff
@@ -211,6 +211,15 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
 		errs = []error{fmt.Errorf("data for tag_id '%s' includes %d imp elements. Only one is allowed", ampId, len(req.Imp))}
 		return
 	}
+
+	// Force HTTPS as AMP requires it, but pubs can forget to set it.
+	if req.Imp[0].Secure == nil {
+		secure := int8(1)
+		req.Imp[0].Secure = &secure
+	} else {
+		*req.Imp[0].Secure = 1
```
mkendall07: lol I like it.

dbemiller:
actually... you should probably doc this in `amp.md` as a difference between `/openrtb2/auction` and `/openrtb2/amp` too ^^

endpoints/openrtb2/amp_auction.go
hhhjort: there is the oddity that the openrtb spec sets this as a pointer, which takes up more memory than an 8 bit int, which doesn't even count the memory the int takes when the pointer is not nil.

Issue #392, opened 2018-03-08
Support Cache Invalidation for Stored Requests


cirla:
To solve the [caveat around caching of stored requests in the OpenRTB AMP endpoint](https://github.com/prebid/prebid-server/issues/376#issuecomment-370034060), Prebid server should support some kind of cache invalidation. This allows for faster debugging and for ensuring that the latest version of a stored request is the one that is run.

**Potential Solutions:**

1. Invalidate cache whenever a stored request is modified
    - This would have to propagate from the instance of PBS processing the request to all instances of PBS running on a host. Requires service discovery or fanning out stored request updates to all instances.
2. Always invalidate cache for given stored requests when `?debug=1` is set on AMP request
    - See #386 
    - **Caveat:** invalidates only on the PBS instance processing the request, causing desync between instance caches

cirla:
Closed by https://github.com/prebid/prebid-server/pull/404

Issue #366, opened 2018-02-23
Allow more flexible Stored Request storage schemes 


dbemiller:
Currently, there are two "types" of stored requests. Some are "request-scoped". For example:

```
{
  ext: {
    prebid: {
      targeting: {
        pricegranularity: "low"
      }
    }
  }
}
```

Others are "imp-scoped". For example:

```
{
  banner: {
    format: [
      { w: 300, h: 250 }
    ]
  },
  ext: {
    appnexus: {
      placementId: 123
    }
  }
}
```

A single HTTP request may include several stored request IDs, and make use of both types at once.

However, the PBS Postgres code today constrains host companies who manage Stored Request data in a few ways:

1. It must be able to fetch all the data in a single SQL query.
2. The query should include an `%ID_LIST%` macro which will be resolved to a list of _all_ the Stored Requests.

I strongly support (1), for performance reasons. However, (2) is less reasonable, and doesn't really have any good reasons motivating it.

In particular, it forces the IDs for "request-scoped" data and "imp-scoped" data to be globally unique. It's not possible (as far as I can tell) to build a SQL query which chooses the right ID in all situations. There isn't much value to this constraint, so it just makes the data unnecessarily difficult to manage.

This ticket is for designing (& possibly building) a feature to relax (2), but only if we can do it _without_ sacrificing the performance benefits of (1).

hhhjort:
Do we know of any benchmarks that would talk to the performance benefit of constraint 1? As I see it, you gain not having a second connection for the second query as a pro, but the con would be Postgres would have to figure out how to parallelize the two subqueries, and then join them, rather than being given the two independent subqueries to run with.

I am not sold that 1 gives us a significant performance benefit. A decent database implementation should be good at efficiently splitting up the compound query, but any decent database client should be running a connection pool which would mitigate the second connection cost. It may be 1 is a significant performance enhancement, I would just like to see some experience or numbers to back this up. I would hate for us to bend over backwards to keep 1, just to find out it saves us only 0.01 ms on each auction.

dbemiller:
No benchmarks, no. Although _we_ happen to care about postgres, bear in mind that the relevant code could be used with any database connection.

I don't think we'll have to bend over backwards to make this happen, though. To throw a simple suggestion out there, we could just make the query resolve two macros: `%REQUEST_ID_LIST%` and `%IMP_ID_LIST%`.

There might be some edge cases, and I obviously haven't written it yet, but... I imagine that would work.

[...]

PR #404, opened 2018-03-09
[Proposal] Rework stored request caches to allow propagation of invalidation/updates

cirla:
Proposed partial solution for #392. Includes implementation of an HTTP endpoint that could be exposed either behind authentication or on a private network to allow cache updates and invalidations.

**Example:** (adding hypothetical implementations for memcached `Cache` and Kafka `EventProducer`)
```go
// fetch from DB
db := db_fetcher.NewPostgresDb(dbCfg)
pgFetcher := db_fetcher.NewFetcher(db, cfg.Postgres.MakeAmpQuery)

// layer in-memory LRU on top of shared memcached instance
lruCache := in_memory.NewLRUCache(lruCfg)
memcachedCache := memcached.NewMemcachedCache(mcCfg)
caches := ComposedCache{lruCache, memcachedCache}

// subscribe caches to updates/invalidations from Kafka stream
kafkaEvents := events.NewKafkaEvents(kafkaCfg) 
listener := Listen(caches, kafkaEvents)

// layer caches on top of fetcher
configFetcher := WithCache(dbFetcher, caches)
```

`EventProducer`s could also be based on some pubsub technology, e.g. Postgres LISTEN/NOTIFY, Kafka, Kinesis.

### DIFF
--- a/pbs_light.go
+++ b/pbs_light.go
@@ -1017,11 +1017,15 @@ func NewFetchers(cfg *config.StoredRequests, db *sql.DB, eventProducers []events
 		byAmpId = stored_requests.WithCache(byAmpId, byAmpIdCache)
 
 		for _, ep := range eventProducers {
-			listeners = append(listeners, events.Listen(byIdCache, ep))
+			listener := events.SimpleEventListener()
+			go listener.Listen(byIdCache, ep)
+			listeners = append(listeners, listener)
 		}
 
 		for _, ep := range ampEventProducers {
-			listeners = append(listeners, events.Listen(byAmpIdCache, ep))
+			listener := events.SimpleEventListener()
+			go listener.Listen(byAmpIdCache, ep)
+			listeners = append(listeners, listener)
 		}
 	}
 	return
--- a/pbs_light.go
+++ b/pbs_light.go
@@ -896,10 +896,10 @@ func serve(cfg *config.Configuration) error {
 	router.ServeFiles("/static/*filepath", http.Dir("static"))
 
 	if cfg.StoredRequests.CacheEventsAPI {
-		router.POST("/storedrequests/openrtb2/:id", handleStoredRequests)
-		router.DELETE("/storedrequests/openrtb2/:id", handleStoredRequests)
-		router.POST("/storedrequests/amp/:id", handleAmpStoredRequests)
-		router.DELETE("/storedrequests/amp/:id", handleAmpStoredRequests)
+		router.POST("/storedrequests/openrtb2", handleStoredRequests)
+		router.DELETE("/storedrequests/openrtb2", handleStoredRequests)
+		router.POST("/storedrequests/amp", handleAmpStoredRequests)
+		router.DELETE("/storedrequests/amp", handleAmpStoredRequests)
 	}
 
 	hostCookieSettings = pbs.HostCookieSettings{
--- a/pbs_light.go
+++ b/pbs_light.go
@@ -970,7 +970,7 @@ const requestConfigPath = "./stored_requests/data/by_id"
 // If it can't generate both of those from the given config, then an error will be returned.
 //
 // This function assumes that the argument config has been validated.
-func NewFetchers(cfg *config.StoredRequests, db *sql.DB, eventProducers []events.EventProducer, ampEventProducers []events.EventProducer) (byId stored_requests.Fetcher, byAmpId stored_requests.Fetcher, listeners []events.EventListener, err error) {
+func NewFetchers(cfg *config.StoredRequests, db *sql.DB, eventProducers []events.EventProducer, ampEventProducers []events.EventProducer) (byId stored_requests.Fetcher, byAmpId stored_requests.Fetcher, listeners []*events.EventListener, err error) {
 	idList := make(stored_requests.MultiFetcher, 0, 3)
 	ampIdList := make(stored_requests.MultiFetcher, 0, 3)
 
--- a/pbs_light.go
+++ b/pbs_light.go
@@ -62,6 +62,8 @@ import (
 	"github.com/prebid/prebid-server/stored_requests/backends/file_fetcher"
 	"github.com/prebid/prebid-server/stored_requests/backends/http_fetcher"
 	"github.com/prebid/prebid-server/stored_requests/caches/in_memory"
+	"github.com/prebid/prebid-server/stored_requests/events"
+	apiEvents "github.com/prebid/prebid-server/stored_requests/events/api"
 	usersyncers "github.com/prebid/prebid-server/usersync"
 )
 
@@ -730,6 +732,7 @@ func init() {
 	// no metrics configured by default (metrics{host|database|username|password})
 
 	viper.SetDefault("stored_requests.filesystem", "true")
+	viper.SetDefault("stored_requests.cache_events_api", false)
 
 	// This Appnexus endpoint works for most purposes. Docs can be found at https://wiki.appnexus.com/display/supply/Incoming+Bid+Request+from+SSPs
 	viper.SetDefault("adapters.appnexus.endpoint", "http://ib.adnxs.com/openrtb2")
@@ -848,10 +851,24 @@ func serve(cfg *config.Configuration) error {
 	}
 	theExchange := exchange.NewExchange(theClient, pbc.NewClient(&cfg.CacheURL), cfg, metricsEngine)
 
-	byId, byAmpId, err := NewFetchers(&(cfg.StoredRequests), db)
+	var handleStoredRequests, handleAmpStoredRequests httprouter.Handle
+	var eventProducers, ampEventProducers []events.EventProducer
+	if cfg.StoredRequests.CacheEventsAPI {
+		var apiEventProducer, ampApiEventProducer events.EventProducer
+		apiEventProducer, handleStoredRequests = apiEvents.NewEventsAPI()
+		eventProducers = append(eventProducers, apiEventProducer)
+		ampApiEventProducer, handleAmpStoredRequests = apiEvents.NewEventsAPI()
+		ampEventProducers = append(ampEventProducers, ampApiEventProducer)
+	}
+	byId, byAmpId, listeners, err := NewFetchers(&(cfg.StoredRequests), db, eventProducers, ampEventProducers)
 	if err != nil {
 		glog.Fatalf("Failed to initialize config backends. %v", err)
 	}
+	defer func() {
+		for _, l := range listeners {
+			l.Stop()
+		}
+	}()
 
 	openrtbEndpoint, err := openrtb2.NewEndpoint(theExchange, paramsValidator, byId, cfg, metricsEngine, pbsAnalytics)
 	if err != nil {
@@ -877,6 +894,13 @@ func serve(cfg *config.Configuration) error {
 	router.GET("/status", status)
 	router.GET("/", serveIndex)
 	router.ServeFiles("/static/*filepath", http.Dir("static"))
+
+	if cfg.StoredRequests.CacheEventsAPI {
+		router.POST("/storedrequests/openrtb2/:id", handleStoredRequests)
+		router.DELETE("/storedrequests/openrtb2/:id", handleStoredRequests)
+		router.POST("/storedrequests/amp/:id", handleAmpStoredRequests)
+		router.DELETE("/storedrequests/amp/:id", handleAmpStoredRequests)
+	}
 
 	hostCookieSettings = pbs.HostCookieSettings{
 		Domain:       cfg.HostCookie.Domain,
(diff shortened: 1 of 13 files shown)


stored_requests/fetcher.go
```diff
@@ -19,52 +19,115 @@ type Fetcher interface {
...
+type CachedFetcher interface {
+	CacheableFetcher
+	Cache
+}
+
+// Subscriptions encapsulates forwarding updates and invalidations to subscribed Caches
+// It provides a convenient way to implement the CacheableFetcher interface by adding
+// it inside an implementation struct
+type Subscriptions struct {
+	subs []Cache
```
cirla: This should probably use channels so propagation can be done concurrently.

stored_requests/fetcher.go
```diff
@@ -19,52 +19,115 @@ type Fetcher interface {
...
+	Invalidate(ctx context.Context, ids []string)
+
+	// Update will update the given values in the cache
+	Update(ctx context.Context, data map[string]json.RawMessage)
+}
+
+// CacheableFetcher is a type of Fetcher that can be composed with one or more
+// Caches via WithCache and provide updates and validations to those Caches
+// when necessary
+type CacheableFetcher interface {
```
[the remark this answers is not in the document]
cirla: I was thinking more on this and I think the invalidation shouldn't be the responsibility of the Fetcher to bubble up as it is in the PR currently. Instead, we should probably have a separate simple interface for an event producer (let's call it `EventProducer` for now for lack of creativity) so that `Fetcher`s can just worry about fetching. Then composition could be done like:

stored_request_cache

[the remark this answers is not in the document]
hhhjort: This looks like a good architecture to me. Also more flexible, so it could easily support a cache invalidation driven from something other the data source, like an explicit cache invalidation API. That would be useful for datasources that don't support events.

cirla:
Here's a rough idea of what I'm going for in the latest iteration:

stored_request_cache

Instead of an explicit `Listener` type, there's a `Listen(Cache, EventProducer)` function which fires off a goroutine that consumes events from the `EventProducer`'s channels and forwards them to `Update` or `Invalidate` on the `Cache` depending on the type of event.

hhhjort:
So you are adding both invalidations and updates as possible events? I would focus on invalidations, I sort of favor updates being handled by invalidation + new fetch request.

Consider for example a publisher actively editing some configurations prior to going live. This would push their new configs to the top of the cache, potentially pushing out other stable configs out the bottom of the cache. This all while there are no live pages accessing these configs.

Not that updates are bad, if the cache is large enough then we don't have to worry about expiring early. And we do gain cutting out a slight bit of latency on the first request, as we won't have to fetch the config. But the gain here is much less than the gain of having an invalidation channel over the nothing we have now.

cirla:
I agree that invalidation events would be the primary use case (and the simpler one), though I added update events because I figured some `EventProducer` implementations might also be privy to the actual updated data (e.g. a Postgres LISTEN/NOTIFY) and could therefore directly update the cache and save a future fetch. I'm happy to take it out or add some better comment docs around it.

stored_requests/fetcher.go
```diff
@@ -19,52 +19,125 @@ type Fetcher interface {
...
+}
+
+func (c *composedCache) Invalidate(ctx context.Context, ids []string) {
+	for _, cache := range c.caches {
+		cache.Invalidate(ctx, ids)
+	}
+}
+
+func (c *composedCache) Update(ctx context.Context, data map[string]json.RawMessage) {
+	for _, cache := range c.caches {
```
hhhjort: This will store a copy of the new update into every cache we have, meaning we have N copies in memory where we will only use one. I would suggest checking each cache to see if the ID exists, and only updating the cache if it is found. If the id was not in any of the caches to begin with, we probably don't need to cache the update.

Further optimizing the update would depend on the use case(s) of having multiple caches. My first thought is we should invalidate later entries once we find a copy in the first cache, but that would break expected behavior if the various caches have different TTLs, or other operating parameters which would have the item leave the first cache prior to leaving a later cache. What use case(s) do you have in mind?

hhhjort: I don't think having an `Update()` is bad, I can see a possibility along the lines of the CPU L1, L2, L3 ... cache structure. A small(er) cache in memory, a larger one on disk, with a more aggressive timeout on the in-memory cache. But I think we are far from being at a state to work on such optimizations yet. Good to have for when we do want it, but if implementing this PR was being slowed down significantly because of it, I would leave it for a later PR.

On the comparison to the CPU caches, I seem to recall Intel and AMD having two different strategies. One would store the same data in L1 and L2, so it would still be found in L2 after it expires in L1. The other would only store in L1, and made the copy to L2 part of the L1 eviction process. That both are used in a heavily optimized process (CPU architecture) show there is no general best choice on this. :)

cirla: Really the only use case I had in mind was putting some kind of cache shared by multiple PBS instances (e.g. memcached) between the in-memory LRU cache and the database. I don't anticipate long chains of caches, I just wanted to make the interface reasonably generic.

Checking which keys in an update are in a cache and then updating only those keys in that cache layer adds some overhead, though it's probably acceptable as it's overhead on `Update`, not on `Get`, which should always be as fast as possible. Perhaps the `Cache` implementations themselves can decide how to handle an `Update` with keys that don't exist (e.g. store them or ignore them).

dbemiller: This looks good to me as is. First get it working. Then get it working well :)

We can't really prove which strategy makes the most sense until we add more detailed metrics and measure it anyway.

stored_requests/fetcher_test.go
```diff
@@ -112,6 +111,53 @@ func TestCacheSaves(t *testing.T) {
...
+		},
+	}
+	c3 := &mockCache{
+		mockGetData: map[string]json.RawMessage{
+			"3": json.RawMessage(`{"id": "3"}`),
+		},
+	}
+	c4 := &mockCache{
+		mockGetData: map[string]json.RawMessage{
+			"1": json.RawMessage(`{"id": "4"}`),
```
dbemiller: nice test!

dbemiller:
This looks great, but... be aware of #366 coming down the pipe.

It doesn't matter whether we fix that first or merge this and fix that later, but the public endpoint you're thinking about adding in https://github.com/cirla/prebid-server/pull/3 may not migrate too well.

stored_requests/events/api/api.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type eventsAPI struct {
+	invalidations chan []string
+	updates       chan map[string]json.RawMessage
+}
+
+// NewEventsAPI creates an EventProducer that generates cache events from HTTP requests.
+// The returned httprouter.Handle must be registered on both POST (update) and DELETE (invalidate)
+// methods and provided an `:id` param via the URL, e.g.:
```
cirla: Might want to add a comment that this should not be exposed as an endpoint without some kind of authentication as that would allow anyone to `POST` malicious configs to the cache.

stored_requests/events/api/api.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if r.Method == "POST" {
+		body, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte("Missing config data.\n"))
+			return
+		}
+
+		// check if valid JSON
```
cirla: We could go beyond just validating that it is JSON and validate the config itself, though this is intended to be an internal/authenticated endpoint and all requests would therefore come from the host/maintainers of the server instance.

stored_requests/fetcher.go
```diff
@@ -19,52 +19,98 @@ type Fetcher interface {
...
+
+// Invalidate will propagate invalidations to all underlying caches
+func (c ComposedCache) Invalidate(ctx context.Context, ids []string) {
+	for _, cache := range c {
+		cache.Invalidate(ctx, ids)
+	}
+}
+
+// Update will propagate updates to all underlying caches
+func (c ComposedCache) Update(ctx context.Context, data map[string]json.RawMessage) {
```
hhhjort: If update will propagate to all underlying caches, who then need to decide if it is a valid update or not (i.e. check to see if the key exists first), then we need a separate `Save()` method to tell the cache "This is a new value you are supposed to save". Otherwise the cache will either save every update that comes through, no matter if it has ever been requested or not, or never save anything as all updates fail the "is already in cache" check.

stored_requests/events/api/api.go
dbemiller: Should be fine either way.

In all honesty, I really doubt that anyone would use this in a prod setting. As soon as you deploy more than one instance, this would require the caller to be able to locate and call this endpoint on all of the instances. That's not generally easy in a containerized/microservice world, where the goal is usually to send a request to _an_ instance, not _all_ instances.

I suspect that this would be more like a development tool, in the same way that the `file_fetcher` is. It's definitely a convenient way to fiddle with the cache on a local instance.

stored_requests/fetcher.go
cirla: It's a lot of overhead to check determine the union of keys in the update and in the cache at each layer and then only forward the subset of updates to the cache's `Update`. Two alternatives that come to mind:
1. Leave as-is and let individual cache implementations decide what to do with an `Update` (e.g. LRUCache's Update could just ignore updates for keys that aren't present, or aggressively cache if it's relatively empty and ignore updates for non-existent keys when relatively full)
2. Change the signature of `Update` to just pass in a single `string` and `json.RawMessage` instead of a batch `map[string]json.RawMessage`.

pbs_light.go
```diff
@@ -1034,8 +1053,15 @@ func NewFetchers(cfg *config.StoredRequests, db *sql.DB) (byId stored_requests.F
...
-		byId = stored_requests.WithCache(byId, in_memory.NewLRUCache(cfg.InMemoryCache))
-		byAmpId = stored_requests.WithCache(byAmpId, in_memory.NewLRUCache(cfg.InMemoryCache))
+		byIdCache := in_memory.NewLRUCache(cfg.InMemoryCache)
+		byAmpIdCache := in_memory.NewLRUCache(cfg.InMemoryCache)
+
+		byId = stored_requests.WithCache(byId, byIdCache)
+		byAmpId = stored_requests.WithCache(byAmpId, byAmpIdCache)
+
+		for _, ep := range eventProducers {
+			listeners = append(listeners, events.Listen(byIdCache, ep), events.Listen(byAmpIdCache, ep))
```
cirla: This may not be desired as there could be overlaps between normal ids and AMP ids. There may need to be two separate `EventProducer`s with each cache listening to the appropriate one.

cindyhere33:
sorry for the delay, reviewing it now..

cirla:
Fixing some minor issues discovered in testing, will push soon and comment with testing methods/outputs.

cirla:
Testing Method:

Build `prebid-server` binary and run in debugger with breakpoints in all LRU Cache methods.

pbs.yaml:
```
stored_requests:
  filesystem: true
  cache_events_api: true
  in_memory_cache:
    ttl: 60
    size_bytes: 2097152
```

`stored_requests/data/by_id/1.json`:
```
{
  "ext": {
    "appnexus": {
      "placementId": 10433394
    }
  }
}
```

`test_requests/1.json`:
```
{
  "id": "test-imp-id",
  "banner": {
    "format": [
      {
        "w": 300,
        "h": 250
      },
      {
        "w": 300,
        "h": 600
      }
    ]
  },
  "ext": {
    "prebid": {
      "storedrequest": {
	"id": "1"
      }
    },
    "appnexus": {
      "placementId": 10433394
    }
  }
}
```

1. Update value not present in cache:
```
$ cat stored_requests/data/by_id/1.json |  http POST http://localhost:8000/storedrequests/openrtb2/1
HTTP/1.1 200 OK
```
Verified stepping in debugger that `Update` was called on the LRU cache and the result was no writes to the cache (as the config was not yet `Save`d) as expected.

2. Load value into cache
```
$ cat test_requests/1.json | http POST http://localhost:8000/openrtb2/auction
HTTP/1.1 400 Bad Request
Invalid request format: request.imp must contain at least one element.
```
Verified stepping in debugger that `Get` was called on the LRU cache, it resulted in a miss, the file fetcher was called, then `Save` was called (for all stored requests as the file fetcher is eager and returns data for all ids regardless of those asked for) as expected.

3. Verify value in cache
```
$ cat test_requests/1.json | http POST http://localhost:8000/openrtb2/auction
HTTP/1.1 400 Bad Request
Invalid request format: request.imp must contain at least one element.
```
Verified stepping in debugger that `Get` was called on the LRU cache, it resulted in a hit, neither `FetchRequests` nor `Save` were called as expected.

3. Update value in cache

4. Verify new value in cache

5. Invalidate value in cache

cirla:
Rebased after #443 merge. Will need to rework the events API to support imps and/or requests. Once that's done I'll run the above test again.

PR #407, opened 2018-03-12
Adds environment variable support for configuration

hhhjort:
Viper supports reading configuration options from environment variables, this enables this feature for Prebid Server.

### DIFF
--- a/pbs_light.go
+++ b/pbs_light.go
@@ -739,6 +739,11 @@ func init() {
 	viper.SetDefault("adapters.conversant.endpoint", "http://media.msg.dotomi.com/s2s/header/24")
 	viper.SetDefault("adapters.conversant.usersync_url", "http://prebid-match.dotomi.com/prebid/match?rurl=")
 	viper.SetDefault("host_cookie.ttl_days", 90)
+
+	// Set environment variable support:
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.SetEnvPrefix("PBS")
+	viper.AutomaticEnv()
 	viper.ReadInConfig()
 
 	flag.Parse() // read glog settings from cmd line
--- a/pbs_light_test.go
+++ b/pbs_light_test.go
@@ -633,9 +633,40 @@ func TestViperInit(t *testing.T) {
 }
 
 func TestViperEnv(t *testing.T) {
-	os.Setenv("PBS_PORT", "7777")
-	os.Setenv("PBS_ADAPTERS_PUBMATIC_ENDPOINT", "not_an_endpoint")
-	os.Setenv("PBS_HOST_COOKIE_TTL_DAYS", "60")
+	// Record environment
+	port, set := os.LookupEnv("PBS_PORT")
+	if set {
+		defer os.Setenv("PBS_PORT", port)
+	} else {
+		defer os.Unsetenv("PBS_PORT")
+	}
+
+	endpt, set := os.LookupEnv("PBS_ADAPTERS_PUBMATIC_ENDPOINT")
+	if set {
+		defer os.Setenv("PBS_ADAPTERS_PUBMATIC_ENDPOINT", endpt)
+	} else {
+		defer os.Unsetenv("PBS_ADAPTERS_PUBMATIC_ENDPOINT")
+	}
+
+	ttl, set := os.LookupEnv("PBS_HOST_COOKIE_TTL_DAYS")
+	if set {
+		defer os.Setenv("PBS_HOST_COOKIE_TTL_DAYS", ttl)
+	} else {
+		defer os.Unsetenv("PBS_HOST_COOKIE_TTL_DAYS")
+	}
+
+	err := os.Setenv("PBS_PORT", "7777")
+	if err != nil {
+		t.Fatal("Error setting evnvironment PBS_PORT")
+	}
+	err = os.Setenv("PBS_ADAPTERS_PUBMATIC_ENDPOINT", "not_an_endpoint")
+	if err != nil {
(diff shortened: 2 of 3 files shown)


pbs_light_test.go
```diff
@@ -623,3 +625,28 @@ func (validator *testValidator) Schema(name openrtb_ext.BidderName) string {
...
 }
+
+// Test the viper setup
+func TestViperInit(t *testing.T) {
+	CompareStrings(t, "Viper error: external_url expected to be %s, found %s", "http://localhost:8000", viper.Get("external_url").(string))
+	CompareStrings(t, "Viper error: adapters.pulsepoint.endpoint expected to be %s, found %s", "http://bid.contextweb.com/header/s/ortb/prebid-s2s", viper.Get("adapters.pulsepoint.endpoint").(string))
+}
+
+func TestViperEnv(t *testing.T) {
+	os.Setenv("PBS_PORT", "7777")
```
dbemiller: hmm..... is there any way to test this _without_ mutating state inside the test? As a general rule, tests which mutate global state are bad news.

Maybe better to set these `ENV` vars inside `./validate.sh`? Or at the very least, use some `defer` statements to reset these to their original values when the test is over.

Also: `os.Setenv` returns an error... it's probably worth checking that and `t.Fatal`-ing if the set didn't work. It's likely to be easier to debug than if `CompareStrings` fails with no other explanation.

dbemiller: On second thought..... meh. I just checked Viper's tests, and they just call `os.Setenv` as well: https://github.com/spf13/viper/blob/master/viper_test.go#L364

It is probably still worthwhile to check the error returns though.

hhhjort: Yes, I had an eye on the viper tests when I wrote mine. Still probably not a bad idea to unset at the end.

pbs_light_test.go
```diff
@@ -623,3 +625,59 @@ func (validator *testValidator) Schema(name openrtb_ext.BidderName) string {
...
+
+// Test the viper setup
+func TestViperInit(t *testing.T) {
+	CompareStrings(t, "Viper error: external_url expected to be %s, found %s", "http://localhost:8000", viper.Get("external_url").(string))
+	CompareStrings(t, "Viper error: adapters.pulsepoint.endpoint expected to be %s, found %s", "http://bid.contextweb.com/header/s/ortb/prebid-s2s", viper.Get("adapters.pulsepoint.endpoint").(string))
+}
+
+func TestViperEnv(t *testing.T) {
+	// Record environment
+	port, set := os.LookupEnv("PBS_PORT")
```
dbemiller: Could minimize duplicate code with a `func clearEnv(key string) (reseter func())`

I'm not sure how the Go syntax works on this... but you _may_ be able to `defer (clearEnv("PBS_PORT"))`. If that doesn't work, you can still reduce these 5 repeated lines to 2.

pbs_light_test.go
```diff
@@ -623,3 +625,59 @@ func (validator *testValidator) Schema(name openrtb_ext.BidderName) string {
...
+
+	ttl, set := os.LookupEnv("PBS_HOST_COOKIE_TTL_DAYS")
+	if set {
+		defer os.Setenv("PBS_HOST_COOKIE_TTL_DAYS", ttl)
+	} else {
+		defer os.Unsetenv("PBS_HOST_COOKIE_TTL_DAYS")
+	}
+
+	err := os.Setenv("PBS_PORT", "7777")
+	if err != nil {
```
dbemiller: Can minimize duplicate code with a `func forceEnv(t *testing.T, key string, val string) { ... }`

pbs_light_test.go
```diff
@@ -623,3 +625,55 @@ func (validator *testValidator) Schema(name openrtb_ext.BidderName) string {
...
+
+	// Basic config set
+	CompareStrings(t, "Viper error: port expected to be %s, found %s", "7777", viper.Get("port").(string))
+	// Nested config set
+	CompareStrings(t, "Viper error: adapters.pubmatic.endpoint expected to be %s, found %s", "not_an_endpoint", viper.Get("adapters.pubmatic.endpoint").(string))
+	// Config set with underscores
+	CompareStrings(t, "Viper error: host_cookie.ttl_days expected to be %s, found %s", "60", viper.Get("host_cookie.ttl_days").(string))
+}
+
+func CompareStrings(t *testing.T, message string, expect string, actual string) {
```
cindyhere33: If this method isn't used outside, can we do `compareStrings` instead?

hhhjort: Good point. :)

pbs_light.go
```diff
@@ -739,6 +739,11 @@ func init() {
 	viper.SetDefault("adapters.conversant.endpoint", "http://media.msg.dotomi.com/s2s/header/24")
 	viper.SetDefault("adapters.conversant.usersync_url", "http://prebid-match.dotomi.com/prebid/match?rurl=")
 	viper.SetDefault("host_cookie.ttl_days", 90)
+
+	// Set environment variable support:
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
```
cirla: Subjective bikeshedding here, feel free to ignore:

Would double underscores be more clear? E.g.: `PBS_HOST_COOKIE__TTL_DAYS` maps to `host_cookie.ttl_days`.

hhhjort: The example from the viper test cases has one underscore. I am a little wary of double underscores, eyeballing the string you can confuse single and double underscores if you don't examine it carefully. Are there places where people use the double underscore convention?

cirla: Single underscores makes sense and it looks like viper knows when to replace with a period and when to leave the underscore alone. Please ignore my comment, LGTM.

PR #410, opened 2018-03-13
Implemented the API for targeting keys without winners

### LINKED ISSUE #374: Add Targeting option to exclude winning bids
Currently, when `req.ext.targeting` exists, we append two types of targeting keys:

`some_key_{bidder}`: On the highest `Bid` for {bidder} on each `Imp`
`some_key`: On the highest `Bid` on each `Imp` overall

Philosophically, this makes the choice that Prebid Server should pick the winning bid. This is essential for Prebid Mobile, because it's very expensive to make App code updates. However it's actually a negative for Prebid.js, which (as a philosophy) intends to let Publishers choose the winning bid.

To support both clients, we should add an option for callers which makes PBS add the first type of targeting params, but _not_ the second type.

dbemiller:
Fixes #374


dbemiller:
yea... there's not much difference between a function on a type and off of one. A function declared on a type is basically the same as a stray function which takes an instance of that type as an argument. 

There's also a bit smaller chance of name collisions in the same package, but.... *shrugs*.

PR #412, opened 2018-03-16
Updates 3 parameter names and adds one new one

hhhjort:
We want to move the appnexus parameters from camelCase to snake_case. This affects 3 parameters (`placementId`, `invCode`, and `trafficSourceCode`). In this PR, the appnexus adapter will accept either version of the parameter name, at some point in the future we will disable the old names. We also add one new parameter: `use_pmt_rule`.

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -42,24 +42,26 @@ type KeyVal struct {
 }
 
 type appnexusParams struct {
-	LegacyPlacementId       int      `json:"placementId"`
-	LegacyInvCode           string   `json:"invCode"`
-	LegacyTrafficSourceCode string   `json:"trafficSourceCode"`
-	PlacementId             int      `json:"placement_id"`
-	InvCode                 string   `json:"inv_code"`
-	Member                  string   `json:"member"`
-	Keywords                []KeyVal `json:"keywords"`
-	TrafficSourceCode       string   `json:"traffic_source_code"`
-	Reserve                 float64  `json:"reserve"`
-	Position                string   `json:"position"`
-	UsePmtRule              *bool    `json:"use_pmt_rule"`
+	LegacyPlacementId       int             `json:"placementId"`
+	LegacyInvCode           string          `json:"invCode"`
+	LegacyTrafficSourceCode string          `json:"trafficSourceCode"`
+	PlacementId             int             `json:"placement_id"`
+	InvCode                 string          `json:"inv_code"`
+	Member                  string          `json:"member"`
+	Keywords                []KeyVal        `json:"keywords"`
+	TrafficSourceCode       string          `json:"traffic_source_code"`
+	Reserve                 float64         `json:"reserve"`
+	Position                string          `json:"position"`
+	UsePmtRule              *bool           `json:"use_pmt_rule"`
+	PrivateSizes            json.RawMessage `json:"private_sizes"`
 }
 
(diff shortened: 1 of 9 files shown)


adapters/appnexus/appnexus.go
```diff
@@ -42,19 +42,24 @@ type KeyVal struct {
...
+	LegacyInvCode           string   `json:"invCode"`
+	LegacyTrafficSourceCode string   `json:"trafficSourceCode"`
+	PlacementId             int      `json:"placement_id"`
+	InvCode                 string   `json:"inv_code"`
+	Member                  string   `json:"member"`
+	Keywords                []KeyVal `json:"keywords"`
+	TrafficSourceCode       string   `json:"traffic_source_code"`
+	Reserve                 float64  `json:"reserve"`
+	Position                string   `json:"position"`
+	UsePmtRule              *bool    `json:"use_pmt_rule"`
```
cirla: Why a pointer?

static/bidder-params/appnexus.json
```diff
@@ -5,14 +5,22 @@
...
   "properties": {
-    "placementId": {
+    "placement_id": {
       "type": "integer",
       "description": "An ID which identifies this placement of the impression"
     },
-    "invCode": {
+    "placementId": {
+      "type": "integer",
+      "description": "Deprecated"
```
cirla: I know it's fairly obvious but maybe add `"description": "Deprecated; use placement_id instead."

hhhjort: I can certainly do that.

adapters/appnexus/appnexus.go
hhhjort: Because we need to track three values: "true", "false", and "not present". When we marshal the parameter and the pointer is null, we can `omitempty` and pass it along as not existing, the same way we got it. If it was not a pointer, a missing `use_pmt_rule` would be forced to be explicitly set to `false` when we marshal it again to send on to Impbus.

cirla: Makes sense, thanks for the context!

adapters/appnexus/appnexus.go
```diff
@@ -42,19 +42,26 @@ type KeyVal struct {
...
+	LegacyTrafficSourceCode string          `json:"trafficSourceCode"`
+	PlacementId             int             `json:"placement_id"`
+	InvCode                 string          `json:"inv_code"`
+	Member                  string          `json:"member"`
+	Keywords                []KeyVal        `json:"keywords"`
+	TrafficSourceCode       string          `json:"traffic_source_code"`
+	Reserve                 float64         `json:"reserve"`
+	Position                string          `json:"position"`
+	UsePmtRule              *bool           `json:"use_pmt_rule"`
+	PrivateSizes            json.RawMessage `json:"private_sizes"`
```
cirla: I see that these are experimental so just leaving a note, not requesting any changes:

Once these become more well defined, I imagine we'd want to provide more concrete types instead of `json.RawMessage` (e.g. `[]PrivateSize`) for validation.

hhhjort: Only if we actually want to do any validation on the values. Otherwise best to leave this as JSON so we save an `UnMarshall`/`Marshall` cycle on the contents.

cirla: Makes sense. As this is an adapter it might make more sense to defer validation of fields like this to whatever service the adapter is calling out to.

PR #414, opened 2018-03-19
Add some basic bid validation

### LINKED ISSUE #225: OpenRTB Support: Filter invalid bids in response
Right now the `/auction/openrtb2` endpoints do nothing to validate bids from Bidders.

We should add validation and make sure all the fields required by the OpenRTB 2.5 spec (`id`, `price`, etc) are there, and remove bids which don't define them.

hhhjort:
This should fix #225


exchange/exchange.go
```diff
@@ -150,6 +144,8 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
 			// Add in time reporting
 			elapsed := time.Since(start)
 			brw.adapterBids = bids
+			// validate bids ASAP, so we don't waste time on invalid bids.
+			err2 := brw.validateBids()
```
cirla: Do we want to also count bid validation errors in the error metrics? If so, this could be something like:

```go
err = append(err, brw.validateBids()...)
```

And you'd no longer need the second loop further down.

exchange/exchange.go
```diff
@@ -186,6 +186,7 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
 	// Wait for the bidders to do their thing
 	for i := 0; i < len(cleanRequests); i++ {
 		brw := <-chBids
+		brw.validateBids()
```
cirla: If the `bidResponseWrapper` has `validateBids()` called on it inside of the goroutine before being sent back through the channel, we shouldn't have to call it a second time here.

exchange/validation.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	if len(validBids) != len(brw.adapterBids.bids) {
+		// If all bids are valid, the two slices should be equal. Otherwise replace the list of bids with the valid bids.
+		brw.adapterBids.bids = validBids
+	}
+	return err
+}
+
+// validateBid will run the supplied bid through validation checks and return true if it passes, false otherwise.
+func (brw *bidResponseWrapper) validateBid(bid *pbsOrtbBid) (bool, error) {
```
cirla: This could be a free function as it never uses `brw`.

exchange/exchange.go
hhhjort: Oops, yes, I had moved it from outside to inside, and forgot to get rid of this line. Thanks.

exchange/validation.go
hhhjort: I figured since it would never be used for anything else, just keep it glued to brw.

exchange/exchange.go
hhhjort: I do that below. I forgot that you could expand the list with `...`, so I did a loop and appended each one. I will have to fix that now.

cirla: The `...` ellipsis [will handle the case of an empty array gracefully](https://play.golang.org/p/Qblz5-pxrLN), so this could just be:
```go
err = append(err, brw.validateBids()...)
```

exchange/validation.go
cindyhere33:
```
I figured since it would never be used for anything else, just keep it glued to brw.
```

It is true, but doesn't feel so right.. 

I believe both receiver arguments and function arguments are just arguments inside the function. So, it should be equivalent to `validateBid(brw *bidResponseWrapper, bid *pbsOrtbBid)`.  

Please correct me if I'm wrong. 😅 


If you would not do `validateBid(brw *bidResponseWrapper, bid *pbsOrtbBid)` when brw is not used inside, you would not want to do `func (brw *bidResponseWrapper) validateBid(bid *pbsOrtbBid)` either.

exchange/validation.go
```diff
@@ -0,0 +1,50 @@
+package exchange
+
+import (
+	"fmt"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type bidResponseWrapper struct {
```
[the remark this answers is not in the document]
hhhjort: It is a hard call. But as per below, not making the validate functions methods, then moving this back to `exchange.go` is the way to go.

exchange/validation.go
hhhjort: I will free the function.

exchange/validation.go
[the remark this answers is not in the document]
cindyhere33: ohh now I see why you had to move it! Go's limitations on using a receiver outside of the file it is defined in. Ahh! 

I personally like `validation.go` to be a separate file. Been on my mind to suggest that for `endpoints/openrtb2/auction.go` as well. But if it's just two methods, I'd suggest for now, it would be better to include it in `exchange.go` itself (owing to consistency..?)

hhhjort:
Decided the validation code was small enough and moved it back into exchange.go

exchange/exchange.go
```diff
@@ -313,3 +319,41 @@ func (e *exchange) makeBid(Bids []*pbsOrtbBid, adapter openrtb_ext.BidderName) (
...
+}
+
+// validateBid will run the supplied bid through validation checks and return true if it passes, false otherwise.
+func validateBid(bid *pbsOrtbBid) (bool, error) {
+	if bid.bid == nil {
+		return false, fmt.Errorf("Empty bid object submitted.")
+	}
+	// These are the three required fields for bids
+	if bid.bid.ID == "" || bid.bid.ImpID == "" || bid.bid.Price == 0.0 {
+		return false, fmt.Errorf("Bid \"%s\" missing required field (id, impid, price)", bid.bid.ID)
```
dbemiller: These messages show up in `ext.debug`, so... I think it's most important to capture:

1. What went wrong?
2. What are the consequences?
3. What action can the publisher take to fix it?

I think it's worth splitting this into 3 `if` clauses so that you can identify specifically which field is missing, and tell them to log a Github issue and/or email the adapter's maintainer about it

PR #415, opened 2018-03-20
Pass new parameters for ad exchange in Adform adapter

red1ynx:


adapters/adform/info.yaml
```diff
@@ -0,0 +1,2 @@
+maintainer:
+  email: "scope.sspp@adform.com"
```
dbemiller: Looks like this was added accidentally. This file got moved to `static/bidder-info/adform.yaml`... so you can delete it here.

dbemiller: ahh nevermind... apparently I can just delete this through the github UI :)

Issue #259, opened 2017-12-13
Support analytics modules


bretg:
**Summary:** Prebid Server (PBS) needs to be able to log at a transaction level to support finer-grained analytics. 

**Motivation:** Browsers coming to PBS can have analytics adapters (via PBJS), so are in pretty good shape from a reporting perspective. But mobile requests don't have a way to generate detailed analytics. The best mobile can currently do is the metrics interface, which isn't granular enough for many reporting requirements.

**Definitions:**

* Event - an action requested by a user. There may be multiple events in a given request. e.g. request, adunit-request, bid-request, bid-response, cookie-sync-request, setuid-request.
* Request - the HTTP/S connection from a browser or app.
* AdUnit-Request - the Request from the user may imply that auctions for many AdUnits should be executed
* Bid-Request - each AdUnit-Request may be given to many different bidders. A Bid-Request is specific to an AdUnit+Bidder.
* Bid-Response - each Bid-Request may receive a response from the bidder, or it may not.
* Dimension - an attribute of an event that's useful for analysis. e.g. date, user platform, country, etc.
* Module - a block of code functionality that can be configured to be active on a given installation. e.g. bidder adapters.
* PBS Installation - a cluster of servers running Prebid Server. e.g. AppNexus and Rubicon Project both have PBS Installations available for publishers to utilize. Other pubs may spin up their own installations.

The request is to spec and implement an analytics interface and module system for Prebid Server. Examples of the types of analytics modules that should be possible:

- a module that just logs events to a local file on disk
- a module that does an HTTPS POST to an analytics service
- a module that forwards requests from certain account IDs back to that publisher's data warehouse

**Proposed Requirements:**

1) One or more analytics modules can be configured to run in a given installation of Prebid Server. For example, it may be useful to have one module that logs to a flat file while another samples requests to be forwarded back to the Publisher.

2) Each PBS installation should be able to configure which analytics module(s) should be informed of events. i.e. It should be possible for the same event to be routed to two different analytics modules.

3) Analytics modules should be allowed to have configuration to define things like endpoint, paths, sampling, account scope, or other behaviors.

4) Every Event must be passed to the analytics system. Configured modules may decide whether they care about the Event(s) and how to process it/them. e.g. an analytics module may not care about /setuid events, so may ignore them.

5) Event latency and status (e.g. succeeded, no-bid, timeout) should be passed to the analytics system.

6) All dimensions available to PBS must be available to the analytics modules. Including, but not limited to: 

* datetime
[...]

PR #302, opened 2018-01-29
Fixes issue #259: Analytics module

cindyhere33:
Refer to issue #259 and https://gist.github.com/cindyhere33/e908474fa49063c83eb7e33650fbfab9 for details. 

Implemented factory pattern to setup analytics modules. A sample module is enabled that logs to a file.

## **Steps to add new module:**
1. Create a corresponding field in `Analytics` in `config.Configuration` in _config.go_.  Config should look like:
``` 
...,
"analytics":{
    "enabled": true,
    "file":{
        "filename":"transactions"
     },
    "test":{
      "config":"parameters"
    }
  }
```
2. Create a struct (for example, `TestLogger{}`) and have it implement the `PBSAnalyticsModule` interface
3. Write a `NewTestLogger{}` factory method and pass it to `Register(key, factoryMethod)` in `registerAnalyticsModules()` inside _pbs_light.go_
4.  Add the key and the config to `analytics` in _pbs_light.go_. 
```
analytics := a.InitializePBSAnalytics(map[string]string{
			a.FILE_LOGGER: cfg.Analytics.File.Config,
                        a.TEST_LOGGER: cfg.Analytics.Test.Config,
		})
```

Shall come up with tests if this looks good. 

I read the argument about consistency vs completeness. If completeness is not a requirement, `LoggableAdapterRequests` can be removed from `AuctionObject`. The only purpose they serve would be to log the bid requests from each adapter in error scenarios.

PR #418, opened 2018-03-20
Support analytics modules for transactional logging

cindyhere33:
Fixes issue #259 and stems from PR #302.

Documentation contains the steps to create a new module. 

The FileLogger in `/analytics/file_logger.go` is an example of an analytics module. It is a sample module that dumps a transaction at a PBS endpoint into a file.


analytics/endpoint_data_objects.go
```diff
@@ -0,0 +1,91 @@
...
+	Error    []error
+	Success  bool
+}
+
+//Loggable object of a transaction at /cookie_sync
+type CookieSyncObject struct {
+	Type    RequestType
+	Status  int
+	Error   []error
+	Bidders string
```
cindyhere33: String - since I couldn't use a structured object owing to cyclic dependencies. Will have to pull out `cookieSyncResponse` from the _main_ package or `PBSBidders` from _pbs_ package in order to make the `Bidders` field structured.

cindyhere33:
Tests coming up..

analytics/endpoint_data_objects.go
```diff
@@ -0,0 +1,88 @@
...
+//Loggable object of a transaction at /openrtb2/amp endpoint
+type AmpObject struct {
+	Type            RequestType
+	Status          int
+	Error           []error
+	Request         *openrtb.BidRequest
+	AuctionResponse *openrtb.BidResponse
+	AmpResponse     map[string]string
+	UserAgent       string
+	Origin          string
```
hhhjort: Wouldn't `UserAgent` and `Origin` be found in `BidRequest`? Is there a reason these are being sent externally?

analytics/endpoint_data_objects.go
hhhjort: Once we can drop the legacy format we can probably restructure this better. :)

analytics/endpoint_data_objects.go
```diff
@@ -0,0 +1,88 @@
...
+
+//Loggable object of a transaction at /cookie_sync
+type CookieSyncObject struct {
+	Type    RequestType
+	Status  int
+	Error   []error
+	Bidders string
+}
+
+func (ao *AuctionObject) ToJson() string {
```
hhhjort: Do you have plans for this `ToJson()` method? As currently written it could be a private method, and perhaps moved to `file_module.go` if no other modules will need it.

analytics/endpoint_data_objects.go
cindyhere33: Yup!!

analytics/endpoint_data_objects.go
cindyhere33: I had that thought too. Will do that.

analytics/endpoint_data_objects.go
cindyhere33: I don't think they are in the BidRequest..

dbemiller: The `UserAgent` is `request.device.ua`. Not sure about the `Origin`.

(I'll do an actual review of this tomorrow)

config/config.go
```diff
@@ -22,12 +22,22 @@ type Configuration struct {
...
 	return cfg.StoredRequests.validate()
 }
 
+type Analytics struct {
+	File FileLogs `mapstructure:"file"`
+}
+
+//Corresponding config for FileLogger as a PBS Analytics Module
+type FileLogs struct {
+	Config string `mapstructure:"filename"`
```
sandraleon: would it make more sense to name this FileName?

analytics/endpoint_data_objects.go
dbemiller: hmmm.... I'm actually pretty troubled by this. There are good reasons to prevent circular dependencies, and changing types to circumvent them is a very slippery slope.

I played around with this a bit locally and got the following import cycle:

```
package github.com/prebid/prebid-server
        imports github.com/prebid/prebid-server/adapters
        imports github.com/prebid/prebid-server/pbs
        imports github.com/prebid/prebid-server/analytics
        imports github.com/prebid/prebid-server/pbs
```

More specifically, this makes `analytics` depend on`pbs.PBSBidder`, while `pbs.SetUID()` also depends on `analytics`. And the reason you want to use `PBSBidder` is because it's used by `main.CookieSync()`.

It'd be one thing if this were all legacy code... but it's actually not. Both `/cookie_sync` and `/setuid` are "live", and we have no plans to delete either one anytime soon.

Stepping back a bit: the general trouble is that `analytics` wants to use the same _contract classes_ as the `endpoints` do. So an `endpoint` Impl must be able to "see" `analytics`, while `analytics` must be able to "see" the endpoint contract classes.

In OpenRTB, this works just fine. The contract types live in `mxmCherry/openrtb` or `openrtb_ext`, while the `endpoints` live in `endpoints/openrtb2`. In the legacy code, they're not cleanly separated. `PBSBidder` is the _contract type_ for `main.CookieSync()`, and it lives in _the same module_ as the `pbs.SetUID()` endpoint impl.

There are a few ways we could untangle this. In an ideal world, we'd move `SetUID()` and `CookieSync()` to `endpoints`... but I think that's a bit overkill for the issue you're hitting here.

I propose we:

- Move `pbs.UserSyncInfo` into the `usersync` module. 
- Make a new struct for `CookieSync()` with _just_ those 3 properties that it needs, and put that in the `usersync` module too.
- Make `CookieSync()` use this struct instead of `PBSBidder`.
- Make `analytics` use this new struct instead of `string`.

Conceptually, this makes the `usersync` module analogous to `openrtb_ext`. Unless I'm missing a dependency somewhere, `main` would depend on `analytics`, `pbs`, and `usersync`. `analytics` would depend on `usersync` only, and you'd get the structured type that you want here.

It may also be worth doing these moves in a separate PR? Your choice there... smaller changesets are usually easier to review and merge, and help minimize conflicts... but I'm not sure which is better.

analytics/endpoint_data_objects.go
```diff
@@ -0,0 +1,54 @@
...
+const (
+	COOKIE_SYNC RequestType = "/cookie_sync"
+	AUCTION     RequestType = "/openrtb2/auction"
+	SETUID      RequestType = "/set_uid"
+	AMP         RequestType = "/openrtb2/amp"
+)
+
+//Loggable object of a transaction at /openrtb2/auction endpoint
+type AuctionObject struct {
+	Type      RequestType
```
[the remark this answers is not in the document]
cindyhere33: That is because a module like FileLogger dumps all transactions to any endpoint into the same file - helps identify to what endpoint the request was made. Not sure if it could help for other modules, can get rid of it if not needed.

endpoints/openrtb2/amp_auction.go
```diff
@@ -46,6 +47,17 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
 
 	// Set this as an AMP request in Metrics.
+
+	ao := analytics.AmpObject{
+		Type:   analytics.AMP,
+		Status: http.StatusOK,
+		Errors: make([]error, 0),
+	}
+
+	defer func() {
```
cirla: I think you can just call `defer deps.analytics.LogAmpObject(&ao)`

endpoints/openrtb2/auction.go
```diff
@@ -107,7 +126,10 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	// That status code can't be un-sent... so the best we can do is log the error.
 	if err := enc.Encode(response); err != nil {
 		glog.Errorf("/openrtb2/auction Error encoding response: %v", err)
+		ao.Errors = append(ao.Errors, fmt.Errorf("/openrtb2/auction Error encoding response: %v", err))
 	}
+
+	deps.analytics.LogAuctionObject(&ao)
```
cirla: This would be called twice due to [the deferred call above](https://github.com/prebid/prebid-server/pull/418/files#diff-09e5d415f33d74fd6df20b2584f318d1R67).

cindyhere33: you're right! updated the code.

endpoints/openrtb2/amp_auction.go
cindyhere33: thanks, got it

cindyhere33:
@dbemiller  I've made the requested changes. Working on improving the tests.

analytics/endpoint_data_objects.go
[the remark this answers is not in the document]
dbemiller: Seems like that would be an implementation detail of the FileLogger, then? If it takes logic which other modules would use too, the code could be written in utils functions.

PR #419, opened 2018-03-20
Add Multi-Fetcher functionality to stored config Fetchers

hhhjort:
This will allow us to compose a fetcher that can pull IDs from multiple datasources.

### DIFF
--- a/stored_requests/multifetcher.go
+++ b/stored_requests/multifetcher.go
@@ -12,12 +12,6 @@ type MultiFetcher []Fetcher
 func (mf *MultiFetcher) FetchRequests(ctx context.Context, ids []string) (map[string]json.RawMessage, []error) {
 	var errs []error
 	result := make(map[string]json.RawMessage, len(ids))
-	missingIDs := 0 // The number of missing IDs that each fetcher reported.
-	// If the number of errors == number of missing IDs, then it is likely that the errors are simply due
-	// to the IDs missing from the return result.
-	numIDs := len(ids)
-	// suspect fetchers ... fetchers that returned errors that don't match the number of missing IDs.
-	suspect := 0
 
 	// Loop over the fetchers
 	for _, f := range *mf {
@@ -27,13 +21,10 @@ func (mf *MultiFetcher) FetchRequests(ctx context.Context, ids []string) (map[st
 				remainingIDs = append(remainingIDs, id)
 			}
 		}
-		missingIDs = missingIDs + len(remainingIDs)
-		if len(errs) > 0 && len(errs) != len(remainingIDs) {
-			// This doesn't look like a simple error per missing ID.
-			suspect++
-		}
 		ids = remainingIDs
 		thisResult, rerrs := f.FetchRequests(ctx, ids)
+		// Drop NotFound errors, as other fetchers may have them. Also don't want multiple NotFound errors per ID.
+		rerrs = dropMissingIDs(rerrs)
 		if len(rerrs) > 0 {
 			errs = append(errs, rerrs...)
 		}
@@ -42,9 +33,24 @@ func (mf *MultiFetcher) FetchRequests(ctx context.Context, ids []string) (map[st
 			result[k] = v
 		}
 	}
-	// If we have all the results and a number of errors == the number of missing IDs, then assume all is good.
-	if len(result) == numIDs && len(errs) <= missingIDs && suspect == 0 {
-		errs = []error{}
+	// Add missing ID errors back in for any IDs that are still missing
+	for _, id := range ids {
+		if _, ok := result[id]; !ok {
+			errs = append(errs, NotFoundError(id))
+		}
 	}
 	return result, errs
+}
+
+// dropMissingIDs will scrub the NotFoundError's from a slice of errors.
+// The order of the errors will not be preserved.
+func dropMissingIDs(errs []error) []error {
+	// filtered errors
+	ferrs := errs[:0]
+	for _, e := range errs {
+		if _, ok := e.(NotFoundError); !ok {
+			ferrs = append(ferrs, e)
+		}
+	}
+	return ferrs
 }
--- a/stored_requests/multifetcher.go
+++ b/stored_requests/multifetcher.go
@@ -12,6 +12,13 @@ type MultiFetcher []Fetcher
 func (mf *MultiFetcher) FetchRequests(ctx context.Context, ids []string) (map[string]json.RawMessage, []error) {
 	var errs []error
 	result := make(map[string]json.RawMessage, len(ids))
+	missingIDs := 0 // The number of missing IDs that each fetcher reported.
+	// If the number of errors == number of missing IDs, then it is likely that the errors are simply due
+	// to the IDs missing from the return result.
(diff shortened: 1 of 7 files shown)


hhhjort:
Added some better handling of the "ID not found" errors, since our code is paranoid of fetcher returning any errors. If we have all the results, and the errors are consistent with missing ID errors, we toss the list of errors and assume all is good.

This will fail if the ID list contains any duplicates, as the count of IDs will not match the count of results. Don't know how to fix that without adding a bunch of code to dedupe the ID list, which will impact performance. Is this a real worry?

stored_requests/multifetcher.go
```diff
@@ -0,0 +1,50 @@
...
+	result := make(map[string]json.RawMessage, len(ids))
+	missingIDs := 0 // The number of missing IDs that each fetcher reported.
+	// If the number of errors == number of missing IDs, then it is likely that the errors are simply due
+	// to the IDs missing from the return result.
+	numIDs := len(ids)
+	// suspect fetchers ... fetchers that returned errors that don't match the number of missing IDs.
+	suspect := 0
+
+	// Loop over the fetchers
+	for _, f := range *mf {
```
cirla: Depending on how many fetchers could be in one `MultiFetcher` and how much latency is involved, we may want to kick off this loop body as a goroutine so they can run concurrently and use a [`WaitGroup`](https://golang.org/pkg/sync/#WaitGroup) to wait until they're all done.

hhhjort: The use case I have in my mind is moving from one backend to another, where the stored requests can't easily be copied live from one backend to the other. This allows a transition period where it will work as long as it exists in one of the locations. So that would be a max of 2. But good idea if we want to crank up the number of fetchers higher.

stored_requests/multifetcher.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+		ids = remainingIDs
+		thisResult, rerrs := f.FetchRequests(ctx, ids)
+		if len(rerrs) > 0 {
+			errs = append(errs, rerrs...)
+		}
+		// Loop over the results
+		for k, v := range thisResult {
+			result[k] = v
+		}
```
cirla: Could break early if all IDs found, e.g.:
```go
if len(result) == numIDs {
    break
}
```

stored_requests/multifetcher.go
```diff
@@ -0,0 +1,50 @@
...
+	// Loop over the fetchers
+	for _, f := range *mf {
+		remainingIDs := make([]string, 0, len(ids))
+		for _, id := range ids {
+			if _, ok := result[id]; !ok {
+				remainingIDs = append(remainingIDs, id)
+			}
+		}
+		missingIDs = missingIDs + len(remainingIDs)
+		if len(errs) > 0 && len(errs) != len(remainingIDs) {
```
cirla: There's a corner case here where `len(remainingIDs) == 0` and a single error returned could either be a missing ID or a more serious problem that we may want to return. It may make more sense to define something like this in fetcher.go:

```go
type NotFoundError string

func (e NotFoundError) Error() string {
    return fmt.Sprintf(`Stored Request with ID="%s" not found.`, e)
}
```
And require the fetchers use this error to signify a missing ID so that we can explicitly check at the end:
```go
cleanedErrs := make([]error, 0, len(errs))
for _, err := errs {
  if _, ok := err.(NotFoundError); ok {
    if _, present := result[err]; present {
      // previously missing ID was found, omit error
      continue
    }
  }
  cleanedErrs = append(cleanedErrs, err)
}
```
and then we could remove all of the code around `suspect` and `missingIDs` and just check:
```go
if len(cleanedErrs) > 0 {
  return result, cleanedErrs
}
return result, nil
```

hhhjort: I had pondered that, and thought perhaps it should go into its own PR since it involved modifying all the fetchers directly. But if you think it fits inside of this PR I am good with that as well.

stored_requests/multifetcher.go
```diff
@@ -0,0 +1,50 @@
...
+	missingIDs := 0 // The number of missing IDs that each fetcher reported.
+	// If the number of errors == number of missing IDs, then it is likely that the errors are simply due
+	// to the IDs missing from the return result.
+	numIDs := len(ids)
+	// suspect fetchers ... fetchers that returned errors that don't match the number of missing IDs.
+	suspect := 0
+
+	// Loop over the fetchers
+	for _, f := range *mf {
+		remainingIDs := make([]string, 0, len(ids))
```
cirla: You could avoid having to recreate remaining IDs each time through the loop by doing [something like this](https://github.com/prebid/prebid-server/pull/404/files#diff-c705d31b368d04ad30e87727936db388R57) to safely delete fetched ids from the slice while iterating over it, then this could be initialized outside of the loop:
```go
remainingIDs := make([]string, len(ids))
copy(remainingIDs, ids)
```

and then updated after the call to `FetchRequests` below.

hhhjort: Wouldn't the copy be essentially a loop over the entries, and then we do a second loop to delete out the ones we don't want? I still think the append is more efficient, given I have sized the slice such that it will not require resizing.

stored_requests/multifetcher.go
cirla: It does increase the scope and it's not a perfect solution as there's no way to statically ensure that `Fetcher` implementations are always using the error when they should, but I think it may be worth it and safer than assuming `len(errs) > 0 && len(errs) == len(remainingIDs)` is OK.

stored_requests/multifetcher.go
cirla: The copy would just happen once before the `for _, f := range *mf` loop; the delete loop would happen inside the `for _, f := range *mf` loop in place of the `for _, id := range ids {` loop below.

hhhjort: The copy would happen once, but the internals of the copy is essentially a `for` loop (if not explicitly a `for` loop) Certainly it is not a single CPU operation, which seems to be how you are counting to say the delete would be more efficient.

cirla: Sorry I'm not wording things clearly, I more meant that currently it's something like this:
```
for each fetcher {
    reset remaining ids to []
    for each id {
        check if id in result and if not, append to remaining ids
    }
    fetch remaining ids
}
```
which loops over all ids in each iteration of the main loop
I was suggesting something like:
```
copy ids to remaining ids
for each fetcher {
    fetch remaining ids
    for each remaining id {
        check if id in fetched ids and if so, add to result and delete from remaining ids
    }
}
```
which only loops over all ids once, then each iteration of the main loop is only looping over a shrinking slice of remaining ids (the delete is constant time).

They're both the equivalent in worst case (all ids missing from all fetchers) but the latter would perform better as each fetcher reduced the number of remaining ids. The latter also has the added benefit that `len(remainingIDs)` would be zero at the end if all IDs were filled instead of containing the remaining IDs from before the final fetcher was called.

To be honest though there's really not much performance difference in practice because we will always have a small number of fetchers (max 2) and a small list of IDs to fetch.

PR #421, opened 2018-03-23
Windows - Resolved load json schema error

abemille:
When trying to run the project on Windows, I received the following error:

```
F0322 22:26:04.060973    9724 pbs_light.go:904] Failed to create the bidder params validator. Failed to load json schema at ./static/bidder-params/adform.json: Reference {0xc042133780 {[]} false true false true false} must be canonical
goroutine 1 [running]:
```

These changes resolve this error.

I also added .vscode/ to the .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ _obj
 _test
 .cover/
 .idea/
+.vscode/
 
 # Architecture specific extensions/prefixes
 *.[568vq]
--- a/openrtb_ext/bidders.go
+++ b/openrtb_ext/bidders.go
@@ -93,7 +93,10 @@ func NewBidderParamsValidator(schemaDirectory string) (BidderParamValidator, err
 		if _, isValid := BidderMap[bidderName]; !isValid {
 			return nil, fmt.Errorf("File %s/%s does not match a valid BidderName.", schemaDirectory, fileInfo.Name())
 		}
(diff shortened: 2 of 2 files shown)


openrtb_ext/bidders.go
```diff
@@ -94,11 +93,11 @@ func NewBidderParamsValidator(schemaDirectory string) (BidderParamValidator, err
 		if _, isValid := BidderMap[bidderName]; !isValid {
 			return nil, fmt.Errorf("File %s/%s does not match a valid BidderName.", schemaDirectory, fileInfo.Name())
 		}
-
-		schemaLoader := gojsonschema.NewReferenceLoaderFileSystem(fmt.Sprintf("file:///%s", fileInfo.Name()), filesystem)
+		toOpen, _ := filepath.Abs(filepath.Join(schemaDirectory, fileInfo.Name()))
```
dbemiller: This should probably return the error, rather than ignoring it

.gitignore
```diff
@@ -8,6 +8,7 @@ _obj
 _test
 .cover/
 .idea/
+.vscode/
```
cirla: You can also setup your local machine to have a `.gitignore` that's always applied to any git repo you're working in:

```sh
touch ~/.gitignore
git config --global core.excludesfile ~/.gitignore
```

And you never have to worry about a project not having `.gitignore` set up for your editor of choice.

PR #424, opened 2018-03-27
Support custom granularity in OpenRTB requests

### LINKED ISSUE #323: Custom Price Granularity
It's time to implement the PBJS [Custom Price Granularity](http://prebid.org/dev-docs/publisher-api-reference.html#customCPMObject) in Prebid Server.

We need someone to:

1) define a new config mechanism (e.g. table structure) for defining buckets
2) define how those buckets map to accounts
3) define default and error behavior

This blocks the currency feature (https://github.com/prebid/prebid-server/pull/281) from being useful outside of the US and a few other markets where currency is close in value to $1.

@dbemiller writes:

> I don’t think (1) or (2) are needed to support custom price buckets. OpenRTB support in PBS is agnostic about the host company’s database structure; the SQL query is configurable, so the host company can use any database structure it wants so long as the data can be fetched with a single SQL query.
> 
> All we need to support custom buckets is to define an API and implement the auction logic for it. The Prebid.js API is quite strong here… so my suggestion would be to just steal it and let users send that JSON in as bidrequest.ext.prebid.targeting.pricegranularity. Happy to talk about alternatives if there are better ideas out there though.

I don't think the client should be sending the price buckets here -- they should be something that the page/app doesn't have hard-coded, but rather tools can be built to manage the buckets from the server side.

hhhjort:
fixed #323


openrtb_ext/request_test.go
```diff
@@ -86,3 +89,66 @@ func TestCacheLegal(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("Failed to Unmarshall granularity: %s", err.Error())
+		}
+		if !reflect.DeepEqual(test.target, resolved) {
+			t.Errorf("Granularity unmarshal failed, the unmarshalled JSON did not match the target\nExpected: %v\nActual  : %v", test.target, resolved)
+		}
+	}
+}
+
+var validGranularityTests []granularityTestData = []granularityTestData{
```
dbemiller: Could you add an explicit `"precision": 0` check? I know it's an edge case, per https://github.com/prebid/prebid-server/issues/323#issuecomment-375764535

would be nice to make sure it's really coming out as "0", and not defaulting to "2".

openrtb_ext/request_test.go
```diff
@@ -86,3 +89,66 @@ func TestCacheLegal(t *testing.T) {
...
+	tests := [][]byte{
+		[]byte(`{}`),
+		[]byte(`[]`),
+		[]byte(`[{"precision": -1, "min":0, "max":20, "increment":0.5}]`),
+		[]byte(`[{"min": 5, "max":1, "increment": 0.1}]`),
+		[]byte(`[{"min":0, "max":20, "increment": -1}]`),
+		[]byte(`[{"min":"0", "max":"20", "increment": "0.1"}]`),
+		[]byte(`[{"min":0, "max":20, "increment":0.1}, {"min":15, "max":30, "increment":1.0}]`),
+		[]byte(`[{"precision": 2, "min":0, "max":10, "increment":0.1}, {"precision": 1, "min":10, "max":50, "increment":1}]`),
+		[]byte(`[{"max":20, "increment":0.1}. {"max":10, "increment":0.02}]`),
```
dbemiller: should there be a test case here for non-overlapping ranges which are out of order? Last I heard, you planned to reject those.

hhhjort: I thought that was what that line was, but adding a test with `min` specified to ensure the non-overlapping part. You can never have too many tests.

exchange/price_granularity_test.go
```diff
@@ -0,0 +1,35 @@
+package exchange
+
+import (
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"testing"
+)
+
+func TestGetPriceBucketString(t *testing.T) {
+	price := 1.87
+	getOnePriceBucket(t, "low", price, "1.50")
```
dbemiller: Seems like you might need to add a test here for `precision: 0` too? Assuming that line I called out in the main file is actually buggy, of course.

docs/endpoints/openrtb2/auction.md
```diff
@@ -139,10 +139,17 @@ to set these params on the response at `response.seatbid[i].bid[j].ext.prebid.ta
 
 ```
 {
-  "pricegraularity": "One of ['low', 'med', 'high', 'auto', 'dense']", // Required property.
+  "pricegraularity": [{
+    "precision": 2,
+    "min":0.0,
+    "max":20.0
+    "increment":0.1 }], // Required property.
```
dbemiller: You're not requiring this anymore, are you? I saw in the code that it defaults to `medium`.

Since we're calling those strings "deprecated" now, it seems useful to list out the default explicitly here.

docs/endpoints/openrtb2/auction.md
```diff
@@ -139,10 +139,17 @@ to set these params on the response at `response.seatbid[i].bid[j].ext.prebid.ta
...
   "includewinners": false // Optional param defaulting to true
 }
 ```
+The list of price granularity ranges must be given in order of increasing `max` values. `max` and `precision` are required. If `precision` is omitted, it will default to `2`. If `min` is omitted, it will default to the previous `max`.
+
+For backwards compatibility the following strings will also be allowed as price granularity definitions. There is no guarantee that these will be honored in the future. "One of ['low', 'med', 'high', 'auto', 'dense']"
```
dbemiller: May be useful to link to here: http://prebid.org/prebid-mobile/adops-price-granularity.html

openrtb_ext/request.go
```diff
@@ -71,14 +71,186 @@ func (ert *ExtRequestTargeting) UnmarshalJSON(b []byte) error {
...
+	if err != nil {
+		return err
+	}
+	if len(gran) > 1 {
+		// We only need to loop over the ranges if we have more than one (inter-range checks)
+		precision := gran[0].Precision
+		var prevMax float64 = 0
+		for i, gr := range gran {
+			if gr.Precision != precision {
+				return errors.New("Price granularity error: precision not consistent across entires")
```
dbemiller: Is this actually an error? It seems like a legitimate input to me.

openrtb_ext/request.go
```diff
@@ -71,14 +71,186 @@ func (ert *ExtRequestTargeting) UnmarshalJSON(b []byte) error {
...
+			}
+			if gr.Max < prevMax {
+				return errors.New("Price granularity error: range list must be ordered with increasing \"max\"")
+			}
+			if gr.Min == 0.0 {
+				// Default min to be the previous max
+				// On the first entry, we will likely overwrite 0.0 with 0.0, which should be ok. Adding a conditional to skip likely won't save any processing time.
+				gr.Min = prevMax
+				gran[i].Min = prevMax
```
dbemiller: Why are there two lines here? Both look to be doing the same thing.

docs/endpoints/openrtb2/auction.md
hhhjort: We always defaulted to medium, so actually was never required. I'll get rid of "required"

docs/endpoints/openrtb2/auction.md
```diff
@@ -125,10 +125,17 @@ to set these params on the response at `response.seatbid[i].bid[j].ext.prebid.ta
...
-  "pricegraularity": "One of ['low', 'med', 'high', 'auto', 'dense']", // Required property.
+  "pricegraularity": [{
+    "precision": 2,
+    "min":0.0,
+    "max":20.0
+    "increment":0.1 }], // Required property.
   "includewinners": false // Optional param defaulting to true
 }
 ```
+The list of price granularity ranges must be given in order of increasing `max` values. `max` and `precision` are required. If `precision` is omitted, it will default to `2`. If `min` is omitted, it will default to the previous `max`.
```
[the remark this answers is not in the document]
hhhjort: You can have a set of granuarities that never uses `precision` or `min`. `min` cannot be negative, but I believe I had a test case where there was no `min`

exchange/price_granularity.go
```diff
@@ -0,0 +1,50 @@
...
+const DEFAULT_PRECISION = 2
+
+// GetCpmStringValue is the externally facing function for computing CPM buckets
+func GetCpmStringValue(cpm float64, config openrtb_ext.PriceGranularity) (string, error) {
+	cpmStr := ""
+	bucketMax := 0.0
+	increment := 0.0
+	precision := config[0].Precision
+	// If we wish to support precision "0", we will need to remove this check
+	if precision == 0 {
```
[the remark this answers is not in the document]
hhhjort: Oh, yes, I did add the code to support 0, then forgot to remove the check. Thanks

openrtb_ext/request.go
hhhjort: It gets a little confusing if the precision changes. I think there is a much greater chance that the precision is a typo if it changes than it being a legitimate desire to have the price bucket string change precision in the middle of the range. I am open to changing it, but my feeling is to set this as an expectation, and if someone has a need submit an issue or PR to change it.

openrtb_ext/request.go
hhhjort: Yes, except `gr` is a copy I believe, so to have the change stick I need to change `gran[i]`. But the proper resolution is to just test `gran[i].Min` below rather than `gr.Min`. Fixing

openrtb_ext/request_test.go
hhhjort: I agree, so see the last check, its already there

openrtb_ext/request.go
dbemiller: That's fair. I checked in with Kendall too, and he said he's never seen anyone want to do it either.

In that case, do you think we should change the API? It seems silly for publishers to define it in multiple places if we force it to be consistent anyway. For example, it could be:

```
{
  "precision": 2,
  "ranges": [
    { "min": 0, "max": 0.5, "increment": 0.1 },
    { "min": 0.5, "max": 1.0, "increment": 0.2 },
    ...
  ]
}
```

hhhjort: I thought this format was mirroring how we define it in prebid.js. I would want to be consistent in layout with what we do there to reduce confusion. But otherwise I am all for pulling precision out.

docs/endpoints/openrtb2/auction.md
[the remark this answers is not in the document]
dbemiller: I didn't notice that... but I love it! Good decision IMO.

A followup question: is there _ever_ a good reason to define a `min`, then? It seems like there's only one "correct" value, and defining it to be anything else will break the request (e.g. by creating overlap or gaps).

If so, maybe we should just delete it from the API?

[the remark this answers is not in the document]
mkendall07: The only possible scenario I can think of is if you didn't want to start the min = 0 for the first bucket. Not sure that's a real use case though.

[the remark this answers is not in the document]
hhhjort: Readability perhaps, or you may want to manually set the next min so it properly lines up as a multiple of the new increment.

so for example your previous range has a max of 5, and an increment of 0.2. You want the new range to have an increment of 0.3, you might want the min to be 5.1 to get things to line up nicely, at the cost of one small bucket. (4.6, 4.8, 5.0. 5.1, 5.4, 5.7, 6.0 ...)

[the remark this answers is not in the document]
dbemiller: If it were... what behavior would you expect from PBS for Bids below this min?

Options which come to mind are...
- Filter those bids out of the auction
- Return those bids exactly
- Use the lowest price bucket anyway

for (1), the "absolute min" would just overlap in purpose with `imp[i].bidfloor`. For (2), I'm not sure how pubs would take advantage of it in DFP. For (3), you may as well not have defined it.

Open to other options though...

dbemiller:
If we removed `min`... doesn't that mean the two are equivalent? `0` _has_ to be the overall min

hhhjort:
The effective `min` is the previous range's `max`, so where to count increments from is still open. I currently use `Min` when I loop over ranges to see if the incoming `cpm` falls within the current range ... it would also be possible to save the previous `Max` in a variable and use that instead. But until we get a final decision on how we count increments, I would rather leave `Min` in internally as we may need it again depending on the final solution.

dbemiller:
oh! I see.

Yeah... counting increments from the _range_ min is definitely more intuitive. Starting from 0 would be super confusing.

On the internal `Min` variables: I agree. If they make the code more straightforward, keep em ^^.  I just don't think they should respond to an undocumented API param.

openrtb_ext/request.go
```diff
@@ -71,14 +71,155 @@ func (ert *ExtRequestTargeting) UnmarshalJSON(b []byte) error {
...
+				return errors.New("Price granularity error: increment must be a nonzero positive number")
+			}
+			// Enforce that we don't read "min" from the request
+			pgraw.Ranges[i].Min = prevMax
+			if pgraw.Ranges[i].Min < prevMax {
+				return errors.New("Price granularity error: overlapping granularity ranges")
+			}
+			prevMax = gr.Max
+		}
+	} else if len(pgraw.Ranges) == 0 {
```
dbemiller: This could just be a simple `else`... if `len(arr)` isn't positive, it'd be 0.

hhhjort: Yes, a hold over from when I only ran checks when `len()>1`

PR #429, opened 2018-03-29
Add openx adapter

haohany:


adapters/openx/openx.go
```diff
@@ -0,0 +1,134 @@
...
+		Uri:     uri,
+		Body:    reqJSON,
+		Headers: headers,
+	}}, errs
+}
+
+// Mutate the imp to get it ready to send to openx.
+func preprocess(imp *openrtb.Imp, reqExt *openxReqExt) error {
+	// We only support banner impressions for now.
+	if imp.Video != nil || imp.Native != nil || imp.Audio != nil {
```
dbemiller: Small note: you may just want `if imp.Banner == nil` here? Publishers may send an `imp` with both `banner` _and_  one of these other formats, if it's a multi-format imp.

adapters/openx/openx.go
```diff
@@ -0,0 +1,134 @@
...
+	var bidResp openrtb.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+
+	bids := make([]*adapters.TypedBid, 0, 5)
+
+	for _, sb := range bidResp.SeatBid {
+		for _, bid := range sb.Bid {
+			bids = append(bids, &adapters.TypedBid{
```
dbemiller: Someone recently submitted a bugfix which you might want to apply here: https://github.com/prebid/prebid-server/pull/413

static/bidder-params/openx.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "delDomain": {
+      "type": "string",
+      "description": "The delivery domain for the customer.",
+      "pattern": "\\.[a-zA-Z]{2,3}$",
+      "format": "hostname"
+    },
+    "customFloor": {
+      "type": "number",
+      "description": "The Minimum CPM price in USD."
```
dbemiller: Should this declare a min value of 0?

haohany:
Thanks for the review! I'll update the PR based on the comments.

dbemiller:
Sure thing. Also, you might want to take a quick look at #433. We'll be merging that in about a month... but you may have some server-side changes to make sure you're returning a real creative ID.

adapters/openx/openxtest/params/race/banner.json
```diff
@@ -0,0 +1,6 @@
+{
```
dbemiller: May be worth adding a `params/race/video.json` now too?

static/bidder-info/openx.yaml
```diff
@@ -0,0 +1,9 @@
+maintainer:
+  email: "team-openx@openx.com"
+capabilities:
+  app:
+    mediaTypes:
+      - banner
+  site:
+    mediaTypes:
+      - banner
```
dbemiller: Looks like this should declare support for `video` now too.

adapters/openx/openx.go
```diff
@@ -0,0 +1,190 @@
...
+// OpenX doesn't support multi-type impressions.
+// If both banner and video exist, take banner as we do not want in-banner video.
+func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	mediaType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Banner == nil && imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			return mediaType
```
hhhjort: This assumes that the media type is always banner or video. I think you want a check in `preprocess` to verify that the imp type is one of the two, and reject it otherwise.

PR #430, opened 2018-03-30
Bid cpm adjustments

### LINKED ISSUE #350: NET / GROSS Bid adjustments
#237 mentions NET revenue in a bid response, I'm wondering if this instead needs to be applied to the bid request. I'm specifically looking at the AMP implementation at the moment where all bid adjustments need to be done on the server. 

In our case we have very few bidders that return NET bids and so in order to equalise the bids we would have to apply the rev share. The reason I'd like control on the bid request is that we, in some cases, apply additional adjustments to bidders where we see discrepancies (could be currency or delivery issues). We also have bidders who return a NET bid but have additional 'ad serving charges' applied to invoices so any NET bid returned in the bid response would still need further adjustment.

dbemiller:
Fixes #350.

I also made some "exchange => adapter" JSON tests, because all of our existing ones are pretty fragile. Many of our tests in this package hit a lot of private methods and implementation details... and I've found them really hard to debug and maintain while working on a few of these recent features (aliases, targeting, and now this).

Although I think we can use this framework to replace most of the tests inside `exchange_test.go`, `utils_test.go`, and `targeting_test.go`, I left all those in for now to prove that I haven't broken anything. If you all like the JSON-style tests overall, I can port them in a folllowup PR.

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -260,7 +260,7 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) error {
 
 func validateBidAdjustmentFactors(adjustmentFactors map[string]float64, aliases map[string]string) error {
 	for bidderToAdjust, adjustmentFactor := range adjustmentFactors {
-		if adjustmentFactor < 0 {
+		if adjustmentFactor <= 0 {
 			return fmt.Errorf("request.ext.prebid.bidadjustmentfactors.%s must not be negative. Got %f", bidderToAdjust, adjustmentFactor)
 		}
 		if _, isBidder := openrtb_ext.BidderMap[bidderToAdjust]; !isBidder {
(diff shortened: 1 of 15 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -223,10 +223,14 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) error {
 		return err
 	} else if bidExt != nil {
 		aliases = bidExt.Prebid.Aliases
-	}
 
-	if err := deps.validateAliases(aliases); err != nil {
-		return err
+		if err := deps.validateAliases(aliases); err != nil {
```
hhhjort: Always good to skip unneeded work (validating aliases or bid adjustments when none are supplied) ASAP :)

monofonik:
Is this new capability documented somewhere? The PR doesn't include any documentation changes from what I can tell. Cheers

dbemiller:
It's described by the last comment in the issue: https://github.com/prebid/prebid-server/issues/350#issuecomment-377330501

You're right, though... it should definitely be in here too: https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/auction.md

PR #432, opened 2018-04-02
Reverted the crid requirement temporarily

dbemiller:
Per #431, we've got to revert this temporarily so that Bidders have a breaking change window to implement this on their servers.

After a month or so, we can revert this and require it again.


dbemiller:
Yeah... that's a good question. I don't think it's realistic to expect bidders to see this in that ticket.

I think my plan is to make the PR which reverts this immediately, tag it with a "breaking change" label, and leave it open for a month until we merge it.

PR #441, opened 2018-04-04
Support some AMP query params

cirla:
Additional query params will be passed to AMP endpoint (see https://github.com/ampproject/amphtml/pull/14155/files).

This uses some of those parameters to fill/override values in the OpenRTB request.

### DIFF
--- a/docs/endpoints/openrtb2/amp.md
+++ b/docs/endpoints/openrtb2/amp.md
@@ -45,5 +45,6 @@ This endpoint supports the following query parameters:
 7. `purl` - the page URL
 8. `timeout` - the publisher-specified timeout for the RTC callout
    - A configuration option `amp_timeout_adjustment_ms` may be set to account for estimated latency so that Prebid Server can handle timeouts from adapters and respond to the AMP RTC request before it times out.
+9. `debug` - When set to `1`, will set `"test": 1` on outgoing OpenRTB requests and will return additional debug information in the response `ext`.
 
 For more information see [this pull request adding the query params to the Prebid callout](https://github.com/ampproject/amphtml/pull/14155) and [this issue adding support for network-level RTC macros](https://github.com/ampproject/amphtml/issues/12374).
--- a/docs/endpoints/openrtb2/amp.md
+++ b/docs/endpoints/openrtb2/amp.md
@@ -31,3 +31,19 @@ A sample response payload looks like this:
 
 In [the typical AMP setup](http://prebid.org/dev-docs/show-prebid-ads-on-amp-pages.html),
 these targeting params will be sent to DFP.
+
+### Query Parameters
+
+This endpoint supports the following query parameters:
+
+1. `h` - `amp-ad` `height`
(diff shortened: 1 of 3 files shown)


endpoints/openrtb2/amp_auction.go
```diff
@@ -276,6 +278,94 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
...
+		if req.Imp[0].Banner != nil {
+			*req.Imp[0].Banner.H = overrideHeight
+		}
+	}
+
+	multiSize := httpRequest.FormValue("ms")
+	if multiSize != "" {
+		// override w and h with format
+		req.Imp[0].Banner.H = nil
+		req.Imp[0].Banner.W = nil
```
dbemiller: I'm not sure this is quite right... at least, I'm finding some things which suggests that AMP treats these as "extra" sizes. For example: https://github.com/ampproject/amphtml/blob/master/examples/doubleclick-multisize.amp.html#L45

In OpenRTB, you're allowed to set `w`, `h`, _and_ `format`... so it might be better to leave these set?

endpoints/openrtb2/amp_auction.go
```diff
@@ -276,6 +278,94 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
...
+	}
+
+	canonicalURL := httpRequest.FormValue("curl")
+	if canonicalURL != "" {
+		// override purl
+		req.Site.Page = canonicalURL
+	}
+
+	if timeout, err := strconv.ParseInt(httpRequest.FormValue("timeout"), 10, 64); err == nil {
+		req.TMax = timeout
```
dbemiller: Do we want to use _exactly_ the AMP timeout here, or should we subtract off some time for the estimated latency?

I know we can't really estimate the latency in any quantitative way... but it seems like pubs would see much better performance overall if we did _something_ here. If our auctions time out at the same time the AMP call does, we'll probably miss a lot of bids.

Maybe worth knocking off 10-20ms here? Or make it configurable so that PBS hosts can decide based on their normal user geos & datacenter spread.

docs/endpoints/openrtb2/amp.md
```diff
@@ -31,3 +31,19 @@ A sample response payload looks like this:
...
+This endpoint supports the following query parameters:
+
+1. `h` - `amp-ad` `height`
+2. `w` - `amp-ad` `width`
+3. `oh` - `amp-ad` `data-override-height`
+4. `ow` - `amp-ad` `data-override-width`
+5. `ms` - `amp-ad` `data-multi-size`
+6. `curl` - the canonical URL of the page
+7. `purl` - the page URL
+8. `timeout` - the publisher-specified timeout for the RTC callout
```
dbemiller: Maybe worth moving `debug=1` into this list too?

I know it's mentioned above... but this list will pop out at people much more than the english paragraphs.

PR #433, opened 2018-04-02
Require Bid.CrID properties

dbemiller:
This PR includes a breaking change for Bidders.

This was originally added in #414, complained about in #427, and reverted in #432. It is scheduled to be merged sometime in early May, to allow Bidders time to make the necessary changes.

See the `docs/developers/add-new-bidder.md` file in the diff for a summary of the changes.

PR #444, opened 2018-04-04
Add E-Planning adapter

matimar:


adapters/eplanning/eplanning.go
```diff
@@ -0,0 +1,147 @@
...
+	return bids, nil
+}
+
+func toOpenRtbBids(ePlanningBids []*EPlanningBid, r *openrtb.BidRequest) []*adapters.TypedBid {
+	bids := make([]*adapters.TypedBid, 0, len(ePlanningBids))
+
+	for i, bid := range ePlanningBids {
+		if bid.Id != "" {
+			openRtbBid := openrtb.Bid{
+				ID:     bid.Id,
```
dbemiller: Doesn't look like you're setting a `CrID` here... so beware that #433 is coming soon. Once it merges, all your bids will be invalid.

static/bidder-params/eplanning.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "EPlanning Adapter Params",
+  "description": "A schema which validates params accepted by the EPlanning adapter",
+  "type": "object",
+  "properties": {
+    "ssp_space_id": {
+      "type": ["integer", "string"],
```
dbemiller: This doesn't line up with the code.

In `imp_eplanning.go`, you've got:

```
SspSpaceId int `json:"ssp_space_id"`
```

So... if someone sends a `string` here, the `Unmarshal` will fail.

static/bidder-info/eplanning.yaml
```diff
@@ -0,0 +1,9 @@
+maintainer:
+  email: "mmartinho@e-planning.net"
+capabilities:
+  app:
+    mediaTypes:
+      - banner
```
dbemiller: Looks like you support `banner` _and_ `video`?

adapters/eplanning/eplanning_test.go
```diff
@@ -0,0 +1,193 @@
...
+type aTagInfo struct {
+	code     string
+	bidfloor float64
+	instl    int8
+	price    float64
+	content  string
+	dealId   string
+}
+
+func TestOpenRTBRequest(t *testing.T) {
```
dbemiller: Please migrate these into `.json` file tests. For examples, see the Openx PR: https://github.com/prebid/prebid-server/pull/429/files#diff-c36dcdbd80e4c5ce762a0b64ba635256

See the rationale & details here: https://github.com/prebid/prebid-server/blob/master/adapters/adapterstest/test_json.go#L17

adapters/eplanning/eplanning.go
```diff
@@ -0,0 +1,147 @@
...
+	return requests, errors
+}
+
+func openRtbToEPlanningRequest(request *openrtb.BidRequest) (*EPlanningRequest, []error) {
+
+	adUnits := make([]*EPlanningAdUnit, 0, len(request.Imp))
+	errors := make([]error, 0, len(request.Imp))
+	for _, imp := range request.Imp {
+		var params openrtb_ext.ExtImpEPlanning
+		err := json.Unmarshal(imp.Ext, &params)
```
hhhjort: You are skipping a layer. You will want to first unmarshal the `bidder` element of `imp.Ext`, and then unmarshal your parameters from that.

```
                    errors = append(errors, err)
		    return nil, errors
	        }

      		var params openrtb_ext.ExtImpEPlanning
		err := json.Unmarshal(bidderExt.Bidder, &params)
		if err != nil {
			errors = append(errors, err)
			continue
		}
```

adapters/eplanning/eplanning.go
```diff
@@ -0,0 +1,147 @@
...
+	Banner     *openrtb.Banner
+}
+
+func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
+	ePlanningRequest, errors := openRtbToEPlanningRequest(request)
+	if len(ePlanningRequest.adUnits) == 0 {
+		return nil, errors
+	}
+
+	reqJSON, err := json.Marshal(request)
```
hhhjort: I think you want to marshal your `ePlanningRequest` here, not the initial request that was passed to you. Otherwise you don't seem to be using the generated `ePlanningRequest` for anything?

adapters/eplanning/eplanning_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	httpRequests, errs := bidder.MakeRequests(request)
+
+	if len(errs) > 0 {
+		t.Errorf("Got unexpected errors while building HTTP requests: %v", errs)
+	}
+	if len(httpRequests) != 1 {
+		t.Fatalf("Unexpected number of HTTP requests. Got %d. Expected %d", len(httpRequests), 1)
+	}
+
```
hhhjort: You don't seem to do any checks that you generated a valid ePlanning request. If you had, you would likely have caught my issues above. Moving to the `.json` file tests will provide this, as they will test against the expected request to be sent to your adapter as defined in your `.json` test file.

adapters/eplanning/eplanning.go
```diff
@@ -8,86 +8,109 @@ import (
...
+
+		requests = append(requests, &requestData)
+	}
+
+	return requests, errors
+}
+
+func verifyImp(imp *openrtb.Imp) (string, error) {
+	// We currently only support banner impressions
+	if imp.Native != nil || imp.Audio != nil || imp.Video != nil {
```
hhhjort: Multi-format Imps may be a thing some day. You might want to bid on the banner option of a multi-format. Also if a new type is developed some day, you will have to add it to your black list. You might want to consider simply rejecting all that have a non nil Banner object.

dbemiller: multi-format Imps _are_ a thing _today_ :)

adapters/eplanning/eplanning.go
```diff
@@ -0,0 +1,152 @@
...
+		addHeaderIfNonEmpty(headers, "DNT", strconv.Itoa(int(request.Device.DNT)))
+	}
+
+	imps := request.Imp
+
+	for source, impIds := range sourceMapper {
+		request.Imp = request.Imp[:0]
+
+		for i := 0; i < len(impIds); i++ {
+			request.Imp = append(request.Imp, imps[impIds[i]])
```
dbemiller: I _think_  there's a bug here, although I'm not certain of it. The `imps := request.Imp` line above only makes a copy of the _slice_--which is basically just a pointer.

So then `imps` and `request.Imp` are pointing to the same data. When you call:

```
request.Imp = request.Imp[:0]
request.Imp = append(request.Imp, {something})
```

then you're actually overriding the first element in memory. If you call `append` twice, you'd override the first and second.

As you loop through the elements, the data `imps` points to will be changing... which may then screw up your `imps[impIds[i]]` call, because you won't be fetching the `Imp` you think you are.

It would be very hard to pick this up by testing, because the `range map` iteration order isn't guaranteed, and you'd need the _second_ element it returns to use an `imp` from the _first_ index.

Does that make sense?

dbemiller: Here's a playground snippet which illustrates the issue: https://play.golang.org/p/MYF9LIOExiv

PR #446, opened 2018-04-05
Added metrics for the number of Imps

### LINKED ISSUE #445: add metric for number of imps per second
we have a metric for the number of requests per second, but it'd be nice to also know the number of imps that we are getting per second

dbemiller:
Fixes #445.

### DIFF
--- a/pbsmetrics/metrics.go
+++ b/pbsmetrics/metrics.go
@@ -1,11 +1,12 @@
 package pbsmetrics
 
 import (
+	"time"
+
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/rcrowley/go-metrics"
 	"github.com/vrischmann/go-metrics-influxdb"
-	"time"
 )
 
 // Labels defines the labels that can be attached to the metrics.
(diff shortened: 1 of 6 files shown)


pbsmetrics/metrics.go
```diff
@@ -113,7 +115,7 @@ const (
 // two groups should be consistent within themselves, but comparing numbers between groups
 // is generally not useful.
 type MetricsEngine interface {
-	RecordRequest(labels Labels)                           // ignores adapter. only statusOk and statusErr fom status
+	RecordRequest(labels Labels, numImps int)              // ignores adapter. only statusOk and statusErr fom status
```
hhhjort: Makes me second guess if the `RecordRequest` and `RecordAdapterRequest` sets of methods should be condensed into just the two functions or not. The split is based on each function storing one piece of event data, which in a tag based metrics system would be one metric with tags attached. Now `RecordRequest` is recording 2 metrics, requests, and the number of imps delivered. I think splitting them is simpler conceptually, but your example suggests condensing the types together may improve extensibility.

Not going to say anything is wrong with this PR, but gives food for thought for future enhancements and cleanup.

dbemiller: hmmm... I actually agree with this.

I don't know what the right path forward is... but I do know I should follow your design until we decide to change it. It's obvious in retrospect, but I didn't realize I was breaking convention.

Is fixed now.

PR #448, opened 2018-04-09
Adding HTTP fetcher

dbemiller:
For people who want to fetch Stored Requests through an API.


pbs_light.go
```diff
@@ -961,6 +962,13 @@ func NewFetchers(cfg *config.StoredRequests, db *sql.DB) (byId stored_requests.F
 		byAmpId = db_fetcher.NewFetcher(db, cfg.Postgres.MakeAmpQuery)
 		ampIdList = append(ampIdList, byAmpId)
 	}
+	if cfg.HTTP != nil {
+		glog.Infof("Loading Stored Requests from Postgres. endpoint=%s, amp_endpoint=%s", cfg.HTTP.Endpoint, cfg.HTTP.AmpEndpoint)
```
cirla: s/Postgres/HTTP

dbemiller: derp... thanks

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func unpackResponse(resp *http.Response) (requestData map[string]json.RawMessage, impData map[string]json.RawMessage, errs []error) {
+	respBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		errs = append(errs, err)
+		return
+	}
+	defer resp.Body.Close()
```
hhhjort: So we do not need to close if there was an error in reading the resp.Body?

dbemiller: mmm.... no, I think that's a bug. Is fixed now.

PR #450, opened 2018-04-10
Adding an HTTP Listener

dbemiller:
This follows up on #404 by adding a file which populates the Stored Request caches via HTTP endpoints.

All the Stored Requests are fetched on startup, and then refreshed periodically.


stored_requests/events/events.go
```diff
@@ -60,10 +60,10 @@ func (e *EventListener) Stop() {
 func (e *EventListener) Listen(cache stored_requests.Cache, events EventProducer) {
 	for {
 		select {
-		case update := <-events.Updates():
-			cache.Update(context.Background(), update.Requests, update.Imps)
-			if e.onUpdate != nil {
-				e.onUpdate()
+		case save := <-events.Saves():
+			cache.Save(context.Background(), save.Requests, save.Imps)
+			if e.onSave != nil {
```
cirla: Should we remove the Update method from the Cache altogether if it's not being used or should we expose Save events in addition to Update events and have producers decide which to use?

stored_requests/events/http/http.go
```diff
@@ -0,0 +1,173 @@
...
+	return e
+}
+
+type httpEvents struct {
+	client        *httpCore.Client
+	ctxProducer   func() (ctx context.Context, canceller func())
+	endpoint      string
+	invalidations chan events.Invalidation
+	lastUpdate    time.Time
+	updates       chan events.Save
```
cirla: If we're removing updates in favor of saves there are a few instances in this file where `updates` is referenced where it's actually a save.

stored_requests/events/events.go
dbemiller: yeah, I'm fine with that. Dead code is almost always more confusing than useful. "You ain't gonna need it" and all that.

stored_requests/events/http/http.go
dbemiller: good call, thanks... updated now.

stored_requests/events/events.go
dbemiller: I chatted with Hans offline... we agreed that deleting was fine.

PR #454, opened 2018-04-12
Analytics packages which are... modular

dbemiller:
This is a strict refactoring... no functional code changes here.

I need to do something like this because there's a dependency issue with the current structure.

Today, the `analytics` package contains:

1. Data objects (like `AuctionObject`)
2. An interface whose functions use those Data objects in the contracts
3. A "constructor" function (`NewPBSAnalytics`) that decides which analytics impls to use.

Between these three, implementations _must_ go into this same package. Otherwise `NewPBSAnalytics` depends on the impl, and the `func (i *imp) SomeFunc(a *AuctionObject)` definitions make the `impl` depend on the `AuctionObject`--which causes a cycle

To fix this, I moved `NewPBSAnalytics` into its own module. Since its job is primarily to turn a `Configuration` into an `Impl`, I named the module `analytics/config`. Now we can put implementations into their own packages... and I moved the `file_module` to prove it.

Wanted to get this sorted out before I started adding a new analytics module.


analytics/config/config.go
```diff
@@ -0,0 +1,48 @@
+package config
+
+import (
+	"github.com/golang/glog"
+	"github.com/prebid/prebid-server/analytics"
+	"github.com/prebid/prebid-server/analytics/filesystem"
+	"github.com/prebid/prebid-server/config"
```
hhhjort: This package's name is also "config" ... It does work, but may cause a little confusion. I don't know a better solution right now though.

analytics/config/config.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+//Modules that need to be logged to need to be initialized here
+func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
+	modules := make(enabledAnalytics, 0)
+	if len(analytics.File.Filename) > 0 {
```
hhhjort: Is it possible for analytics.File to be null? Looking at the definition, I don't think so, as there are no pointers in that chain. Might be worth a test if you haven't tested yet.

analytics/config/config.go
dbemiller: yeah... I'm in the same situation there. The purpose of the package would literally be "to build up the implementations based on the config"... I could call it `construction`? `instantiation`? `building`?

There are a bunch of packages which follow this format, and where this pattern could make sense (fetchers, caches, adapters, httprouting...), so let me know if you do come up with something you like more.

analytics/config/config.go
dbemiller: It can't. Go won't compile if you say `struct.NonPointerField = nil`, because `nil` isn't assignable to structs. If you don't set it, you get the zero object-- a non-nil struct whose properties are all zero objects.

PR #455, opened 2018-04-12
Rubicon adapter: Algorithm for multiple ad sizes

cindyhere33:
Implemented rubicon specific algorithm for multiple ad sizes


adapters/rubicon/rubicon.go
```diff
@@ -190,11 +190,31 @@ var rubiSizeMap = map[rubiSize]int{
...
-	if sz, ok := rubiSizeMap[rubiSize{w: uint16(s.W), h: uint16(s.H)}]; ok {
-		return sz, nil
+//MAS algorithm
+func findPrimary(alt []int) (int, []int) {
+	min, pos, primary := 0, 0, 0
+	for i, size := range alt {
+		if size == 15 {
+			primary = 15
+			pos = i
+			break
```
bretg: why does this one break while size 9 and 2 don't?

adapters/rubicon/rubicon_test.go
```diff
@@ -411,6 +414,132 @@ func TestParseSizes(t *testing.T) {
...
+				SIZE_ID[2],
+			},
+			output{2, []int{33}, false},
+		},
+		{
+			[]openrtb.Format{
+				SIZE_ID[8],
+			},
+			output{8, []int{}, false},
+		},
```
bretg: This is a good set of tests -- suggest adding one more which is a list of sizes that's not 2,9,15.

adapters/rubicon/rubicon.go
cindyhere33: this is for selection of primary ad size. When we encounter a 15, we take 15 for primary and break. But for 9 and 2, there is still a chance that 15 might occur much later in the array so we don't break immediately.

adapters/rubicon/rubicon_test.go
cindyhere33: That one's already covered in https://github.com/prebid/prebid-server/blob/36b03e07d07876d78532fb33fc19e0123269a7fe/adapters/rubicon/rubicon_test.go#L815

bretg:
Ok - signing off.

PR #463, opened 2018-04-17
Expand in-memory cache to allow unbounded sizes

dbemiller:
As the title says...

Now that we have Events for concurrent updates & invalidations, there may be good reason to use a stable, unbounded "map" as their cache.

This change lets people pass 0 or negative sizes to get an in-memory cache with no size limit.


dbemiller:
ok... I think this design works better.


I thought about it a bit... it makes much more sense to reuse the existing config values. If the `in_memory_cache` is configured with no TTL or max size, then the `sync.Map` implementation will be used behind the scenes.

With this design, that `cachetest` package is not strictly necessary (yet). However, I left it as is because I do think it's a better structure, and we have some other test-related files (e.g. `mockCache` inside the `exchange`) that should move there in the future.

stored_requests/caches/cachestest/reliable.go
```diff
@@ -0,0 +1,112 @@
+package cachestest
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
```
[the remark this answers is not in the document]
dbemiller: That I'm not sure of. I know it compiles files with the name `_test.go` into a separate pacakge secretly, and I _think_ it leaves out packages which are never imported anywhere by "prod" packages.

I stole this pattern & naming convention from `httptest`: https://golang.org/pkg/net/http/httptest/

It's not _quite_ the same use... but it's similar. `httptest` contains some `http` stuff which _other_ packages are intended to use in their `*_test.go` files, whereas these are some utilities used by the `*_test.go` files in the `caches` package, but... I still think it's as good a place to put the shared code as any other.

stored_requests/caches/memory/cache_test.go
```diff
@@ -0,0 +1,99 @@
...
+	for i := 0; i < numWrites; i++ {
+		cache.Save(context.Background(), mapForVal(i), mapForVal(-i))
+	}
+	done <- s
+}
+
+func invalidateLots(cache stored_requests.Cache, done chan<- struct{}, numWrites int) {
+	var s struct{}
+	for i := 0; i < numWrites; i++ {
+		cache.Invalidate(context.Background(), sliceForVal(i), sliceForVal(i))
```
hhhjort: read and write you use `i, -i`, why are you using `i, i` here?

dbemiller: no good reason... just thought I may as well invalidate half the values to broaden the test scope a bit.

I can see how it'd be confusing though. D'you think I should change it, or add a comment explaining?

stored_requests/caches/cachestest/reliable.go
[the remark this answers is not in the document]
hhhjort: As long as we don't import `cachetest` into a non-cahce file, I imagine it should be good.

stored_requests/caches/memory/cache_test.go
hhhjort: With imp invalidation taking positive `i`, we will never invalidate an imp we ever set. Not sure if that is supposed to be a feature of the test or not.

It might make sense to shuffle an array of integers of length `numWrites` for each of the three functions, so as to mix reads, writes, and invalidates. If read, write, and invalidate all take the same time to execute, then I would exepect the current code would write `i`, read`i`, and then invalidate `i`. And we would just test that each one of these can be done while the other types of commands are working on other ids. And if reads are faster than writes, we will soon get to the point where we are reading unset data all the time.

Might make sense to shuffle, and have 3 read routines and 3 write routines to cover all possibilities. I don't think multiple invalidates will be needed, as they will be rare and unlikely to run in parallel.

dbemiller: these sound like reasonable ideas... done now (if I understood your suggestion properly).

For what it's worth, though... while working on this, I wrote a version with a race condition (using a `map` with no `RWMutex`) and the race detector seemed to pick it up 5 times in a row, with no failures. So we're probably safe.

stored_requests/caches/memory/cache_test.go
```diff
@@ -54,36 +49,44 @@ func TestRaceUnboundedConcurrency(t *testing.T) {
...
 
+	doRaceTest(t, cache)
+}
+
+func doRaceTest(t *testing.T, cache stored_requests.Cache) {
 	done := make(chan struct{})
-	go writeLots(cache, done, 100)
-	go readLots(cache, done, 100)
-	go invalidateLots(cache, done, 100)
+	sets := [][]int{rand.Perm(100), rand.Perm(100), rand.Perm(100)}
```
hhhjort: I would create separate random sets for each goroutine, otherwise not much benefit beyond going in order, as they are all still going in the same order, just not the 1, 2, 3, 4 order.

PR #467, opened 2018-04-18
Minimize log messages

### LINKED ISSUE #142: Improve error handling and reporting
Adapters return errors for two reasons.

1. The user's request was bad (missing/invalid params)
2. An actual error occurred (`json.Marshal` failed, the remote server was down, etc)

Right now, there's no way to distinguish between these errors, so we're [logging everything](https://github.com/prebid/prebid-server/blob/master/pbs_light.go#L323).

We should introduce a specific error type for (1), and then only log errors which are _not_ of that type. That should prevent log pollution with "error" messages like this:

```
W1003 19:47:37.726751      14 pbs_light.go:323] Error from bidder appnexus. Ignoring all bids: json: cannot unmarshal string into Go struct field appnexusParams.placementId of type int
```

dbemiller:
Fixes #142 by making the logging much less verbose.

I don't intend for this to have any functional changes on adapters. I just wanted to make the errors they return be more semantically correct so that our core code can handle them differently.

For now, that just involves silencing the logs if the errors don't give the PBS host any action items.


hhhjort:
So this is just for legacy adapters?

dbemiller:
Sort of.

Legacy is flooding the logs because of this: https://github.com/prebid/prebid-server/pull/467/files#diff-444feda808b76687bcc8827c1025876cL354

Most of the adapters respond to OpenRTB requests through legacy code paths, though.

Once the logs are silent (or rather, sensible), I actually plan to _add_ a log message to the OpenRTB code paths. Currently that code returns all errors rather than logging them.

hhhjort:
I notice you only filter the error messages in the legacy auctions, so this won't reduce any log spam for openrtb auctions.

dbemiller:
Yep yep. That's just cause OpenRTB auctions aren't logging anything now :)

Issue #420, opened 2018-03-21
Additional parameters for AMP RTC 


bretg:
AMP RTC supports a number of dynamic parameters that Prebid Server could/should support. Here's the URL that's been configured in AMP:

```
/openrtb2/amp?tag_id=REQUEST_ID&w=ATTR(width)&h=ATTR(height)&ow=ATTR(data-override-width)&oh=ATTR(data-override-height)&ms=ATTR(data-multi-size)&slot=ATTR(data-slot)&targeting=TGT&curl=CANONICAL_URL&timeout=TIMEOUT&adc=ADCID&purl=HREF'
```
Specifically, the following are desired for this release: w, h, ms, curl, slot, timeout:

- w add into the openrtb packet at imp.banner.format[0].w
- h add into the openrtb packet at imp.banner.format[0].h
- ms (multiple-sizes) - takes values like "970x90, 728x90". Parse sizes and add to imp.banner.format array
- curl add as site.page
- slot add as imp.tagid
- timeout add as tmax
- targeting - this one's bigger. It's for passing first party data in, but will be a separate issue.
- ow, oh, and purl are TBD - also for a separate issue

bretg:
So we're implementing this on the Java side, and there's a details in the Go implementation that I'm not sure is quite right. Specifically, setting imp.banner.h and imp.banner.w

OpenRTB 2.5 recommends that the format be specified, and having both format and h/w be specified is unclear. The standard seems to say that h/w are only used if format isn't specified, but SSPs may have implemented that differently. I think it's cleaner to specify only the format object.

So this is what we're building into Java:

1) only imp.banner.format[] is ever set. Never imp.banner.w/h
2) if oh/ow is defined, they are set as imp.banner.format[0] and ms/w/h are ignored
3) else if ms is defined, it's used to set imp.banner.format[]
4) else use w/h define imp.banner.format[0]

dbemiller:
> The standard seems to say that h/w are only used if format isn't specified

Where does it say that?

bretg:
https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf section 3.2.6: format is the `recommended` field.

```
Array of format objects (Section 3.2.10) representing the
banner sizes permitted. If none are specified, then use of the
h and w attributes is highly recommended.
```
And then both `w` and `h` say 
```
recommended if no format objects are specified.
```
[...]

PR #470, opened 2018-04-19
Rework AMP size-related query param handling

cirla:
Address feedback on #420


dbemiller:
Beware that some docs changes are coming in #472 which don't quite align with this.

At the very least, we'll want to remove the `format` from that sample Stored Request payload when this goes in, because (per #420), the sizes sent by AMP will always override the Stored Request sizes anyway.

cirla:
Closing in favor of #534

PR #471, opened 2018-04-19
metrics code cleanup and better logging

hhhjort:
Cleaned u some metrics code, shouldn't change anything or fix the blank metrics issue, but perhaps if there is some subtle issue I cannot see.

Also improved the logging to help narrow down where/when the blank metrics are happening.

### DIFF
--- a/pbs_light.go
+++ b/pbs_light.go
@@ -311,9 +311,9 @@ func (deps *auctionDeps) auction(w http.ResponseWriter, r *http.Request, _ httpr
 				CookieFlag:    labels.CookieFlag,
 				AdapterStatus: pbsmetrics.AdapterStatusOK,
 			}
-			if blables.Adapter == "" {
+			if blabels.Adapter == "" {
 				// "districtm" is legal, but not in BidderMap. Other values will log errors in the go_metrics code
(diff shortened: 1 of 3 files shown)


pbs_light.go
```diff
@@ -312,7 +311,10 @@ func (deps *auctionDeps) auction(w http.ResponseWriter, r *http.Request, _ httpr
 				CookieFlag:    labels.CookieFlag,
 				AdapterStatus: pbsmetrics.AdapterStatusOK,
 			}
-			bidderLabels[bidder.BidderCode] = &blabels
+			if blables.Adapter == "" {
```
dbemiller: eh... it might be more straightforward to just put this in the declaration: `Adapter: openrtb_ext.BidderName(bidder.BidderCode),`

That should have the same effect regardless of whether or not the code is defined in the map.

dbemiller:
...but there are some test failures :(

hhhjort:
Ack, typos, forgot to validate before committing.

pbs_light.go
hhhjort: Was worried if we ever have the case where BidderCode is not the same as the cannonical BidderName

PR #477, opened 2018-04-23
Added EventProducer using Postgres Listen/Notify

dbemiller:
As the title says.

This lets PBS listen for Postgres changes so that there's no lagtime waiting for the cache to expire before the app starts seeing the new data.


dbemiller:
Per some other feedback here... not going forward with this.

I'll leave this branch alive so that it can be picked back up and merged if anyone else wants it.

PR #481, opened 2018-04-25
Postgres polling EventProducer 

dbemiller:
This introduces two new EventProducers for Postgres. One fetches all the configs to populate the cache on startup, and the other pings the database for updates periodically to keep the caches up to date. The goal here is to poll the DB for updates periodically so that the in-memory cache doesn't prevent PBS from seeing updates for its entire TTL.

This is only a breaking change for hosts who are using Postgres as a Stored Request backend. The breaking change is in the config.

Before these changes:

```
stored_requests:
  postgres:
    dbname: "prebid"
    host: "localhost"
    port: 5432
    user: "some-user"
    password: "some-password"
    query: SELECT id, requestData, type FROM stored_requests WHERE id in %REQUEST_ID_LIST%;
    amp_query: SELECT id, requestData, type FROM amp_requests WHERE id in %REQUEST_ID_LIST%;
```

After these changes:

```
stored_requests:
  postgres:
    connection:
      dbname: "prebid"
      host: "localhost"
      port: 5432
      user: "some-user"
      password: "some-password"
    fetcher:
      query: SELECT id, requestData, type FROM stored_requests WHERE id in %REQUEST_ID_LIST%;
      amp: SELECT id, requestData, type FROM amp_requests WHERE id in %REQUEST_ID_LIST%;
```

A sample config using the new feature would look like this:

```
stored_requests:
  postgres:
    connection:
      dbname: "prebid"
      host: "localhost"
      port: 5432
      user: "some-user"
      password: "some-password"
    fetcher:
      query: SELECT id, requestData, type FROM stored_requests WHERE id in %REQUEST_ID_LIST%;
      amp: SELECT id, requestData, type FROM amp_requests WHERE id in %REQUEST_ID_LIST%;
    initialize_caches:
      timeout_ms: 5000
      query: SELECT id, requestData, type FROM stored_requests;
      amp_query: SELECT id, requestData, type FROM amp_requests;
    poll_for_updates:
      refresh_rate_seconds: 30
      timeout_ms: 1000
      query: SELECT id, requestData, type FROM stored_requests WHERE last_updated > $1;
      amp_query: SELECT id, requestData, type FROM amp_requests WHERE last_updated > $1;
```


stored_requests/config/config.go
```diff
@@ -99,11 +106,16 @@ func newCache(cfg *config.StoredRequests) stored_requests.Cache {
 		return &nil_cache.NilCache{}
 	}
 
-	glog.Infof("Using a Stored Request in-memory cache. Max size for StoredRequests: %d bytes. Max size for Stored Imps: %d bytes. TTL: %d seconds.", cfg.InMemoryCache.RequestCacheSize, cfg.InMemoryCache.ImpCacheSize, cfg.InMemoryCache.TTL)
+	if cfg.InMemoryCache.RequestCacheSize <= 0 && cfg.InMemoryCache.ImpCacheSize <= 0 && cfg.InMemoryCache.TTL <= 0 {
```
hhhjort: Should these log messages be in `memory.NewCache()` where the actual logic is implemented, rather than replicating that logic in this piece of code just for the purpose of logging this message?

dbemiller: Much cleaner this way, yeah.

Also improves the log messages, since I can split the lines describing Stored Requests and Stored Imps.

dbemiller:
I wish we could send out an email, but... we actually don't know who the PBS host companies are :/. We only have contact emails for the Bidders, who wouldn't really care about this for any reason.

I know you've suggested mailing lists in the past... I'm definitely in favor of those, and also making one for PBS host companies... but as of now, I'm not sure what we can do beyond add labels like this and clear upgrade instructions.

PR #486, opened 2018-04-30
Update test_json to work with non-OpenRTB bidders

FlorentFlament:
The code is generic enough to allow testing non-OpenRTB compliant
bidders. This patch mostly changes the type of the HttpResponse.Body
from openRTB.BidResponse to json.RawMessage, to avoid having the test
interpreting the response as an OpenRTB one. This wasn't required as
the response is interpreted appropriately by the adapter's MakeBids
method, which take a []byte as the body.

### DIFF
--- a/adapters/adapterstest/test_json.go
+++ b/adapters/adapterstest/test_json.go
@@ -14,7 +14,7 @@ import (
 	"net/http"
 )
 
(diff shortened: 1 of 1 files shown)


FlorentFlament:
thanks !

PR #492, opened 2018-05-03
Support not sending all bidder keys in targetinng

hhhjort:
Adds ext.prebid.targeting.includebidderkeys, defaults to true (current behavior) and supports setting it false. You will need includewinners set true (true by default) or no targeting set for a valid request.

### DIFF
--- a/exchange/targeting_test.go
+++ b/exchange/targeting_test.go
@@ -59,22 +59,6 @@ func assertKeyExists(t *testing.T, bid *openrtb.Bid, key string, expected bool) 
 	}
 }
 
-// Verify we do not set winning bid keys if includewinners is set false
-func TestExcludeWinners(t *testing.T) {
-	bids := runTargetingAuction(t, mockBids, false, false, true, false)
-
-	// Make sure that the cache keys exist on the bids where they're expected to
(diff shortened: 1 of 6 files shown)


exchange/targeting_test.go
```diff
@@ -57,9 +59,25 @@ func assertKeyExists(t *testing.T, bid *openrtb.Bid, key string, expected bool)
 	}
 }
 
+// Verify we do not set winning bid keys if includewinners is set false
```
dbemiller: Can we move this into an `exchange/exchangetest`? You can basically copy/paste/tweak the `targeting-with-winners.json` one.

I realized while working on some recent features that all these tests of private methods make the `exchange` code really hard to maintain. Since the `exchangetest` files are independent of `exchange` implementation, we can move around the private code without having to update the tests... which in turn makes refactors much safer.

exchange/targeting_test.go
```diff
@@ -15,28 +15,30 @@ import (
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
+// Using this set of bids in more than one test
```
dbemiller: You could probably undo all the changes to `exchange/targeting_test.go` if you wanted, but... *shrugs*. Not the biggest deal. Just more complicated now than it needs to be.

PR #493, opened 2018-05-07
Adform adapter: bid's crid and priceType request parameter

volhalink:


static/bidder-params/adform.json
```diff
@@ -7,6 +7,10 @@
     "mid": {
       "type": ["integer", "string"],
       "description": "An ID which identifies the placement selling the impression"
+    },
+    "priceType": {
+      "type": "string",
+      "description": "An expected price type (net or gross) of bids."
```
dbemiller: Hey, could you give us some rationale behind this?

When we were talking about this internally, we couldn't think of any good reasons for a Bidder to return Gross. There's a section in the [add a new bidder](https://github.com/prebid/prebid-server/blob/master/docs/developers/add-new-bidder.md#adding-a-new-bidder) doc on it:

> NOTE: To make everyone's lives easier, Bidders are expected to make Net bids (e.g. "If this ad wins, what will the publisher make?), not Gross ones. Publishers can correct for Gross bids anyway by setting Bid Adjustments to account for fees.

Would be curious to hear your thoughts on that.

-----------------------------------------------------------------

Assuming you do really want to keep this param, though, this should be an enum in the JSON schema: https://spacetelescope.github.io/understanding-json-schema/reference/generic.html#enumerated-values

BZSky: Hey! About price type param – we are going to add description of it into s2s config instructions on prebid.js readme.md after this will merge. There, we will point out that "net" is expected to be used for header bidding. 

The rationale is that our delivery chain is heavily based on gross prices, and we added net price option to our ad server specifically for header bidding. Using it as the general default is not an option currently. On top of that, we cannot be sure that we can differentiate all header bidding requests from non-HB ones, to set net prices for HB traffic automagically, but instead, we are strongly participating in header bidding configs on publisher pages, thus we are able to communicate with our clients to use correct priceType parameter, leaving safe gross as a default.

---

adapters/adform/adform.go
```diff
@@ -317,6 +352,12 @@ func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []erro
 			secure = true
 		}
 
+		_, valid := isPriceTypeValid(adformAdUnit.PriceType)
```
dbemiller: You can remove this logic from the OpenRTB code path now. Those JSON schemas are enforced, so PBS will return a 400 before it even calls your Bidder if it's invalid.

You _do_ need to validate on the legacy path above, though.

static/bidder-params/adform.json
dbemiller: ok... that makes sense.

Since Prebid Server is a header bidding library, I'm a little confused why you wouldn't just hardcode it into your adapter... seems like exposing the option to publishers just introduces room for errors.

There's no real policy against it though, so... *shrugs*. Thanks for the explanation.

PR #494, opened 2018-05-07
Fixing inventory/visitor data types

bretg:
This patch fixes a valid openrtb request that currently fails with a data type error:
```
{
    "id": "7b3e93dc-10cc-42cd-b855-76ddea2e3f7d",
    "source": {
        "tid": "7b3e93dc-10cc-42cd-b855-76ddea2e3f7d"
    },
    "tmax": 1000,
    "imp": [
        {
            "id": "test-div",
            "ext": {
                "rubicon": {
                    "accountId": 1001,
                    "siteId": 113932,
                    "zoneId": 535510,
                    "inventory": {
                        "rating": [
                            "5-star"
                        ],
                        "prodtype": [
                            "tech",
                            "mobile"
                        ]
                    },
                    "visitor": {
                        "ucat": [
                            "new"
                        ],
                        "search": [
                            "iphone"
                        ]
                    }
                }
            },
            "secure": 0,
            "banner": {
                "format": [
                    {
                        "w": 300,
                        "h": 250
                    },
                    {
                        "w": 300,
                        "h": 600
                    }
                ]
            }
        }
    ],
    "test": 1,
    "site": {
        "publisher": {
            "id": "1001"
        },
        "page": "http://localhost/prebid_server_test_1001_gdpr.html"
    },
    "user": {
        "ext": {
            "digitrust": {
                "id": "B+pL2o/72Ece/VNhE1FjdK1hDI99YIQ5DAN/BQTjdgWKiqKIkh49YTC4jJnwuXXxC9Chg036stdDBpJujn0z9pt4D4xEwgtoJQof6vUD5NGZdA8m4k+bb6NeJloUMi5jz4hb60FMUw/FfGO7Uk7stupS5MV83vmdZjV0ebaJ1ClU9DDGUA59RghiHNnGrgzscX93TztObnbVLazeh51tf0Sfg7+vZqUvFjYOiqfrYI0hxgUnMcMbKDnhs2B7PRiMrerljqq5/c3jALl0anb+12lICQtyjod7vHE/51KN22RwRycjbziJI4HBpuCJCMLvlKRKGaX8rRLjtu+EWWTfLQ==",
                "keyv": 4,
                "pref": 0
            }
        }
    }
}
```

### DIFF
--- a/static/bidder-params/rubicon.json
+++ b/static/bidder-params/rubicon.json
@@ -23,14 +23,14 @@
       "type": "object",
       "description": "An object defining arbitrary targeting key/value pairs related to the page",
       "additionalProperties": {
-        "type": "string"
+        "type": "array"
       }
     },
     "visitor": {
       "type": "object",
       "description": "An object defining arbitrary targeting key/value pairs related to the visitor",
       "additionalProperties": {
-        "type": "string"
+        "type": "array"
       }
     },
     "video": {


bretg:
@dbemiller - this isn't a breaking change - it's a bugfix. Currently the parameters described at http://prebid.org/dev-docs/bidders.html#rubicon don't work for the openrtb endpoint. Valid openrtb requests like the above fail with:
```
Invalid request format: request.imp[0].ext.rubicon failed validation.
inventory: Invalid type. Expected: string, given: array
inventory: Invalid type. Expected: string, given: array
visitor: Invalid type. Expected: string, given: array
visitor: Invalid type. Expected: string, given: array
```

bretg:
I've tested this on our fork and both legacy and openrtb correctly pass visitor and inventory values with this change.

dbemiller:
It certainly _could be_ a breaking change. In the `prebid.adnxs.com` UI, publishers can enter any JSON they want for AMP. This would break the request of anybody using this param.

I'll grant that it's pretty unlikely, though, and probably wouldn't affect many (if any) pubs. We have no real way of knowing, though.

bretg:
If a pub has entered objects-with-string-attributes into `prebid.adnxs.com` for Rubicon openrtb requests, those requests are failing with an error. Making this change will fix that error.

The legacy endpoint supports either objects-with-string-attributes or objects-with-array-attributes (converts the strings to arrays as needed) -- or in any case, it's not affected by the change to the schema.

dbemiller:
Errors from inside the Rubicon adapter, or from your server, would cause a `response.ext.errors.rubicon` message in the response, but other bidders will still work as normal.

Requests rejected by the `bidder-params/rubicon.json` file will cause a 400 response for the whole auction, so no bids from _any_ adapter will come back.

To be clear: I support this breaking change. It's a v0.x app and the alternative would be making the code accept both types, which is worse overall. Better to just rip off the band-aid now. But it definitely could make some publisher requests perform worse than they used to.

bretg:
Merging.

PR #497, opened 2018-05-08
GDPR support for legacy adapters

pycnvr:
Please review the following change that passes along the two GDPR parameters for legacy adapters.  Essentially the adapters do not need to change as long as the openrtb request generated by MakeOpenRTBGeneric is used directly.

* Add Regs to PBSRequest
* Pass along Regs in legacy conversion
* In MakeOpenRTBGeneric, picks up User.Ext if available, and also picks up Regs.

### DIFF
--- a/adapters/openrtb_util.go
+++ b/adapters/openrtb_util.go
@@ -148,11 +148,17 @@ func MakeOpenRTBGeneric(req *pbs.PBSRequest, bidder *pbs.PBSBidder, bidderFamily
 			},
 			AT:   1,
 			TMax: req.TimeoutMillis,
+			Regs: req.Regs,
 		}, nil
 	}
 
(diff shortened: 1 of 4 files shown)


dbemiller:
@bretg take note of this one... this affects the legacy API for GDPR.

pycnvr:
Great!  Thanks.

dbemiller:
@pycnvr Beware the nuances discussed in https://github.com/prebid/prebid-server/issues/501

The legacy `/auction` endpoint returns cookie sync URLs as well, so it still may not be GDPR-safe, depending on what the Client is doing.

 I can still merge this because I don't think it hurts anything, but there might need to be more work here before we can really call this endpoint "GDPR-safe".

PR #500, opened 2018-05-10
Fixed some nil-pointers in the Adform adapter

dbemiller:
Saw some errors about nil-dereferencing in the logs recently... this fixes those.

Also includes some updates to the `test_json` file because the adform request body is empty, and the JSON tests didn't handle that very well.

@volhalink @vstatkevich I'm not sure what request you actually expect on your server if the `User` or `Device` are nil. You should check this and update it. This just stops your code from crashing if they are.

### DIFF
--- a/adapters/adform/adform.go
+++ b/adapters/adform/adform.go
@@ -369,14 +369,42 @@ func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []erro
 
 	return &adformRequest{
 		adUnits:       adUnits,
-		ip:            request.Device.IP,
-		advertisingId: request.Device.IFA,
-		userAgent:     request.Device.UA,
+		ip:            getIPSafely(request.Device),
+		advertisingId: getIFASafely(request.Device),
+		userAgent:     getUASafely(request.Device),
 		isSecure:      secure,
 		referer:       referer,
-		userId:        request.User.BuyerUID,
+		userId:        getBuyerUIDSafely(request.User),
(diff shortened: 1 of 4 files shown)


adapters/adform/adform.go
```diff
@@ -369,16 +369,44 @@ func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []erro
...
+		userAgent:     getUASafely(request.Device),
 		isSecure:      secure,
 		referer:       referer,
-		userId:        request.User.BuyerUID,
+		userId:        getBuyerUIDSafely(request.User),
 		tid:           tid,
 	}, errors
 }
 
+func getIPSafely(device *openrtb.Device) string {
```
cirla: I wish there were a simpler way to do this generically besides a `reflect` hack like:

```go
func getStringFieldSafely(i interface{}, field string) string {
    if i == nil {
        return ""
    }

    v := reflect.ValueOf(i)
    f := reflect.Indirect(v).FieldByName(field)
    return f.String()
}

...

ip:            getStringFieldSafely(request.Device, "IP"),
advertisingId: getStringFieldSafely(request.Device, "IFA"),
userAgent:     getStringFieldSafely(request.Device, "UA"),
userId:        getStringFieldSafely(request.User, "BuyerID"),
```

dbemiller: you and me both :(... not sure if I'm just bad with Go and there are better ways to do things, but this language is painful sometimes.

PR #505, opened 2018-05-10
Added GDPR to the app config

### LINKED ISSUE #503: GDPR cookie sync configuration
This issue is for adding GDPR support to the Prebid Server config. For an overview of GDPR support in Prebid Server, see #501

# Requirements

Prebid Server host companies must be able to:

1. Identify themselves as one of the members of the [global vendors list](https://vendorlist.consensu.org/vendorlist.json)
2. Define how Prebid Server should behave if the Request doesn't carry any GDPR info in it.

# Working design

We plan to add the following config options:

```
gdpr:
  host_vendor_id: 23
  usersync_if_ambiguous: true
```

If these are undefined, we will assume that the publisher doesn't care about GDPR and continue to sync users all the time.

If defined:
- `host_vendor_id` must be an integer > 0 which identifies the Prebid Server host company in the Global Vendors List.
- `usersync_if_ambiguous` will determine whether or not usersyncs are done if the incoming request has no `gdpr` info.

`usersync_if_ambiguous` is optional, and will be treated as `false` if `host_vendor_id` is defined.

See #504 for changes to the core logic in the endpoints.

dbemiller:
Fixes #503.

No functional changes here... just getting the app config ready for more GDPR support.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -27,13 +27,32 @@ type Configuration struct {
 	MaxRequestSize       int64              `mapstructure:"max_request_size"`
 	Analytics            Analytics          `mapstructure:"analytics"`
 	AMPTimeoutAdjustment int64              `mapstructure:"amp_timeout_adjustment_ms"`
+	GDPR                 GDPR               `mapstructure:"gdpr"`
 }
 
 func (cfg *Configuration) validate() error {
 	if cfg.MaxRequestSize < 0 {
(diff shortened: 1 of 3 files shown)


mkendall07:
Looks like you have a test failure.

dbemiller:
d'oh... thanks. I fixed that earlier this morning, but that must have typo'd my `git push branch`.

Issue #504, opened 2018-05-10
GDPR Endpoint Logic


dbemiller:
This issue describes incoming changes to the cookie sync endpoints for GDPR support. For an overview of GDPR in Prebid Server, see #501.

For the complementary changes to Prebid.js, see https://github.com/prebid/Prebid.js/issues/2516.

# Changes to /cookie_sync

This endpoint will respect two new optional parameters.

```
{
  "gdpr": 1,
  "gdpr_consent": "BONciguONcjGKADACHENAOLS1rAHDAFAAEAASABQAMwAeACEAFw"
}
```

`gdpr` should be `1` if GDPR is in effect, `0` if it's not, and undefined if clients are unsure.

`gdpr_consent` should be the [unpadded base64-URL](https://tools.ietf.org/html/rfc4648#section-5) 
 encoded [vendor consent string](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/Consent%20string%20and%20vendor%20list%20formats%20v1.1%20Final.md#vendor-consent-string-format-). It is required if `gdpr` is `1`, and optional otherwise. If `gdpr` is undefined, it is optional, but highly encouraged.

# Changes to /setuid

This endpoint will respect the same values as `/cookie_sync`, but as query params.

For example: `GET /setuid?bidder=adnxs&uid=12345&gdpr=1&gdpr_consent=BONciguONcjGKADACHENAOLS1rAHDAFAAEAASABQAMwAeACEAFw`

# Behavior

**If `gdpr` is `0`**: The `gdpr_consent` string will be ignored, and both endpoints will work exactly like they do today.

**If `gdpr` is `1`**: Both endpoints will no-op unless the`gdpr_consent` string gives the PBS host permission to save cookies. The `/cookie_sync` endpoint will only return syncs for bidders with permission for purpose 3 ("Ad selection, delivery, reporting")

**If `gdpr` is not defined**: The Prebid Server host can configure the behavior to assume that GDPR is in effect or not (see #503). If they think GDPR is in effect, behave like a `1`. If not, behave like a `0`. 

Callers should always send in the `gdpr_consent` string if it's available because it will increase your cookie sync rate with publishers who are playing it safe legally.

# Questions and complex details

1. Will the endpoints delete the cookie if the host company doesn't have consent?

[...]

PR #512, opened 2018-05-17
GDPR-friendly /cookie_sync

dbemiller:
This partially implements #504, updating the `/cookie_sync` endpoint. See the issue (or the docs in the PR) for the behavior.

It does _not_ yet forward the gdpr/consent info into the URLs so that it gets passed back into `/setuid`.  I believe those changes will mostly be in the `Usersyncer` interface, so I'll save them for the next changeset.


endpoints/cookie_sync.go
```diff
@@ -1,112 +1,166 @@
...
+		co.Errors = append(co.Errors, errors.New("Failed to read request body"))
+		http.Error(w, "Failed to read request body", http.StatusBadRequest)
+		return
+	}
+	biddersJSON, err := parseBidders(bodyBytes)
+	if err != nil {
+		co.Status = http.StatusBadRequest
+		co.Errors = append(co.Errors, errors.New("Failed to check request.bidders in request body. Was your JSON well-formed?"))
+		http.Error(w, "Failed to check request.bidders in request body. Was your JSON well-formed?", http.StatusBadRequest)
+	}
```
hhhjort: Should there be a return here? If we are declaring an error condition in the response here, what use is further processing?

gdpr/impl_test.go
```diff
@@ -183,3 +183,10 @@ func assertBoolsEqual(t *testing.T, expected bool, actual bool) {
 		t.Errorf("Expected %t, got %t", expected, actual)
 	}
 }
+
+func assertStringsEqual(t *testing.T, expected string, actual string) {
```
hhhjort: Is this function being used? This is the only change in the file, and it is a pure addition, so `assertStringsEqual()` couldn't have been there before.

usersync/usersyncers/syncer.go
```diff
@@ -26,6 +26,16 @@ func NewSyncerMap(cfg *config.Configuration) map[openrtb_ext.BidderName]usersync
 	}
 }
 
+func ParseIdsGDPR(syncers map[openrtb_ext.BidderName]usersync.Usersyncer) map[openrtb_ext.BidderName]uint16 {
+	gdprAwareSyncers := make(map[openrtb_ext.BidderName]uint16, len(syncers))
```
hhhjort: Not entirely happy with this name, as to me it implies the values should be  syncers, not GDPR IDs. perhaps rename to `gdprAwareSyncerIDs`?

hhhjort: Hah, actually referred to the first variable in the function rather than the function name itself, but this is good too. :)

PR #514, opened 2018-05-18
Enhances the API used by the http stored request fetcher

hhhjort:
This is a breaking change for the API the http stored request fetcher uses. As it is not a user facing API change, I am not sure if this should have the "breaking api change" tag or not.

Oh, possibly needs docs ... I am not seeing any docs currently describing the http fetcher API, so maybe?


dbemiller:
I split the "breaking change" labels up by "adapter", "host", and "publisher" for this reason.

This would be a breaking change for any hosts who are using this module through the app config. Given how new it is, though, I'd be surprised if that were anyone but us. Still might be a good idea to let it sit for a little while to let anyone complain if it affects them.

PR #519, opened 2018-05-21
Removed legitimate interest, updated "allow sync" logic

dbemiller:
This makes two changes based on recent finds:

1. It removes the logic which lets syncs happen based on legitimate interest.
2. It updates the logic so that bidder syncs happen if they have consent _to access the cookie_--not based on ad personalization.

------------------------------------------------------------------------------------------
## Why ignore legitimate interest? 

There are some [new pubvendors.json docs](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/pubvendors.json%20v1.0%20Draft%20for%20Public%20Comment.md#sspdsp-integration-), which say

> If a publisher whitelists a vendor claiming legitimate interest then Legitimate Interest can take precedence over the information in the Consent String

I thought legitimate interest took precedence over consent... but it looks like this `pubvendors.json` spec gives publishers a way to "object" to vendors who claim legitimate interest that they don't trust.

Since we aren't supporting `pubvendors.json` (at least yet), we can't really support this... so this plays it safe and limits syncs to explicit consent.

If this upsets Bidders, we could let Usersyncers whose sync endpoints understand consent strings bypass this check... but based on lots of conversations I had (and the changes for (2)), I don't think it'll be an issue.

------------------------------------------------------------------------------------------
## Why consider Cookie Access for sync URLs?

Nearly every Usersync endpoint reads from the Cookie. I feel dumb for not realizing this. But as a result, it really doesn't make sense to drop a sync URL for someone who can't read from the Cookie.

Similarly, ad targeting & personalization have nothing to do with the sync itself. That only matters during the Auction, which has its own consent string.

### DIFF
--- a/gdpr/impl.go
+++ b/gdpr/impl.go
@@ -23,38 +23,24 @@ type permissionsImpl struct {
 }
 
 func (p *permissionsImpl) HostCookiesAllowed(ctx context.Context, consent string) (bool, error) {
-	// If we're not given a consent string, respect the preferences in the app config.
-	if consent == "" {
-		return p.cfg.UsersyncIfAmbiguous, nil
-	}
-
-	data, err := base64.RawURLEncoding.DecodeString(consent)
-	if err != nil {
-		return false, err
-	}
-
-	parsedConsent, err := vendorconsent.Parse([]byte(data))
-	if err != nil {
-		return false, err
-	}
-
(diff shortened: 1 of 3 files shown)


dbemiller:
hold off on merging this for a bit... I'm going to talk this over with some more people today.

Issue #518, opened 2018-05-21
Oath's Prebid Server adapter


smithaammassamveettil:
Description:
Introducing new adaptor for Oath bidder for Prebidding through Prebid Server(Appnexus prebid server), following openRTB2.5 specification for bid requests. This adaptor will reach out to Oath's DSP platform to fetch bids during bidding process.

BidderName: oath

dbemiller:
Fixed by #520

PR #520, opened 2018-05-21
 [Issue-518] Oath's Prebid Server adapter

smithaammassamveettil:
Ticket/Issue :

[Issue-518]Oath's Prebid Server adapter

Description:

Introducing new adaptor for Oath bidder for Prebidding for Prebid Server(Appnexus prebid server), following openRTB2.5 specification for bid requests. This adaptor will reach out to Oath's DSP platform to fetch bids during bidding process.

Contact Info:
Please contact smithaa@oath.com for any issues/questions/changes on this PR


cirla:
Closes #518

adapters/oath/oath.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if response.StatusCode == http.StatusBadRequest {
+		return nil, []error{&adapters.BadInputError{
+			Message: fmt.Sprintf("Unexpected status code: %d. ", response.StatusCode),
+		}}
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return nil, []error{fmt.Errorf("Unexpected status code: %d. ", response.StatusCode)}
```
dbemiller: Should this be an `adapters.BadServerResponseError`?

adapters/oath/oath.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	var bidResp openrtb.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
```
dbemiller: Same here... this looks like an `adapters.BadServerResponseError`

adapters/oath/oath.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// getMediaTypeForImp figures out which media type this bid is for.
+func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	mediaType := openrtb_ext.BidTypeBanner //default type
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			} else if imp.Native != nil {
```
dbemiller: The rest of your code, and your `info.yaml` file, looks like you only support `Banner` or `Video`... why does this check `Native`?

adapters/oath/oath.go
```diff
@@ -0,0 +1,178 @@
...
+			Message: fmt.Sprintf("No impression in the bid request"),
+		}
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	var bannerImps []openrtb.Imp
+	var videoImps []openrtb.Imp
+
+	for _, imp := range request.Imp {
```
hhhjort: It seems the purpose of this loop is to simply validate that at least one valid Imp exists. May be more efficient to simply declare `validImps := false`, and then when you find a banner or video imp, set `validImps = true` rather than building these two lists that you will never actually use. Avoiding the allocation extra memory as the lists grows may save a milisecond that can be spent waiting for the response before the timeout hits. :)

usersync/usersync.go
```diff
@@ -36,3 +36,20 @@ type CookieSyncBidders struct {
 	NoCookie     bool          `json:"no_cookie,omitempty"`
 	UsersyncInfo *UsersyncInfo `json:"usersync,omitempty"`
 }
+
+type syncer struct {
```
dbemiller: This block of code isn't used anywhere... did you add it accidentally?

adapters/oath/oath.go
```diff
@@ -0,0 +1,178 @@
...
+	if response.StatusCode != http.StatusOK {
+		return nil, []error{fmt.Errorf("Unexpected status code: %d. ", response.StatusCode)}
+	}
+
+	var bidResp openrtb.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(5)
```
hhhjort: Will you ever find more than one SeatBid in your response from Oath? if not, then
`adapters.NewBidderResponseWithBidsCapacity(len(bidResp.SeatBid.Bid))`
would be an accurate way to set the capacity.

usersync/usersync.go
hhhjort: This conflicts with the definition of `syncer` in `usersync/usersyncers/syncer.go` as well as being broken since this definition of the struct does an infinite recursion for `syncer.GDPRVendorID()`.

smithaammassamveettil:
Sure, updated accordingly

smithaammassamveettil:
First of all Thanks for reviewing the PR. Looks like I am not able to edit
the PR directly under each comments listed in the PR. So replying to the
emails here as another option to edit the PR comment.

Regarding your below comment, I have a few clarifications required on
configuration settings. As you mentioned below we wont be using single URL
for all prebid servers. Would like to use different URLs based on colo. Can
you please direct me how to set those configurations? I couldn't find any
documentation or sample setting on how exactly need to specify the
configuration, if we want to use *OathAdapter.URI. *


*"Bit confused... you have a hardcoded URL here with appnexus in the path,
but then an unused URI param on the OathAdapter struct which is passed in
from the app config, but never used.*

*Do you intend to hardcode a single URL for all Prebid Server host
companies, or use different ones for different hosts?*

*If you're hardcoding a single one, delete the NewOathBidder endpoint param
and OathAdapter.URI field, since they're not used.*

*If you want different hosts to hit different endpoints, delete this and
change your MakeRequests impl to use the OathAdapter.URI param rather than
this hardcoded one."*

dbemiller:
yeah... the documentation on configs is very sparse. All we have is [this](https://github.com/prebid/prebid-server/blob/master/docs/developers/configuration.md).

For this PR, just add a `viper.setDefault` for your endpoint alongside [all these other ones](https://github.com/prebid/prebid-server/blob/master/pbs_light.go#L641). If you want PBS hosts to use different ones based on region, put a good "default" there in the code and then leave the others in comments.

The actual config options are all set by the PBS host during deployment. There are a bunch of ways to do that... but it's outside the scope of a PR. Check out the docs for Viper docs, and feel free to open another issue if you still have questions after that.

smithaammassamveettil:
Yea, typo... Just fixed.

smithaammassamveettil:
Looks like this commit overridden your prev approval :-) Please review and
approve...
Also what is the next step after this PR approval?

kizzard:
Before merging.. might I suggest reviewing the adapter name to avoid confusion with other Oath supply-side products, Perhaps "brightroll" or "dspplus" would be more appropriate?

dbemiller:
looks good to me. We'll wait to merge until @smithaammassamveettil confirms the adapter name. I can re-approve if you want to rename it.

We wait on PRs until they either have 2 approvals, or 1 approval and 5 business days pass. Then we'll merge it, and eventually tag a new release. It will go "live" with different PBS hosts whenever they decide to deploy it. The best way to tell is to ping `/info/bidders` (e.g. http://prebid.adnxs.com/pbs/v1/info/bidders for Appnexus' instance).


One interesting issue here is that I don't see your Bidder in Prebid.js yet... which makes you the first Prebid Server Bidder who's _not_ in Prebid.js. Since almost all the web traffic on PBS comes through Prebid.js, I would be surprised if you got any non-mobile traffic.

I'll open an issue to discuss with the Prebid.js core team... ideally they should be able to hit PBS demand without making everyone add their Bidder to both projects... but we'll see what they say.

smithaammassamveettil:
I have renamed the adapter to "brightroll" and all the related changes are
in now. Please re-review and approve. Thanks for your time!

Also can you please add me to the other Issue#2659 you opened to track the
Prebid.js change. I dont have access to edit any PR directly. And not
receiving any mails in my emailid smithaa@oath.com when someone updates
that PR, so not able to track or update the PR through email also. So
please add me there or give me appropriate access to edit any PRs in the
repo.

dbemiller:
@smithaammassamveettil you can subscribe to receive emails for any issue or PR you care about: https://help.github.com/articles/subscribing-to-and-unsubscribing-from-notifications/

If you need to update Prebid.js it's the same process as Prebid Server. Fork, update, and open a PR for it. That said... I think someone from the prebid.js core team is looking into it for you. Subscribe to that issue if you want updates.

adapters/brightroll/brightroll.go
```diff
@@ -0,0 +1,170 @@
...
+	headers.Add("Accept", "application/json")
+	headers.Add("x-openrtb-version", "2.5")
+
+	if request.Device != nil {
+		addHeaderIfNonEmpty(headers, "User-Agent", request.Device.UA)
+		addHeaderIfNonEmpty(headers, "X-Forwarded-For", request.Device.IP)
+		addHeaderIfNonEmpty(headers, "Accept-Language", request.Device.Language)
+		addHeaderIfNonEmpty(headers, "DNT", strconv.Itoa(int(request.Device.DNT)))
+	}
+	fmt.Println("**** URI ::: " + thisURI)
```
dbemiller: Looks like this slipped into your most recent commits? It shouldn't be here though... it'll pollute the app logs.

smithaammassamveettil:
Thank you for approving!

It was a good catch, added as part of local testing after renaming adapter.
Somehow missed to remove it. Thanks!

dbemiller:
@smithaammassamveettil I'm not sure which PR you want to edit. https://github.com/prebid/Prebid.js/issues/2659 is an _issue_, made for discussion & comments. There are no PRs related to this yet.

> its a full fledged Client side adapter, which actually
doing the complete end to end bidding process without interacting with
server side adapter

Almost correct :). It's actually a full fledged Client side _auction runner_. Bidders can submit adapters there as well.

> As you know, what we are anticipating from server side adapter...

This is all correct.

> I hope, in this server side bidding process, Publisher just need to identify the participating bidders to call and the corresponding params to pass to corresponding adapter right? Do you think we need the full client side adapter for that? Is there any specific setting or Prebid.js implementation need to do from our side for the same? Or is it something, you can do in your configuration looking at our adapter implictly for the publishers as per demand? Or is there a lightweight Prebid.js we can use for server side adapter? How does it work?

You're good to go :). We don't need anything else from you.

I opened https://github.com/prebid/Prebid.js/issues/2659 because there are some design issues with Prebid.js which will prevent it from playing nicely with this PBS adapter. The Prebid.js core team knows about them, and is working on some fixes now.

There's no need for you to build a client-side adapter in Prebid.js unless you want to :). You may get more publishers using your endpoint if you do, but we plan to make the Prebid ecosystem work without it.

PR #524, opened 2018-05-23
Made auction max timeout configurable.

dbemiller:
This adds a configurable max timeout for auction requests. "0" means "no limit".

This lets PBS hosts effectively cap the max auction timeout they're willing to allow, which is important for some technical reasons.

I also found that the `DefaultTimeout` config option was unused throughout the code, so I deleted it.


dbemiller:
That's not a bad idea, at such high scale... I'll add it in a followup.

Issue #526, opened 2018-05-23
Travis Build failure on go version go1.9


smithaammassamveettil:
While building our new oath adaptor codebase, one of the build jobs(one building on go version 1.9) is failing: Attaching the build job link below. 
https://travis-ci.org/prebid/prebid-server/builds/382830737
The failing build job(Job #2211.1) doesn't list  any exact reason for failure also.

You can see the build is done successfully on go1.10.1 version as per the Travis build log on the same link. 
I tried the same build command  on my local for both go versions (go version go1.9.6 darwin/amd64 and go version go1.10.1 darwin/amd64 ) and is able to successfully build on both.

dbemiller:
very interesting...

Are you sure you're running `./validate.sh --nofmt --cov`? This fails for me consistently on go1.9.3.

I don't get debug output either, which is very frustrating. However, running `./validate.sh` (with _no_ `--nofmt` flag) does make a formatting fix to `usersync/usersync.go`

dbemiller:
#527 fixes the issue with bad debugging output.

smithaammassamveettil:
Yes, I tried those 2 commands
  ./validate.sh (without --nofmt flag)

also
  ./validate.sh --nofmt  --cov

Both are successfull on my local

dbemiller:
Bizarre... it looks like you've got it fixed now though.

Fair to close this?

smithaammassamveettil:
Yes,
looking at your comment above "*However,
running ./validate.sh (with no --nofmt flag) does make a formatting fix to *
*usersync/usersync.go"*, I did some cleanups in this file, which helped to
fix the issue.

Thanks for pointing out this.

Also it will be great to have the reason listing clearly in the build
failure log. Hope your other PR will do that. Thanks!

PR #527, opened 2018-05-23
Fixed ./validate.sh --nofmt output bug.

dbemiller:
#526 noted that `./validate.sh --nofmt` produced blank output. I got this while running it on their branch:

```
dbemiller-mac:prebid-server dbemiller$ ./validate.sh --nofmt
```

This PR fixes that. Looks like it was still printing a hardcoded list of files, so formatting issues in these new directories wouldn't get printed properly.

### DIFF
--- a/validate.sh
+++ b/validate.sh
@@ -38,7 +38,7 @@ if $AUTOFMT; then
     done
   fi
 else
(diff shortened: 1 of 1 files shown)


dbemiller:
the diff is the other way around... we _were_ only running it on pbs_light.go and two modules

PR #529, opened 2018-05-24
Silenced some false positive errors

dbemiller:
This is an absolutely ridiculous solution to an absolutely ridiculous problem.

TCP connections can be closed by callers. If that happens, then the server gets an error when _it_ tries to close the connection.

This fragile and hacky workaround is the best I found in [five and a half years of discussion](https://github.com/golang/go/issues/4373). It even appears [in the standard libs](https://github.com/golang/net/blob/master/http2/server.go#L626).

I just want the app logs to be filled with messages that mean things  ¯\\_(ツ)_/¯

### DIFF
--- a/server/listener.go
+++ b/server/listener.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"net"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


server/listener.go
```diff
@@ -25,7 +26,13 @@ func (l *monitorableConnection) Close() error {
 	if err == nil {
 		l.metrics.RecordConnectionClose(true)
 	} else {
-		glog.Errorf("Error closing connection: %v", err)
+		// If the connection was closed by the client, it's not a real/actionable error.
+		// Although there are no official APIs to detect this, this ridiculous workaround appears
+		// in the core Go libs: https://github.com/golang/go/issues/4373#issuecomment-347680321
```
cirla: 🤣

PR #531, opened 2018-05-24
Deleting the /getuids endpoint

dbemiller:
`/getuids` is an undocumented endpoint which I don't think anyone ever uses.

I'm not sure why it exists, as it pre-dates my involvement in the project. My best guess is that someone used it to help debugging at some point.

Gonna delete it unless someone complains in the next few weeks.


khatibda:
i was debating whether to use it to track match rates and the impact on bidder eCPM.  perhaps there is another way to get this data.

dbemiller:
hmm... we could make auctions return it in the `bidresponse.ext` somewhere, so that client-side scripts (like prebid.js analytics modules) could access it.

That should be more reliable too, since prebid.js does cookie syncs immediately after the auction.

khatibda:
sounds good!

dbemiller:
Logged as #567

PR #534, opened 2018-05-25
Resolve AMP size query params

### LINKED ISSUE #420: Additional parameters for AMP RTC

dbemiller:
This includes the changes from #470, fixed a few inconsistencies with https://github.com/prebid/prebid-server/issues/420#issuecomment-383087697, and closes #420.

### DIFF
--- a/docs/endpoints/openrtb2/amp.md
+++ b/docs/endpoints/openrtb2/amp.md
@@ -8,11 +8,11 @@ For a User's Guide, see the [AMP feature docs](http://prebid.org/dev-docs/show-p
 The `tag_id` ID must reference a [Stored BidRequest](../../developers/stored-requests.md#stored-bidrequests).
 For a thorough description of BidRequest JSON, see the [/openrtb2/auction](./auction.md) docs.
 
-Optionally, a param `debug=1` may be set, setting `"test": 1` on the request and resulting in [additional debug output](auction.md#debugging).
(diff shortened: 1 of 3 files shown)


dbemiller:
`ms` -> `data-multi-size` is actually already defined [in our docs](https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/amp.md#query-parameters).

Ironically it looks like it never made it into AMP, though, so... I'm adding it here: https://github.com/ampproject/amphtml/pull/15599

PR #536, opened 2018-05-29
Make Pulsepoint params accept values like "200X150"

### LINKED ISSUE #535: Pulsepoint JSON schema is incorrect
Found this while debugging https://github.com/prebid/Prebid.js/issues/2623

```
Invalid request format: request.imp[0].ext.pulsepoint failed validation.
cf: Does not match pattern '^[0-9]+x[0-9]+$'
```

The code which uses this size is: `var size = strings.Split(strings.ToLower(params.AdSize), "x")`

Although that technically works... in both `Prebid.js` and the `pulsepoint_test.go` file, they use capital `X` chars: `bidder.AdUnits[0].Params = json.RawMessage("{\"cp\": 2001, \"cf\": \"728X90\"}")`

This is causing at least some of the problems found in https://github.com/prebid/Prebid.js/issues/2623

dbemiller:
In https://github.com/prebid/Prebid.js/issues/2623, one of the bidder params uses a capital `X` to separate the width & height of `cf` param.

This looks like it's what the `pulsepoint.md` example in their Prebid.js file expects... so I figured we'd just change it here.

We need to accept both forms, of course, since there may be `x` sizes in the DB. This fixes #535

### DIFF
--- a/static/bidder-params/pulsepoint.json
+++ b/static/bidder-params/pulsepoint.json
@@ -14,7 +14,7 @@
     },
     "cf": {
       "type": "string",
-      "pattern": "(^[0-9]+X[0-9]+$)|(^[0-9]+x[0-9]+$)",
+      "pattern": "^[0-9]+[xX][0-9]+$",
       "description": "The size of the ad slot being sold. This should be a string like 300X250"
     }
   },
--- a/static/bidder-params/pulsepoint.json
+++ b/static/bidder-params/pulsepoint.json
@@ -14,8 +14,8 @@
     },
     "cf": {
       "type": "string",
-      "pattern": "^[0-9]+x[0-9]+$",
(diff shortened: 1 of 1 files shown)


static/bidder-params/pulsepoint.json
```diff
@@ -14,8 +14,8 @@
     },
     "cf": {
       "type": "string",
-      "pattern": "^[0-9]+x[0-9]+$",
-      "description": "The size of the ad slot being sold. This should be a string like 300x250"
+      "pattern": "(^[0-9]+X[0-9]+$)|(^[0-9]+x[0-9]+$)",
```
hhhjort: would `"pattern": "(^[0-9]+[xX][0-9]+$)"` be simpler?

dbemiller: definitely, yeah. Thanks

Issue #375, opened 2018-03-02
Bidder for beachfront.com


muncha:
I'm building a A bidder for beachfront.com

This is for the openrtb2/auction endpoint only. The legacy methods will return, but don't do anything. I'm currently testing against prebid.js #2102 for banner and video. I still have all the test cases to write.

dbemiller:
cool! Thanks for letting us know.

Just FYI: if you don't want to support the legacy methods, then you don't need to implement them on your adapter. You can just register the MisconfiguredAdapter, found [here](https://github.com/prebid/prebid-server/blob/master/adapters/legacy.go#L79) directly into the legacy exchange, found [here](https://github.com/prebid/prebid-server/blob/master/pbs_light.go#L774)

dbemiller:
@muncha any update on this?

muncha:
Sorry - I missed that update request! 
I'm actually getting stuff together right now to put in a pull request.

muncha:
I'm at 85.0% coverage right now. I'll have that to > 90%  shortly and I should be able to put in the request tomorrow.

muncha:
I've got my pull request in - #538.

(fingers crossed)

dbemiller:
This was merged and released. Closing.

PR #538, opened 2018-05-30
Beachfront bidder

muncha:
This is the bidder for beachfront media, #375. I'm only implementing the openrtb/auction endpoint and ignoring the legacy stuff.


muncha:
The test that failed - "./validate.sh --nofmt --cov --race 10" worked when I ran it locally.

dbemiller:
interesting... perhaps a similar failure to #526?

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,541 @@
...
+	return true
+}
+
+func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
+	errs := make([]error, 0, len(request.Imp))
+
+	var beachfrontRequests BeachfrontRequests
+	var reqJSON []byte
+
+	a.URI = func() string {
```
hhhjort: You have a race condition here. If you have two requests, one for banner and and other for video, come in at the same time, then the two will be overriding the adapter object's URI property. You should make this a local variable so that you don't have this cross talk between `MakeRequests()` calls.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+
+	return beachfrontReqs, err
+}
+
+func getBannerRequest(req *openrtb.BidRequest) (BeachfrontBannerRequest, error) {
+	var beachfrontReq BeachfrontBannerRequest
+	dec := json.NewDecoder(strings.NewReader(beachfrontBannerRequestTemplate))
+	for {
+		if err := dec.Decode(&beachfrontReq); err == io.EOF {
```
hhhjort: It all that `beachfrontBannerRequestTemplate` is used for is populating default values into `beachfrontReq`, wouldn't it be better to simply either define a default instance of the struct, and then deep copy it into `beachfrontReq`, or define a `defaultBannerReq()` function to allocate the struct and fill in the non null default entries? Saves having to process a string through the JSON decoder for every request.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+	}
+
+	// step through the prebid request "imp" and inject into the beachfront request.
+	for i, imp := range req.Imp {
+		if imp.Video != nil {
+			return beachfrontReq, nil
+		} else if imp.Banner != nil {
+			if i > 0 {
+				beachfrontReq.Slots = append(beachfrontReq.Slots, BeachfrontSlot{})
```
hhhjort: what happens here if there is a native `Imp` as part of the request? You expand the size of `beachfrontReq.Slots`, by 1 in order to have have the slot index available, but then reference `i` directly for the slot. If there was a native (or something else not video or banner) inbetween, `i` will have advanced by 2 since the last append, and you will get an array out of bounds error.

There are two solutions, start with setting `Slots` to be an array of size `len(req.Imp)`, or reference the slot as `beachfrontReq.Slots[len(beachfrontReq.Slots)-1]`

Also from the way you have defined your template, the first slot will have an extra 0x0 size in the array, since it won't be an empty `beachfrontSlot{}` like you are appending for all slots beyond the first.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,541 @@
...
+/*
+Prepare the request that has been received from Prebid.js, translating it to the beachfront format
+*/
+func getVideoRequest(req *openrtb.BidRequest) (BeachfrontVideoRequest, error) {
+	var beachfrontReq BeachfrontVideoRequest
+	var videoImps int8 = 0
+	dec := json.NewDecoder(strings.NewReader(beachfrontVideoRequestTemplate))
+
+	for {
+		if err := dec.Decode(&beachfrontReq); err == io.EOF {
```
hhhjort: See my comment above about decoding a default banner template in `getBannerRequest()`. It all should apply to this video request template.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,541 @@
...
+	impressions, only a single URL will be returned. Hopefully the beachfront endpoint can be modified
+	in the future to return multiple video ads
+
+	*/
+	for i, imp := range req.Imp {
+		if imp.Video != nil {
+			beachfrontReq.Id = req.ID
+
+			// The template has 1 Imp struct, so use that one first, then add them as needed.
+			if videoImps > 0 {
```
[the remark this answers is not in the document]
muncha: @hhhjort Thanks for all of that.

@dbemiller I ran validate.sh as above with version 1.9.3, and it also worked fine locally. I'll continue working on that bug after the changes hhhjort points out.

muncha:
@hhhjort Thanks for all of that.

@dbemiller I ran validate.sh as above with version 1.9.3, and it also worked fine locally. I'll continue working on that bug after the changes hhhjort points out.

muncha:
@dbemiller - I was just setting up some test pages for the beachfront people to look at and therefore did a pull and build on the staging server. I went ahead and ran that validation command there, and I got the error. See commit 15e261d message for details.

This looks to me like a bug in Go, as far Mac vs. Linux builds not applying consistent format rules.

.gitignore
```diff
@@ -42,3 +42,5 @@ inventory_url.yaml
 # static/version.txt
 
 .idea/
+
+analytics/
```
dbemiller: Looks like this got committed with your `gofmt` fix... I'm assuming it wasn't intentional :)

muncha: Yep. I'll get that out for next commit.

muncha:
@dbemiller That last Travis CI fail was a result of one of my test cases failing because I forgot to add the len() based indexing to the video method, not part of the Mac/Linux weirdness that I'm seeing. But like I said in the message for ff0eba0 - I'll hold off on the commits because this looks weird.

muncha:
@hhhjort - I think that this covers everything that you mentioned above. 

@dbemiller  - As for the weird bit where I had to reformat a comment in pbs_light.go that was not mine in order to pass the validation, that still seems weird to me, but it was just formatting, and it was just a comment so I'm not going to get hung up on it.

adapters/beachfront/params_test.go
```diff
@@ -0,0 +1,15 @@
+package beachfront
+
+import (
+	"testing"
+)
+
+func TestValidParams(t *testing.T) {
```
dbemiller: Looks like some tests are missing here?

docs/bidders/beachfront.md
```diff
@@ -0,0 +1,2 @@
+# Beachfront bidder
```
dbemiller: Do you have docs to add here? If not, just delete it.

muncha: I've added docs and parameter testing.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,486 @@
...
+	} else {
+		/* Beachfront currently returns banner ads in a sparse format which is just the openRTB seatbid
+		object. It needs to be wrapped up in openrtb format.
+		*/
+		if err = json.Unmarshal(response.Body, &beachfrontResp); err != nil {
+			return openrtbResp, err
+		}
+
+		openrtbResp.ID = id
+		for range beachfrontResp {
```
dbemiller: It seems like this can all be simplified quite a bit... though correct me if I'm wrong.

`MakeBids` needs to return a `BidderResponse`, which is basically a `(currency, list of bids)` pair. Your banner endpoint returns a list of bids. However, you're wrapping each `Bid` into its own `SeatBid` (here & in `postprocessBanner`), and then _unwrapping_ it in `MakeBids`.

Wouldn't it be simpler, and more efficient, to just return the list of bids?

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,486 @@
...
+			openrtbResp.SeatBid = append(openrtbResp.SeatBid, openrtb.SeatBid{})
+		}
+
+		return postprocessBanner(openrtbResp, beachfrontResp, id)
+	}
+}
+
+func postprocessBanner(openrtbResp openrtb.BidResponse, beachfrontResp []BeachfrontResponseSlot, id string) (openrtb.BidResponse, error) {
+	r, _ := regexp.Compile("\\\"([0-9]+)")
```
dbemiller: regexes are cheap to use, but expensive to compile. You should use `regexp.MustCompile` and declare this loose in your file, treating it like a pseudo-const.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,486 @@
...
+		return openrtbResp, err
+	}
+
+	/* there will only be one seatBid because beachfront only returns a single video ad
+	but if that were to change this should work on all of them:
+	*/
+	for i := range openrtbResp.SeatBid {
+		for j := range openrtbResp.SeatBid[i].Bid {
+			openrtbResp.SeatBid[i].Bid[j].ImpID = xtrnal.Imp[i].ImpId
+			openrtbResp.SeatBid[i].Bid[j].CrID = xtrnal.Imp[i].ImpId
```
dbemiller: Whoa, hang on... `CrID` supposed to identify the _creative_ that was returned. `ImpId` is given by pubs to identify the impression.

It's not good for anything if you just overwrite it with the impression ID here.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+func getBannerRequest(req *openrtb.BidRequest) (BeachfrontBannerRequest, error) {
+	var bannerImpsIndex int = 0
+	var beachfrontReq BeachfrontBannerRequest = NewBeachfrontBannerRequest()
+	/*
+	 step through the prebid request "imp" and inject into the beachfront request. If we got to here,
+	 then we have already stepped through the requested imps and verified that none are Video, so no
+	 reason to check that here, but there could be Audio or Native (or maybe they are filtered out before
+	 I get here based on the capabilities in bidder-info/beachfront.yaml? Regardless, I'll leave
```
dbemiller: Unfortunately, not yet :(. That is indeed a feature we've talked about, though, and it's a great idea.

I realized it wasn't logged anywhere, though, so I opened #555 for it.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
+	errs := make([]error, 0, len(request.Imp))
+
+	var beachfrontRequests BeachfrontRequests
+	var reqJSON []byte
+	var uri string
+
+	uri = func() string {
```
dbemiller: little strange... why the inline function & immediate execution here?

Seems like you should either move this into a named helper function, or just extract the logic so that `MakeRequests` runs it directly.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,486 @@
...
+	 step through the prebid request "imp" and inject into the beachfront request. If we got to here,
+	 then we have already stepped through the requested imps and verified that none are Video, so no
+	 reason to check that here, but there could be Audio or Native (or maybe they are filtered out before
+	 I get here based on the capabilities in bidder-info/beachfront.yaml? Regardless, I'll leave
+	 place holders here) .
+	*/
+
+	for _, imp := range req.Imp {
+		if imp.Audio != nil {
+			// Place holder
```
dbemiller: per [the docs](https://github.com/prebid/prebid-server/blob/master/adapters/bidder.go#L17):

> The errors should contain a list of errors which explain why this bidder's bids will be
> "subpar" in some way. For example: the request contained ad types which this bidder doesn't support.

If `MakeRequests` or `MakeBids` return errors _and_ useful HTTP reqs/Bids, the core logic will make use of both. Bids will still happen, and the errors will be returned as strings to help with debugging.

.gitignore
```diff
@@ -39,6 +39,6 @@ pbs.*
 inventory_url.yaml
 
 # autogenerated version file
-# static/version.txt
```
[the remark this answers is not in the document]
muncha: I checked out your current master .gitignore, locally for now. I think I deleted that line unintentionally when removing some stuff I had added for testing.

adapters/beachfront/beachfront.go
muncha: Yes it would. Thanks.

muncha:
@dbemiller - I think those requests are covered.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,510 @@
...
+type BeachfrontResponseSlot struct {
+	CrID  string  `json:"crid"`
+	Price float64 `json:"price"`
+	W     uint64  `json:"w"`
+	H     uint64  `json:"h"`
+	Slot  string  `json:"slot"`
+	Adm   string  `json:"adm"`
+}
+
+func (a *BeachfrontAdapter) Name() string {
```
dbemiller: Name() is dead code... you can delete it.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,510 @@
...
+	return "beachfront"
+}
+
+func (a *BeachfrontAdapter) SkipNoCookies() bool {
```
dbemiller: `SkipNoCookies()` is dead code... you can delete it.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,510 @@
...
+			isVideo = true
+			bidtype = openrtb_ext.BidTypeVideo
+			break
+		}
+	}
+
+	bids, err = postprocess(response, externalRequest, internalRequest.ID, isVideo)
+
+	if err != nil {
+		return nil, []error{fmt.Errorf("Failed to process the beachfront response\n%s", err)}
```
dbemiller: This should be an `adapters.BadServerResponseError`, right?

From the docs:

> // BadServerResponseError should be used when returning errors which are caused by bad/unexpected behavior on the remote server.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,510 @@
...
+func postprocessVideo(bids []openrtb.Bid, externalRequest *adapters.RequestData, id string) ([]openrtb.Bid, error) {
+	var xtrnal BeachfrontVideoRequest
+	var err error
+	var crid string
+
+	if err = json.Unmarshal(externalRequest.Body, &xtrnal); err != nil {
+		return bids, err
+	}
+
+	glog.Info(bids)
```
dbemiller: Please remove the `glog`s. They make the app logs unusable.

muncha:
Thank you!

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,495 @@
...
+
+	for _, imp := range req.Imp {
+		if imp.Audio != nil {
+			return beachfrontReq, &adapters.BadInputError{
+				Message: fmt.Sprintf("Beachfront doesn't support audio Imps. Ignoring Imp ID=%s", imp.ID),
+			}
+		} else if imp.Native != nil {
+			return beachfrontReq, &adapters.BadInputError{
+				Message: fmt.Sprintf("Beachfront doesn't support native Imps. Ignoring Imp ID=%s", imp.ID),
+			}
```
hhhjort: Until #555 is fixed, these returns will mean you will not bid on any banner imps on a multi-imp request even if only one of the imps is an unsupported format. Not fatal, but I am not sure that is what you want.

muncha: That is not what I want. Please hold off until my next push.

muncha:
@hhhjort - Improved error handling. Thank you for pointing that out. Ready to go if it looks ok to you.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,512 @@
...
+		*/
+		reqJSON, err = json.Marshal(beachfrontRequests.Banner)
+	}
+
+	if imps == 0 {
+		errs = append(errs, errors.New("No valid impressions were found"))
+		return nil, errs
+	}
+
+	if err != nil {
```
dbemiller: I don't see you using `err` anymore. So it should always be `nil`, and this never executes.

Can remove it.

adapters/beachfront/beachfront.go
```diff
@@ -0,0 +1,512 @@
...
+	}}, errs
+}
+
+/*
+We have received a prebid request. It needs to be converted to a beachfront request. This is complicated
+by the fact that we have different servers for video/display and they have different contracts.
+*/
+func preprocess(req *openrtb.BidRequest, uri string) (BeachfrontRequests, []error, int) {
+	var beachfrontReqs BeachfrontRequests
+	var errs = make([]error, 0, len(req.Imp))
```
dbemiller: You're not actually appending to this slice anywhere... just re-assigning it. So you could just drop the `make` call and set it to `nil` for more efficiency.

dbemiller:
just FYI: we're waiting to merge this because `master` right now has been acting a bit shaky in prod.

We're also on vacation next week... but should get this in soon after.

muncha:
Got it. Thank you and enjoy your vacation.

PR #539, opened 2018-05-31
Defines defaults for all config options and removes pointers from the…

### LINKED ISSUE #475: App config values aren't used as expected
To reproduce:

1. Define a `pbs.yaml` file which _doesn't_ include an entry for `status_response: "stuff"`
2. Build PBS and set the environment variable with `export PBS_STATUS_RESPONSE=something`
3. Run `GET localhost:8000/status`

Expected:

- The `/status` endpoint should return "something".

Observed:

- A 204 is returned because the config value was empty.

------------------------------------------------------------

Even more unusual: define a `pbs.yaml` which _does_ include an entry for `status_response: "something-else"`

Restart the PBS app. Now the `PBS_STATUS_RESPONSE` environment variable will be in effect.

hhhjort:
… Configuration struct

This will allow Viper to read environment variables into the config for all config options. Fixes #475 

Copying this to the top so people can easily find the breaking change:
Before this, A config like:

stored_requests:
  postgres:
    in_memory_cache:
      ttl_seconds: ###
      request_cache_size_bytes: ###
      imp_cache_size_bytes: ###
would create an in-memory stored request cache.

After this, it needs a Type should be:

stored_requests:
  postgres:
    in_memory_cache:
      type: "unbounded"   // could also be "none" or "lru"
      ttl_seconds: ###
      request_cache_size_bytes: ###
      imp_cache_size_bytes: ###
The ttl_seconds, request_cache_size_bytes, and imp_cache_size_bytes will only apply to lru caches.

That makes this a breaking change for PBS hosts.


stored_requests/config/config.go
```diff
@@ -84,12 +84,12 @@ func newFetchers(cfg *config.StoredRequests, client *http.Client, db *sql.DB) (f
 		idList = append(idList, fFetcher)
 		ampIDList = append(ampIDList, fFetcher)
 	}
-	if cfg.Postgres != nil {
+	if cfg.Postgres.FetcherQueries.QueryTemplate != "" {
```
dbemiller: Do you think we should split this into two clauses now? Someone might define `QueryTemplate` and not `AmpQueryTemplate`, or vice versa.

Otherwise we should add some `validate()` logic to make sure that both or neither are defined... but I lean towards changing the logic here so that it works either way.

stored_requests/config/config.go
```diff
@@ -84,12 +84,12 @@ func newFetchers(cfg *config.StoredRequests, client *http.Client, db *sql.DB) (f
...
 		glog.Infof("Loading Stored Requests via Postgres.\nQuery: %s\nAMP Query: %s", cfg.Postgres.FetcherQueries.QueryTemplate, cfg.Postgres.FetcherQueries.AmpQueryTemplate)
 		idList = append(idList, db_fetcher.NewFetcher(db, cfg.Postgres.FetcherQueries.MakeQuery))
 		ampIDList = append(ampIDList, db_fetcher.NewFetcher(db, cfg.Postgres.FetcherQueries.MakeAmpQuery))
 	}
-	if cfg.HTTP != nil {
+	if cfg.HTTP.Endpoint != "" {
```
dbemiller: Same issue goes here, with `Endpoint` and `AmpEndpoint`

stored_requests/config/config.go
```diff
@@ -101,45 +101,44 @@ func newFetchers(cfg *config.StoredRequests, client *http.Client, db *sql.DB) (f
 }
 
 func newCache(cfg *config.StoredRequests) stored_requests.Cache {
-	if cfg.InMemoryCache == nil {
+	if cfg.InMemoryCache.RequestCacheSize == 0 {
```
dbemiller: eh... this will actually cause problems. See the docs on `InMemoryCache`:

```
// RequestCacheSize is the max number of bytes allowed in the cache for Stored Requests. Values <= 0 will have no limit
```

stored_requests/config/config.go
```diff
@@ -101,45 +101,44 @@ func newFetchers(cfg *config.StoredRequests, client *http.Client, db *sql.DB) (f
...
 		eventProducers = append(eventProducers, newEventsAPI(router, "/storedrequests/openrtb2"))
 		ampEventProducers = append(ampEventProducers, newEventsAPI(router, "/storedrequests/amp"))
 	}
-	if cfg.HTTPEvents != nil {
+	if cfg.HTTPEvents.RefreshRate != 0 {
 		eventProducers = append(eventProducers, newHttpEvents(client, cfg.HTTPEvents.TimeoutDuration(), cfg.HTTPEvents.RefreshRateDuration(), cfg.HTTPEvents.Endpoint))
 		ampEventProducers = append(ampEventProducers, newHttpEvents(client, cfg.HTTPEvents.TimeoutDuration(), cfg.HTTPEvents.RefreshRateDuration(), cfg.HTTPEvents.AmpEndpoint))
 	}
-	if cfg.Postgres != nil {
+	if cfg.Postgres.CacheInitialization.Query != "" {
```
dbemiller: Same issue with `Query` and `AmpQuery` here

dbemiller:
Fixes #475

pbs_light.go
```diff
@@ -618,16 +618,58 @@ func init() {
 	viper.AddConfigPath(".")
 	viper.AddConfigPath("/etc/config")
 
+	// Some defaults will be set just so they are accessable via environment variables
+	// (basically so viper knows they exist)
```
dbemiller: Could you start this comment with `// Fixes #475: {description}`?

It's searchable by grep/find/whatever, and points confused engineers to the discussion which caused the code.

stored_requests/config/config.go
hhhjort: I think our logic breaks in other places if someone defines one and not the other. Before if the struct was not null, we just assumed we had both and configured the fetchers appropriately. We will likely have to dive deeper into our code if we want to support piecemeal configs.

stored_requests/config/config.go
hhhjort: I will have to add a `use_memory_cache` property, as it is a valid config to have all other cache properties be their zero values (just not all zero at the same time as that would be unlimited cache size with no expirery)

stored_requests/config/config.go
dbemiller: if it doesn't work today, separate PR.

stored_requests/config/config.go
dbemiller: yeah, that sounds right.

Alternatively you could add a `cfg.InMemoryCache.Type` with values `none`, `unbounded` and `lru`, and then `cfg.InMemoryCache.LRU` with the LRU-specific configs.

I just mention it because that pattern was used consistently in `prebid-cache`, and it worked out pretty well... but either one would work for now.

stored_requests/config/config.go
hhhjort: Fixing the base issues was enough work for this PR. Adding support for partial configs (something we don't support today) left for another PR.

config/config.go
```diff
@@ -198,3 +198,96 @@ func (cfg *Cache) GetBaseURL() string {
 func (cfg *Configuration) GetCachedAssetURL(uuid string) string {
 	return fmt.Sprintf("%s/cache?%s", cfg.CacheURL.GetBaseURL(), strings.Replace(cfg.CacheURL.Query, "%PBS_CACHE_UUID%", uuid, 1))
 }
+
+// Set the default config values for the viper object we are using.
+func ConfigViper(v *viper.Viper) {
```
dbemiller: opinion: `SetDefaults` might be a bit more clear?

config/config_test.go
```diff
@@ -9,7 +9,12 @@ import (
 )
 
 func TestDefaults(t *testing.T) {
-	cfg, err := New(newViperWithDefaults())
+	v := viper.New()
+	ConfigViper(v)
+	v.SetDefault("auction_timeouts_ms.max", 250)
```
dbemiller: This seems a bit weird...

This test targets the `ConfigViper` method, but this sets extra defaults and then asserts on them. Seems like these should go away, and the assertions replaced with the intended default values.

config/stored_requests.go
```diff
@@ -316,6 +314,8 @@ func makeIdList(numSoFar int, numArgs int) string {
 }
 
 type InMemoryCache struct {
+	// Identify the type of memory cache. "none", "unbounded", "lru"
+	Type string `mapstructure:"type"`
```
dbemiller: I think this should be an enum? To identify the 3 legal values.

config/config.go
dbemiller: `SetupViper` also sounds good.

I'm just shooting for something which publicizes the fact that it actually _mutates_ the argument object. `Set`, `Setup`, `Init`, etc. do that. `Configure` isn't bad either... but `Config` is used as a noun too frequently.

config/config_test.go
hhhjort: The intended default values for these are the zero values, or a URL that has apparently changed in the past. I thought it was a weak test to test that zero values were still zero values, and I wasn't sure about changing the actual default values in this PR. Could just check for the current URL, and the test failure will prompt us to update the test if the URL changes again.

config/stored_requests.go
hhhjort: Yes, would be more proper with an enum. :)

config/config_test.go
dbemiller: ahh, gotcha. I agree, that makes sense.

In that case let's just remove those assertions. There's no value in testing test code.

config/stored_requests.go
hhhjort: Actually, on second thought I think an enum would add overhead without any benefit. The config value only gets set from Viper, so any incoming string will be forced into the `MemCacheType` type, eliminating any constraints on the value that go "enum" could enforce. The code never changes this later, which is where type safety comes in. And finally, we have the `validate()` method we call after reading in the configuration that will enforce the value to be one of the list.

config/config_test.go
hhhjort: Already replace, found a couple of ints with non-zero defaults to test for instead.

dbemiller:
Before this, A config like:

```
stored_requests:
  postgres:
    in_memory_cache:
      ttl_seconds: ###
      request_cache_size_bytes: ###
      imp_cache_size_bytes: ###
```

would create an in-memory stored request cache.

After this, it needs a `Type` should be:

```
stored_requests:
  postgres:
    in_memory_cache:
      type: "unbounded"   // could also be "none" or "lru"
      ttl_seconds: ###
      request_cache_size_bytes: ###
      imp_cache_size_bytes: ###
```

The `ttl_seconds`, `request_cache_size_bytes`, and `imp_cache_size_bytes` will only apply to `lru` caches.

That makes this a breaking change for PBS hosts.

hhhjort:
Oh, yes, that is true, thanks. I was hoping to do all this without introducing a breaking change, but then found that there were cases where a non-"none" cache could have a `0` in any one of the current parameters, so I needed to introduce the type. Didn't reconsider the label after I changed it.

PR #516, opened 2018-05-21
Adapter for Somo Audience Bidder

bryandelong:
Official Somoaudience Adapter

PR #544, opened 2018-06-04
Cleaning up some maps and syncer docs

dbemiller:
This PR just has a few minor cleanups.

I alphabetized the various maps, because this reduces the chance of code conflicts when multiple adapters get submitted.

I also noticed in #516 that they were following the docs, but the docs for `FamilyName()` on the `Usersyncer` were outdated. This updates the docs, leaves a TODO there, and updates their usersync key to be the bidder name instead, to be consistent with all the other bidders out there.

### DIFF
--- a/exchange/adapter_map.go
+++ b/exchange/adapter_map.go
@@ -31,8 +31,8 @@ func newAdapterMap(client *http.Client, cfg *config.Configuration) map[openrtb_e
 		openrtb_ext.BidderAdform:      adaptBidder(adform.NewAdformBidder(client, cfg.Adapters["adform"].Endpoint), client),
 		openrtb_ext.BidderAdtelligent: adaptBidder(adtelligent.NewAdtelligentBidder(client), client),
 		openrtb_ext.BidderAppnexus:    adaptBidder(appnexus.NewAppNexusBidder(client, cfg.Adapters["appnexus"].Endpoint), client),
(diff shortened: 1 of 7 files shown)


exchange/adapter_map.go
```diff
@@ -28,9 +28,13 @@ import (
 
 func newAdapterMap(client *http.Client, cfg *config.Configuration) map[openrtb_ext.BidderName]adaptedBidder {
 	return map[openrtb_ext.BidderName]adaptedBidder{
-		openrtb_ext.BidderAppnexus: adaptBidder(appnexus.NewAppNexusBidder(client, cfg.Adapters["appnexus"].Endpoint), client),
+		openrtb_ext.BidderAdform:      adaptBidder(adform.NewAdformBidder(client, cfg.Adapters["adform"].Endpoint), client),
+		openrtb_ext.BidderAdtelligent: adaptBidder(adtelligent.NewAdtelligentBidder(client), client),
+		openrtb_ext.BidderAppnexus:    adaptBidder(appnexus.NewAppNexusBidder(client, cfg.Adapters["appnexus"].Endpoint), client),
 		// TODO #267: Upgrade the Conversant adapter
```
cirla: Silly thing but this comment is separated from its adapter line

dbemiller: good catch... fixed.

PR #547, opened 2018-06-06
Bad request metrics

dbemiller:
Currently we take metrics of request counts and errors. This splits "error" into "bad request input" and "real, unexpected error."

I also noticed that the existing metrics overlapped quite a bit. If I wanted to know how many "successful" requests were made to an endpoint, I would have to _subtract_ the "errors" from the "totals".

In general it's much easier for a time-series database to aggregate than to subtract. You can get a "total request count" graph by replacing the `.ok.` with a `.*.`... but I'm not even sure how to build a `.ok.` from a `.total.` and `.errors`.

So... I also reorganized the request counts a bit, and delete old ones. This means that the following metrics disappear:

- `prebidserver.requests`
- `prebidserver.error_requests`
- `prebidserver.ortb_requests`
- `prebidserver.amp_requests`

And these new ones will appear instead:

- `prebidserver.requests.ok.openrtb2`
- `prebidserver.requests.badinput.amp`
- `prebidserver.requests.err.legacy`

I believe that the following queries can be used to reproduce the old behavior:

- `prebidserver.requests`=> `prebidserver.requests.*.*`
- `prebidserver.ortb_requests` => `prebidserver.requests.*.openrtb2`
- `prebidserver.amp_requests` => `prebidserver.requests.*.amp`
- `prebidserver.error_requests` => `prebidserver.requests.(err|badinput).*`


dbemiller:
yeah... I was back and forth on it, but it probably should.

bretg:
looks good.

PR #550, opened 2018-06-07
More granular adapter error metrics

dbemiller:
This PR captures better adapter metrics, and fixes some bugs I ran into along the way.

It deletes a few old metrics, and introduces the following:

`prebidserver.adapter.{bidder}.requests.gotbids`: a bid request was made and bids were returned
`prebidserver.adapter.{bidder}.requests.nobid`: a bid request was made, but no bids were returned
`prebidserver.adapter.{bidder}.requests.badinput`: a bid request was made with some invalid input
`prebidserver.adapter.{bidder}.requests.badserverresponse`: a bid request was made and the server responded with an error or something unexpected.
`prebidserver.adapter.{bidder}.requests.timeout`: a bid request was made and the bidder timed out.
`prebidserver.adapter.{bidder}.requests.unknown`: a bid request was made and some unknown error happened.

One risky change in here is that the "account X adapter" request counter was also split into `gotbids` and `nobid`. If you have lots of accounts, this could potentially cause problems in your metrics collector.

Some _loose_ translation from old => new metrics are:

`prebidserver.adapter.{bidder}.requests` => `prebidserver.adapter.{bidder}.requests.(gotbids|nobid)`
`prebidserver.adapter.{bidder}.error_requests` => `prebidserver.adapter.{bidder}.requests.(badinput|badserverresponse|unknown)`
`prebidserver.adapter.{bidder}.timeout_requests` => `prebidserver.adapter.{bidder}.requests.timeout`

However, these are not exact either. Previously, a nobid-with-errors would get logged as _one of_ those error conditions. Now we will log a nobid event _and_ an error event.

---------------------------------------------------------------------
This also fixed a bug in bid price logging. In `/openrtb2/auction`, bidders are allowed to return bids _and_ errors. For example, "you gave us a banner & video Imp, and we don't bid on video... but here are some banner bids."

Previously, the code was only logging bids and bid prices if there were no errors. Now it logs all bids.


bretg:
1) It would be good to know how often each adapter was called for which type of request. Similar to PR #547 , could we add the request-type to the adapter? e.g.
```
requests.adapter.*.requests.type.{ortb-web,ortb-app,amp,legacy}
```
Note this a little different than what's in #547 as it breaks openrtb2 into ortb-web and ortb-app. Any chance of tweaking that at the request level? And maybe also adding `type`:
```
requests.type.{ortb-web,ortb-app,amp,legacy}
```

2) likewise for accounts
```
requests.adapter.*.requests.type.{ortb-web,ortb-app,amp,legacy}
```

pbsmetrics/go_metrics.go
```diff
@@ -312,20 +314,22 @@ func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
...
-	case AdapterStatusTimeout:
-		am.TimeoutMeter.Mark(1)
+		aam.NoBidMeter.Mark(1)
+	case AdapterBidPresent:
+		am.GotBidsMeter.Mark(1)
+		aam.GotBidsMeter.Mark(1)
+	default:
+		glog.Warningf("No go-metrics logged for AdapterBids value: %s", labels.AdapterBids)
 	}
+	for errType := range labels.AdapterErrors {
```
hhhjort: This may be an issue for any metrics backend that supports true labels. Normally you would want a label "error" with the value indicating the error type, or the value "none". To support multiple error types, you would need either multiple labels, one for each error type and treat the value for each as true/false. Or record multiple requests, one for each error type.

Considering that each individual request to a bidder should only be able to result in one error, may be better to record each success as a metrics event, and each error as a metrics event. With the current implementation we are losing info about the success rate.

Take for an example an adapter that on average generates 4 requests per auction, and generates on average 1.5 errors. If we look at the metrics, we will see an error rate of 150%, which is of course crazy. The true error rate was actually 37.5%. Tone this down to 0.5 errors on average and the metrics will report a 50% error rate when only 12.5% of requests to the bidder generated an error. Better to record on a per request basis if we care about more than "did the adapter perform flawlessly in the auctions?"

dbemiller:
> requests.adapter.*.requests.type.{ortb-web,ortb-app,amp,legacy}

Logged as #551. Since it's not a breaking change, there's no sense in smooshing it into this PR.

> requests.account.*.requests.type.{ortb-web,ortb-app,amp,legacy}

Logged as #552. Same reason.

> requests.type.{ortb-web,ortb-app,amp,legacy}

This sounds good & I did it now. #547 already made similar breaking changes and hasn't been released yet... so we should consolidate them.

`prebidserver.requests.ok.openrtb2` is changed into `prebidserver.requests.ok.openrtb2-web` and `prebidserver.requests.ok.openrtb2-app`. Same with the `requests.badinput.*` and `requests.err.*` metrics.

pbsmetrics/go_metrics.go
dbemiller: > This may be an issue for any metrics backend that supports true labels.

Can't the backend choose how it logs events? This is an implementation for `go-metrics`, which doesn't support labels.

> Considering that each individual request to a bidder should only be able to result in one error

This isn't true... officially, [the docs](https://github.com/prebid/prebid-server/blob/master/adapters/bidder.go#L17) say:

> The errors should contain a list of errors which explain why this bidder's bids will be "subpar" in some way.

The goal for adapters is to return good error messages that will explain to a Publisher what went wrong. It could be one or many. A good example might be "you sent us a banner/video imp, but we don't support video" _and_ "our server returned a 5xx when we tried to fetch the banner bids".

--------------------------------------------------------------------

> With the current implementation we are losing info about the success rate.

You're right... but I don't think the problem is with the `AdapterErrors` list. Just this implementation of it. We could easily collect this data with:

```
if len(labels.AdapterErrors) == 0 {
  // log success
} else {
  // log error
}
```

That said... I must admit that I don't see the value in it. If you knew the error rate was high, what would you do about it? Suggest that other Pubs avoid it? Tell the Bidder about it? What if the cause was just a single high-traffic publisher who messed up their bidder params?

We can chat if you disagree, but... I can't think of any actionable data that we're losing here.

hhhjort: As for returning multiple errors, let me modify that by saying multiple types of errors. If one request returns multiple reasons why the request was subpar, that is still just the one type of error. We certainly cannot get both a 5xx error and an error why the request was subpar in a single request.

As for "If you knew the error rate was high, what would you do about it?", what is the purpose of having metrics on these errors if not to track the error rate? I don't see how making the error metrics more granular is useful if we are not tracking the error rate accurately.

Issue #549, opened 2018-06-07
[CI] Travis CI to publish binaries in releases


benjaminch:
Hey there,

Is there any plan / previous discussions regarding publishing binaries in releases tags?
In my opinion, it would help everyone plugged into prebid server to setup continuous integration tests in house.

https://docs.travis-ci.com/user/deployment/releases/

What do you think?

Cheers,
Benjamin

benjaminch:
Alright ! 
I will try to have a look to this one then.
Thanks

dbemiller:
yep yep. My experience with docker has been _really_ positive. Would definitely recommend it.

If you still think binaries would be useful though, circle back and we can see what we can do.

benjaminch:
Thing is, on our end, if we can avoid relying on Docker and have directly binaries it's better. This way, we are sure that the app is in the expected state and have just to pull and deploy it from the latest release and then launch our integration tests.
In my experience, Docker is really nice as well but can be tricky when we need to know how the Docker container was built.

Anyways, I drafted a PR for this purpose it works on my end (in our fork). We would just need to update the Github encrypted token so Travis can publish sources in releases in Github and changing the branches on which you want to publish binaries.

dbemiller:
Alrighty, that's totally fine. For what it's worth, there is [a Dockerfile](https://github.com/prebid/prebid-server/blob/master/Dockerfile) in this repo, so you _can_ see exactly how it's built :).

...but you're right, certain info (like "which Go version was used to compile the binary?") does get lost. And I can also understand not wanting to rely on Docker.

Which branch has the changes in your fork?

benjaminch:
Oh, I just pushed the PR: on our Fork, the branch is https://github.com/criteo-forks/prebid-server/tree/publishing-releases-binaries

Let me know what you think.
Cheers

dbemiller:
hmm... so I released version 0.32.0 today, and the binaries didn't show up in the release note.

Any idea why not? I triggered the release with:

```
git tag 0.32.0
git push origin 0.32.0
```

on the `master` branch

benjaminch:
It seems it appears a bit after, but to me, it seems that binaries are there !
https://github.com/prebid/prebid-server/releases

![image](https://user-images.githubusercontent.com/1393795/41297873-6aba2ef8-6e60-11e8-9e09-a5f14f8a0b56.png)

dbemiller:
[...]

PR #553, opened 2018-06-08
Publishing releases binaries with Travis #549

benjaminch:
Publishing releases binaries with Travis #549

This change allows the following:
 * Deploy binaries only if a tag has been created
 * Make binaries available into the releases tags

Couple changes needs to be introduced prior merging such as:
 * Changing encrypted Github token for deploy (https://docs.travis-ci.com/user/deployment/releases/)
 * Changing branches for which you want to deploy binaries when a tag is created

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: go
+
 go:
-    - "1.9"
-    - "1.10"
+    - '1.9'
+    - '1.10'
 
 env:
     - DEP_VERSION="0.4.1"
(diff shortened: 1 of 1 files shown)


dbemiller:
Nice! Thanks. I'll have to dig into API keys a bit. I don't actually have admin access to this repo... so I'll have to read some docs and see if it's something I can do, or something i need to poke @mkendall07 about when he gets back from vacation.

.travis.yml
```diff
@@ -1,17 +1,34 @@
...
+    - go get github.com/mitchellh/gox
 
 install:
-    - dep ensure
+    - $GOPATH/bin/dep ensure
 
 script:
-    - ./validate.sh --nofmt --cov --race 10
+    - "./validate.sh --nofmt --cov --race 10"
+    - gox -os="linux" -arch="386" -output="{{.Dir}}_{{.OS}}_{{.Arch}}" -ldflags "-X main.Rev=`git rev-parse --short HEAD`" -verbose ./...;
```
dbemiller: Did you try this line in a `before_deploy` script instead? I haven't used this feature before, but the docs say that should make it run only when a release is happening.

You could probably move the `go get` line there too. Otherwise they'll run on every PR, which slows things down.

benjaminch:
Alright, thanks @dbemiller ! 
Please let me me know if I can help you with anything !

.travis.yml
benjaminch: Very good point, I’ll update this

dbemiller:
Actually... I think it turned out to be really easy. I stole these commits and opened #554 with the needed changes.

dbemiller:
This was included by #554... so gonna close it.

benjaminch:
Thanks !

PR #554, opened 2018-06-08
Include binaries in releases

dbemiller:
This includes #553, with a few extra changes:

- The API key to one I just made & encrypted
- The repo to `prebid/prebid-server`
- The release branch to `master`
- Moves some commands to `before_deploy` so that they don't slow down normal PR builds

Thanks @benjaminch!

If you have time to test out using `before_deploy` in your fork, I'd definitely appreciate it. If not, not a huge deal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,18 +17,21 @@ install:
 
 script:
     - "./validate.sh --nofmt --cov --race 10"
+
+before_deploy:
+    - go get github.com/mitchellh/gox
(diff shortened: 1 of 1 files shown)


benjaminch:
Thanks @dbemiller !
Just wondering if we should make possible to read the github encrypted token from env instead of hard coding it. This way, every forks would be able to leverage this feature.
I will check if this is doable.

dbemiller:
solid. Gonna merge this then... we'll see how it works next time we release.

PR #556, opened 2018-06-12
Fixed the validation logic on unbounded caches

dbemiller:
Bug from #539 which is blocking release... unbounded caches don't require a TTL. In fact they don't support one at all.

I also tightened up the validation on cache size, just because it was a little loose.

### DIFF
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -330,8 +330,14 @@ func (inMemCache *InMemoryCache) validate() error {
 	case "none":
 		return nil
 	case "unbounded":
-		if inMemCache.TTL <= 0 {
-			return errors.New("Stored requests In-Memory caches need a TTL when unbounded")
+		if inMemCache.TTL != 0 {
+			return fmt.Errorf("stored_requests.in_memory_cache must be 0 for unbounded caches. Got %d", inMemCache.TTL)
+		}
+		if inMemCache.RequestCacheSize != 0 {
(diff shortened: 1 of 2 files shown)


config/stored_requests.go
```diff
@@ -330,8 +330,14 @@ func (inMemCache *InMemoryCache) validate() error {
 	case "none":
 		return nil
 	case "unbounded":
-		if inMemCache.TTL <= 0 {
-			return errors.New("Stored requests In-Memory caches need a TTL when unbounded")
+		if inMemCache.TTL != 0 {
```
hhhjort: So ... unbounded memory caches need to have to TTL and no max size, so we cache everything forever until the OOM killer gets to it?

Should be `if inMemCache.TTL == 0`

config/stored_requests_test.go
```diff
@@ -75,6 +75,59 @@ func TestPostgressConnString(t *testing.T) {
 	assertHasValue(t, params, "sslmode", "disable")
 }
 
+func TestInMemoryCacheValidation(t *testing.T) {
+	assertNilErr(t, (&InMemoryCache{
```
hhhjort: This test will fail when you fix validation above.

config/stored_requests.go
dbemiller: Yes, that's exactly how it works today :). It's just implemented using a `sync.Map`, which doesn't support any of those options

hhhjort: Ahh, ok. Nevermind then.

PR #558, opened 2018-06-12
Consolidates 2 metrics engine functions and fixes missing metric from…

hhhjort:
… exchange

`RecordAdapterBidsReceived()` was missing from the openrtb code. Also this is mostly redundant with `RecordAdapterBidAdm()` as the total count from the second is the metric from the first. This collapses the counting into `RecordAdapterBidAdm()`, which happens to fix the missing `RecordAdapterBidsReceived()` from openrtb.

### DIFF
--- a/pbsmetrics/go_metrics.go
+++ b/pbsmetrics/go_metrics.go
@@ -331,21 +331,6 @@ func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
 	}
 }
 
-// RecordAdapterBidsReceived implements a part of the MetricsEngine interface. This tracks the number of bids received
-// from a bidder.
-func (me *Metrics) RecordAdapterBidsReceived(labels AdapterLabels, bids int64) {
-	am, ok := me.AdapterMetrics[labels.Adapter]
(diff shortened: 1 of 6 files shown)


pbsmetrics/go_metrics.go
```diff
@@ -331,21 +331,6 @@ func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
...
-	// Adapter metrics
-	am.BidsReceivedMeter.Mark(bids)
-	// Account-Adapter metrics
-	aam := me.getAccountMetrics(labels.PubID).adapterMetrics[labels.Adapter]
-	aam.BidsReceivedMeter.Mark(bids)
-}
-
 // RecordAdapterBidAdm implements a part of the MetricsEngine interface.
 // This tracks how many bids from each Bidder use `adm` vs. `nurl.
 func (me *Metrics) RecordAdapterBidAdm(labels AdapterLabels, bidType openrtb_ext.BidType, hasAdm bool) {
```
dbemiller: Should we rename this to `recordAdapterBidReceived`?

PR #562, opened 2018-06-13
Cookie sync URL gdpr support for Brightroll adapter

smithaammassamveettil:
Description:

Updated Brightroll cookie sync URL to add the expecting gdpr signals as URL parameters.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -74,7 +74,7 @@ adapters:
     usersync_url: http://facebook.com/ortb/prebid-s2s
     platform_id: abcdefgh1234
   brightroll:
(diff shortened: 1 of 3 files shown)


dbemiller:
These changes look fine, but... just making sure. You're not actually changing the `SetDefault` line inside `config.go` in this PR... and when I look in the code, it already has these properties:

`v.SetDefault("adapters.brightroll.usersync_url", "http://east-bid.ybp.yahoo.com/sync/appnexuspbs?gdpr={{gdpr}}&euconsent={{gdpr_consent}}&url=")`

Is the goal here to fix a bug where the `url.QueryEscape(userSyncUrl)` escaped your `{` and `}` characters, so then `resolveMacros` didn't substitute them properly?

PR #563, opened 2018-06-14
Changes configuration logging to be reflect-based.

### LINKED ISSUE #561: Use reflection to log Configuration values
Right now, we log the app config manually. This is prone to fall out of sync as we add/move properties around.

As suggested by [this comment](https://github.com/prebid/prebid-server/pull/560#pullrequestreview-128445961), it would be awesome if we could add a tag and log these via reflection.

> ExternalURL string `mapstructure:"external_url" log:"true"`

hhhjort:
Fixes #561

### DIFF
--- a/config/structlog.go
+++ b/config/structlog.go
@@ -1,4 +1,4 @@
-package structlog
+package config
 
 import (
 	"fmt"
@@ -9,15 +9,14 @@ import (
 	"github.com/golang/glog"
 )
 
-var mapregex *regexp.Regexp
-var blacklistregexp []*regexp.Regexp
-
-func init() {
-	mapregex = regexp.MustCompile(`mapstructure:"([^"]+)"`)
-	blacklistregexp = make([]*regexp.Regexp, 1, 1)
-	blacklistregexp[0] = regexp.MustCompile("password")
+var mapregex *regexp.Regexp = regexp.MustCompile(`mapstructure:"([^"]+)"`)
+var blacklistregexp []*regexp.Regexp = []*regexp.Regexp{
+	regexp.MustCompile("password"),
 }
 
+// LogGeneral will log nearly any sort of value, but requires the name of the root object to be in the
+// prefix if you want that name to be logged. Structs will append .<fieldname> recursively to the prefix
+// to document deeper structure.
 func LogGeneral(v reflect.Value, prefix string) {
 	logGeneralWithLogger(v, prefix, glog.Infof)
 }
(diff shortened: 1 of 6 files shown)


structlog/structlog.go
```diff
@@ -0,0 +1,104 @@
+package structlog
```
dbemiller: I know this isn't exactly config, but... I'd still recommend keeping it in the `config` package for now.

Motivations being:

1. We have no other use-cases on the radar in the near future.
2. By using `mapstructure`, its usage is basically tied to Viper structs anyway.
3. If we really generalized it enough to be reusable, it should probably be its own repo/project. However, it's tough to see what's "too specialized" in it until you have more use-csaes.

So... IMO, better to keep it in `config` until we have some more use-cases.

structlog/structlog.go
```diff
@@ -0,0 +1,104 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+var mapregex *regexp.Regexp
+var blacklistregexp []*regexp.Regexp
+
+func init() {
```
dbemiller: Let's not (ever, if we can avoid it) use `init()` functions... they complicate the code's lifecycle, and tbh I have no idea what value they actually add. I'm [not alone](https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html) here either.

In this case, you can easily define them inline:

```
var mapregex *regexp.Regexp = regexp.MustCompile(`mapstructure:"([^"]+)"`)
var blacklistregexp []*regexp.Regexp = []*regexp.Regexp{
  regexp.MustCompile("password"),
}
```

(ironically, I understand that the article also recommends "no package-level variables" too. I think he's right about that in general, except that immutable/threadsafe ones like these are ok. If _all_ package-level vars were bad, then we'd never declare `const`s either... which I think is a little ridiculous)

structlog/structlog.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func extendPrefix(prefix string, field string) string {
+	if len(strings.Trim(prefix, " \t")) == 0 {
+		return fmt.Sprintf("%s%s", prefix, field)
+	}
+	return fmt.Sprintf("%s.%s", prefix, field)
+}
+
+func logInt(v reflect.Value, prefix string, logger func(msg string, args ...interface{})) {
```
dbemiller: I'm very sympathetic to small functions, but... these are only called once, aren't they? Subjective, but... I think the code reads better if they're inlined.

structlog/structlog.go
```diff
@@ -0,0 +1,104 @@
...
+func LogGeneral(v reflect.Value, prefix string) {
+	logGeneralWithLogger(v, prefix, glog.Infof)
+}
+
+func logGeneralWithLogger(v reflect.Value, prefix string, logger func(msg string, args ...interface{})) {
+	switch v.Kind() {
+	case reflect.Struct:
+		logStructWithLogger(v, prefix, logger)
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		logInt(v, prefix, logger)
```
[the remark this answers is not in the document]
hhhjort: I did it that way to resemble the struct case and be more easily extensible if we want further functionality in the future. I was initially expecting to need more than just one short line though, so if you really think it would be better inlined, I don't have a problem with that.

structlog/structlog.go
hhhjort: I'll modify

config/structlog.go
```diff
@@ -0,0 +1,79 @@
...
+
+var mapregex *regexp.Regexp = regexp.MustCompile(`mapstructure:"([^"]+)"`)
+var blacklistregexp []*regexp.Regexp = []*regexp.Regexp{
+	regexp.MustCompile("password"),
+}
+
+// LogGeneral will log nearly any sort of value, but requires the name of the root object to be in the
+// prefix if you want that name to be logged. Structs will append .<fieldname> recursively to the prefix
+// to document deeper structure.
+func LogGeneral(v reflect.Value, prefix string) {
```
dbemiller: Can be private now

hhhjort: I did it the other way around. :)

PR #565, opened 2018-06-15
[Travis] Adding support for forks testing

benjaminch:
This CL allows to add testing support in Travis for forks leveraging
the `dep ensure --add` option to support this use case.

As of today, `dep ensure` will pull the official prebid-server repo and
compile it but won't rely on the forked repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,10 @@ before_install:
     - go get github.com/mitchellh/gox
 
 install:
-    - $GOPATH/bin/dep ensure
+    # Making sure to dep ensure with the current repository URL to avoid any packages issues
+    # if the build is triggered from a fork.
+    # dep will make sure to get the branch from which the commit originates.
+    - if [ $TRAVIS_REPO_SLUG == "prebid/prebid-server" ]; then dep ensure; else eval "dep ensure -add github.com/prebid/prebid-server:github.com/$TRAVIS_REPO_SLUG@$TRAVIS_BRANCH"; fi
 
 script:
     - "./validate.sh --nofmt --cov --race 10"


dbemiller:
hey @benjaminch: this looks a little funky to me... what exactly are you trying to do here?

`dep` uses `Gopkg.toml` and/or `Gopkg.lock` to create a `vendor` directory at the project root. It then downloads all the dependencies needed to make this project run.

I wouldn't expect `dep` to install `prebid/prebid-server` as a dependency at all--in this repo or a fork. We'd expect PBS to be used like a Service (deploy it and send HTTP requests to/from it), not imported like a library.

benjaminch:
Hey @dbemiller thanks for jumping on this one !

Actually, it seems that the travis testing doesn't work on forks in case of multi-package commits.
It seems that dep ensure was actually pulling the `prebid/prebid-server` sources into vendors instead of `fork/prebid-server` . Maybe I am mistaken but I don't see how it would manage not to do so?

I am doing some tests on this front and will share the results to you asap. If it works, then I will cancel this PR.

Do you have any examples of golang forks working properly in a multiple packages commits tests on travis?

Cheers

dbemiller:
I don't, no. In general I'm not really sure how forks work with Go--in particular `import` statements and `go get`.

Throughout the source code, the imports [are hardcoded as `prebid/prebid-server`](https://github.com/prebid/prebid-server/blob/master/pbs_light.go#L29).

`dep` is just a dependency manager. My guess is that it sees those imports in your codebase, realizes that you're in `criteo-forks/prebid-server`, and _thinks_ that `prebid/prebid-server` is a library that you're trying to import. Then it tries to download `prebid/prebid-server` source code to satisfy the import.

Honestly I'm not sure what the best way around this is. I did find [this blog post](https://blog.sgmansfield.com/2016/06/working-with-forks-in-go/), which had a nice developer workflow... but I'm not sure how you would integrate it with CI. Still, might be useful for you?

benjaminch:
Actually, you can see a failling example here:
- Commit : https://github.com/criteo-forks/prebid-server/commit/6adf0cc2b7a3a9e2fce5330267b805bb0b515124 
- Travis test : https://travis-ci.org/criteo-forks/prebid-server/builds/392770689?utm_source=github_status&utm_medium=notification

And the stacktrace looks fishy to me: 

```
$ $GOPATH/bin/dep ensure
Solving failure: No versions of github.com/prebid/prebid-server met constraints:
	0.32.1: Could not introduce github.com/prebid/prebid-server@0.32.1, as its subpackage github.com/prebid/prebid-server/adapters/criteo is missing. (Package is required by (root).)
	0.32.0: Could not introduce github.com/prebid/prebid-server@0.32.0, as its subpackage github.com/prebid/prebid-server/adapters/criteo is missing. (Package is required by (root).)
	0.31.0: Could not introduce github.com/prebid/prebid-server@0.31.0, as its subpackage github.com/prebid/prebid-server/adapters/criteo is missing. (Package is required by (root).)
	0.30.0: Could not introduce github.com/prebid/prebid-server@0.30.0 due to multiple problematic subpackages:
	Subpackage github.com/prebid/prebid-server/adapters/brightroll is missing. (Package is required by (root).)	Subpackage github.com/prebid/prebid-server/adapters/criteo is missing. (Package is required by (root).)	Subpackage github.com/prebid/prebid-server/adapters/somoaudience is missing. (Package is required by (root).)
```

I am pretty sure it has something to do with what dep pulls.

I'll continue to investigate to see if there is another solution.
By the way, if I apply this change (in travis) along with my multi-packages commit, Travis tests passes just fine.

hhhjort:
You could try something like this. Assuming your fork is in `.../github.com/criteo-forks/prebidserver/`, create a simlink `.../github.com/prebid` which points to `.../github.com/criteo-forks`, and set your working directory to be in this simlink when you do your builds. That way all the scripts think they are working in the original project and don't try importing it as a dependency. Not sure how exactly you would incorporate simlinks into your build automation exactly, I don't have experience with that.

Another option might be to put a symlink in the `vendor` directory, so the tool thinks the dependency is already satisfied without having to download the original repo.

benjaminch:
Thanks @hhhjort, I thought about the sys link, the issue I have with that is that it requires an extra setup in the itself, i think it would be better to think about a generic solution which will work out of the box once the repo is forked without any extra setup.

Maybe we can create this syslink in the Travis yaml, was that what you were thinking off? 
The other option is to rely on an env var set in Travis which would be set in the forked Travis  repo but at the end it’s the same as what I did in this commit.

Do you see any issue with the current implementation ? 
I am trying to find the more generic way of doing it but cannot find anything simpler without having to update the forked repo only.

Thanks !

dbemiller:
> And the stacktrace looks fishy to me:
> ...
> I am pretty sure it has something to do with what dep pulls.

I don't think I was clear.

I do agree that failure is caused by dep pulls. I don't believe that this change addresses the root of the problem.

The core issue is that the source code imports `prebid/prebid-server`. When you fork the project, the source code in that fork  _still_ imports `prebid/prebid-server`.

`dep` is just doing its job here; it fetches anything you import. The problem is that you _don't want_ our code imported when you're running and testing your fork.

benjaminch:
Ok clear. However, if I add the change in the Travis file to my commit, the build works just fine.
I am open to any generic solution which can make it work properly without having to do anything in the forked repo. 
The dep add option seems to do the job properly if we are on a fork.
The other option will be not to run Travis on forks but it would be a pity.
I’ll continue to check if there is any other solution around other open sources projects, but open to any ideas.

Thanks

hhhjort:
The problem with this solution is every module in the forked project will be importing the original `prebid/prebid-server` modules rather than importing your project's version of them. This is fine if the forked code just has small isolated changes, but it will break once you make a change in two modules where the change in one module depends on the change in the other. Not sure how to avoid that without some simlink trickery, or a script to rewrite all the imports to import from the fork's path rather than the original path.

dbemiller:
yeah... @benjaminch I believe you would hit the problem if you run this in your fork:

- Add `const SomeValue = "foo"` somewhere in `openrtb_ext`. 
- Add a test somewhere in `usersync` which asserts the value of that const. Make sure that the test file imports `github.com/prebid/prebid-server/openrtb_ext`.

If I'm right, I'd expect the CI in your fork to fail, even though your code changes are perfectly valid.

benjaminch:
@dbmiller, I will try tomorrow to test that !
@hhhjort, I think I was not clear enough but I agree with you that dep pulls the wrong libs. Hence the option added to ensure which will do the black magic pulling the fork repo as if it was the official one.

In the dep ensure command help, we have:

<import path>[:alt source URL][@<constraint>]

So to add github.com/foo/bar from location github.com/fork/bar you have to add it like this:

dep ensure -add github.com/foo/bar:github.com/fork/bar
The source location will be added as source attribute in the Gopkd.toml.

Cf: https://stackoverflow.com/questions/49475290/go-dep-and-forks-of-libraries

hhhjort:
Ahh, I missed that bid of magic. That may work as long as you are always testing on the master branch of your fork. If not, then you will run into the issue that you are pulling the master branch of your fork, so when testing against something else you link the wrong version. Could be an issue if travis is trying to verify that a branch can be merged into master or something, as all the linked modules will be the current master version, not the potential new master I suspect.

benjaminch:
I’ll dig further on this one, we are able to get the branch from Travis already, I just now need to be sure to pull the proper branch now, either we do it via an extra hit command or maybe there is a way to specify it via dep ensure.

benjaminch:
Hey guys, just found that we can properly specify the branch along with the repo. 
I just updated the code. 
Also note that it works on the fork, meaning tests are passing. It shouldn't have any impacts on the official repo since we add the `dep ensure -add` option only if we are in a forked repo.

Let me know what do you think.

Cheers

benjaminch:
Hey @dbemiller !

I do understand your point, however, I wanted to have an exact copy of the original repo without any workarounds nor overrides, so we can checkout your master without having to over think about what needs to stay or not. Maybe there is a more generic solution or something cleaner but digging on the subject, I didn't found out anything else.

I think it will allow every forks to be able to test their devs, but I do understand if you want to wait and see if other people needs this feature or not.

Cheers

dbemiller:
yeah, that also makes sense. It's a bit of a zero-sum game, then. We'd be adding a small bit of dead logic to the core project in order to save forks from having to add a small workaround.

I'll discuss with the team & see what they think.

benjaminch:
Thanks @dbemiller, let me know !

dbemiller:
After some discussion, we decided not to accept this one. Reasons include:

1. `.travis.yml` is intended to be specific to the repo in which it lives. This expands ours beyond its intent.
2. If a future PR proposes larger changes to support their fork, we don't want to have the "but Criteo was allowed do it" argument.
3. We think you have two very reasonable workarounds. You can trigger CI by opening PRs against this repo, or edit the `.travis.yml` inside your fork to make CI work for you on internal ones.

benjaminch:
Hey @dbemiller, ok fair enough, thanks for the feedback :)

PR #566, opened 2018-06-15
Moves metrics definitions to its own package to avoid circular depend…

hhhjort:
…encies adding new metrics backends.

Ran into an issue developing a second metrics backend into its own package. There are definitions needed to implement the metrics interface, but they are defined in `pbsmetrics`. The master metrics package cannot keep them and import packages implementing the interface.

We should also move the go-metrics package to its own package at some point.

### DIFF
--- a/endpoints/cookie_sync.go
+++ b/endpoints/cookie_sync.go
@@ -17,6 +17,7 @@ import (
 	"github.com/prebid/prebid-server/gdpr"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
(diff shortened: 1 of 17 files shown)


dbemiller:
I hate to say it, but... could we just move the `NewMetrics` function into a new module?

This is the same dependency issue I've come across in the `analytics` and `stored_requests` packages. We have some core types, some implementations of those types, and a factory function which builds the implementation given some config options.

For those other packages, I added a `config` package and moved the `NewWhatever(cfg)` function into it. I think for sanity's sake, we should be consistent with that pattern until we find good reason to use a better one.

On the bright side, it's a much smaller changeset. Much less code changed elsewhere to move a few functions rather than all the constants.

PR #572, opened 2018-06-18
Infer imp.secure state

### LINKED ISSUE #569: Appnexus PBS ads are failing on https site (returning http script)
We have an https site and are running appnexus demand through appnexus PBS endpoint.  The ad property in the bid response object has an http: script but our site is https:, so we're getting a javascript error on any ad that wins and attempts to render:

image

### LINKED ISSUE #570: /openrtb2/auction should infer `secure` state of Imps from request
According to [the OpenRTB spec](https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf#page=18), `request.imp[i].secure` is:

> Flag to indicate if the impression requires secure HTTPS URL creative assets and markup, where 0 = non-secure, 1 = secure. If omitted, the secure state is unknown, but non-secure HTTP support can be assumed.

As a general rule, /openrtb2/auction does "infer" properties as long as they're not specified by the request. For example, it fills out the request with IP addresses, domain, user agent, auction type, etc, based on the HTTP headers, connection info, and the fact that it knows that it's a header bidding library.

I think we should do the same for `secure` status. If omitted, we can use the protocol of the request to infer whether or not to require `https` creatives.

dbemiller:
Fixes #569, #570.

This can be reviewed... but should wait for @bretg to see it before we merge anything.

### DIFF
--- a/docs/endpoints/openrtb2/auction.md
+++ b/docs/endpoints/openrtb2/auction.md
@@ -318,7 +318,7 @@ The error message in the response should describe how to "fix" the request to ma
 If the message is unclear, please [log an issue](https://github.com/prebid/prebid-server/issues)
 or [submit a pull request](https://github.com/prebid/prebid-server/pulls) to improve it.
 
-#### Imp Secure Status
+#### Determining Bid Security (http/https)
 
 In the OpenRTB spec, `request.imp[i].secure` says:
 
--- a/docs/endpoints/openrtb2/auction.md
+++ b/docs/endpoints/openrtb2/auction.md
@@ -318,6 +318,16 @@ The error message in the response should describe how to "fix" the request to ma
 If the message is unclear, please [log an issue](https://github.com/prebid/prebid-server/issues)
 or [submit a pull request](https://github.com/prebid/prebid-server/pulls) to improve it.
 
+#### Imp Secure Status
+
+In the OpenRTB spec, `request.imp[i].secure` says:
+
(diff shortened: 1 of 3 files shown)


bretg:
Here are the scenarios that should be tested for implicit determination of the secure flag:

- Insecure page, insecure ad iframe: secure=0
- Insecure page but secure ad iframe: secure=1
- Secure page, secure ad iframe: secure=1
- Secure page but insecure ad iframe: shouldn't happen, but send secure=1

hhhjort:
If PBS is being called from an iframe, will it even have access to tell if the enclosing page is secure or not? I thought one of the features of the iframe is that it isolates its content from access to the whole page.

dbemiller:
Nope... the server only has access to what the client sends it. In this case it's HTTP headers and the Prebid Server URL itself.

It would be up to Prebid.js to send `http` or `https` requests to PBS depending on the page details.

bretg:
So this is a real (edge case) scenario we ran into last year on our pre-Prebid wrapper:

- non-secure page
- secure iframe calling wrapper
- non-secure assets come back from the exchange caused browser errors

Our solution was to implement an override so pubs could make this work as expected in the admittedly unexpected situation.

dbemiller:
I opened https://github.com/prebid/Prebid.js/issues/2750 for Prebid.js to address.

I really don't know what we can do about this on the server.

PR #573, opened 2018-06-18
Adds map support to config logging.

### LINKED ISSUE #561: Use reflection to log Configuration values

hhhjort:
The adaptor configs are stored in a map, so were left out of logging. This fixes #561


hhhjort:
Added the test for a map case. Did surface a limitation that I panic if trying to log an unexported map. Doesn't make sense yet to have an unexported map in the config code, so just left a note warning of this limitation in the code.

PR #576, opened 2018-06-19
Changing custom parameter name

smithaammassamveettil:
Description:

Changing Brightroll adapter's custom parameter name from "publisherName" to "publisher"

### DIFF
--- a/adapters/brightroll/brightroll.go
+++ b/adapters/brightroll/brightroll.go
@@ -70,21 +70,21 @@ func (a *BrightrollAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapte
 	err = json.Unmarshal(bidderExt.Bidder, &brightrollExt)
 	if err != nil {
 		err = &adapters.BadInputError{
-			Message: "ext.bidder.publisherName not provided",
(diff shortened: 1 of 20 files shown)


adapters/brightroll/brightroll.go
```diff
@@ -70,21 +70,21 @@ func (a *BrightrollAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapte
...
 		err = &adapters.BadInputError{
-			Message: "ext.bidder.publisherName not provided",
+			Message: "ext.bidder.publisher not provided",
 		}
 		errors = append(errors, err)
 		return nil, errors
 	}
 
-	if brightrollExt.PublisherName == "" {
+	if brightrollExt.Publisher == "" {
```
dbemiller: Just FYI: You can(/probably should?) include this in your `static/bidder-params/brightroll.json` schema definition: https://spacetelescope.github.io/understanding-json-schema/reference/regular_expressions.html

PBS will return a 400 if publishers send you an empty one... so you'll be less likely to get calls from people who think their auctions are working, but don't see you returning bids.

PR #578, opened 2018-06-20
Prometheus adapter for metrics.

hhhjort:
This gives the option to use Prometheus as the backend metrics engine.


pbsmetrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,242 @@
...
+// NewMetrics constructs the appropriate options for the Prometheus metrics. Needs to be fed the promethus config
+// Its own function to keep the metric creation function cleaner.
+func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
+	// define the buckets for timers
+	timerBuckets := prometheus.LinearBuckets(0.05, 0.05, 20)
+	timerBuckets = append(timerBuckets, []float64{1.5, 2.0, 3.0, 5.0, 10.0, 50.0}...)
+
+	standardLabelNames := []string{"source", "type", "pubid", "browser", "cookie", "status"}
+
+	adapterLabelNames := []string{"source", "type", "pubid", "browser", "cookie", "status", "adapter"}
```
dbemiller: I stumbled on this in the Prometheus docs, which is a little concerning: https://prometheus.io/docs/practices/naming/#labels

> Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), **such as user IDs**, email addresses, or other unbounded sets of values.

The fact that they specifically include user IDs is a little telling.

Is it a breaking change to add labels later? If so, it might be a good idea to leave that one out and see how Prometheus works under "normal" circumstances. Then we can add it back in and see if it's really a problem.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,242 @@
...
+		adapterLabelNames, timerBuckets,
+	)
+	metrics.Registry.MustRegister(metrics.adaptTimer)
+	metrics.adaptBids = newCounter(cfg, "adapter_bids_recieved_total",
+		"Number of bids recieved from each bidder.",
+		bidLabelNames,
+	)
+	metrics.Registry.MustRegister(metrics.adaptBids)
+	metrics.adaptPrices = newHistogram(cfg, "adapter_prices",
+		"Value of the highest bids from each bidder.",
```
dbemiller: What makes this just "the highest" bids?

pbsmetrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,242 @@
...
+	bidLabels := prometheus.Labels{
+		"source":  string(labels.Source),
+		"type":    string(labels.RType),
+		"pubid":   labels.PubID,
+		"browser": string(labels.Browser),
+		"cookie":  string(labels.CookieFlag),
+		"status":  string(labels.AdapterStatus),
+		"adapter": string(labels.Adapter),
+		"bidtype": string(bidType),
+		"hasadm":  boolToString(hasAdm),
```
dbemiller: `hasadm` would probably be clearer as a `markup_type` label with the values `adm` and `nurl`.

More flexible too. For example, we might want to deprecate `adm` someday and replace it with `html`, `native`, `custom` (in the face of facebook's wonky response JSON), etc.

pbsmetrics/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,423 @@
...
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/pbsmetrics"
+)
+
+var gaugeValueRegexp = regexp.MustCompile("gauge:<value:([0-9]+) >")
+var counterValueRegexp = regexp.MustCompile("counter:<value:([0-9]+) >")
+var histogramValueRegexp = regexp.MustCompile("histogram:<sample_count:([0-9]+)")
+
+func TestConnectionMetrics(t *testing.T) {
+	proMetrics := newTestMetricsEngine().(*Metrics)
```
dbemiller: Can you get rid of all these casts if you just change the return type on `newTestMetricsEngine` to `*Metrics`?

pbsmetrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,242 @@
...
+func (me *Metrics) RecordConnectionClose(success bool) {
+	if success {
+		me.connCounter.Dec()
+	} else {
+		me.connError.WithLabelValues("close_error").Inc()
+	}
+}
+
+func (me *Metrics) RecordRequest(labels pbsmetrics.Labels) {
+	me.requests.With(resolveLabels(labels)).Inc()
```
dbemiller: so... one major warning here. The implementation on this `With()` function ultimately handles a lock:

```
// getOrCreateMetricWithLabelValues retrieves the metric by hash and label value
// or creates it and returns the new one.
//
// This function holds the mutex.
func (m *MetricVec) getOrCreateMetricWithLabels(hash uint64, labels Labels) Metric {
	m.mtx.RLock()
	metric, ok := m.getMetricWithLabels(hash, labels)
	m.mtx.RUnlock()
```

I don't think this should be handled in this PR... but if we see weird issues (especially on startup, when all these metric/label combinations are being created), we might have to be smarter about this.

I've been warned that global locks will wreck performance in these high-traffic systems... so it might be worth opening an issue about, and logging with a `// TODO #XYZ` comment so that we don't forget.

In the future (assuming you remove `publisher_id` from the labels), we could call this function with all the label combinations on app startup so that we're only grabbing read-locks while serving real traffic.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,242 @@
...
+	cookieSync    prometheus.Counter
+	userID        *prometheus.CounterVec
+}
+
+// NewMetrics constructs the appropriate options for the Prometheus metrics. Needs to be fed the promethus config
+// Its own function to keep the metric creation function cleaner.
+func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
+	// define the buckets for timers
+	timerBuckets := prometheus.LinearBuckets(0.05, 0.05, 20)
+	timerBuckets = append(timerBuckets, []float64{1.5, 2.0, 3.0, 5.0, 10.0, 50.0}...)
```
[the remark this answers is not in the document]
hhhjort: Keeping the config small until we have a definite use case to change them. Easier to add config options later than be stuck supporting them for eternity.

pbsmetrics/prometheus/prometheus.go
hhhjort: A quick reading of the code brought me to this conclusion. I think I missed the loop, will need to fix the description, Thanks.

pbsmetrics/prometheus/prometheus.go
hhhjort: Two problems, one: it is possible that both are present. Two: our current code just checks if `adm` is set, so unless we want to change what we are recording here, this is what we are checking. If we want better `adm`/`nurl` metrics, that should probably be its own PR.

server/prometheus.go
```diff
@@ -0,0 +1,27 @@
...
+	var proMetrics *prometheusMetrics.Metrics
+
+	proMetrics = metrics.PrometheusMetrics
+
+	if proMetrics == nil {
+		glog.Fatal("Prometheus metrics configured, but a Prometheus metrics engine was not found. Cannot set up a Prometheus listener.")
+	}
+	return &http.Server{
+		Addr:    cfg.Host + ":" + strconv.Itoa(cfg.Metrics.Prometheus.Port),
+		Handler: promhttp.HandlerFor(proMetrics.Registry, promhttp.HandlerOpts{}),
```
dbemiller: I think `ErrorLog`, `MaxRequestsInFlight`, and `Timeout` are pretty important here... just so that we can diagnose errors and make sure that the prometheus mothership doesn't bury the server.

server/prometheus.go
```diff
@@ -0,0 +1,27 @@
...
+func newPrometheusServer(cfg *config.Configuration, metrics *metricsconfig.DetailedMetricsEngine) *http.Server {
+	var proMetrics *prometheusMetrics.Metrics
+
+	proMetrics = metrics.PrometheusMetrics
+
+	if proMetrics == nil {
+		glog.Fatal("Prometheus metrics configured, but a Prometheus metrics engine was not found. Cannot set up a Prometheus listener.")
+	}
+	return &http.Server{
+		Addr:    cfg.Host + ":" + strconv.Itoa(cfg.Metrics.Prometheus.Port),
```
dbemiller: This port needs to be exposed in the `Dockerfile`. Otherwise outside stuff (like Prometheus) won't be able to get at it.

server/prometheus.go
hhhjort: Are you asking to add metrics to record the prometheus interface activity, or something else?

dbemiller: They're properties you can set on `HandlerOpts`

pbsmetrics/config/metrics.go
```diff
@@ -33,14 +35,84 @@ func NewMetricsEngine(cfg *config.Configuration, adapterList []openrtb_ext.Bidde
...
+}
+
+// DetailedMetricsEngine is a MultiMetricsEngine that preserves links to unerlying metrics engines.
+type DetailedMetricsEngine struct {
+	metricsEngine     pbsmetrics.MetricsEngine
+	GoMetrics         *pbsmetrics.Metrics
+	PrometheusMetrics *prometheusmetrics.Metrics
+}
+
+// RecordRequest passthrough
```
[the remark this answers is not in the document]
hhhjort: The tricky thing here is that pbsmetrics.MetricsEngine is an interface, not a struct. It can hold 4 different concrete types. So I don't see how this construction could work, and if it did, how would we fill the interface with the actual type?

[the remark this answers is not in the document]
dbemiller:
```
		return &DetailedMetricsEngine{
			MetricsEngine:     &engineList,
			PrometheusMetrics: proMetrics,
			GoMetrics:         goMetrics,
		}
```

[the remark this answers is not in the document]
dbemiller: related: I found a last-minute bug in this code... you're never actually setting `GoMetrics` and `PrometheusMetrics` in the `DetailedMetricsEngine`. You should fix that regardless.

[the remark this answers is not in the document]
hhhjort: Good catch, though I fixed it differently. :)

PR #581, opened 2018-06-22
[adform adapter] Pass gdpr & digitrust to bid request

volhalink:
1. [adform adapter] Send GDPR consent and digitrust in request to Adform.
2. [pbs/pbsrequest.go] Fix a bug in ParseGDPR(): GDPR is stored as `*int8` in `req.Regs.Ext` and couldn't be read using `val, err := jsonparser.GetString(data, key)`


adapters/adform/adform.go
```diff
@@ -210,21 +253,45 @@ func toPBSBidSlice(adformBids []*adformBid, r *adformRequest) pbs.PBSBidSlice {
 // COMMON
 
 func (r *adformRequest) buildAdformUrl(a *AdformAdapter) string {
-	adUnitsParams := make([]string, 0, len(r.adUnits))
-	for _, adUnit := range r.adUnits {
-		str := fmt.Sprintf("mid=%s", adUnit.MasterTagId)
-		adUnitsParams = append(adUnitsParams, base64.URLEncoding.WithPadding(base64.NoPadding).EncodeToString([]byte(str)))
+	var uriObj *url.URL
+	uriObj, err := url.Parse(a.URI)
```
dbemiller: Two requests on this:

1. Can you put something like "check your config's `adapters.adform.endpoint` value" in the error message? That makes fixing things like this much easier.
2. Can you parse this and panic in the constructor? You'll have to make a copy of it inside this function before mutating it... but we're trying to make sure that PBS crashes immediately on bad config values--before it starts serving traffic.

PR #587, opened 2018-06-26
 Banner impression update to look for only top one

smithaammassamveettil:
Description:

Making a change in brightroll adaptor to take the top width and height for banner impressions. Moving them to banner.w and banner.h fields for oRTB2.3 support


dbemiller:
Tests are failing because this introduces a race condition.

The `imp.Banner` object is shared across different adapters. You'll need to copy it and mutate your copy to get rid of it.

smithaammassamveettil:
@dbemiller I tried changing the code to work on a copy of the request for updating in the adaptor, as you mentioned above. Still race condition exists. 
Looks like the issue is with the pointer type for impression.Banner.H and impression.Banner.W. They are of type *uint64
When I try to set the "impression.Banner.H" and "impression.Banner.W" value to the first value available in impression.Banner.Format[0].H and impression.Banner.Format[0].W, those are of type uint64, looks like the race condition issues arising from there.
I am not sure why the openRTB implementation in Prebid core code having impression.Banner.H and impression.Banner.W declared as *uint64. 
(I just learned Go for this project, so not an expert in Go, trying to understand). since our System is is on openRTB2.3, trying to change the request by setting Width and Height into Banner object instead of  them in Format array.

dbemiller:
@smithaammassamveettil see how other bidders handle this. For example: https://github.com/prebid/prebid-server/blob/master/adapters/appnexus/appnexus.go#L353

`reqCopy := *request` only makes a shallow copy of the request. Since the `Banner` object is the one which you're mutating, that's the one you need to copy.

smithaammassamveettil:
@dbemiller Thank you so much for letting me know how to fix the banner issue. Appreciate it!
I was working on that bug after committing initial race condition fix. Now updated the imp copy fix as mentioned and it is working now.

dbemiller:
sure thing. lgtm

Issue #555, opened 2018-06-11
Preprocess bid requests for valid media types


dbemiller:
A comment which @muncha put in his adapter code in #538 reminded me of this. It's a good idea, and I realized it wasn't logged anywhere.

In the `static/bidder-info/{bidder}.yaml` files, Bidders describe which types of media types & platforms they support. For example:

```
capabilities:
  app:
    mediaTypes:
      - banner
  site:
    mediaTypes:
      - banner
      - video
```

We should add these into the core auction logic, so that bidders never see requests for types & platforms they're not expecting. It would also centralize the error messages we return to pubs so that we can make them consistent across all the bidders.

dbemiller:
#651 was merged on August 13, and released on August 16.

I see no instances of `CAPABILITY` in our app logs... so it's safe to say that the middleware is working. It should be safe to delete those logic blocks now.

PR #594, opened 2018-06-28
Adding a decorator which enforces the {bidder}.yaml info files

dbemiller:
This contributes to #555.

Goal here is to make an object which wraps a Bidder and screens invalid bids before calling it.

In a future PR, this can be used to simplify adapter code. For example, in the APN adapter, we have:

```
	if imp.Audio != nil {
		return "", &adapters.BadInputError{
			Message: fmt.Sprintf("Appnexus doesn't support audio Imps. Ignoring Imp ID=%s", imp.ID),
		}
	}
```

Other adapters do similar. This offers a way to consolidate that logic into one place and standardize the error messages.

### DIFF
--- a/adapters/info.go
+++ b/adapters/info.go
@@ -36,17 +36,13 @@ func (i *InfoAwareBidder) MakeRequests(request *openrtb.BidRequest) ([]*RequestD
 	var allowedMediaTypes []openrtb_ext.BidType
 	if request.Site != nil {
 		if i.info.Capabilities.Site == nil {
-			return nil, []error{&BadInputError{
-				Message: "this bidder does not support site requests",
-			}}
+			return nil, []error{BadInput("this bidder does not support site requests")}
 		}
 		allowedMediaTypes = i.info.Capabilities.Site.MediaTypes
 	}
 	if request.App != nil {
 		if i.info.Capabilities.App == nil {
-			return nil, []error{&BadInputError{
-				Message: "this bidder does not support app requests",
-			}}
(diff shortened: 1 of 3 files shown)


adapters/info.go
```diff
@@ -1,13 +1,130 @@
...
+		request.Imp = filteredImps
+		errs = append(errs, newErrs...)
+	}
+	reqs, delegateErrs := i.Bidder.MakeRequests(request)
+	return reqs, append(errs, delegateErrs...)
+}
+
+// pruneImps trims invalid media types from each imp, and returns true if any of the
+// Imps have _no_ valid Media Types left.
+func (i *InfoAwareBidder) pruneImps(imps []openrtb.Imp, allowedTypes []openrtb_ext.BidType) (int, []error) {
```
hhhjort: Could this be made more efficient by returning a list of the indexes that need to be pruned, so the second stage can loop over that rather than loop over the entire list of imps and recheck each one?

dbemiller: I did think about that... but my hunch was that that would actually be less efficient. These lists are short, `malloc` is slow, and it's _probably_ not a huge deal because this code only runs if publishers send bad requests--which should hopefully only be done during testing and fixed before deploying to the internet.

I couldn't really tell you for sure which is more efficient without writing a benchmark for it, though. I can do that if you think it's needed.

hhhjort: The malloc is a good point, I had not considered that. It probably will not be all that rare as lazy publishers will probably activate all bidders on all Imps as long as there is a single imp that each bidder can bid on.

PR #595, opened 2018-06-28
Build go binary inside alpine container

larrabee:
In old Dockerfile, you build go binary outside of the docker. Because of this, the binary can be linked with glibc and it will crash when it starts inside the alpine container (alpine uses musl, that not compatible with glibc).
Solution: build the go binary in an alpine environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,19 @@
-FROM alpine
+FROM alpine:edge AS build
+WORKDIR /go/src/github.com/prebid/prebid-server/
+RUN apk add -U --no-cache go git fftw-dev musl-dev dep
+ENV GOPATH /go
+COPY ./ ./
+RUN dep ensure
+RUN go build .
+
+
+FROM alpine:edge AS release
 MAINTAINER Brian O'Kelley <bokelley@appnexus.com>
-ADD prebid-server prebid-server
+WORKDIR /usr/local/bin/
+COPY --from=build /go/src/github.com/prebid/prebid-server/prebid-server .
 COPY static static/
 COPY stored_requests/data stored_requests/data
+RUN apk add -U --no-cache ca-certificates
 EXPOSE 8000
-ENTRYPOINT ["/prebid-server"]
+ENTRYPOINT ["/usr/local/bin/prebid-server"]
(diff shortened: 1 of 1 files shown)


dbemiller:
Definitely like the idea of building inside the `Dockerfile`. We've been doing it externally, but it makes sense to have one for the project.

Any reason why you're including all that C stuff? As of [Go 1.5](https://golang.org/doc/go1.5#net), everything's generally simpler if you build with `CGO_ENABLED=0`

larrabee:
Hello.
1. I think Dockerfile should contain full build scenario.
2. I did not know that the PBS does not use CGOs (in the readme it does not). Should we update the build instruction, it does not work with the current docker file?
3. For further use. If you ever need to use the CGO in the project.
4. Build in the docker does not make the image heavier or harder.

dbemiller:
*nods*, gotcha. I think you're right.

We will merge this... but production started having some issues yesterday, and I don't want to merge a big build change like this until the dust settles.

We'll be on vacation next week, but should get this in soon after.

Dockerfile
```diff
@@ -1,8 +1,19 @@
-FROM alpine
+FROM alpine:edge AS build
+WORKDIR /go/src/github.com/prebid/prebid-server/
+RUN apk add -U --no-cache go git fftw-dev musl-dev dep
```
dbemiller: Not a big deal, but... why is `fftw-dev` in here?

We're not doing any fast fourier transforms... and the `docker build .` seems to work without it.

Dockerfile
```diff
@@ -1,8 +1,19 @@
...
+ENV GOPATH /go
+COPY ./ ./
+RUN dep ensure
+RUN go build .
+
+
+FROM alpine:edge AS release
```
cirla: `edge` is the development branch of alpine--I don't think we'd want to use this for release images.

larrabee:
1) @dbemiller Yep, you are right. I forget remove this when copy-paste from another project ;)
2) @cirla edge was chosen because 3.8 has not released and 3.6 dont have dep in official repos. 3.8 has been released and i change alpine  version from edge to 3.8.

PR #604, opened 2018-07-09
Handle legacy => openrtb multiformat imps better.

### LINKED ISSUE #603: Fix MakeOpenRTBGeneric multiformat
If `singleMediaTypeImp` is true, then `MakeOpenRTBGeneric` returns a broken OpenRTB request on multiformat imps. See https://github.com/prebid/prebid-server/blob/master/adapters/openrtb_util.go#L88

In OpenRTB, every `Imp` object must have a unique `id`. Otherwise it's impossible to interpret the Bids in the response. On a multiformat imp, this function will return two imps with the same ID.

This only affects adapters which were never properly upgraded to support `/openrtb2`: `conversant`, `audienceNetwork`, `index`, `lifestreet`, `pubmatic`, and `pulsepoint`.

dbemiller:
Fixes #603.

We'll have to release this careful and watch the bid responses when we do. It changes `/auction` behavior for all bidders, and `/openrtb2/auction` for bidders who haven't updated yet (see #603 for the list).

but... I have to imagine that the risk is worth it. In OpenRTB, you can't interpret bid responses unless each Imp had a unique ID. The code path which I deleted here was splitting a single legacy AdUnit into two Imps with the same ID... which should cause some nonsense to happen.

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -74,7 +74,7 @@ type appnexusImpExt struct {
 
 func (a *AppNexusAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder *pbs.PBSBidder) (pbs.PBSBidSlice, error) {
 	supportedMediaTypes := []pbs.MediaType{pbs.MEDIA_TYPE_BANNER, pbs.MEDIA_TYPE_VIDEO}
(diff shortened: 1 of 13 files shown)


dbemiller:
Should wait to merge this until after the next release... this is pretty big, and we've got quite a lot of changes already.

PR #605, opened 2018-07-09
Fixed beachfront imports

dbemiller:
I noticed this while working on #604.

`pkg/errors` is a third-party lib... which we're not using. It's a transient dependency... so PBS doesn't build unless you run `dep ensure` again to pick it up. But when you do, it changes your lockfile.

So... this reverts it to plain old `errors`.

### DIFF
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -2,13 +2,14 @@ package beachfront
 
 import (
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


dbemiller:
yeah... I realized it was a small/standalone change, so I made a separate PR for it.  Figured that one might be more contentious, but this should merge quickly (and before the next release)

PR #610, opened 2018-07-10
Fixes a corner case where UsersyncIfAmbiguous is ignored 

hhhjort:
if the consent string is supplied without the gdpr flag, we would always assume `gdpr: 1`. This fixes it too match the documented behavior.


endpoints/cookie_sync.go
```diff
@@ -88,6 +93,14 @@ func (deps *cookieSyncDeps) Endpoint(w http.ResponseWriter, r *http.Request, _ h
 		http.Error(w, "gdpr_consent is required if gdpr=1", http.StatusBadRequest)
 		return
 	}
+	// If GDPR is ambiguous, lets untangle it here.
+	if parsedReq.GDPR == nil {
+		if deps.gDPR.UsersyncIfAmbiguous {
+			parsedReq.GDPR = &zero
+		} else {
+			parsedReq.GDPR = &one
```
cirla: This may get us into a position where `*parsedReq.GDPR == 1` but `gdpr_consent == ""`, which throws an error in the above block. This is probably ok if `permissions.HostCookiesAllowed` (via `filterForGDPR`) handles this case without error, otherwise in the `UsersyncIfAmbiguous == false` case we could set `*parsedReq.GDPR == 0` and ``*parsedReq.Bidders == nil`.

hhhjort: Thats why I put this just below the above block. :)  Thought it best to avoid `nil` as either we would have to push the config option and the logic deeper into the code, or risk the original issue creeping back. We already have lots of checks `if consent == ""` so I think we are on solid ground for the case of `gdpr: 1` and no consent string, but it is a good point.

dbemiller: I think @hhhjort has the right idea here. Or at least: the block above returns a 400. I don't think that makes sense if gdpr/consent are undefined, because it's a perfectly valid request.

endpoints/cookie_sync.go
```diff
@@ -20,20 +20,25 @@ import (
...
+		hostCookie:      &cfg.HostCookie,
+		gDPR:            &cfg.GDPR,
 		syncPermissions: syncPermissions,
 		metrics:         metrics,
 		pbsAnalytics:    pbsAnalytics,
 	}
 	return deps.Endpoint
 }
 
+var one = 1
```
dbemiller: I know what you wrote is more efficient, but... I think you should just move these inside the `Endpoint()`.

It would be very easy for someone to write `parsedReq.GDPR = {thing}` and accidentally change the value of this global "constant". A bug like that would be nearly impossible to track down.

cirla:
The builds are getting terminated on Travis (I did a restart and same error again):

```
...
ok  	github.com/prebid/prebid-server/endpoints/openrtb2	1.023s [no tests to run]
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

what happens when you run `./validate.sh --nofmt --cov --race 10` locally?

dbemiller:
Yeah... I tracked this down. See #615.

dbemiller:
No substantive changes in those last two commits... just pulling so that this branch gets #619 and we can see if the CI passes.

PR #613, opened 2018-07-12
PBS crashes on startup if prometheus isn't configured

### LINKED ISSUE #614: Prometheus metrics configured, but a Prometheus metrics engine was not found
Probably the result of https://github.com/prebid/prebid-server/commit/c40c8373caa8fe153c7333f23dbff380aa0ebf27 

All my setting for Prometheus are default ones, the port is set to zero. For some reason server.go tries to create a new Prometheus server https://github.com/prebid/prebid-server/blob/master/server/server.go#L37 no matter what settings are listed in the config

dbemiller:
b/c it called `newPrometheusServer` without checking the config, and triggered:

```
	if proMetrics == nil {
		glog.Fatal("Prometheus metrics configured, but a Prometheus metrics engine was not found. Cannot set up a Prometheus listener.")
	}
```

Fixes #614

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -34,8 +34,6 @@ func Listen(cfg *config.Configuration, handler http.Handler, adminHandler http.H
 	mainServer := newMainServer(cfg, handler)
 	go shutdownAfterSignals(mainServer, stopMain, done)
 
-	prometheusServer := newPrometheusServer(cfg, metrics)
-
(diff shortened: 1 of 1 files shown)


dbemiller:
Travis is hanging because of something in the exchange... so I'm pretty certain these aren't causing the Travis errors. Gonna merge this because it's upsetting others.

PR #618, opened 2018-07-13
Close the database safely

dbemiller:
Tiny bugfix for an issue I found while debugging something else.

This has pretty minimal impact... if the app config doesn't include a database connection, then there's a nil dereference when trying to shut down the server.

Still... a little annoying.


dbemiller:
I tested this locally, commenting out the `TestRaceIntegration` test to verify that it works.

Issue #615, opened 2018-07-12
CI is failing due to timeouts


dbemiller:
Right now, our tests are behaving _extremely_ slow (even locally and on master). You can reproduce with:

```
go test ./exchange/ -timeout 10s
```

I traced this back to the `TestRaceIntegration` test, which runs:

```

noBidServer := func(w http.ResponseWriter, r *http.Request) {
  w.WriteHeader(204)
}
server := httptest.NewServer(http.HandlerFunc(noBidServer))
defer server.Close()
ex := NewExchange(server.Client(), ...)
```

This is... misleading. Although it creates a mock endpoint, it never actually configures the adapters to use the `server.URL` as their endpoint.  This gives it external dependencies.

Someone's servers must be slow today, which is causing the tests to hang. We should mock this out so that it doesn't depend on external servers anymore.

dbemiller:
Was fixed a while ago... I forgot to close this.

PR #619, opened 2018-07-13
Commented out the broken test.

dbemiller:
Because we run tests before release, and this test is broken because of external dependencies.

I can confirm that it's due to external dependencies because `./validate.sh` still hangs if I, `git checkout 0.35.3`, and that version was definitely working in the past.

### DIFF
--- a/exchange/exchange_test.go
+++ b/exchange/exchange_test.go
@@ -56,28 +56,30 @@ func TestNewExchange(t *testing.T) {
 //
 // The "known" file names right now are "banner.json" and "video.json". These files should hold params
 // which the Bidder would expect on banner or video Imps, respectively.
-func TestRaceIntegration(t *testing.T) {
-	noBidServer := func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


exchange/exchange_test.go
```diff
@@ -56,28 +56,30 @@ func TestNewExchange(t *testing.T) {
 //
 // The "known" file names right now are "banner.json" and "video.json". These files should hold params
 // which the Bidder would expect on banner or video Imps, respectively.
-func TestRaceIntegration(t *testing.T) {
```
cirla: Could alternatively do a `t.SkipNow()` with a `//TODO` comment

dbemiller: much better idea ^^.

I don't think config makes sense here... this isn't something which varies from one deployment to another. `t.SkipNow` is much cleaner though

dbemiller:
yeah. A real fix will come with #615... but I got a publisher request to release, so I want to make it happen ASAP.

Will wait for the CI to pass before merging, just to make certain.

PR #622, opened 2018-07-16
Removed most external test dependencies

### LINKED ISSUE #621: App config should be consistent with bidder params
Noted by @dmitriyshashkin in slack:

> One small inconsistency. config.go has a default param for `adapters.index.usersync_url`, yet syncer.go looks for `adapters.indexexchange.usersync_url`

This does need to be fixed... but more generally, we should make sure all the bidders are using their bidder codes in the app config, to reduce the chance that this happens in the future.

In the meantime: PBS hosts should make sure they set the app config for `adapters.indexExchange`, because the defaults in the code won't work.

dbemiller:
This fixes #621

This contains a breaking change to the app config:

- `adapters.facebook.*` params are renamed to `adapters.audiencenetwork.*`

It also fixes the `adapters.indexExchange.*` defaults. As #621 mentions, we're calling `SetDefault` using `index`, but the actual code is searching for `indexexchange`.

------------------------------------------------------

This also fixes _most_ of #615. `Somoaudience` servers were the guilty party causing the CI to hang... so this fixes that, and stops skipping the test.

I updated most of the bidders to use configurable endpoints so that we could make them call a mock server in `TestRaceIntegration`. The only exception is `Beachfront`, which uses _two_ endpoints . (one for video and one for banner).

The config structure needs to support that before we can mock their endpoints... but I think that's big enough that it deserves its own PR.


dbemiller:
Thank you... I totally forgot about that.

PR #628, opened 2018-07-19
Recover from panics inside goroutines

dbemiller:
Today, bad Bidder code can cause the server to crash (e.g. nil-pointer dereferences).

The root cause is `panics`, which crash the process if they're not caught. Although Go's HTTP Server catches panics from the handlers, there's nothing to recover inside the goroutines we create ourselves.

This adds some safeguards so they get logged as errors rather than crashing the process.

I do plan to add this to the metrics... but was undecided on whether to do it in this PR or another one. Let me know either way.

### DIFF
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"runtime/debug"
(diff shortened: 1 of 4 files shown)


cirla:
Metrics could be another PR to keep these digestible

PR #629, opened 2018-07-19
Fix for Issue 574

rachelrj:
Most of the updates were on our side to allow for the impressions to be recorded. Also needed this update to render the creative correctly. I think this should be it.


adapters/sovrn/sovrn.go
```diff
@@ -250,6 +250,7 @@ func (s *SovrnAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalReq
 	for _, sb := range bidResp.SeatBid {
 		for i := 0; i < len(sb.Bid); i++ {
 			bid := sb.Bid[i]
+			bid.AdM, _ = url.QueryUnescape(bid.AdM)
```
dbemiller: No error handling here?

Should this be an `adapters.BadServerResponeError`?

rachelrj: Good point. But.. I'd rather just drop that specific bid than error on the entire response.

rachelrj: That work?

dbemiller: It's ok with me.

May I ask why you wouldn't want to use the `BadServerResponseError`? I don't see any drawbacks... but there are definitely a few benefits.

To be clear: your bids will still count if you return bids _and_ errors. These just help debugging & diagnosing errors..

dbemiller: Merged... but also feel free to submit another PR adding the error if it changes your mind.

PR #630, opened 2018-07-20
Moves adapter error types into its own module.

hhhjort:
This is being done in anticipation of being able to define error codes and return them as part of the PBS return. Should be helpful for some monitoring and automatic debugging applications.

### DIFF
--- a/errortypes/errortypes.go
+++ b/errortypes/errortypes.go
@@ -1,0 +1,72 @@
+package errortypes
+
+// These define the error codes for all the errors enumerated in this package
+// NoErrorCode is to reserve 0 for non error states.
+const (
+	NoErrorCode = iota
+	TimeoutCode
+	BadInputCode
+	BadServerResponseCode
+)
+
+// We should use this code for any Error interface that is not in this package
+const UnknownErrorCode = 999
+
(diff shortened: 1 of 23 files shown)


errortypes/errortypes.go
```diff
@@ -0,0 +1,72 @@
...
+	BadInputCode
+	BadServerResponseCode
+)
+
+// We should use this code for any Error interface that is not in this package
+const UnknownErrorCode = 999
+
+// PBSError provides an interface to use if we want to deal with any error type created in this package.
+type PBSError interface {
+	Error() string
```
dbemiller: Any reason to declare the `Error()` method on here?

I'm not 100% sure how you plan to use this... but it seems extraneous to me. You'll already know the type is `erro`, which tells you that this function exists.

Not something you have to deal with right away... but in your future PR, bear in mind that you might just be able to delete it.

hhhjort: That is a fair point. There would be something fishy going on if something should fall into this interface does not implement `Error()`, but yes would not be strictly needed for the intended application.

dbemiller: Alright. We'll keep an eye on it in your next review. For now let's merge this so you don't end up with nasty conflicts.

Issue #473, opened 2018-04-20
Return bidder statuses in OpenRTB


dbemiller:
# Motivation

Header Bidding is often a hybrid auction which takes place on the Client (Prebid.js) and Server (PBS). Since Prebid.js adds new demand to the auction, it's impossible for Prebid Server to know who the true winner is.

Because "the winner" is a key part of analytics, Prebid Server must make other analytics data available in the response as well.

# Proposal

Since OpenRTB requires at least one `bid` in a `response.seatbid`, and we still want to return this data even if the bidder makes no bids, we will need to put this in `response.ext`. 

Since we've already defined `ext.responsetimemillis.{bidder}`, I think we should stick with that convention.

```
{
  "timeouts": {
    "appnexus": true,
    "index": false
  }
}
```

For errors, I'd hope to make a breaking change in the existing `response.ext.errors` API. If this causes anyone problems, please speak up.

```
{
  "errors": {
    "appnexus": [
      {
        "code": 1,
        "message": "A hybrid Banner/Audio Imp was offered, but Appnexus doesn't support Audio."
      }
    ]
  }
}
```

This would technically be a breaking change, because today the response would be:

```
{
  "errors": {
    "appnexus": "A hybrid Banner/Audio Imp was offered, but Appnexus doesn't support Audio."
  }
}
```

...but my hunch is that the project is small enough that nobody would mind yet. @anwzhang, @jaiminpanchal27, @mkendall07, @bretg as the primary consumers, please tell me if that's not ok with you.

The intent here is that the `response.ext.errors.{bidder}.code` would use a rigorous, well-defined list of error types, while `response.ext.errors.{bidder}.message` is a human-readable explanation of what went wrong. Changes to `code` _would_ constitute a breaking change to the PBS API, while changes to the `message` text would _not_.

dbemiller:
This was fixed in #631.

PR #631, opened 2018-07-23
Adds error code output to openrtb auction debugs.

hhhjort:
This expands the reporting of errors in debug requests to include error codes. This will allow for some automation in analyzing debug reports, as scripts can focus on the error codes rather than trying to pattern match against error messages.

### DIFF
--- a/docs/endpoints/openrtb2/auction.md
+++ b/docs/endpoints/openrtb2/auction.md
@@ -232,16 +232,14 @@ It will become impossible to fetch bids from Appnexus within that Request.
 `response.ext.responsetimemillis.{bidderName}` tells how long each bidder took to respond.
 These can help quantify the performance impact of "the slowest bidder."
 
+#### Bidder Errors
+
 `response.ext.errors.{bidderName}` contains messages which describe why a request may be "suboptimal".
 For example, suppose a `banner` and a `video` impression are offered to a bidder
 which only supports `banner`.
 
 In cases like these, the bidder can ignore the `video` impression and bid on the `banner` one.
 However, the publisher can improve performance by only offering impressions which the bidder supports.
-
-#### Bidder Errors
-
-`response.ext.errors` contains any errors reported by the bidders as they processed the request, broken out by bidder.
 
 For example, a request may return this in `response.ext`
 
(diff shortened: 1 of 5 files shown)


exchange/exchange.go
```diff
@@ -232,7 +232,7 @@ func errorsToMetric(errs []error) map[pbsmetrics.AdapterError]struct{} {
 	ret := make(map[pbsmetrics.AdapterError]struct{}, len(errs))
 	var s struct{}
 	for _, err := range errs {
-		if err == context.DeadlineExceeded {
+		if _, ok := err.(*errortypes.Timeout); ok {
```
dbemiller: This should be using the `DecodeError` method now.

hhhjort: What do you mean?

dbemiller:
Probably should delete the existing docs, with this new section & update: https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/auction.md#bidder-response-times

docs/endpoints/openrtb2/auction.md
```diff
@@ -232,13 +232,45 @@ It will become impossible to fetch bids from Appnexus within that Request.
...
+```
+
+The codes currently defined are:
+
+```
+0	NoErrorCode
+1	TimeoutCode
+2	BadInputCode
+3	BadServerResponseCode
+999 UnknownErrorCode
```
dbemiller: This spacing looks a little weird... is it mixing tabs & spaces?

hhhjort: Hmmm, could be ... did a little cutting and pasting to put that together, it is possbile. It looked ok in my editor, but tabs can certainly vary between editors.

dbemiller: Markdowns are tricky... I tend to just look them up on my branch in github: https://github.com/prebid/prebid-server/blob/error-codes/docs/endpoints/openrtb2/auction.md#bidder-errors

There are visual studio code plugins which you can use to see the rendered version too though.

dbemiller:
This fixes #473

PR #636, opened 2018-08-01
Creat a Fail-To_Bid error

hhhjort:
Adapters should either generate one or more http requests to their bidder to get bids, or return errors indicating why they could not bid on the supplied bid request (or both). However this is not enforced. This PR adds a failedToBid error to catch this case, and signal that an adapter is misbehaving in this manner. This should hopefully never happen, but now we will be able to catch and debug if it ever does.

### DIFF
--- a/pbsmetrics/metrics.go
+++ b/pbsmetrics/metrics.go
@@ -146,6 +146,7 @@ const (
 	AdapterErrorBadInput          AdapterError = "badinput"
 	AdapterErrorBadServerResponse AdapterError = "badserverresponse"
 	AdapterErrorTimeout           AdapterError = "timeout"
+	AdapterErrorFailedToBid       AdapterError = "failedtobid"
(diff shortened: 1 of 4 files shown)


pbsmetrics/metrics.go
```diff
@@ -146,6 +146,7 @@ const (
 	AdapterErrorBadInput          AdapterError = "badinput"
 	AdapterErrorBadServerResponse AdapterError = "badserverresponse"
 	AdapterErrorTimeout           AdapterError = "timeout"
+	AdapterErrorFailedToBid       AdapterError = "failedtobid"
```
dbemiller: Name seems a bit confusing to me... if you were unfamiliar with the project, how do you think this would compare with `nobid`?

Maybe `failedtorequest` would be more accurate?

hhhjort: Yes, that is what I use elsewhere, Not sure how I got off track here. Thanks

PR #637, opened 2018-08-02
Use jsonparser fork

dbemiller:
This switches our project to a `jsonparser` fork which includes https://github.com/buger/jsonparser/pull/137, because that project's maintainer hasn't responded in a while.

I also tried replacing `jsonparser` with [gjson](https://github.com/tidwall/gjson)... but our benchmark showed an extra millisecond of overhead, so... *shrugs*.

I believe this should fix some panics we've been seeing in the app logs. Once it merges we can switch back to the main project.

It upgrades lots of other packages too. Honestly this happened by accident when I deleted `Gopkg.lock` and ran `dep ensure`. If you think that's too risky, I can go back and update _just_ this package... but IMO we're due for some dependency upgrades anyway.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -45,6 +45,15 @@
   branch = "master"
   name = "github.com/evanphx/json-patch"
 
+# The main project has an open PR with a bugfix that affects us.
+# This jsonparser fork includes that PR.  Once merged, this can
+# return to the main project.
+# See: https://github.com/buger/jsonparser/pull/137
+[[constraint]]
+  branch = "master"
+  name = "github.com/buger/jsonparser"
+  source = "https://github.com/dbemiller/jsonparser.git"
+
 [[constraint]]
   branch = "master"
   name = "github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -45,6 +45,15 @@
...
   name = "github.com/evanphx/json-patch"
 
+# The main project has an open PR with a bugfix that affects us.
+# This jsonparser fork includes that PR.  Once merged, this can
+# return to the main project.
+# See: https://github.com/buger/jsonparser/pull/137
+[[constraint]]
+  branch = "master"
+  name = "github.com/buger/jsonparser"
+  source = "https://github.com/dbemiller/jsonparser.git"
```
cirla: Do we want this fork to live in the prebid github org? Probably not worth it if we're intending to switch back to upstream after a single PR is merged.

dbemiller: yeah... I considered that too. I chose not to put it in the prebid org just because I didn't want to pollute the namespace. Hopefully this is a short-lived fork.

We can always move it if it turns out to be a long-lived fork and we care that much. At this point, we don't even know for certain that this will get rid of the panics, or what other bugs it might cause... so that seemed a little premature.

PR #640, opened 2018-08-02
[sovrn] accept camel- or lower-case tagid's

aprakash-sovrn:


adapters/sovrn/sovrn.go
```diff
@@ -287,7 +291,11 @@ func preprocess(imp *openrtb.Imp) (string, error) {
 		}
 	}
 
-	imp.TagID = sovrnExt.TagId
+	if len(sovrnExt.TagId) > 0 {
```
dbemiller: It's ok as is, but... if `TagId` is deprecated, wouldn't you want `tagid` to have preference in case both exist?

aprakash-sovrn: good call, let me flip those

PR #641, opened 2018-08-02
Add adapter errors to the AMP response

### LINKED ISSUE #616: AMP endpoint hides errors even in debug mode
So currently the response for amp endpoint includes 1) targeting 2) http calls 3) resolved request. Errors received from adapters are passed to the `AmpAuction` function but ignored. 

Which makes debugging extremely difficult, there is no way to understand what exactly went wrong. The only hint is the absence of the query to one of the bidders in the HTTP calls list. The two possible workarounds for a publisher I see right now are 1) compile your own PBS instance and add some glog.Info calls 2) send request to openrtb2/auction endpoint with the mention of the stored request used for amp. Both solutions are rather cumbersome and require good technical knowledge.

I would suggest adding a new section to the amp endpoint's response that would contain all the errors that happened during the auction.

hhhjort:
AMP should ignore the field, but allows for debugging without trying to turn the request into an openrtb request. Fixes #616

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -167,7 +167,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	var extResponse openrtb_ext.ExtBidResponse
 	eRErr := json.Unmarshal(response.Ext, &extResponse)
 	if eRErr != nil {
-		ao.Errors = append(ao.Errors, fmt.Errorf("AMP response: failed to unpack OpenRTB response.ext: %v", eRErr))
+		ao.Errors = append(ao.Errors, fmt.Errorf("AMP response: failed to unpack OpenRTB response.ext, debug info cannot be forwarded: %v", eRErr))
 	}
 
 	// Now JSONify the targets for the AMP response.
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -165,19 +165,15 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	}
 	// Extract any errors
 	var extResponse openrtb_ext.ExtBidResponse
-	var respErrors map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderError
 	eRErr := json.Unmarshal(response.Ext, &extResponse)
 	if eRErr != nil {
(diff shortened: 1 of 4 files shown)


endpoints/openrtb2/amp_auction.go
```diff
@@ -162,17 +163,28 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 			}
 		}
 	}
+	// Extract any errors
+	var extResponse openrtb_ext.ExtBidResponse
+	var respErrors map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderError
+	eRErr := json.Unmarshal(response.Ext, &extResponse)
+	if eRErr != nil {
+		glog.Errorf("AMP response: failed to unpack OpenRTB response.ext: %v", eRErr)
```
dbemiller: I don't think we need to `glog` this... but I do think we should append it to the response errors under the `prebid` key so the publisher knows an error occurred, even if they don't know exactly what it is.

endpoints/openrtb2/amp_auction.go
```diff
@@ -162,17 +163,28 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 			}
 		}
 	}
+	// Extract any errors
+	var extResponse openrtb_ext.ExtBidResponse
+	var respErrors map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderError
```
dbemiller: you can simplify this by removing `respErrors`. Just replace `Errors:    respErrors,` with `Errors:    extResponse.Errors`,

endpoints/openrtb2/amp_auction.go
```diff
@@ -162,17 +163,28 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
+		Errors:    respErrors,
 	}
 
 	ao.AmpTargetingValues = targets
 
 	// add debug information if requested
-	if req.Test == 1 {
-		var extResponse openrtb_ext.ExtBidResponse
-		if err := json.Unmarshal(response.Ext, &extResponse); err == nil && extResponse.Debug != nil {
+	if req.Test == 1 && eRErr == nil {
```
dbemiller: why?

hhhjort: If we had an error above in unmarshalling the ext, we have already logged it and then this message could be deceptive (its not that errors was missing, but that the whole ext was damaged). So test for both if debug was set, no errors unpacking, and we have debug info.

PR #643, opened 2018-08-03
Cache vast xml

### LINKED ISSUE #324: Support video cache
**Behavior:**

The `cacheMarkup` flag causes PrebidServer to cache the bids and remove the creative body from the response to the client.

The PrebidServerBidAdapter places `cacheMarkup` at the global s2sConfig level.

**Problem:**

When the same web page contains both video (cached) and banner (uncached), the `s2sConfig.cacheMarkup` flag has to be changed in between requestBids() calls.

Proposed solution:

Define `cacheMarkup` per media type.

This allows for more control over when Prebid Server caches and responds with `hb_cache_url` and `hb_cache_id`. The client should to be able to tell the server what it wants to do for responses of different media types.

The proposed mechanism is to add a prebidServer-options extension to the bidRequest to allow the client to define what should happen for each mediaType. E.g. 
```
bidRequest: {
  ext: {
     prebidServer: {
         options: {
             bidCaching: {
                 banner: {
                      cacheMarkup: false
                 },
                 video: {
                     cacheMarkup: true
                     // could be more options in the future, e.g.  ttl: 300 and cacheScope: [all|markup]
                 }
             }
         }
     }
   }
}
```

When `cacheMarkup` is true for the given bid response based on mediaType, Prebid Server caches and responds with `hb_cache_url` and `hb_cache_id`.

There could also be a default cacheMarkup value for each mediaType:
- banner: false
- video: true
- native: false

dbemiller:
Fixes #324.

This does not implement `ttlseconds`, which was described in the ticket as optional. To support that, we'll need to add it to the whole stack (down through Prebid Cache)... so it would be another few PRs.


exchange/exchange.go
```diff
@@ -94,6 +95,10 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 			return nil, fmt.Errorf("Error decoding Request.ext : %s", err.Error())
 		}
 		bidAdjustmentFactors = requestExt.Prebid.BidAdjustmentFactors
+		if requestExt.Prebid.Cache != nil {
+			shouldCacheBids = requestExt.Prebid.Cache.Bids != nil
+			shouldCacheVAST = requestExt.Prebid.Cache.VastXML != nil
+		}
 		shouldCacheBids = requestExt.Prebid.Cache != nil && requestExt.Prebid.Cache.Bids != nil
```
hhhjort: Isn't this line redundant with the above?

dbemiller: yes, thank you. will remove

PR #646, opened 2018-08-06
Required some size to exist on banner requests

dbemiller:
In #644, I found this test which wasn't testing what it thought it was.

It intended to make sure we reject sizeless banners... but it was failing for other reasons.

Unfortunately, this is now a breaking change :(. Still probably better late than never. We can measure its true impact whenever we release it.

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -355,6 +355,11 @@ func validateBanner(banner *openrtb.Banner, impIndex int) error {
 		return fmt.Errorf("request.imp[%d].banner uses unsupported property: \"hmax\". Use the \"format\" array instead.", impIndex)
 	}
 
+	hasRootSize := banner.H != nil && banner.W != nil
+	if !hasRootSize && len(banner.Format) == 0 {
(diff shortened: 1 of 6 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -355,6 +355,11 @@ func validateBanner(banner *openrtb.Banner, impIndex int) error {
 		return fmt.Errorf("request.imp[%d].banner uses unsupported property: \"hmax\". Use the \"format\" array instead.", impIndex)
 	}
 
+	hasRootSize := banner.H != nil && banner.W != nil
```
hhhjort: shouldn't this be `banner.H != nil || banner.W != nil`? We need both for a size.

dbemiller: this is defining `hasRootSize`, in the positive. Like you said, we need both to have one.

hhhjort: Doh, my bad.

dbemiller: all good. It did point me to the fact that you could still set these to `0`, which isn't helpful either.

I added some logic for that & a few more tests.

PR #647, opened 2018-08-06
Sets a config file name so we can redirect to non-existent file for t…

### LINKED ISSUE #608: Config tests should ignore filesystem
When developing locally, it's convenient to put a `pbs.yml` in the project root with your config values.

Currently, there's a "unit" test which fails based on the contents in this file:

```
--- FAIL: TestViperInit (0.00s)
	pbs_light_test.go:685: Viper error: external_url expected to be http://localhost:8000, found {other-value}
FAIL
FAIL	github.com/prebid/prebid-server	0.069s
```

We should fix these tests so that they don't check the files on the filesystem.

hhhjort:
…ests. This fixes #608

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -237,7 +237,7 @@ func (cfg *Configuration) GetCachedAssetURL(uuid string) string {
 // Set the default config values for the viper object we are using.
 func SetupViper(v *viper.Viper, filename string) {
 	if filename != "" {
-		v.SetConfigName(name)
+		v.SetConfigName(filename)
 		v.AddConfigPath(".")
 		v.AddConfigPath("/etc/config")
 	}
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -235,11 +235,12 @@ func (cfg *Configuration) GetCachedAssetURL(uuid string) string {
...
 
 // Set the default config values for the viper object we are using.
-func SetupViper(v *viper.Viper) {
-	v.SetConfigName("pbs")
-	v.AddConfigPath(".")
-	v.AddConfigPath("/etc/config")
-
+func SetupViper(v *viper.Viper, filename string) {
+	if filename != "" {
+		v.SetConfigName(name)
```
cirla: Looks like build is failing here, should be `filename`

Issue #648, opened 2018-08-06
CORS error on version 0.40.0 


mercuryyy:
Failed to load http://prebid.adnxs.com/pbs/v1/openrtb2/auction: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. Origin 'http://www.mydomain.com' is therefore not allowed access. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.

Started at 1PM UST Time Aug 6

Please Advise we have 7 bidders on s2s got hit with a big revenue drop until we saw the error. 

@dbemiller

dbemiller:
Looks like this was caused by the `cors` library update: https://github.com/prebid/prebid-server/compare/0.39.1...0.40.0#diff-bd247e83efc3c45ae9e8c47233249f18R225

We rolled back our deployed version, and added some integration tests which run from a browser and make sure we get alerted sooner if it happens again.

For other PBS hosts: don't use version 0.40.0. We'll fix this and release a 0.40.1 patch soon.

PR #650, opened 2018-08-08
Cors regression test

dbemiller:
Preventing #648 more safely.

### DIFF
--- a/pbs_light_test.go
+++ b/pbs_light_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/mxmCherry/openrtb"
 	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/prebid/prebid-server/cache/dummycache"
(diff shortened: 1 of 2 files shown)


pbs_light_test.go
```diff
@@ -688,6 +689,23 @@ func TestPanicRecovery(t *testing.T) {
...
+	req.Header.Set("Access-Control-Request-Method", "POST")
+	req.Header.Set("Access-Control-Request-Headers", "origin")
+	req.Header.Set("Origin", "https://publisher-domain.com")
+
+	if !assert.NoError(t, err) {
+		return
+	}
+	cors.ServeHTTP(rr, req)
+	assert.NotEqual(t, "*", rr.Header().Get("Access-Control-Allow-Origin"))
+}
```
hhhjort: Do we want to test if the `Access-Control-Allow-Origin` matches the `Origin` we set in the request?

dbemiller: Yes... very good idea. Done

PR #651, opened 2018-08-09
Bidder capabilities

hhhjort:
This enforces the bidder capabilities listed in bidder-info.yaml. Server logging has been added to the redundant checks within the adapter code so we can verify they are not needed and remove them with a later PR. Note that this does not cover legacy adapters that have been wrapped with the legacy->openRTB adapter.

### DIFF
--- a/adapters/info.go
+++ b/adapters/info.go
@@ -23,28 +23,28 @@ import (
 func EnforceBidderInfo(bidder Bidder, info BidderInfo) Bidder {
 	return &InfoAwareBidder{
 		Bidder: bidder,
-		info:   info,
+		info:   parseBidderInfo(info),
 	}
 }
 
 type InfoAwareBidder struct {
 	Bidder
-	info BidderInfo
+	info parsedBidderInfo
 }
 
 func (i *InfoAwareBidder) MakeRequests(request *openrtb.BidRequest) ([]*RequestData, []error) {
-	var allowedMediaTypes []openrtb_ext.BidType
+	var allowedMediaTypes parsedSupports
 	if request.Site != nil {
-		if i.info.Capabilities.Site == nil {
+		if i.info.site.enabled == false {
 			return nil, []error{BadInput("this bidder does not support site requests")}
 		}
-		allowedMediaTypes = i.info.Capabilities.Site.MediaTypes
+		allowedMediaTypes = i.info.site
 	}
 	if request.App != nil {
-		if i.info.Capabilities.App == nil {
+		if i.info.app.enabled == false {
 			return nil, []error{BadInput("this bidder does not support app requests")}
 		}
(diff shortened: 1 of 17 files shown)


adapters/info.go
```diff
@@ -190,3 +189,30 @@ func containsMediaType(haystack []openrtb_ext.BidType, needle openrtb_ext.BidTyp
 	}
 	return false
 }
+
+// Structs to handle parsed bidder info, so we aren't reparsing every request
```
cirla: Good catch!

adapters/info.go
```diff
@@ -23,28 +23,28 @@ import (
...
-	info BidderInfo
+	info parsedBidderInfo
 }
 
 func (i *InfoAwareBidder) MakeRequests(request *openrtb.BidRequest) ([]*RequestData, []error) {
-	var allowedMediaTypes []openrtb_ext.BidType
+	var allowedMediaTypes parsedSupports
 	if request.Site != nil {
-		if i.info.Capabilities.Site == nil {
+		if i.info.site.enabled == false {
```
dbemiller: fairly silly nitpick, but... `if !i.info.site.enabled`

config/config.go
```diff
@@ -235,11 +235,12 @@ func (cfg *Configuration) GetCachedAssetURL(uuid string) string {
 }
 
 // Set the default config values for the viper object we are using.
-func SetupViper(v *viper.Viper) {
-	v.SetConfigName("pbs")
-	v.AddConfigPath(".")
-	v.AddConfigPath("/etc/config")
-
+func SetupViper(v *viper.Viper, filename string) {
```
dbemiller: Little puzzled... what happened here?

Does this go away if you merge master into the branch? I'd've sworn we already merged these changes elsewhere.

pbs_light.go
```diff
@@ -693,8 +693,10 @@ func serve(revision string, cfg *config.Configuration) error {
 		glog.Fatalf("Failed to create the bidder params validator. %v", err)
 	}
 
+	bidderInfos := adapters.ParseBidderInfos("./static/bidder-info", openrtb_ext.BidderList())
```
dbemiller: This _should_ be safe, but... please do build & run this once in a docker container locally, just to make sure the files & paths still work.

config/config.go
hhhjort: Ok, I am really puzzled as config/config.go was not part of the commit when I created it.

pbs_light.go
hhhjort: This is the exact same line as from below, just moved up so `exchange.NewExchange` has access to it. Is there something about moving it that might break something?

dbemiller: ah, of course... no, you're fine then. Ignore me

dbemiller:
Contributes to #555

PR #655, opened 2018-08-17
AdkernelAdn adapter

ckbo3hrk:


adapters/adkernelAdn/adkernelAdn.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+		resImps = append(resImps, imp)
+		resImpExts = append(resImpExts, *impExt)
+
+	}
+	return &resImps, &resImpExts, &errors
+}
+
+func validateImpression(imp *openrtb.Imp, impExt *openrtb_ext.ExtImpAdkernelAdn) error {
+	if imp.Banner == nil && imp.Video == nil {
```
hhhjort: This should now be handled by PR #651 to standardize these basic capability checks, and make sure the advertised capabilities match what the code actually checks for. You can add `glog.Warning("Adkernl CAPABILITY VIOLATION: only banner and video imps supported")` at this time so we can verify that your adapter is never sent and impression that does not allow either of these types. A later PR will pull all these checks from the adapter code to save the adapters a few cycles.

adapters/adkernelAdn/adkernelAdn.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	return nil
+}
+
+//Group impressions by AdKernel-specific parameters `pubId` & `host`
+func dispatchImpressions(imps *[]openrtb.Imp, impsExt *[]openrtb_ext.ExtImpAdkernelAdn) map[openrtb_ext.ExtImpAdkernelAdn][]openrtb.Imp {
+	res := make(map[openrtb_ext.ExtImpAdkernelAdn][]openrtb.Imp)
+	for idx := range *imps {
+		imp := (*imps)[idx]
+		compatImpression(&imp)
```
hhhjort: You are ignoring any errors returned by `compatImpression()` currently?

adapters/adkernelAdn/adkernelAdn.go
```diff
@@ -0,0 +1,242 @@
...
+		Body:    reqJSON,
+		Headers: headers}, nil
+}
+
+func createBidRequest(prebidBidRequest *openrtb.BidRequest, params *openrtb_ext.ExtImpAdkernelAdn, imps *[]openrtb.Imp) *openrtb.BidRequest {
+	bidRequest := *prebidBidRequest
+	bidRequest.Imp = *imps
+	for idx := range bidRequest.Imp {
+		imp := &bidRequest.Imp[idx]
+		imp.TagID = imp.ID
```
hhhjort: This should result in `imp` being a copy of the imp in `bidRequest`, and then you set `TagID` in this copy. I don't see you using `imp` again, so I believe this won't do what you expect. You could try `bidRequest.Imp[idx].TagID = bidRequest.Imp[idx].ID`.

adapters/adkernelAdn/adkernelAdn.go
ckbo3hrk: Looks like my mistake. Thank you

adapters/adkernelAdn/adkernelAdn.go
ckbo3hrk: @hhhjort My intention there was to write compact code but it looks like pointers are confusing. Regarding `bidRequest.Imp[idx].TagID = bidRequest.Imp[idx].ID` I have a noob question - does golang optimize double `bidRequest.Imp[idx]` op?

hhhjort: That is a good question, I don't know for sure myself. Since one of the design goals of golang is to be a relatively simple language with just the minimum features needed, that allows for easier optimization in the compiler. Then again it is a relatively new language, so it is likely they haven't optimized every possible case.

adapters/adkernelAdn/adkernelAdn.go
ckbo3hrk: Addressed in 076145eba887ea2ac2eab38ffaef77e9a9ef631c

adapters/adkernelAdn/adkernelAdn.go
ckbo3hrk: Addressed in 076145eba887ea2ac2eab38ffaef77e9a9ef631c

adapters/adkernelAdn/adkernelAdn.go
```diff
@@ -0,0 +1,254 @@
...
+			errs = append(errs, err)
+		} else {
+			result = append(result, bidRequest)
+		}
+	}
+	return result, errs
+}
+
+// getImpressionsInfo checks each impression for validity and returns impressions copy with corresponding exts
+func getImpressionsInfo(imps *[]openrtb.Imp) (*[]openrtb.Imp, *[]openrtb_ext.ExtImpAdkernelAdn, *[]error) {
```
dbemiller: The code throughout can be simplified if you drop the `*`s in the method signatures. A pointer to a slice doesn't make much sense as a type, because slices [are already pointers](https://blog.golang.org/go-slices-usage-and-internals#TOC_4.).

adapters/adkernelAdn/adkernelAdn.go
```diff
@@ -0,0 +1,254 @@
...
+		imp.TagID = imp.ID
+	}
+	if bidRequest.Site != nil {
+		bidRequest.Site.Publisher = nil
+		bidRequest.Site.Domain = ""
+	}
+	if bidRequest.App != nil {
+		bidRequest.App.Publisher = nil
+	}
+	//nullify as prebid-server passes base64 adapter->uid map for all adapters
```
dbemiller: er... how sure are you that you want to do this?

We didn't document t his feature well, so I can see how you got this impression... but actually the base64 adapter->uid map is just a fallback. Companies hosting PBS _should_ have their app config set so that PBS reads the user's ID from their domain's cookie.

adapters/adkernelAdn/info.yaml
```diff
@@ -0,0 +1,2 @@
+maintainer:
```
dbemiller: You can delete this file.

`static/bidder-info/adkernelAdn.yaml` is important. `adapters/adkernelAdn/info.yaml` isn't used for anything... so it's misleading to have.

Did you see some bad docs which made you think you needed this? If so let me know... I should update them.

adapters/adkernelAdn/adkernelAdn.go
ckbo3hrk: Fixed in e341d28059d36a1fb21dc9a01d4479cac76c2059

adapters/adkernelAdn/adkernelAdn.go
ckbo3hrk: My mistake - I tested local PBS instance without properly configure `host_cookie.*`. With updated config there is no need to nullify this field.

PR #658, opened 2018-08-22
GDPR enhancements

hhhjort:
This will remove some Personally Identifying information from bid requests sent to bidder adapters if GDPR consent is missing.

### DIFF
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -17,6 +17,7 @@ import (
 	"github.com/prebid/prebid-server/adapters"
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/gdpr"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
 	"github.com/prebid/prebid-server/prebid_cache_client"
@@ -35,10 +36,12 @@ type IdFetcher interface {
 }
 
 type exchange struct {
-	adapterMap map[openrtb_ext.BidderName]adaptedBidder
-	me         pbsmetrics.MetricsEngine
-	cache      prebid_cache_client.Client
-	cacheTime  time.Duration
+	adapterMap          map[openrtb_ext.BidderName]adaptedBidder
+	me                  pbsmetrics.MetricsEngine
+	cache               prebid_cache_client.Client
+	cacheTime           time.Duration
+	gDPR                gdpr.Permissions
+	UsersyncIfAmbiguous bool
 }
 
 // Container to pass out response ext data from the GetAllBids goroutines back into the main thread
@@ -53,13 +56,15 @@ type bidResponseWrapper struct {
 	bidder       openrtb_ext.BidderName
 }
 
-func NewExchange(client *http.Client, cache prebid_cache_client.Client, cfg *config.Configuration, metricsEngine pbsmetrics.MetricsEngine, infos adapters.BidderInfos) Exchange {
+func NewExchange(client *http.Client, cache prebid_cache_client.Client, cfg *config.Configuration, metricsEngine pbsmetrics.MetricsEngine, infos adapters.BidderInfos, gDPR gdpr.Permissions) Exchange {
 	e := new(exchange)
 
 	e.adapterMap = newAdapterMap(client, cfg, infos)
 	e.cache = cache
 	e.cacheTime = time.Duration(cfg.CacheURL.ExpectedTimeMillis) * time.Millisecond
 	e.me = metricsEngine
+	e.gDPR = gDPR
+	e.UsersyncIfAmbiguous = cfg.GDPR.UsersyncIfAmbiguous
 	return e
 }
 
@@ -77,6 +82,18 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
 	cleanRequests, aliases, errs := cleanOpenRTBRequests(bidRequest, usersyncs, blabels, labels)
+
+	// Clean PI from bidrequests if not allowed per GDPR
+	gdpr, consent, _ := extractGDPR(bidRequest, e.UsersyncIfAmbiguous)
+	// TODO : do something with the error value
+	if gdpr == 1 {
+		for bidder, bidReq := range cleanRequests {
+			if ok, err := e.gDPR.PersonalInfoAllowed(ctx, bidder, consent); !ok && err == nil {
+				cleanPI(bidReq)
+			}
+		}
+	}
+
 	// List of bidders we have requests for.
 	liveAdapters := make([]openrtb_ext.BidderName, len(cleanRequests))
 	i := 0
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -81,18 +81,7 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
-	cleanRequests, aliases, errs := cleanOpenRTBRequests(bidRequest, usersyncs, blabels, labels)
-
-	// Clean PI from bidrequests if not allowed per GDPR
-	gdpr, consent, _ := extractGDPR(bidRequest, e.UsersyncIfAmbiguous)
-	// TODO : do something with the error value
-	if gdpr == 1 {
-		for bidder, bidReq := range cleanRequests {
-			if ok, err := e.gDPR.PersonalInfoAllowed(ctx, bidder, consent); !ok && err == nil {
-				cleanPI(bidReq)
-			}
-		}
-	}
+	cleanRequests, aliases, errs := cleanOpenRTBRequests(ctx, bidRequest, usersyncs, blabels, labels, e.gDPR, e.UsersyncIfAmbiguous)
 
 	// List of bidders we have requests for.
 	liveAdapters := make([]openrtb_ext.BidderName, len(cleanRequests))
--- a/exchange/gdpr.go
+++ b/exchange/gdpr.go
@@ -7,25 +7,29 @@ import (
 	"github.com/mxmCherry/openrtb"
 )
 
-// ExtractGDPR will pull the gdpr info from an openrtb request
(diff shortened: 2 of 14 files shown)


exchange/exchange.go
```diff
@@ -77,6 +82,18 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
 	cleanRequests, aliases, errs := cleanOpenRTBRequests(bidRequest, usersyncs, blabels, labels)
+
+	// Clean PI from bidrequests if not allowed per GDPR
+	gdpr, consent, _ := extractGDPR(bidRequest, e.UsersyncIfAmbiguous)
+	// TODO : do something with the error value
+	if gdpr == 1 {
+		for bidder, bidReq := range cleanRequests {
+			if ok, err := e.gDPR.PersonalInfoAllowed(ctx, bidder, consent); !ok && err == nil {
```
cirla: If `err != nil` here should we clean by default?

exchange/exchange.go
```diff
@@ -77,6 +82,18 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
 	cleanRequests, aliases, errs := cleanOpenRTBRequests(bidRequest, usersyncs, blabels, labels)
+
+	// Clean PI from bidrequests if not allowed per GDPR
+	gdpr, consent, _ := extractGDPR(bidRequest, e.UsersyncIfAmbiguous)
+	// TODO : do something with the error value
```
cirla: If err != nil here should we clean by default?

exchange/gdpr.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Zero the last byte of an IP address
+func cleanIP(fullIP string) string {
+	i := strings.LastIndex(fullIP, ".")
+	return fullIP[0:i] + ".000"
+}
+
+// Zero the last two sections of an IPv6 address
+func cleanIPv6(fullIP string) string {
+	i := strings.LastIndex(fullIP, ":")
```
cirla: The comment says "last two sections" but this looks like it's just replacing the last group.

exchange/gdpr.go
```diff
@@ -0,0 +1,92 @@
...
+	return fullIP[0:i] + ":0000"
```
cirla: I believe we can omit the extra zeroes and just do `":"`

exchange/gdpr.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Return a cleaned Geo object pointer
+func cleanGeo(geo *openrtb.Geo) *openrtb.Geo {
```
cirla: This is rounding the Lat/Lon to two decimal places?

gdpr/impl.go
```diff
@@ -69,12 +82,48 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
 }
 
-type alwaysAllow struct{}
+func (p *permissionsImpl) allowPI(ctx context.Context, vendorID uint16, consent string) (bool, error) {
+	// If we're not given a consent string, respect the preferences in the app config.
+	if consent == "" {
+		return p.cfg.UsersyncIfAmbiguous, nil
+	}
+
+	parsedConsent, err := vendorconsent.ParseString(consent)
```
cirla: Outside the scope of this PR perhaps, but do we always re-parse the entire string every time we check for a specific permission in it? I vaguely remember us having a "parsed GDPR" struct, but I may be mistaken.

exchange/exchange.go
hhhjort: If the bidRequest is malformed enough that we can't tell if `gdpr` is on, off, or missing, then I am not sure what we can do with it.

exchange/exchange.go
hhhjort: perhaps if we have GDPR enabled when ambiguous. It should be rare that looking this up should generate an error.

exchange/gdpr.go
hhhjort: last 2 bytes, one section ... will fix.

exchange/gdpr.go
hhhjort: Yes

gdpr/impl.go
hhhjort: That is a good point. We would reparse the consent string for every bidder. It would change the method signature from the other allow methods, but would save a few cycles.

exchange/exchange.go
```diff
@@ -77,6 +82,18 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
 	cleanRequests, aliases, errs := cleanOpenRTBRequests(bidRequest, usersyncs, blabels, labels)
+
+	// Clean PI from bidrequests if not allowed per GDPR
```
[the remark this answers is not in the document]
dbemiller: ahhh... ok.

Another option is to make a smaller struct with narrower scope, and put the functions on it. For example: we could make an `auction` distinct from an `exchange`.

Right now, all the code is organized like `func (e *exchange) someFunc(auctionSpecificArgs)`.
Another reasonable way would be `func (a *auction) someFunc()`, where `auctionSpecificArgs` are actually properties on the `auction` itself.

I wouldn't worry about the method args here... for this PR, I think it's important to put the logic alongside the rest of the logic which does the same thing. You could look into refactoring later to reduce the number of method args and make it more readable.

Some good news: I noticed this module getting messy, and expected it to need some refactoring someday... so I moved the majority of the tests in this package into that JSON format. If you do decide it's time to refactor things, you shouldn't need to update many tests as long as you don't change the `HoldAuction` method signature.

exchange/gdpr_test.go
```diff
@@ -0,0 +1,119 @@
+package exchange
+
+import (
+	"testing"
+
+	"github.com/mxmCherry/openrtb"
+)
+
+func TestExtractGDPRFound(t *testing.T) {
```
dbemiller: https://github.com/prebid/prebid-server/pull/658#discussion_r212363297 reminded me...

It sounds like we both think the `exchange` code can be a little hard to read right now. I would recommend making JSON-style tests for these, so that the tests will still work if/when the code gets too messy and needs refactoring.

I wouldn't block the merge for it, though... it'll just make future refactors of this module easier.

exchange/utils.go
```diff
@@ -28,6 +30,18 @@ func cleanOpenRTBRequests(orig *openrtb.BidRequest, usersyncs IdFetcher, blables
 	}
 
 	requestsByBidder, errs = splitBidRequest(orig, impsByBidder, aliases, usersyncs, blables, labels)
+
+	// Clean PI from bidrequests if not allowed per GDPR
+	gdpr, consent, _ := extractGDPR(orig, usersyncIfAmbiguous)
+	// TODO : do something with the error value
```
dbemiller: Can just delete the `TODO` if you're not going to fix it now... these never get done later.

IMO PBS should probably just act by `usersyncIfAmbiguous` on error though.

exchange/gdpr.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/mxmCherry/openrtb"
+)
+
+// ExtractGDPR will pull the gdpr info from an openrtb request
+func extractGDPR(bidRequest *openrtb.BidRequest, usersyncIfAmbiguous bool) (gdpr int, consent string, err error) {
```
dbemiller: small note... it's a little weird that this is one function call. Parsing `gdpr` and parsing `consent` are two disjoint tasks.

Two smaller functions are generally better than one big one. More flexible, reusable, and easy to test.

exchange/gdpr.go
```diff
@@ -0,0 +1,92 @@
...
+		// Need to duplicate pointer objects
+		device := *bidRequest.Device
+		bidRequest.Device = &device
+		bidRequest.Device.DIDMD5 = ""
+		bidRequest.Device.DIDSHA1 = ""
+		bidRequest.Device.DPIDMD5 = ""
+		bidRequest.Device.DPIDSHA1 = ""
+		bidRequest.Device.IP = cleanIP(bidRequest.Device.IP)
+		bidRequest.Device.IPv6 = cleanIPv6(bidRequest.Device.IPv6)
+		if bidRequest.Device.Geo != nil {
```
dbemiller: Can move these `!= nil` checks into `cleanGeo`. It makes `cleanGeo` nil-safe, and avoids the duplicate check in this function.

exchange/gdpr.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// ExtractGDPR will pull the gdpr info from an openrtb request
+func extractGDPR(bidRequest *openrtb.BidRequest, usersyncIfAmbiguous bool) (gdpr int, consent string, err error) {
+	var re regsExt
+	if bidRequest.Regs != nil {
+		err = json.Unmarshal(bidRequest.Regs.Ext, &re)
+	}
+	if re.GDPR == nil || err != nil {
+		if usersyncIfAmbiguous {
```
dbemiller: I think this is backwards... if `usersyncIfAmbiguous` is true, shouldn't the "effective" `gdpr` be `0`?

exchange/utils.go
dbemiller: actually... now that you mention it, I think we _do_ kill it completely.

- https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/auction.go#L670
- https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/auction.go#L641

So... doesn't really matter how you handle errors here. Input validation is the first thing we do.

exchange/gdpr.go
hhhjort: I even had a test for this, but my test had the same bug. :)

exchange/gdpr.go
```diff
@@ -0,0 +1,95 @@
...
+	"encoding/json"
+	"strings"
+
+	"github.com/mxmCherry/openrtb"
+)
+
+// ExtractGDPR will pull the gdpr flag from an openrtb request
+func extractGDPR(bidRequest *openrtb.BidRequest, usersyncIfAmbiguous bool) (gdpr int) {
+	var re regsExt
+	var err error
```
dbemiller: This `err` isn't used anywhere anymore.

Did you mean to `panic` on error in here?

hhhjort: It is used ... `if ... || err != nil`

exchange/gdpr_test.go
hhhjort: JSON-style tests will require a lot of extra work to test everything that needs testing, so will save that for another day.

PR #661, opened 2018-08-27
Test go 1.11, update dep

cirla:
Go 1.11 was recently released. Begin testing all PRs/builds against it. Update to recent version of `dep` in travis-ci, went back to upstream jsonparser, and updated some deps.

Leaving go 1.9 in travis for now, we should figure out when we want to stop officially supporting go 1.9. The builds are 2x slower than 1.10.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ go:
 env:
   - DEP_VERSION="0.5.0"
 
-matrix:
-  allow_failures:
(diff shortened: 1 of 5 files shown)


dbemiller:
just fyi... you probably don't need to `allow_failures`, at least for that error. Just replace `		glog.Infof(msg, args)` with `glog.Infof(msg, args...)`

PR #667, opened 2018-09-05
Beachfront panic fix

hhhjort:
Resolves a panic for any request whose site.page does not include the protocol piece.


hhhjort:
@muncha : Flagging you as it seems to be your adapter. :)

dbemiller:
should add a regression test for this

PR #673, opened 2018-09-10
Move router and legacy auction to router directory

zachbadgett:
@cirla @dbemiller 

With the growing popularity of prebid I think the prebid-server project can benefit from isolating some components so it can be reusable. I think one of the main goals should be complete adapter isolation, this will allow others to add adapters without needing to modify prebid-server code.

My original pull request was going to be based off: https://github.com/zachbadgett/prebid-server/commit/57a580ca4fd46a3d1cb07e3d3c2c52c383a0a3ba
However, I felt like that may be too many changes at once.

If you guys like this idea I can add pieces one by one and we can discuss each part.

The first part is the current source code changes; moving all router and legacy logic to the router directory. If everyone likes this I'll move and fix all the tests.


dbemiller:
Cool! Very nice to see someone working on code structure.... we definitely need it.

In general I'm definitely a fan of more, smaller packages... and everyone agrees we should be pulling things out of the root package and making modules for them.

`router` looks like as good a package boundary as anything else. We do need to bind routes to endpoints, apply middleware, and things like that.

Only concern I have is moving all the legacy auction code in there. At its logical conclusion, does that mean you'd want to move [all the other endpoint implementations](https://github.com/prebid/prebid-server/tree/master/endpoints) into that package too? If so, it seems like that package would become pretty huge rather quickly.

zachbadgett:
I was considering the legacy auction a special use case only because it existed in `pbs_light.go` I'm all for pulling it out.

dbemiller:
ah! ok.

In that case, could you put it in `/endpoints/auction.go` instead?

The convention today is that every endpoint is inside `/endpoints` in a file where the path matches its route. So for example, `<host>/openrtb2/auction` is in `/endpoints/openrtb2/auction.go`, and `<host>/status` is in `/endpoints/status`.

I'm definitely open to discussing better structures there... but without an overall plan, I'd rather move towards consistency. The only reason legacy auctions were still in `pbs_light.go` was because we thought we'd be able to delete it from the project much sooner and didn't want to spend the time moving them.

zachbadgett:
Alright, that sounds good to me. I'll move it there :D

zachbadgett:
@dbemiller Should I squash the commits?

hhhjort:
Don't worry about squashing the commits, we always squash when merging a PR.

I kind of feel that `newExchangeMap()` should be moved to `/endpoints`, and called directly when creating the `Auction` object, I don't think we need it anywhere else. While we are cleaning up this code anyway. :) Won't insist on it if you guys feel it is out of scope.

zachbadgett:
@hhhjort I didn't move it because the goal is to remove it completely. Statically defining a map of adapters will limit the extendability of the library.

PR #676, opened 2018-09-12
Fixes panic with cleaning blank IPs

hhhjort:


dbemiller:
don't think this needs to wait for a second review... problem is obvious, fix is simple, and regression tests make sense.

PR #679, opened 2018-09-14
Fixing a few stray issues

dbemiller:
I found these during a last-minute review of #673.

Rather than bug @zachbadgett I just figured I'd change them right after merging. This fixes two issues:

1. `router.Shutdown()` was never called.
2. The schema directory only worked in tests, but not at runtime.

(2) is an issue because the tests run in the directory where they're defined, but the static binary gets output to the root directory... so the paths are actually different for each case.

Those changes stop PBS from crashing on startup.


zachbadgett:
Ah, I thought I fixed the schema directory. Thanks for fixing those!

PR #681, opened 2018-09-17
Allows sending user.buyerid for AMP requests

hhhjort:
DONT MERGE UNTIL #682 HAS BEEN TESTED AND DOES NOT RESOLVE THE ISSUE

AMP currently does not support generating/sending IAB consent strings. After merging #658 AMP performance tanked in the EU. This is likely due to the PR removing `user.buyerid` where consent is not set, and being impossible to set consent on AMP requests. We still leave in place the other PI information cleaning code.

This is intended as a patch for the AMP consent issue until AMP supports sending full consent information to 3rd party requests. Ideally they will support the IAB consent string, but we cannot discount that they may roll their own format.

The motivation behind allowing `user.buyerid` to go through is that we cannot know the user's consent, however if the ID exists then the user must have given consent for the bidder on at least one website that does prebid, we just can't confirm that the current website is also covered. Also, the `buyerid` is not itself personally identifying information, it can just be considered potentially identifying if the bidder already has some personal information associated with the user ID. So there exists an argument that sending the buyer ID does not violate the PI portion of the GDPR, but we mask it for non-AMP just to be safe.

The other GDPR pitfall is tracking consent. Prebid server does not log the `buyerid`, so it is not directly tracking the user. We send the bidder request with the GDPR flag set, but no consent string, so the bidder should consider no tracking consent has been given. So in theory no tracking data should be saved, but we do enable a bidder to track this user if the bidder is willing to violate the GDPR and the user has consent to tracking by this bidder, but not consented on this specific site in question. This should be a rare case.

### DIFF
--- a/exchange/gdpr.go
+++ b/exchange/gdpr.go
@@ -49,12 +49,14 @@ type regsExt struct {
 }
 
 // cleanPI removes IP address last byte, device ID, buyer ID, and rounds off lattitude/longitude
-func cleanPI(bidRequest *openrtb.BidRequest) {
+func cleanPI(bidRequest *openrtb.BidRequest, isAMP bool) {
 	if bidRequest.User != nil {
 		// Need to duplicate pointer objects
 		user := *bidRequest.User
 		bidRequest.User = &user
-		bidRequest.User.BuyerUID = ""
+		if isAMP == false {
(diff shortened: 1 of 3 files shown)


hhhjort:
#682 did not resolve

PR #684, opened 2018-09-17
Support sending a TTL parameter to prebid cache

hhhjort:


exchange/auction.go
```diff
@@ -137,6 +144,27 @@ func maybeMake(shouldMake bool, capacity int) []prebid_cache_client.Cacheable {
 	return nil
 }
 
+func cacheTTL(impTTL int64, bidTTL int64, buffer int64) (ttl int64) {
```
cirla: Just wanted to clarify I'm understanding the logic correctly:

1. If both imp and bid TTL are <= 0, ttl is 0
2. If just one of imp or bid TTL is <= 0, use the one that's not (+ buffer)
3. If both imp and bid TTL are > 0, use the lesser of the two (+ buffer)

prebid_cache_client/client.go
```diff
@@ -129,7 +130,13 @@ func encodeValueToBuffer(value Cacheable, leadingComma bool, buffer *bytes.Buffe
 
 	buffer.WriteString(`{"type":"`)
 	buffer.WriteString(string(value.Type))
-	buffer.WriteString(`","value":`)
+	if value.TTLSeconds > 0 {
+		buffer.WriteString(`","ttlseconds":`)
+		buffer.WriteString(string(value.TTLSeconds))
```
dbemiller: This doesn't do what you think it does... see: https://play.golang.org/p/khS3hM1y7v9

Needs to be `strconv.Itoa(value.TTLSeconds)`. Please add a test to catch this.

exchange/auction.go
dbemiller: One way you can make this more readable is to "return" as soon as you know what the return value is.

```
if impTTL <= 0 {
  return bidTTL
}
if bidTTL <= 0 {
  return impTTL
}
if impTTL < bidTTL {
  return impTTL
}
return bidTTL
```

Advantage: others don't have to read through _all_ the code to decipher the behavior.
Disadvantage: it's tough do add the buffer logic properly.

You could work around that by splitting `addBuffer(base int64, buffer in64) int64` into its own function, though.

I'm ok with it as is, too, but... food for thought.

exchange/auction_test.go
```diff
@@ -29,3 +36,137 @@ func TestMakeVASTNurl(t *testing.T) {
...
+type mockCache struct {
+	items []prebid_cache_client.Cacheable
+}
+
+func (c *mockCache) PutJson(ctx context.Context, values []prebid_cache_client.Cacheable) []string {
+	c.items = values
+	return []string{"", "", "", "", ""}
+}
+
+// JSONBytesEqual compares the JSON in two byte slices.
```
dbemiller: check out the libraries :)

https://godoc.org/github.com/stretchr/testify/assert#JSONEq

hhhjort: Ack, I tried finding something like this, but failed. Probably because I tried too hard to find if assert could check if one particular object existed in a list ... and then devolved in this direction trying to find some formula that worked.

hhhjort:
About not seeing a way to create a better test without a ton of work, I think 90% of the work I put into this PR was writing the tests for it. :)

exchange/auction_test.go
hhhjort: Nevermind, this does not work for my purpose, thats why I did not find/use it. For the majority of times I call this function, it is expected to return a `false` value. The true test is if I get the correct number of true results from this function.

dbemiller: ah! That makes sense.

I logged an issue here... we'll see if they're open to adding support: https://github.com/stretchr/testify/issues/673

PR #688, opened 2018-09-19
Move user syncs to each respective adapter directory

zachbadgett:
Continued progress on adapter isolation.

Code complete and unit tests updated, still need to build and run it.


internal/testutil/usersync.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type usersyncTest struct {
+	t          *testing.T
+	usersyncer usersync.Usersyncer
+	syncInfo   *usersync.UsersyncInfo
+}
+
+// Assert common values match.
+func (u *usersyncTest) Assert(url, syncType string, vendor uint16, supportCORS bool) {
```
zachbadgett: Maybe separate each value out into their own method `AssertFamilyName`

adapters/pubmatic/usersync.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"net/url"
+
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/usersync"
+)
+
+func NewPubmaticSyncer(cfg *config.Configuration) usersync.Usersyncer {
```
zachbadgett: Another way to do this method is to have a signature like:
```
func NewPubmaticSyncer(externalURL string, adapter config.Adapter) usersync.Usersyncer
```
which would avoid the need for some adapters to do a lookup on the adapter map.

internal/testutil/usersync.go
```diff
@@ -0,0 +1,44 @@
+package testutil
```
zachbadgett: I don't like the idea of a "testutil" package...however, it's in the internal directory which is not exposed to the public api.

Thoughts?

dbemiller: I'm a fan of code sharing. In the vast majority of cases, utils are much better than duplicate code.

I'm not too worried about the package API, either, because this is an application  and not a library. Our REST APIs are _very_ important... but the package APIs are less so because we don't expect people to use this code by importing it. Any package structure within this project is purely for our own sanity.

That said: if you take a step back, do you think these utilities really improve things? Compare these code snippets (loosely based on the `sovrn` syncer).

Without utils (using [testify](https://godoc.org/github.com/stretchr/testify/assert)):
```
info := syncer.GetUsersyncInfo("0", "")
assert.Equal(t, "//ap.lijit.com/pixel?redir=external.com%2Fsetuid%3Fbidder%3Dsovrn%26gdpr%3D0%26gdpr_consent%3D%26uid%3D%24UID", info.URL)
assert.Equal(t, "redirect", info.Type)
assert.False(t, info.SupportCORS)
assert.Equal(t, 13, syncer.GDPRVendorID())
```

With utils:
```
u := testutil.UsersyncTest(t, syncer, syncer.GetUsersyncInfo("0", ""))
u.Assert(
  "//ap.lijit.com/pixel?redir=external.com%2Fsetuid%3Fbidder%3Dsovrn%26gdpr%3D0%26gdpr_consent%3D%26uid%3D%24UID",
  "redirect",
  13,
  false,
)
```

Advantages of no utils:
- Fewer lines of code
- Clearer to read
- No need to implement the utils function
- No need to find a good package location for the utils implementation

Advantages of utils:
- When a field gets added to the usersyncer, we'd update `u.Assert` and nobody's tests would compile until _every_ adapter is running the assertion

adapters/pubmatic/usersync.go
dbemiller: Either way works. I do agree with your comment here... reducing the work adapters have to do is definitely an advantage. I can't think of any disadvantages, either, so I'll give a small edge to `NewPubmaticSyncer(externalURL string, adapter config.Adapter)`.

It's probably not a huge deal, though.

dbemiller:
Hey @zachbadgett... this still looks like a work-in-progress to me, but let me know when you want a review on it. Or re-edit the title to remove the `[WIP]`

adapters/pubmatic/usersync.go
zachbadgett: I think the main reason I kept it that way was in case there were only configuration additions. I say we keep it for now and change it later if it makes sense.

internal/testutil/usersync.go
zachbadgett: I was mostly avoiding adding another dependency, didn't realize testify was already being used so I will update it.

One quick comment about your statement: 
> I'm not too worried about the package API, either, because this is an application and not a library.
Treating the components as libraries and stabilizing the API will allow the community to use the code here and contribute back improvements. Keeping it isolated as an application will cause users to fork and less likely to contribute anything back (I know I sound like a broken record at this point, however I really feel like this will improve the health of the project).

dbemiller: That's an interesting perspective... I wouldn't have thought this would make it hard for people to contribute to the project.

You're not a broken record at all. From your previous comments, I thought your main issue was that you or someone you knew was trying to add a Bidder and finding it annoying to change lots of different places in the code. I didn't realize you planned to import some packages and rely on them directly. Thanks for the perspective.

I do agree that the healthier an application architecture is, the more it resembles a bunch of libraries with a thin layer tying them all together... so whether or not we ultimately fold package APIs into our "v1" compatibility, it's definitely helpful to organize the code as if we plan to.

zachbadgett:
@dbemiller this is done unless you have any other changes/comments.

endpoints/openrtb2/openrtb2_go111_test.go
```diff
@@ -0,0 +1,23 @@
+// +build go1.11
```
dbemiller: Can you split this Go 1.11 fix off into a separate PR?  It's a valuable change... but commits do get squashed on merge, and it's unrelated.

Small note: I'd prefer something like `message_go1.11` rather than `message_alt`. We already know that this bug has been fixed in Go 1.11.1 (see https://github.com/golang/go/issues/27275), so... it's a pretty specific fix.

Would also be nice to lin this issue somewhere in the file so we remember why the workaround exists.

zachbadgett: I was thinking the same thing for the message name. I went with message_alt because it will apply to go1.11+, now that I know it was an actual bug and it's been fix it makes a lot more sense to do message_go1.11.

adapters/syncer_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"testing"
+
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/usersync"
+)
+
+func TestSyncers(t *testing.T) {
+	syncers := make(map[openrtb_ext.BidderName]usersync.Usersyncer)
+	for _, bidderName := range map[string]openrtb_ext.BidderName{} {
```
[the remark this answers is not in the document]
zachbadgett: Not forgotten about, see my comment above. The map is ultimately moved to the adapters package, however importing the static map from where it lives now would cause a circular dependency.

usersync/usersyncers/syncer_test.go
```diff
@@ -7,40 +7,30 @@ import (
...
-func TestSyncers(t *testing.T) {
-	cfg := &config.Configuration{}
-	syncers := NewSyncerMap(cfg)
-	for _, bidderName := range openrtb_ext.BidderMap {
-		if _, ok := syncers[bidderName]; !ok {
-			t.Errorf("No syncer exists for adapter: %s", bidderName)
-		}
+func TestNewSyncerMap(t *testing.T) {
+	cfg := &config.Configuration{
+		Adapters: map[string]config.Adapter{
```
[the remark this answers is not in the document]
zachbadgett: In an effort to isolate adapters entirely I was thinking of doing something like this: https://github.com/zachbadgett/prebid-server/commit/57a580ca4fd46a3d1cb07e3d3c2c52c383a0a3ba#diff-25c7c8df3a616bd75098f16d158d0a0bR29

This could still achieve your goal of "impossible" to do wrong by doing a panic on `Register`.

The `usersync/usersyncers/` would ultimately be removed.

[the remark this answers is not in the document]
zachbadgett: Keep in mind the design decisions I made in that commit attempts to change other code as little as possible (especially in the adapters themselves).

adapters/syncer_test.go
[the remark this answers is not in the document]
dbemiller: aha, yeah that makes sense. I wasn't sure which one of the two tests you wanted to be the "real" one.

Can you just delete it? It doesn't make much sense to have tests without assertions.

adapters/syncer_test.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+}
+
+// Bidders may have an ID on the IAB-maintained global vendor list.
+// This makes sure that we don't have conflicting IDs among Bidders in our project,
+// since that's almost certainly a bug.
+func TestVendorIDUniqueness(t *testing.T) {
+	syncers := make(map[openrtb_ext.BidderName]usersync.Usersyncer)
+	idMap := make(map[uint16]openrtb_ext.BidderName, 0)
+	for name, syncer := range syncers {
```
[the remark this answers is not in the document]
dbemiller: Given the circular dependency issue: could you either move this test back into the `usersyncers` package for now, or else move the static map into this package to avoid the circular dependency?

Until the `Register`/`panic` changes are in, this test _does_ have some value... and I have a few good reasons not to go that route, so I'd rather treat this PR as an isolated change set while we discuss.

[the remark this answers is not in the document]
zachbadgett: Sure thing!

usersync/usersyncers/syncer_test.go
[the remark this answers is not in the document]
dbemiller: yeah, that's fair.

I'm trying to make sure that the code is left in a good state after each PR. Someone might try to submit a Bidder between now and the time the next PR merges, or the next PR might never come.

We might also spend a while discussing tradeoffs. For `init` & `Register` in particular, I know we'll have at least some back and forth. I'm most definitely not an experienced Gopher... but I can think of some good language-agnostic reasons to avoid that structure, and the benefits that I can think of don't outweigh the costs. As an avid Go supporter, I'm curious to learn about how you weigh them :).

That's a long-winded way of saying: if the future changes don't pan out, I'd rather not lose the value provided by this test.

hhhjort:
Looks like some of David's comments haven't been addressed yet? Are there still changes in the works, or is the current state what you think the final state should be?

zachbadgett:
This is the final state, I fixed what he commented on. I'll resolve the conflicts today.

dbemiller:
oh! Sorry. I didn't know you were waiting on me here.

Open issues I see are these:

- https://github.com/prebid/prebid-server/pull/688#discussion_r220617758
- https://github.com/prebid/prebid-server/pull/688/files#r220590955

Goals are:

- Don't waste cycles running a test with no assertions
- Have a test which fails if someone contributes a Bidder but forgets their Usersyncer.

zachbadgett:
Heh...I moved the tests but didn't commit the reverting of them. My mistake! Should be fixed now.

adapters/syncer_test.go
```diff
@@ -0,0 +1 @@
+package adapters
```
dbemiller: Got an empty file here too... not sure if I only just noticed this now, or it became empty in those last changes.

dbemiller: gotcha, ok. no worries

PR #689, opened 2018-09-19
Fixes a test to match the IP cleaning patch

hhhjort:


hhhjort:
One line test fix, one review should be enough

Issue #399, opened 2018-03-09
Support Native version 1.2


dbemiller:
Our OpenRTB library has added Native 1.2 support: https://github.com/mxmCherry/openrtb/releases/tag/v10.0.0

Right now, our input validation only checks against Native v1.1. We should upgrade it to validate v1.2 fields as well. We'll also have to fix the other breaking changes in the mxmcherry library (mainly: `openrtb.RawJSON` got replaced with `json.RawMessage`)

ORTB spec for Native v1.2 is here: https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf

muncha:
After doing my upstream pull into my working fork, I was unable to build because of the RawJSON part of this. For testing purposes, it worked for me to roll back mxmCherry/openrtb to:

https://github.com/mxmCherry/openrtb/commit/c426d335a3506bd009710ee7a6b5fcf943e45c91

muncha:
After merging upstream master into my working fork, I was unable to build because of the RawJson change as well as:
endpoints/openrtb2/auction.go:337:30: undefined: request.ContextType
endpoints/openrtb2/auction.go:344:37: undefined: request.PlacementType

For testing purposes I'm getting around that for now by rolling back mxmCherry/openrtb to:
https://github.com/mxmCherry/openrtb/commit/2f6127d966dc46d16e90f71acc6b304177460723

dbemiller:
@muncha this project doesn't use the latest version of the OpenRTB library yet. The `Gopkg.lock` file should tie it to the version it's intended to work with.

Did you follow the README while setting up PBS? https://github.com/prebid/prebid-server#installation

The `dep ensure` line should download the version which PBS is compatible with.

When this ticket is implemented, we'll bump the version in `Gopkg.toml` to 10.0.0, and then we'll have to fix PBS code because of the breaking changes you mentioned. See https://github.com/prebid/prebid-server/blob/master/Gopkg.toml#L57

basantashah:
```github.com/prebid/prebid-server/openrtb_ext
# github.com/prebid/prebid-server/openrtb_ext
../../../../../Go/src/github.com/prebid/prebid-server/openrtb_ext/bidders.go:88:32: undefined: openrtb.RawJSON
../../../../../Go/src/github.com/prebid/prebid-server/openrtb_ext/bidders.go:138:70: undefined: openrtb.RawJSON```

I am still not able to fix this. `HELP`

PR #693, opened 2018-09-21
Native 1.2 input validation

dbemiller:
This breaking change only affects Native bid requests. I got no responses when asking reddit's adops slack if anyone was using PBS for Native yet... so this shouldn't affect many people.

This contributes support for #399. I wouldn't exactly call #399 "fixed," though, until we validate bid responses too. I would put that in a separate PR though.


endpoints/openrtb2/auction.go
```diff
@@ -475,13 +505,38 @@ func validateNativeAsset(asset nativeRequests.Asset, impIndex int, assetIndex in
...
+
 func validateNativeAssetTitle(title *nativeRequests.Title, impIndex int, assetIndex int) error {
 	if title.Len < 1 {
 		return fmt.Errorf("request.imp[%d].native.request.assets[%d].title.len must be a positive integer", impIndex, assetIndex)
 	}
 	return nil
 }
 
+func validateNativeEventTracker(tracker nativeRequests.EventTracker, impIndex int, eventIndex int) error {
+	if tracker.Event < 1 || tracker.Event > 4 {
```
cirla: I know these are defined in the spec but maybe 
```go
if tracker.Event < native.EventTypeImpression || tracker.Event > native.EventTypeViewableVideo50 {
```
and similarly in other places using the named constants.

dbemiller: akh... good call. Will fix

PR #701, opened 2018-09-26
Add message_go1.11 to fix test failing on go1.11

zachbadgett:
Unfortunately, the build flags (at least from what I can tell), do not allow specifying minor versions. When `go1.11.1` is released we can remove the hack and possibly add a warning in the README explaining why `go1.11` is not supported/why the tests fail on that specific version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ go:
 env:
   - DEP_VERSION="0.5.0"
 
-matrix:
-  allow_failures:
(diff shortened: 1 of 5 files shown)


dbemiller:
No reason to hold up this merge... but this version could be removed from the `allow_failures` block in the `.travis.yml` now.

zachbadgett:
Will remove it :D

PR #702, opened 2018-10-02
Remove "capability violation" logic

### LINKED ISSUE #555: Preprocess bid requests for valid media types

dbemiller:
None of these `glog` messages have shown up in our app logs... so it's pretty safe to assume the `adapters.EnforceBidderInfo` middleware is working as intended.

Changes here:

- Deletes the adapter-specific logic which validates Impression types.
- Removes adapter tests which target that logic
- Minor refactor of the `newAdapterMap` function

Motivation for the refactor is so that new Bidders don't "forget" to add the middleware. It's pretty important now.

This fixes #555.


hhhjort:
Test coverage is provided by the generic capability checks. When we moved to generic we kept in these checks and logging to confirm that these old checks were no longer getting used. This is just the final cleanup

PR #703, opened 2018-10-04
Prevent Imps with duplicate IDs

### LINKED ISSUE #602: Return 400 if the input request has Imps with duplicate IDs
During a debugging session with a publisher, we found that some Prebid.js setups will send a request like this:

```
{
  "imp": [
    {
      "id": "abc"
    },
    {
      "id": "abc"
    }
  ]
}
```

I'll open a separate issue in the `Prebid.js` project to fix this... but from our end, we should return a 400 on requests with duplicate Imp IDs. It would have made the debugging process much faster.

dbemiller:
Fixes #602.

Although this is a breaking change, a caller who sends a request like this couldn't possibly interpret the results anyway.

Each `response.seatbid[i].bid[j]` contains an `impid`, which refers to an ID from the Request. If the Request has two Imps with the same ID, there's no way to know which Imp that Bid applies to.

This just makes debugging easier for callers who make this mistake.


hhhjort:
This has gone past 30 days

PR #704, opened 2018-10-04
Initial RhythmOne Prebid Server Adapter

dineshkumarjayaram:
Initial RhythmOne Prebid Server Adapter support


adapters/rhythmone/rhythmone.go
```diff
@@ -0,0 +1,218 @@
...
+			err = json.Unmarshal(bidderExt.Bidder, &rhythmoneExt)
+			if err != nil {
+				err = &errortypes.BadInput{
+					Message: fmt.Sprintf("placementId | zone | path not provided in imp id=%s. Abort all Request", imp.ID),
+				}
+				errors = append(errors, err)
+				return nil, "", errors
+			}
+
+			if rhythmoneExt.PlacementId == "" || rhythmoneExt.Zone == "" || rhythmoneExt.Path == "" {
```
dbemiller: This constraint should go in your `static/bidder-params/rhythmone.json` schema as a `minlength`: https://json-schema.org/understanding-json-schema/reference/string.html#length

Then you can remove this logic here. If the params don't validate against that schema, your adapter won't be called.

adapters/rhythmone/rhythmone.go
```diff
@@ -0,0 +1,218 @@
...
+	mediaType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Banner != nil {
+				mediaType = openrtb_ext.BidTypeBanner
+			} else if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			if imp.Banner != nil && imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeBanner
```
dbemiller: This `if imp.Banner != nil && imp.Video != nil {` block is a no-op, and can be deleted.

If both are non-nil, the `if imp.Banner != nil {` block will already have set this to BidTypeBanner.

adapters/rhythmone/rhythmone.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+			return mediaType
+		}
+	}
+	return mediaType
+}
+
+func NewRhythmoneAdapter(config *adapters.HTTPAdapterConfig, endpoint string) *RhythmoneAdapter {
```
dbemiller: This function is unused, and can be deleted.

exchange/adapter_map.go
```diff
@@ -61,5 +62,6 @@ func newAdapterMap(client *http.Client, cfg *config.Configuration, infos adapter
 			infos[string(openrtb_ext.BidderRubicon)]), client),
 		openrtb_ext.BidderSomoaudience: adaptBidder(adapters.EnforceBidderInfo(somoaudience.NewSomoaudienceBidder(cfg.Adapters[string(openrtb_ext.BidderSomoaudience)].Endpoint), infos[string(openrtb_ext.BidderSomoaudience)]), client),
 		openrtb_ext.BidderSovrn:        adaptBidder(adapters.EnforceBidderInfo(sovrn.NewSovrnBidder(client, cfg.Adapters[string(openrtb_ext.BidderSovrn)].Endpoint), infos[string(openrtb_ext.BidderSovrn)]), client),
+		openrtb_ext.BidderRhythmone:    adaptBidder(adapters.EnforceBidderInfo(rhythmone.NewRhythmoneBidder(client, cfg.Adapters[string(openrtb_ext.BidderRhythmone)].Endpoint), infos[string(openrtb_ext.BidderRhythmone)]), client),
```
dbemiller: Alphabetize by key please.

openrtb_ext/bidders.go
```diff
@@ -37,6 +37,7 @@ const (
 	BidderRubicon      BidderName = "rubicon"
 	BidderSomoaudience BidderName = "somoaudience"
 	BidderSovrn        BidderName = "sovrn"
+	BidderRhythmone    BidderName = "rhythmone"
```
dbemiller: alphabetize by key please

usersync/usersyncers/rhythmone_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"testing"
+)
+
+func TestRhythmoneSyncer(t *testing.T) {
+	an := NewRhythmoneSyncer("https://sync.1rx.io/usersync2/rmphb?redir=", "localhost")
+	syncInfo := an.GetUsersyncInfo("1", "BOPVK28OVJoTBABABAENBs-AAAAhuAKAANAAoACwAGgAPAAxAB0AHgAQAAiABOADkA")
+	url := "https://sync.1rx.io/usersync2/rmphb?redir=localhost%2Fsetuid%3Fbidder%3Drhythmone%26gdpr%3D1%26gdpr_consent%3DBOPVK28OVJoTBABABAENBs-AAAAhuAKAANAAoACwAGgAPAAxAB0AHgAQAAiABOADkA%26uid%3D%5BRX_UUID%5D"
+	assertStringsMatch(t, url, syncInfo.URL)
```
dbemiller: Use `testify` for cleaner assertions: https://godoc.org/github.com/stretchr/testify/assert

`assert.Equal()` should do the trick here.

openrtb_ext/imp.go
```diff
@@ -2,10 +2,11 @@ package openrtb_ext
 
 // ExtImp defines the contract for bidrequest.imp[i].ext
 type ExtImp struct {
```
dbemiller: It looks like this struct is unused throughout the project... you can just delete it.

Rajeshk08:
Updated code review comments as per dbemiller. 
Please review

openrtb_ext/imp.go
Rajeshk08: ExtImp struct is used in openrtb_ext/imp_rhythmone.go file for adding custom extension object in bid request.

adapters/rhythmone/rhythmone.go
```diff
@@ -0,0 +1,148 @@
...
+			} else if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			return mediaType
+		}
+	}
+	return mediaType
+}
+
+func NewRhythmoneBidder(client *http.Client, endpoint string) *RhythmoneAdapter {
```
dbemiller: `client` is unused... please remove it.

dineshkumarjayaram:
Review comments incorporated, please review.

adapters/rhythmone/rhythmone.go
```diff
@@ -0,0 +1,148 @@
...
+func NewRhythmoneBidder(endpoint string) *RhythmoneAdapter {
+	return &RhythmoneAdapter{
+		endPoint: endpoint,
+	}
+}
+
+func (a *RhythmoneAdapter) preProcess(req *openrtb.BidRequest, errors []error) (*openrtb.BidRequest, string, []error) {
+	numRequests := len(req.Imp)
+
+	requestImpCopy := req.Imp
```
hhhjort: Since `req.Imp` is a slice, and a slice is essentially a pointer to an existing array structure, `requestImpCopy := req.Imp` will end up referencing the same underlying array as `req.Imp` So it is not a true copy. 

You only seem to use `reqestImpCopy` for `imp := requestImpCopy[i]`, which will make `imp` a copy shallow of the original `imp`. So there does not seem to be a reason to define `reqestImpCopy`.

adapters/rhythmone/rhythmone.go
```diff
@@ -0,0 +1,148 @@
...
+		bidderExtCopy := openrtb_ext.ExtBid{
+			Bidder: rhythmoneExtCopy,
+		}
+		impExtCopy, err := json.Marshal(&bidderExtCopy)
+		if err != nil {
+			errors = append(errors, err)
+			return nil, "", errors
+		}
+		imp.Ext = impExtCopy
+		req.Imp[i] = imp
```
hhhjort: This could be dangerous, but we do supply each adapter with its own shallow copy of the request imps, so this line will not cause any conflicts. So this is good. :)

adapters/rhythmone/params_test.go
```diff
@@ -0,0 +1,52 @@
...
+			t.Errorf("Schema allowed unexpected params: %s", invalidParam)
+		}
+	}
+}
+
+var validParams = []string{
+	`{"placementId":"123", "zone":"12345", "path":"34567"}`,
+}
+
+var invalidParams = []string{
```
hhhjort: Since you require all 3 paramaters, a test where you have the incorrect type for one parameter could fail because the other two parameters are missing rather than because you use the wrong type. You should add valid values for the other two parameters when you test for an invalid parameter value so you can be sure you are testing that you fail because of that rather than possibly failing only because you omitted the other required parameters.

dineshkumarjayaram:
@hhhjort  - Review comments incorporated, please review.

hhhjort:
We have merge conflicts now

hhhjort:
Ack, looks like an older PR that just merged is what actually caused the conflicts, and the change shouldn't have merged. Not sure why git merged it that way, but we should be keeping `EnforceBidderInfo` Will need to git that fixed, sorry for the confusion.

hhhjort:
Never mind the last comment, that change was a valid part of that PR, so it is needed. I just got a little confused.

Issue #280, opened 2018-01-16
Open RTB: Add currency support


sandraleon:
This issue is to track adding currency support in Prebid Server.

There will be a new currency object located under bidrequest.ext:
```
	"currency": {
		"rates": {
			"USD": {
				"JPY": 110
			}
		}
	}
```

The ad server currency **must** be set either in bidrequest.cur (must be **one** ISO 4217 alpha codes string in array)
```
    "cur": ["USD"],
```

or in config file
```
    "ad_server_currency": "USD",
```

The currency specified in the bid request will take precedence.

Bidders who support multiple currencies will have to specify the currency for their bids. The currency set (in the request or config) will be provided to bidders so that they can pass that info on to their bidder servers. 

Bid price gets converted based on conversion rates available. Conversion rates can be provided in the bid request as shown above. If rates are not provided there, the latest conversion rates will be fetched.

dbemiller:
I'd like to question the idea of a `granularityMultiplier`.

In Prebid.js right now, I see a few ways to [set price granularity](http://prebid.org/dev-docs/publisher-api-reference.html#setConfig-Price-Granularity):

1. Choose one of the pre-defined buckets (low, med, high, etc)
2. Define custom buckets explicitly (by passing min, max, and interval)
3. Set a currency conversion + `granularityMultiplier`.

It seems to me like (2) is both clearer and more flexible than (3). Is that true, or is there some value here I'm not seeing?

If there's no value add, and we plan to implement custom buckets too, then we should skip the `granularityMultiplier` here. Complexity without value makes for a bad user experience.

dbemiller:
It looks like #281 has stalled... so I'll describe the feature here for posterity in case anyone wants to take it on.

This is really two feature requests. These could easily be implemented separately, to keep the work in small chunks.

1. If `request.cur` contains some acceptable currencies, PBS should be able to auto-convert bids into one of those currencies using some sensible default rates
2. If  `request.prebid.ext.currency.rates` and `request.cur` are defined, publishers should be able to specify their own currency conversion rates. In this case, the values from (1) will be ignored.

If (2) is implemented without (1), then `request.cur` should be required any time `request.prebid.ext.currency.rates` is defined.
---------------------------------------------

For (1), there are some acceptance criteria for general feature performance.

1. Conversion rates are be fetched from `GET http://currency.prebid.org/latest.json`
2. The rates must be fetched before the `/status` endpoint signals that PBS is ready to respond to auctions.
[...]

PR #705, opened 2018-10-05
Introduce currency converter mechanism #280

benjaminch:
This CL introduces a mechanism to fetch currencies rates
periodically from a remote source.
Those rates are stored in `CurrencyConverter` struct which
can be injected in PBS to get rates.

This doesn't introduce any changes in PBS workflow just yet, but in another CL later for clarity.
#280

### DIFF
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -1,0 +1,162 @@
+package currencies
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"sync/atomic"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+// RateConverter holds the currencies conversion rates dictionnary
+type RateConverter struct {
+	httpClient       httpClient
+	done             chan bool
+	updateNotifier   chan<- int
+	fetchingInterval time.Duration
+	syncSourceURL    string
+	rates            atomic.Value // Should only hold Rates struct
+	lastUpdated      atomic.Value // Should only hold time.Time
+}
+
+// NewRateConverter returns a new RateConverter
+func NewRateConverter(
+	httpClient httpClient,
+	syncSourceURL string,
+	fetchingInterval time.Duration,
+) *RateConverter {
+	return NewRateConverterWithNotifier(
+		httpClient,
+		syncSourceURL,
+		fetchingInterval,
+		nil, // no notifier channel specified, won't send any notifications
+	)
+}
+
+// NewRateConverterWithNotifier returns a new RateConverter
+// it allow to pass an update chan in which the number of ticks will be passed after each tick
+// allowing clients to listen on updates
+// Do not use this method
+func NewRateConverterWithNotifier(
+	httpClient httpClient,
+	syncSourceURL string,
+	fetchingInterval time.Duration,
+	updateNotifier chan<- int,
+) *RateConverter {
+	rc := &RateConverter{
+		httpClient:       httpClient,
+		done:             make(chan bool),
+		updateNotifier:   updateNotifier,
+		fetchingInterval: fetchingInterval,
+		syncSourceURL:    syncSourceURL,
+		rates:            atomic.Value{},
+		lastUpdated:      atomic.Value{},
+	}
+
+	// In case host do not want to support currency lookup
+	// we just stop here and do nothing
+	if rc.fetchingInterval == time.Duration(0) {
+		return rc
+	}
+
+	rc.Update()                   // Make sure to populate data before to return the converter
+	go rc.startPeriodicFetching() // Start periodic ticking
+
+	return rc
+}
+
+// fetch allows to retrieve the currencies rates from the syncSourceURL provided
+func (rc *RateConverter) fetch() (*Rates, error) {
+	request, err := http.NewRequest("GET", rc.syncSourceURL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	response, err := rc.httpClient.Do(request)
+	if err != nil {
+		return nil, err
+	}
+
+	defer response.Body.Close()
+
+	bytesJSON, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	updatedRates := &Rates{}
+	err = json.Unmarshal(bytesJSON, updatedRates)
+	if err != nil {
+		return nil, err
+	}
+
+	return updatedRates, err
+}
+
+// Update updates the internal currencies rates from remote sources
+func (rc *RateConverter) Update() error {
+	rates, err := rc.fetch()
+	if err == nil {
+		rc.rates.Store(rates)
+		rc.lastUpdated.Store(time.Now())
+	} else {
+		glog.Errorf("Error updating conversion rates: %v", err)
+	}
+
+	return err
+}
+
+// startPeriodicFetching starts the periodic fetching at the given interval
+// triggers a first fetch when called before the first tick happen in order to initialize currencies rates map
+// returns a chan in which the number of data updates everytime a new update was done
+func (rc *RateConverter) startPeriodicFetching() {
+
+	ticker := time.NewTicker(rc.fetchingInterval)
+	updatesTicksCount := 0
+
+	for {
+		select {
+		case <-ticker.C:
+			// Retries are handled by clients directly.
+			rc.Update()
+			updatesTicksCount++
+			if rc.updateNotifier != nil {
+				rc.updateNotifier <- updatesTicksCount
+			}
+		case <-rc.done:
+			if ticker != nil {
+				ticker.Stop()
+				ticker = nil
+			}
+			return
+		}
+	}
+}
+
+// StopPeriodicFetching stops the periodic fetching while keeping the latest currencies rates map
+func (rc *RateConverter) StopPeriodicFetching() {
+	rc.done <- true
+	close(rc.done)
+}
+
+// LastUpdated returns time when currencies rates were updated
+func (rc *RateConverter) LastUpdated() time.Time {
+	if lastUpdated := rc.lastUpdated.Load(); lastUpdated != nil {
+		return lastUpdated.(time.Time)
+	}
+	return time.Time{}
+}
+
+// Rates returns current rates
+func (rc *RateConverter) Rates() *Rates {
+	if rates := rc.rates.Load(); rates != nil {
+		return rates.(*Rates)
+	}
+	return nil
+}
+
+type httpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
--- a/currencies/rate_converter_test.go
+++ b/currencies/rate_converter_test.go
@@ -1,0 +1,543 @@
+package currencies_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/prebid/prebid-server/currencies"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFetch_Success(t *testing.T) {
+
+	// Setup:
+	calledURLs := []string{}
+	mockedHttpServer := httptest.NewServer(http.HandlerFunc(
+		func(rw http.ResponseWriter, req *http.Request) {
+			calledURLs = append(calledURLs, req.RequestURI)
+			rw.WriteHeader(http.StatusOK)
+			rw.Write([]byte(
+				`{
+					"dataAsOf":"2018-09-12",
+					"conversions":{
+						"USD":{
+							"GBP":0.77208
+						},
+						"GBP":{
+							"USD":1.2952
+						}
+					}
+				}`,
+			))
+		}),
+	)
+
+	defer mockedHttpServer.Close()
+
+	expectedRates := currencies.Rates{
+		DataAsOf: time.Date(2018, time.September, 12, 0, 0, 0, 0, time.UTC),
+		Conversions: map[string]map[string]float64{
+			"USD": {
+				"GBP": 0.77208,
+			},
+			"GBP": {
+				"USD": 1.2952,
+			},
+		},
+	}
+
+	// Execute:
+	rateConverter := currencies.NewRateConverter(
+		&http.Client{},
+		mockedHttpServer.URL,
+		time.Duration(0),
+	)
+	beforeExecution := time.Now()
+	err := rateConverter.Update()
+
+	// Verify:
+	assert.Equal(t, 1, len(calledURLs), "sync URL should have been called %d times but was %d", 1, len(calledURLs))
+	assert.Nil(t, err, "err should be nil")
+	assert.NotEqual(t, rateConverter.LastUpdated(), (time.Time{}), "LastUpdated() should return a time set")
+	assert.True(t, rateConverter.LastUpdated().After(beforeExecution), "LastUpdated() should be after last update")
+	rates := rateConverter.Rates()
+	assert.NotNil(t, rates, "Rates() should not return nil")
+	assert.Equal(t, expectedRates, *rates, "Rates() doesn't return expected rates")
+}
+
+func TestFetch_Fail404(t *testing.T) {
+
+	// Setup:
+	calledURLs := []string{}
+	mockedHttpServer := httptest.NewServer(http.HandlerFunc(
+		func(rw http.ResponseWriter, req *http.Request) {
+			calledURLs = append(calledURLs, req.RequestURI)
+			rw.WriteHeader(http.StatusNotFound)
+		}),
+	)
+
+	defer mockedHttpServer.Close()
+
+	// Execute:
+	rateConverter := currencies.NewRateConverter(
+		&http.Client{},
+		mockedHttpServer.URL,
+		time.Duration(0),
+	)
+	err := rateConverter.Update()
+
+	// Verify:
+	assert.Equal(t, 1, len(calledURLs), "sync URL should have been called %d times but was %d", 1, len(calledURLs))
+	assert.NotNil(t, err, "err shouldn't be nil")
+	assert.Equal(t, rateConverter.LastUpdated(), (time.Time{}), "LastUpdated() shouldn't return a time set")
+	assert.Nil(t, rateConverter.Rates(), "Rates() should return nil")
+}
+
+func TestFetch_FailErrorHttpClient(t *testing.T) {
+
+	// Setup:
+	calledURLs := []string{}
+	mockedHttpServer := httptest.NewServer(http.HandlerFunc(
+		func(rw http.ResponseWriter, req *http.Request) {
+			calledURLs = append(calledURLs, req.RequestURI)
+			rw.WriteHeader(http.StatusBadRequest)
+		}),
+	)
+
+	defer mockedHttpServer.Close()
+
+	// Execute:
+	rateConverter := currencies.NewRateConverter(
+		&http.Client{},
+		mockedHttpServer.URL,
+		time.Duration(0),
+	)
+	err := rateConverter.Update()
+
+	// Verify:
+	assert.Equal(t, 1, len(calledURLs), "sync URL should have been called %d times but was %d", 1, len(calledURLs))
+	assert.NotNil(t, err, "err shouldn't be nil")
+	assert.Equal(t, rateConverter.LastUpdated(), (time.Time{}), "LastUpdated() shouldn't return a time set")
+	assert.Nil(t, rateConverter.Rates(), "Rates() should return nil")
+}
+
+func TestFetch_FailBadSyncURL(t *testing.T) {
+
+	// Setup:
+
+	// Execute:
+	rateConverter := currencies.NewRateConverter(
+		&http.Client{},
+		"justaweirdurl",
+		time.Duration(0),
+	)
+	err := rateConverter.Update()
+
+	// Verify:
+	assert.NotNil(t, err, "err shouldn't be nil")
+	assert.Equal(t, rateConverter.LastUpdated(), (time.Time{}), "LastUpdated() shouldn't return a time set")
+	assert.Nil(t, rateConverter.Rates(), "Rates() should return nil")
+}
+
+func TestFetch_FailBadJSON(t *testing.T) {
+
+	// Setup:
+	calledURLs := []string{}
+	mockedHttpServer := httptest.NewServer(http.HandlerFunc(
+		func(rw http.ResponseWriter, req *http.Request) {
+			calledURLs = append(calledURLs, req.RequestURI)
+			rw.WriteHeader(http.StatusOK)
+			rw.Write([]byte(
+				`{
+					"dataAsOf":"2018-09-12",
+					"conversions":{
+						"USD":{
+							"GBP":0.77208
+						},
+						"GBP":{
+							"USD":1.2952
+						},
+						"badJsonHere"
+					}
+				}`,
+			))
+		}),
+	)
+
+	defer mockedHttpServer.Close()
+
+	// Execute:
+	rateConverter := currencies.NewRateConverter(
+		&http.Client{},
+		mockedHttpServer.URL,
+		time.Duration(0),
+	)
+	err := rateConverter.Update()
+
+	// Verify:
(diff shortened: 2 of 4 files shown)


benjaminch:
#280

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+		if conversion, present := rc.rates.Conversions[from][to]; present == true {
+			return &conversion, nil
+		}
+		return nil, fmt.Errorf("conversion %s->%s not present in rates dictionnary", from, to)
+	}
+	return nil, errors.New("rates are nil")
+}
+
+type httpClient interface {
+	Get(url string) (*http.Response, error)
```
dbemiller: Nice use of interface! I only recently learned how awesome it is that you can define ad-hoc interfaces like this wherever you need them, and I feel like it was a major step forward in my Go development skills :).

One issue I see in this PR is that there's nothing cancelling the request if `currency.prebid.org` hangs for some reason. This also affects  your `TODO: Handle retry strategy`, since you'll want to cancel the first request when you do a retry.

All that said: I think `Do(req *Request) (*Response, error)` might make a better interface function here. You need an `http.Request` if you want to [attach Contexts](https://golang.org/pkg/net/http/#Request.WithContext), which you'll need to cancel requests before you retry them.

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, err
+	}
+	rc.updateRates(rates)
+
+	return updatesNotifier, nil
+}
+
+// StopPeriodicFetching stops the periodic fetching while keeping the latest currencies rates map
+func (rc *RateConverter) StopPeriodicFetching() {
+	rc.done <- true
```
dbemiller: What's the purpose for this?

Seems like we'd start fetching rates when the app starts, and continue until the server shuts down.

Do you plan to use this to for graceful shutdown? If so, you'd need another channel which blocks until the goroutine from `StartPeriodicFetching` totally returns.

currencies/rates.go
```diff
@@ -0,0 +1,24 @@
...
+// note that `DataAsOfRaw` field is needed when parsing remote JSON as the date format if not standard and requires
+// custom parsing to be properly set as Golang time.Time
+type Rates struct {
+	DataAsOfRaw string                        `json:"dataAsOf"`
+	DataAsOf    *time.Time                    `json:"-"`
+	Conversions map[string]map[string]float32 `json:"conversions"`
+}
+
+// ComputeDataAsOf computes `DataAsOf` by parsing `DataAsOfRaw` which is a raw string representing a date
+func (c *Rates) ComputeDataAsOf() error {
```
dbemiller: +1 to computing the Time so people don't have to think about the string.
-1 to making callers run `ComputeDataAsOf()` every time after they Unmarshal a response.

You could just implement the [Unmarshaler](https://golang.org/pkg/encoding/json/#Unmarshaler) and do this work inside there. As a bonus it'd also let you remove `DataAsOfRaw`, which doesn't really have much value once it's been parsed.

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Rates returns current rates
+func (rc *RateConverter) Rates() *Rates {
+	return rc.rates
+}
+
+// GetRate returns the conversion rate between two currencies
+// returns an error in case the conversion rate between the two given currencies is not in the currencies rates map
+func (rc *RateConverter) GetRate(from string, to string) (*float32, error) {
```
dbemiller: Why a pointer return here?

currencies/currency_converter_test.go
```diff
@@ -0,0 +1,557 @@
...
+	// Verify:
+	if !mockedHttpClient.VerifyUrlCalled(requestURL) {
+		t.Errorf("sync URL `%s` wasn't called", requestURL)
+		return
+	}
+	if !mockedHttpClient.VerifyHasNoErrors() {
+		t.Error("expects no error while fecthing remote source")
+		return
+	}
+	if err != nil {
```
dbemiller: Check out Testify: https://godoc.org/github.com/stretchr/testify/assert

It's already a project dependency, and makes tests so much cleaner.

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+func (rc *RateConverter) fetch() (*Rates, error) {
+	response, err := rc.httpClient.Get(rc.syncSourceURL)
+	if err != nil {
+		return nil, err
+	}
+
+	defer response.Body.Close()
+
+	updatedRates := &Rates{}
+	err = json.NewDecoder(response.Body).Decode(&updatedRates)
```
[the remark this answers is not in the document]
benjaminch: Oh ! I didn't know that, good to know !

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+func (rc *RateConverter) Rates() *Rates {
+	return rc.rates
+}
+
+// GetRate returns the conversion rate between two currencies
+// returns an error in case the conversion rate between the two given currencies is not in the currencies rates map
+func (rc *RateConverter) GetRate(from string, to string) (*float32, error) {
+	if rc.rates != nil {
+		rc.RLock()
+		defer rc.RUnlock()
```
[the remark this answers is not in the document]
benjaminch: Yes, indeed !

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// StopPeriodicFetching stops the periodic fetching while keeping the latest currencies rates map
+func (rc *RateConverter) StopPeriodicFetching() {
+	rc.done <- true
+}
+
+// LastUpdated returns time when currencies rates were updated
+func (rc *RateConverter) LastUpdated() *time.Time {
+	return rc.lastUpdated
```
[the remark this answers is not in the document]
benjaminch: Good point ! Let's see how it looks like. 
Is there any issue to have those if it includes lock mechanism though ? (via `atomic.value`)

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+		rc.lastUpdated = &now
+	}
+
+	return rates, err
+}
+
+// StartPeriodicFetching starts the periodic fetching at the given interval
+// triggers a first fetch when called before the first tick happen in order to initialize currencies rates map
+// returns a chan in which the number of data updates everytime a new update was done
+func (rc *RateConverter) StartPeriodicFetching() (chan int, error) {
```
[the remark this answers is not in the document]
benjaminch: Yes, why not. 
It means that if we remove this Start function, then everytime we stop the fetching, we will need to recreate a new converter. Maybe it's ok ?

currencies/rate_converter.go
benjaminch: I'll have a look yes, good point.

currencies/rate_converter.go
benjaminch: We might not need it indeed. From your previous comment, If we don't need the start, we might not need the Stop.
But I still think that for convenience, we should still return a chan in which we send ticks. So if the fetching starts directly on object construction, then, we can return this chan in `rateConverter, tickerChan, err := NewRateConverter(...)` what do you think ?

Also, if we want to align on the `time.ticker` implementation, the later does have the `Stop()` function, so maybe good to keep.

currencies/rates.go
benjaminch: Totally agree, I just forgot to get back to this implementing the `UnMarshaller`.

currencies/rate_converter.go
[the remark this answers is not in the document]
hhhjort: I might suggest returning immediately if `rc.fetchingInterval` is zero. We definitely should handle the zero case, as there likely will be some hosts that do not want to support currency lookup, and would rather not waste the cycles and bandwidth fetching the data.

currencies/rate_converter.go
benjaminch: I wanted to be able to differenciate not found rate without having to pass a 0.
Maybe I can change the signature to remove the pointer:

`func (rc *RateConverter) GetRate(from string, to string) (float32, bool, error)`

currencies/currency_converter_test.go
benjaminch: yes, thanks :)

currencies/rate_converter.go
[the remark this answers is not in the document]
benjaminch: Good point !

currencies/rate_converter.go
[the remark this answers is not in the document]
dbemiller: I think you'd have to either use two `atomic.Value` props, or roll them all into the same struct.

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// NewRateConverter returns a new RateConverter
+func NewRateConverter(
+	httpClient httpClient,
+	syncSourceURL string,
+	fetchingInterval time.Duration,
+	retryFetchingInterval time.Duration,
+) *RateConverter {
+	return &RateConverter{
```
hhhjort: We should fetch the initial set of currency data before returning this object. That way we know the object is ready to use before any callers try using it.

benjaminch: yes, agree

currencies/rate_converter.go
dbemiller: I don't have a strong opinion... but I'd lean towards leaving it in and fixing it. Graceful shutdowns are nice to have. For this use-case, though, I'd call them very low priority.

If you keep it, my main concern is that it returns "too early" to be used as a graceful shutdown function. For example: imagine it gets called while the `go func()` inside `StartPeriodicFetching` is in the middle of making an HTTP call/rate update.

This function would send a signal on `rc.done`... but it will return before the `go func()` ever finishes its call and processes the "done" signal. So there's no real way for the caller to know when the shutdown has truly finished.

> So if the fetching starts directly on object construction, then, we can return this chan in rateConverter, tickerChan, err := NewRateConverter(...) what do you think ?

That's an option... my main concern is that channels have a limited buffer space. If callers don't read the values from that `tickerChan`, it will eventually fill up and block future updates from happening.

Is there a use-case for "listening" on updates? One option is to make callers pass in the channel:

```
func NewRateConverter(func NewRateConverter(
  httpClient httpClient,
  syncSourceURL string,
  updater <-chan {any},
  ...
) *RateConverter
```

Callers could just send in a `NewTicker().C` for `updater`, and it would replace your current `fetchingInterval`. If the caller also wants to be alerted of updates, they could [fan out](https://blog.golang.org/pipelines) the events from a ticker, pass one into here, and listen to the other themselves.

currencies/rate_converter.go
dbemiller: aha, I see.

Couldn't you use the `error` for that, though? If error is non-nil, the rate wasn't found. If the error is nil and rate is 0, then it's _really_ 0.

...in practice "real" 0 rates should be impossible, though.

benjaminch: Yes I can, I was just thinking that errors and found / not found are not exactly the same in this context, but we can try.

currencies/rate_converter.go
dbemiller: very good catch!

currencies/rate_converter.go
benjaminch: Done, I allowed clients to pass a chan in which they will be listening notifications.
There are now two functions `NewRateConverter` and `NewRateConverterWithNotifier` to cronstruct new `RateConverter`. By default, `NewRateConverter` doesn't track updates whereas `NewRateConverterWithNotifier` will, so clients can pick the one they want and will also avoid having a forgotten chan provided by the client.

currencies/rate_converter.go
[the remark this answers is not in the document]
benjaminch: Done

currencies/rate_converter.go
```diff
@@ -0,0 +1,165 @@
...
+	sync.RWMutex
+	httpClient            httpClient
+	updates               chan *Rates
+	done                  chan bool
+	ticker                *time.Ticker
+	fetchingInterval      time.Duration
+	retryFetchingInterval time.Duration
+	syncSourceURL         string
+	rates                 *Rates
+	lastUpdated           *time.Time
```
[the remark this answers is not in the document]
benjaminch: Done

currencies/rate_converter.go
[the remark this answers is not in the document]
benjaminch: Done

currencies/rate_converter.go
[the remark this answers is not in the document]
benjaminch: I put each of those variables within an atomic.Value. Let me know if you want me to remove those two getters functions or not.

currencies/rate_converter.go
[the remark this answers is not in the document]
benjaminch: Done

currencies/rate_converter.go
benjaminch: Done

currencies/rate_converter.go
benjaminch: Done, I allowed clients to pass a chan in which they will be listening notifications.
There are now two functions NewRateConverter and NewRateConverterWithNotifier to cronstruct new RateConverter. By default, NewRateConverter doesn't track updates whereas NewRateConverterWithNotifier will, so clients can pick the one they want and will also avoid having a forgotten chan provided by the client.

currencies/rates.go
benjaminch: Done

currencies/rate_converter.go
benjaminch: Done

currencies/rate_converter.go
benjaminch: Done

benjaminch:
Hey @dbemiller & @hhhjort, sorry for the delay. Just updated the PR taking into account your feedbacks, let me know if it looks better for you :)

Thanks

currencies/rate_converter.go
```diff
@@ -0,0 +1,205 @@
...
+
+	defer response.Body.Close()
+
+	bytesJSON, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	updatedRates := &Rates{}
+	err = json.Unmarshal(bytesJSON, &updatedRates)
```
hhhjort: `updatedRates` is already a pointer, you don't want a pointer to a pointer

currencies/rate_converter.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// updateRates allows to update the internal currencies rates map with the one provided
+func (rc *RateConverter) updateRates(updatedRates *Rates) {
+	rc.rates.Store(updatedRates)
+	rc.lastUpdated.Store(time.Now())
+}
+
+// Update updates the internal currencies rates from remote sources
+func (rc *RateConverter) Update() (*Rates, error) {
```
hhhjort: Is the `*Rates` return value used for anything? For that matter do you ever try checking the `error` state when you call it?

currencies/rate_converter.go
```diff
@@ -0,0 +1,205 @@
...
+		lastUpdated:           atomic.Value{},
+	}
+
+	// In case host do not want to support currency lookup
+	// we just stop here and do nothing
+	if rc.fetchingInterval == time.Duration(0) {
+		return rc
+	}
+
+	rc.Update()                // Make sure to populate data before to return the converter
```
hhhjort: You should probably at least log an error if the `Update()` returns an error.

currencies/rate_converter.go
```diff
@@ -0,0 +1,205 @@
...
+// returns a chan in which the number of data updates everytime a new update was done
+func (rc *RateConverter) startPeriodicFetching() error {
+
+	rc.ticker = time.NewTicker(rc.fetchingInterval)
+	updatesTicksCount := 0
+
+	go func() {
+		for {
+			select {
+			case <-rc.ticker.C:
```
hhhjort: Would it make sense to call `rc.Update()` here rather than spawning off two go routines to handle one update? Do these even need to be handled in go routines? All that it would allow is multiple updates to happen in parallel if necessary, and I don't see a use case for that.

I would suggest just calling `rc.Update()` on a tick, so you can have all your update code in one spot. That will allow the initial `Update()` call benefit from any retry logic implemented in the future. If a new update tick comes along while one update is still processing, it doesn't seem useful to spawn a second update attempt. The context created within the Update code will handle the case where one request hangs, so it will eventually be retried.

currencies/rate_converter.go
```diff
@@ -0,0 +1,205 @@
...
+func (rc *RateConverter) Rates() *Rates {
+	if rates := rc.rates.Load(); rates != nil {
+		return rates.(*Rates)
+	}
+	return nil
+}
+
+// GetRate returns the conversion rate between two currencies
+// returns an error in case the conversion rate between the two given currencies is not in the currencies rates map
+func (rc *RateConverter) GetRate(from string, to string) (float64, error) {
```
hhhjort: I am thinking we may want to hang this function on the `Rates` struct rather than the `RateConverter` object. This way, an auction can call to get the `Rates` once, at the expense of one atomic transaction, rather than invoking the atomic machinery for each conversion call. This would also avoid the race condition where the exchange rates could change half way through an auction, not that this would be a serious bug.

currencies/rate_converter.go
benjaminch: Oh yes, fixed !

currencies/rate_converter.go
benjaminch: I don't use `*Rates` but in tests. I can totally remove it you are right. 

Regarding error, I added error logging / check in `Update()` function directly. I also still think returning the error is not bad, it's used to test the function and it also allows to be able to know if there was an error or not in the calling function (clients might want to handle the error or do something about it).

currencies/rate_converter.go
benjaminch: Yes, done !

currencies/rate_converter.go
benjaminch: Indeed, it would be way simpler; I updated it with what you suggested, let me know if I did it right.

currencies/rate_converter.go
benjaminch: Good point, just move this function to `Rates` struct.

benjaminch:
Should be better now, let me know what you think :)

currencies/rate_converter.go
```diff
@@ -0,0 +1,178 @@
...
+	return err
+}
+
+// startPeriodicFetching starts the periodic fetching at the given interval
+// triggers a first fetch when called before the first tick happen in order to initialize currencies rates map
+// returns a chan in which the number of data updates everytime a new update was done
+func (rc *RateConverter) startPeriodicFetching() error {
+
+	rc.ticker = time.NewTicker(rc.fetchingInterval)
+	updatesTicksCount := 0
```
dbemiller: There are some race conditions in here... `rc.ticker` is written in this scope, but read from inside the goroutine. In general, shared memory is dangerous.

I think the `ticker` should be a local variable. By saving it on `rc`, race conditions are very likely. `updateTicksCount` should be confined to the goroutine where it's used as well.

A few other suggestions:

1. You're always returning `nil`.... so the return value is useless. You may as well remove it.
2. In idiomatic Go, synchronous methods are the "default." I'm no expert... but I think this function should block forever, and you should run it with `go rc.startPeriodicFetching()`. I'd be curious if @zachbadgett had some insight here.

currencies/rate_converter_test.go
```diff
@@ -0,0 +1,477 @@
...
+	"net/http"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/prebid/prebid-server/currencies"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFetch_Success(t *testing.T) {
```
dbemiller: I've seen a few race conditions already... but they're notoriously hard to pick out by eye. I trust good automated tests much more than myself.

Please add a test which exercises the concurrency. Name it something like `TestRace*()`, and spin up some goroutines which do a bunch of reads & updates in parallel. The CI runs tests which start with `TestRace` under the `-race` flag, so this would really help prevent prod bugs. Race conditions crash the entire process, so... it's extremely important that we catch these.

currencies/rate_converter.go
zachbadgett: @dbemiller is exactly right. Another thing about this goroutine is that it will cause a CPU spike due to the `default` keyword. The goroutine will no longer "park"/yield to the scheduler. This function should look more like this:
```go
func (rc *RateConverter) startPeriodicFetching(fetchingInterval time.Duration) {
	t := time.NewTicker(fetchingInterval)
	var updatesTicksCount int
	for {
		select {
		case <-t.C:
			// TODO: Handle retry strategy
			rc.Update()
			updatesTicksCount++
			if rc.updateNotifier != nil {
				rc.updateNotifier <- updatesTicksCount
			}
		case <-rc.done:
			t.Stop()
			return
		}
	}
}
```

dbemiller: oh man... good catch! I didn't even see that.

Thanks for taking a look!

currencies/rate_converter.go
zachbadgett: If `StopPeriodicFetching` is kept this should be changed to do to a close to ensure all goroutines are broadcasted to.
```go
close(rc.done)
```
Even if there is currently only one goroutine it will help protect future work if more are added or if `startPeriodicFetching` is called more than once.

currencies/rate_converter_test.go
```diff
@@ -0,0 +1,477 @@
...
+		time.Duration(5)*time.Minute,
+	)
+	defer rateConverter.StopPeriodicFetching()
+	rates := rateConverter.Rates()
+
+	// Verify:
+	assert.Nil(t, rates, "rates should be nil")
+}
+
+type HttpClientMock struct {
```
zachbadgett: Any reason to not use https://golang.org/pkg/net/http/httptest/ ?

currencies/rate_converter.go
benjaminch: Oh I didn't know about that, thanks guys, it's way simpler :)

currencies/rate_converter.go
```diff
@@ -0,0 +1,178 @@
...
+
+	go func() {
+		for {
+			select {
+			case <-rc.ticker.C:
+				// TODO: Handle retry strategy
```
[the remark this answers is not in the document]
benjaminch: Yes, indeed it's a very good idea :)

currencies/rates.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Rates holds data as represented on http://currency.prebid.org/latest.json
+// note that `DataAsOfRaw` field is needed when parsing remote JSON as the date format if not standard and requires
+// custom parsing to be properly set as Golang time.Time
+type Rates struct {
+	DataAsOf    time.Time                     `json:"dataAsOf"`
+	Conversions map[string]map[string]float64 `json:"conversions"`
+}
+
+type Time struct {
```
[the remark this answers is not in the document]
benjaminch: Oops, yes indeed

currencies/rate_converter.go
hhhjort: Looks good to me. :)

currencies/rate_converter_test.go
benjaminch: Yes !

benjaminch: Yep, good point

currencies/rate_converter_test.go
benjaminch: No, not really, I will replace this mock with the httptest package

currencies/rate_converter_test.go
benjaminch: Done, 50 workers running `Updates` and `GetRate`:)

currencies/rate_converter.go
benjaminch: Done

currencies/rate_converter.go
benjaminch: Good point :) 
Done

currencies/rate_converter.go
benjaminch: Thanks @hhhjort you are right, I just changed it, thanks

PR #711, opened 2018-10-10
Formalizing support for `site.ext.amp`

### LINKED ISSUE #708: Pass AMP signal in ORTB request to adapters
Currently PBS doesn't flag requests as coming from AMP-RTC vs. standard Prebid Server. This issue is to track adding an AMP=true or AMP=1 type of signal to the ORTB request when it's sourced from an AMP PBS stored request.

Since `site.amp` [does not exist yet](https://github.com/ampproject/amphtml/blob/master/ads/google/a4a/docs/RTBExchangeGuide.md) in ORTB 2.5 we most likely should use an `ext` field so that each adapter can track whether the request originated from AMP or not.

dbemiller:
This fixes #708

Technically the `ext` was already going to the bidders... so this just adds input validation & documents it.

Along the way, I found that our AMP tests were testing lots of payloads with `request.app` defined. Since AMP implies `site`, these tests don't make sense. I added an explicit check for `request.app` in the code.

This seems like a breaking change... but I don't think it is, because a _real_ AMP page would send the `curl` query param, and there was already code which defined `site` in that case. If the Stored Request already defined `app`, then it would cause both to be defined, which would make the request invalid anyway.

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -681,7 +681,11 @@ func (deps *endpointDeps) validateAliases(aliases map[string]string) error {
 }
 
 func (deps *endpointDeps) validateSite(site *openrtb.Site) error {
-	if site != nil && site.ID == "" && site.Page == "" {
+	if site == nil {
+		return nil
+	}
+
+	if site.ID == "" && site.Page == "" {
 		return errors.New("request.site should include at least one of request.site.id or request.site.page.")
 	}
 	if len(site.Ext) > 0 {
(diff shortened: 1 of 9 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -787,6 +797,21 @@ func setSiteImplicitly(httpReq *http.Request, bidReq *openrtb.BidRequest) {
 			}
 		}
 	}
+	if bidReq.Site != nil {
+		setAmpExt(bidReq.Site, "0")
+	}
+}
+
+func setAmpExt(site *openrtb.Site, value string) {
```
hhhjort: Any particular reason you put this in auction.go rather than amp_auction.go?

dbemiller: No reason... but that makes sense. I'll move it.

It's used in both files... started out named something else, but with "amp" in the name I can definitely see how it fits better in the `amp_auction` file.

Issue #657, opened 2018-08-21
IndexExchange vs ix


bretg:
For PBJS 1.0, IndexExchange changed the name of their PBJS adapter to `ix`.

This is causing compatibility issues with calling PBS via the prebidServerBidAdapter. Two specific issues:

1) 'ix' isn't recognized on the Prebid Server side
2) the `siteId` parameter is defined on the client side, but the server expects `siteID`

dbemiller:
A few related tickets redirect here... so linking these comments for visibility:
- https://github.com/prebid/Prebid.js/issues/3044#issuecomment-422952684
- https://github.com/prebid/Prebid.js/pull/2916#issuecomment-426264997

I've heard nothing about this personally... but technically, according to the docs we don't even support 0.x anymore: http://prebid.org/dev-docs/bidders.html

At this point it should be safe to just make the breaking change and rename the Index adapter to `ix`.

akira28:
Seems that they are working on it on their fork: https://github.com/indexexchange/prebid-server/commit/aeb02bb78714f297ffc027174afb2a30a220e922

dbemiller:
This has been merged & should be resolved now.

PR #712, opened 2018-10-10
Rename indexExchange to ix (#657)

ix-certification:
- Fixes the issues addressed in #657.
- Rename package from `indexExchange` to `ix`
- Fixes `siteId` being incorrect type, Prebid.js sends as a string.
- We do not currently support video for Prebid Server.
- Added default endpoint for ix.
- Response will match ad unit code based on bid id.


dbemiller:
@ix-certification This will break far too many things :/.

Prebid Mobile and AMP use Stored Requests. Every PBS host company has their own data, and all that data will still use `indexExchange`. If you just delete it, suddenly `indexExchange` will be an unrecognized Bidder, and everyone's pages will break.

We could deprecate it, and delete it in the distant future (after everyone host PBS has migrated their DB data)... but I don't think it's fair to delete it suddenly and with no window for migration.

dbemiller:
One thing @mkendall07 did discover this morning: according to our metrics, `indexExchange` returns _no_ valid bids at all, to anyone.

So another option is that we could relax the input validation. Today, PBS returns a 4xx if you use an "unknown" bidder. It was designed that way to catch dumb typos or missing aliases.... but if we're ok losing that, we could run the other bidders and return a `req.ext` error describing the wrong bidder instead. That change would be backwards compatible, and once we made them, these changes wouldn't really "break" anything, since `indexExchange` isn't bidding today anyway.

hhhjort:
After some offline discussion, it seems dbemiller's last idea is the best. With a slight modification, we will relax the constraint to not 4xx any "deprecated" bidders, which will include the old name for the index exchange bidder, while leaving the general case intact (truely unknown bidders will still cause a 4xx). Once that PR goes in, we can merge this if no other issues come up in review.

hhhjort:
Submitted PR #716 in support of this PR, so existing requests won't break when the name change happens.

adapters/ix/ix.go
```diff
@@ -17,31 +17,31 @@ import (
...
+type IxAdapter struct {
 	http *adapters.HTTPAdapter
 	URI  string
 }
 
 // used for cookies and such
-func (a *IndexAdapter) Name() string {
-	return "indexExchange"
+func (a *IxAdapter) Name() string {
+	return "ix"
```
dbemiller: Should return the constant, `openrtb_ext.BidderIx` here.

adapters/ix/ix.go
```diff
@@ -136,38 +136,40 @@ func (a *IndexAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder *pb
...
+						BidID:             bidID,
+						AdUnitCode:        adunit.Code,
+						BidderCode:        bidder.BidderCode,
+						Price:             bid.Price,
+						Adm:               bid.AdM,
+						Creative_id:       bid.CrID,
+						Width:             bid.W,
+						Height:            bid.H,
+						DealId:            bid.DealID,
+						CreativeMediaType: "banner",
```
dbemiller: Should use the `openrtb_ext.BidTypeBanner` constant here.

Issue #713, opened 2018-10-11
Bug with request.imp.ext validation


hhhjort:
In /endpoints/aution.go:
```
	if len(bidderExts) < 1 {
		errL = append(errL, fmt.Errorf("request.imp[%d].ext must contain at least one bidder", impIndex))
		return errL
	}
```
If the `imp.ext` contains a `prebid` section, but no actual bidders we won't catch the lack of bidders condition. Fixing this will be a breaking change, as that will cause requests which currently work (multiple imp requests with one "empty" imp) to return 4xx errors. But otherwise it is a simple fix.

PR #716, opened 2018-10-12
Enables proper processing for deprecated (disabled) bidders.

hhhjort:
This will allow us to not 4xx on receiving params for disabled/deprecated bidders. This will allow #657 to go through without causing current requests to start returning 4xx errors.

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -660,7 +660,7 @@ func (deps *endpointDeps) validateImpExt(imp *openrtb.Imp, aliases map[string]st
 				}
 			} else {
 				if msg, isDisabled := deps.disabledBidders[bidder]; isDisabled {
-					errL = append(errL, &errortypes.NonFatalValidation{Message: msg})
+					errL = append(errL, &errortypes.BidderTemporarilyDisabled{Message: msg})
 					disabledBidders = append(disabledBidders, bidder)
 				} else {
 					return []error{fmt.Errorf("request.imp[%d].ext contains unknown bidder: %s. Did you forget an alias in request.ext.prebid.aliases?", impIndex, bidder)}
@@ -681,7 +681,7 @@ func (deps *endpointDeps) validateImpExt(imp *openrtb.Imp, aliases map[string]st
 		imp.Ext = extJSON
 	}
 
-	// TODO: Fix issue #713 here
+	// TODO #713 Fix this here
 	if len(bidderExts) < 1 {
 		errL = append(errL, fmt.Errorf("request.imp[%d].ext must contain at least one bidder", impIndex))
 		return errL
@@ -1001,7 +1001,7 @@ func writeError(errs []error, w http.ResponseWriter) bool {
 // Checks to see if an error in an error list is a fatal error
 func fatalError(errL []error) bool {
 	for _, err := range errL {
-		if _, ok := err.(*errortypes.NonFatalValidation); !ok {
+		if errortypes.DecodeError(err) != errortypes.BidderTemporarilyDisabledCode {
 			return true
 		}
 	}
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -22,6 +22,7 @@ import (
 	nativeRequests "github.com/mxmCherry/openrtb/native/request"
 	"github.com/prebid/prebid-server/analytics"
 	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/exchange"
 	"github.com/prebid/prebid-server/openrtb_ext"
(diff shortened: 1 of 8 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -964,3 +997,13 @@ func writeError(errs []error, w http.ResponseWriter) bool {
 	}
 	return false
 }
+
+// Checks to see if an error in an error list is a fatal error
+func fatalError(errL []error) bool {
+	for _, err := range errL {
+		if _, ok := err.(*errortypes.NonFatalValidation); !ok {
```
dbemiller: I stumbled on this interesting blog post on error handling a little while back: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

See the sections titled "Problems with error types", "Conclusion: avoid error types", and "Assert errors for behaviour, not type"

errortypes/errortypes.go
```diff
@@ -88,6 +89,21 @@ func (err *FailedToRequestBids) Code() int {
 	return FailedToRequestBidsCode
 }
 
+// NonFatalValidation is used at the request validation step, where we want to continue processing as best we
+// can rather than returning a 4xx, and still return an error message.
+// The initial usecase is to flag deprecated bidders.
+type NonFatalValidation struct {
```
dbemiller: I think this name is too vague... for example: `BadInput` errors are also a type of "non-fatal validation" errors, from the caller's point of view.

I think this should be something more specific, like `BidderTemporarilyDisabled`. Also check out the error-handling blog post from my other comment. Using that strategy, you can make this specialized and still have validation code that extends well when other non-fatal cases come up.

endpoints/openrtb2/sample-requests/invalid-whole/unknown-bidder.json
```diff
@@ -0,0 +1,38 @@
+{
+    "description": "Copy of the prebid test ad, with the addition of an unknown bidder",
+  
+    "message": "Invalid request: request.imp[0].ext contains unknown bidder: unknownbidder. Did you forget an alias in request.ext.prebid.aliases?\n",
```
dbemiller: This test looks like it overlaps with `imp-ext-unknown-bidder.json`... did you mean to hit another edge case here?

endpoints/openrtb2/auction.go
```diff
@@ -646,14 +656,37 @@ func (deps *endpointDeps) validateImpExt(ext json.RawMessage, aliases map[string
...
+			delete(bidderExts, bidder)
+		}
+		extJSON, err := json.Marshal(bidderExts)
+		if err != nil {
+			return []error{err}
+		}
+		imp.Ext = extJSON
+	}
+
+	// TODO: Fix issue #713 here
```
dbemiller: By convention: `// TODO #XYZ: {message}`.

That way we can regex search for `// TODO #{digit}+` to find them all.

endpoints/openrtb2/auction.go
```diff
@@ -224,57 +226,66 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
 
 	for index, imp := range req.Imp {
-		if err := deps.validateImp(&imp, aliases, index); err != nil {
-			return err
+		errs := deps.validateImp(&imp, aliases, index)
+		if len(errs) > 0 {
+			errL = append(errL, errs...)
+		}
+		if fatalError(errs) {
+			return errL
```
dbemiller: Any reason to return here?

If it's not too much trouble, it seems like it'd be better to keep running validation so we can return all the errors at once.

hhhjort: Old behavior was to stop processing on the first error, I was trying to preserve that behavior.

dbemiller: ahh ok. no worries then

endpoints/openrtb2/auction.go
hhhjort: Done

endpoints/openrtb2/auction.go
hhhjort: Hah, we happen to be 90% there already ... well I'll do it by importing `errortypes`, but if it ever becomes an issue we can copy a few lines from there and break the dependence `errortypes`

endpoints/openrtb2/sample-requests/invalid-whole/unknown-bidder.json
hhhjort: This test fails when not using the correct default alias. This test passes when you do use the correct alias. It overlaps, because I needed one valid bidder to exist so this test will still pass when #713 is fixed.

I suppose I could have just modified `imp-ext-unknown-bidder.json`, but this way we will explicitly test the case where both valid and invalid bidders are provided. It is a little counterintuitive that we use a bad request to test that in a certain case the request is valid. :)

PR #718, opened 2018-10-16
Default aliases

hhhjort:
This implements a configurable default set of aliases to be known by PBS. The aliases are handled similar to stored requests, with the aliases merged into the request, but any aliases defined in the request, raw or stored, will overwrite the default aliases. Also includes a flag that will add the aliases to the bidder info endpoints, as a way to advertise what bidders are available on the server by default.


docs/developers/default-aliases.md
```diff
@@ -0,0 +1,30 @@
+# Server Based Default Aliases
+
+Some aliases are common, so it makes some sense to predefine them on the server, rather than adding them to all incoming requests (or all stored requests). We can define the aliases in a file to be read at server startup and injected into all requests.
```
dbemiller: So... I'm a bit puzzled by some of the design decisions here. Broadly speaking, I have two independent concerns:

------------------------------------

For features like these, there are two key questions to ask:

1. Should the values vary from deployment to deployment?
2. Will we need to change these values on-the-fly, without doing a release?

Based on these, we can categorize the type of values we're working with:

If both are "no", it's a **Constant**.
If only (1) is "yes", it's **Configuration**.
If both are "yes," it's **Data**. (I don't know if there's a standard name for this case... but it's what you'd put in a database, or fetch from another Service via HTTP, or whatever) 

**Constants** can be hardcoded... but I think everyone agrees that this feature doesn't fit that category.

**Configuration** in modern apps is done via environment variables. See [the 12 factor app docs](https://12factor.net/config).

We never discussed whether this should be **Data**... but I can see arguments in either direction. It doesn't matter much to me either way.

What I do think does matter, though, is that we use the same tools to accomplish the same goals.

If Config: this feature shouldn't force people to read config from files. Even though environment variables are ideal, we do use Viper to accommodate legacy infrastructures because PBS is open-source. If these values are config, they should come straight through Viper too.

If Data: Consider that people will want to use different backends, just like they do with Stored Requests. It's fine if "Files" are the only backend you want to implement now... but let's choose config values which can scale to other backends without breaking changes. Alternatively: you could probably extend today's Stored Request backends to add support here.

------------------------------------------------------------------

The other issue on my mind is Parsimony (see [this link](https://engineering.squarespace.com/blog/2018/six-attributes-of-beautiful-systems)). We have a few other use-cases for "global" Request values today.

1. `request.at` (signals first-price or second-price auction. In header bidding, "first price" makes a good default).
2. `request.tmax` (currently `auction_timeouts_ms.default` in the Viper config)
3. `regs.ext.gdpr` (currently `gdpr.usersync_if_ambiguous` in the Viper config)

All these features, as well as this new one you're adding, could be the same feature if you just merge the JSON in at the top level.

I'm not suggesting you tackle all that in this PR... but if you agree, it would make sense to implement this feature as a top-level Request merge. Then later we can deprecate/delete the other config flags, which will make PBS overall more Parsimonious.

endpoints/info/bidders_test.go
```diff
@@ -128,9 +143,83 @@ func TestGetBidderAccuracy(t *testing.T) {
...
+	params := []httprouter.Param{{
+		Key:   "bidderName",
+		Value: bidder,
+	}}
+
+	r := httptest.NewRecorder()
+	endpoint(r, req, params)
+
+	var fileData adapters.BidderInfo
+	if err := json.Unmarshal(r.Body.Bytes(), &fileData); err != nil {
```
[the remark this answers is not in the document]
hhhjort: Odd it showing this all as additions, where all this test code was actually untouched. We can look into converting it to testify though.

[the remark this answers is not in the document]
dbemiller: ahh ok. If that's the case don't worry about it then. There's plenty of test code we _could_ convert, if we wanted to :).

I'd only be concerned if the problem were growing.

router/router.go
```diff
@@ -261,3 +263,41 @@ func SupportCORS(handler http.Handler) http.Handler {
...
+func readDefaultRequest(defReqConfig config.DefReqConfig) (map[string]string, []byte) {
+	defReq := &defReq{}
+	aliases := make(map[string]string)
+	if defReqConfig.Type == "file" {
+		if len(defReqConfig.FileSystem.FileName) == 0 {
+			return aliases, []byte{}
+		}
+		defReqJSON, err := ioutil.ReadFile(defReqConfig.FileSystem.FileName)
+		if err != nil {
+			glog.Errorf("error reading aliases from file %s: %v", defReqConfig.FileSystem.FileName, err)
```
dbemiller: Should fail fast & crash on invalid configs, for easier debugging.

Here are some "error" cases:

1. `type` is something besides `file` or empty string. (prevents typos like "fild")
2. `type` is `file`, but `filesystem.filename` is empty/undefined.
3. `type` is `file`, but the `filesystem.filename` doesn't exist on the filesystem.

adapters/info.go
```diff
@@ -166,6 +166,7 @@ func (infos BidderInfos) SupportsWebMediaType(bidder openrtb_ext.BidderName, med
 type BidderInfo struct {
 	Maintainer   *MaintainerInfo   `yaml:"maintainer" json:"maintainer"`
 	Capabilities *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
+	AliasOf      string            `json:"alias_of,omitempty"`
```
dbemiller: most everyone uses camel-case... including this same endpoint (with `mediaType`). Let's keep it consistent and use `aliasOf`.

docs/developers/default-request.md
```diff
@@ -0,0 +1,44 @@
...
+This allows a defaut stored request to be defined that allows the server to set up some defaults for all incoming requests. A request specified stored request will override these defaults, and of course any options specified directly in the stored request override both. The default stored request is only read on server startup, it is meant as an installation static default rather than a dynamic tuning option.
+
+A common use case is to "hard code" aliases into the server. This saves having to specify them on all incoming requests, and/or on all stored requests. To help support automation and alias discovery we can flag that any aliases found in the file be added to the bidder info endpoints.
+
+## Config Options
+
+Three config options are exposed to support this feature.
+```
+default_request:
+    type: "file"
```
dbemiller: little inconsistent... do you think `file` should be `filesystem` here, to keep it consistent with the key name?

Could also leave `type` as is, but rename `filesystem` to `file`, and `filename` to `name`.

docs/developers/default-request.md
```diff
@@ -0,0 +1,44 @@
...
+                "districtm": "appnexus"
+            }
+        }
+    }
+}
+```
+This will be JSON merged into the `ext` element of the incomming requests to define common aliases for the server.
+
+The `info` option determines if the alised bidders will be exposed on the `/info` endpoints. If true the alias name will be added to the list returned by
+`/info/bidders` and the info JSON for the core bidder will be coppied into `/info/bidder/{biddername}` with the addition of the field
```
dbemiller: Any thoughts on `/bidders/params` here?

If the goal is to make aliases seem just like real bidders when this flag is flipped, it seems to me like they should appear there too... but I'm not totally sold on it.

endpoints/info/bidders_test.go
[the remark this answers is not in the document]
dbemiller: Oh, don't worry about converting old code. I didn't realize.

It's showing as additions because... well, they're additions :). Maybe you thought you renamed `TestGetBidderAccuracy` to `testGetBidderAccuracy`? Both exist, though... so it looks like you copy/pasted it rather than renaming.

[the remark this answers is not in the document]
hhhjort: Well, I renamed the function, and then introduced the original name as a new function. Because I needed to write a second test function that duplicated 95% of the code, I converted the original test into a more flexible helper function, called by two calling functions whose names spell out the details of what is being tested in each case. I woudn't think that git would be parsing function names, just noticing that this block of code was untouched but a bunch of lines added inbetween the function name and the block of code.

adapters/info.go
hhhjort: Ahh, other parts of the code base we used snake case, but better to be more locally consistent.

docs/developers/default-request.md
hhhjort: Yes, missed that, since it wasn't handled in `bidders.go`

router/router.go
```diff
@@ -77,6 +77,16 @@ func NewJsonDirectoryServer(schemaDirectory string, validator openrtb_ext.Bidder
 		}
 		data[bidder] = json.RawMessage(validator.Schema(bidderName))
 	}
+
+	// Add in any default aliases
+	for aliasName, bidderName := range aliases {
+		bidderData, ok := data[bidderName]
+		if !ok {
+			glog.Fatalf("Default alias (%s) exists referencing unknown bidder: %s", aliasName, bidderName)
```
dbemiller: Good catch!

endpoints/openrtb2/amp_auction.go
```diff
@@ -35,12 +35,14 @@ type AmpResponse struct {
 
 // NewAmpEndpoint modifies the OpenRTB endpoint to handle AMP requests. This will basically modify the parsing
 // of the request, and the return value, using the OpenRTB machinery to handle everything inbetween.
-func NewAmpEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string) (httprouter.Handle, error) {
+func NewAmpEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte) (httprouter.Handle, error) {
```
mkendall07: at what point do you decide to send an object instead of an ever increasing list of arguments? Food for thought.

Issue #722, opened 2018-10-23
Currency tests failing sporadically


dbemiller:
I saw lots of errors like this while running `./validate.sh` locally. There seem to be two unrelated problems.

The first is an `os.SyscallError` when trying to fetch conversion rates:

```
        	Error Trace:	rate_converter_test.go:526
        	            				asm_amd64.s:1333
        	Error:      	Expected nil, but got: &url.Error{Op:"Get", URL:"http://127.0.0.1:53676", Err:(*net.OpError)(0xc0004c2000)}
        	Test:       	TestRace
        	Messages:   	root error was: connect: can't assign requested address
```

This is on a Macbook Pro. The message: `root error was: connect: can't assign requested address` comes from some custom code I added to unwrap the `net.OpError` that shows up as a pointer in the original message.

I suspect Travis didn't have these because it runs tests on a different OS.

------------------------------------

The second does not reproduce consistently... but it's not rare either. I suspect Travis didn't catch it because the failures are intermittent.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1409538]

goroutine 136136 [running]:
github.com/prebid/prebid-server/currencies.(*Rates).GetRate(0x0, 0x14de558, 0x3, 0x14de4fe, 0x3, 0x4, 0xc00048cef0, 0x10784df)
	/Users/dbemiller/go/src/github.com/prebid/prebid-server/currencies/rates.go:47 +0x48
github.com/prebid/prebid-server/currencies_test.TestRace.func2(0xc0006a8b40, 0xc0001ce500, 0xc00009c8f0, 0xc000180300, 0x0)
	/Users/dbemiller/go/src/github.com/prebid/prebid-server/currencies/rate_converter_test.go:525 +0x303
created by github.com/prebid/prebid-server/currencies_test.TestRace
	/Users/dbemiller/go/src/github.com/prebid/prebid-server/currencies/rate_converter_test.go:512 +0x227
FAIL	github.com/prebid/prebid-server/currencies	20.826s
```

This is _probably_ due to a race condition that the race detector doesn't pick up, but nevertheless exists. 

------------------------------

I'm not sure whether or not these are expected... but I'll put them down here as extra info. There are a big string of log messages before the test finally fails:

```
E1023 09:38:02.214017    6796 rate_converter.go:105] Error updating conversion rates: unexpected end of JSON input
[...]

PR #723, opened 2018-10-23
Removes the buggy currency code

hhhjort:
The race condition test was failing as `mockedHttpServer` was being hit faster than it could respond, resulting in many requests failing (sporatically). There may also be a legitimate race condition hiding in there, as some segfaults were occasionally observed. (see issue #722 ) pulling the currency code until these issues can be resolved.

@benjaminch


benjaminch:
Hey @hhhjort,

Sorry about that ... If I understand properly, currency code is raising exceptions during tests because of the number of requests generated by the tests ?
If so I can lower the number of calls to assert the race conditions. 
What will be the right number of workers / intervals to test properly race condition without generating errors elsewhere ?

Cheers

hhhjort:
The problem stems from concurrent requests to the mock server. If you lower the number of workers/interval frequency to insure that one call to the mock server completes before the next time it is called, will that mean we have essentially lowered the race condition test to effectively a single threaded entity? I expect the http processing through the loop back device makes the http call the longest process running. If that never overlaps itself, how often do we expect that processing the rates will land just on the border of the rate update and thus exercise the race condition potential?

I would rather see a solution where the rate of currency conversions isn't limited to match the rate of sequential calls to update the rates. Perhaps have one goroutine run a tight loop calling the mock server, and a collection of other goroutines calling for rate conversions only. The code already limits to only one update call at a time, so no need to allow two updates to try to process concurrently.

dbemiller:
Let's move the discussion to #722.

PR #724, opened 2018-10-23
Introduce currency converter mechanism

benjaminch:
This CL introduces a mechanism to fetch currencies rates periodically from a remote source.
Those rates are stored in `CurrencyConverter` struct which can be injected in PBS to get rates.

This doesn't introduce any changes in PBS workflow just yet.

It also address issues described in #722 regarding httptestClient overload in tests generating
errors while running `validate.sh` script.

Issue: #280

### DIFF
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -1,0 +1,162 @@
+package currencies
+
+import (
+	"encoding/json"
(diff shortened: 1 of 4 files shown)


benjaminch:
Thanks :)

PR #729, opened 2018-10-30
I was appending when I should have been referencing. 

muncha:
This resulted in a null bid being tried for every real bid, and an error in the result JSON for every bid, in addition to the real bid.

Also, just a note - should Gopkg.lock be added to the project .gitignore?

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,495 +2,295 @@
 
 
 [[projects]]
-  digest = "1:04aea75705cb453e24bf8c1506a24a5a9036537dbc61ddf71d20900d6c7c3ab9"
   name = "github.com/DATA-DOG/go-sqlmock"
   packages = ["."]
-  pruneopts = "UT"
   revision = "d76b18b42f285b792bf985118980ce9eacea9d10"
   version = "v1.3.0"
 
 [[projects]]
   branch = "master"
-  digest = "1:d6afaeed1502aa28e80a4ed0981d570ad91b2579193404256ce672ed0a609e0d"
   name = "github.com/beorn7/perks"
   packages = ["quantile"]
-  pruneopts = "UT"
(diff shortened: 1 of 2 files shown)


dbemiller:
No. `Gopkg.lock` guarantees that everyone gets the same exact code when they check out a PBS commit and run `dep ensure`. 

Changes to `Gopkg.lock` are code changes to the project just like any other. That file shouldn't change in a PR like this.

Based on your changes, it looks like you're on an old version of `dep`. If you run `dep version | grep version`, it should say `v0.5.0` if you're on the latest. If it doesn't, try upgrading it. You'll probably find that `Gopkg.lock` stops changing then.

muncha:
Thank you. I had installed dep through apt in Ubuntu 18.04 it is only using 0.3.2. Updated to 0.5.0.

tl;dr - I'm putting in a fresh pull request.

After deleting my Gopkg.lock and re-running dep ensure, locally it looked OK. The file sizes matched between branch master and branch beachfront-bidder, and a git diff showed no differences at that level. So I did  a push to my upstream remote (beachfront).  I realized after that that I should have done a git diff between what I had and origin. While the file sizes were the same, the formatting of the Gopkg.lock showed a lot of differences. It may have been overkill, but considering how trivial the actual changes that I'm pushing here are, I figured I'd just start over again. I deleted the beachfront-bidder branch off of beachfront's github, deleted my local, re-cloned master and did it from scratch. It looks fine now.  I'll put up a new pull request in a sec.

dbemiller:
That's fine. No worries. There are some docs on `dep ensure` here, if you're interested: https://golang.github.io/dep/docs/daily-dep.html#using-dep-ensure

tl;dr: if you delete `Gopkg.lock` and run `dep ensure`, you will probably not get back the exact same lock file as you had before. If you run `dep ensure` on a project that already has a `Gopkg.lock` in it, it should not change.

PR #730, opened 2018-10-30
Adds support for generating imp.displaymanagerver for AppNexus

hhhjort:
Simple little addition to the Appnexus adapter

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/buger/jsonparser"
 	"github.com/prebid/prebid-server/pbs"
 
 	"golang.org/x/net/context/ctxhttp"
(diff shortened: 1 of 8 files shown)


bretg:
Thanks for bringing this up @dbemiller. I agree it should be treated as a real API feature. Please update the docs, tests. Unclear we need the server to constrain request.app.ext.prebid.source at this point since it's optional.

hhhjort:
So we should move this to incoming request validation rather than simply being an AppNexus thing?

dbemiller:
_validation_ should be with the incoming request validation. There should be some structs in `openrtb_ext` for it, and PBS should return 400s if someone sends in anything besides a `string`, etc.

The code that moves these fields into `displaymanagerver` is definitely specific to the Appnexus adapter though.

adapters/appnexus/appnexus.go
```diff
@@ -248,8 +249,17 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters
 	memberIds := make(map[string]bool)
 	errs := make([]error, 0, len(request.Imp))
 
+	// AppNexus openrtb2 endpoint expects imp.displaymanagerver to be populated, but some SDKs will put it in imp.ext.prebid instead
+	var defaultDisplayManagerVer string
+	if request.App != nil {
+		source, err1 := jsonparser.GetString(request.App.Ext, "prebid", "source")
+		version, err2 := jsonparser.GetString(request.App.Ext, "prebid", "version")
+		if (err1 == nil) && (err2 == nil) {
+			defaultDisplayManagerVer = fmt.Sprintf("%s-%s", source, version)
```
dbemiller: FYI I actually profiled this a few months ago, and `source + "-" + version` runs much faster. Not a huge deal, but... *shrugs*

PR #731, opened 2018-10-31
Corrected redundant append bug, and fixed Gopkg.lock corruption from previous PR

muncha:
When I did git diff to origin, there where problems. The Gopkg.lock was formatted differently. I just did a fresh clone from origin, and redid my trivial change in beachfront.go. A diff with origin now looks right.


dbemiller:
You may want a regression test to make sure this bug doesn't reappear later... but your call. It'll just help keep your adapter more stable over time.

PR #732, opened 2018-10-31
patches the beachfront usersyncer to not return a URL when none is co…

hhhjort:
…nfigured.

There is no default usersync URL for Beachfront. this will now return `""` as the syncer URL rather than the platform ID, which is more generally recognized as an unusable URL.


dbemiller:
Could use a test for the `/cookie_sync` handler too, but... I have pretty high confidence that that code does what it's trying to do.

PR #735, opened 2018-11-01
Defined cookie sync URL in config

muncha:
Also cleared deprecated comment in use sync.

### DIFF
--- a/adapters/beachfront/usersync.go
+++ b/adapters/beachfront/usersync.go
@@ -14,7 +14,7 @@ func NewBeachfrontSyncer(cfg *config.Configuration) usersync.Usersyncer {
 	b := string(openrtb_ext.BidderBeachfront)
 	usersyncURL := cfg.Adapters[b].UserSyncURL
 	platformID := cfg.Adapters[b].PlatformID
-	// By default, beachfront has a platformID, but no URL. This is only useful if we have both
+
 	url := ""
 	if len(usersyncURL) > 0 {
 		url = fmt.Sprintf("%s%s", usersyncURL, platformID)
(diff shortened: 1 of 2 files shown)


dbemiller:
Feel free to remove the logic that was added in #732 too... it's not needed anymore. We just added that because we didn't know you'd be so proactive about fixing it :)

muncha:
Thanks. I'm going to leave #732 for now. We've got our first live user on this, and they're using the appnexus server. So I'm just going to watch this for a bit and see if there's anything else I need to change as far as cookie syncs. If not I'll clean it up in a few days.

Can you give me some idea of when this might be live on the appnexus deployment?

dbemiller:
@hhhjort is in doing the releases for appnexus... should get his feedback

hhhjort:
Just released the fix for no beachfront usersync URL (#732), so probably sometime next week for this to hit the code. If there is a rush to get usersyncs fully working on the appnexus instance, we can always update the config directly instead of relying on the default being found in the code.

muncha:
@hhhjort 
Thank you, but no need to change anything there just yet. I just had a little talk with the beachfront guys and it appears that this syncing is still not going to be right on their end. Hopefully I'll have another sync related PR next week that actually does what it needs to do.

dbemiller:
ok. should we close this for now then? Or could you put `WIP` in the title until it's ready to be re-reviewed?

Thanks.

muncha:
@dbemiller I think that you should go ahead and close this. I'll have a new PR in a few days, but it will include some additional stuff not related to usersyncs - as well as a few minor usersync changes. I'll put in a new PR for that.

Also @hhhjort  -Please define the beachfront platform_id as 155 in appnexus' server config. I've changed that to the default for now, so no rush.

PR #737, opened 2018-11-02
Fixed aAdkernelAdn usersync bug

### LINKED ISSUE #736: [adkernelAdn] Usersyncs not working
In development, with normal configs, Adkernel is returning usersync URLs like this:

```
        {
            "bidder": "adkernelAdn", 
            "no_cookie": true, 
            "usersync": {
                "type": "redirect", 
                "url": "http%3A%2F%2Flocalhost%3A8000%2Fsetuid%3Fbidder%3DadkernelAdn%26uid%3D%7BUID%7D"
            }
        }
```

They should be:

```
        {
            "bidder": "adkernelAdn", 
            "no_cookie": true, 
            "usersync": {
                "type": "redirect", 
                "url": "https://tag.adkernel.com/syncr?gdpr=0&gdpr_consent=&r=http%3A%2F%2Flocalhost%3A8000%2Fsetuid%3Fbidder%3DadkernelAdn%26uid%3D%7BUID%7D"
            }
        }
```

The root cause is that the AdkernelAdn syncer isn't reading the config properly. Fix incoming shortly.

dbemiller:
Fixes #736.

This is subtle... but apparently Viper manages all config options as lowercase. I left some `// Fixes` and `// Prevents` comments because this looks like an easy bug to re-introduce later.

### DIFF
--- a/adapters/adkernelAdn/usersync.go
+++ b/adapters/adkernelAdn/usersync.go
@@ -13,7 +13,8 @@ import (
 const adkernelGDPRVendorID = uint16(14)
 
 func NewAdkernelAdnSyncer(cfg *config.Configuration) usersync.Usersyncer {
-	usersyncURL := cfg.Adapters[string(openrtb_ext.BidderAdkernelAdn)].UserSyncURL
+	// Fixes #736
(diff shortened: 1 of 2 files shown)


dbemiller:
It does, yes. But their usersyncer is already doing this :)

PR #742, opened 2018-11-06
Added a configurable "concurrent incoming connection" limter

dbemiller:
Use case here is to help protect against sudden traffic spikes. Without it, traffic spikes can grind things to a halt by causing us to open up many new connections to Bidders all at once.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -401,12 +401,13 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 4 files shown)


dbemiller:
Let's not merge this... I did some research and found some much more scalable ways to shed load during traffic spikes.

No sense in adding config values just to deprecate/remove them later.

Issue #740, opened 2018-11-06
Validation error on user-ext-consent-int.json


muncha:
While running my validation after getting my usersync and contractual issues worked out, I ran into this validation error. It's a super easy fix (unless I'm missing something), and I have it in a branch, isolated from my other changes that I can put in as a PR. Or maybe you guys want to take care of it, or maybe I should just include it in the PR I should be putting in before noon today with my beach front changes?

```
ok  	github.com/prebid/prebid-server/endpoints/info	(cached)
--- FAIL: TestBadRequests (0.27s)
    auction_test.go:152: 
        	Error Trace:	auction_test.go:176
        	            				auction_test.go:152
        	Error:      	Not equal: 
        	            	expected: "Invalid request: request.user.ext object is not valid: json: cannot unmarshal number into Go struct field ExtUser.consent of type string\n"
        	            	actual  : "Invalid request: request.user.ext object is not valid: json: cannot unmarshal number into Go value of type string\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-Invalid request: request.user.ext object is not valid: json: cannot unmarshal number into Go struct field ExtUser.consent of type string
        	            	+Invalid request: request.user.ext object is not valid: json: cannot unmarshal number into Go value of type string
        	            	 
        	Test:       	TestBadRequests
        	Messages:   	file sample-requests/invalid-whole/user-ext-consent-int.json had bad response body
E1106 09:05:07.827908   11379 auction.go:138] /openrtb2/auction Critical error: Critical, unrecoverable error.
FAIL
FAIL	
```

muncha:
Well, I don't know what's going on here. This must be something weird with my local as if I try changing the message to what will pass locally it fails on Travis. So I can't put my PR in yet. Here is the error that I get when I try my PR (#743)  with the naive fix, which is the opposite of what I get locally :

`--- FAIL: TestBadRequests (0.32s)
	auction_test.go:152: 
			Error Trace:	auction_test.go:176
			            				auction_test.go:152
			Error:      	Not equal: 
			            	expected: "Invalid request: request.user.ext object is not valid: json: cannot unmarshal number into Go value of type string
"
			            	actual  : "Invalid request: request.user.ext object is not valid: json: cannot unmarshal number into Go struct field ExtUser.consent of type string
"
			            	
			            	Diff:
			            	--- Expected
			            	+++ Actual
[...]

PR #743, opened 2018-11-06
Cookie sync

muncha:
I am not including the fix for issue #740 in this as, although by changing the expected message in the JSON I was able to pass validation while I was testing, the proper fix may involve changing the type to what is actually indicated ( ExtUser.consent ), or some other more involved fix.

What this does include are various bugs which were revealed in initial live testing, mostly related to cookie syncs and both the video and banner endpoint contracts. Where possible I've tried to use parts of the openRTB package structs instead of building custom structs, but still a lot of them are just different enough for that not to be possible. I've updated all validation to reflect these changes and checked coverage. I also set the default platform_id to 155 which indicates the appnexus server and points our cookie sync redirect to: https://prebid.adnxs.com/pbs/v1/setuid?

### DIFF
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -64,7 +64,6 @@ type BeachfrontVideoImp struct {
 
 type BeachfrontVideoDevice struct {
 	UA         string `json:"ua"`
-	Devicetype int    `json:"deviceType"`
 	IP         string `json:"ip"`
 }
 
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -16,10 +16,13 @@ import (
 const Seat = "beachfront"
 const BidCapacity = 5
 
-// const BannerEndpoint = "https://display.bfmio.com/prebid_display"
-const BannerEndpoint = "https://qa.bfmio.com/prebid_display"
-// const VideoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id="
-const VideoEndpoint = "https://qa.bfmio.com/bid.json?exchange_id="
+const BannerEndpoint = "https://display.bfmio.com/prebid_display"
+
+// const BannerEndpoint = "https://qa.bfmio.com/prebid_display"
+
+const VideoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id="
+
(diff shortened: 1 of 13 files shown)


dbemiller:
#740 is a bug in Go 1.11, actually. There's a small note on it here: https://github.com/prebid/prebid-server/blob/master/docs/developers/automated-tests.md#automated-tests

Upgrade to 1.11.1 and the tests will pass again.

muncha:
Cool. Thanks.

hhhjort:
We don't have a regular schedule, we release as seems necessary. We do have a code freeze around thanksgiving and again around christmas. Unless we find a lot of issues in the PR, I can't imagine it being delayed into the thanksgiving freeze.

adapters/beachfront/beachfront.go
```diff
@@ -36,40 +42,32 @@ type BeachfrontRequests struct {
...
 
+// Soooo close, but not quite openRTB
 type BeachfrontVideoImp struct {
 	Video    BeachfrontSize `json:"video"`
 	Bidfloor float64        `json:"bidfloor"`
-	Id       int            `json:"id"`
-	ImpId    string         `json:"impid"`
+	Id       int            `json:"id"` // A sequential count of which imp on the page this is. Since the bfm
+	// 	video endpoint only returns one response for one imp, this is
+	//	never going to happen. This will always be 0. @TODO - Alex
```
dbemiller: There are a few TODOs in here... do you want us to wait until those are resolved before we merge, or did you plan to resolve them later?

If they've already been resolved, remove the TODOs?

adapters/beachfront/beachfront.go
```diff
@@ -16,12 +16,18 @@ import (
 const Seat = "beachfront"
 const BidCapacity = 5
 
-const BannerEndpoint = "https://display.bfmio.com/prebid_display"
+const BannerEndpoint = "http://display.bfmio.com/prebid_display"
+
+// const BannerEndpoint = "http://qa.bfmio.com/prebid_display"
```
dbemiller: Any reason these are left commented, rather than just deleted?

muncha:
Sorry for the delay - the TODO's are just for down the line as you seem to have correctly assumed.

adapters/beachfront/beachfront.go
muncha: No real reason. I'll leave similar out in future PRs.

PR #747, opened 2018-11-12
Check size match as per config before creating request

ix-certification:
Fixing a bug identified in the IX adapter where requests were being sent for sizes that were not present in the adapter config.

Additionally, some validation code was moved outside the loop containing it, because it did not need to be inside the loop.

### DIFF
--- a/static/bidder-params/ix.json
+++ b/static/bidder-params/ix.json
@@ -19,5 +19,5 @@
       "description": "An array of two integer containing the dimension"
     }
   },
-  "required": ["siteId", "size"]
+  "required": ["siteId"]
 }
--- a/static/bidder-params/ix.json
+++ b/static/bidder-params/ix.json
@@ -6,8 +6,18 @@
   "properties": {
     "siteId": {
       "type": "string",
+      "minLength": 1,
       "description": "An ID which identifies the site selling the impression"
+    },
+    "size": {
+      "type": "array",
+      "items": {
+        "type": "integer"
+      },
+      "minItems": 2,
+      "maxItems": 2,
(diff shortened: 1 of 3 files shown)


hhhjort:
Can you make the size parameter optional for now, and have the adapter return an error if you truely can't do anything without the size? If we just flip it from non-existent to required, all requests that have have been set up that include ix, but no size will cause the request to fail for all bidders, not just yours.

Once publishers have had plenty of time to fix their requests to the new format we can consider making it required.

static/bidder-params/ix.json
```diff
@@ -6,8 +6,18 @@
...
+      "items": {
+        "type": "integer"
+      },
+      "minItems": 2,
+      "maxItems": 2,
+      "description": "An array of two integer containing the dimension"
     }
   },
-  "required": ["siteId"]
+  "required": ["siteId", "size"]
```
dbemiller: Bidder params can't have breaking changes. Since they're in Stored Requests, they may be in host companies' DBs. It's unrealistic to make host companies do DB migrations, at least not without a significant breaking change window.

Can you add this as an optional param, and make your adapter return `BadInput` errors on requests that don't have it?

ix-certification:
> Can you make the size parameter optional for now, and have the adapter return an error if you truely can't do anything without the size? If we just flip it from non-existent to required, all requests that have have been set up that include ix, but no size will cause the request to fail for all bidders, not just yours.
> 
> Once publishers have had plenty of time to fix their requests to the new format we can consider making it required.

Fixed in dbf18c4

PR #748, opened 2018-11-14
Adding GumGum server adapter

cchang320:


adapters/gumgum/gumgum.go
```diff
@@ -0,0 +1,145 @@
...
+			if request.Imp[i].Banner != nil {
+				bannerCopy := *request.Imp[i].Banner
+				if bannerCopy.W == nil && bannerCopy.H == nil && len(bannerCopy.Format) > 0 {
+					format := bannerCopy.Format[0]
+					bannerCopy.W = &(format.W)
+					bannerCopy.H = &(format.H)
+				}
+				request.Imp[i].Banner = &bannerCopy
+				validImps = append(validImps, request.Imp[i])
+				trackingId = zone
```
hhhjort: You might want to do some checks if people are passing inconsistent values across the imps. Shouldn't happen in theory, but users can end up doing all sorts of crazy things.

cchang320: Is it possible for the user to send different values (unless they decide to randomly generate this value in which case none of these values would be valid on GumGum's end)? Since whatever is in the imp.ext object is hard coded when the user sets up the client for the site right?

hhhjort: There is one `imp.ext` for each imp, so technically it could be possible to send inconsistent values. It would be a strong indication that there is something seriously wrong with their prebid implementation to generate a mess like that. I am not going to deny this PR if there is no check, just wanted to offer it as an idea to consider.

static/bidder-params/gumgum.json
```diff
@@ -0,0 +1,24 @@
...
+  "title": "GumGum Adapter Params",
+  "description": "A schema which validates params accepted by the GumGum adapter",
+  "type": "object",
+  "properties": {
+    "zone": {
+      "type": "string",
+      "description": "A tracking id used to identify GumGum zone.",
+      "minLength": 8
+    },
+    "deviceHeight": {
```
dbemiller: Hey, out of curiosity: why do you use `deviceWidth` and `deviceHeight` as bidder params here?

OpenRTB already defines these as `request.device.w` and `request.device.h`... so you wouldn't need anything bidder-specific to capture them.

Also: the device size will vary from user to user. These bidder params are defined by Publishers define for a particular page. How would you expect them to fill these out?

cchang320: Sure I can clarify. I was testing my adapter with my coworker who created a test page running the Prebid client and we noticed that the device object never got sent in the bid request. We spoke with a publisher that wants to call us through Prebid server and asked if they can send over the device info in the request and they told us that they do not control what gets sent over in the request since they run Prebid.js. So I was under the impression that Prebid.js is not sending the device info in the request. 

Regarding how the publisher would set this up, the way my coworker defined the device width and height params was by using the functions window.innerWidth() and window.innerHeight(). 

Please let me know if there was perhaps some issue with our test page setup and the device info does actually get sent over through Prebid.js.

dbemiller: Aha! Yeah, that makes sense. Thanks!

Your test page was probably fine... Prebid.js just isn't sending these _yet_. I opened https://github.com/prebid/Prebid.js/pull/3336 to add them, though.

Feel free to test using that branch, if you want to stay ahead of the PRs & merge ^^. I haven't worked with Prebid.js in a long time, but I believe it should fix your problem.

cchang320: Gotcha! Will you be merging your changes soon? I’m out of office this week so I can test with your branch on Monday. If I still don’t see the device info included in the request, how should we proceed?

dbemiller: ugh... so: yes, I plan to merge these changes as soon as possible. They're small, so I can nag the Prebid.js team for a review if necessary.

...but right now the tests don't pass, and I can't even build it locally because of https://github.com/prebid/Prebid.js/issues/3330. The Prebid.js folks are waiting on https://github.com/AveVlad/gulp-connect/pull/259, which is outside of our control.

So... I definitely will try to get this in quickly, but can't promise a timeline either.

cchang320: Ah I see. I know you said it’s hard to give a timeline but do you think the issues will be resolved by the end of December? Trying to merge my adapter before the end of Q4 :) Also in the case you think it won’t be merged by the end of December, would it be terrible to merge my adapter as is and get rid of the device width and height in the ext object once your changes have been merged and the device info is being sent over?

dbemiller: We merged it now :). They're actually releasing today too, so these changes should be live in v1.35.0 in the next few hours. If the tests fail during release validation, it might take a few days... but shouldn't be any longer than that.

cchang320: Awesome news! Okay I will go ahead and remove the device width and height from the extension object and see how it goes. Thanks!

adapters/gumgum/gumgum.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	if len(validImps) == 0 {
+		return nil, errs
+	}
+
+	request.Imp = validImps
+
+	if request.Site != nil {
```
dbemiller: I notice in your `static/bidder-info` that you also support apps... did you want to do something with the tracking ID for those here too? Maybe in `req.App`?

It seems a bit weird that the `zone` info doesn't make it anywhere into the request that goes to your servers on App requests.

cchang320: Sorry that's an oversight on my end. GumGum will only support site for now so I'll correct our bidder-info file. Thanks for catching that! If I modify that file will you have to reapprove?

dbemiller: yeah, but it's fine. We're active :)

PR #752, opened 2018-11-16
First pass at config disabled bidders

### LINKED ISSUE #379: Ability to turn off bidders
Some installations of Prebid Server may not want to run all available bidder adapters.

Scenarios:
- installation of Prebid Server not approved to use one or more bidder adapters
- installation of Prebid Server has no customers interested in demand from one or more bidder adapters

Ideally, Prebid Server could be built without the full list of bidder adapters, but minimally an installation should be able to turn off requests for a given bidder adapter in the config.

If a bidder is requested that is not configured, an error should result. Ideally the error would be "adapter not active". This is different than when an unknown adapter is requested.

hhhjort:
Fixes #379


router/router.go
```diff
@@ -173,6 +173,25 @@ func New(cfg *config.Configuration) (r *Router, err error) {
 	disabledBidders := map[string]string{
 		"indexExchange": "Bidder \"indexExchange\" has been deprecated and is no longer available. Please use bidder \"ix\" and note that the bidder params have changed.",
 	}
+	bidderList := openrtb_ext.BidderList()
```
dbemiller: I don't see any tests for this... it's enough code that there should probably be some.

I'd recommend splitting it into its own function anyway... both to help make it testable and to avoid adding lines to this already-large function. Could do something like:

`disabledBidders := readDisabledBidders(cfg.Adapters)`


On an unrelated note: code is more readable if you declare variables as close to the point where they're used as possible. You're not really using this until `adapters.ParseBidderInfos(p, bidderList)`, so... put it down there.

exchange/adapter_map.go
```diff
@@ -69,11 +69,18 @@ func newAdapterMap(client *http.Client, cfg *config.Configuration, infos adapter
 	}
 
 	allBidders := make(map[openrtb_ext.BidderName]adaptedBidder, len(ortbBidders)+len(legacyBidders))
+
 	for name, bidder := range legacyBidders {
-		allBidders[name] = adaptLegacyAdapter(bidder)
+		// Clean out any disabled bidders, or bidders that are otherwise missing infos (Only reason for not having infos should be that they are disabled)
```
dbemiller: re: my tech talk... this is a great example of a confusing Message.

1. If they see the name & type "BidderInfos," why would either a caller or callee think "enabled bidders"?
2. The `cfg` is already a function argument, and it has `Disabled` info in it very clearly.

Better to read from the config here.

endpoints/openrtb2/auction_test.go
```diff
@@ -197,15 +274,36 @@ func readFile(t *testing.T, filename string) []byte {
...
 
 // doRequest populates the app with mock dependencies and sends requestData to the /openrtb2/auction endpoint.
-func doRequest(t *testing.T, requestData []byte, aliased bool) (int, string) {
+func (gr *getResponseFromDirectory) doRequest(t *testing.T, requestData []byte) (int, string) {
 	aliasJSON := []byte{}
-	if aliased {
+	if gr.aliased {
 		aliasJSON = []byte(`{"ext":{"prebid":{"aliases": {"test1": "appnexus", "test2": "rubicon", "test3": "openx"}}}}`)
 	}
+	disabledBidders := map[string]string{
```
dbemiller: This is a lot of copy/paste code from the `router`... should try to find some way to share it, if it's possible.

I won't have time to think much about it before vacation... but this is usually a sign that some code lives in a module where it doesn't belong, or is private and ought to be public so other modules can use it.

endpoints/openrtb2/auction_test.go
```diff
@@ -137,41 +148,107 @@ func TestImplicitUserId(t *testing.T) {
 
 // TestGoodRequests makes sure we return 200s on good requests.
 func TestGoodRequests(t *testing.T) {
-	assertResponseFromDirectory(t, "sample-requests/valid-whole/exemplary", getRequestPayload, nilReturner, http.StatusOK, true)
-	assertResponseFromDirectory(t, "sample-requests/valid-whole/supplementary", noop, nilReturner, http.StatusOK, true)
+	exemplary := &getResponseFromDirectory{
```
dbemiller: good change!

At some point it might be worth experimenting with testify's Suite to see it if would make things here nicer: https://godoc.org/github.com/stretchr/testify/suite

but... not this PR.

endpoints/openrtb2/amp_auction.go
```diff
@@ -35,14 +35,17 @@ type AmpResponse struct {
...
 // NewAmpEndpoint modifies the OpenRTB endpoint to handle AMP requests. This will basically modify the parsing
 // of the request, and the return value, using the OpenRTB machinery to handle everything inbetween.
-func NewAmpEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte) (httprouter.Handle, error) {
+func NewAmpEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte, bidderMap map[string]openrtb_ext.BidderName) (httprouter.Handle, error) {
 	if ex == nil || validator == nil || requestsById == nil || cfg == nil || met == nil {
 		return nil, errors.New("NewAmpEndpoint requires non-nil arguments.")
 	}
 
 	defRequest := defReqJSON != nil && len(defReqJSON) > 0
+	if bidderMap == nil {
```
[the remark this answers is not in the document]
hhhjort: Yes, I thought perhaps letting nil be a standin for the default, but you are right it could get confusing.

router/router.go
```diff
@@ -195,19 +214,19 @@ func New(cfg *config.Configuration) (r *Router, err error) {
 	pbsAnalytics := analyticsConf.NewPBSAnalytics(&cfg.Analytics)
 
 	// Hack because of how legacy handles districtm
-	bidderList := openrtb_ext.BidderList()
-	bidderList = append(bidderList, openrtb_ext.BidderName("districtm"))
+	LegacyBidderList := openrtb_ext.BidderList()
```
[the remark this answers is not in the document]
hhhjort: Oops, good catch

exchange/adapter_map.go
hhhjort: My hack saved a lot of headache. Apparently we have several tests that depend on `newAdapaterMap()` working given a blank config. Cleaning up the mess that results when enforcing that `cfg.Adapters[<name>].Disabled` exist and be false. Should hopefully improve our code to be a bit more strict. Surprised that the function works at all on a blank config.

router/router.go
hhhjort: I have refactored this code.

PR #753, opened 2018-11-27
Configurable sync url

dbemiller:
Makes the cookie sync URL configurable. If the config value isn't defined, it defaults to the same external URL it uses today.


dbemiller:
I'll see what comes up in review, but... #763 may be a better approach to this. In my opinion in probably is, but... I'm leaving this open for now in case there's pushback about the breaking change.

dbemiller:
Closed in favor of #763

PR #759, opened 2018-12-04
Adds a limit feature to the /cookie_sync endpoint.

hhhjort:
This is to facilitate clients not overloading a user with syncs the first time they are encountered.


endpoints/cookie_sync.go
```diff
@@ -192,6 +197,18 @@ func (req *cookieSyncRequest) filterForGDPR(permissions gdpr.Permissions) {
...
 }
 
+// filterToLimit will enforce a max limit on cookiesyncs supplied, picking a random subset of syncs to get to the limit if over.
+func (req *cookieSyncRequest) filterToLimit() {
+	if req.Limit <= 0 {
+		return
+	}
+	for len(req.Bidders) > req.Limit {
+		cut := rand.Intn(len(req.Bidders))
+		req.Bidders = append(req.Bidders[:cut], req.Bidders[cut+1:]...)
```
benjaminch: Maybe there is a more efficient way of taking randomly the `req.Limit` elements leveraging `rand.Perm` returning a random permutation of the numbers 0 through N (not including N). It will avoid creating new slices and the `O(n^2)` operation.

```
biddersToKeep = make([]req.Bidders, req.Limit)
for n, idx := range rand.Perm(len(req.Bidders)) {
      if n == req.Limit {
          break
      }
      biddersToKeep[n] = req.Bidders[idx]
  }

req.Bidders = biddersToKeep
```

benjaminch: There is maybe also a way to shuffle the slice in place index by index until we reached the limit.
Something like
```
randIndex = rand.Intn(len(req.Bidders))
vals[n], vals[randIndex] = vals[randIndex], vals[n]
```

docs/endpoints/cookieSync.md
```diff
@@ -26,6 +27,9 @@ must supply a JSON object to define the list of bidders that may need to be sync
 If `gdpr` is  omitted, callers are still encouraged to send `gdpr_consent` if they have it.
 Depending on how the Prebid Server host company has configured their servers, they may or may not require it for cookie syncs.
 
+`limit` is optional. If present and greater than zero, it will limit the number of syncs returned to `limit`, dropping random syncs to
```
dbemiller: let's not specify "dropping random syncs" here... I think that over-constrains the API a little.

For example: the only use-case (at least for now) is page performance. Clients use this to tell us how many syncs they're ok having. Given that, it may be a good idea to refresh syncs which haven't expired yet, but will expire soon, and prioritize them based on the expiry deadline.

If you specify "random", that'd be a breaking change.

endpoints/cookie_sync.go
```diff
@@ -108,6 +109,9 @@ func (deps *cookieSyncDeps) Endpoint(w http.ResponseWriter, r *http.Request, _ h
 
 	parsedReq.filterExistingSyncs(deps.syncers, userSyncCookie)
 	parsedReq.filterForGDPR(deps.syncPermissions)
+	if parsedReq.Limit > 0 {
```
dbemiller: Since `if req.Limit <= 0` is checked inside `filterToLimit()`, you don't need to check `> 0` here.

endpoints/cookie_sync.go
```diff
@@ -192,6 +197,18 @@ func (req *cookieSyncRequest) filterForGDPR(permissions gdpr.Permissions) {
 	}
 }
 
+// filterToLimit will enforce a max limit on cookiesyncs supplied, picking a random subset of syncs to get to the limit if over.
+func (req *cookieSyncRequest) filterToLimit() {
+	if req.Limit <= 0 {
```
dbemiller: You can return immediately if the `Limit` is larger than the sync list length. Trims off the more expensive work.

endpoints/cookie_sync_test.go
```diff
@@ -186,6 +193,14 @@ func assertSameElements(t *testing.T, expected []string, actual []string) {
 	}
 }
 
+func assertSyncsLen(t *testing.T, responseBody []byte, limit int) {
+	t.Helper()
+	found := len(parseSyncs(t, responseBody))
+	if found != limit {
```
dbemiller: There's an `assert.Len()` in `testify` that simplifies this.

endpoints/cookie_sync.go
dbemiller: yeah... you're quickly converging on the Fisher-Yates shuffle: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm

Run that over the first `Limit` elements only, and then `req.Bidders = req.Bidders[:limit]` to set the slice. Either way, there are definitely ways to do this in O(limit) time and O(1) space.

endpoints/cookie_sync.go
hhhjort: Yes, I was going to remove that when I moved the check, then forgot to go back and actually remove it. Thanks.

endpoints/cookie_sync.go
hhhjort: I don't think it trims off anything. if `len(req.Bidders)` is under the limit, then we skip the `for` loop and return immediately already. I was thinking of adding such a check, but then thought harder on what the loop was actually doing and came to the conclusion it was unecessary

endpoints/cookie_sync.go
hhhjort: Though if on average we discard more than 2/3rds of the bidders, then starting from the front would be more efficient.

endpoints/cookie_sync.go
dbemiller: yeah, definitely ignore this now. In the old version it was making an array & copying stuff. Now it's negligible, and if I had to guess would probably be a net negative

endpoints/cookie_sync.go
dbemiller: We probably would discard more than 2/3s of the bidders in most cases... we're already at 20 bidders, which would be a sync limit of ~7. That's already quite a few requests for a page to make.

I wouldn't worry about it too much though. It's fine as is, or fine if you want to change it.

dbemiller:
bleh... sorry. tried to fix the merge conflict in the github editor and apparently failed. should be fixed now

hhhjort:
Prebid.js issue opened: https://github.com/prebid/Prebid.js/issues/3367

PR #760, opened 2018-12-06
[Currency] Introduce currencies conversions in PBS workflow

benjaminch:
This CL allows to start using the rates converter.
The default currency remains `USD`.

By default, the rate converter will fetch every 30 minutes the official
prebid.org currencies rates at http://currency.prebid.org/latest.json.

The rate converter can be disabled by setting `rates_converter.fetch_url` to 0.

I will also do another PR to add a new endpoint in order to expose the rate converter settings along with the actual rates to ease debug and such if it's ok for you.

For more details on this feature, please refer to #280

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -38,6 +38,7 @@ type Configuration struct {
 	Analytics            Analytics          `mapstructure:"analytics"`
 	AMPTimeoutAdjustment int64              `mapstructure:"amp_timeout_adjustment_ms"`
 	GDPR                 GDPR               `mapstructure:"gdpr"`
+	CurrencyConverter    CurrencyConverter  `mapstructure:"currency_converter"`
 	DefReqConfig         DefReqConfig       `mapstructure:"default_request"`
 }
 
@@ -73,6 +74,7 @@ func (cfg *Configuration) validate() configErrors {
 		errs = append(errs, fmt.Errorf("cfg.max_request_size must be >= 0. Got %d", cfg.MaxRequestSize))
 	}
 	errs = cfg.GDPR.validate(errs)
+	errs = cfg.CurrencyConverter.validate(errs)
 	return errs
 }
 
@@ -135,7 +137,19 @@ type Analytics struct {
 	File FileLogs `mapstructure:"file"`
 }
 
-//Corresponding config for FileLogger as a PBS Analytics Module
+type CurrencyConverter struct {
+	FetchURL             string `mapstructure:"fetch_url"`
+	FetchIntervalSeconds int    `mapstructure:"fetch_interval_seconds"`
+}
+
+func (cfg *CurrencyConverter) validate(errs configErrors) configErrors {
+	if cfg.FetchIntervalSeconds < 0 {
+		errs = append(errs, fmt.Errorf("currency_converter.fetch_interval_seconds must be in the range [0, %d]. Got %d", 0xffff, cfg.FetchIntervalSeconds))
+	}
+	return errs
+}
+
+// FileLogs Corresponding config for FileLogger as a PBS Analytics Module
 type FileLogs struct {
 	Filename string `mapstructure:"filename"`
 }
@@ -452,6 +466,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.usersync_if_ambiguous", false)
 	v.SetDefault("gdpr.timeouts_ms.init_vendorlist_fetches", 0)
 	v.SetDefault("gdpr.timeouts_ms.active_vendorlist_fetch", 0)
+	v.SetDefault("currency_converter.fetch_url", "https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json")
+	v.SetDefault("currency_converter.fetch_interval_seconds", 0) // #280 Not activated for the time being
 	v.SetDefault("default_request.type", "")
 	v.SetDefault("default_request.file.name", "")
 	v.SetDefault("default_request.alias_info", false)
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -19,6 +19,7 @@ type RateConverter struct {
 	syncSourceURL    string
 	rates            atomic.Value // Should only hold Rates struct
 	lastUpdated      atomic.Value // Should only hold time.Time
+	constantRates    Conversions
 }
 
 // NewRateConverter returns a new RateConverter
@@ -33,6 +34,13 @@ func NewRateConverter(
 		fetchingInterval,
 		nil, // no notifier channel specified, won't send any notifications
 	)
+}
+
+// NewRateConverterDefault returns a RateConverter with default values.
+// By default there will be no currencies conversions done.
+// `currencies.ConstantRate` will be used.
+func NewRateConverterDefault() *RateConverter {
+	return NewRateConverter(&http.Client{}, "", time.Duration(0))
 }
 
 // NewRateConverterWithNotifier returns a new RateConverter
@@ -58,6 +66,7 @@ func NewRateConverterWithNotifier(
 	// In case host do not want to support currency lookup
 	// we just stop here and do nothing
 	if rc.fetchingInterval == time.Duration(0) {
+		rc.constantRates = NewConstantRates()
 		return rc
 	}
 
@@ -149,8 +158,12 @@ func (rc *RateConverter) LastUpdated() time.Time {
 	return time.Time{}
 }
 
-// Rates returns current rates
-func (rc *RateConverter) Rates() *Rates {
+// Rates returns current conversions rates
+func (rc *RateConverter) Rates() Conversions {
+	if rc.constantRates != nil {
+		// Converter is not active, returning the constant rates
+		return rc.constantRates
+	}
 	if rates := rc.rates.Load(); rates != nil {
 		return rates.(*Rates)
 	}
@@ -159,4 +172,11 @@ func (rc *RateConverter) Rates() *Rates {
 
 type httpClient interface {
 	Do(req *http.Request) (*http.Response, error)
+}
+
+// Conversions allows to get a conversion rate between two currencies.
+// if one of the currency string is not a currency or if there is not conversion between those
+// currencies, then an err is returned and rate is 0.
+type Conversions interface {
+	GetRate(from string, to string) (float64, error)
 }
--- a/docs/endpoints/openrtb2/auction.md
+++ b/docs/endpoints/openrtb2/auction.md
@@ -343,6 +343,13 @@ This adds two optional properties:
 
 These fields will be forwarded to each Bidder, so they can decide how to process them.
 
+### OpenRTB Ambiguities
+
+This section describes the ways in which Prebid Server **implements** OpenRTB spec ambiguous parts.
+
+- `request.cur`: If `request.cur` is not specified in the bid request, Prebid Server will consider it as being `USD` whereas OpenRTB spec doesn't mention any default currency for bid request.
+```request.cur: ['USD'] // Default value if not set```
+
 ### OpenRTB Differences
 
 This section describes the ways in which Prebid Server **breaks** the OpenRTB spec.
--- a/exchange/bidder.go
+++ b/exchange/bidder.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/currencies"
 	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"golang.org/x/net/context/ctxhttp"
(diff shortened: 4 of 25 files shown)


router/router.go
```diff
@@ -217,12 +218,14 @@ func New(cfg *config.Configuration) (r *Router, err error) {
 	exchanges = newExchangeMap(cfg)
 	theExchange := exchange.NewExchange(theClient, pbc.NewClient(&cfg.CacheURL), cfg, r.MetricsEngine, bidderInfos, gdprPerms)
 
-	openrtbEndpoint, err := openrtb2.NewEndpoint(theExchange, paramsValidator, fetcher, cfg, r.MetricsEngine, pbsAnalytics, disabledBidders, defReqJSON)
+	rateConverter := currencies.NewRateConverter(&http.Client{}, cfg.RatesConverter.FetchURL, time.Duration(cfg.RatesConverter.FetchIntervalSeconds)*time.Second)
```
dbemiller: Any reason this doesn't use `theClient`?

From [the docs](https://golang.org/pkg/net/http/):

> Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used.

Also... if you don't set the `Transport.TLSClientConfig` the same way `theClient` does, I believe PBS hosts who use an `https://` URL inside the Docker container might get errors.

benjaminch:
Thanks for your feedback @dbemiller. 
I wanted to include the endpoint after this one in order to make it lighter. I understand that it would help, so either I can submit the PR with the endpoint before this one or directly in this one, as you prefer.
Maybe before this one ?

Cheers

dbemiller:
yeah, please make a separate PR for the endpoint :).

benjaminch:
Will do ! thx

dbemiller:
Were you planning another PR to complete the currency conversion _feature_, though? Or is this one supposed to be feature complete.

Asking because I still don't see `request.Cur` being read anywhere. All the code still assumes `USD`--and not just as a default.

benjaminch:
I wanted to have:

- This PR introducing the mechanics inside the PBS workflow
- Another introducing the endpoint to debug / check what's going on
- Another introducing the mapping of request.Cur
- A last one to activate this mechanics (activate the currency conversion param via config if not done already)

Does it work for you? Do you prefer less steps? I personally prefer to have more smaller non breaking PRs and a last very small one activating the feature.
Do you think qbout any other step needed or way to test this feature?

config/config.go
```diff
@@ -390,6 +404,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.usersync_if_ambiguous", false)
 	v.SetDefault("gdpr.timeouts_ms.init_vendorlist_fetches", 0)
 	v.SetDefault("gdpr.timeouts_ms.active_vendorlist_fetch", 0)
+	v.SetDefault("rates_converter.fetch_url", "http://currency.prebid.org/latest.json")
+	v.SetDefault("rates_converter.fetch_interval_seconds", 30*60) // 30 minutes
```
benjaminch: Missed this one but should be 0 in a first time (desactivated), we will activated it in the last PR for this feature

dbemiller:
> Does it work for you? Do you prefer less steps? I personally prefer to have more smaller non breaking PRs and a last very small one activating the feature.

Those steps sound totally fine to me. Multiple smaller PRs are good.

exchange/bidder.go
```diff
@@ -85,7 +86,7 @@ type bidderAdapter struct {
 	Client *http.Client
 }
 
-func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.BidRequest, name openrtb_ext.BidderName, bidAdjustment float64) (*pbsOrtbSeatBid, []error) {
+func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.BidRequest, name openrtb_ext.BidderName, bidAdjustment float64, rateConverter *currencies.RateConverter) (*pbsOrtbSeatBid, []error) {
```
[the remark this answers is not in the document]
hhhjort: Yes, that was the purpose of breaking out the `Rates` object. `requestBids()` would user the `rateConverter` once to get the `Rates` object, then use that set of rates for all the conversions. This also benefits performance as you avoid the atomic `load` of the rate data happening more than once in an auction.

docs/developers/rates-converter.md
```diff
@@ -0,0 +1,49 @@
+# Rates Converter Mechanics
+
+Prebid server supports rates conversions when receiving bids.
+
+## Default currency
+
+The default currency is `USD`. It means that any bids coming without an explicit currency will be interpreted as being `USD`.
```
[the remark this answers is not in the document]
benjaminch: Just had a look at OpenRTB 2.5 specs for example, and it seems that `USD` is the default currency as well. Did you meant there is something else deviating from OpenRTB specs?

config/config.go
```diff
@@ -37,6 +37,7 @@ type Configuration struct {
 	Analytics            Analytics          `mapstructure:"analytics"`
 	AMPTimeoutAdjustment int64              `mapstructure:"amp_timeout_adjustment_ms"`
 	GDPR                 GDPR               `mapstructure:"gdpr"`
+	RatesConverter       RatesConverter     `mapstructure:"rates_converter"`
```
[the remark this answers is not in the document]
benjaminch: Done

config/config.go
```diff
@@ -390,6 +404,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.usersync_if_ambiguous", false)
 	v.SetDefault("gdpr.timeouts_ms.init_vendorlist_fetches", 0)
 	v.SetDefault("gdpr.timeouts_ms.active_vendorlist_fetch", 0)
+	v.SetDefault("rates_converter.fetch_url", "http://currency.prebid.org/latest.json")
```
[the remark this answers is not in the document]
benjaminch: Done

config/config.go
benjaminch: Done

docs/developers/rates-converter.md
```diff
@@ -0,0 +1,49 @@
+# Rates Converter Mechanics
```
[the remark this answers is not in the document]
benjaminch: Done. Also flagged this feature as not yet ready

exchange/bidder.go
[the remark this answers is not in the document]
benjaminch: Sure !
So I introduced a `ConstantRates` which returns 1 if both from and to currencies are the same, otherwise it returns not found.
It allowed to reduce the complexity in the bidder code which receive directly a `Conversions` interface which is either a `ConstantRates` or `Rates`.
The `Rates()` is called once per auction in `exchange.HoldAuction()` to avoid any discrepancies between bidders and the unnecessary multiple fetch per auction.

I kept some tests in `bidder_test` tackling common use cases (checked in `currencies` package and in PBS workflow in bidder just to make sure.

So the bidder code change is minimal but is backed with unit tests and most of the logic is handled in the `Conversions` implementations
Let me know if you guys prefer this implementation (I will be happy to change it otherwise)

router/router.go
benjaminch: Oh, no, no reasons, I just replaced it to use `theClient `

docs/developers/rates-converter.md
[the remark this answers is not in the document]
dbemiller: ah! yeah, sorry about that. I did misread what you wrote here. Bids _do_ default to USD in the spec.

I thought you were talking about `bidrequest.cur` here. It's an optional property... but the spec doesn't define a default. We're treating undefined as if it were `["USD"]`.

On second thoguht, that's not really a "deviation"... the spec is just defined vaguely. It should definitely make it into the docs somewhere though.

currencies/rate_converter.go
```diff
@@ -160,3 +173,8 @@ func (rc *RateConverter) Rates() *Rates {
 type httpClient interface {
 	Do(req *http.Request) (*http.Response, error)
 }
+
+// Conversions allows
```
dbemiller: Unfinished comment.

docs/endpoints/openrtb2/auction.md
```diff
@@ -375,6 +375,10 @@ In the OpenRTB spec, `request.imp[i].secure` says:
 In Prebid Server, an `https` request which does not define `secure` will be forwarded to Bidders with a `1`.
 Publishers who run `https` sites and want insecure ads can still set this to `0` explicitly.
 
+#### Default currency
+
+In the OpenRTB spec, `request.imp[i].secure` says:
```
dbemiller: Stray/unfinished docs. ...looks like a copy/paste thing?

docs/developers/rates-converter.md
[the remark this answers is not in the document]
benjaminch: Ok, let me know where you want me to add that fact. It's specified in the rates-converter docs but nowhere else as of today.

router/router.go
```diff
@@ -219,12 +220,16 @@ func New(cfg *config.Configuration) (r *Router, err error) {
 	exchanges = newExchangeMap(cfg)
 	theExchange := exchange.NewExchange(theClient, pbc.NewClient(&cfg.CacheURL), cfg, r.MetricsEngine, bidderInfos, gdprPerms)
 
-	openrtbEndpoint, err := openrtb2.NewEndpoint(theExchange, paramsValidator, fetcher, cfg, r.MetricsEngine, pbsAnalytics, disabledBidders, defReqJSON, bidderMap)
+	currencyConverter := currencies.NewRateConverter(theClient, cfg.CurrencyConverter.FetchURL, time.Duration(cfg.CurrencyConverter.FetchIntervalSeconds)*time.Second)
+
+	openrtbEndpoint, err := openrtb2.NewEndpoint(theExchange, paramsValidator, fetcher, cfg, r.MetricsEngine, pbsAnalytics, currencyConverter, disabledBidders, defReqJSON, bidderMap)
```
dbemiller: The code is quite a bit simpler if you send the `currencyConverter` directly into `NewExchange`, rather than through `NewEndpoint` and `NewAmpEndpoint`.

Do any of your future changes planning to use the currency conversions in the endpoint code? I can't think of any reason you'd want to do that... but let me know if I'm missing something.

docs/developers/rates-converter.md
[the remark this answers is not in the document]
benjaminch: Ok I see :)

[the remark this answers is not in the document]
benjaminch: I just added a section `OpenRTB Ambiguities` in `/openrtb2/auction.md` mentioning that PBS will consider no currency in the request as being `USD` by default.
Let me know if the doc is clear enough :)

exchange/bidder.go
```diff
@@ -134,36 +134,25 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+					// Empty currency means default currency `USD`
+					bidResponse.Currency = defaultCurrency
 				}
 
-				// TODO: #281 - Once currencies rate conversion is out, this shouldn't be an issue anymore, we will only
-				// need to convert the bid price based on the currency.
-				if firstHTTPCallCurrency == bidResponse.Currency {
+				// Try to get a conversion rate
+				if conversionRate, err := conversions.GetRate(bidResponse.Currency, "USD"); err == nil {
+					// Conversion rate found, using it for conversion
```
dbemiller: Please leave a `TODO: #280` comment somewhere around here.

Before this feature is closed, this code should try to convert every to element of `request.cur`, and use the first one which succeeds. No need to do that in this PR... but I don't want to lose track of it.

exchange/bidder_test.go
```diff
@@ -230,78 +237,389 @@ func TestConnectionClose(t *testing.T) {
 	}
 }
 
-// TestMultiCurrencies makes sure that bidderAdapter.requestBid returns errors if the bidder pass several currencies in case of multi HTTP calls.
+// TestTryGetConversionRate
+func TestTryGetConversionRate(t *testing.T) {
+	// TODO:
```
dbemiller: Should either do this or delete the empty test.

exchange/bidder.go
benjaminch: Yes, just added ! I would prefer put it in another PR (the one doing all the `cur` mapping so it will be cleared maybe.

exchange/bidder_test.go
benjaminch: Yes, sorry about that, just removed. It was meant to test a bidder function to get the conversion rate, but not needed anymore since the API is clear enough, no need to encapsulate more.

bretg:
Note that the primary location of the currency file has changed from

http://currency.prebid.org/latest.json

to

https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json

benjaminch:
Well noted @bretg thanks, I will update the URL in a dedicated PR

dbemiller:
..... d'oh. Sorry @benjaminch . I just merged #786 and caused you a conflict here.

benjaminch:
No worries @dbemiller, it’s expected I need to update references to new URL here as well ;)

benjaminch:
Just updated with the new PrebidJS currency file URL :)

benjaminch:
Thanks :)

PR #763, opened 2018-12-07
Configurable sync URLs

### LINKED ISSUE #509: AMP user sync mixed content
We've noticed a number of user syncs being initialised for Prebid in AMP via RTC are not secure - this is throwing up mixed content errors.

Samples:

http://prebid-match.dotomi.com/prebid/match?rurl=https%3A%2F%2Fprebid.adnxs.com%2Fpbs%2Fv1%2Fsetuid%3Fbidder%3Dconversant%26uid%3D

http://sync.e-planning.net/um?uidhttps%3A%2F%2Fprebid.adnxs.com%2Fpbs%2Fv1%2Fsetuid%3Fbidder%3Deplanning%26uid%3D%24UID

To clarify, these are observed from PrebidAppNexus RTC vendor.

dbemiller:
This would be a breaking change for Prebid Server host companies. It's an alternative to #753 that I think puts the project in a better place moving forward.

## How may this break things?

If merged, Prebid Server host companies will need to update any `config.adapters.{bidder}.usersync_url` config parameters that they define. If you haven't defined any of these, you have nothing to update. If you do define them, keep reading.

If you deploy with any `config.adapters.{bidder}.usersync_url` values defined, you must:

1. Replace `{{gdpr}}` with `{{.GDPR}}`.
2. Replace `{{gdpr_consent}}` with `{{.GDPRConsent}}`.
3. Define the config value to be the _full_ URL that you want Prebid Server to return for that Bidder when `/cookie_sync` is called. Before these changes, some bidders only interpreted these changes as config values as "URL fragments." After these changes, all Bidders will expect the full URL.

## What does this improve?

A few things.

First: It makes Prebid Server config more flexible so that it's easier to run in test environments (e.g. load & integration tests). The config values can be pointed to mock servers without those servers having to be updated whenever a Bidder's actual sync URL changes.

Second:  We've been having some bugs with AMP due to invalid usersync URLs. Publishers are seeing these in the terminal and get worried. Most recently, these are coming from [some code in the 33across usersyncer](https://github.com/prebid/prebid-server/blob/master/adapters/33across/usersync.go#L19). We're looking for ways to minimize these errors. To catch this most recent bug today, we would have to write custom validation logic for this Bidder (`config.adapters.33across.partner_id` is a required config value, but `config.adapters.{everyone-else}.partner_id` is not). After these changes, it'll be easy to implement some simple `isValidURL()` validation logic that works across all the bidders.

Third: As noted above, that `partner_id` is specific to `33across`. This was a totally legitimate addition to the project, given how their sync URL works. However, it also means they had to make changes to the PBS config in order to contribute their Bidder to the project. With the whole URL configurable, it would have been easier for them.

Fourth: It takes a major step towards an architecture that doesn't require Usersyncers for Bidders at all. After these changes, every Usersyncer in the project looks like:

```
func NewBidderSyncer(temp *template.Template) usersync.Usersyncer {
	return adapters.NewSyncer("bidderName", vendorConsentID, temp, syncType)
}
```

In a future change, we can move `vendorConsentID` and `syncType` into [the bidder-info files](https://github.com/prebid/prebid-server/tree/master/static/bidder-info) and remove Usersyncers from the code completely. This would farther simplify things and make it easier to contribute Bidders to the project.

Fifth: It lets us use [Golang Templates](https://golang.org/pkg/text/template/) to populate GDPR info rather than having us implement our own.

This also fixes #509 by making all the defaults use `https`.


dbemiller:
Travis builds will be fixed once #764 merges.

PR #770, opened 2018-12-13
Stop counting invalid bids in the metrics

dbemiller:
This isn't my priority at the moment... but I saw this line which looks, at first glance, like a bug.

Making this PR as a placeholder so it doesn't get forgotten. It seems like bids which are rejected as invalid are still being counted as valid in the metrics.

This needs some investigation, when there's more time. If it is really a bug, also needs a regression test (& maybe some refactoring to make it less likely in the future).


dbemiller:
Should be superceded by #772... but this is the "quick fix" in case that's a bad idea for some reason.

PR #776, opened 2019-01-07
Adding adapter for Yieldmo.

Winston-Yieldmo:


adapters/yieldmo/yieldmo.go
```diff
@@ -0,0 +1,141 @@
...
+	errs = append(errs, errors...)
+
+	return adapterRequests, errors
+}
+
+func (a *YieldmoAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
+	var errs []error
+
+	// Make a copy as we don't want to change the original request
+	reqCopy := *request
```
hhhjort: Please note that this only makes a shallow copy. Anything that is a pointer in `request` will still point to the original datastructure. `[]Imp` is technically a pointer. However PBS makes shallow copies of every `Imp` in the request for every adaptor (which requires a shallow copy of `request` itself, so it can point to the copied slice) so the request you get is already unique to your adapter. You only need to make a copy if you plan on changing the data, and afterward want to access the original values.

adapters/yieldmo/yieldmo.go
```diff
@@ -0,0 +1,141 @@
...
+
+		impExtJSON, err := json.Marshal(impExt)
+		if err != nil {
+			return &errortypes.BadInput{
+				Message: err.Error(),
+			}
+		}
+
+		imp.Ext = impExtJSON
+		request.Imp[i] = imp
```
hhhjort: Given my previous comment, it should be safe to just `request.Imp[i].Ext = impExtJSON` as PBS has already insured you are working on your own private copy of `Imp`, saving some copy steps. I am happy to see care taken against generating race conditions by ensuring parallel routines do not try to modify the same data.

adapters/yieldmo/yieldmo.go
Winston-Yieldmo: Thanks, I've removed the unnecessary copying.

adapters/yieldmo/yieldmo.go
Winston-Yieldmo: Cleaned this up.

PR #778, opened 2019-01-07
Adds checking Legitimate interests in checking vedor purposes in GDPR…

hhhjort:
… PI checks.


dbemiller:
gonna merge this now because it's an important bug, but... should add a regression test when you have the time.

PR #780, opened 2019-01-07
Update E-Planning adapter for new API

Nistenf:
Update E-Planning adapter for new API.
Let us know if you need something clarified.


sebaperez:
Hi @dbemiller , is there an ETA for adapter's update? I know you are busy but we are in a hurry with this

dbemiller:
I'm not involved with the project in any significant way anymore. @hhhjort is the one who'd be reviewing, and @bretg will probably want to be involved when working out a general policy for handling breaking changes to bidder params.

Bidder param breaking changes cause two issues:

1. Existing publisher requests break.
2. Any existing Stored Requests in the databases of every PBS host will suddenly break.

We see no traffic to e-planning on our servers, so we don't really care about either of these... but other host companies might.

dbemiller:
I'm glad the tests are thorough :).

It's not possible to catch this type of issue with unit tests, though :/.

hhhjort:
I think it would be good to drop the requirement for the `ci` in your bidder-params, and either keep some backwards compatibility support in your adapter, or generate an error message from within the adapter saying `ci` is now required. The difficulty is Prebid Server will verify incoming requests against the defined `bidder-params`, and if the request fails for any bidder at that level, the whole request is considered bad and rejected. This means if anyone is using `exchange_id` today, this change will invalidate their whole request (not just the eplanning part) which will be bad behavior.

At a later date you can submit a PR to make `ci` a required parameter once we all feel confident no one is still using the old API (or everyone has had plenty of warning to change). Or if you can confirm that you are presently seeing 0 traffic using `exchange_id` today from prebid server installs, then I may feel better going forward as is.

sebaperez:
@hhhjort I can confirm we have 0 traffic using exchange_id. All partnerships we have are through PBJS adapter.
These updates are for our first integration using PBS

hhhjort:
Ok, then I will let the `bidder-params` change go through. See the inline comments though.

adapters/eplanning/eplanning.go
```diff
@@ -14,102 +18,190 @@ import (
...
+	Height       uint64 `json:"h,omitempty"`
 }
 
 func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
 	errors := make([]error, 0, len(request.Imp))
 	totalImps := len(request.Imp)
-	sourceMapper := make(map[string][]int)
+	spacesStrings := make([]string, 0, totalImps)
+	totalRequests := 0
+	spaceNameToImpID = make(map[string]string)
```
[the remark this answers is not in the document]
Nistenf: The need for persistence is because I need access to that map in the MakeBids function. Is the adapter instance also reused between threads?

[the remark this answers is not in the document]
dbemiller: MakeBids() is really `MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData)`.

`internalRequest` is identical to the `MakeRequests()` argument.

So... it should be possible to re-extract whatever info you need from it, at the cost of some CPU cycles.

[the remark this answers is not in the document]
Nistenf: I did as suggested and extracted the needed information from the internalRequest

sebaperez:
Hello, @dbemiller and @hhhjort when will this PR be able to merge into master? just want to have an estimate on the timeframe. Thank you!

hhhjort:
Hoping to get a second review before merging. It is sort of unfortunate timing as dbemiller is transitioning away from the project, but we are getting new people in. I think I will merge by the end of the week even if I can't get a second review.

sebaperez:
@hhhjort thanks for your update!

adapters/eplanning/eplanning.go
```diff
@@ -14,102 +18,186 @@ import (
...
 	}
 
+	requests := []*adapters.RequestData{&requestData}
+
 	return requests, errors
 }
 
-func verifyImp(imp *openrtb.Imp) (string, error) {
+func cleanName(name string) string {
+	for regexp, replacementString := range cleanNameRegexps {
```
dbemiller: As @hhhjort said, I'm transitioning away from the project... but this would be a fairly major bug. The iteration order on maps isn't guaranteed, so your three regexes might not run in order.

This looks like a problem, because the 2nd regex might create some `_` chars, while the first one doesn't.

On a related note: can you add a regex to your json-schema to validate inputs more strictly, rather than trying to transform them here? If so, that's a much better approach.

Nistenf: If we add a regex to the json-schema, wouldn't nonmatching inputs be rejected? We'd prefer to keep those nonmatching adunits and transform them into what our backend expects, if possible.

dbemiller: yes, you're correct. Nonmatching inputs would be rejected.

The way forward is up to you/@hhjort... I can think of a few reasons to strongly recommend JSON schema validation instead.

1. If you change this logic in any way, it would be a breaking API change for publishers. Since PBS host companies will have params saved in their Stored Request backends, this is a pretty serious issue. No promises that the project maintainers would allow it.

2. JSON schemas are returned by the [/bidders/params endpoint](https://prebid.adnxs.com/pbs/v1/bidders/params)... so a UI, for example, can run this validation before anything gets saved in the Stored Request database. Hardcoded validation inside your adapter won't have that benefit.

3. It's consistent with all the other APIs in PBS today, which are "fail loud and fast."

hhhjort: So you want to commit to supporting invalid adunit names and have your adapter be responsible for doing the correct thing in converting them to valid names?

Nistenf: Yes, we'd prefer the keep the cleanName logic to be consistent with the rest of our systems in which me interact with DFP AdUnits.

PR #787, opened 2019-01-10
PBS-249:  better error messages for invalid JSON

josephveach:
Still planning another test...


endpoints/openrtb2/auction.go
```diff
@@ -905,6 +911,12 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	if hasStoredBidRequest {
 		resolvedRequest, err = jsonpatch.MergePatch(storedRequests[storedBidRequestId], requestJson)
 		if err != nil {
+			if !json.Valid(requestJson) {
+				err = errBadIncomingRequest
+			} else if !json.Valid(storedRequests[storedBidRequestId]) {
+				err = errBadStoredRequest
```
hhhjort: It may be better to have this dynamic, and report back the stored request ID that is causing an issue. More of an issue on the bad stored Imp than the bad stored request, as a request could be referencing multiple stored Imps

endpoints/openrtb2/auction.go
```diff
@@ -879,6 +879,12 @@ func setImpsImplicitly(httpReq *http.Request, imps []openrtb.Imp) {
 	}
 }
 
+var errBadIncomingRequest = fmt.Errorf("Invalid JSON in Incoming Request")
+var errBadStoredRequest = fmt.Errorf("Invalid JSON in Stored Request")
+var errBadIncomingImp = fmt.Errorf("Invalid JSON in Incoming Imp Request")
+var errBadStoredImp = fmt.Errorf("Invalid JSON in Stored Imp Request")
+var errBadDefaultRequest = fmt.Errorf("Invalid Default Request")
```
hhhjort: Probably want to add that the issue is with the server config, not anything related to the request.

`Invalid JSON in Default Request: The server has been configured with an invalid default for all requests.`

endpoints/openrtb2/auction.go
```diff
@@ -924,6 +941,11 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
 		if err != nil {
+			if !json.Valid(imps[idIndices[i]]) {
+				err = errBadIncomingImp
```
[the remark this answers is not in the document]
josephveach: When the incoming JSON is invalid, it would require a lot of trickery to be able to place the path into the error message.  I did add paths for stored requests and stored imps.  For incoming requests, the error message simply states the incoming request had invalid JSON and then displays the specific error and offset within the request that Unmarshal() produces.

endpoints/openrtb2/auction.go
```diff
@@ -885,6 +885,23 @@ func setImpsImplicitly(httpReq *http.Request, imps []openrtb.Imp) {
 	}
 }
 
+func getJsonSyntaxError(testJSON []byte) (bool, string, []error) {
```
hhhjort: Do you ever return anything besides `nil` for `[]error`?

endpoints/openrtb2/auction.go
```diff
@@ -919,6 +946,15 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	if deps.defaultRequest {
 		aliasedRequest, err := jsonpatch.MergePatch(deps.defReqJSON, resolvedRequest)
 		if err != nil {
+			hasErr, Err, _ := getJsonSyntaxError(resolvedRequest)
+			if hasErr {
+				err = fmt.Errorf("Invalid JSON in Incoming Request: %s", Err)
+			} else {
+				hasErr, Err, _ = getJsonSyntaxError(deps.defReqJSON)
+				if hasErr {
+					err = fmt.Errorf("Invalid JSON in Default Alias: %s", Err)
```
hhhjort: Perhaps "Invalid JSON in default request settings: %s"

endpoints/openrtb2/auction.go
josephveach: No.  Should I drop the third return value?

PR #791, opened 2019-01-15
Implements metrics to record GDPR status of cookie sync requests

hhhjort:
We can see per adapter how often an adapter user sync is GDPR blocked in the `/cookie-sync` endpoint.

### DIFF
--- a/endpoints/cookie_sync.go
+++ b/endpoints/cookie_sync.go
@@ -108,7 +108,19 @@ func (deps *cookieSyncDeps) Endpoint(w http.ResponseWriter, r *http.Request, _ h
 	}
 
 	parsedReq.filterExistingSyncs(deps.syncers, userSyncCookie)
+	adapterSyncs := make(map[openrtb_ext.BidderName]bool)
+	for _, b := range parsedReq.Bidders {
+		// assume all bidders will be GDPR blocked
+		adapterSyncs[openrtb_ext.BidderName(b)] = true
+	}
 	parsedReq.filterForGDPR(deps.syncPermissions)
+	for _, b := range parsedReq.Bidders {
+		// surviving bidders are not GDPR blocked
+		adapterSyncs[openrtb_ext.BidderName(b)] = false
+	}
+	for b, g := range adapterSyncs {
+		deps.metrics.RecordAdapterCookieSync(b, g)
+	}
 	parsedReq.filterToLimit()
 
 	csResp := cookieSyncResponse{
--- a/pbsmetrics/config/metrics.go
+++ b/pbsmetrics/config/metrics.go
@@ -6,9 +6,9 @@ import (
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
-	"github.com/prebid/prebid-server/pbsmetrics/prometheus"
-	"github.com/rcrowley/go-metrics"
-	"github.com/vrischmann/go-metrics-influxdb"
+	prometheusmetrics "github.com/prebid/prebid-server/pbsmetrics/prometheus"
+	metrics "github.com/rcrowley/go-metrics"
+	influxdb "github.com/vrischmann/go-metrics-influxdb"
 )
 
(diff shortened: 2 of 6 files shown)


pbsmetrics/config/metrics.go
```diff
@@ -6,9 +6,9 @@ import (
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
-	"github.com/prebid/prebid-server/pbsmetrics/prometheus"
-	"github.com/rcrowley/go-metrics"
-	"github.com/vrischmann/go-metrics-influxdb"
+	prometheusmetrics "github.com/prebid/prebid-server/pbsmetrics/prometheus"
+	metrics "github.com/rcrowley/go-metrics"
+	influxdb "github.com/vrischmann/go-metrics-influxdb"
```
dbemiller: what made the imports change? not that it's a big deal.. things still compile so I assume it still works.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -12,19 +12,20 @@ import (
...
+	imps            *prometheus.CounterVec
+	requests        *prometheus.CounterVec
+	reqTimer        *prometheus.HistogramVec
+	adaptRequests   *prometheus.CounterVec
+	adaptTimer      *prometheus.HistogramVec
+	adaptBids       *prometheus.CounterVec
+	adaptPrices     *prometheus.HistogramVec
+	adaptErrors     *prometheus.CounterVec
+	cookieSync      prometheus.Counter
+	adaptCookieSync *prometheus.CounterVec
```
dbemiller: I'd personally be a bit confused by these names... in the code,  yes, but more importantly in the Prometheus metrics themselves.

In the dashboard you'd see `adapter_cookie_sync` and `usersync_total`, but the interesting difference between the two is that one logs `/cookie_sync` calls while the other logs `/setuid` stuff.

Maybe something like `cookie_sync_returns` and `setuid_attempts`? Or something else more clear.

endpoints/cookie_sync.go
```diff
@@ -108,7 +108,19 @@ func (deps *cookieSyncDeps) Endpoint(w http.ResponseWriter, r *http.Request, _ h
...
+	}
 	parsedReq.filterForGDPR(deps.syncPermissions)
+	for _, b := range parsedReq.Bidders {
+		// surviving bidders are not GDPR blocked
+		adapterSyncs[openrtb_ext.BidderName(b)] = false
+	}
+	for b, g := range adapterSyncs {
+		deps.metrics.RecordAdapterCookieSync(b, g)
+	}
 	parsedReq.filterToLimit()
```
dbemiller: This logic might make some the metrics misleading...  cause (as an outsider) I would probably expect the `/cookie_sync allowed by GDPR` numbers to match up with the `/setuid attempts` numbers.

Or at least, it seems _useful_ if the numbers lined up, cause we could use them to diagnose bidders with broken sync endpoints.

Not sure how you want to address these, or if it doesn't matter much to you... but maybe add another `true/false` label for `limited_by_count`?

I can also understand the argument for limiting the number of labels/metrics for performance,  though, so... your call.

hhhjort: Well, there is a problem with limited by count. The client could implement its own "limit syncs" function after getting the syncs, especially if they were not aware of the functionality within the endpoint. Or may slowly fire off the syncs one at a time and may not get them all off before a page load breaks the chain and triggers a new `/cookie-sync` call. So I don't expect it will be possible for the counts to line up even accounting for `limited_by_count`.

dbemiller: ah... yeah, good points.

pbsmetrics/config/metrics.go
hhhjort: That is a good question, I did not consciously make this change.

Line 40 below shows that the old prometheus import was being accessed by the label `prometheusmetrics` before the change, so I am guessing new linting in either upgraded go package or VSCode now triggers the move to make inobvious module names explicit.

PR #796, opened 2019-01-22
Sonobi Adapter

JonGoSonobi:


adapters/sonobi/sonobi.go
```diff
@@ -0,0 +1,296 @@
...
+	var err error
+
+	var adapterRequests []*adapters.RequestData
+
+	for _, imp := range request.Imp {
+		// Sonobi doesn't allow multi-type imp. Banner takes priority over video.
+		if imp.Banner != nil {
+		} else if imp.Video != nil {
+		} else {
+			err := fmt.Errorf("Sonobi only supports banner and video imps. Ignoring imp id=%s", imp.ID)
```
hhhjort: The base PBS should now take care of this for you. It will not pass any media type objects you don't support, and drop the particular imp if nothing is left that you do.

adapters/sonobi/sonobi.go
```diff
@@ -0,0 +1,296 @@
...
+// MakeRequests Makes the OpenRTB request payload
+func (a *SonobiAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
+	var errs []error
+	var sonobiExt openrtb_ext.ExtImpSonobi
+	var err error
+
+	var adapterRequests []*adapters.RequestData
+
+	for _, imp := range request.Imp {
+		// Sonobi doesn't allow multi-type imp. Banner takes priority over video.
```
hhhjort: I think this means you want code someplace to set the video object to `nil` if the banner object exists. Or it may just be a note about what your bidder will do if it gets both in a bid request from PBS?

josephveach: I find this comment confusing, too.  It led me to expect that there should be some sort of code for the immediately following lines, where imp.Banner and imp.Video are non-nil.  Instead, it seems to simply refer to the fact that getMediaTypeForImp(), near the end of this file, uses banner if both banner and video are set. 

If there really is nothing to be done when imp.Banner and imp.Video are non-nil, it would be much clearer if the code either had comments indicating so (between lines 182/183 and between 183/184) or if the tests were rewritten to be "if imp.Banner == nil && imp.Video == nil".

adapters/sonobi/sonobi.go
```diff
@@ -0,0 +1,296 @@
...
+		if imp.Banner != nil {
+		} else if imp.Video != nil {
+		} else {
+			err := fmt.Errorf("Sonobi only supports banner and video imps. Ignoring imp id=%s", imp.ID)
+			errs = append(errs, err)
+		}
```
josephveach: Should there be a "continue" statement here?  If there was an error, it doesn't seem that the loop should continue.

adapters/sonobi/sonobi.go
josephveach: In case it's relevant, note that this instance of "err" (line 185, 186) is not the same as the one declared at the top of the function (line 176), which is the one used at lines 195 and 200, below.  I don't think there's anything technically incorrect here, but it's potentially confusing for future maintenance.  Perhaps renaming this one, or just changing the ":=" to "=" would be less risky!?

JonGoSonobi: Must have been left over code from a while back when the first efforts to make this adapter was being made. I removed the check for media types since they will be filtered to the ones that Sonobi supports before hand.

adapters/sonobi/sonobi.go
JonGoSonobi: Must have been left over code from a while back when the first efforts to make this adapter was being made. I removed the check for media types since they will be filtered to the ones that Sonobi supports before hand.

adapters/sonobi/sonobi.go
```diff
@@ -0,0 +1,292 @@
...
+	mediaType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner == nil && imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			return mediaType
+		}
+	}
+	return mediaType
```
josephveach: If we get to this final return, it means we failed to find a match for impID.  Given that, is it correct to return mediaType?

JonGoSonobi: True. This case should never happen though, I think. Anyways, I pushed a commit to handle it just incase.

adapters/sonobi/sonobi.go
```diff
@@ -0,0 +1,300 @@
...
+		err = json.Unmarshal(unit.Params, &params)
+		if err != nil {
+			return nil, err
+		}
+
+		// Fixes some segfaults. Since this is legacy code, I'm not looking into it too deeply
+		if len(sonobiReq.Imp) <= i {
+			break
+		}
+		sonobiReq.Imp[i].TagID = params.TagID
```
josephveach: No test of the contents of params has been performed.  Is it safe to do this assignment without checking whether params.TagID is set / valid?

adapters/sonobi/sonobi.go
```diff
@@ -0,0 +1,300 @@
...
+	var sonobiExt openrtb_ext.ExtImpSonobi
+	var err error
+
+	var adapterRequests []*adapters.RequestData
+
+	// Sonobi currently only supports 1 imp per request to sonobi.
+	// Loop over the imps from the initial bid request to form many adapter requests to sonobi with only 1 imp.
+	for _, imp := range request.Imp {
+		// Make a copy as we don't want to change the original request
+		reqCopy := *request
```
hhhjort: `request` and each `imp` are already shallow copies when passed to the adapters, so no need to recopy them. Good work in trying to avoid race conditions though, thanks! :)

adapters/sonobi/sonobi.go
hhhjort: Go always initializes struct data to a valid null value. So the assignment will not result in a segfault. Beyond that, the actual bidder is likely in a better position to say if a `Tag.ID` is a valid ID or not. Also the bodder paramas JSON spec below (in static) ensures that there was a `TagID` property in the `ext`.

adapters/sonobi/sonobi.go
```diff
@@ -0,0 +1,300 @@
...
+		reqCopy.Imp = append(make([]openrtb.Imp, 0), imp)
```
hhhjort: It may or may not be more efficient to use `append(make([]openrtb.Imp, 0, 1), imp)` This explicitly tells make to ensure memory is reserved for one `imp`, so `append()` does not need to allocate any new memory.

adapters/sonobi/sonobi.go
JonGoSonobi: I make a copy of the request here so that I don't change the request.Imp that I'm currently iterating over. 

Changing reqCopy to just modifying "request" directly causes only 1 imp to be returned in a bid request.

adapters/sonobi/sonobi.go
josephveach: Okay, it makes sense to not burden PBS with validation, so given that params.TagID exists, I have no objection here.

JonGoSonobi: Aren't params validated before we reach this point by the static/bidder-params/sonobi.json spec?

Also during my test I dont ever actually hit this function when I put a debug point at the top of "Call". When is this function used? I wrote this function a few months back before the definition of an adapter changed to using MakeRequests. Only reason I left is it because its a required function in the past. Seems like its not anymore. So I'll remove Call

adapters/sonobi/sonobi.go
JonGoSonobi: Yes thats the purpose. Each iteration of the incoming request's imps needs to produce multiple outgoing requests to Sonobi that are similar, except with only 1 imp. 

Not copying the incoming request on each iteration causes an issue where the final returned bids back to /openrtb2/auction only include 1 of the potentially many bids.

josephveach: Oh, sorry, I misunderstood your previous comment to mean you were about to change from using reqCopy to modifying request.

hhhjort:
The `Call()` function is used in the legacy endpoint. We do want to deprecate it, but it has been slow getting people off of it. Just so you know why there is one in some adapters. :)

Issue #755, opened 2018-11-28
Improved support for interstitial ads


hhhjort:
Typically interstitial ads are full screen ads. Typical screen sizes don't always map to common ad sizes. Currently the way around this is to include a list of acceptable dimensions in the `format` object. This is less than ideal as publishers will have to guess what sizes will be available with a high demand, and bidders may have to contend with a long list of sizes if publishers go with a shotgun approach in hitting a high demand size for their space.

A better solution would be to define a minwidth/minheight percent setting to pass on to bidders, which could then optimize their search for demand to cover the range of acceptable sizes rather than working off a list of sizes. This will save tuning the acceptable size list as interstitials become more common and device sizes proliferate, which will change the ad size landscape and optimal solutions over time.

hhhjort:
A couple of locations for these parameters would be `bidrequest.device.ext.prebid.interstitial.minwidthperc/minheightperc` or `bidrequest.ext.prebid.minwidthperc`. I suggest the device object so they can live in close proximity to the screen width/height they relate to.

It might also make sense to attach them under `imp.ext.prebid` since they are only relevant to interstitial impressions. However the minimum allowable percentage is unlikely to be a per imp setting, rather be consistent across a publisher or a site/app. Keeping it outside of the `imp` context makes config management a little simpler, as it will be associated with other publisher/site configs rather than the imp level configs.

dbemiller:
+1 to `request.device.ext...` 

Everything about this sounds great.

bretg:
Please help me connect the dots for how bidders + DSPs are supposed to utilize minwidthperc/minheightperc.

I see imp.instl in OpenRTB 2.5, but that's just a flag. From my reading of the standard, someone has to be responsible for actually defining sizes.

hhhjort:
The implementation is to say that any size between the min size and the screen size is a valid size for the auction. The savings on the bidder is that the validity check can simply be does the creative fall in the allowed range, rather than having a list of all possible sizes that fit in the range to be checked in turn. Saving on the publisher is that they don't need to revisit the list of allowed sizes as screen sizes and popular ad sizes changes change over time.

This may not always work out well for the bidders, depending on how entrenched the current logic for matching ad sizes is. But supporting ranges will likely be a good thing to support in the long run as I can see native taking advantage of size ranges for creative content.

hhhjort:
[...]

PR #797, opened 2019-01-22
Interstitial support for banner Imps

hhhjort:
This implements #755


config/interstitial.go
```diff
@@ -0,0 +1,267 @@
...
+// configurable solution once the need develops and the use cases can be defined.
+
+type InterstitialSize struct {
+	Width  uint64
+	Height uint64
+}
+
+type InterstitialSizes []InterstitialSize
+
+var ResolvedInterstitialSizes = InterstitialSizes{
```
mkendall07: where was the size list sourced from? Might be worth adding a comment.

hhhjort: It was sourced from delivery statistics from Appnexus. A list of sizes sorted by size (larger first) and frequency(more common sizes first) with what seemed to be a good weighting factor balancing the two factors.

PR #799, opened 2019-01-26
* added gzip support

olafbuitelaar:
I would like to optionally add gzipped responses.
With this PR it's possible to set enable_gzip: true in the pbs.yml, which makes the server gzip the response when possible.
the default behavior is unchanged.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -11,6 +11,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/NYTimes/gziphandler"
 	"github.com/golang/glog"
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/pbsmetrics"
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -11,6 +11,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/NYTimes/gziphandler"
```
hhhjort: I believe we will need an addition to Gopkg.toml since the gziphandler is not part of the project yet.

Otherwise a nice simple addition, sorry for letting it sit so long.

olafbuitelaar: Sorry for missing that, i've included the dependency. I think for now minor updates are ok, checking the state of the lib, outstanding PR's and velocity..their last update was 6 days ago, which seemed ok.

PR #801, opened 2019-01-29
Consumable adapter implementation

nealeu:
Hi,

We'd like to add the Consumable adapter in this PR. 

We believe we've met all the criteria for adding an adapter.

Also, as we wanted to cover HTTP headers in our testing, we've included the commits from #795 which implement checking of those parts of the test files. 

Please let us know if there's anything else you need for this to be merged.

Thanks,

Neale.


adapters/consumable/consumable.go
```diff
@@ -0,0 +1,292 @@
...
+			AdTypes:   getSizeCodes(impression.Banner.Format), // was adTypes: bid.adTypes || getSize(bid.sizes) in prebid.js
+		}
+	}
+
+	bodyBytes, err := json.Marshal(body)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	println("== header ==")
```
hhhjort: Were these supposed to be `fmt.Printf()`s?

hhhjort: Otherwise please only log to console messages that the PBS host needs to read in production. Else the logs become unreadable due to spam and useless for server maintenance.

nealeu: Are you referring to the println()?

Sorry. That should have been removed.  Will sort them.

nealeu:
Hi,
That should do it, subject to the build being okay.
Cheers,
Neale

adapters/consumable/consumable.go
```diff
@@ -0,0 +1,288 @@
...
+	bidderResponse := adapters.NewBidderResponse()
+	var errors []error
+
+	for impID, decision := range serverResponse.Decisions {
+		imp := getImp(impID, internalRequest.Imp)
+		if imp == nil {
+			errors = append(errors, &errortypes.BadServerResponse{
+				Message: fmt.Sprintf("ignoring bid id=%s, request doesn't contain any impression with id=%s", "TODO: bid.ID", impID),
+			})
+			fmt.Printf("%s", errors[0])
```
hhhjort: We should drop this error print ... the error should be visible in the PBS response and it is not clear that this is actionable by the PBS host.

adapters/consumable/consumable.go
```diff
@@ -0,0 +1,288 @@
...
+			return &imp
+		}
+	}
+	return nil
+}
+
+func extractExtensions(impression openrtb.Imp) (*adapters.ExtImpBidder, *openrtb_ext.ExtImpConsumable, []error) {
+	var bidderExt adapters.ExtImpBidder
+	if err := json.Unmarshal(impression.Ext, &bidderExt); err != nil {
+		fmt.Printf("err: %s", err) // TODO: DELETE
```
hhhjort: This one even has a TODO :)

adapters/consumable/consumable.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, nil, []error{&errortypes.BadInput{
+			Message: err.Error(),
+		}}
+	}
+
+	var consumableExt openrtb_ext.ExtImpConsumable
+	if err := json.Unmarshal(bidderExt.Bidder, &consumableExt); err != nil {
+		fmt.Printf("err: %s", err) // TODO: DELETE
```
hhhjort: And here

nealeu: Hmm.. I don't think I get a medal for this.

Sorry.. you shouldn't be left picking this up at review.  I'll run everything up this evening.  I just did a search for println() before ...

adapters/consumable/consumable.go
```diff
@@ -0,0 +1,284 @@
...
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("error while decoding response, err: %s", err),
+		}}
+	}
+
+	bidderResponse := adapters.NewBidderResponse()
+	var errors []error
+
+	for impID, decision := range serverResponse.Decisions {
+		imp := getImp(impID, internalRequest.Imp)
```
josephveach: The value of imp is used only when both decision.Pricing and decision.Pricing.ClearPrice are non-nil.  Please move this code (current 206-211) into the code that runs after the check of Pricing and ClearPrice (current 215) so that it is executed only when necessary.  This will save the cost of iterating through the imps in getImp() in some cases.

nealeu: Good spot.  Change pushed

PR #803, opened 2019-02-04
Brightroll bidder changes for customized request

smithaammassamveettil:
Accommodating some customized bid request changes for brightroll bidder

### DIFF
--- a/adapters/brightroll/brightroll.go
+++ b/adapters/brightroll/brightroll.go
@@ -89,6 +89,7 @@ func (a *BrightrollAdapter) MakeRequests(requestIn *openrtb.BidRequest) ([]*adap
 	}
 
 	request.AT = 1 //Defaulting to first price auction for all prebid requests
+
 	if brightrollExt.Publisher == "adthrive" {
 		request.BCat = getBlockedCategoriesForAdthrive()
 	}
--- a/adapters/brightroll/brightroll.go
+++ b/adapters/brightroll/brightroll.go
@@ -152,9 +152,9 @@ func (a *BrightrollAdapter) MakeBids(internalRequest *openrtb.BidRequest, extern
 	return bidResponse, nil
 }
 
-//Cafemedia customized request, need following blocked categories
+//customized request, need following blocked categories
 func setBcat(request *openrtb.BidRequest) {
-	request.BCat = []string{"IAB8-5", "IAB8-18", "IAB15-1", "IAB7-30", "IAB14-1", "IAB22-1", "IAB3-7", "IAB7-3", "IAB14-3", "IAB11", "IAB11-1", "IAB11-2", "IAB11-3", "IAB11-4", "IAB11-5", "IAB23", "IAB23-1", "IAB23-2", "IAB23-3", "IAB23-4", "IAB23-5", "IAB23-6", "IAB23-7", "IAB23-8", "IAB23-9", "IAB23-10", "IAB7-39", "IAB9-30", "IAB7-44", "IAB25", "IAB25-1", "IAB25-2", "IAB25-3", "IAB25-4", "IAB25-5", "IAB25-6", "IAB25-7", "IAB26", "IAB26-1", "IAB26-2", "IAB26-3", "IAB26-4"}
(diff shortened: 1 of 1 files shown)


smithaammassamveettil:
@dbemiller Requesting any pointers on this. When the build runs , receiving below error. Seems like a race condition situation. But I believe makeRequests() method is working on a copy of request object all the time. The error message is not saying any details about the issue. Not sure why/where  it is happening. Is there any other way I can troubleshoot this?
0.72s$ ./validate.sh --nofmt --cov --race 10
adapters/brightroll/brightroll.go

smithaammassamveettil:
@dbemiller I can see above build passed already by clicking on the "Details" link, taking me to https://travis-ci.org/prebid/prebid-server/builds/490218085?utm_source=github_status&utm_medium=notification. But still status showing as "The Travis CI build is in prog..."

smithaammassamveettil:
Can I get a review on this Pull Request please...

dbemiller:
@smithaammassamveettil I rarely work on this project anymore. @hhhjort is leading it now.

At the moment his priority is to onboard several new engineers to the project. I'm sure he'll get to it as soon as he has bandwidth.

smithaammassamveettil:
@dbemiller Thanks for letting me know that. I will wait for the review from them then.

smithaammassamveettil:
@hhhjort, Can I please get a review on this?

smithaammassamveettil:
@hhhjort , Thanks for your review and approval. Appreciate it. Looks like I dont have permission to merge to master. Can you please merge the PR as well? When can I expect, this change in Prebid production for live traffic testing?

hhhjort:
We prefer to have 2 reviews before merging. Hope to get this out sometime this week.

smithaammassamveettil:
@hhhjort Whoelse should I request here for additional reviews?

hhhjort:
I have already asked Joe. :)

adapters/brightroll/brightroll.go
```diff
@@ -37,13 +65,21 @@ func (a *BrightrollAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapte
...
 			}
+			if brightrollExt.Publisher == "adthrive" {
+				bannerCopy.BAttr = getBlockedCreativetypesForAdThrive()
+
+			}
 			request.Imp[i].Banner = &bannerCopy
 			validImpExists = true
 		} else if request.Imp[i].Video != nil {
 			validImpExists = true
+			videoCopy := *request.Imp[i].Video
```
josephveach: Setting videoCopy is unnecessary unless the comparison on the following "if" evaluates to true.  Please move this line down into the "then" portion of the following "if". 

Likewise, the use of bannerCopy above could be similarly optimized, though it's a little more complicated because there are two cases where it might be needed.  Please fix that, too, assuming that a significant percentage of the time bannerCopy will not be changed.

Finally, unrelated to this Pull Request but FWIW:  is sufficient testing of the width and height values being done in the unchanged code related to a banner ad?  For example, are there no worries of only bannerCopy.W or bannerCopy.H being set without the other, nor worries of bannerCopy.Format[0] lacking one or both of W or H?

smithaammassamveettil: Updated for videoCopy setting as you suggested. 
Using bannerCopy as such, as it is kind of validation section.

josephveach: Can you please elaborate on bannerCopy?  Its value is changed in only two cases:  (a) H and W are both null and Format is set, and (b) Publisher is adthrive.  I don't know but I would guess these two cases are more the exception than the rule.  If so, in the interest of performance, it is not desirable to copy a structure, make no change to it, and copy it back on a busy server.

smithaammassamveettil: Copying the value to W and H is going to happen all the time in our case. Since our system not supporting oRTB2.5 for now. So need to create the bannerCopy whenever impression is of type banner.

smithaammassamveettil:
Thanks for approving @josephveach. 
@hhhjort Can you please review again? and please merge if you are going to approve. Thanks!

smithaammassamveettil:
Thank you all.. @hhhjort When can I expect this merged change in Prebid production server?

hhhjort:
Probably tomorrow.

PR #805, opened 2019-02-08
BeachFront new cookie sync endpoint.

muncha:
This updates to the new cookie sync endpoint. The existing endpoint will continue to work for now.


muncha:
This updates the beachfront cookie sync endpoint from the stop gap endpoint to a new dedicated endpoint. Also, I cleaned up a bit of error handling on a bad response.

muncha:
@hhhjort - Could you give an estimate of when this might be live on appnexus' server?

hhhjort:
We are having some issues with our test environment. I am hoping to get it out today or tomorrow.

hhhjort:
It has been released

muncha:
@hhhjort I'm still seeing the previous url (https://sync.bfmio.com/syncb?...) in the response for beachfront when I post to https://prebid.adnxs.com/pbs/v1/cookie_sync instead of the new endpoint (https://sync.bfmio.com/sync_s2s?...). Do you know why? Could this be caching on your server?

hhhjort:
Oh yes, thanks for catching that. We have it configured directly in the config file rather than using the default because of some networking details on our end. And that wasn't updated after the release.

patmmccann:
confirmed changed, thanks!

PR #808, opened 2019-02-12
Fix e-planning sync url and improve request parameters

Nistenf:
We were contacted by Hans Hjort regarding an apparently missing equals sign in our default user sync url which led us to realize that the original implementation was using a wrong url. This PR fixes the url and changes a couple of our bidder request parameters.


Nistenf:
Hi @hhhjort, the build seems to have failed due to a download error, could it be rerun? Thanks

```
$ curl -L -s https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 -o $GOPATH/bin/dep
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
The command "curl -L -s https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 -o $GOPATH/bin/dep" failed and exited with 56 during .
```

hhhjort:
Unfortunately I am not finding where I can kick off a rerun. I can confirm that your analysis seems correct and the rerun should fix it. If you have a pointer on how to trigger the rerun, please share it, otherwise a minor commit should trigger a new run on its own.

Nistenf:
I merged upstream to trigger a new run

config/config.go
```diff
@@ -325,7 +325,7 @@ func (cfg *Configuration) setDerivedDefaults() {
...
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderBrightroll, "https://pr-bh.ybp.yahoo.com/sync/appnexuspbs?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&url="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dbrightroll%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24%7BUID%7D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConversant, "https://prebid-match.dotomi.com/prebid/match?rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconversant%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
-	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderEPlanning, "https://sync.e-planning.net/um?uid"+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Deplanning%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderEPlanning, "https://ads.us.e-planning.net/getuid/1/5a1ad71d2d53a0f5?"+externalURL+"/setuid?bidder=eplanning&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&uid=$UID")
```
hhhjort: Is there a reason not to `QueryEscape` the external URL?

Nistenf: Because our backend, for this endpoint, takes everything after the ? as the redirect url and does not unescape it

PR #812, opened 2019-02-14
pass bid request currency to adform adserver - default to USD

Aadeshp:


adapters/adform/adformtest/supplemental/user-nil.json
```diff
@@ -44,7 +44,7 @@
   "httpCalls": [
     {
       "expectedRequest": {
-        "uri": "http://adx.adform.net/adx?CC=1&fd=1&gdpr=1&gdpr_consent=abc2&ip=&pt=gross&rp=4&stid=&bWlkPTE"
+        "uri": "http://adx.adform.net/adx?CC=1&fd=1&gdpr=1&gdpr_consent=abc2&ip=&pt=gross&rp=4&stid=&bWlkPTEmcmN1cj1VU0Q"
```
mkendall07: curious why this changed?

Aadeshp: we have to pass `rcur` in the base64 encoded tag blob, so the base64 blob is just a bit longer

mkendall07: ah gotcha.

Issue #388, opened 2018-03-07
Handle setting up config errors in adapters more elegantly


dbemiller:
One of our Mobile engineers tried to run an Audience Network auction locally. Since they don't define [a default config param](https://github.com/prebid/prebid-server/blob/master/pbs_light.go#L729), it failed with an obtuse error message.

I improved this for Audience Network in #332, but... I believe we can handle this more generally, for all adapters, at a global level (e.g. the `exchange/adapter_map.go`).

dmitriyshashkin:
Another case when this can be useful is when someone tries to use index exchange adapter without specifying the endpoint. Right now this doesn't produce any errors, the only clue that something might be wrong is `'uri': ''` in `httpcalls` part of the response. And it's not quite easy without diving deep into the code to understand what's wrong. By default indexexchange's endpoint is not defined, so anyone who tries to use index in self-hosted PBS will encounter this problem.

PR #817, opened 2019-02-26
PBS-68: Disable Rubicon and AudienceNetwork adapters by default

mansinahar:
* Validate adapter endpoints and user sync URLs at start up

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,12 +6,15 @@ import (
 	"net/url"
 	"reflect"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/prebid/prebid-server/macros"
 	"github.com/prebid/prebid-server/openrtb_ext"
-	"github.com/prebid/prebid-server/util"
 	"github.com/spf13/viper"
+
+	validator "github.com/asaskevich/govalidator"
 )
 
 // Configuration
@@ -175,7 +178,7 @@ const (
 	dummyHost        string = "dummyhost.com"
 	dummyPublisherID int    = 12
 	dummyGDPR        string = "0"
-	dummyGDPRConsent string = "some gdpr consent string"
+	dummyGDPRConsent string = "someGDPRConsentString"
 )
 
 type Adapter struct {
@@ -204,31 +207,34 @@ type Adapter struct {
 	Disabled bool `mapstructure:"disabled"`
 }
 
-// validateAdapterEndpoint validates to make sure that an adapter has a valid endpoint
+// validateAdapterEndpoint makes sure that an adapter has a valid endpoint
 // associated with it
 func validateAdapterEndpoint(endpoint string, adapterName string, errs configErrors) configErrors {
 	if endpoint == "" {
-		errs = append(errs, fmt.Errorf("There's no default endpoint available for %s. Calls to this bidder/exchange will fail. "+
+		return append(errs, fmt.Errorf("There's no default endpoint available for %s. Calls to this bidder/exchange will fail. "+
 			"Please set adapters.%s.endpoint in your app config", adapterName, adapterName))
-	} else {
-		// Check if endpoint has macros. If it does then resolve with dummy values
-		// for validation
-		if util.ContainsMacro(endpoint) {
-			endpointTemplate, err := util.BuildTemplate(endpoint)
-			if err != nil {
-				errs = append(errs, fmt.Errorf("Invalid endpoint template: %s. %s", endpoint, err.Error()))
-				return errs
-			} else {
-				if endpoint, err = util.ResolveMacros(*endpointTemplate, util.EndpointTemplateParams{Host: dummyHost, PublisherID: dummyPublisherID}); err != nil {
-					errs = append(errs, fmt.Errorf("Unable to resolve endpoint: %s. %s", endpoint, err.Error()))
-					return errs
-				}
-			}
-		}
+	}
 
-		if _, err := url.ParseRequestURI(endpoint); err != nil {
-			errs = append(errs, fmt.Errorf("The endpoint specified for %s is invalid. %s", adapterName, err.Error()))
-		}
+	// Create endpoint template
+	endpointTemplate, err := template.New("endpointTemplate").Parse(endpoint)
+	if err != nil {
+		return append(errs, fmt.Errorf("Invalid endpoint template: %s for adapter: %s. %v", endpoint, adapterName, err))
+	}
+	// Resolve macros (if any) in the endpoint URL
+	resolvedEndpoint, err := macros.ResolveMacros(*endpointTemplate, macros.EndpointTemplateParams{Host: dummyHost, PublisherID: dummyPublisherID})
+	if err != nil {
+		return append(errs, fmt.Errorf("Unable to resolve endpoint: %s for adapter: %s. %v", endpoint, adapterName, err))
+	}
+	// Validate the resolved endpoint
+	//
+	// Validating using both IsURL and IsRequestURL because IsURL allows relative paths
+	// whereas IsRequestURL requires absolute path but fails to check other valid URL
+	// format constraints.
+	//
+	// For example: IsURL will allow "abcd.com" but IsRequestURL won't
+	// IsRequestURL will allow "http://http://abcd.com" but IsURL won't
+	if !validator.IsURL(resolvedEndpoint) || !validator.IsRequestURL(resolvedEndpoint) {
+		errs = append(errs, fmt.Errorf("The endpoint: %s for %s is not a valid URL", resolvedEndpoint, adapterName))
 	}
 	return errs
 }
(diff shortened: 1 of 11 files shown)


mansinahar:
This PR addresses https://github.com/prebid/prebid-server/issues/388. 

- It's hard to validate all config params for adapters on a global level because there are some params which are required only by specific adapters. For example `partnerID` is only required by AudienceNetwork and `XAPI` is only required by Rubicon. For this reason, this PR proposes to disable these special adapters by default and relies on the author to enable them and configure them correctly if they wish to use it. It also adds validation for endpoints and user_sync URLs since each adapter must have a valid endpoint (required) and a valid user_sync URL (if there is one) associated with it.

- This PR also adds some common utility functions for resolving macros in an effort to have a consistent way of doing so across the code base.

util/macros.go
```diff
@@ -0,0 +1,48 @@
+package util
```
hhhjort: Lets not create a "util" package ... nobody will know what to expect inside of it, and nobody will know how to find the code inside if they weren't already familiar with it.

Maybe a "macros" or "templates" package?

mansinahar: @hhhjort Sure. Makes sense! I will update the PR.

mansinahar: @hhhjort moved macros to it's own package and added some unit tests for it.

macros/macros.go
```diff
@@ -0,0 +1,48 @@
...
+type UserSyncTemplateParams struct {
+	GDPR        string
+	GDPRConsent string
+}
+
+// MacroRegExp is the regexp to check if a string contains a macro or not
+const MacroRegExp string = `\{\{\.(\w+)\}\}`
+
+// BuildTemplate builds a text template for a given string
+func BuildTemplate(templateStr string) (*template.Template, error) {
```
dbemiller: Just food for thought... compare these two equivalent lines of code from a caller:

```
t, err := macros.BuildTemplate(s)
t, err := template.New("aTemplate").Parse(s)
```

Are the keystrokes it saves you worth the cost of testing + maintaining it?

Personally I don't think so, but... your call in the end.

macros/macros.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return "", err
+	}
+	res := strBuilder.String()
+	return res, nil
+}
+
+// ContainsMacro checks and returns if the provided string contains a macro or not
+func ContainsMacro(str string) bool {
+	r, _ := regexp.Compile(MacroRegExp)
```
dbemiller: I think this function can be deleted (see other comments). But if not, you should `regexp.MustCompile` it somewhere outside the function. Regexes execute fast, but compile slow. No need to re-compile it on every execution.

macros/macros_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	for _, test := range testCases {
+		res, err := ResolveMacros(test.aTemplate, test.params)
+
+		if test.hasError {
+			assert.NotNil(t, err, "Error shouldn't be nil")
+			assert.Empty(t, res, "Result should be empty")
+		} else {
+			assert.Nil(t, err, "Err should be nil")
+			assert.Equal(t, res, test.result, fmt.Sprintf("String after resolving macros should be %s", test.result))
```
dbemiller: no need to `fmt.Sprintf` here... the `assert.*` methods have this functionality already... so you can just:

```
assert.Equal(t, res, test.result, "String after resolving macros should be %s", test.result)
```

util/macros.go
dbemiller: May be worth reading this someday: https://blog.golang.org/package-names

It mentions this... but also has lots of other advice.

macros/macros.go
mansinahar: Yeah, I agree. This is more of a leftover from another approach that I was trying. Will remove that method. Thanks for pointing it out!

macros/macros_test.go
mansinahar: Ah, good to know. Thanks!

config/config.go
```diff
@@ -195,6 +204,74 @@ type Adapter struct {
...
+// validateAdapterEndpoint validates to make sure that an adapter has a valid endpoint
+// associated with it
+func validateAdapterEndpoint(endpoint string, adapterName string, errs configErrors) configErrors {
+	if endpoint == "" {
+		errs = append(errs, fmt.Errorf("There's no default endpoint available for %s. Calls to this bidder/exchange will fail. "+
+			"Please set adapters.%s.endpoint in your app config", adapterName, adapterName))
+	} else {
+		// Check if endpoint has macros. If it does then resolve with dummy values
+		// for validation
+		if macros.ContainsMacro(endpoint) {
```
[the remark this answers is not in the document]
mansinahar: Well, yes. BuildTemplate and ResolveMacros would still work and I agree that it won't affect performance because this is at start up. But, performing these two steps for each and every adapter endpoint/userSync URL which may or may not have macros to start with seems like not needed to me, no?

macros/macros.go
mansinahar: Yes, I agree. Thanks for that suggestion! Have made that change.

config/config.go
```diff
@@ -195,6 +204,74 @@ type Adapter struct {
...
+			endpointTemplate, err := macros.BuildTemplate(endpoint)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Invalid endpoint template: %s. %s", endpoint, err.Error()))
```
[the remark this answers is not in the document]
mansinahar: I have changed that to `fmt.Errorf("Error parsing template: %s", err.Error())` but not totally sure if that's any better. Open to suggestions on this one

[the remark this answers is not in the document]
dbemiller: oh... no, I was referring to the two lines together.

These two statements are equivalent:

```
errs = append(errs, err)
return errs
```

and

```
return append(errs, err)
```

Your `fmt.Errorf` was fine either way. Only suggestion I'd have there is to use `%v`  and `err` . as  the arg, rather than `%s` and `err.Error()`

config/config.go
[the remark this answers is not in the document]
dbemiller: Yeah, it's debatable. It'd cost the machine some cycles, but also reduce the [cyclomatic complexity](https://docs.codeclimate.com/docs/cyclomatic-complexity) of your function, which correlates to how many test cases you need to write to get full coverage.

My other worry is that the adapter code ultimately runs `macros.ResolveMacros` _without_ this check. So if Go's templates have some other restriction on string formats, you might actually skip some important validation here.

For example, suppose Go templates didn't allow `/` unless it were escaped like `\/`. If an adapter expects a template, but the value in the config doesn't have `{{ }}` vars, it might crash with an obscure error message because the validation here would miss it.

config/config.go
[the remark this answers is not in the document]
mansinahar: Ah ok! Looks like I misread your previous comment. Sorry, my bad!

config/config.go
```diff
@@ -195,6 +205,73 @@ type Adapter struct {
...
+	} else {
+		// Check if endpoint has macros. If it does then resolve with dummy values
+		// for validation
+		if macros.MacroRegExp.MatchString(endpoint) {
+			endpointTemplate, err := template.New("endpointTemplate").Parse(endpoint)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("Error parsing template: %s", err.Error()))
+				return errs
+			}
+			if endpoint, err = macros.ResolveMacros(*endpointTemplate, macros.EndpointTemplateParams{Host: dummyHost, PublisherID: dummyPublisherID}); err != nil {
```
josephveach: I believe it's better to use "_" rather than endpoint for the return value from ResolveMacros() in the above line (same comment in the next function, line 248).  ResolveMacros() currently returns the empty string when err in non-nil so it would seem the error messages printed (lines 224, 249) will not properly identify the bad argument.

mansinahar: @josephveach Good catch! Actually, I was using `endpoint` there so that we get the resolved endpoint to validate in the next step (line 230). So, we can't replace it with `_` but I can hold that value in a temp variable and assign it to endpoint if err is nil.

Thanks for pointing it out!

josephveach: Yeah, somehow I didn't finish reading the two functions.  One further suggestion is to add more detail to the error messages (i.e., print both adapterName value and the offending string value) so that it's more clear exactly what failed.

mansinahar:
@dbemiller @josephveach @hhhjort, I believe I've addressed all the review comments and this is ready for a second pass. Thanks!

docs/bidders/audienceNetwork.md
```diff
@@ -4,3 +4,5 @@
 
 Audience Network will not bid on requests made from device simulators.
 When testingfor Mobile bids, you must make bid requests using a real device.
+
+**Note:** Audience Network is disabled by default. Please enable it in the app config if you wish to use it. Make sure you provide the partnerID for the auctions to run correctly.
```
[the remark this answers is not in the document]
mansinahar: Sure. Created that: https://github.com/prebid/prebid.github.io/pull/1189

PR #818, opened 2019-02-26
Bid Category Mapping

jmaynardxandr:
This PR adds initial support for mapping a bid response category into the corresponding primary ad server category id.

* Appnexus adapter maps Appnexus brand_id field to an IAB category.
* Prebid server loads IAB to primary ad server (freewheel for now) mapping data using a new CategoryFetcher interface leveraging the stored_requests support.
* Prebid server maps the IAB category to the associated ad server specific category id.
* "hb_pb_cat_dur" targeting key added to targeting values.
* Category mapping only applies when the "includebrandcategory" field exists in the request's ext.prebid.targeting object.


exchange/exchange.go
```diff
@@ -305,6 +312,125 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
...
+}
+
+func getPrimaryAdServer(adServerId int) (string, error) {
+	switch adServerId {
+	case 1:
+		return "freewheel", nil
+	case 2:
+		return "dfp", nil
+	default:
+		return "", fmt.Errorf("Primary ad server is incorrect")
```
hhhjort: Probably should change this to "Primary ad server is not recognized", or perhaps "unknown". Incorrect doesn't seem to be the right word to use.

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -16,45 +16,93 @@ import (
...
-	if err != nil {
-		return nil, err
-	}
-	storedImpData, err := collectStoredData(directory + "/stored_imps")
-	if err != nil {
-		return nil, err
-	}
-
-	return &eagerFetcher{storedReqData, storedImpData}, nil
+	storedData, err := collectStoredData(directory, FileSystem{make(map[string]FileSystem), make(map[string]json.RawMessage)}, nil)
```
hhhjort: You should make a variant of `collectStoredData()` that will unmarshal the json on startup, so we don't unmarshal on every request. It may even reduce memory requirements to store as a map rather than a raw JSON blob.

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -16,45 +16,93 @@ import (
...
+		fileName := primaryAdServer
+		if len(publisherId) != 0 {
+			fileName = primaryAdServer + "_" + publisherId
+		}
+
+		if file, ok := primaryAdServerDir.Files[fileName]; ok {
+
+			tmp := make(map[string]string)
+
+			if err := json.Unmarshal(file, &tmp); err != nil {
```
hhhjort: Since this is called once per (winning) bid, this will cause us to unmarshal the same JSON multiple times per request. See above about unmarshalling at startup and storing that result.

stored_requests/backends/file_fetcher/fetcher.go
hhhjort: Another option is to use `collectStoredData()`, but then go through the unmarshalling process on the result to get the maps to finally store in the fetcher.

stored_requests/multifetcher.go
```diff
@@ -35,6 +35,10 @@ func (mf MultiFetcher) FetchRequests(ctx context.Context, requestIDs []string, i
 	return
 }
 
+func (mf MultiFetcher) FetchCategories(primaryAdServer, publisherId, iabCategory string) (string, error) {
+	return "", nil
```
hhhjort: All the possible fetchers already have nil returns if they don't support category fetching. The `MultiFetcher` should just loop over the list of fetchers and combine them as it does in `FetchRequests()`, so that the multifetcher will work as more category fetchers are added.

stored_requests/config/config.go
```diff
@@ -34,15 +34,20 @@ import (
...
+			cfg.StoredRequests.Postgres.ConnectionInfo.Host,
+			cfg.StoredRequests.Postgres.ConnectionInfo.Port,
+			cfg.StoredRequests.Postgres.ConnectionInfo.Username)
+		db = newPostgresDB(cfg.StoredRequests.Postgres.ConnectionInfo)
+	}
+	eventProducers, ampEventProducers := newEventProducers(&cfg.StoredRequests, client, db, router)
+	cache := newCache(&cfg.StoredRequests)
+	ampCache := newCache(&cfg.StoredRequests)
+	fetcher, ampFetcher = newFetchers(&cfg.StoredRequests, client, db)
+	categoriesFetcher, _ = newFetchers(&cfg.CategoryMapping, client, db)
```
hhhjort: I am thinking we should cast this to a categories fetcher at this point, so there is no confusion that this is a category fetcher, not a standard fetcher.

Actually, to be secure, I think we should define a `globalFetcher` interface that implements both fetcher and categoryFetcher, and have that be what `newFetchers()` returns. The makes explicit that `newFetchers()` supports both, and then we immediately cast them into the fetchers we want at this point. This would also force us to make fetcher decorators like cache support need to explicitly support category fetching from the get go, but otherwise we can run into the case where caching only supports the plain fetcher, so when it is applied, the result cannot be cast to a category fetcher, and you have a crash based on a seemingly valid config not working as intended.

hhhjort: To be honest, at this point I would say we want to have `newFetchers()` to return a standard, amp, or category fetcher, and have us call it three times. But that would require refactoring a significant chunk of code, and probably changing the layout of the config file, so out of scope for this PR.

exchange/exchange.go
jmaynardxandr: Updated, also added what the bad value was when it's unrecognized to the error message.

stored_requests/multifetcher.go
jmaynardxandr: Good catch, changed this function to loop through the list of fetchers.

stored_requests/config/config.go
jmaynardxandr: That's a good call out. I like your idea of having newFetcher() that returns a single instance and we call it three times, but agreed that it will affect a lot of areas of the code so probably best done in a separate PR to avoid bloating this one.

stored_requests/backends/file_fetcher/fetcher.go
Kalypsonika: @hhhjort this code is now refactored: 
- Categories unmarshalling is now lazy and works ones per file at the first request to this file
- Unmarshalled categories are now stored in map
- Unmarshalled files are deleted from fileSystem variable after successful unmarshalling

stored_requests/backends/file_fetcher/fetcher.go
Kalypsonika: @hhhjort code refactored to unmarshall same categories file once. Please see previous comment

stored_requests/config/config.go
hhhjort: There is still the suggestion to have `newFetchers()` return an interface that supports both methods, and then immediately cast them to a normal fetcher interface or a category fetcher interface.

config/config.go
```diff
@@ -490,7 +491,10 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("datacache.filename", "")
 	v.SetDefault("datacache.cache_size", 0)
 	v.SetDefault("datacache.ttl_seconds", 0)
+	v.SetDefault("category_mapping.filesystem", true)
+	v.SetDefault("category_mapping.directorypath", "./static/category-mapping")
 	v.SetDefault("stored_requests.filesystem", false)
+	v.SetDefault("stored_requests.directorypath", "./stored_requests/data/by_id")
```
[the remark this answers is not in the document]
jmaynardxandr: These are setting defaults, so pbs config file could just set stored_requests.filesystem to true and it will default to using this path.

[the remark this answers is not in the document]
hhhjort: This means we get a default directory path set if we simply supply `stored_requests.filesystem=true` So we have a default value for the directory even if it is turned off by default. Makes turning it on potentially a little simpler, and advertises a "standard" location.

I can see arguments either way

exchange/exchange.go
```diff
@@ -305,6 +312,125 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
...
+
+	//If includebrandcategory is present in ext then CE feature is on.
+	if brandCatExt == (openrtb_ext.ExtIncludeBrandCategory{}) {
+		return res, nil //if not present continue the existing processing without CE.
+	}
+
+	//if includebrandcategory present but primaryadserver/publisher not present then error out the request right away.
+	primaryAdServer, err := getPrimaryAdServer(brandCatExt.PrimaryAdServer) //1-Freewheel 2-DFP
+	if err != nil || primaryAdServer == "" {
+		return res, err
```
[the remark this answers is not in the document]
hhhjort: As the code is today that should not be a possible state. But it probably would be good to check for `err==nil` and generate one to return for completeness. The primary point is we won't try to do the category mapping if no primary adserver exists, so we exit the request right away even if we don't properly error out.

exchange/exchange.go
```diff
@@ -306,6 +313,125 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
...
+				}
+				duration = bidderExt.CreativeInfo.Video.Duration
+
+			}
+
+			bidIabCat := bid.Cat
+			if len(bidIabCat) != 1 {
+				//TODO: add metrics
+				//on receiving bids from adapters if no unique IAB category is returned  or if no ad server category is returned discard the bid
+				bidsToRemove = append(bidsToRemove, bidInd)
```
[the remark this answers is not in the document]
hhhjort: We generally abort and return errors if we cannot unmarshal data. I don't see us hard erroring on a failed duration check outside of that. The theory being is if the data is so munged that we cannot unmarshal the JSON, it sort of calls into question all the data (or at least all the data from the bidder in question). This should be a rare occurrence and better to not bid than let faulty data through the process. If the problem is simply we cannot determine what we need to know from the supplied data, yet the data format was good, then we don't worry that all the data might be contaminated.

Basically a read error says we should abort, but successfully read data that is not fit for purpose simply disallows the object.

exchange/exchange.go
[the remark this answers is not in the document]
jmaynardxandr: Agreed I think the intention was to return an error here if no primaryAdServer, good catch.

exchange/exchange.go
```diff
@@ -305,6 +312,125 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
 	return bidResponse, err
 }
 
+func applyCategoryMapping(requestExt openrtb_ext.ExtRequest, seatBids *map[openrtb_ext.BidderName]*pbsOrtbSeatBid, categoriesFetcher stored_requests.CategoryFetcher) (map[string]string, error) {
+	res := make(map[string]string)
+
+	//If includebrandcategory is present in ext then CE feature is on.
```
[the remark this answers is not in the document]
jmaynardxandr: Yeah that is a little confusing, @Kalypsonika please update this to spell out ext.prebid.targeting.includebrandcategory and competitive exclusion. Thanks!

exchange/exchange.go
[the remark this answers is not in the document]
jmaynardxandr: True, unmarshal errors stop the process but a zero duration value is allowed through, for now. I think we'll probably be removing bids with a zero duration in a future pr though similar to what we do when we can't find a category.

exchange/exchange.go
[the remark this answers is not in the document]
Kalypsonika: Done

exchange/exchange.go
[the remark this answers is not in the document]
Kalypsonika: Done

stored_requests/config/config.go
```diff
@@ -34,15 +34,20 @@ import (
...
-func NewStoredRequests(cfg *config.StoredRequests, client *http.Client, router *httprouter.Router) (fetcher stored_requests.Fetcher, ampFetcher stored_requests.Fetcher, db *sql.DB, shutdown func()) {
-	if cfg.Postgres.ConnectionInfo.Database != "" {
-		glog.Infof("Connecting to Postgres for Stored Requests. DB=%s, host=%s, port=%d, user=%s", cfg.Postgres.ConnectionInfo.Database, cfg.Postgres.ConnectionInfo.Host, cfg.Postgres.ConnectionInfo.Port, cfg.Postgres.ConnectionInfo.Username)
-		db = newPostgresDB(cfg.Postgres.ConnectionInfo)
-	}
-	eventProducers, ampEventProducers := newEventProducers(cfg, client, db, router)
-	cache := newCache(cfg)
-	ampCache := newCache(cfg)
-	fetcher, ampFetcher = newFetchers(cfg, client, db)
+func NewStoredRequests(cfg *config.Configuration, client *http.Client, router *httprouter.Router) (fetcher stored_requests.Fetcher, ampFetcher stored_requests.Fetcher, db *sql.DB, shutdown func(), categoriesFetcher stored_requests.CategoryFetcher) {
```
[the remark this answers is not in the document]
Kalypsonika: Done

PR #821, opened 2019-02-28
renaming yieldmo bidder param from placementid to placementId

Winston-Yieldmo:
The Yieldmo js adapter uses the bidder param placementId, and the Yieldmo server to server adapter to use placementid. This branch corrects that mistake and unifies them to both use placementId.

### DIFF
--- a/adapters/yieldmo/params_test.go
+++ b/adapters/yieldmo/params_test.go
@@ -40,7 +40,7 @@ func TestInvalidParams(t *testing.T) {
 }
 
 var validParams = []string{
-	`{"placementid": "123"}`,
+	`{"placementId": "123"}`,
 }
 
 var invalidParams = []string{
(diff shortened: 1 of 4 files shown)


hhhjort:
Is anyone currently hitting your adapter using the old `placementid` name? If so, then PBS will reject all incoming bid requests that include a reference to yieldmo's `placementid` as per `bidder-params/yieldmo.json`. It would not be nice to break all of a publisher's tags to support this change.

Winston-Yieldmo:
@hhhjort Thank you for taking a look at the pull request. Nobody is currently hitting our adapter with the old field name. We noticed the issue while working to connect to our first publisher via prebid server. I understand that in the future we will need to make all changes backwards compatible so that we do not break the adapter for existing partners, but in this case we are able to clean up and remove the old field, because we don't have any existing partners.

PR #822, opened 2019-03-03
[Currency support] Adding a debug endpoint

benjaminch:
This CL introduces a new admin endpoint exposing current currency rate
converter internals such as:
- Sync source URL
- Internal rates
- Update frequency
- Last update

It will help for the future rollout of currencies support.

More details: #280

### DIFF
--- a/endpoints/currency_rates.go
+++ b/endpoints/currency_rates.go
@@ -1,0 +1,74 @@
+package endpoints
+
+import (
+	"encoding/json"
+	"net/http"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prebid/prebid-server/currencies"
+)
+
+// currencyRatesInfo holds currency rates information.
+type currencyRatesInfo struct {
+	Active           bool                           `json:"active"`
+	Source           *string                        `json:"source,omitempty"`
+	FetchingInterval *time.Duration                 `json:"fetchingIntervalNs,omitempty"`
+	LastUpdated      *time.Time                     `json:"lastUpdated,omitempty"`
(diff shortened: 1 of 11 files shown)


endpoints/currency_rates.go
```diff
@@ -0,0 +1,74 @@
...
+	"encoding/json"
+	"net/http"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prebid/prebid-server/currencies"
+)
+
+// currencyRatesInfo holds currency rates information.
+type currencyRatesInfo struct {
```
hhhjort: Took me a while to figure out why you have two info structs. Finally realized because they are private, but in different packages.

benjaminch: Yes indeed, I introduced it in order not to have the endpoint package tied to the currency package implementation :) 
It's also easier to test

endpoints/currency_rates.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if rateConverter == nil {
+		return currencyRatesInfo
+	}
+
+	currencyRatesInfo.Active = true
+
+	infos := rateConverter.GetInfo()
+	if infos == nil {
+		return currencyRatesInfo
```
josephveach: Does it make sense for Active to be true but nothing else in the struct to be defined in this case?

benjaminch:
Hey !
I think so in the sense where there is a rate converter set rather than
nil. The converter might be constant but it’s active and not the default
nil behavior

What do you think ?

Cheers !

PR #826, opened 2019-03-04
PBS-64: Fix HTML-encoding errors

guscarreon:
Using the NewEncoder function that will help correct all '&', '<', and '>' characters.

### DIFF
--- a/exchange/exchange_test.go
+++ b/exchange/exchange_test.go
@@ -58,17 +58,14 @@ func TestNewExchange(t *testing.T) {
 //From exchange_test.go func TestPanicRecoveryHighLevel(t *testing.T) {
 //but remember we don't need to create everything in therek, we'll just create the request and use makeBidResponse to test
 func TestCharacterEscape(t *testing.T) {
-	/* The objective is to get to execute e.buildBidResponse(ctx.Background(), liveA... ) (*openrtb.BidResponse, error) to do so we first
-	   create all the parameters it needs. Such as:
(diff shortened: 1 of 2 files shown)


dbemiller:
One best practice I'd recommend: whenever you fix a bug, add a test which makes sure it doesn't come back again, and label it with a well-known comment string.

For example, if you search for `// Prevents #` in the project, you'll find some of the other regression tests for bugs we've had. If everyone's consistent about it, it's incredibly helpful. Cause the next time someone refactors the code, the test will break and they'll have a Github issue they can check with all the details.

guscarreon:
David, I don't know the ```// Prevents #``` notation. In your review, if you have time to do one, could you please point out what format or comment conventions to follow on the test case function?

dbemiller:
@guscarreon like this: https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/amp_auction_test.go#L86

It's not an official notation or anything... just a convention I followed in case we wanted to find them all with `grep` or something like that.

exchange/exchange_test.go
```diff
@@ -155,7 +155,8 @@ func TestCharacterEscape(t *testing.T) {
 		t.Errorf("exchange.buildBidResponse returned %d errors", len(errList))
 	}
 
-	if bytes.Contains(bid_resp.Ext, []byte("&")) {
+	//if bytes.Contains(bid_resp.Ext, []byte("&")) && !bytes.Contains(bid_resp.Ext, []byte("u0026")) {
```
hhhjort: Good fix, not sure what the purpose of this comment is?

PR #828, opened 2019-03-06
Update mxmCherry from v10 to v12-alpha

cheung-chifung:
I updated mxmCherry's version for better OpenRTB support in the future.
Though the current version is still alpha, we need to update it sooner or later.


hhhjort:
Good call out. We probably want t wait for it to become a full release unless the new version fixes and identified prebid issue.

cheung-chifung:
I changed the update target to v11
https://github.com/prebid/prebid-server/pull/838

PR #831, opened 2019-03-07
PBS-278: metrics for recovered bidder panics

josephveach:
New meter _PanicMeter_, added to the AdapterMetrics struct, is incremented via the _RecordAdapterPanic_() interface.


endpoints/auction.go
```diff
@@ -199,56 +199,56 @@ func (a *auction) auction(w http.ResponseWriter, r *http.Request, _ httprouter.P
 				}
 			}
 			sentBids++
-			bidderRunner := recoverSafely(func(bidder *pbs.PBSBidder, blables pbsmetrics.AdapterLabels) {
+			bidderRunner := a.recoverSafely(func(bidder *pbs.PBSBidder, adapterlabels pbsmetrics.AdapterLabels) {
```
hhhjort: Should be `adapterLabels`, as the go standard is to use camelcase

josephveach: This discussion should really be in https://github.com/prebid/prebid-server/pull/832, as these changes of _blables_ and _blabels_ to _adapterlabels_  originate there.  You're right that _adapterLables_ is preferable to _adapterlabels_.  Further investigation into variable naming reveals that maybe both names are too long (see https://talks.golang.org/2014/names.slide#6 and the subsequent two slides) and that maybe _alabels_ (or _aLabels_ ?) should be used.  Do you agree?

josephveach: I've updated the variable name to _alabels_ per further discussion with @hhhjort

exchange/exchange_test.go
```diff
@@ -146,12 +146,35 @@ func newRaceCheckingRequest(t *testing.T) *openrtb.BidRequest {
...
+	defer server.Close()
+
+	cfg := &config.Configuration{
+		CacheURL: config.Cache{
+			ExpectedTimeMillis: 20,
+		},
+		Adapters: blankAdapterConfig(openrtb_ext.BidderList()),
+	}
+
+	theMetrics := pbsmetrics.NewMetrics(metrics.NewRegistry(), openrtb_ext.BidderList())
```
mansinahar: Any reason why we'd not just use the DummyMetricsEngine here for the test?

exchange/exchange_test.go
```diff
@@ -146,12 +146,35 @@ func newRaceCheckingRequest(t *testing.T) *openrtb.BidRequest {
...
+	e := NewExchange(server.Client(), nil, cfg, theMetrics, adapters.ParseBidderInfos("../static/bidder-info", openrtb_ext.BidderList()), gdpr.AlwaysAllow{}, currencies.NewRateConverterDefault()).(*exchange)
```
mansinahar: I'd just use an empty http client here like `&http.Client{}` since this test is just testing the `recoverSafely()` method and that doesn't seem to be using it. This will help reduce the first 4 lines of the test and can make it a little simpler to read but will leave it up to you.

hhhjort:
For some reason it can't figure out `auction.go`  Otherwise it looks good.

exchange/exchange_test.go
josephveach: You're right, this is a better way to write this test.  Change made.

exchange/exchange_test.go
josephveach: The other panic recovery test I changed above in endpoints/auction_test.go uses the DummyMetricsEngine.  I wanted to exercise a different RecordAdapterPanic() with this test.  The former (auction_test.go) executes the no-op RecordAdapterPanic() in pbsmetrics/config/metrics.go; the latter executes RecordAdapterPanic() in pbsmetrics/go_metrics.go.

josephveach: But, to answer your question, I don't know of a reason why the DummyMetricsEngine wouldn't work here.

mansinahar: That makes sense. Can we just leave a comment explaining that though? Because if someone looks at these tests and they see two different MetricsEngine being used, they might get confused and wonder why. In that case a clarifying comment might be helpful. What do you think?

josephveach: Looking more closely at each file, I see that exchange_test.go uses the same metrics engine in each of the four tests in which NewExchange() is called.  OTOH, in auction_test.go, either the dummy metrics engine (twice) or nil (once) is used.  I think people will more likely be comparing tests within the individual files than between these files and so I'm not convinced adding a comment for this one test referring to another test in a file in a separate directory will be particularly beneficial.   Maybe someday we should provide more explanations of the logic behind all the PBS tests, though given how tests are added in incremental fashion, doing so may open many new cans of worms.

PR #832, opened 2019-03-07
PBS-336: correct bad variable name

josephveach:
Simply a change of all the **blabel** and **blable** variable names passed as in the function argument to _recoverSafely_(), lines 202-252, to **adapterlabels**.

Since you're reviewing this, please consider the merits of the idea of moving this 50-line function call into a separate function.  I think this code is very confusing in its current form, which is that the call to _recoverSafely_() serves to simply insert some defer() code into the 50-line function passed to it, and then returns the same function.  Why not simply write a separate function with the defer() code already in place?  I can't spot a technical reason to not move the function out of calling function _auction_().  The benefits of moving it out are (1) the problem being fixed by this change would have been caught by the compiler and (2) it would be much easier for future maintenance to understand what's going on here.  If that's agreeable, we will open a separate, less urgent, ticket to address this change which also would be done for similar code in exchange/exchange.go's _getAllBids_() which calls its own _recoverySafely_().

### DIFF
--- a/endpoints/auction.go
+++ b/endpoints/auction.go
@@ -199,16 +199,16 @@ func (a *auction) auction(w http.ResponseWriter, r *http.Request, _ httprouter.P
 				}
 			}
 			sentBids++
-			bidderRunner := recoverSafely(func(bidder *pbs.PBSBidder, adapterlabels pbsmetrics.AdapterLabels) {
+			bidderRunner := recoverSafely(func(bidder *pbs.PBSBidder, alabels pbsmetrics.AdapterLabels) {
 				start := time.Now()
 				bidList, err := ex.Call(ctx, req, bidder)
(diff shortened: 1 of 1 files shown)


josephveach:
I changed the name from adapterlabels to alabels to conform with recommendations from  https://talks.golang.org/2014/names.slide#6.

mansinahar:
@josephveach LGTM except shouldn't we name this variable `aLabels` rather than `alabels` to comply with Go's use of camelcase naming?

josephveach:
I changed _alabels_ to _aLabels_.  I originally thought _alabels_ would be okay given it was replacing _blabels_ and that the first part (i.e., 'a') isn't really a word but better to follow the rules so there's more credibility when we ask others to follow the same rules.

Issue #820, opened 2019-02-28
Error with bidder aliases and GDPR scrubbing of personal information.


hhhjort:
In the file `exchange/utils.go` the function `cleanOpenRTBRequests()` we have a chunk of code to verify if the user has consented to give data to the bidder if the GDPR applies to the request. This is determined by `gDPR.PersonalInfoAllowed(ctx, bidder, consent)`.

We store user mappings only for core bidders, as any aliased bidders would have the same `buyeruid` as the core bidder. But we are not resolving the alias before checking, and most likely users have only consented to the core bidder, not one of the aliased names. We need to resolve the core bidder, using `resolveBidder(bidder, aliases)` so that `gDPR.PersonalInfoAllowed` can look up consent based on the actual bidder, not an aliased name.

PR #834, opened 2019-03-08
PBS-335: Fix GDPR info lookup to use core bidder when aliases are used

mansinahar:
This PR addresses #820

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,6 +10,14 @@
   version = "v1.3.0"
 
 [[projects]]
+  digest = "1:af6e785bedb62fc2abb81977c58a7a44e5cf9f7e41b8d3c8dd4d872edea0ce08"
+  name = "github.com/NYTimes/gziphandler"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "dd0439581c7657cb652dfe5c71d7d48baf39541d"
+  version = "v1.1.1"
+
+[[projects]]
   branch = "master"
   digest = "1:5264e1bb1289de58fd25724e060fe45524a9c2b39bad8562a6fd9d6a62d7a3b6"
   name = "github.com/asaskevich/govalidator"
@@ -467,6 +475,7 @@
   analyzer-version = 1
   input-imports = [
     "github.com/DATA-DOG/go-sqlmock",
+    "github.com/NYTimes/gziphandler",
     "github.com/asaskevich/govalidator",
     "github.com/blang/semver",
     "github.com/buger/jsonparser",
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -36,6 +36,7 @@ func cleanOpenRTBRequests(ctx context.Context, orig *openrtb.BidRequest, usersyn
 	consent := extractConsent(orig)
 	if gdpr == 1 {
 		for bidder, bidReq := range requestsByBidder {
+			// Fixes #820
 			coreBidder := resolveBidder(bidder.String(), aliases)
 			if ok, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, consent); !ok && err == nil {
 				cleanPI(bidReq, labels.RType == pbsmetrics.ReqTypeAMP)
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -36,7 +36,8 @@ func cleanOpenRTBRequests(ctx context.Context, orig *openrtb.BidRequest, usersyn
 	consent := extractConsent(orig)
 	if gdpr == 1 {
 		for bidder, bidReq := range requestsByBidder {
-			if ok, err := gDPR.PersonalInfoAllowed(ctx, bidder, consent); !ok && err == nil {
+			coreBidder := resolveBidder(bidder.String(), aliases)
+			if ok, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, consent); !ok && err == nil {
 				cleanPI(bidReq, labels.RType == pbsmetrics.ReqTypeAMP)
 			}
 		}
--- a/exchange/utils_test.go
+++ b/exchange/utils_test.go
@@ -13,7 +13,7 @@ import (
 
 // permissionsMock mocks the Permissions interface for tests
 //
-// It always disallows BrightRoll for GDPR consent
+// It only allows appnexus for GDPR consent
 type permissionsMock struct{}
 
 func (p *permissionsMock) HostCookiesAllowed(ctx context.Context, consent string) (bool, error) {
@@ -25,10 +25,10 @@ func (p *permissionsMock) BidderSyncAllowed(ctx context.Context, bidder openrtb_
 }
 
 func (p *permissionsMock) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, consent string) (bool, error) {
-	if bidder == "brightroll" {
-		return false, nil
+	if bidder == "appnexus" {
+		return true, nil
 	}
-	return true, nil
+	return false, nil
 }
 
 func assertReqWithoutAliases(t *testing.T, reqByBidders map[openrtb_ext.BidderName]*openrtb.BidRequest) {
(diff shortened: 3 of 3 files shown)


exchange/utils_test.go
```diff
@@ -1,11 +1,123 @@
...
+			assert.NotNil(t, err, "Error shouldn't be nil")
+		} else {
+			assert.Nil(t, err, "Err should be nil")
+			test.bidReqAssertions(t, reqByBidders)
+		}
+	}
+}
+
+// newAdapterAliasBidRequest builds a BidRequest with aliases
+func newAdapterAliasBidRequest(t *testing.T) *openrtb.BidRequest {
```
mansinahar: We've a couple such requests and responses - some in the test functions and some as part of helper functions like this one - that we build and use for test cases in this package and I was wondering if we should put them all together in separate files like `sample_requests_test.go` and `sample_responses_test.go` (and maybe also have a dynamic req/res builder type function) for better organization and readability. But, this might require quite a bit of refactoring and therefore would like to know your thoughts @hhhjort @dbemiller @josephveach @mkendall07 @guscarreon before diving into it.

hhhjort: The adapter test suite is probably the most feature rich and organized suite. It also benefits from all adapters being called through the same API, so you only need to run the test suite through the API.

On the exchange side it would be nice to have something similar. Unfortunately in the exchange loop, the adapter handlers will make calls out to the bidders to get bids back. To have proper exchange tests, we would need to mock out all these bidder responses, and ofcou

exchange/utils_test.go
```diff
@@ -1,11 +1,123 @@
...
+func (p *permissionsMock) HostCookiesAllowed(ctx context.Context, consent string) (bool, error) {
+	return true, nil
+}
+
+func (p *permissionsMock) BidderSyncAllowed(ctx context.Context, bidder openrtb_ext.BidderName, consent string) (bool, error) {
+	return true, nil
+}
+
+func (p *permissionsMock) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, consent string) (bool, error) {
+	if bidder == "brightroll" {
```
hhhjort: I would have it return true for "appnexus", and false otherwise. Otherwise a test that typos on "brightroll" will have strange behavior, and the real test returns false for everything but the specific bidders that were consented to.

Also help catch an issue if `corebidder` doesn't return the correct value here for some reason. As long as corebidder returns something other than "brightroll", we pass. We would rather tighten up to have corebidder return "appnexus", or else we fail.

mansinahar: PR Updated!

Gopkg.lock
```diff
@@ -467,6 +475,7 @@
   analyzer-version = 1
   input-imports = [
     "github.com/DATA-DOG/go-sqlmock",
+    "github.com/NYTimes/gziphandler",
```
mkendall07: what's this used for? I don't see it in the code diff.

dbemiller: Looks like an accident... I agree, I wouldn't have expected `Gopkg.lock` changes here.

Lockfile generally only changes if you've decided to update dependencies (which  probably belongs in its own PR) or because you're changing `Gopkg.toml`.

exchange/utils.go
```diff
@@ -36,7 +36,8 @@ func cleanOpenRTBRequests(ctx context.Context, orig *openrtb.BidRequest, usersyn
 	consent := extractConsent(orig)
 	if gdpr == 1 {
 		for bidder, bidReq := range requestsByBidder {
-			if ok, err := gDPR.PersonalInfoAllowed(ctx, bidder, consent); !ok && err == nil {
+			coreBidder := resolveBidder(bidder.String(), aliases)
```
dbemiller: Similar to this comment: https://github.com/prebid/prebid-server/pull/826#issuecomment-471576897

You may also want to put `// Fixes #` comments over fixes where the code could go "either way." Like... if it's not immediately obvious which code is correct or why.

They're not as useful as  `// Prevents #` comments... but it are occasionally nice to have.  It's basically a signal that "hey... if you're refactoring this code, look at this issue and make sure you don't re-introduce it."

exchange/utils_test.go
dbemiller: There are a bunch of valid input requests here: https://github.com/prebid/prebid-server/tree/master/endpoints/openrtb2/sample-requests/valid-whole

They currently serve two purposes:

1. There's a test in `endpoints`  which makes sure that they all pass input validation.
2. I had hoped someday to generate user-facing docs from the `exemplary` ones. Like... "here's the  prototypical  way to build a  request using feature X." The `supplementary` ones wouldn't make good user-facing docs, but they're valid nonetheless, and exist to help fill out test coverage. Never got around to making that happen, thoughh :(.

Nothing preventing you from using them in other tests too, though!

mansinahar: @dbemiller Yes, we can definitely use those json requests in this or other tests. I didn't really want to use that here because a bid request object made much more sense to me for this particular test case. 

Also, I was actually talking about the request/response objects that we define directly in our tests for some test cases. Like I am doing here with the `newAdapterAliasBidRequest` function. This test case is also leveraging another sample request defined by the `newRaceCheckingRequest` function. There are a couple other request/response objects in this package which we can potentially pull out and put together in corresponding `sample_requests_test.go` and `sample_responses_test.go` files, for example. Quite similar to how we have bundled the json request and responses together. Although, I think that will require some amount of refactoring and we might not want to do that as part of this PR.

Gopkg.lock
mansinahar: @mkendall07  @dbemiller. This is used in `server.go` and is a result of I running `dep ensure` on my machine to make sure I have all the updated dependencies. Wasn't aware of the fact that dependencies are suppose to be updated as part of a dedicated PR. I will reverse this. Thanks for pointing it out.

exchange/utils_test.go
dbemiller: yeah, for sure. I wouldn't refactor all that in this PR either.

Just for some context... those `sample-requests` got added to the project fairly late, relative to all the tests with hardcoded objects. If I were starting from scratch I think I would have started with JSON and stuck with it.

The fact is that hardly any tests actually care about all those properties. They generally just want "any valid request... but let me set these 3-4 properties so I know what I'm working with." Since the `sample-requests` payloads run through other input validation tests, they're guaranteed to be valid. If you just define an object in a test file somewhere, the tests which use it are less focused. They might fail because there was a breaking API change and the object in the test wasn't valid anymore.... or because the object always had an invalid property, but none of the other tests used it in such a way that it mattered to them yet.

Just my 2 cents... not a strong opinion, and definitely not something I'd suggest blocking this PR to do.

Gopkg.lock
dbemiller: oh man.... that's a quirk in `dep` that I never knew about o.o.

Apparently it'll just auto-fetch anything you `import`, regardless of whether or not you've put it in your `Gopkg.toml`. I was treating it like `npm`, which I'm more familiar with, and assumed you had to list the dependency explicitly to pull it down.

That's... quite dangerous, because it means the builds aren't reproducible now.  I take it back... you definitely want to get that in your lockfile as soon as possible :). It shouldn't _update_ dependencies... it'll just pin them to a known version.

Good thing [native go modules](https://github.com/golang/go/wiki/Modules) are in the process of replacing `dep`, eh?

mansinahar: Yes, `dep ensure` will usually fetch something that you've imported but not specified in the `Gopkg.toml` file. I think the thought process with `dep` here is that if your code is using it, `dep` will be like, _I see you're importing this dependency and using it. I will just install it for you. And I will abide by any rules for installing it if you've mentioned them in the `Gopkg.toml` file._ 

And for reproducible builds, I believe you'd have to commit and track `vendor` as part of the github repo too.

As for the `gziphandler`, we do have that in our `Gopkg.toml` file. I am now confused why didn't it exist in the lock file until now though 😅

dbemiller: very odd...

akh...  sounds like a sloppy code review on our part. Whichever PR updated the `Gopkg.toml` must not have also  updated the `Gopkg.lock` like it should have. Probably my bad.

> And for reproducible builds, I believe you'd have to commit and track vendor as part of the github repo too.

technically true, yeah :). Should have said "more reproducible" builds. Without the lockfile  your  build won't reproduce  if they, say, put out a new version of their library.  Or  if they don't have tagged versions yet, put out any new commits at all.

mansinahar: Yes. You definitely need the `Gopkg.lock` file. It is what defines the current state of all your imports. Just need to always be mindful that a `dep ensure` won't always lead to an exact same build 😅 

Anyway, I have kept the `Gopkg.lock` file changes in this PR.

exchange/utils_test.go
```diff
@@ -1,11 +1,137 @@
...
+	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/pbsmetrics"
+	"github.com/stretchr/testify/assert"
 )
 
+// permissionsMock mocks the Permissions interface for tests
+//
+// It only allows appnexus for GDPR consent
+type permissionsMock struct{}
```
dbemiller: A very similar class exists in other tests: https://github.com/prebid/prebid-server/blob/master/endpoints/cookie_sync_test.go#L187

This is one of those things I really don't like about Go... various language rules conspire to make it really hard to write shareable test code that's clearly _test_ code, and not meant for prod use.

The stdlib pattern is to make a new package. So for example mock stuff for `http` is stored in `http/httptest`: https://golang.org/pkg/net/http/httptest/

Might be worth doing that with this too, to avoid repeated code.

mansinahar: Ha! Thanks for pointing that out. Didn't know there existed another mock for that interface. And now that I am looking more closely, it seems there are a couple more mocks for that interface.

https://github.com/prebid/prebid-server/blob/master/endpoints/auction_test.go#L398
https://github.com/prebid/prebid-server/blob/master/endpoints/setuid_test.go#L178

I will try to combine these mocks into one. Also, I'd like to know thoughts about using [testify's mock](https://godoc.org/github.com/stretchr/testify/mock). I think it has some nice mocking features that we can leverage.

mansinahar: I tried combining all those different mocks of the `Permissions` interface in one but I think that will require quite a bit of refactoring and shouldn't be part of this PR so will leave it out of this one.

dbemiller: I had no idea that `mock` package existed... it looks as well designed as the rest of the `testify` project is.

Definitely could be useful. I can't really judge how useful without trying it in a few places.

PR #837, opened 2019-03-13
Fix `gamoshi` bidder's CI failure

cheung-chifung:
Currently, following errors occur in CI.
I fixed it with replacing it with a dummy ID.
P.S. I also add back the omitted `gziphandler` dependency

```
Unable to resolve user sync URL: https://rtb.gamoshi.io/pix/{{.supplyPartnerId}}/scm?gdpr={{.GDPR}}&consent={{.GDPRConsent}}&rurl=http%3A%2F%2Flocalhost%3A8000%2Fsetuid%3Fbidder%3Dgamoshi%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%5Bgusr%5D for adapter: gamoshi. template: userSyncTemplate:1:29: executing "userSyncTemplate" at <.supplyPartnerId>: can't evaluate field supplyPartnerId in type macros.UserSyncTemplateParams
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,6 +10,14 @@
   version = "v1.3.0"
 
 [[projects]]
+  digest = "1:af6e785bedb62fc2abb81977c58a7a44e5cf9f7e41b8d3c8dd4d872edea0ce08"
+  name = "github.com/NYTimes/gziphandler"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "dd0439581c7657cb652dfe5c71d7d48baf39541d"
+  version = "v1.1.1"
+
+[[projects]]
   branch = "master"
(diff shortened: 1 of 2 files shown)


Gopkg.lock
```diff
@@ -9,6 +9,14 @@
   revision = "d76b18b42f285b792bf985118980ce9eacea9d10"
   version = "v1.3.0"
 
+[[projects]]
```
hhhjort: Wouldn't be the proper PR for this, but should be locked down anyway, so might as well take it here.

hhhjort:
Pinging @sa1omon so that you are aware. I guess the verification code was merged at about the same time as your code, so Travis didn't catch the issue. In any case the usersync was broken in the original submission, just no one caught it.

PR #843, opened 2019-03-15
NewFetchers code refactoring:

Kalypsonika:
-Returns single general fetcher instead of two specific
-Now uses to create category mapping fetcher

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -27,15 +27,15 @@ type Configuration struct {
 	EnableGzip  bool       `mapstructure:"enable_gzip"`
 	// StatusResponse is the string which will be returned by the /status endpoint when things are OK.
 	// If empty, it will return a 204 with no content.
-	StatusResponse  string          `mapstructure:"status_response"`
-	AuctionTimeouts AuctionTimeouts `mapstructure:"auction_timeouts_ms"`
-	CacheURL        Cache           `mapstructure:"cache"`
-	RecaptchaSecret string          `mapstructure:"recaptcha_secret"`
-	HostCookie      HostCookie      `mapstructure:"host_cookie"`
-	Metrics         Metrics         `mapstructure:"metrics"`
-	DataCache       DataCache       `mapstructure:"datacache"`
-	StoredRequests  StoredRequests  `mapstructure:"stored_requests"`
-	CategoryMapping StoredRequests  `mapstructure:"category_mapping"`
+	StatusResponse  string             `mapstructure:"status_response"`
+	AuctionTimeouts AuctionTimeouts    `mapstructure:"auction_timeouts_ms"`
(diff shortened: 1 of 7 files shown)


config/stored_requests.go
```diff
@@ -37,25 +37,74 @@ type StoredRequests struct {
...
+type HTTPEventsConfigSlim struct {
 	Endpoint    string `mapstructure:"endpoint"`
 	RefreshRate int64  `mapstructure:"refresh_rate_seconds"`
 	Timeout     int    `mapstructure:"timeout_ms"`
 }
 
-func (cfg HTTPEventsConfig) TimeoutDuration() time.Duration {
+// HTTPEventsConfig configures stored_requests/events/http/http.go
+type HTTPEventsConfig struct {
+	HTTPEventsConfigSlim
```
hhhjort: I like this embedding.

config/config.go
```diff
@@ -493,8 +493,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("datacache.filename", "")
 	v.SetDefault("datacache.cache_size", 0)
 	v.SetDefault("datacache.ttl_seconds", 0)
-	v.SetDefault("category_mapping.filesystem", true)
-	v.SetDefault("category_mapping.directorypath", "./static/category-mapping")
+	v.SetDefault("category_mapping.filesystem.enabled", true)
```
hhhjort: We should probably default to false, since this should be set up individually by any PBS host that wants to implement this.

stored_requests/config/config.go
```diff
@@ -22,48 +22,152 @@ import (
...
+		shutdown1()
+		shutdown2()
+		shutdown3()
+	}
+
+	return
+}
+
+func resolvedStoredRequestsConfig(cfg *config.Configuration) (auc, amp config.StoredRequestsSlim) {
+	sr := &cfg.StoredRequests
```
hhhjort: If we only use `sr`, and not the original `cfg`, it may be best to pass just that.

rshivapurkar:
@hhhjort As discussed , let's merge this for now. We will address the 2 minor changes in a separate PR in the week of 04/01

stored_requests/config/config.go
```diff
@@ -22,48 +22,152 @@ import (
...
-		shutdown1()
-		shutdown2()
-		if db != nil {
+		if shutdown1 != nil {
+			shutdown1()
+		}
+		if dbConnection.db != nil {
+			db := dbConnection.db
+			dbConnection.db = nil
+			dbConnection.conn = ""
```
josephveach: Maybe the above two lines should execute only if/after the db.Close() below is confirmed to be successful!?

PR #844, opened 2019-03-18
Adform adapter currency support

BZSky:


adapters/adform/adform.go
```diff
@@ -586,7 +585,10 @@ func toOpenRtbBidResponse(adformBids []*adformBid, r *openrtb.BidRequest) *adapt
 		}
 
 		bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{Bid: &openRtbBid, BidType: openrtb_ext.BidTypeBanner})
+		currency = bid.Currency
```
mansinahar: It might be useful to add a check to make sure we don't get bids back for a request that are associated with different currencies. Because in the very unlikely case that that might happen, there would be no way to know about it (except maybe from a user).

adapters/adform/adform_test.go
```diff
@@ -57,7 +58,7 @@ var adformTestData aBidInfo
 // Legacy auction tests
 
 func DummyAdformServer(w http.ResponseWriter, r *http.Request) {
```
mansinahar: A minor suggestion: Since this is a helper function just for tests, it would be good to make it private so that it's not accessible outside this package.

adapters/adform/adform_test.go
```diff
@@ -374,6 +377,9 @@ func createOpenRtbRequest(testData *aBidInfo) *openrtb.BidRequest {
 	bidRequest.Regs = &regs
 	bidRequest.User.Ext = getUserExt()
 
+	bidRequest.Cur = make([]string, 1)
```
mansinahar: Just food for thought - You can also combine these lines to achieve the same: `bidRequest.Cur = []string{testData.currency}`

adapters/adform/adform_test.go
```diff
@@ -514,7 +520,14 @@ func assertAdformServerRequest(testData aBidInfo, r *http.Request) *string {
...
 
-	if ok, err := equal("CC=1&adid=6D92078A-8246-4BA4-AE5B-76104861E7DC&fd=1&gdpr=1&gdpr_consent=abc&ip=111.111.111.111&pt=gross&rp=4&stid=transaction-id&bWlkPTMyMzQ0JnJjdXI9VVNE&bWlkPTMyMzQ1JnJjdXI9VVNE&bWlkPTMyMzQ2JnJjdXI9VVNE", r.URL.RawQuery, "Query string"); !ok {
+	var midsWithCurrency = ""
+	if isOpenRtb {
+		midsWithCurrency = "bWlkPTMyMzQ0JnJjdXI9RVVS&bWlkPTMyMzQ1JnJjdXI9RVVS&bWlkPTMyMzQ2JnJjdXI9RVVS"
+	} else {
+		midsWithCurrency = "bWlkPTMyMzQ0JnJjdXI9VVNE&bWlkPTMyMzQ1JnJjdXI9VVNE&bWlkPTMyMzQ2JnJjdXI9VVNE" // no way to pass currency in legacy adapter
+	}
+
+	if ok, err := equal("CC=1&adid=6D92078A-8246-4BA4-AE5B-76104861E7DC&fd=1&gdpr=1&gdpr_consent=abc&ip=111.111.111.111&pt=gross&rp=4&stid=transaction-id&"+midsWithCurrency, r.URL.RawQuery, "Query string"); !ok {
```
mansinahar: Again, just a minor thing but check out testify's assert package: https://godoc.org/github.com/stretchr/testify/assert

Makes assertions easier in tests and also will print much better and useful error messages.

BZSky: Noted for future, thanks! No capacity sadly to introduce it this time.

adapters/adform/adform_test.go
BZSky: We'll skip it for now to not rebuild and retest for a minor change

adapters/adform/adform_test.go
BZSky: Noted for future, thanks! No capacity sadly to introduce it this time

adapters/adform/adform.go
BZSky: If this check is about validating that bids match what we add to rcur, then we have it in our ad server

PR #845, opened 2019-03-18
Eliminates some race condition issues in adkernel adapter.

hhhjort:


hhhjort:
We don't have the race condition tester running in the build scripts

mkendall07:
why is that?

adapters/adkernelAdn/adkernelAdn.go
```diff
@@ -108,12 +108,15 @@ func dispatchImpressions(imps []openrtb.Imp, impsExt []openrtb_ext.ExtImpAdkerne
...
 	imp.Ext = nil //do not forward ext to adkernel platform
 	if imp.Banner != nil {
-		return compatBanerImpression(imp.Banner)
+		return compatBanerImpression(imp)
 	}
 	return nil
 }
 
-func compatBanerImpression(banner *openrtb.Banner) error {
+func compatBanerImpression(imp *openrtb.Imp) error {
```
mansinahar: Looks like this function name has a typo. I believe it should be `compatBannerImpression` instead of `compatBanerImpression`. Can we also refactor this as part of this PR if it doesn't require a lot of effort?

PR #846, opened 2019-03-18
allow sellers to pass bidder exts in imp[...].ext.prebid.bidder

Aadeshp:


bretg:
We discussed this in the Prebid Server PMC - this interface (imp[...].ext.prebid.bidder) is not considered official -- i.e. we're not planning on changing Prebid SDK or PBJS because the transition period would be years where we'd have to pass both the original location and the new location.

For the record, imp[...].ext.prebid.bidder is a better location for the values and I wish we'd have started out there, but we want to avoid a potentially infinite transition for a feature that doesn't actually provide customer benefits.

endpoints/openrtb2/auction.go
```diff
@@ -668,11 +668,29 @@ func (deps *endpointDeps) validateImpExt(imp *openrtb.Imp, aliases map[string]st
...
 	}
+
 	var bidderExts map[string]json.RawMessage
 	if err := json.Unmarshal(imp.Ext, &bidderExts); err != nil {
 		return []error{err}
 	}
 
+	// Also accept bidder exts within imp[...].ext.prebid.bidder
+	// NOTE: This is not part of the official API, we are not expecting clients
+	// migrate from imp[...].ext.${BIDDER} to imp[...].ext.prebid.bidder.${BIDDER}
```
mkendall07: maybe just say "at this time" at the end?

Aadeshp: okie dok

exchange/utils.go
```diff
@@ -125,42 +132,86 @@ func splitImps(imps []openrtb.Imp) (map[string][]openrtb.Imp, []error) {
...
+	bidderExts map[string]json.RawMessage,
+	rawPrebidExt json.RawMessage,
+	out *map[string][]openrtb.Imp) []error {
+
+	var prebidExt map[string]json.RawMessage
+	var errs []error
+
+	// We don't want to include other demand partners' bidder params
+	// in the sanitized imp
+	if err := json.Unmarshal(rawPrebidExt, &prebidExt); err == nil {
```
hhhjort: I was going to ask if we should push this up to only do it once, but then I realized that the loop over bidders is now happening inside this function, rather than out in `cleanOpenRTBRequests()`

endpoints/openrtb2/auction.go
```diff
@@ -668,11 +668,31 @@ func (deps *endpointDeps) validateImpExt(imp *openrtb.Imp, aliases map[string]st
...
+	// https://github.com/prebid/prebid-server/pull/846#issuecomment-476352224
+	if rawPrebidExt, ok := bidderExts["prebid"]; ok {
+		var prebidExt openrtb_ext.ExtImpPrebid
+		if err := json.Unmarshal(rawPrebidExt, &prebidExt); err == nil && prebidExt.Bidder != nil {
+			for bidder, ext := range prebidExt.Bidder {
+				if ext == nil {
+					continue
+				}
+
+				bidderExts[bidder] = ext
```
mansinahar: Since with this PR, bidder extensions can possibly be in two places (`imp[...].ext.prebid.bidder` or `imp[...].ext.bidder`), are we assuming that `imp[...].ext.prebid.bidder` will always override `imp[...].ext.bidder`? I do understand that ideally this shouldn't happen but just wondering what if it does.

Aadeshp: yup exactly, but as you said this should never happen. Also, in reality only impbus is going to be using ext.prebid.bidder for the foreseeable future

exchange/utils.go
```diff
@@ -18,7 +18,14 @@ import (
 //   1. BidRequest.Imp[].Ext will only contain the "prebid" field and a "bidder" field which has the params for the intended Bidder.
 //   2. Every BidRequest.Imp[] requested Bids from the Bidder who keys it.
 //   3. BidRequest.User.BuyerUID will be set to that Bidder's ID.
-func cleanOpenRTBRequests(ctx context.Context, orig *openrtb.BidRequest, usersyncs IdFetcher, blables map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels, labels pbsmetrics.Labels, gDPR gdpr.Permissions, usersyncIfAmbiguous bool) (requestsByBidder map[openrtb_ext.BidderName]*openrtb.BidRequest, aliases map[string]string, errs []error) {
+func cleanOpenRTBRequests(ctx context.Context,
+	orig *openrtb.BidRequest,
+	usersyncs IdFetcher,
+	blables map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels,
```
mansinahar: This looks like a typo. Also, following the Go naming convention, I believe this should be `bLabels`. Sorry, I missed this earlier.

exchange/utils.go
```diff
@@ -125,42 +132,86 @@ func splitImps(imps []openrtb.Imp) (map[string][]openrtb.Imp, []error) {
...
-			if intendedBidder == "prebid" {
+		imp := imps[i]
+		impExt := impExts[i]
+
+		rawPrebidExt, ok := impExt["prebid"]
+
+		if ok {
+			var prebidExt openrtb_ext.ExtImpPrebid
+
+			if err := json.Unmarshal(rawPrebidExt, &prebidExt); err == nil && prebidExt.Bidder != nil {
```
josephveach: Should any action be taken if Unmarshal() returns an error?

exchange/utils.go
josephveach: Should any action be taken if Unmarshal() returns an error?

exchange/utils.go
```diff
@@ -125,42 +132,86 @@ func splitImps(imps []openrtb.Imp) (map[string][]openrtb.Imp, []error) {
...
+	// We don't want to include other demand partners' bidder params
+	// in the sanitized imp
+	if err := json.Unmarshal(rawPrebidExt, &prebidExt); err == nil {
+		delete(prebidExt, "bidder")
+
+		var err error
+		if rawPrebidExt, err = json.Marshal(prebidExt); err != nil {
+			errs = append(errs, err)
+		}
+	}
```
josephveach: Should we bail out here if errs is not nil?

exchange/utils.go
hhhjort: Might be a potential performance boost to refactor `ext.prebid` handling, but may be beyond the scope of this PR.

exchange/utils.go
hhhjort: prebid is optional, and errors rare, and most times prebid is used to set behavior in main PBS and not the adapters, so probably best to carry on even if it gets borked.

PR #847, opened 2019-03-18
Video auction endpoint 

Kalypsonika:
Implementation of new Video auction endpoint


endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,495 @@
...
+func createImpressionTemplate(imp openrtb.Imp, video openrtb_ext.SimplifiedVideo) openrtb.Imp {
+	imp.Video = &openrtb.Video{}
+	imp.Video.W = video.W
+	imp.Video.H = video.H
+	imp.Video.Protocols = video.Protocols
+	imp.Video.MIMEs = video.Mimes
+	return imp
+}
+
+func loadStoredImp(storedImpId string) openrtb.Imp {
```
hhhjort: You should just grab the stored imp from the same source as the openrtb code grabs the stored imp. That is what the plan is, no need for a stub.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,495 @@
...
+		for _, bid := range seatBid.Bid {
+
+			var tempRespBidExt openrtb_ext.ExtBid
+			if err := json.Unmarshal(bid.Ext, &tempRespBidExt); err != nil {
+				return nil, err
+			}
+
+			impId := bid.ImpID
+			podNum := strings.Split(impId, "_")[0]
+			podInd, _ := strconv.ParseInt(podNum, 0, 64)
```
hhhjort: why `podInd` rather than `podID`?

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,495 @@
...
+
+func loadStoredVideoRequest(storedRequestId string) ([]byte, error) {
+	jsonString := []byte(`{"accountid": "11223344", "site": {"page": "mygame.foo.com"}}`)
+	return jsonString, nil
+}
+
+func getVideoStoredRequestId(request []byte) (string, error) {
+	req := openrtb_ext.StoredRequestId{}
+
+	if err := json.Unmarshal(request, &req); err != nil {
```
hhhjort: jsonparser is probably a better fit here than parsing the entire request.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,495 @@
...
+	errL := deps.validateVideoRequest(req)
+	if len(errL) > 0 {
+		errs = append(errs, errL...)
+	}
+	return
+}
+
+func (deps *endpointDeps) validateVideoRequest(req *openrtb_ext.BidRequestVideo) []error {
+	errL := []error{}
+	if req.AccountId == "" {
```
hhhjort: I thought we dropped AccountId?

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,495 @@
...
+		err := errors.New("request missing required field: accountid")
+		errL = append(errL, err)
+	}
+	if req.StoredRequestId == "" {
```
hhhjort: We wanted a config option if this should be required or not.

endpoints/openrtb2/video_auction.go
rshivapurkar: @hhhjort don't we need it for collecting metrics. Matt mentioned that we would need to collect data per client. do you think we would use some other pre-existing id for the same?

hhhjort: I thought for analytics we were going to look up account from the stored request. (On the backend, outside of PBS we have the connection)

endpoints/openrtb2/video_auction.go
Kalypsonika: Added stored impression loading instead of using stub

endpoints/openrtb2/video_auction.go
Kalypsonika: Done

endpoints/openrtb2/video_auction.go
Kalypsonika: Done

endpoints/openrtb2/video_auction.go
Kalypsonika: Done

endpoints/openrtb2/video_auction.go
Kalypsonika: Done

exchange/exchange.go
```diff
@@ -345,57 +356,86 @@ func applyCategoryMapping(requestExt openrtb_ext.ExtRequest, seatBids *map[openr
...
+						//if mapping required but no mapping file is found then discard the bid
+						bidsToRemove = append(bidsToRemove, bidInd)
+						continue
+					}
 				}
+			}
+
+			// TODO: consider should we remove bids with zero duration here?
+
+			pb, err = GetCpmStringValue(bid.bid.Price, targData.priceGranularity)
```
josephveach: I see GetCpmStringValue never returns an error.  Maybe should replace _err_ above with an underscore since it doesn't seem necessary to check err.

josephveach: OTOH, should there be checks of _pb_ to make sure it's at least not an empty string?  Seems like an empty string value is a possibility.

rshivapurkar: @josephveach Thanks for spotting. 
We will definitely change err to underscore as in no case we would get an err. However since this function was a pre-existing function we would like to defer to @hhhjort to specify the behavior for empty pb. Imo if no price then the bid seems to be invalid. I would recommend to address this in a separate PR as this behavior should have always been present even prior  to this change.

hhhjort: I do recall some discussions on if a zero bid bid is valid or not. I don't recall what the final result was off the top of my head. The function should return a non blank `pb` as long as a valid price granularity exists. I am not sure how tight the validation on granularities is, so it might be possible.

adapters/appnexus/appnexus.go
```diff
@@ -301,21 +324,64 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters
...
+				}
+			}
+		} else {
+			errs = append(errs, err)
+			return nil, errs
+		}
 	}
 
+	imps := request.Imp
+	initialCapacity := (len(imps) + maxImpsPerReq - 1) / maxImpsPerReq
```
hhhjort: `maxImpsPerReq - 1` is so we round up instead of down?

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	requestJson, err := ioutil.ReadAll(lr)
+	if err != nil {
+		errL := []error{err}
+		handleError(labels, w, errL, ao)
+		return
+	}
+
+	resolvedRequest := requestJson
+	if deps.cfg.VideoStoredRequestRequired {
```
hhhjort: We should always try to resolve stored requests, We should error out on a failure to find a stored request only if one was required.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	//unmarshal and validate combined result
+	videoBidReq, errL := deps.parseVideoRequest(resolvedRequest)
+	if len(errL) > 0 {
+		handleError(labels, w, errL, ao)
+		return
+	}
+
+	var bidReq = &openrtb.BidRequest{}
+
```
hhhjort: If we are passing in `defReqJSON` we should copy it into `bidReq` at this stage, so we can take advantage of its default values.

adapters/appnexus/appnexus.go
Kalypsonika: @hhhjort this is initial capacity for future array of requests, small memory optimization.
Let's say there are 35 impressions and limit impressions per request equals to 10. In this case we need to create 4 requests with 10, 10, 10 and 5 impressions. 
With this formula initial capacity=(35+10-1)/10 = 4
Please let us know if you see a better way to do this.

endpoints/openrtb2/video_auction.go
Kalypsonika: Got it. Done.

endpoints/openrtb2/video_auction.go
Kalypsonika: Got it! Done.

adapters/appnexus/appnexus.go
hhhjort: It is fine for me, I made the comment mostly because it took me several minutes to figure out the purpose of adding that value in there, and thought the comment might save other reviewers a little time. And on the off chance there was something else going on that I wasn't seeing.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,524 @@
...
+		handleError(labels, w, errL, ao)
+		return
+	}
+
+	resolvedRequest := requestJson
+
+	//load additional data - stored simplified req
+	storedRequestId, err := getVideoStoredRequestId(requestJson)
+
+	if err != nil && deps.cfg.VideoStoredRequestRequired {
```
hhhjort: `deps.cfg.VideoStoredRequestRequired` requires that a valid stored video request ID was supplied, not just that they filled in the parameter with some random string. We want to use this to tie the request to a specific client. We will know the client associated with all valid request IDs. We cannot tie a client to a random unknown request ID.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,524 @@
...
+	//unmarshal and validate combined result
+	videoBidReq, errL := deps.parseVideoRequest(resolvedRequest)
+	if len(errL) > 0 {
+		handleError(labels, w, errL, ao)
+		return
+	}
+
+	var bidReq = &openrtb.BidRequest{}
+	if deps.defaultRequest {
+		hasErr, Err := getJsonSyntaxError(deps.defReqJSON)
```
hhhjort: I am not sure we need this because we are unmashalling into an empty structure. In the main code we are merging two JSON blobs and the error message was confusing as to which JSON blob was bad.

hhhjort:
Just a couple more tweaks. :)

openrtb_ext/bid_request_video.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/mxmCherry/openrtb"
+)
+
+type BidRequestVideo struct {
+	// Attribute:
+	//   accountid
+	// Type:
+	//   string; required
+	// Description:
+	//   Unique ID of the stored request
```
josephveach: Fix above line!?

mkendall07: I thought we ended saying this wasn't going to be an `accountid` but a sort of parent level config id?

openrtb_ext/bid_request_video.go
```diff
@@ -0,0 +1,268 @@
...
+	//    where 0 = live mode, 1 = test mode.
+	Test int8 `json:"test,omitempty"`
+
+	// Attribute:
+	//   test
+	// Type:
+	//   integer; default 0
+	// Description:
+	//    Indicator of test mode in which auctions are not billable,
+	//    where 0 = live mode, 1 = test mode.
```
josephveach: Fix above description!?

openrtb_ext/bid_request_video.go
rshivapurkar: Yes. We use just the storedrequestid. We need to delete this.

router/router.go
```diff
@@ -233,8 +233,14 @@ func New(cfg *config.Configuration, rateConvertor *currencies.RateConverter) (r
 		glog.Fatalf("Failed to create the amp endpoint handler. %v", err)
 	}
 
+	videoEndpoint, err := openrtb2.NewVideoEndpoint(theExchange, paramsValidator, fetcher, cfg, r.MetricsEngine, pbsAnalytics, disabledBidders, defReqJSON, bidderMap, categoriesFetcher)
+	if err != nil {
+		glog.Fatalf("Failed to create the simplified openrtb endpoint handler. %v", err)
```
josephveach: Identical error message as for an earlier call.  Please make this specific to video so they can be distinguished.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,524 @@
...
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/pbsmetrics"
+	"github.com/prebid/prebid-server/stored_requests"
+	"github.com/prebid/prebid-server/usersync"
+)
+
+func NewVideoEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte, bidderMap map[string]openrtb_ext.BidderName, categories stored_requests.CategoryFetcher) (httprouter.Handle, error) {
+
+	if ex == nil || validator == nil || requestsById == nil || cfg == nil || met == nil {
+		return nil, errors.New("NewSimplifiedEndpoint requires non-nil arguments.")
```
josephveach: "NewSimplifiedEndpoint" instead of "NewVideoEndpoint"?

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,524 @@
...
+	if err == nil {
+		storedRequest, err := loadStoredVideoRequest(storedRequestId)
+		if err != nil {
+			errL := []error{err}
+			handleError(labels, w, errL, ao)
+			return
+		}
+
+		//merge incoming req with stored video req
+		resolvedRequest, err = jsonpatch.MergePatch(storedRequest, requestJson)
```
josephveach: Test for error here?

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,523 @@
...
+		//load stored impression
+		storedImpressionId := string(pod.ConfigId)
+		storedImp, errs := deps.loadStoredImp(storedImpressionId)
+		if errs != nil {
+			return nil, errs
+		}
+
+		numImps := pod.AdPodDurationSec / minDuration
+
+		impDivNumber := numImps / len(videoDur)
```
josephveach: If number of impressions (numImps) is less than the size of videoDur, we get impDivNumber == 0.  That is a possibility, I think.  When we divide by impDivNumber on line 263, something bad probably happens.  If so, we probably need to recognize a zero value here and correct it to one.

I suppose it might not be a bad idea to also check that minDuration is non-zero since we divide by that on line 253.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,523 @@
...
+		impsArray := make([]openrtb.Imp, numImps)
+		for impInd := range impsArray {
+			newImp := createImpressionTemplate(storedImp, videoData)
+			impsArray[impInd] = newImp
+			if reqExactDur {
+				//floor := int(math.Floor(ind/impDivNumber))
+				durationIndex := impInd / impDivNumber
+				if durationIndex > len(videoDur)-1 {
+					durationIndex = len(videoDur) - 1
+				}
```
josephveach: Above comment about dividing by zero arose while trying to determine whether above three lines are ever necessary.  As is, I don't think they are, but adding the special case suggested above might change that, though I suspect it doesn't.

adapters/appnexus/appnexus.go
Kalypsonika: Comment added

endpoints/openrtb2/video_auction.go
Kalypsonika: Done

openrtb_ext/bid_request_video.go
Kalypsonika: Deleted

openrtb_ext/bid_request_video.go
Kalypsonika: Fixed

router/router.go
Kalypsonika: Fixed

endpoints/openrtb2/video_auction.go
Kalypsonika: Fixed

endpoints/openrtb2/video_auction.go
Kalypsonika: Error handler added

endpoints/openrtb2/video_auction.go
Kalypsonika: @josephveach this is a handler for special case we we have let say 7 imps and 2 durations:[15,30]. we cannot divide 7 by 2, so with this code we will have first 3 imps duration equal to 15 and other 4 imps duration equals to 30.

endpoints/openrtb2/video_auction.go
Kalypsonika: `If number of impressions (numImps) is less than the size of videoDur` - I think this is exceptional and edge case. We need to discuss this. Which ad duration to use in this case.

endpoints/openrtb2/video_auction.go
Kalypsonika: As discussed one on one with @hhhjort nothing is required to change here

endpoints/openrtb2/video_auction.go
rshivapurkar: Thanks @josephveach . Good catch. For now, let's throw an error here if such a case arises.
As an incremental approach we need to add better validation on the incoming request. Also while doing so i will sync it with what we have on the client side.

@Kalypsonika Veronika will go ahead and add the error so that we don't get into a bad state.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,524 @@
...
+	if err != nil {
+		return err
+	}
+	if len(bidExt) > 0 {
+		bidRequest.Ext = bidExt
+	}
+
+	bidRequest.Test = videoRequest.Test
+
+	bidRequest.TMax = 5000
```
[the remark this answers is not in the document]
rshivapurkar: @josephveach Let's read it from the request and if not present default to 5000 for now.

endpoints/openrtb2/video_auction.go
josephveach: My point here was that I don't think it's mathematically possible for durationIndex to ever exceed (len(videoDur)-1.  It's not a big deal.  I was just trying to eliminate an unnecessary check.

endpoints/openrtb2/video_auction.go
josephveach: I'm not sure this should necessarily be an error.  Example:  AdPodDurationSec = 60, videoDur = (60, 40, 30, 20).  In this case, minDuration will be 20 and numImps will be 3.  Since len(videoDur) is 4, impDivNumber will be zero.  Why not simply correct impDivNumber to 1 here?

endpoints/openrtb2/video_auction.go
Kalypsonika: @josephveach this is possible for example if there are 7 imps and durations array [15, 30] - size 
 2. 
7/2=3, 3 - is incorrect index, we don't have it in array of durations.

endpoints/openrtb2/video_auction.go
hhhjort: Or should we increase the number of imps to ensure that `impDivNumbere` is at least one? This will give us an imp for each duration.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,528 @@
...
+				//floor := int(math.Floor(ind/impDivNumber))
+				durationIndex := impInd / impDivNumber
+				if durationIndex > len(videoDur)-1 {
+					durationIndex = len(videoDur) - 1
+				}
+				impsArray[impInd].Video.MaxDuration = int64(videoDur[durationIndex])
+				impsArray[impInd].Video.MinDuration = int64(videoDur[durationIndex])
+				//fmt.Println(podIndex, "  ", impInd, "duration ", videoDur[durationIndex])
+			} else {
+				impsArray[impInd].Video.MaxDuration = int64(maxDuration)
```
hhhjort: Should we also set `MinDuration` here?

exchange/exchange.go
```diff
@@ -139,8 +139,8 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	conversions := e.currencyConverter.Rates()
 
 	adapterBids, adapterExtra := e.getAllBids(auctionCtx, cleanRequests, aliases, bidAdjustmentFactors, blabels, conversions)
-	auc := newAuction(adapterBids, len(bidRequest.Imp))
 	bidCategory, err := applyCategoryMapping(requestExt, &adapterBids, *categoriesFetcher, targData)
```
hhhjort: You may need `applyCategoryMapping()` to return the new `adapterBids`, as I recall it is a slice, which will do a lot with pointers.

endpoints/openrtb2/video_auction.go
Kalypsonika: As I understand no, this is a case for `requeredExactDuration==false`, `minDuration` is 0 by default

rshivapurkar: Yes. The min duration should be set to default 0. This is in sync with the product approved behavior for client side sahara.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,529 @@
...
+	glog.Errorf("/openrtb2/video Critical error: %v", errors)
+	ao.Status = http.StatusInternalServerError
+	ao.Errors = append(ao.Errors, errL...)
+}
+
+func (deps *endpointDeps) createImpressions(videoReq *openrtb_ext.BidRequestVideo) (imps []openrtb.Imp, errL []error) {
+	videoDur := videoReq.PodConfig.DurationRangeSec
+	minDuration, maxDuration := minMax(videoDur)
+	reqExactDur := videoReq.PodConfig.RequireExactDuration
+	videoData := videoReq.Video
```
josephveach: Perhaps my comment below about checking for minDuration to be non-zero wasn't well placed.  Assuming it's possible for a zero value to be in videoDur, we can get a divide by zero panic at line 252.  I suggest checking for z value here and returning an error if it is found.

endpoints/openrtb2/video_auction.go
```diff
@@ -0,0 +1,529 @@
...
+
+		numImps := pod.AdPodDurationSec / minDuration
+
+		impDivNumber := numImps / len(videoDur)
+		if impDivNumber == 0 && reqExactDur {
+			// In case of impressions number is less than durations array, we bump up impressions number up to duration array size
+			// with this handler we will have one impression per specified duration
+			numImps = len(videoDur)
+			impDivNumber = 1
+		}
```
josephveach: If it's necessary to change this file again for minDuration check for zero, this code can be made a little better performing by simply declaring impDivNumber variable earlier, then setting it only if reqExactDur is true.  This approach bypasses both an unnecessary division operation and a check for zero.

rshivapurkar: @josephveach let's address this after this release ?

endpoints/openrtb2/video_auction.go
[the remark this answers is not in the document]
Kalypsonika: Done

endpoints/openrtb2/video_auction.go
Kalypsonika: Done

endpoints/openrtb2/video_auction.go
rshivapurkar: Let's avoid the panic by making sure we do not divide by zero. However we should do a better cleanup on the incoming request for allowed duration array in the next phase.

Kalypsonika: Validation added

PR #851, opened 2019-03-21
Fix regression in exchange/utils_test.go due to updated openrtb version

mansinahar:


mansinahar:
@guscarreon, I was trying to do the same yesterday when I created the PR :D. But Go doesn't allow you to take address of constants which makes sense. See https://stackoverflow.com/a/35146856 for more on this.

PR #856, opened 2019-03-25
Enable 33across usersync

thomas-33across:
This MR enable 33across user sync with default partnerid


adapters/33across/usersync_test.go
```diff
@@ -13,7 +13,7 @@ func Test33AcrossSyncer(t *testing.T) {
 	syncInfo, err := syncer.GetUsersyncInfo("", "")
 	assert.NoError(t, err)
 	assert.Equal(t, "https://ssc-cms.33across.com/ps/?ri=123&ru=localhost%2Fsetuid%3Fbidder%3Dttx%26uid%3D33XUSERID33X", syncInfo.URL)
```
hhhjort: Do you want to update the test URL to match the new default?

thomas-33across: thank you, I had update the test case.

PR #859, opened 2019-03-27
PBS-353: caching changes for Sahara

josephveach:
Two functional changes:
- back out the doCache() change that add pb portion of custom cache key
- save only the hbCacheID portion of the custom cache key in a.vastCacheIds[]

Had to update one of the tests to allow for doCache() no longer adding pb, so updated its sister test with a matching, more meaningful variable.

### DIFF
--- a/exchange/auction.go
+++ b/exchange/auction.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	uuid "github.com/gofrs/uuid"
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


exchange/auction.go
```diff
@@ -163,7 +160,12 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
 		a.vastCacheIds = make(map[*openrtb.Bid]string, len(vastIndices))
 		for index, bid := range vastIndices {
 			if ids[index] != "" {
-				a.vastCacheIds[bid] = ids[index]
+				if competitiveExclusion && strings.HasSuffix(ids[index], hbCacheID) {
```
hhhjort: I wonder if the CPU cost of `HasSuffix()` is worth the safety of verifying that the cache key is the requested cache key above.

rshivapurkar: @hhhjort it may be. This is because if we do not validate, we may save on CPU here but the SSAI vendor would waste time fetching the bid/creative which has issues. This could result into e2e latency, which has been raised as a major concern by most clients we tried pitching this to. 

So it would be a reasonable tradeoff imo.

hhhjort: Well, if we do validate, then we put in to `hb_cache_id` the cache ID that prebid cache returned, rather than the cache ID we requested to be used. This will only help the client if they are comparing all the individual `hb_cache_id`s we return and verifying that they are all indeed identical. If they don't do this, then they waste time trying to fetch a creative that does not exist.

josephveach: The point behind the HasSuffix() check is to determine which of the bids has the custom key, not to verify the correct value of the key.  Only a subset of the bids (the winning bids) has the custom key.

hhhjort: If `competitiveExclusion` is on, then the only valid vast keys are the ones with the custom key. If we somehow have a key generated that is not the custom key, it will not be usable in the `competitiveExclusion` use case.

PR #863, opened 2019-03-29
Fixes bugs introduced with the stored requests refactor

hhhjort:


stored_requests/config/config.go
```diff
@@ -59,6 +59,8 @@ func CreateStoredRequests(cfg *config.StoredRequestsSlim, client *http.Client, r
 		// Error out if config is trying to use multiple database connections for different stored requests (not supported yet)
 		if conn != dbConnection.conn {
 			glog.Fatal("Multiple database connection settings found in Stored Requests config, only a single database connection is currently supported.")
+		} else {
+			db = dbConnection.db
```
josephveach: Isn't it more logical and very slightly more efficient (db gets assigned only once) if this else statement goes with the previous if-statement?

hhhjort: No, they are separate conditions/code paths. The first if is for the initial creation of a database connection. The second if is if a previous invocation of the function set up the database connection, and now we want to reuse it.

hhhjort: Sorry, misunderstood the recommendation.

PR #865, opened 2019-04-02
Adds a Stored Video Request Fetcher

hhhjort:
Also a minor refactor.


config/config.go
```diff
@@ -524,6 +525,30 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("stored_requests.http_events.amp_endpoint", "")
 	v.SetDefault("stored_requests.http_events.refresh_rate_seconds", 0)
 	v.SetDefault("stored_requests.http_events.timeout_ms", 0)
+	v.SetDefault("stored_video.filesystem.enabled", false)
+	v.SetDefault("stored_video.filesystem.path", "")
+	v.SetDefault("stored_video.directorypath", "./stored_requests/data/by_id")
```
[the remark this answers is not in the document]
hhhjort: Good catch, thanks.

The query would be the same form as the normal fetcher, SELECT id, data, datatype FROM .... where the details would depend on the implementation of the data store. `datatype` would always be `request` unless we want to refactor the fetcher setup so we have `imp`s being pulled from the same query as video requests. I would be against that, as then the stored `imp`s could not also be used on openrtb side. Well, it would probably be possible with a fancy enough query, but the current architecture enforces that the same datastore provides video imps and openrtb imps.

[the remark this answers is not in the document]
hhhjort: The other way to go is to have the main fetcher return three types, "request", "imp", and "video" and do away with a separate video fetcher. But that path leads to havign a fetcher that returns "request", "imp", "video", "amp", "category". I think it's probably better to have smaller targeted, semi-redundant configs than a huge all-in-one fetcher that demands support for use cases not all hosts will have.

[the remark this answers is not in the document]
Kalypsonika: I was able to make it work with this query: `SELECT aa.uuid, aa.config as data, 'request' as dataType FROM s2sconfig_config aa WHERE aa.uuid IN %REQUEST_ID_LIST%;` (should  it be accounts_account table?).
I like the idea of `FetchRequests()` refactoring, because it becomes more complicated and tangled.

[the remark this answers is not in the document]
hhhjort: Do recall if we make the postgres fetcher one big multipurpose fetcher, we must make all fetchers into big multipurpose fetchers. Otherwise the config will become much much more complicated and tangled.

Also are you saying a big multipurpose fetcher is more complicated and tangled, or simple single purpose fetchers are more complicated and tangled?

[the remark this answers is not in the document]
Kalypsonika: I'm ok with current implementation. 
It's just becomes more complicated to figure out how does it work, what parameter to pass in case we need stored requests, stored imps, video stored requests and categories.  
For the DB fetcher I would like to see simple single select query rather than complicated multi select and union query and then separate parser for data type ("request" or "imp"). 
I understand it's now a trade of between complicated config and complicated fetcher. 
I think if we will have more data to fetch in future we will need to refactor big fetcher to simple single purpose fetchers.

endpoints/openrtb2/video_auction.go
```diff
@@ -29,14 +30,14 @@ import (
 
 var defaultRequestTimeout int64 = 5000
 
-func NewVideoEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte, bidderMap map[string]openrtb_ext.BidderName, categories stored_requests.CategoryFetcher) (httprouter.Handle, error) {
+func NewVideoEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte, bidderMap map[string]openrtb_ext.BidderName, categories stored_requests.CategoryFetcher, videoFetcher stored_requests.Fetcher) (httprouter.Handle, error) {
```
mansinahar: One minor suggestion: Can we maybe consider putting all the fetchers next to each other in the method parameter list for this and the other auction and amp endpoint methods? Not that it affects the execution of the code but I think it will really improve readability.

Probably the same also in the endpointDeps struct

endpoints/openrtb2/video_auction_test.go
```diff
@@ -194,7 +196,9 @@ func mockDeps(t *testing.T, ex *mockExchangeVideo) *endpointDeps {
 		false,
 		[]byte{},
 		openrtb_ext.BidderMap,
-		nil}
+		empty_fetcher.EmptyFetcher{},
```
mansinahar: Should we also be using a mock for category fetcher in these tests to make sure the categories are fetched and returned in the auction response as expected?

config/config.go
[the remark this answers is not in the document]
Kalypsonika: Oh I mean separate fetchers by purpose, like: stored request fetcher, stored impressions fetcher and video stored request fetcher. So DB fetcher for each of those will have a simple sql query and there will no mess with which parameter to pass. 
```FetchRequests(ctx context.Context, requestIDs []string)```
and 
```FetchImpressions(ctx context.Context, requestIDs []string, impIDs []string))```
and
```FetchVideoRequests(ctx context.Context, videoRequestIDs []string)``` (if we need to store in separate table )
instead of
```FetchRequests(ctx context.Context, requestIDs []string, impIDs []string))```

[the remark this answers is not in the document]
Comment (Kalypsonika) [reply config/config.go]: Sorry, I expressed myself unclear in previous comment. I meant refactor fetchers functions, not fetchers itself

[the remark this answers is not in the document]
Comment (Kalypsonika) [reply config/config.go]: `Perhaps if the list of fetchers continues to expand it will be worth simplifying the fetcher in this manner.` - this is exactly what I meant!

File: endpoints/openrtb2/video_auction.go (hunk 2 above)
Comment (hhhjort) [reply endpoints/openrtb2/video_auction.go]: Makes sense

File: stored_requests/config/config.go [hunk 4]
```diff
@@ -26,7 +26,7 @@ import (
 
 // This gets set to the connection string used when a database connection is made. We only support a single
 // database currently, so all fetchers need to share the same db connection for now.
-var dbConnection struct {
+type dbConnection struct {
```
josephveach: Two thoughts:
(1) Comment above the type declaration seems a little inconsistent, though I'm not sure what to suggest.
(2) Why eliminate the global variable?  The new approach seems to allow for future possible programming errors if NewStoredRequests() should somehow be called twice.  In that case, I believe the first call to CreateStoredRequests() in the second call to NewStoredRequests() would pass an uninitialized dbc value leading to a second DB connection attempt even though one already exists.  That danger doesn't seem to exist with the current global variable in place.  As this code is written, I don't think two calls to NewStoredRequests() can occur, though, so this is more a precautionary suggestion.

config/config.go
```diff
@@ -524,6 +525,29 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("stored_requests.http_events.amp_endpoint", "")
 	v.SetDefault("stored_requests.http_events.refresh_rate_seconds", 0)
 	v.SetDefault("stored_requests.http_events.timeout_ms", 0)
+	v.SetDefault("stored_video.filesystem.enabled", false)
```
josephveach: The use of the names StoredVideo and stored_video in this file and similarly named entries in other files below (e.g., videoFetcher) seem potentially confusing as they may be perceived as being related to stored videos rather than stored video requests (if not now, maybe someday in the future).  Perhaps it might be better to choose names that clearly indicate these all refer to stored video requests!?

stored_requests/config/config.go
hhhjort: Global variables are bad. :) The way it was, it was mostly fine, but could potentially cause issues for testing if multiple test threads where running where the tests all assumed they were the only ones messing with `dbConnection` things.

With this change, we are 90% of the way to supporting multiple databases. Yes, a second call to `NewStoredRequests()` would create a second valid database connect, a second set of fetchers, and a second shutdown function. This should all be fine if you treat them as a second set of these things that need to be managed, but of course bad if you then conflate the members of the two sets.

I feel that getting rid of the mutable global state is worth the risk of accidentally calling `NewStoredRequests()` twice and not handling the results properly.

config/config.go
hhhjort: I can't imagine a scenario where we are storing actual videos on PBS. The closet we come is caching VAST XML on Prebid Cache. But if you can think of a name that is clearer and not an overlong monstrosity I am happy to update the config name.

josephveach: Maybe StoredVideoReq, stored_video_req, and videoReqFetcher?  An alternative would be comments in the code where each item is introduced.  If I'm the only one who cares, it's no bid deal if you don't change anything.

hhhjort: I think I will go the comment route unless someone else comments thinking that this might be necessary.

stored_requests/config/config.go
```diff
@@ -42,32 +42,30 @@ var dbConnection struct {
 //
 // As a side-effect, it will add some endpoints to the router if the config calls for it.
 // In the future we should look for ways to simplify this so that it's not doing two things.
-func CreateStoredRequests(cfg *config.StoredRequestsSlim, metricsEngine pbsmetrics.MetricsEngine, client *http.Client, router *httprouter.Router) (fetcher stored_requests.AllFetcher, db *sql.DB, shutdown func()) {
+func CreateStoredRequests(cfg *config.StoredRequestsSlim, metricsEngine pbsmetrics.MetricsEngine, client *http.Client, router *httprouter.Router, dbc *dbConnection) (fetcher stored_requests.AllFetcher, shutdown func()) {
```
josephveach: The change of no longer returning the DB connection makes the comment above the function incorrect (line 37).  This gives me the opportunity to propose that maybe the function should retain the old behavior of returning the DB connection rather than modifying the structure pointed to in the new argument (which would then not need to be a pointer).  I find it a little confusing that three things are set or changed by this function but only two of the changes are available via return value and the other is in one of the function arguments.  Either way a change is need to update either the function description or the function itself.

stored_requests/config/config.go
```diff
@@ -107,7 +105,7 @@ func CreateStoredRequests(cfg *config.StoredRequestsSlim, metricsEngine pbsmetri
 //
 // As a side-effect, it will add some endpoints to the router if the config calls for it.
 // In the future we should look for ways to simplify this so that it's not doing two things.
-func NewStoredRequests(cfg *config.Configuration, metricsEngine pbsmetrics.MetricsEngine, client *http.Client, router *httprouter.Router) (db *sql.DB, shutdown func(), fetcher stored_requests.Fetcher, ampFetcher stored_requests.Fetcher, categoriesFetcher stored_requests.CategoryFetcher) {
+func NewStoredRequests(cfg *config.Configuration, metricsEngine pbsmetrics.MetricsEngine, client *http.Client, router *httprouter.Router) (db *sql.DB, shutdown func(), fetcher stored_requests.Fetcher, ampFetcher stored_requests.Fetcher, categoriesFetcher stored_requests.CategoryFetcher, videoFetcher stored_requests.Fetcher) {
```
josephveach: Description of this function needs to be updated to reflect additional return value.

stored_requests/config/config.go
hhhjort: I will have to fix the comment. We need the `dbConnection` object so we can confirm that we can reuse the same db object (verify the connection string). It is also used in the shutdown functions in order to avoid trying to shut down the same database multiple times.

Basically `dbConnection` was a global mutable variable, which could have caused issues in testing. It is only used by `CreateStoredRequests()` and `NewStoredRequests()`. So I made it local to `NewStoredRequests()` and passed it on to `CreateStoredRequests()` to avoid issues with mutable global variables.

mansinahar: My two cents on this:

@hhhjort @josephveach What do you think about breaking this function up into two functions: `NewDBConn` and `CreateFetcher` or something along those lines? I think the confusion is arising from the fact that this function is doing both of those things and IMHO, the function name `CreateStoredRequests` doesn't convey either of those. I am not sure if this would be as simple as splitting the method in to two and then in the `NewStoredRequests` function calling the `NewDBConn` function followed by `CreateFetchers` for various fetchers or a little more complicated than that based on where all is this function used. But nevertheless, if it's the former than it might be worth doing.

Thinking of the `dbConn` as a singleton, I think that approach would make more sense and then it becomes clearer to pass the `dbConn` object to `CreateFetcher` and why it'd need it. Also, at that point `CreateFetcher` won't modify the `dbConn` object at all. It will just use it. 

And lastly, of course if it is too much effort to do in this PR then we can always do it as a separate one. But, would love to know what do you guys think about this?

config/config.go
mansinahar: @hhhjort Not suggesting that this change is necessary in this PR but I think I agree with @josephveach and even though we don't see the potential of storing videos in future, it'd still improve readability if we rename to something more on the lines of `stored_video_req`. But as I said, doesn't have to be in this PR. Maybe we can leave this for another PR. Refactoring and adding features in the same PR can be a little too much most of the times.

josephveach: Is it really "refactoring" given the code in question is being added here for the first time?  :-)

stored_requests/config/config.go
josephveach: I agree that this code has been increasing in complexity and could be changed to make it more readable and intuitive.  It may be better to wait for it to reach equilibrium rather than pouring a lot of effort into it now and risking needing to redesign it again for the next update.

config/config.go
hhhjort: I will leave potentially renaming the struct fields for another PR. But the fields in the config file would be an API change, so should get them right on launch. So the above line will be modified along with its friends.

stored_requests/config/config.go
hhhjort: I would say refactoring it out into two functions might be best for a new PR. DOn't want the scope of this to creep too far, especially after having it sit pending for so long.

stored_requests/config/config.go
hhhjort: fixed

stored_requests/config/config.go
mansinahar: Sounds reasonable!

PR #866, opened 2019-04-03
Optimize cache requests

guscarreon:
Please take a look at doCache, test case needs fpr,ating and an assert call. Morgan laws could have been used in doCache to avoid nesting the entire function in that gigantic true branch of the first if statement but I believe this is more readable. Please let me know your thoughts


exchange/auction.go
```diff
@@ -60,111 +60,118 @@ func (a *auction) setRoundedPrices(priceGranularity openrtb_ext.PriceGranularity
...
+		}
+		for ImpName, topBidsPerImp := range a.winningBidsByBidder {
+			//	winningBidsByBidder map[ImpName]map[openrtb_ext.BidderName]*pbsOrtbBid
+			for bidderName, topBidPerBidder := range topBidsPerImp {
+				impID := topBidPerBidder.bid.ImpID
+				isOverallWinner := a.winningBids[impID] == topBidPerBidder
+				//  PBS-330 add a condition so that we don't cache what is not an overallWinner
+				//if targData.includeBidderKeys
+				//if targData.includeWinners && overallWinner
+				if includeBidderKeys || (includeWinners && isOverallWinner) { //overallWinner := a.winningBids[impId]
```
josephveach: Above line can simply be:
if includeBidderKeys || isOverallWinner {

exchange/auction.go
```diff
@@ -60,111 +60,118 @@ func (a *auction) setRoundedPrices(priceGranularity openrtb_ext.PriceGranularity
...
+					var customCacheKey string
+					var catDur string
+					useCustomCacheKey := false
+					if competitiveExclusion && topBidPerBidder == a.winningBids[impID] {
```
josephveach: use **isOverallWinner** here instead of **topBidPerBidder == a.winningBids[impID]**

exchange/auction.go
```diff
@@ -60,111 +60,118 @@ func (a *auction) setRoundedPrices(priceGranularity openrtb_ext.PriceGranularity
...
-			var catDur string
-			useCustomCacheKey := false
-			if competitiveExclusion && topBidPerBidder == a.winningBids[impID] {
-				// set custom cache key for winning bid when competitive exclusion applies
-				catDur = bidCategory[topBidPerBidder.bid.ID]
-				if len(catDur) > 0 {
-					customCacheKey = fmt.Sprintf("%s_%s", catDur, hbCacheID)
-					useCustomCacheKey = true
-				}
+	if (bids || vast) && (includeBidderKeys || includeWinners) {
```
josephveach: I don't understand the use of includeBidderKeys and includeWinners in this function:
- purpose of this change is to cache only winners but we will cache non-winners if includeBidderKeys is set
- if neither of these is set, we won't cache winners; this is a change from previous behavior, isn't it?

exchange/auction.go
```diff
@@ -60,111 +60,118 @@ func (a *auction) setRoundedPrices(priceGranularity openrtb_ext.PriceGranularity
...
+		}
+
+		// Grab the imp TTLs
+		for _, imp := range bidRequest.Imp {
+			expByImp[imp.ID] = imp.Exp
+		}
+		for ImpName, topBidsPerImp := range a.winningBidsByBidder {
+			//	winningBidsByBidder map[ImpName]map[openrtb_ext.BidderName]*pbsOrtbBid
+			for bidderName, topBidPerBidder := range topBidsPerImp {
+				impID := topBidPerBidder.bid.ImpID
```
josephveach: Above line can be eliminated if you replace **ImpName** on line 90 with **impID**.  Doing so undoes the change I made recently for https://jira.corp.appnexus.com/browse/PBS-348 based on the misperception that winningBidsByBidder is not indexed by impression ID.  See that ticket if you want further details.
If you don't remove this line as described, ImpName should be replaced by an underscore both because it isn't used otherwise and also because ImpName and the associated comment are a little misleading.

exchange/auction.go
guscarreon: current version of the `addKeys` function found in `exchange/utils.go` also uses this logic to create targets. I don't know if this logic that is currently in place in `addKeys` is correct or makes perfect sense, I just replicated it.

hhhjort: If we are setting keys for all bidders, then we need to cache all bidders. But it may be best to preserve the initial logic flow and bail out of the function if we have nothing to do rather than putting everything inside an if statement.

exchange/auction.go
hhhjort: Yes, we know from the logic above `(includeBidderKeys || includeWinners)` must be true.

exchange/auction.go
guscarreon: Joe. It is probably best if I remove ImpName for the moment, don't you think? Maybe it'd be better to treat that as a separate issue.

josephveach: Argument in favor of changing ImpName to impID is that it is slightly more efficient:  impID is set only once instead of each time in each loop.  If you have any doubts about correctness, see the other references to winningBidsByBidder in auction.go and targeting.go, most notably setTargeting().

exchange/auction.go
guscarreon: But are we sure ```includeBidderKeys || (includeWinners && isOverallWinner)``` is true? I'm a bit confused.

hhhjort: From above, `(includeBidderKeys || includeWinners)` must be true. if we also know that `(includeBidderKeys || isOverallWinner)` is true, then `(includeBidderKeys || (includeWinners && isOverallWinner))` must be true.

exchange/auction.go
```diff
@@ -91,53 +91,56 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
-					useCustomCacheKey = true
-				}
-			}
-			if bids {
-				if jsonBytes, err := json.Marshal(topBidPerBidder.bid); err == nil {
-					if useCustomCacheKey {
-						// not allowed if bids is true; log error and cache normally
-						errs = append(errs, errors.New("cannot use custom cache key for non-vast bids"))
+			isOverallWinner := a.winningBids[impID] == topBidPerBidder
+			if includeBidderKeys || (includeWinners && isOverallWinner) {
```
hhhjort: Again this can be shortened to `includeBidderKeys || isOverallWinner`. If `includeBidderKeys` is false, then we know `includeWinners` is true.

exchange/auction.go
josephveach: In other words, we're in this part of the function only because we know at least one of includeBidderKeys and includeWinners is true (line 64).  We evaluate the second part of the OR only if the first part is false (includeBidderKeys).  Given includeBidderKeys must be false to get here, we can conclude that includeWinners is true, so we now only need to know whether isOverallWinner is also true.

exchange/auction.go
josephveach: You could simply test for the opposite in the above line and, if true, do a continue.  This would reduce the need for indenting and might make comparing the new and old code quite a bit easier since no additional indentation would be needed.  In other words, replace above line with:
if ! includeBidderKeys && ! isOverallWinner {
    continue
}

guscarreon: You are right Hans, let me change this.

guscarreon: That's probably more elegant. Let me change it

guscarreon: changed. The new version of the code implements your suggestion Joe

guscarreon: changed

exchange/auction.go
guscarreon: done

exchange/auction.go
guscarreon: done

exchange/auction.go
guscarreon: done

exchange/auction.go
```diff
@@ -60,11 +60,11 @@ func (a *auction) setRoundedPrices(priceGranularity openrtb_ext.PriceGranularity
...
 }
 
-func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client, bids bool, vast bool, bidRequest *openrtb.BidRequest, ttlBuffer int64, defaultTTLs *config.DefaultTTLs, bidCategory map[string]string) []error {
-	if !bids && !vast {
+func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client, targData *targetData, bidRequest *openrtb.BidRequest, ttlBuffer int64, defaultTTLs *config.DefaultTTLs, bidCategory map[string]string) []error {
+	if !((bids || vast) && (includeBidderKeys || includeWinners)) {
 		return nil
 	}
-
+	var bids, vast, includeBidderKeys, includeWinners bool = targData.includeCacheBids, targData.includeCacheVast, targData.includeBidderKeys, targData.includeWinners
```
josephveach: Does this even compile?  Variables are being tested before they're declared.

exchange/auction_test.go
```diff
@@ -124,62 +142,80 @@ func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec, bid
...
-				if strings.HasPrefix(cFound.Key, cExpected.Key) {
-					keymatch = true
-				}
-			} else {
-				if len(cFound.Key) == 0 {
-					// Key is expected to be empty
-					keymatch = true
-				}
+			for _, cExpected := range specData.ExpectedCacheables {
+				compareString = string(cExpected.Data) + string(cExpected.TTLSeconds)
```
hhhjort: I see you are just checking `Data`  and `TTLSeconds`. `Type` is not likely to be an issue. This does miss out on checking that custom keys are being set. You were not able to get the mock cache client to unrandomize the uuid in the custom key?

guscarreon: I did not Hans. I didn't get the mock cache client to unrandomize the uuid in the custom key. Is it possible for the `gofrs/uuid` library to not deliver randomized UUID codes? Is this what your mean?

hhhjort: No, I was thinking it might be possible for the mock client to do some pattern matching to recognize the uuid part and replace it. But I had not looked into it too closely.

apr2gus.txt
```diff
@@ -0,0 +1,2 @@
+PASS
+ok  	github.com/prebid/prebid-server	0.020s
```
mansinahar: Can you remove this and the .swp file, please?

exchange/auction.go
```diff
@@ -60,11 +60,11 @@ func (a *auction) setRoundedPrices(priceGranularity openrtb_ext.PriceGranularity
 	a.roundedPrices = roundedPrices
 }
 
-func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client, bids bool, vast bool, bidRequest *openrtb.BidRequest, ttlBuffer int64, defaultTTLs *config.DefaultTTLs, bidCategory map[string]string) []error {
-	if !bids && !vast {
+func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client, targData *targetData, bidRequest *openrtb.BidRequest, ttlBuffer int64, defaultTTLs *config.DefaultTTLs, bidCategory map[string]string) []error {
+	var bids, vast, includeBidderKeys, includeWinners bool = targData.includeCacheBids, targData.includeCacheVast, targData.includeBidderKeys, targData.includeWinners
```
mansinahar: This can simply be written as `bids, vast, includeBidderKeys, includeWinners := targData.includeCacheBids, targData.includeCacheVast, targData.includeBidderKeys, targData.includeWinners`. In Go, inside functions you can use the `:=` short assignment statement that will automatically infer the type of the values on the right and assign them to the variables on the left.

exchange/auction_test.go
```diff
@@ -124,62 +142,82 @@ func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec, bid
...
-	if found != len(specData.ExpectedCacheables) {
-		fmt.Printf("Expected:\n%v\n\n", specData.ExpectedCacheables)
-		fmt.Printf("Found:\n%v\n\n", cache.items)
-		t.Errorf("%s:  All expected cacheables not found. Expected %d, found %d.", fileDisplayName, len(specData.ExpectedCacheables), found)
+	for _, cExpected := range specData.ExpectedCacheables {
+		for _, cFound := range cache.items {
+			// make sure Key value is as expected
+			_, err := regexp.MatchString(cExpected.Key+`[a-zA-Z0-9_-]*`, cFound.Key)
+			if err != nil {
+				continue
```
mansinahar: I am not sure what are we trying to do here. If for an expectedCacheable, you don't find in the actual cache items a cache key with the regexp cExpected.Key+`[a-zA-Z0-9_-]*`, you just continue and move on to the next expectedCacheable item.

If we want to assert that for every expectedCacheable, there exists a an actual cache item with the cache key that matches the cExpected.Key+`[a-zA-Z0-9_-]*` regexp then we should do something like:

```
for _, cExpected := range specData.ExpectedCacheables {
	cacheKeyFound := false
	for _, cFound := range cache.items {
		// make sure Key value is as expected
		_, err := regexp.MatchString(cExpected.Key+`[a-zA-Z0-9_-]*`, cFound.Key)
		if err != nil {
			continue
		}
		cacheKeyFound = true
		break
	}
	if !cacheKeyFound {
		t.Errorf("%s cache key either not found in the cache result or doesn't match the expected key format.", cExpected.Key)
	}
}
```

mansinahar:
@guscarreon can you also please change the PR title to be a little more meaningful as that is what will become the commit message once we merge this? Thanks!

exchange/auction.go
guscarreon: It does compile now

apr2gus.txt
guscarreon: removed

exchange/auction.go
guscarreon: I know, I just thought it's more readable this way. What do you think?

hhhjort: I suppose Gus' way makes it explicit that they are booleans, but that should be easy to infer from the context below.

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
```diff
@@ -0,0 +1,95 @@
...
+		}
+	],
+	"defaultTTLs": {
+		"banner": 300,
+		"video": 3600,
+		"audio": 1800,
+		"native": 300
+	},
+	"targetDataIncludeWinners": true,
+	"targetDataIncludeBidderKeys": true,
```
josephveach: I have several questions about this test:

(1)  I don't understand how any of the bids get cached.  If I understand correctly, doCache() will be called with "bids" set to false and "vast" set to true (ultimately coming from line 93 in auction_test.go).  Line 125 in doCache() in auction.go has this test:
			if vast && topBidPerBidder.bidType == openrtb_ext.BidTypeVideo {
so we shouldn't enter the vast caching logic since all these bids have bidType of banner.  Why / how do these bids get cached?

(2) The expected cache keys are either 11_sports_22_abcd-efgh-ijkl-mnop or 33_news_44abcd-efgh-ijkl-mnop.  Given the categories are 11_sports_22 and 33_news_44, why is there no underscore between 44 and abcd in the latter while there is an underscore between 22 and abcd in the former?

(3) Shouldn't there also be a test in this directory with targetDataIncludeBidderKeys set to false to confirm that only the winning bids get cached with multiple imps?  Basically, it would be the same test as this (if this works) but with targetDataIncludeBidderKeys set to false and only two expected cached values.  I think reducing the number of cached values is the primary change for this PR that we want to confirm, though this case with targetDataIncludeBidderKeys is also necessary.  

Also, please update the PR description as already requested by Mansi.

exchange/auction.go
guscarreon: Sure Joe. It'll get corrected in PBS-354

exchange/auction_test.go
guscarreon: Mansi this function has been modified. Thanks for pointing out this flaw. Indeed the for loop was not exactly asserting anything and after careful thought, it turns out it was not even needed. Please let me know if you agree with the new approach.

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
hhhjort: 1) It is true that it does not make sense to VAST cache banner ads. However a request could be multiformat that accepts banner or video. In this case, it makes sense to have vast caching turned on for the video support. The down stream client will just have to be smart enough to figure out if the ad is banner or video, and display the content appropriately.

We could ad code to vast cache only if the ad is video if we have that bit of info available. But that is a separate optimization, so I am happy to let that be its own PR.

exchange/auction.go
mansinahar: @guscarreon If you think it is more readable this way then sure. I don't mind it 👍

exchange/auction_test.go
mansinahar: @guscarreon I see that the cache key format assertion has now been removed and as per our discussion offline, I don't think I am the best person here to judge if those assertions are needed or not. I'd guess they are but I think @hhhjort is a better person to answer that.

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
-					continue
-				}
-			}
+	if len(specData.ExpectedCacheables) > len(cache.items) {
+		//t.Errorf("%s:  Less elements were cached than expected \n", fileDisplayName)
+		t.Errorf("%s:  [CACHE_ERROR] Less elements were cached than expected \n", fileDisplayName)
+	} else if len(specData.ExpectedCacheables) < len(cache.items) {
+		t.Errorf("%s:  [CACHE_ERROR] More elements were cached than expected \n", fileDisplayName)
+	} else { // len(specData.ExpectedCacheables) == len(cache.items)
+		var ht map[string]*cacheComparator = make(map[string]*cacheComparator)
```
hhhjort: Is there a reason this is a map of pointers to `cacheComparator`s rather than simply a map of `cacheComparator`s?

hhhjort: As Gus found, it seems it actually needs to be a pointer for the . assignments to work properly.

exchange/auction_test.go
guscarreon: Mansi your observations here were very important to correct all test cases in this document

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
```diff
@@ -0,0 +1,100 @@
...
+	"targetDataIncludeCacheBids": true,
+	"targetDataIncludeCacheVast": false
```
josephveach: As above for customcachkeytest files, these two values are wrong.

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
+				ht[compareString].actualKeys = append(ht[compareString].actualKeys, cFound.Key)
+			}
+		}
+		for k, cachedElements := range ht {
+			if cachedElements.freq > 0 {
+				t.Errorf("%s:  [CACHE_ERROR] Cache inconsistensy. Element %s was not expected to get cached\n", fileDisplayName, k)
+			} else if cachedElements.freq < 0 {
+				t.Errorf("%s:  [CACHE_ERROR] Cache inconsistensy. We cached some more elements (%s) than expected.\n", fileDisplayName, k)
+			} else if targData.includeCacheVast {
+				for i := 0; i < cachedElements.freq; i++ {
```
josephveach: My comment from the previous review disappeared!  It stated that this code is never reached because cachedElements.freq is necessarily 0 if we're in this else statement.  Probably the length of expected keys is the right counter to loop on.

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
-			winningBids[pbsBid.Bid.ImpID] = bid
+		cpm := bid.bid.Price
+		wbid, ok := winningBids[bid.bid.ImpID]
+		if !ok || cpm > wbid.bid.Price {
+			winningBids[bid.bid.ImpID] = bid
+		}
+		if bidMap, ok := winningBidsByImp[bid.bid.ImpID]; ok {
+			bestSoFar, ok := bidMap[pbsBid.Bidder]
+			if !ok || cpm > bestSoFar.bid.Price {
+				bidMap[pbsBid.Bidder] = bid
```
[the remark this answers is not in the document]
guscarreon: Correct. I changed it to that. It is way more readable.

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
 			bidType: pbsBid.BidType,
 		}
-		if _, ok := winningBids[pbsBid.Bid.ImpID]; !ok {
-			winningBids[pbsBid.Bid.ImpID] = bid
+		cpm := bid.bid.Price
+		wbid, ok := winningBids[bid.bid.ImpID]
+		if !ok || cpm > wbid.bid.Price {
+			winningBids[bid.bid.ImpID] = bid
+		}
+		if bidMap, ok := winningBidsByImp[bid.bid.ImpID]; ok {
```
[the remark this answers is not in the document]
guscarreon: Done

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
+			if !ok || cpm > bestSoFar.bid.Price {
+				bidMap[pbsBid.Bidder] = bid
+			}
+		} else {
+			winningBidsByImp[bid.bid.ImpID] = make(map[openrtb_ext.BidderName]*pbsOrtbBid)
+			winningBidsByImp[bid.bid.ImpID][pbsBid.Bidder] = bid
 		}
-		winningBidsByBidder[pbsBid.Bid.ID][pbsBid.Bidder] = bid
+
+		winningBidsByImp[pbsBid.Bid.ImpID][pbsBid.Bidder] = bid
```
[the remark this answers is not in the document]
guscarreon: It does rewrites logic and can screw-up the end result. Can't believe I missed it, good call.

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
+		}
+		for k, cachedElements := range ht {
+			if cachedElements.freq > 0 {
+				t.Errorf("%s:  [CACHE_ERROR] Cache inconsistensy. Element %s was not expected to get cached\n", fileDisplayName, k)
+			} else if cachedElements.freq < 0 {
+				t.Errorf("%s:  [CACHE_ERROR] Cache inconsistensy. We cached some more elements (%s) than expected.\n", fileDisplayName, k)
+			} else if targData.includeCacheVast {
+				for i := 0; i < cachedElements.freq; i++ {
+					found := false
+					for j := 0; j < cachedElements.freq; j++ {
```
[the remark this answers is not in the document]
guscarreon: Thank God you spotted this, it was completely wrong, originally it was correct but for some reason I modified it in the final version.

exchange/cachetest/defaultvideo.json
```diff
@@ -27,16 +27,22 @@
...
         {
             "Type": "json",
-            "TTLSeconds": 660
+            "TTLSeconds": 660,
+            "Data":"\"\u003cVAST version=\"3.0\"\u003e\u003cAd\u003e\u003cWrapper\u003e\u003cAdSystem\u003eprebid.org wrapper\u003c/AdSystem\u003e\u003cVASTAdTagURI\u003e\u003c![CDATA[]]\u003e\u003c/VASTAdTagURI\u003e\u003cImpression\u003e\u003c/Impression\u003e\u003cCreatives\u003e\u003c/Creatives\u003e\u003c/Wrapper\u003e\u003c/Ad\u003e\u003c/VAST\u003e\"" 
         }, {
             "Type": "json",
-            "TTLSeconds": 3660
+            "TTLSeconds": 3660,
+            "Data": "\"\u003cVAST version=\"3.0\"\u003e\u003cAd\u003e\u003cWrapper\u003e\u003cAdSystem\u003eprebid.org wrapper\u003c/AdSystem\u003e\u003cVASTAdTagURI\u003e\u003c![CDATA[]]\u003e\u003c/VASTAdTagURI\u003e\u003cImpression\u003e\u003c/Impression\u003e\u003cCreatives\u003e\u003c/Creatives\u003e\u003c/Wrapper\u003e\u003c/Ad\u003e\u003c/VAST\u003e\""
```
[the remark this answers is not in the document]
guscarreon: We are currently caching Json encoded data of the bids (winning bids only from now on because of the modifications made by this PR). The `prebid_cache_client.Cacheable` struct we currently use to cache bids includes this Data field were we marshall bid data such as cat, imp, price and so on. This data field was included in the expected cachables of the test Json files in order to get compared to what we actually cache.

exchange/cachetest/customcachekey.json
```diff
@@ -27,16 +27,23 @@
     "expectedCacheables": [
         {
             "Type": "json",
-            "TTLSeconds": 660
-        }, {
+            "TTLSeconds": 660,
+            "Data":"{\"id\":\"appbid001\",\"impid\":\"oneImp\",\"price\":7.64,\"cat\":[\"11_sports_22\"]}"
```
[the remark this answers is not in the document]
guscarreon: They are necessary, in order to escape the quotes. The string inside the "Data" field is what the marshalled bid looks like.

exchange/customcachekeytest/customcachekey.json
```diff
@@ -28,16 +28,22 @@
...
         "banner": 300,
         "video": 3600,
         "audio": 1800,
         "native": 300
-    }
+    },
+    "targetDataIncludeWinners":true,
+    "targetDataIncludeBidderKeys":true,
+    "targetDataIncludeCacheBids":true,
+    "targetDataIncludeCacheVast":false
```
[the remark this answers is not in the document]
guscarreon: The code that checks keys? I don't understand. What program exactly? Inside what function?

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
guscarreon: You mean `"targetDataIncludeCacheBids": true` and `"targetDataIncludeCacheVast": false` ? Why are they wrong?

exchange/auction_test.go
guscarreon: It is, good call.

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
 		bid = &pbsOrtbBid{
 			bid:     pbsBid.Bid,
 			bidType: pbsBid.BidType,
 		}
-		if _, ok := winningBids[pbsBid.Bid.ImpID]; !ok {
-			winningBids[pbsBid.Bid.ImpID] = bid
+		cpm := bid.bid.Price
+		wbid, ok := winningBids[bid.bid.ImpID]
+		if !ok || cpm > wbid.bid.Price {
+			winningBids[bid.bid.ImpID] = bid
```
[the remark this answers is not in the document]
guscarreon: Done

exchange/auction_test.go
```diff
@@ -94,92 +112,143 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
 	return &spec, nil
 }
 
-func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec, bids bool, vast bool) {
+func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec) {
 	// bid := make([]pbsOrtbBid, 5)
 	var bid *pbsOrtbBid
 	winningBids := make(map[string]*pbsOrtbBid)
-	winningBidsByBidder := make(map[string]map[openrtb_ext.BidderName]*pbsOrtbBid)
+	winningBidsByImp := make(map[string]map[openrtb_ext.BidderName]*pbsOrtbBid)
```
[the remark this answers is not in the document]
guscarreon: Rolled back the variable name.

exchange/cachetest/defaultvideo.json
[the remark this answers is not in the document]
josephveach: Why are the values identical for the two different bids?

exchange/customcachekeytest/customcachekey.json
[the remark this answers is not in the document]
josephveach: lines 227+ in auction_test.go.

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
josephveach: Because for non-vast requests, we don't generate custom cache keys.

exchange/auction_test.go
```diff
@@ -94,92 +112,151 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
+				ht[compareString] = &cacheComparator{0, make([]string, 0), make([]string, 0)}
+			}
+			ht[compareString].freq -= 1
+			if targData.includeCacheVast {
+				ht[compareString].actualKeys = append(ht[compareString].actualKeys, cFound.Key)
+			}
+		}
+		for k, cachedElements := range ht {
+			if cachedElements.freq > 0 {
+				t.Errorf("%s:  [CACHE_ERROR] Cache inconsistensy. Element %s was not expected to get cached\n", fileDisplayName, k)
```
josephveach: I just noticed that this error message is confusing. It should say that the element was cached fewer times than expected.  Also, here and two lines below, "inconsistensy" should be "inconsistency".

exchange/auction_test.go
```diff
@@ -94,92 +112,151 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
+					found := false
+					for j := 0; j < len(cachedElements.actualKeys); j++ {
+						if strings.HasPrefix(cachedElements.expectedKeys[i], cachedElements.actualKeys[j]) {
+							found = true
+						}
+					}
+					if !found {
+						t.Errorf("[CACHE_ERROR] Key \"%s\" was expected to be cached but was not.\n", cachedElements.expectedKeys[i])
+					}
+				}
```
josephveach: I just realized that the above code confirms only that expected custom keys are found, but does not check that unexpected custom keys were found.  This can probably be done most easily by comparing length of expectedKeys array to length of actualKeys array.  I suspect you can do that before the above loop and execute the loop only if the lengths match.

exchange/auction_test.go
```diff
@@ -94,92 +112,151 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
+			// init cacheComparator element before hashing it. If needed
+			if _, ok := ht[compareString]; !ok {
+				ht[compareString] = &cacheComparator{0, make([]string, 0), make([]string, 0)}
+			}
+			ht[compareString].freq -= 1
+			if targData.includeCacheVast {
+				ht[compareString].actualKeys = append(ht[compareString].actualKeys, cFound.Key)
+			}
+		}
+		for k, cachedElements := range ht {
```
josephveach: Please add brief comments for each of the three for-loops above explaining the purpose of each.  At line 203, we create a map of expected cache values and associated custom cache keys, if any, and the number of times they should occur, at line 216, we examine the actual cache values and decrement the corresponding counters in the map of expected values, and at line 229, we make sure the final counter values are zero, which would mean we got the exact counts we expected.

exchange/cachetest/defaultvideo.json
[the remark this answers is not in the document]
guscarreon: I copied what the system cached. In all honesty I don understand what the banner request caches, but not what the video is caching. In the case of the banners I wrote the Data field of the expected output myself but not in this file that tests for video cache.

exchange/customcachekeytest/customcachekey.json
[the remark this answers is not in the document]
guscarreon: Joe, these are the lines 227 and beyond of auction_test.go. What part is wrong exactly?
```224             ht[compareString].freq -= 1
225             if targData.includeCacheVast {
226                 ht[compareString].actualKeys = append(ht[compareString].actualKeys, cFound.Key)
227             }
228         }
229         for k, cachedElements := range ht {
230             if cachedElements.freq > 0 {
231                 t.Errorf("%s:  [CACHE_ERROR] Cache inconsistensy. Element %s was not expected to get cached\n", fileDisplayName, k)
232             } else if cachedElements.freq < 0 {
233                 t.Errorf("%s:  [CACHE_ERROR] Cache inconsistensy. We cached some more elements (%s) than expected.\n", fileDisplayName, k)
234             } else if targData.includeCacheVast {
235                 for i := 0; i < len(cachedElements.expectedKeys); i++ {
236                     found := false
237                     for j := 0; j < len(cachedElements.actualKeys); j++ {
238                         if strings.HasPrefix(cachedElements.expectedKeys[i], cachedElements.actualKeys[j]) {
239                             found = true
240                         }
241                     }
242                     if !found {
243                         t.Errorf("[CACHE_ERROR] Key \"%s\" was expected to be cached but was not.\n", cachedElements.expectedKeys[i])
244                     }
245                 }
246             }
247         }
248     }
249 }
```

exchange/auction_test.go
guscarreon: Lines 193 to 197 perform the checks you mention we only go ahead to compare the cached values in line 198 when lengths are not equal 
```191     _ = testAuction.doCache(ctx, cache, targData, &specData.BidRequest, 60, &specData.DefaultTTLs, bidCategory)
192
193     if len(specData.ExpectedCacheables) > len(cache.items) {
194         t.Errorf("%s:  [CACHE_ERROR] Less elements were cached than expected \n", fileDisplayName)
195     } else if len(specData.ExpectedCacheables) < len(cache.items) {
196         t.Errorf("%s:  [CACHE_ERROR] More elements were cached than expected \n", fileDisplayName)
197     } else { // len(specData.ExpectedCacheables) == len(cache.items)
198         var ht map[string]*cacheComparator = make(map[string]*cacheComparator)
199         var formattedData string
200         var compareString string
201
202         for _, cExpected := range specData.ExpectedCacheables {
203             formattedData = strings.Replace(string(cExpected.Data), "\\", "", -1)
204 +-- 40 lines: formattedData = strings.Replace(formattedData, " ", "", -1)----------------------------------------------------------------------------------------------------------------------------
244                 }
245             }
246         }
247     }
248 }
```

exchange/auction_test.go
guscarreon: Jeez, the spelling mistake is embarrassing. I just corrected it. I believe that both error messages in lines 230 and 232 are correct aren't they? What changes do you think could make them more descriptive?
```228         for k, cachedElements := range ht {
229             if cachedElements.freq > 0 {
230                 t.Errorf("%s:  [CACHE_ERROR] Cache inconsistency. Element %s was not expected to get cached\n", fileDisplayName, k)
231             } else if cachedElements.freq < 0 {
232                 t.Errorf("%s:  [CACHE_ERROR] Cache inconsistency. We cached some more elements (%s) than expected.\n", fileDisplayName, k)
233             } else if targData.includeCacheVast {
234                 for i := 0; i < len(cachedElements.expectedKeys); i++ {
```

exchange/auction_test.go
guscarreon: comments added. PR updated

exchange/cachetest/defaultvideo.json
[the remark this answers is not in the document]
josephveach: So it seems there are multiple issues here. 
 
First problem is that the last two lines of this file, as well as those in all the other files in cachetest/, are wrong.  In the old auction_test.go code you can see that _TestCacheJSON_() calls _runCacheSpec(t, fileDisplayName, specData, true, false)_.  

Secondly, it should be intuitive that two different bids should not have identical values cached.  Again, if you look at the old code, you'll see that the data portion of cached vast bids was never tested.  Since this should not be vast (perhaps you already can figure it out since you understand banner request caches), the expected data is the marshal'd bid.  

Third, now that I understand that the "Data" field you've added to the JSON files is simply the marshal'd bid, I believe adding the "Data" fields, which are simply marshal'd versions of bid data already present is not a good idea.  There is a potential problem that some future JSON marshal() code will change the order of the fields.  In that case, the old approach would be okay because the same marshaling procedure will be applied to both the test data and the real data, while this new approach will cause failures because only the real data will use the new marshaling.  I believe you should undo that change and go back to the original implementation that did the marshal operation in real time, i.e.:
                cjson, _ := json.Marshal(bid.bid)
                specData.ExpectedCacheables[i].Data = cjson

exchange/auction_test.go
josephveach: Line 230 in your snippet above should say that the element was cached fewer times than expected.

exchange/auction_test.go
josephveach: I'm referring to the comparisons of **expectedKeys** and **actualKeys** (which is the code that was in the loop immediately above my comment), not the examination of lengths of **specData.ExpectedCacheables** and **cache.items**.

exchange/cachetest/defaultvideo.json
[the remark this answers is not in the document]
guscarreon: _Joe, answers shown below:_
**1)** The last two lines of every json file end up becoming the last two boolean values of the old `runCacheSpec(t, fileDisplayName, specData, true, false)` function because its signature ended up changing to `runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec)` in this new version the last two values found in the json files, get assigned to the target data object in lines 182 and 183 below:
```
118 func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec) {
119     var bid *pbsOrtbBid
120     winningBidsByImp := make(map[string]*pbsOrtbBid)
121 +-- 38 lines: winningBidsByBidder := make(map[string]map[openrtb_ext.BidderName]*pbsOrtbBid)-------------------
159     targData := &targetData{
160         priceGranularity: openrtb_ext.PriceGranularity{
161             Precision: 2,
162             Ranges: []openrtb_ext.GranularityRange{
163                 {
164                     Min:       0,
165                     Max:       5,
166                     Increment: 0.05,
167                 },
168                 {
169                     Min:       5,
170                     Max:       10,
171                     Increment: 0.1,
172                 },
173                 {
174                     Min:       10,
175                     Max:       20,
176                     Increment: 0.5,
177                 },
178             },
179         },
180         includeWinners:    specData.TargetDataIncludeWinners,
181         includeBidderKeys: specData.TargetDataIncludeBidderKeys,
182         includeCacheBids:  specData.TargetDataIncludeCacheBids,
183         includeCacheVast:  specData.TargetDataIncludeCacheVast,
184     }
185 
```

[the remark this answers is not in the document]
guscarreon: **2)** Excepting for the Data field, the video tests below don't cache the exact same data, se are comparing the cache in two parts:
  + The first part is the Data and the TTLseconds fields in lines 207 and 222 of auction _test.go
  + If the first part matches and we are testing a video request, we compare a second part of the cache, the keys. This is performed in the true branch of the else statment in line 239 in auction_test.go 
```
117 // finds the highest bid of every Imp.
118 func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec) {
119 +--115 lines: var bid *pbsOrtbBid-------------------------------------------------------------------------------------
234         for k, cachedElements := range ht {
235             if cachedElements.freq > 0 {
236                 t.Errorf("%s:  [CACHE_ERROR] Cache inconsistency. Element %s was not expected to get cached\n", fileDisplayName, k)
237             } else if cachedElements.freq < 0 {
238                 t.Errorf("%s:  [CACHE_ERROR] Cache inconsistency. We cached some more elements (%s) than expected.\n", fileDisplayName, k)
239             } else if targData.includeCacheVast {
240                 for i := 0; i < len(cachedElements.expectedKeys); i++ {
241                     found := false
242                     for j := 0; j < len(cachedElements.actualKeys); j++ {
243                         if strings.HasPrefix(cachedElements.expectedKeys[i], cachedElements.actualKeys[j]) {
244                             found = true
245                         }
246                     }
247                     if !found {
248                         t.Errorf("[CACHE_ERROR] Key \"%s\" was expected to be cached but was not.\n", cachedElements.expectedKeys[i])
249                     }
250                 }
251             }
252         }
253     }
254 }
```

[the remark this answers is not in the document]
guscarreon: **3)** That's not exactly a good idea because the data doesn't come cached in the exact same order than listed in the json file. If we just marshal the fields as they come in the json file and try to compare side by side, we'll get something like the following:
```
              Expected 2                                                                      Actual 2
Type      | json                                                                  | json
Data      | "{ \"id\": \"bidTwo\", \"impid\": \"oneImp\", \"price\": 5.64 }" | {"id":"bidOne","impid":"oneImp","price":7.64,"exp":600} 
TTLSeconds| 360                                                                   | 660
Key       | ""                                                                    | "" 
          |                                                                       |   
Type      | json                                                                  | json
Data      | "{ \"id\": \"bidOne\", \"impid\": \"oneImp\", \"price\": 7.64, \"exp\": 600}" | {"id":"bidTwo","impid":"oneImp","price":5.64} 
TTLSeconds| 660                                                                   | 360
Key       | ""                                                                    | "" 
          |                                                                       |   
```
The data is clearly there but it wasn't cached in the exact same order. If we do a direct comparison of the 5.64 bid vs the 7.64 bid they won't match. And if we want to go find a match of the first element of the left in the entire "Actual" list on the right, we'll have an O(n^2) process that is quite inefficient. The O(n+m) hashing approach that we coded is far better specially if we increase the size of the tests in the future. Therefore, I thought the approach of listing in the expected cachables part of the json file what you expect to assert in the test output, was better.

exchange/customcachekeytest/customcachekey.json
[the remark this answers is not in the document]
guscarreon: The old `runCacheSpec(t, fileDisplayName, specData, true, false)` function is no longer executed in `TestCustomCacheKeyJSON()` because it was completely reingeneered and its signature ended up changing to `runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec)` . In this new version the last two values found in the json filescarry the last two boolean values hardcoded in the old `runCacheSpec(t, fileDisplayName, specData, true, false)` . Therefore, those boolean values are now coded in the json files themselves rather than the runCacheSpec function.

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
guscarreon: We could but the outcome would not change at all. Do you want me to update the PR again?

exchange/auction_test.go
guscarreon: What do you think about:
```
t.Errorf("%s:  [CACHE_ERROR] Cache inconsistency. Element '%s' was not cached as was expected. \n", fileDisplayName, k)
``` 
Consider this scenario:
```
Expected      Actual
bidA           bidA
bidB           bidB
bidC
```
This will turn into a hash talbe like this:
```
ht       freq    expectedKeys
bidA360    0          CatA
bidB220    0          CatB
bidC360    1          CatC    //Element 'bidC360' was not cached as we expected
```

In this other scenario:
```
Expected      Actual
bidA           bidA
bidB           bidB
               bidC
```
This will turn into a hash talbe like this:
```
ht       freq    expectedKeys
bidA360    0          CatA
bidB220    0          CatB
bidC360   -1          CatC    //We cached some more elements ('bidC360') than we expected
```

exchange/auction_test.go
guscarreon: We are comparing them, its comparison is reflected in the 'freq' field with the difference that the freq field is better because it gives us the exact number of elements that mismatch, which we probably should mention in the comments, right? Again as stated in a response above, I agree with you that the error messages should change to:
```
	if cachedElements.freq > 0 {
		t.Errorf("%s:  [CACHE_ERROR] Cache inconsistency. We cached %d less element of %s than expected\n", fileDisplayName,cachedElements.freq, k)
	} else if cachedElements.freq < 0 {
		t.Errorf("%s:  [CACHE_ERROR] Cache inconsistency. We cached element %d more elements of %s than expected\n", fileDisplayName,cachedElements.freq, k)
	} else if targData.includeCacheVast {
		for i := 0; i < len(cachedElements.expectedKeys); i++ {
``` 
So we can have an output like shown below:
Cached values:
```
Expected      Actual
bidA           bidA
bidA           bidA
bidB           bidB
bidC           bidB
bidC
```
Hash table:
```
ht       freq    expectedKeys  actualKeys
bidA360    0          CatA       CatA
bidB220   -1          CatB       CatB
bidC360    2          CatC    
```
Output message: `We cached 1 more elements 'CatB' than we expected` . I realize that we also have a mismatch in BidC but once bidB gets sorted out, BidC will get its own error message when we run the tests again.

exchange/cachetest/defaultvideo.json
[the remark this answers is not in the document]
josephveach: (1)  Understood, but you've entered the wrong values on the last two lines.

(2) This is really tangential since the real problem is due to the wrong entries on the last two lines.  However, you should be aware that PBS caches only the data.  The key values, when present in the tests, represent the custom cache keys that are created by doCache() in certain cases to be used for retrieving the cached data.  See the useCustomCacheKey variable in doCache() if you want to understand how/when that occurs.  Since you've forced me to look a little further, I now see that the reason the Data values are identical for different bids is probably due to the fact that the NURL field of the bids must be either empty or identical.  If you look at doCache(), you'll see in the "bids" case, the Data value comes from
 **json.Marshal(topBidPerBidder.bid)** 
so values for different bids will always be different, while in the "vast" case, it comes from:
**vast := makeVAST(topBidPerBidder.bid)**
**jsonBytes, err := json.Marshal(vast)**
where makeVast() is simply:
func makeVAST(bid *openrtb.Bid) string {
        if bid.AdM == "" {
                return `<VAST version="3.0"><Ad><Wrapper>` +
                        `<AdSystem>prebid.org wrapper</AdSystem>` +
                        `<VASTAdTagURI><![CDATA[` + bid.NURL + `]]></VASTAdTagURI>` +
                        `<Impression></Impression><Creatives></Creatives>` +
                        `</Wrapper></Ad></VAST>`
        }
        return bid.AdM
}
You can see that if bid.NURL is empty or identical for both bids, the Data values cached will be identical.  Presumably, that means that bid.NURL must exist and be unique in real (non-test) cases (though I suppose it could mean we have a flawed caching approach, too).  Thus, if you want to test the Data values for vast bids, you simply need to specify a unique bid.NURL value for each bid and then confirm the presence of what would be created from the above calls.  Or you could go back to skipping checking the Data values for vast bids, as the old test code did.

(3) You may be right, in which case both approaches are flawed.  If so, the correct approach is to unmarshal both sets of data into appropriate structures and compare the corresponding fields of interest in the two structures.  BTW, the performance characteristics here are pretty irrelevant given that the numbers will always be quite small and that these are only tests.

exchange/auction_test.go
josephveach: There's no connection between cachedElements.freq and either expectedKeys or actualKeys.  If there were, you wouldn't need any code to compare keys when cachedElements.freq is 0.  Example:
- only expected data ht value is bid001 and it has no expected key.  When it gets added, freq is incremented to 1.
- only cached value is bid001 and it has a key.  When it is found, freq is decremented to 0.
This test should fail because we weren't expecting a key value but we got one.  Current code would miss that.

exchange/auction_test.go
```diff
@@ -94,92 +112,131 @@ func loadCacheSpec(filename string) (*cacheSpec, error) {
...
-			if cExpected.TTLSeconds == cFound.TTLSeconds {
-				found++
+			for j := 0; j < len(cache.items); j++ {
+				if formattedExpectedData == string(cache.items[j].Data) &&
+					specData.ExpectedCacheables[i].TTLSeconds == cache.items[j].TTLSeconds &&
+					specData.ExpectedCacheables[i].Type == cache.items[j].Type {
+					if specData.ExpectedCacheables[i].Type == prebid_cache_client.TypeJSON {
+						matched++
+					}
+					if specData.ExpectedCacheables[i].Type == prebid_cache_client.TypeXML &&
```
josephveach: The above four lines should be omitted.  Regardless of the "type", we should check the key value.  Otherwise, we will miss cases where we incorrectly get a key for JSON or, if keys are ever possible for anything other than XML, we'll fail to check their values.  Simply remove the above four lines and insert an "if" at the start of the following line.

exchange/auction_test.go
```diff
@@ -192,6 +249,12 @@ type mockCache struct {
 	items []prebid_cache_client.Cacheable
 }
 
+type cacheComparator struct {
+	freq         int
+	expectedKeys []string
+	actualKeys   []string
+}
```
josephveach: I think cacheComparator struct is no longer used and should be removed.

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
josephveach: This file is still wrong as is.  In its current state, it's a repeat of cachetest/multibid.json but with expected key values that are wrong but never evaluated because of the current state of auction_test.go.  Reversing the last two lines would cause it to make sense and not be a repeat of cachetest/multibid.json, though it would probably now be necessary to change the Type values from json to xml (and maybe the bidTypes of the bids), and we would still be still missing the other associated test cases we've come to expect (i.e., tests toggling the values of the two lines above the last two lines).

josephveach: Also, if the file is to be kept as a test for custom cache keys for multiple imp requests, the category values of the bids should be unique, and that will affect the expected key values.  In addition, in that case, only the winning bids should have custom (expected) key values.

exchange/customcachekeytest/customcachekey.json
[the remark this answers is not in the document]
guscarreon: corrected boolean values on the  JSON files

exchange/impcustomcachekeytest/multiImpCustomCacheKey.json
guscarreon: The multibid test has been corrected. It still is a multi-imp, multi-bid all banner request. To test a multi-imp, multi-bid vast request, a new json file has been added so now we have two different tests in that folder: 
```
▾ exchange/
  ▸ cachetest/
  ▸ customcachekeytest/
  ▸ exchangetest/
  ▾ impcustomcachekeytest/
      multiImpBanners.json
      multiImpVast.json
```

exchange/auction_test.go
guscarreon: The hash table approach has been modified for an O(n^2) approach where fields are compared directly with one another.

exchange/auction_test.go
guscarreon: Didn't look that far ahead to when we implement anything other than XML, we should also be aware that, in that hypothetical scenario, the else branch in line 207 could also need modifications. Anyways. I guess that discussion shall be left for that day, in the meantime, the correction has been made
```
201         for i := 0; i < len(specData.ExpectedCacheables); i++ {
202             if specData.ExpectedCacheables[i].Type == prebid_cache_client.TypeJSON {
203                 ExpectedData := strings.Replace(string(specData.ExpectedCacheables[i].Data), "\\", "", -1)
204                 ExpectedData = strings.Replace(ExpectedData, " ", "", -1)
205                 formattedExpectedData = ExpectedData[1 : len(ExpectedData)-1]
206             } else {
207                 formattedExpectedData = string(specData.ExpectedCacheables[i].Data)
208             }
```

exchange/auction_test.go
guscarreon: removed

guscarreon:
Tests inside `exchange/impcustomcachekeytest/` have been corrected

exchange/auction_test.go
```diff
@@ -209,15 +209,10 @@ func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec) {
...
-					}
-					if specData.ExpectedCacheables[i].Type == prebid_cache_client.TypeXML &&
-						len(specData.ExpectedCacheables[i].Key) <= len(cache.items[j].Key) &&
-						specData.ExpectedCacheables[i].Key == cache.items[j].Key[:len(specData.ExpectedCacheables[i].Key)] {
-						matched++
-					}
+					specData.ExpectedCacheables[i].Type == cache.items[j].Type &&
+					len(specData.ExpectedCacheables[i].Key) <= len(cache.items[j].Key) &&
+					specData.ExpectedCacheables[i].Key == cache.items[j].Key[:len(specData.ExpectedCacheables[i].Key)] {
+					matched++
```
josephveach: This is a very minor complaint and probably not worth changing.  I just realized that the key comparison could give a false positive if the expected key (prefix) is a subset of the actual key (prefix).  E.g., if expected key is 13_sports_22 but the actual key prefix is 13_sports_222, the test would pass incorrectly.  So, ideally, an underscore should be the last character of the expected key and/or (preferably) the logic here should add an underscore to the end of the expected key if one is not already present before comparing to the actual key.  In practice, I'm doubt it's worth the trouble.  The current logic already achieves a greater precision than what we had before.

exchange/impcustomcachekeytest/multiImpBanners.json
```diff
@@ -0,0 +1,72 @@
...
+			"cat": ["33_news_44"]
+		},
+		"bidType": "banner",
+		"bidder": "pubmatic"
+	}, {
+		"bid": {
+			"id": "bidThree",
+			"impid": "oneImp",
+			"price": 2.3,
+			"cat": ["33_news_44"]
```
josephveach: Another minor nit probably not requiring a change:  logic in PBS will prevent two bids for the same imp from having the same cat value so the scenario here could never actually happen.  Since cat will not be used because none of these bids will get cached anyway, it's probably not worth worrying about.

exchange/auction_test.go
guscarreon: I see. It's a fair point I can still change it if you want.

exchange/impcustomcachekeytest/multiImpBanners.json
guscarreon: it's not a big change and I could still make it. What do you think?

exchange/auction_test.go
josephveach: It's hard to imagine a PBS bug sneaking through because of this, though it's theoretically possible.  If a test had two keys where one was a subset of the other (like 13_sports_22 and 13_sports_222), it would fail because the cached key matching the latter would also match the former and so throw off the count.  However, that failure would be found by the person introducing the test.  IMHO, I don't think it's worth the trouble but we'll need at least one more reviewer to agree.  :-)

exchange/impcustomcachekeytest/multiImpBanners.json
josephveach: Unless someone else objects, I'm okay with the code as is.

PR #868, opened 2019-04-03
Allow app requests for Conversant

johnwier:
Allow inapp requests for banner and video to be send to the Conversant servers

### DIFF
--- a/adapters/conversant/conversant.go
+++ b/adapters/conversant/conversant.go
@@ -51,14 +51,6 @@ func (a *ConversantAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidde
 		return nil, err
 	}
 
-	// Without this, the code crashes with a nil-pointer dereference below, on
-	// cnvrReq.Site.ID = params.SiteID
-	if cnvrReq.Site == nil {
-		return nil, &errortypes.BadInput{
-			Message: "Conversant doesn't support App requests",
-		}
-	}
-
 	// Create a map of impression objects for both request creation
 	// and response parsing.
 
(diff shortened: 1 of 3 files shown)


johnwier:
I can't migrate to the new methods at the moment, but I will get it into our backlog.  For now I've fixed the yaml file.

adapters/conversant/conversant.go
```diff
@@ -51,14 +51,6 @@ func (a *ConversantAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidde
...
 
-	// Without this, the code crashes with a nil-pointer dereference below, on
-	// cnvrReq.Site.ID = params.SiteID
-	if cnvrReq.Site == nil {
-		return nil, &errortypes.BadInput{
-			Message: "Conversant doesn't support App requests",
-		}
-	}
-
 	// Create a map of impression objects for both request creation
```
guscarreon: If we eliminate this statement, we open the door to the posibility of returning from the Call function despite having `PBSRequestInstance.Site` equal to nil. Aren't we taking the risk of de-referencing the null Site field in line 547 of `conversant_test.go` ? Maybe under the current test conditions we aren't, but a different parsed request could make us take this risk. Isn't it?

hhhjort: I am not sure what you mean? There is a test that is currently hardcoded to expect a non-null `site` from a request that is hard coded to contain a `site` object. That could be an issue if someone naively modifies the test, and it fails unexpectedly at the dereference. But such a comment would belong on the test, not here, as it is a bug with the test, not a bug with removing this code.

The purpose of this PR is to allow app requests for this adapter. An app request is defined as a request that has `site==nil` and `app!=nil`. The code to be removed previously enforced that app requests were not allowed, so must be removed for this PR to function.

guscarreon: Fair point

PR #870, opened 2019-04-04
Change eplanning adapter sync URL

Nistenf:
A new improved endpoint was created for our usersync process and this PR changes the syncer to reflect that.


Nistenf:
Thanks for the reviews. I have a question: does the externalURL in config.go include the protocol in your instance?

hhhjort:
Yes, our configuration includes the protocol.

Nistenf:
> Yes, our configuration includes the protocol.

Great, thank you

PR #872, opened 2019-04-08
Video endpoint request validation refactoring

Kalypsonika:
Validation refactoring

-Added pod level error: return error for that particular pod. process the good pods normally
-Added validation for 0 and negative ad pod duration
-Added mime types clean up
-Added validation for duplicated pod ids
-Added validation for missing pod ids
-Added pod config id handling
-Added handler for the case when all pods are incorrect

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -129,7 +129,7 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		}
 	}
 	//unmarshal and validate combined result
-	videoBidReq, errL := deps.parseVideoRequest(resolvedRequest)
+	videoBidReq, errL, podErrors := deps.parseVideoRequest(resolvedRequest)
 	if len(errL) > 0 {
 		handleError(labels, w, errL, ao)
 		return
@@ -147,13 +147,26 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 	//create full open rtb req from full video request
 	mergeData(videoBidReq, bidReq)
 
+	initialPodNumber := len(videoBidReq.PodConfig.Pods)
+	if len(podErrors) > 0 {
+		//remove incorrect pods
+		videoBidReq = cleanupVideoBidRequest(videoBidReq, podErrors)
+	}
+
 	//create impressions array
-	imps, errs := deps.createImpressions(videoBidReq)
-	if errs != nil {
-		errL = append(errL, errs...)
+	imps, podErrors := deps.createImpressions(videoBidReq, podErrors)
+
+	if len(podErrors) == initialPodNumber {
+		resPodErr := make([]string, 0)
+		for _, podEr := range podErrors {
+			resPodErr = append(resPodErr, strings.Join(podEr.PodErrors, ", "))
+		}
+		err := errors.New(fmt.Sprintf("all pods are incorrect: %s", strings.Join(resPodErr, "; ")))
+		errL = append(errL, err)
 		handleError(labels, w, errL, ao)
 		return
 	}
+
 	bidReq.Imp = imps
 	bidReq.ID = "bid_id" //TODO: look at prebid.js
 
@@ -199,7 +212,7 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 	}
 
 	//build simplified response
-	bidResp, err := buildVideoResponse(response)
+	bidResp, err := buildVideoResponse(response, podErrors)
 	if err != nil {
 		errL := []error{err}
 		handleError(labels, w, errL, ao)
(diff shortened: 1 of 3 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -480,22 +522,28 @@ func createBidExtension(videoRequest *openrtb_ext.BidRequestVideo) ([]byte, erro
...
 		errs = append(errs, errL...)
 	}
 	return
 }
 
-func (deps *endpointDeps) validateVideoRequest(req *openrtb_ext.BidRequestVideo) []error {
+type PodError struct {
+	PodId     int
+	PodIndex  int
+	PodErrors []string
```
mansinahar: Can we please call this either `errMsgs` or something along those lines? It's just that `PodError.PodErrors` can become real difficult to follow real soon when reading the code

endpoints/openrtb2/video_auction.go
```diff
@@ -514,18 +562,37 @@ func (deps *endpointDeps) validateVideoRequest(req *openrtb_ext.BidRequestVideo)
...
-			err := fmt.Errorf("request missing required field: PodConfig.Pods.PodId, Pod index: %d", ind)
-			errL = append(errL, err)
+			err := fmt.Sprintf("request missing required field: PodConfig.Pods.PodId, Pod index: %d", ind)
+			podErr.PodErrors = append(podErr.PodErrors, err)
 		}
 		if pod.AdPodDurationSec == 0 {
-			err := fmt.Errorf("request missing required field: PodConfig.Pods.AdPodDurationSec, Pod index: %d", ind)
-			errL = append(errL, err)
+			err := fmt.Sprintf("request missing required field: PodConfig.Pods.AdPodDurationSec, Pod index: %d", ind)
+			podErr.PodErrors = append(podErr.PodErrors, err)
```
mansinahar: I am assuming it is possible that someone set the ad pod duration in the request as `0`? In that case we should probably modify the error message to capture both possibilities of ad pod duration being missing or being set to '0'.

endpoints/openrtb2/video_auction.go
```diff
@@ -147,13 +147,26 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+		videoBidReq = cleanupVideoBidRequest(videoBidReq, podErrors)
+	}
+
 	//create impressions array
-	imps, errs := deps.createImpressions(videoBidReq)
-	if errs != nil {
-		errL = append(errL, errs...)
+	imps, podErrors := deps.createImpressions(videoBidReq, podErrors)
+
+	if len(podErrors) == initialPodNumber {
```
mansinahar: Wouldn't it make more sense to check `len(imps) == 0` here? The `initialPodNumber` could be higher here as compared to that of the request based on which pods were removed above in line 153, no?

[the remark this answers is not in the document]
Kalypsonika: Mansi, thank you for your comments! 
Good catch, `len(imps) == 0` will do the same, but logically it will more tangled, because one pod may require more than one impression to be created. `podErrors` is a list of errors per pod and this code handles the case when all pods are incorrect. Also pod error may occur not just because of impression creation error. I would not mix this logic. 
`initialPodNumber` variable is initialized before any incorrect pods clean up, so I don't see any problems with this.  
Please let me know what you think.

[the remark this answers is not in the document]
mansinahar: Ah! I see what you mean about podErrors. Initially, I missed that podErrors was being passed to `createImpressions` and was thinking that `podErrors` returned from `createExpressions` might not reflect errors for all pods but only for the pods left after the clean up on line 153. You are right, this should be fine.

Also, regarding your comment about one pod potentially requiring  more than one imps, if all pods were incorrect then `createImpressions` should return `0` imps, right? And so `len(imps) == 0` should also give the same result, correct? (You don't need to change it to `len(imps)==0`. I am just making sure I understood you correctly)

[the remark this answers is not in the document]
Kalypsonika: I've just tried it and it all works including unit tests! Which is a good sign, we don't need to process request if we have no impressions to send downstream. 
Thank you for this, good catch!

endpoints/openrtb2/video_auction.go
```diff
@@ -364,6 +390,22 @@ func buildVideoResponse(bidresponse *openrtb.BidResponse) (*openrtb_ext.BidRespo
...
+	// If there were incorrect pods, we put them back to response with error message
+	if len(podErrors) > 0 {
+		for _, podEr := range podErrors {
+			pE := make([]string, 0, len(podEr.PodErrors))
+			for _, er := range podEr.PodErrors {
+				pE = append(pE, er)
+			}
+			adPodEr := &openrtb_ext.AdPod{
+				PodId:  int64(podEr.PodId),
+				Errors: pE,
```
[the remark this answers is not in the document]
Kalypsonika: Good catch! Removed.

endpoints/openrtb2/video_auction.go
Kalypsonika: Yes, done.

endpoints/openrtb2/video_auction.go
Kalypsonika: When AdPodDurationSec is missed in request, after unmarshalling it to object we have zero as default value. This is the same if it is set to zero initially. there is no way to distinguish it in code. 
I changed error message to this:
```request missing or incorrect required field: PodConfig.Pods.AdPodDurationSec, Pod index: %d```

Kalypsonika: ConfigId as well

Kalypsonika: @mansinahar Good comments, thank you!

mansinahar: Yeah, exactly. It's hard to figure out if the `0` was set by the user or is just the default value. Thanks for updating the error message to capture both possibilities.

PR #873, opened 2019-04-09
Improved error handling on Beachfront adapter

muncha:
Checking http statuses, in particular the possibility of a 200 with a empty array which should be treated as an empty response.

### DIFF
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -23,7 +23,7 @@ const VideoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id="
 
 const VideoEndpointSuffix = "&prebidserver"
 
-const beachfrontAdapterName = "BF_PREBID_S2S1"
+const beachfrontAdapterName = "BF_PREBID_S2S"
 const beachfrontAdapterVersion = "0.3.0"
 
 type BeachfrontAdapter struct {
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -16,8 +16,7 @@ import (
 const Seat = "beachfront"
 const BidCapacity = 5
 
-const BannerEndpoint = "http://yourmomshouse.com/mock/display.php"
-// const BannerEndpoint = "https://display.bfmio.com/prebid_display"
+const BannerEndpoint = "https://ioms.bfmio.com/prebid_display"
 const VideoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id="
 
 const VideoEndpointSuffix = "&prebidserver"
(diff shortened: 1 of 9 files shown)


adapters/beachfront/beachfront.go
```diff
@@ -499,6 +527,13 @@ func extractVideoCrid(nurl string) string {
 	return strings.TrimSuffix(chunky[2], ":")
 }
 
+// Thank you, brightroll.
```
mansinahar: Can I ask what's the purpose of this comment? If it's just a leftover of some sort, can we please remove it? Thanks!

adapters/beachfront/beachfront.go
```diff
@@ -394,12 +401,33 @@ func getVideoRequest(req *openrtb.BidRequest) (BeachfrontVideoRequest, []error,
 func (a *BeachfrontAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
 	var bids []openrtb.Bid
 	var bidtype = getBidType(internalRequest)
+
+	/*
+		Beachfront is now sending an empty array and 200 as their "no results" response. This should catch that.
+	*/
+
+	if response.StatusCode == http.StatusOK && len(response.Body) <= 2 {
```
mansinahar: Is there a reason why we can't just check `len(response.Body) == 0` rather than `len(response.Body) <= 2` ?

adapters/beachfront/beachfront.go
```diff
@@ -140,15 +142,14 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapte
 	var beachfrontRequests BeachfrontRequests
 	var reqJSON []byte
 	var uri string
-	var errs = make([]error, 0)
+	var errs = make([]error, 0, len(request.Imp))
 	var err error
```
guscarreon: I'm not sure if it's uncommon to have three or more Imps in a given request but, according to the code flow, I believe We'll have at most two Error object elements in this slice, so maybe we are wastiong memory by assigning more than we need whenever we have three or more Imps in a request

adapters/beachfront/beachfront.go
```diff
@@ -394,12 +401,33 @@ func getVideoRequest(req *openrtb.BidRequest) (BeachfrontVideoRequest, []error,
...
+	*/
+
+	if response.StatusCode == http.StatusOK && len(response.Body) <= 2 {
+		return nil, nil
+	}
+
+	if response.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
```
guscarreon: Can we reduce:
```408
409     if response.StatusCode == http.StatusOK && len(response.Body) <= 2 {
410         return nil, nil
411     }
412
413     if response.StatusCode == http.StatusNoContent {
414         return nil, nil
415     }
416
```

To a single statement?
```
if (response.StatusCode == http.StatusNoContent) || (response.StatusCode == http.StatusOK && len(response.Body) <= 2) {
      return nil, nil
}
```

muncha:
@guscarreon thanks for all that. I'll be putting in an "improvements" PR shortly.

PR #875, opened 2019-04-11
Video cache response validation

Kalypsonika:


endpoints/openrtb2/video_auction.go
```diff
@@ -134,6 +134,10 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		handleError(labels, w, errL, ao)
 		return
 	}
+	if len(podErrors) > 0 {
+		//remove incorrect pods
+		videoBidReq = cleanupVideoBidRequest(videoBidReq, podErrors)
+	}
```
guscarreon: This if statement is exactly the same as the one in line 155 below and nothing seems to happen with podErrors between here and there. Are the two checks necessary?

Kalypsonika: Good catch. This appeared because of incorrect merge of merge conflicts. Fixed now.

endpoints/openrtb2/video_auction.go
```diff
@@ -366,6 +366,9 @@ func buildVideoResponse(bidresponse *openrtb.BidResponse, podErrors []PodError)
 			if err := json.Unmarshal(bid.Ext, &tempRespBidExt); err != nil {
 				return nil, err
 			}
+			if tempRespBidExt.Prebid.Targeting["hb_uuid"] == "" {
```
mansinahar: Hey @Kalypsonika! I just noticed that we have constants defined for different Targeting Keys in `openrtb_ext/bid.go`. I'd suggest we probably use that here instead of the hardcoded string.
 
Also, I understand that we use hard coded strings for these in most of the places in the code and I think we should probably refactor that. Maybe, I will put up a PR for the same some time in the future but that shouldn't stop us from changing it here now. What do you think?

endpoints/openrtb2/video_auction_test.go
```diff
@@ -363,6 +363,119 @@ func TestVideoEndpointValidationsPodErrors(t *testing.T) {
...
+
+	bid3.Ext = extBid3
+	bids = append(bids, bid3)
+
+	seatBid.Bid = bids
+	seatBids = append(seatBids, seatBid)
+	openRtbBidResp.SeatBid = seatBids
+
+	bidRespVideo, err := buildVideoResponse(&openRtbBidResp, podErrors)
+	assert.Equal(t, nil, err, "Error shopuld be nil")
```
mansinahar: `testify/assert` package also has a [`assert.NoError(t, err)`](https://godoc.org/github.com/pgpst/pgpst/internal/github.com/stretchr/testify/assert#NoError) assertion available. Not saying it needs to be changed in this PR to use this but just thought might be interesting to know.

Also, similar for length check below: [`assert.Len(t, bidRespVideo.AdPods, 1, "err msg")`](https://godoc.org/github.com/pgpst/pgpst/internal/github.com/stretchr/testify/assert#Assertions.Len)

endpoints/openrtb2/video_auction_test.go
```diff
@@ -363,6 +363,119 @@ func TestVideoEndpointValidationsPodErrors(t *testing.T) {
...
+	podErrors = append(podErrors, podErr1)
+
+	podErr2 := PodError{}
+	podErr2.PodId = 333
+	podErr2.PodIndex = 2
+	podErrors = append(podErrors, podErr2)
+
+	bidRespVideo, err := buildVideoResponse(&openRtbBidResp, podErrors)
+	assert.Equal(t, nil, err, "Error shopuld be nil")
+	assert.Equal(t, 3, len(bidRespVideo.AdPods), "AdPods length should be 3")
```
mansinahar: This is more of a question: In this particular test case I see that you're sending 2 bids and both of them has `hb_uuid` targeting key set so then in the `buildVideoResponse` function, the if condition on line 369 should be true for both the bids and no adPods should be created, correct? But here we are expecting 3 adPods and I am likely missing something here. Would you mind explaining to me why is that so that I can understand this better?

endpoints/openrtb2/video_auction.go
Kalypsonika: @mansinahar Hi Mansi! I totally agree, we don't want to see any hardcoded strings like this. I'll change all of them in the video endpoint.

endpoints/openrtb2/video_auction_test.go
Kalypsonika: Good point! Changed!

endpoints/openrtb2/video_auction.go
mansinahar: Thank you so much. Appreciate it!

endpoints/openrtb2/video_auction_test.go
Kalypsonika: Oh I see what you mean! In this test case the idea is to test only pod level errors, not cache errors. Bid data should be correct, content doesn't matter. 
 There are also 2 more test cases to test cache errors: `TestVideoBuildVideoResponseMissedCacheForAllBids` - no cache for all bids and no pod erros
`TestVideoBuildVideoResponseMissedCacheForOneBid` - just one bid cache error and no pods errors
In `buildVideoResponse` function we need to handle both cache and pod errors to build proper response. I just try to test it with different combinations. 

To be more clear about purpose of `TestVideoBuildVideoResponsePodErrors` test case, I added one more assert:
`assert.Len(t, bidRespVideo.AdPods[0].Targeting, 2, "First ad pod should be correct and contain 2 targeting elements")`
Hope this explanation makes sense for you.

PR #878, opened 2019-04-17
Makes 'size' a required parameter for the ix adapter

hhhjort:
Back in PR #747 in November, ix updated their adapter to require sizes. At the time we blocked updating `/bidders/params` to set `size` as required, as it would make all legacy configs (ix configs lacking a `size` parameter) fail the entire request, not just the ix adapter part. It has been 5 months now, publishers should have had plenty of time to update their configs.

@ix-certification : will be leaving this PR up unmerged for a couple of weeks in case anyone has an issue with this change.

### DIFF
--- a/static/bidder-params/ix.json
+++ b/static/bidder-params/ix.json
@@ -19,5 +19,5 @@
       "description": "An array of two integer containing the dimension"
     }
   },
-  "required": ["siteId"]
+  "required": ["siteId", "size"]
 }


ix-prebid-support:
Hello we are actually removing the size requirement altogether. This will be included in a separate PR based on https://github.com/ix-prebid-support/prebid-server/tree/ix-remove-size-param

josephveach:
This request has been withdrawn in favor of https://github.com/prebid/prebid-server/pull/883.

PR #880, opened 2019-04-19
[Currency support] Map currency from bid request

benjaminch:
This CL aims to take into account `request.Cur. It's yet another step forward the full currency support CF #280.

Please let me know what you think about this one and whether I forgot anything.

Cheers :)

### DIFF
--- a/exchange/bidder.go
+++ b/exchange/bidder.go
@@ -129,20 +129,31 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
 		}
 
 		if httpInfo.err == nil {
-
 			bidResponse, moreErrs := bidder.Bidder.MakeBids(request, httpInfo.request, httpInfo.response)
 			errs = append(errs, moreErrs...)
 
 			if bidResponse != nil {
-
+				// Setup default currency as `USD` is not set in bid request nor bid response
 				if bidResponse.Currency == "" {
-					// Empty currency means default currency `USD`
 					bidResponse.Currency = defaultCurrency
+				}
+				if len(request.Cur) == 0 {
+					request.Cur = []string{defaultCurrency}
 				}
 
 				// Try to get a conversion rate
-				// TODO(#280): try to convert every to element of request.cur, and use the first one which succeeds
-				if conversionRate, err := conversions.GetRate(bidResponse.Currency, "USD"); err == nil {
+				// Try to get the first currency from request.cur having a match in the rate converter,
+				// and use it as currency
+				var conversionRate float64
+				var err error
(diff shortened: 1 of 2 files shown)


exchange/bidder.go
```diff
@@ -129,20 +129,31 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
 
 				// Try to get a conversion rate
-				// TODO(#280): try to convert every to element of request.cur, and use the first one which succeeds
-				if conversionRate, err := conversions.GetRate(bidResponse.Currency, "USD"); err == nil {
+				// Try to get the first currency from request.cur having a match in the rate converter,
+				// and use it as currency
+				var conversionRate float64
+				var err error
+				for _, bidReqCur := range request.Cur {
+					if conversionRate, err = conversions.GetRate(bidResponse.Currency, bidReqCur); err == nil {
```
hhhjort: Should we worry that the typical case is `bidResponse.Currency == bidReqCur`, or is the processing to get out the conversion rate `1` light enough that the performance hit isn't worth an extra check at this point to skip `GetRate()` or the loop entirely?

benjaminch: Hey !
Well, for sure the processing is more than a plain inline if here, but I think it worth passing via the method rather than inline for following:
- The method does also check currency iso code existence (bad currencies names would end up being discarded using the `rates` method where it won't be unless adding more logic inline doing pretty much the same with the same perfs hit)
- Having the logic / responsibility in one place (also easier to test :))
- Finally, for the perfs hit, I guess it won't be much is both currencies are the same, costing 2 currencies parsing and string comparison: https://github.com/prebid/prebid-server/blob/master/currencies/rates.go#L50 
```
// GetRate returns the conversion rate between two currencies
// returns an error in case the conversion rate between the two given currencies is not in the currencies rates map
func (r *Rates) GetRate(from string, to string) (float64, error) {
	var err error
	fromUnit, err := currency.ParseISO(from)
	if err != nil {
		return 0, err
	}
	toUnit, err := currency.ParseISO(to)
	if err != nil {
		return 0, err
	}
	if fromUnit.String() == toUnit.String() {
		return 1, nil
	}
	if r.Conversions != nil {
		if conversion, present := r.Conversions[fromUnit.String()][toUnit.String()]; present == true {
			return conversion, err
		}

		return 0, fmt.Errorf("Currency conversion rate not found: '%s' => '%s'", fromUnit.String(), toUnit.String())
	}
	return 0, errors.New("rates are nil")
}
```
What do you guys think?

benjaminch:
Thanks @hhhjort ;)

PR #881, opened 2019-04-19
[Tests] Moving test cases comments to dedicated field

benjaminch:
This CL only introduces a change in couple tests having test cases
comments moving those comments in a dedicated field in the test case so
it will be printed by assert on fail which ease the debug
investigations.

### DIFF
--- a/exchange/bidder_test.go
+++ b/exchange/bidder_test.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"testing"
 	"time"
 
@@ -254,8 +253,8 @@ func TestMultiCurrencies(t *testing.T) {
 		rates                     currencies.Rates
 		expectedBids              []bid
 		expectedBadCurrencyErrors []error
+		description               string
 	}{
-		// Bidder respond with the same currency (default one) on all HTTP responses
 		{
 			bids: []bid{
 				{currency: "USD", price: 1.1},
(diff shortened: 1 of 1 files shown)


exchange/bidder_test.go
```diff
@@ -660,9 +663,9 @@ func TestMultiCurrencies_RateConverterNotSet(t *testing.T) {
 		)
 
 		// Verify:
-		assert.Equal(t, false, (seatBid == nil && tc.expectedBidsCount != 0), fmt.Sprint("Case:", strings.Join(tc.bidCurrency, ",")))
-		assert.Equal(t, tc.expectedBidsCount, uint(len(seatBid.bids)), fmt.Sprint("Case:", strings.Join(tc.bidCurrency, ",")))
-		assert.ElementsMatch(t, tc.expectedBadCurrencyErrors, errs, fmt.Sprint("Case:", strings.Join(tc.bidCurrency, ",")))
+		assert.Equal(t, false, (seatBid == nil && tc.expectedBidsCount != 0), fmt.Sprint("Case:", strings.Join(tc.bidCurrency, ",")), tc.description)
```
mansinahar: Like this approach of adding the descriptions to the table test cases. One question though, since the descriptions are now added, do we still need the existing error messages: `fmt.Sprint("Case:", strings.Join(tc.bidCurrency, ","))` ?

benjaminch: Actually, you are right, we don't need it any longer, I can remove it :)
Thanks !

benjaminch: Just updated the PR, thanks @mansinahar :)

mansinahar: Awesome. Thanks @benjaminch !

benjaminch:
Thanks @mansinahar :)

mansinahar:
Of course! Thanks for the PR @benjaminch :)

PR #882, opened 2019-04-19
Implementation of Categories Http fetcher

Kalypsonika:
-Added code to fetch data using http
-Added previously loaded categories to run http request just at the first time

### DIFF
--- a/stored_requests/backends/http_fetcher/fetcher.go
+++ b/stored_requests/backends/http_fetcher/fetcher.go
@@ -91,10 +91,10 @@ func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer
 	var dataName, url string
 	if publisherId != "" {
 		dataName = fmt.Sprintf("%s_%s", primaryAdServer, publisherId)
-		url = fmt.Sprintf("%s/%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer, publisherId)
+		url = fmt.Sprintf("%s/%s/%s.json", strings.TrimSuffix(fetcher.Endpoint, "?"), primaryAdServer, publisherId)
 	} else {
 		dataName = primaryAdServer
-		url = fmt.Sprintf("%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer)
+		url = fmt.Sprintf("%s/%s.json", strings.TrimSuffix(fetcher.Endpoint, "?"), primaryAdServer)
 	}
 
 	if data, ok := fetcher.Categories[dataName]; ok {
--- a/stored_requests/backends/http_fetcher/fetcher.go
+++ b/stored_requests/backends/http_fetcher/fetcher.go
@@ -88,13 +88,13 @@ func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer
 
 	//in NewFetcher function there is a code to add "?" at the end of url
 	//in case of categories we don't expect to have any parameters, that's why we need to remove "?"
-	url := fmt.Sprintf("%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer)
-
-	dataName := primaryAdServer
-
+	var dataName, url string
 	if publisherId != "" {
 		dataName = fmt.Sprintf("%s_%s", primaryAdServer, publisherId)
 		url = fmt.Sprintf("%s/%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer, publisherId)
+	} else {
+		dataName = primaryAdServer
+		url = fmt.Sprintf("%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer)
 	}
 
 	if data, ok := fetcher.Categories[dataName]; ok {
--- a/stored_requests/backends/http_fetcher/fetcher.go
+++ b/stored_requests/backends/http_fetcher/fetcher.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/prebid/prebid-server/stored_requests/backends/file_fetcher"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -60,7 +59,7 @@ type HttpFetcher struct {
 	client     *http.Client
 	Endpoint   string
 	hasQuery   bool
-	Categories map[string]map[string]file_fetcher.Category
+	Categories map[string]map[string]stored_requests.Category
 }
 
 func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []string, impIDs []string) (requestData map[string]json.RawMessage, impData map[string]json.RawMessage, errs []error) {
@@ -84,7 +83,7 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
 
 func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer, publisherId, iabCategory string) (string, error) {
 	if fetcher.Categories == nil {
-		fetcher.Categories = make(map[string]map[string]file_fetcher.Category)
+		fetcher.Categories = make(map[string]map[string]stored_requests.Category)
 	}
 
 	if publisherId == "" {
@@ -110,7 +109,7 @@ func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer
 	defer httpResp.Body.Close()
 
 	respBytes, err := ioutil.ReadAll(httpResp.Body)
-	tmp := make(map[string]file_fetcher.Category)
+	tmp := make(map[string]stored_requests.Category)
 
 	if err := json.Unmarshal(respBytes, &tmp); err != nil {
 		return "", fmt.Errorf("Unable to unmarshal categories for adserver: '%s', publisherId: '%s'", primaryAdServer, publisherId)
--- a/stored_requests/backends/http_fetcher/fetcher.go
+++ b/stored_requests/backends/http_fetcher/fetcher.go
@@ -86,17 +86,20 @@ func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer
 		fetcher.Categories = make(map[string]map[string]stored_requests.Category)
 	}
 
-	if publisherId == "" {
-		publisherId = primaryAdServer
+	//in NewFetcher function there is a code to add "?" at the end of url
+	//in case of categories we don't expect to have any parameters, that's why we need to remove "?"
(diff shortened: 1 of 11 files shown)


stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -56,9 +57,10 @@ func NewFetcher(client *http.Client, endpoint string) *HttpFetcher {
 }
 
 type HttpFetcher struct {
-	client   *http.Client
-	Endpoint string
-	hasQuery bool
+	client     *http.Client
+	Endpoint   string
+	hasQuery   bool
+	Categories map[string]map[string]file_fetcher.Category
```
jmaynardxandr: We should move the file_fetcher.Category type into a common module so that http_fetcher doesn't have to depend on file_fetcher.

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -80,8 +82,44 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
+	if publisherId == "" {
+		publisherId = primaryAdServer
+	}
+
+	dataName := fmt.Sprintf("%s_%s", primaryAdServer, publisherId)
+	if data, ok := fetcher.Categories[dataName]; ok {
+		return data[iabCategory].Id, nil
+	}
+
+	url := fmt.Sprintf("%s/%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer, publisherId)
```
jmaynardxandr: Probably just removing the ? isn't going to work if the fetcher.Endpoint contains query params in the URL, we should just have this function throw an appropriate error if the endpoint has a ? here (instead of trying to handle that case).

Kalypsonika: I made it because in `NewFetcher` function there is this code: 

```
func NewFetcher(client *http.Client, endpoint string) *HttpFetcher {
	// Do some work up-front to figure out if the (configurable) endpoint has a query string or not.
	// When we build requests, we'll either want to add `?request-ids=...&imp-ids=...` _or_
	// `&request-ids=...&imp-ids=...`, depending.
	urlPrefix := endpoint
	if strings.Contains(endpoint, "?") {
		urlPrefix = urlPrefix + "&"
	} else {
		urlPrefix = urlPrefix + "?"
	}

	...

	return &HttpFetcher{
		client:   client,
		Endpoint: urlPrefix,
	}
} 
```
because of this there is always `?` at the end of url. 
In my understanding in case of Categories fetcher we expect to have url without parameters, so I have to handle this explicitly. 
Do you think I need to refactor existing functionality?

stored_requests/backends/http_fetcher/fetcher.go
Kalypsonika: Agree. Done

stored_requests/backends/http_fetcher/fetcher.go
jmaynardxandr: Oh ok I see why that's there then. I guess the difference is category fetcher is expecting it's variable parts to be in the URL path where as other fetchers are passing them in URL query params. Maybe this is fine then, but please add a comment to the effect that you are stripping away the question mark that gets added in NewFetcher code. Also we're assuming endpoints used for category mappings aren't going to contain query params, which maybe is ok since they'll probably just be statically hosted files.

Kalypsonika: Comment added.

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -80,8 +81,45 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
 
-func (fetcher *HttpFetcher) FetchCategories(primaryAdServer, publisherId, iabCategory string) (string, error) {
-	return "", nil
+func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer, publisherId, iabCategory string) (string, error) {
+	if fetcher.Categories == nil {
+		fetcher.Categories = make(map[string]map[string]stored_requests.Category)
+	}
+
+	if publisherId == "" {
+		publisherId = primaryAdServer
```
hhhjort: This means we could have something like `freewheel_freewheel` as the `dataName` below. Is this correct, or should it resolve to just `freewheel`?

Kalypsonika: You're right, fixed!

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -80,8 +81,48 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
+
+	dataName := primaryAdServer
+
+	if publisherId != "" {
+		dataName = fmt.Sprintf("%s_%s", primaryAdServer, publisherId)
+		url = fmt.Sprintf("%s/%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer, publisherId)
+	}
+
+	if data, ok := fetcher.Categories[dataName]; ok {
+		return data[iabCategory].Id, nil
```
mansinahar: Overall the PR LGTM. I just have one small question though: Are we sure that the `data` will always have a value for `iabCategory`? Should we rather just do a small check before we return? Both here and towards the end of the function.

Kalypsonika: @mansinahar yes, it's good to have this check. Thank you!

mansinahar: Thanks @Kalypsonika !

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -80,8 +81,54 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
 
-func (fetcher *HttpFetcher) FetchCategories(primaryAdServer, publisherId, iabCategory string) (string, error) {
-	return "", nil
+func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer, publisherId, iabCategory string) (string, error) {
+	if fetcher.Categories == nil {
+		fetcher.Categories = make(map[string]map[string]stored_requests.Category)
+	}
+
+	//in NewFetcher function there is a code to add "?" at the end of url
+	//in case of categories we don't expect to have any parameters, that's why we need to remove "?"
```
josephveach: A few comments about this code:

1. NewFetcher() is capable of returning "&" as the last character as well as "?" (in which case there will be a "?" somewhere earlier in the string.  I'm guessing from the above comment that this will never happen for categories.  Is that correct?  If so, might it be better (i.e., probably more efficient and more readable) to use strings.TrimSuffix() to simply remove the "?" from the end of the string?  If "?" is not guaranteed to be at the end of the string, then shouldn't everything from the "?" in the line forward be removed?  It might be safest to allow for the possibility of "?" not being at the end of the line (in case of future changes) and use strings.IndexByte() to find the "?" and chop off the end of the line from there; this works regardless of the location of "?".

2. The url variable is assigned twice when publisherId has a value, both times making the same call to strings.Replace().  The data name variable is also assigned a value twice.  It would be a little more efficient to simply declare url and dataName as strings, then use an if-statement to assign each variable once.

3. Taking efficiency considerations to an extreme, if it's likely that fetcher.Categories[dataName] is likely to be set more often than not (I don't know how one would know this), it might be even more efficient to put off setting the value of url until it is needed, in case we never get there.  Drawback to this is it's less efficient (two if-statements needed to check publisherId) when fetcher.Categories[dataName] is unset.

Kalypsonika: Hi Joe, @josephveach thank you for your comment!
1) Please see above discussion with Jason about this. There is a code snipped with logic of adding "?" at the end of URL: https://github.com/prebid/prebid-server/pull/882#discussion_r277105193
We don't expect to have URL with parameters for http fetcher, here is how url builder looks like:
```
url = fmt.Sprintf("%s/%s/%s.json", strings.Replace(fetcher.Endpoint, "?", "", -1), primaryAdServer, publisherId)
``` 
2) Refactored

3) URL builder is just a simple string concatenation. We can separate it and have 2 publisher id checks. I would leave it as it is now. 

Please let me know what do you think.

josephveach: Hi Veronika,  Let me clarify my thoughts regarding (1): 
Looking purely at the NewFetcher() code, it seems that the returned value of Endpoint can be in two forms:
(a) if there is no query string in the original "endpoint" argument to NewFetcher() (i.e., endpoint was of form <someURL>), then Endpoint is simply
<someURL>?
which works fine with the given code, but
(b) if "endpoint" contained a query string, i.e., was of form <someURL>?<someQuery>, then Endpoint would be
<someURL>?<someQuery>&
which I believe does not work properly with the given code.

The comment simply states "we don't expect to have any parameters" in the category case, which does not seem absolute or clear to me (e.g., does "don't expect" mean it's impossible? does "parameters" equate to what would be in the query?).  Furthermore, as I've already indicated, the use of strings.Replace() probably doesn't achieve the desired effect on the second form (b) above of the returned value of Endpoint.  

IMHO, it would be more clear to someone who is not terribly familiar with the underlying code that gets us here if (1) this code used strings.TrimSuffix() instead of strings.Replace(), which would emphasize the '?' occurs only as the last character and, possibly, (2) if the comment were a little more strong, i.e., something like "in case of categories, there is never a query string in the endpoint so it always ends with '?'", which is more certain than "we don't expect".   Actually, looking at previous discussion, I prefer Jason's explanation above: "category fetcher is expecting its variable parts to be in the URL path whereas other fetchers are passing them in URL query params", though maybe "expecting" could be replaced by something stronger in a code comment.   
I also expect strings.TrimSuffix() to run faster than strings.Replace() because it doesn't need to examine the entire string, though I haven't confirmed that.

BTW, I see that NewFetcher() also does an unnecessary second assignment to urlPrefix.  The first portion of the function could be changed to simply:

       var urlPrefix string
       if strings.Contains(endpoint, "?") {
		urlPrefix = endpoint + "&"
	} else {
		urlPrefix = endpoint + "?"
	}

Kalypsonika: Hi Joe, `strings.Replace` is changed to `strings.TrimSuffix`. 
Refactoring existing code in `NewFetcher` is beyond the scope of this PR and will require to refactor `buildRequest` function as well.
Also there are no specific requirements to categories URL, so maybe url builder will be changed in future.

josephveach: My original comment suggested to chop off the end of the line starting at the "?" if there was a chance of the "?" ever not being the last character.  You seem to be saying that could be a possibility in the future, in which case this solution could lead to a problem at that time.  In any case, I've made my suggestion and I have no other objections so I'll give my approval of the PR.

Kalypsonika: Thank you Joe!
With current implementation we just want to delete last "?". If it's not there - url is returned unchanged. Please see this documentation: https://golang.org/pkg/strings/#TrimSuffix

PR #883, opened 2019-04-24
removed filtering requests from ix.size param

ix-prebid-support:
Hello!

The Index Exchange endpoint can now dynamically bid on all slot sizes provided. The non-standard ext.size attribute has been deprecated for server integrations. Please let us know how to ensure this is included in the next prebid-server release as it is approved.

Thank-you kindly
--
IX Prebid Support Team

### DIFF
--- a/adapters/ix/ix.go
+++ b/adapters/ix/ix.go
@@ -37,8 +37,7 @@ func (a *IxAdapter) SkipNoCookies() bool {
 }
 
 type indexParams struct {
-	SiteID string    `json:"siteId"`
-	Size   [2]uint64 `json:"size"`
+	SiteID string `json:"siteId"`
 }
 
 type ixBidResult struct {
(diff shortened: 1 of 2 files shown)


adapters/ix/ix.go
```diff
@@ -100,11 +99,6 @@ func (a *IxAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder *pbs.P
 		}
 
 		for sizeIndex, format := range unit.Sizes {
-			// ensure request are sent only for valid size as specified in the params
-			if ok := isValidIXSize(format, params.Size); !ok {
```
mansinahar: If this is the only place `isValidIXSize` method was used then I'd suggest we also remove that function definition as part of this PR

hhhjort: I can't see it anywhere else.

mansinahar: Cool. Thanks for conforming @hhhjort. @ix-prebid-support, PR looks good. If you can just please remove the `isValidIXSize` function definition, we can go ahead and merge it. Thanks!

ix-prebid-support: Thanks for taking the time to review this PR. We will update the branch so this can be merged.

mansinahar: @ix-prebid-support can I ask what's the approx ETA on updating the branch? We're planning on doing a release today and it would be good to have this in as part of that

PR #886, opened 2019-04-30
PubMatic Adslot validation

PubMatic-OpenWrap:
Make AdSlot as optional parameter and make changes in adslot validation. The valid format for the optional string adslot are :  'adslot@WxH' and 'adslot'

If Width and Height parameters are not present for Banner, assign them to the Banner object from the first size in the array format sizes.

### DIFF
--- a/adapters/pubmatic/pubmatic.go
+++ b/adapters/pubmatic/pubmatic.go
@@ -368,6 +368,51 @@ func (a *PubmaticAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters
 	}}, errs
 }
 
+// validateAdslot validate the adslot object
+// valid formats are 'adslot@WxH', 'adslot' and no adslot
+func validateAdSlot(adslot string, imp *openrtb.Imp) error {
+	if len(adslot) == 0 {
+		return nil
+	}
+
+	if !strings.Contains(adslot, "@") {
+		imp.TagID = strings.TrimSpace(adslot)
+		return nil
+	}
+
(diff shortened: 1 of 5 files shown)


adapters/pubmatic/pubmatic.go
```diff
@@ -368,6 +368,73 @@ func (a *PubmaticAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters
...
+	}
+
+	if !strings.Contains(adSlotStr, "@") {
+		imp.TagID = adSlotStr
+		return nil
+	}
+
+	adSlot := strings.Split(adSlotStr, "@")
+	if len(adSlot) == 2 && adSlot[0] != "" && adSlot[1] != "" {
+		imp.TagID = strings.TrimSpace(adSlot[0])
```
mansinahar: Do we need to call `TrimSpace` after every `strings.Split` call? There shouldn't be any spaces since you already do a `TrimSpace` on the original `adSlot` string itself, right?

PubMatic-OpenWrap: In most cases, the adslot string will be like "test@WxH" and in those cases, TrimSpace is not required more than once but in the rare case where adslot string, for some reason, is something like " test@ 34 x 45 ", TrimSpace called on the entire string will only handle the leading and trailing spaces but when the string is split, the width will be "34 " and height "45 ". For those cases when I have retained the TrimSpace after the spilts.

But there are two places where we can do without TrimSpace and I will remove it there as we handle spaces for individual elements (width and height)

PR #887, opened 2019-04-30
ImproveDigital adapter updates

improve-prebid-dev:
Hi,

Please make code review for our changes:
 - Add support imp.video
 - Add support for mobile/app
 - Add support for multibid responses with known capacity
 - Extend optional parameters
 - Extend nobid/noad cases while server under active development

Thanks


adapters/improvedigital/improvedigital.go
```diff
@@ -78,3 +100,26 @@ func NewImprovedigitalBidder(endpoint string) *ImprovedigitalAdapter {
 		endpoint: endpoint,
 	}
 }
+
+func getMediaTypeForImp(impID string, imps []openrtb.Imp) (openrtb_ext.BidType, error) {
+	for _, imp := range imps {
```
hhhjort: How well does this code work for multiformat imps? For example, if given a multiformat `imp` that includes both banner and video, is your bidder guaranteed to return a banner type bid?

improve-prebid-dev: Our ADServer don`t support multiformat, only banner will be used ignoring video/native/etc. So adapter have the some logic as server to prefer banner by default. Agree, case is good but right now not expected in real life and will make nobid (204 status) with alarm to check publisher site configuration

hhhjort:
One note, since you are tightening up the JSON schema, have you seen any traffic from PBS yet? If a request breaks the schema, the entire request is considered invalid, not just your adapter's piece.

adapters/improvedigital/params_test.go
```diff
@@ -45,9 +46,14 @@ var invalidParams = []string{
 	`[]`,
 	`true`,
 	`2`,
+	`{"size": {"w": 10, "h": 5}}`,
+	`{"other_optional": true}`,
```
josephveach: The fact that the above two entries were formerly valid but are not valid any more is concerning.  My impression is that this is a new adapter that may not yet have had any traffic, in which case this change is okay.  Can you please confirm that is the case?

improve-prebid-dev: We don`t have PBS traffic. Just preparation and want to be strict on start.

"size" without "placementID" or (pkey+pubID) is invalid case, here is pb.js adapter https://github.com/prebid/Prebid.js/blob/master/modules/improvedigitalBidAdapter.js#L21 with the same logic.
Server also need placementID or (pkey+pubID), without them will be nobid/noad

"other_optional" also become invalid in case of bad configuration/typo/undocumented stuff 
Server will ignore "other_optional", but maybe somebody expect other result

PR #889, opened 2019-04-30
Ix remove size param

ix-prebid-support:
Sorry for the delay on this!

### DIFF
--- a/adapters/ix/ix.go
+++ b/adapters/ix/ix.go
@@ -37,8 +37,7 @@ func (a *IxAdapter) SkipNoCookies() bool {
 }
 
 type indexParams struct {
-	SiteID string    `json:"siteId"`
-	Size   [2]uint64 `json:"size"`
+	SiteID string `json:"siteId"`
 }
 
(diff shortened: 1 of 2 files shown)


ix-prebid-support:
@mansinahar let me know if it's too late for this!

hhhjort:
We are releasing the initial PR now ... this bit is just code cleanup, will get it into the next release of PBS. Thanks for the cleanup.

ix-prebid-support:
Sounds good and thank-you so much both @hhhjort @mansinahar

mansinahar:
@ix-prebid-support You'd probably just need to rebase though. Thanks again for the PR!

Aadeshp:
we should also remove the "size" parameter from the IX schema https://github.com/prebid/prebid-server/blob/master/static/bidder-params/ix.json

hhhjort:
Yes, would be good to drop it if size no longer serves a use.

ix-prebid-support:
Appreciate the insights here, will incorporate feedback and resubmit this!

PR #891, opened 2019-05-02
Fixed "invalid BidType: " error for lifestreet adapter

Lifestreet:
On the live traffic, we figured out that a client receives an error from our adapter:
`"errors": {
            "lifestreet": [{
                "code": 999,
                "message": "invalid BidType: "
            }]
        },`

This change should fix the issue


Lifestreet:
It seems gofmt didn't run, closing request and will create a new one

PR #894, opened 2019-05-08
[Currency support] Activate multi-currencies support

benjaminch:
This CL is the last piece to activate currency conversion support in
PBS. It activates currency support per default.

Currency rates will be fetched once per hour (PrebidJS file is updated
once a day).

Issue: #280


benjaminch:
Thanks ! :)

PR #897, opened 2019-05-13

guscarreon:
PBS internally determines if the impression from a sender includes the publisher_id RTB parameter, and if publisher_id is present in an ENV/config variable set by the PBS owner, skip the GDPR check internally but leave the gdpr=1/0 flag intact.


gdpr/impl.go
```diff
@@ -41,6 +41,12 @@ func (p *permissionsImpl) BidderSyncAllowed(ctx context.Context, bidder openrtb_
 func (p *permissionsImpl) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, consent string) (bool, error) {
 	id, ok := p.vendorIDs[bidder]
 	if ok {
+		// Linear seach the bidderName in the GDPR TrustVendors array. These trusted vendors come from the configuration .yaml file
```
hhhjort: We want to trust all bidders for certain publisher IDs. Not certain bidders always.

guscarreon:
Created PR #906 instead

PR #906, opened 2019-05-17

guscarreon:
**This pull request substitutes #897**
PBS internally determines if the impression from a sender includes the publisher_id RTB parameter, and if publisher_id is present in an ENV/config variable set by the PBS owner, skip the GDPR check internally but leave the gdpr=1/0 flag intact. This ticket is intended to substitute PBS-372 since it was rejected in favor of this approach.

### DIFF
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -45,7 +45,16 @@ func cleanOpenRTBRequests(ctx context.Context,
 		for bidder, bidReq := range requestsByBidder {
 			// Fixes #820
 			coreBidder := resolveBidder(bidder.String(), aliases)
-			if ok, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, consent); !ok && err == nil {
+
+			var publisher_id string
+			if bidReq.Site != nil && bidReq.Site.ID != "" {
+				publisher_id = bidReq.Site.ID
+			} else if bidReq.App != nil {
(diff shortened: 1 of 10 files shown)


exchange/utils.go
```diff
@@ -45,7 +45,16 @@ func cleanOpenRTBRequests(ctx context.Context,
 		for bidder, bidReq := range requestsByBidder {
 			// Fixes #820
 			coreBidder := resolveBidder(bidder.String(), aliases)
-			if ok, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, consent); !ok && err == nil {
+
+			var publisher_id string
+			if bidReq.Site != nil && bidReq.Site.ID != "" {
```
hhhjort: It is `bidReq.Site.Publisher.Id`, we are identifying the publisher of the site, not the site itself.

exchange/utils.go
```diff
@@ -45,7 +45,16 @@ func cleanOpenRTBRequests(ctx context.Context,
...
+				publisher_id = bidReq.Site.ID
+			} else if bidReq.App != nil {
+				publisher_id = bidReq.App.ID
```
hhhjort: Likewise It is `bidReq.App.Publisher.Id`, we are identifying the publisher of the app.

Issue #895, opened 2019-05-10
Go 1.9.1 compatibility broken


ffeast:
Hi

From the https://github.com/prebid/prebid-server#installation guide:
`First install Go 1.9.1 or later and dep`

However, it doesn't seem to build with 1.9.1:

```
./validate.sh 
# github.com/prebid/prebid-server/macros
macros/macros.go:22:16: undefined: strings.Builder
# github.com/prebid/prebid-server/macros
macros/macros.go:22:16: undefined: strings.Builder
FAIL    github.com/prebid/prebid-server [build failed]
FAIL    github.com/prebid/prebid-server/adapters [build failed]
FAIL    github.com/prebid/prebid-server/adapters/33across [build failed]

```

It seems to fail because of `strings.Builder` used in `macros.go`
The suggestion is to either update the docs or switch to `bytes.Buffer` if 1.9.1 still needs to be supported

benjaminch:
Hello there ! 
I would be happy to work on this one ! :)

PR #910, opened 2019-05-20
[fix] broken Go 1.9 compatibility

benjaminch:
This CL addresses Go 1.9 compatibility issue.

Issue: #895


mansinahar:
Thank you so much @benjaminch for the PR! LGTM. Just one additional thing though, If we want to support Go 1.9 then we should also add that version in `travis.yml` so that these kind of issues are caught early.

benjaminch:
Thanks ! Will add the version to the Travis ;)

benjaminch:
just updated @mansinahar thanks :)

hhhjort:
Do we still have people using go 1.9 to compile PBS? It is two versions back, but if there is a use case for it I am ok with it.

benjaminch:
Hey @hhhjort , according to #895 the documentation was stating 1.9, so I guess some people are still using it.
Maybe we can merge this one, update the doc with the latest version to use and then remove 1.9 from the CI?

benjaminch:
And here's the ticket for following steps: #920 
:)

PR #912, opened 2019-05-21
feat(adapters): submit unruly adapter

lewismoore10100:
Submitting a new adapter from Unruly


lewismoore10100:
User syncing is not implemented, will address.

PR #914, opened 2019-05-21
Striving for readability and simplification, function doCache has been modified with no extra expense of time nor space complexity.

guscarreon:
### Issue description:
The `doCache` function found in `exchange/auction.go` must traverse this hash table by `ImpID` and not by `bidID` as it currently does.

### Implemented changes
Striving for readability and simplification, function `doCache` has been modified with no extra expense of time nor space complexity. Names of two fields of the auction struct where also changed for the sake of readability

### DIFF
--- a/exchange/auction.go
+++ b/exchange/auction.go
@@ -79,15 +79,13 @@ func (a *auction) doCache(
 	}
 
 	var errs []error
-	var expectNumBids int
-	var expectNumVast int
-	var bidIndices map[int]*openrtb.Bid
-	var vastIndices map[int]*openrtb.Bid
+	var cacheErr error
+	var expectNumBids, expectNumVast, newlyCached int
+	var bidIndices, vastIndices map[int]*openrtb.Bid
 	var toCache []prebid_cache_client.Cacheable
 	var expByImp map[string]int64
 	var hbCacheID string
-	var competitiveExclusion bool
-	var bids_to_cache_arr []*pbsOrtbBid
+	var competitiveExclusion, isNonVast, isVast bool
 
 	expectNumBids = valOrZero(targData.includeCacheBids, len(a.roundedPrices))
 	expectNumVast = valOrZero(targData.includeCacheVast, len(a.roundedPrices))
@@ -97,7 +95,7 @@ func (a *auction) doCache(
 	expByImp = make(map[string]int64)
 	competitiveExclusion = false
 
-	if len(bidCategory) > 0 {
+	if len(bidCategory) > 0 && targData.includeCacheVast {
 		// assert:  category of winning bids never duplicated
 		if rawUuid, err := uuid.NewV4(); err == nil {
 			hbCacheID = rawUuid.String()
@@ -112,55 +110,38 @@ func (a *auction) doCache(
 		expByImp[imp.ID] = imp.Exp
 	}
 
-	/* bidders vs winners			*/
+	// if targData.includeBidderKeys is true, we should cache all bids, both winning and losing. In other words, we'll cache
+	// banners and/or videos found in impsToBiddersTopBids map[string]map[openrtb_ext.BidderName]*pbsOrtbBid
 	if targData.includeBidderKeys {
-		// cache banners and/or videos found in impsToBiddersTopBids map[string]map[openrtb_ext.BidderName]*pbsOrtbBid
 		for impID := range a.impsToBiddersTopBids {
-			for _, bid_to_cache := range a.impsToBiddersTopBids[impID] {
-				bids_to_cache_arr = append(bids_to_cache_arr, bid_to_cache)
-			}
-		}
-	} else if targData.includeWinners {
-		// cache banners and/or videos found in impsToTopBids map[string]*pbsOrtbBid
-		for _, bid_to_cache := range a.impsToTopBids {
-			bids_to_cache_arr = append(bids_to_cache_arr, bid_to_cache)
-		}
-	}
-
-	for _, bid_to_cache := range bids_to_cache_arr {
-		/* banners vs videos 			*/
-		if targData.includeCacheBids { //banner
-			if jsonBytes, err := json.Marshal(bid_to_cache.bid); err == nil {
-				toCache = append(toCache, prebid_cache_client.Cacheable{
-					Type:       prebid_cache_client.TypeJSON,
-					Data:       jsonBytes,
-					TTLSeconds: cacheTTL(expByImp[bid_to_cache.bid.ImpID], bid_to_cache.bid.Exp, defTTL(bid_to_cache.bidType, defaultTTLs), ttlBuffer),
-				})
-				bidIndices[len(toCache)-1] = bid_to_cache.bid
-			} else {
-				errs = append(errs, err)
-			}
-		}
-		if targData.includeCacheVast && bid_to_cache.bidType == openrtb_ext.BidTypeVideo { //video
-			if jsonBytes, err := json.Marshal(makeVAST(bid_to_cache.bid)); err == nil {
-				catDur, ok := bidCategory[a.impsToTopBids[bid_to_cache.bid.ImpID].bid.ID]
-				if ok && (a.impsToTopBids[bid_to_cache.bid.ImpID] == bid_to_cache) {
-					toCache = append(toCache, prebid_cache_client.Cacheable{
-						Type:       prebid_cache_client.TypeXML,
-						Data:       jsonBytes,
-						TTLSeconds: cacheTTL(expByImp[bid_to_cache.bid.ImpID], bid_to_cache.bid.Exp, defTTL(bid_to_cache.bidType, defaultTTLs), ttlBuffer),
-						Key:        fmt.Sprintf("%s_%s", catDur, hbCacheID),
-					})
-				} else {
-					toCache = append(toCache, prebid_cache_client.Cacheable{
-						Type:       prebid_cache_client.TypeXML,
-						Data:       jsonBytes,
-						TTLSeconds: cacheTTL(expByImp[bid_to_cache.bid.ImpID], bid_to_cache.bid.Exp, defTTL(bid_to_cache.bidType, defaultTTLs), ttlBuffer),
-					})
+			for _, bidToCache := range a.impsToBiddersTopBids[impID] {
+				isNonVast, isVast, newlyCached, cacheErr = a.cacheBid(bidToCache, targData.includeCacheBids, targData.includeCacheVast, competitiveExclusion, &toCache, expByImp, defaultTTLs, ttlBuffer, bidCategory, hbCacheID)
+				if isNonVast {
+					bidIndices[len(toCache)-newlyCached] = bidToCache.bid
+					newlyCached--
 				}
-				vastIndices[len(toCache)-1] = bid_to_cache.bid
-			} else {
-				errs = append(errs, err)
+				if isVast {
+					vastIndices[len(toCache)-newlyCached] = bidToCache.bid
+				}
+				if cacheErr != nil {
+					errs = append(errs, cacheErr)
+				}
+			}
+		}
+	} else {
+		// targData.includeBidderKeys is false, therefore, targData.includeWinners is true and we should cache only winning bids
+		// which are found in a.impsToTopBids
+		for _, bidToCache := range a.impsToTopBids {
+			isNonVast, isVast, newlyCached, cacheErr = a.cacheBid(bidToCache, targData.includeCacheBids, targData.includeCacheVast, competitiveExclusion, &toCache, expByImp, defaultTTLs, ttlBuffer, bidCategory, hbCacheID)
+			if isNonVast {
+				bidIndices[len(toCache)-newlyCached] = bidToCache.bid
+				newlyCached--
+			}
+			if isVast {
+				vastIndices[len(toCache)-newlyCached] = bidToCache.bid
+			}
+			if cacheErr != nil {
+				errs = append(errs, cacheErr)
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


exchange/auction.go
```diff
@@ -88,61 +111,56 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
-			useCustomCacheKey := false
-			if competitiveExclusion && isOverallWinner {
-				// set custom cache key for winning bid when competitive exclusion applies
-				catDur = bidCategory[topBidPerBidder.bid.ID]
-				if len(catDur) > 0 {
-					customCacheKey = fmt.Sprintf("%s_%s", catDur, hbCacheID)
-					useCustomCacheKey = true
-				}
+		}
+	} else if targData.includeWinners {
```
hhhjort: What if `targData.includeWinners` and `targData.incudeBidderKeys` are both true?

guscarreon: When `targData.incudeBidderKeys` is true, we will also include winners in the `bids_to_cahce_arr` therefore, even if we don't go into the true branch of the `else if` statement in line 123, those bids will be copied into `bids_to_cahce_arr` because `impsToBiddersTopBids`, which was traversed in line 118, includes both regular bids and top bids.

exchange/auction.go
```diff
@@ -60,20 +64,39 @@ func (a *auction) setRoundedPrices(priceGranularity openrtb_ext.PriceGranularity
...
-	competitiveExclusion := false
+	var expectNumBids int
+	var expectNumVast int
+	var bidIndices map[int]*openrtb.Bid
+	var vastIndices map[int]*openrtb.Bid
+	var toCache []prebid_cache_client.Cacheable
+	var expByImp map[string]int64
 	var hbCacheID string
+	var competitiveExclusion bool
+	var bids_to_cache_arr []*pbsOrtbBid
```
josephveach: Variable name police comment:  bids_to_cache_arr should be bidsToCacheArr.

exchange/auction.go
josephveach: The above line can simply be
} else {
because we know targData.includeWinners is true if we get here.

I think Hans' comment above indicates that the comments in the code could be improved.  I believe the point is that when includeBidderKeys is set, we need to cache all bids but, otherwise, we need only cache the winners.  I suggest, in the interests of readability, changing the comment at line 117 (and maybe 124) to reflect that and eliminating the comment at line 115, or eliminating 117 and 124 entirely and replacing with a very descriptive 115.

exchange/auction.go
```diff
@@ -88,61 +111,56 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
+		// cache banners and/or videos found in impsToTopBids map[string]*pbsOrtbBid
+		for _, bid_to_cache := range a.impsToTopBids {
+			bids_to_cache_arr = append(bids_to_cache_arr, bid_to_cache)
+		}
+	}
+
+	for _, bid_to_cache := range bids_to_cache_arr {
+		/* banners vs videos 			*/
+		if targData.includeCacheBids { //banner
+			if jsonBytes, err := json.Marshal(bid_to_cache.bid); err == nil {
```
josephveach: The old code had a check here to detect an error if trying to use custom cache key for a non-vast bid.  Is there a reason the check is gone?

exchange/auction.go
```diff
@@ -88,61 +111,56 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
-			if bids {
-				if jsonBytes, err := json.Marshal(topBidPerBidder.bid); err == nil {
-					if useCustomCacheKey {
-						// not allowed if bids is true; log error and cache normally
-						errs = append(errs, errors.New("cannot use custom cache key for non-vast bids"))
-					}
+		}
+		if targData.includeCacheVast && bid_to_cache.bidType == openrtb_ext.BidTypeVideo { //video
+			if jsonBytes, err := json.Marshal(makeVAST(bid_to_cache.bid)); err == nil {
+				catDur, ok := bidCategory[a.impsToTopBids[bid_to_cache.bid.ImpID].bid.ID]
```
josephveach: It's probably more efficient to assign catDur only once it's known to be needed.  So, probably better to check if it's a winning bid first, then assign catDur.

exchange/auction.go
```diff
@@ -88,61 +111,56 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
+				if ok && (a.impsToTopBids[bid_to_cache.bid.ImpID] == bid_to_cache) {
 					toCache = append(toCache, prebid_cache_client.Cacheable{
-						Type:       prebid_cache_client.TypeJSON,
+						Type:       prebid_cache_client.TypeXML,
 						Data:       jsonBytes,
-						TTLSeconds: cacheTTL(expByImp[impID], topBidPerBidder.bid.Exp, defTTL(topBidPerBidder.bidType, defaultTTLs), ttlBuffer),
+						TTLSeconds: cacheTTL(expByImp[bid_to_cache.bid.ImpID], bid_to_cache.bid.Exp, defTTL(bid_to_cache.bidType, defaultTTLs), ttlBuffer),
+						Key:        fmt.Sprintf("%s_%s", catDur, hbCacheID),
```
josephveach: A problem can occur here if hbCacheID is nil.  Old code caught this via useCustomCacheKey variable.

exchange/auction.go
```diff
@@ -88,61 +111,56 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
+		}
+	} else if targData.includeWinners {
+		// cache banners and/or videos found in impsToTopBids map[string]*pbsOrtbBid
+		for _, bid_to_cache := range a.impsToTopBids {
+			bids_to_cache_arr = append(bids_to_cache_arr, bid_to_cache)
+		}
+	}
+
+	for _, bid_to_cache := range bids_to_cache_arr {
+		/* banners vs videos 			*/
```
josephveach: Recognizing that when includeBidderKeys is false, it's possible to avoid looking at non-winners is a good catch.  To refine that even further, I suggest that this entire loop (lines 130 to 166) should be replaced with a function that is called from lines 120 and 126.  I believe this would improve readability and also eliminate the need for and cost of building and iterating through the bids_to_cache_arr array.  Another possible benefit would be one argument to the function could be a flag that indicates whether it was called from line 120 or line 126 because, in the latter case, we already know it's a winning bid without needing to check as at line 147.

exchange/auction.go
```diff
@@ -16,39 +16,43 @@ import (
 )
 
 func newAuction(seatBids map[openrtb_ext.BidderName]*pbsOrtbSeatBid, numImps int) *auction {
-	winningBids := make(map[string]*pbsOrtbBid, numImps)
-	winningBidsByBidder := make(map[string]map[openrtb_ext.BidderName]*pbsOrtbBid, numImps)
+	impsToTopBids := make(map[string]*pbsOrtbBid, numImps)
+	impsToBiddersTopBids := make(map[string]map[openrtb_ext.BidderName]*pbsOrtbBid, numImps)
```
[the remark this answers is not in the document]
guscarreon: I find it useful to implicitly have a hint of what variable types are being mapped in a hash table. But I can roll back the names if you want. Comments were added.

exchange/auction.go
guscarreon: Done

exchange/auction.go
guscarreon: Done

exchange/auction.go
guscarreon: Good catch! Error check included again

exchange/auction.go
guscarreon: Agree, done.

exchange/auction.go
guscarreon: I believe strings can, at worst, be blank ("") and not NULL, useCustomCacheKey's functionality ahs been included here in the form of lines 201 and 202
```
199     if incVastBids && bidToCache.bidType == openrtb_ext.BidTypeVideo { //video
200         if jsonBytes, err := json.Marshal(makeVAST(bidToCache.bid)); err == nil {
201             _, isTopBid := a.impsToTopBids[bidToCache.bid.ImpID]
202             if catDur, ok := bidCategory[a.impsToTopBids[bidToCache.bid.ImpID].bid.ID]; ok && isTopBid {
203                 *toCache = append(*toCache, prebid_cache_client.Cacheable{
exchange/auction.go
```

exchange/auction.go
guscarreon: Originally, before putting up this PR, I did think about that possibility but said function started to become a bit bloated and I dropped it. Now that you suggest it, went back to that approach and finally added it. I ran into trouble with a couple of test cases but now it is passing them all again.

exchange/auction.go
```diff
@@ -176,6 +175,52 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
+			})
+			chachedBid = true
+		} else {
+			anError = err
+		}
+
+	}
+	if incVastBids && bidToCache.bidType == openrtb_ext.BidTypeVideo { //video
+		if jsonBytes, err := json.Marshal(makeVAST(bidToCache.bid)); err == nil {
+			_, isTopBid := a.impsToTopBids[bidToCache.bid.ImpID]
```
josephveach: If this function includes an argument that indicates the bid is already known to be a winner (which is true when iterating through impsToTopBids), then it's not necessary to make the above check when that argument is set to true.  In the long run, that should be a performance win.

exchange/auction.go
```diff
@@ -176,6 +175,52 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
+			if catDur, ok := bidCategory[a.impsToTopBids[bidToCache.bid.ImpID].bid.ID]; ok && isTopBid {
+				*toCache = append(*toCache, prebid_cache_client.Cacheable{
+					Type:       prebid_cache_client.TypeXML,
+					Data:       jsonBytes,
+					TTLSeconds: cacheTTL(expByImp[bidToCache.bid.ImpID], bidToCache.bid.Exp, defTTL(bidToCache.bidType, defaultTTLs), ttlBuffer),
+					Key:        fmt.Sprintf("%s_%s", catDur, hbCacheID),
```
josephveach: It is possible for hbCacheID to not be set, in which case a problem occurs here.  This possibility needs to be avoided.

exchange/auction.go
```diff
@@ -244,10 +289,10 @@ func defTTL(bidType openrtb_ext.BidType, defaultTTLs *config.DefaultTTLs) (ttl i
 }
 
 type auction struct {
-	// winningBids is a map from imp.id to the highest overall CPM bid in that imp.
-	winningBids map[string]*pbsOrtbBid
-	// winningBidsByBidder stores the highest bid on each imp by each bidder.
-	winningBidsByBidder map[string]map[openrtb_ext.BidderName]*pbsOrtbBid
+	// We'll store the hightest bid comming from each Imp in the request
```
josephveach: Please correct the spelling errors (hightest, comming) and add that these maps map **imp IDs** to what they're mapped to (highest bid for the specified imp in first case, a map of the highest bid from each bidder for the specified imp in the second case).  Bids don't _come from_ the imps; they are _for_ the imps.

exchange/auction.go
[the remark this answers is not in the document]
josephveach: Let's let other reviewers weigh in here.

exchange/auction.go
```diff
@@ -88,79 +109,57 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
-					customCacheKey = fmt.Sprintf("%s_%s", catDur, hbCacheID)
-					useCustomCacheKey = true
+
+	// if targData.includeBidderKeys is true, we should cache all bids, both winning and losing. In other words, we'll cache
+	// banners and/or videos found in impsToBiddersTopBids map[string]map[openrtb_ext.BidderName]*pbsOrtbBid
+	if targData.includeBidderKeys {
+		for impID := range a.impsToBiddersTopBids {
+			for _, bidToCache := range a.impsToBiddersTopBids[impID] {
+				isNonVast, isVast, newlyCached, cacheErr = a.cacheBid(bidToCache, targData.includeCacheBids, targData.includeCacheVast, competitiveExclusion, &toCache, expByImp, defaultTTLs, ttlBuffer, bidCategory, hbCacheID)
+				if isNonVast {
```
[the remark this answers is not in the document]
guscarreon: I think passing 3 extra parameters to `cacheBid` overcomplicates it and makes its signature less readable, doesn't it?

exchange/auction.go
guscarreon: Good catch. I'm still including the check whenever the new flag is false because we'll sometimes call this function from the for statement in line 117 that traverses `a.impsToBiddersTopBids` making this check to be needed anyways

exchange/auction.go
```diff
@@ -176,6 +175,52 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
...
+			chachedBid = true
+		} else {
+			anError = err
+		}
+
+	}
+	if incVastBids && bidToCache.bidType == openrtb_ext.BidTypeVideo { //video
+		if jsonBytes, err := json.Marshal(makeVAST(bidToCache.bid)); err == nil {
+			_, isTopBid := a.impsToTopBids[bidToCache.bid.ImpID]
+			if catDur, ok := bidCategory[a.impsToTopBids[bidToCache.bid.ImpID].bid.ID]; ok && isTopBid {
```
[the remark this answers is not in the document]
guscarreon: Yep, before updating the PR I tried inverting the order of the operands in the `&&` operator but didn't know how to so I finally refactored to two consecutive if statements, which I don't like. But I guess golang has its ways.

exchange/auction.go
guscarreon: No it isn't, the current logic prevents it

mkendall07:
Think we are spending too much time on this PR in proportion to its value. Going to close. If you have a good justification for re-opening please let me know.

PR #915, opened 2019-05-21
Unruly adaptor submit

lewismoore10100:
Add Unruly Adaptor with User Sync end points


lewmoore:
Hi @josephveach 

I've just committed some changes to the PR, let me know if there is anything else we need to add.

Thanks,

josephveach:
My apologies:  I forgot to also mention that there should be other tests in adapters/unruly/unrulytest/ using the tool described here:  https://github.com/prebid/prebid-server/blob/master/adapters/adapterstest/test_json.go#L50 and referenced here:  https://github.com/prebid/prebid-server/blob/master/docs/developers/automated-tests.md.  Please see pretty much any other adapter's code in the corresponding directory for examples.

lewmoore:
Thanks @josephveach taking a look at this now.

Also, noticed in some other adapters and in the RunJSONBidderTest function that there are references to /exemplary and /supplemental directories. Do we need to add these as well with example json files?

josephveach:
Hi @lewmoore, Yes, you may have already found that RunJSONBidderTest processes the son files in /exemplary and /supplemental directories.

adapters/unruly/unruly.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	return false
+}
+
+func (a *UnrulyAdapter) BuildRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
+	var errs []error
+	reqJSON, err := json.Marshal(request)
+	if err != nil {
+		errs = append(errs, err)
+		return nil, errs
```
hhhjort: You could `return nil, []error{err}` instead of declaring an empty errs, and then appending the error to it. Then below return `nil` instead of `errs` since `errs` would have to be empty if we got to that point.

adapters/unruly/unruly.go
```diff
@@ -0,0 +1,180 @@
...
+			errs = append(errs, errors...)
+		}
+	}
+	return adapterRequests, errs
+}
+
+func getMediaTypeForImpWithId(impID string, imps []openrtb.Imp) (openrtb_ext.BidType, error) {
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner != nil {
```
hhhjort: For multiformat requests, are you guaranteed to always return a banner?

adapters/unruly/unruly.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+
+func (a *UnrulyAdapter) ReplaceImp(imp openrtb.Imp, request *openrtb.BidRequest) *openrtb.BidRequest {
+	reqCopy := *request
+	reqCopy.Imp = append(make([]openrtb.Imp, 0, 1), imp)
+	return &reqCopy
+}
+
+func (a *UnrulyAdapter) CheckImpExtension(request *openrtb.BidRequest) bool {
```
hhhjort: is `CheckImpExtension()` used anywhere besides the test to insure it is working?

adapters/unruly/unruly.go
hhhjort: Actually, from your bidder info below, it appears you only support video?

adapters/unruly/unruly_test.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+func TestJsonSamples(t *testing.T) {
+	adapterConfig := new(adapters.HTTPAdapterConfig)
+	adapterstest.RunJSONBidderTest(t, "unrulytest", NewUnrulyAdapter(adapterConfig, "http://targeting.unrulymedia.com/openrtb/2.2"))
```
hhhjort: This will run tests defined in JSON that can test a wide range of issues. See all the JSON files in other adapters in `<adapterName>test` directories of the adapters. You may be able to convert many of your tests below into JSON format, and perhaps make them even more comprehensive. The JSON also helps other people understand what is and is not supposed to work as a request for your adapter.

lewmoore:
Thanks @hhhjort , addressed the points in the above commit. Let me know if theres anything else.

hhhjort:
Some merge conflicts have developed, but it otherwise looks good.

lewmoore:
Thanks @hhhjort , regarding the merge conflicts, I can't see the diff as it's saying that it can only be resolved by those with write access to the repo?

hhhjort:
What you want to do is pull the current prebid:master branch down into your branch and merge it. Then commit those changes on your branch and the merge conflicts will be resolved. You might have to first update the master branch of your clone to the prebid master branch, and then merge that local master branch in, not sure if you can do it in one step.

lewmoore:
Hi @hhhjort we have merged the updated master changes with our branch and pushed the changes above.

PR #916, opened 2019-05-21

guscarreon:
PBS internally determines if the impression from a sender includes the `publisher_id` RTB parameter, and if `publisher_id` is present in an ENV/config variable set by the PBS owner, skip the GDPR check internally but leave the gdpr=1/0 flag intact. This ticket is intended to substitute **PBS-372** since it was rejected in favor of this approach.

### DIFF
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -45,7 +45,16 @@ func cleanOpenRTBRequests(ctx context.Context,
 		for bidder, bidReq := range requestsByBidder {
 			// Fixes #820
 			coreBidder := resolveBidder(bidder.String(), aliases)
-			if ok, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, consent); !ok && err == nil {
+
+			var publisher_id string
+			if bidReq.Site != nil && bidReq.Site.Publisher != nil && bidReq.Site.Publisher.ID != "" {
+				publisher_id = bidReq.Site.Publisher.ID
+			} else if bidReq.App != nil && bidReq.App.Publisher != nil {
+				publisher_id = bidReq.App.Publisher.ID
+			} else {
+				publisher_id = ""
+			}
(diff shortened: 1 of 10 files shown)


exchange/utils.go
```diff
@@ -45,7 +45,16 @@ func cleanOpenRTBRequests(ctx context.Context,
 		for bidder, bidReq := range requestsByBidder {
 			// Fixes #820
 			coreBidder := resolveBidder(bidder.String(), aliases)
-			if ok, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, consent); !ok && err == nil {
+
+			var publisher_id string
```
mansinahar: Can this var name please be changed to `publisherID` based on Go's naming convention?

gdpr/impl_test.go
```diff
@@ -175,11 +175,21 @@ func TestAllowPersonalInfo(t *testing.T) {
...
 	assertNilErr(t, err)
 	assertBoolsEqual(t, false, allowPI)
 
-	allowPI, err = perms.PersonalInfoAllowed(context.Background(), openrtb_ext.BidderPubmatic, "BOS2bx5OS2bx5ABABBAAABoAAAABBwAA")
+	allowPI, err = perms.PersonalInfoAllowed(context.Background(), openrtb_ext.BidderPubmatic, "", "BOS2bx5OS2bx5ABABBAAABoAAAABBwAA")
+	assertNilErr(t, err)
+	assertBoolsEqual(t, true, allowPI)
+
+	// Assert that an item that otherwise would not be allowed PI access, gets approved because it is found in the GDPR.NonStandardPublishers array
+	perms.cfg.NonStandardPublishers = []string{"siteID", "fake-site-id", "appNexusAppID"}
```
mansinahar: I'd recommend we add only `appNexusAppID` to the `NonStandardPublisherMap` directly since only that value is tested in this test.

guscarreon: You mean the `perms.cfg.NonStandardPublishers` array?

PR #917, opened 2019-05-22
PBS-409: add flag for AMP requests

josephveach:
This looks more complicated than it really is.  There are 42 changed files and two new files.  At most simple level, this PR:
(1) introduces a new adapters struct used to pass info about PBS entry point to bidder adapters,
(2) modifies the appnexus adapter to pass is_amp=1 to bidder in bid Ext when new adapters struct indicates request came through AMP endpoint,
(3) adds a few tests for the new functionality, and
(4) updates about function calls in about 30 files to accept but ignore a new argument.

Specific changes:
- added new ExtraRequestInfo struct to adapters/info.go
- updated _getAllBids_() in exchange/exchange.go to set adapters.ExtraRequestInfo appropriately based on info in metrics structure and pass a pointer to it to requestBid()
- updated _requestBid_() in exchange/bidder.go to accept new adapters.ExtraRequestInfo argument and pass it to _MakeRequests_() in each adapter
- updated appnexus adapter (adapters/appnexus/appnexus.go) _MakeRequests_()
  - to have new request bidder Ext value IsAmp / is_amp
  - to propagate IsAmp / is_amp to bidder with value 1 for AMP requests
  - changed existing request bidder Ext values to be _omitempty_ to minimize size of Ext info sent with AMP requests
- updated 25 other adapters' _MakeRequests_() simply to accept new adapters.ExtraRequestInfo argument (though they don't use it)
- updated 5 other existing adapter tests with _MakeRequests_() calls to accept additional argument
- updated test_json.go to run tests in new _adapters/<bidder>/<biddertest>/amp_ directory with adapters.ExtraRequestInfo set (this can be updated in the future to be more generic as we add additional values) and added 2 test files for appnexus adapter


adapters/rubicon/rubicon_test.go
```diff
@@ -1046,7 +1046,8 @@ func TestOpenRTBRequest(t *testing.T) {
 		},
 	}
 
-	reqs, errs := bidder.MakeRequests(request)
+	reqInfo := adapters.ExtraRequestInfo{}
+	reqs, errs := bidder.MakeRequests(request, reqInfo)
```
guscarreon: Again, if for some reason, a `RequestType` field can't be added to the `openrtb.BidRequest` struct itself, can we save us the hassle of having an extra local variable here and get rid of the `reqInfo` declaration? If so, line 1049 would be gone and we could just have:
```
1050        reqs,errs := bidder.MakeRequests(request, adapters.ExtraRequestInfo{})
```

exchange/bidder_test.go
```diff
@@ -146,7 +147,8 @@ func TestMultiBidder(t *testing.T) {
 	}
 	bidder := adaptBidder(bidderImpl, server.Client())
 	currencyConverter := currencies.NewRateConverterDefault()
-	seatBid, errs := bidder.requestBid(context.Background(), &openrtb.BidRequest{}, "test", 1.0, currencyConverter.Rates())
+	reqInfo := adapters.ExtraRequestInfo{}
+	seatBid, errs := bidder.requestBid(context.Background(), &openrtb.BidRequest{}, "test", 1.0, currencyConverter.Rates(), reqInfo)
```
guscarreon: Same here. If we really need the `ExtraRequestInfo` structure to exist we can get rid of the instantiation of `reqInfo` in line 150 and just pass `adapters.ExtraRquestInfo{}` as the last parameter just as they passed  `&openrtb.BidRequest{}` as second parameter

adapters/info.go
```diff
@@ -217,3 +218,7 @@ func parseBidderInfo(info BidderInfo) parsedBidderInfo {
 	}
 	return parsedInfo
 }
+
+type ExtraRequestInfo struct {
+	PbsEntryPoint pbsmetrics.RequestType
+}
```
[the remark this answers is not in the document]
josephveach: Hi Gus,  The OpenRTB.BidRequest structure is an industry standard independent of PBS that we are not free to change.  While it would ease this particular task if we could update the structure, it wouldn't make sense to have PBS-specific fields in that struct.  See https://www.iab.com/wp-content/uploads/2015/05/OpenRTB_API_Specification_Version_2_3_1.pdf.

adapters/info_test.go
```diff
@@ -21,9 +21,10 @@ func TestAppNotSupported(t *testing.T) {
 		},
 	}
 	constrained := adapters.EnforceBidderInfo(bidder, info)
+	reqInfo := adapters.ExtraRequestInfo{}
 	bids, errs := constrained.MakeRequests(&openrtb.BidRequest{
 		App: &openrtb.App{},
-	})
+	}, reqInfo)
```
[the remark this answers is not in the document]
josephveach: I agree this makes at least some of the code a little more clear.  I updated the six test files that could have been modified per your suggestion.

josephveach:
Hi Gus,  Thanks for valiantly trying to simplify these changes.  I have a few objections to your proposed solution as it currently stands, including:
(1) It would still be necessary to update every structure that implements MakeRequests().  That means all the same adapter/_bidder_/_bidder.go_ files would need to be changed as in the current solution.
(2) I view the ExtraRequestInfo structure as a positive rather than a negative.  Strictly speaking, to satisfy the ticket, it would have sufficed to add a boolean variable indicating whether the request originated from AMP.   The PbsEntryPoint variable instead provides more info immediately, allowing detection of other entry points.  Having PbsEntryPoint in a structure allows for possible additional information to be added to the structure in the future without needing to change the interface everywhere again.

adapters/info.go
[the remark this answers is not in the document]
hhhjort: Also the `openrtb` package is a third party package, so we would need to fork it, and then track any updates in order to make any changes to it.

hhhjort:
I was just pondering if `adapters.ExtraRequestInfo` should be passed as a pointer rather than as struct. It doesn't make much matter now, but if it expands we will need to copy the struct on every adapter call. Passing a pointer can help save on work.

josephveach:
I agree, it probably makes more sense to pass a pointer rather than a struct.

adapters/appnexus/appnexus.go
```diff
@@ -345,6 +354,16 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters
...
+	} else if isAMP != 0 {
+		// if isAMP is 0, bidder doesn't need it included
+		var err error
+		reqExt.Appnexus = &appnexusReqExtAppnexus{}
+		reqExt.Appnexus.IsAMP = isAMP
+		request.Ext, err = json.Marshal(reqExt)
+		if err != nil {
+			errs = append(errs, err)
+			return nil, errs
+		}
```
guscarreon: If the entire point of `isAmp` is to serve the purpose of a boolean value, can't we do so? We can declare it in the struct to not be a pointer so its default value is `false` and not `nil`. That is, unless there is an additional purpose `isAMP` serves that makes it absolutely neccessary for us to use an `int`.
```
   94
   95 type appnexusReqExtAppnexus struct {
   96     IncludeBrandCategory    *bool `json:"include_brand_category,omitempty"`
   97     BrandCategoryUniqueness *bool `json:"brand_category_uniqueness,omitempty"`
   98     IsAMP                   bool `json:"brand_category_uniqueness"`  //<-- declare it as bool so it defaults to false and not nil
   99 }
  100
adapters/appnexus/appnexus.go [+][RO]
```

hhhjort: This is driven by the requirement from impbus, it is not a decision Joe made.

PR #919, opened 2019-05-28
Using Site publisher or app publisher as account in openrtb

guscarreon:
Currently we only track account for legacy requests. We are seeing issues now where it would be good to track accounts in `openrtb`. Unfortunately `openrtb` doesn't have this as a required field, but we can grab `request.site.publisher.id` or `request.app.publisher.id` as the account. We can use "UNKNOWN" for when this field is not populated.

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -89,7 +89,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 
 	start := time.Now()
 	labels := pbsmetrics.Labels{
-		Source:        pbsmetrics.DemandUnknown,
+		Source:        pbsmetrics.DemandWeb,
 		RType:         pbsmetrics.ReqTypeAMP,
 		PubID:         pbsmetrics.PublisherUnknown,
 		Browser:       pbsmetrics.BrowserOther,
@@ -143,13 +143,12 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
-	labels.Source = pbsmetrics.DemandWeb
 	if usersyncs.LiveSyncCount() == 0 {
 		labels.CookieFlag = pbsmetrics.CookieFlagNo
 	} else {
 		labels.CookieFlag = pbsmetrics.CookieFlagYes
 	}
-	if req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
+	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 		labels.PubID = req.Site.Publisher.ID
 	}
 
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -149,7 +149,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	} else {
 		labels.CookieFlag = pbsmetrics.CookieFlagYes
 	}
-	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
+	if req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 		labels.PubID = req.Site.Publisher.ID
 	}
 
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -91,7 +91,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	labels := pbsmetrics.Labels{
 		Source:        pbsmetrics.DemandUnknown,
 		RType:         pbsmetrics.ReqTypeAMP,
-		PubID:         "",
+		PubID:         pbsmetrics.PublisherUnknown,
 		Browser:       pbsmetrics.BrowserOther,
 		CookieFlag:    pbsmetrics.CookieFlagUnknown,
 		RequestStatus: pbsmetrics.RequestStatusOK,
@@ -151,8 +151,6 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	}
 	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 		labels.PubID = req.Site.Publisher.ID
-	} else {
-		labels.PubID = "UNKNOWN"
 	}
 
 	response, err := deps.ex.HoldAuction(ctx, req, usersyncs, labels, &deps.categories)
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -143,16 +143,18 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
-	if req.App != nil {
-		labels.Source = pbsmetrics.DemandApp
+	labels.Source = pbsmetrics.DemandWeb
+	if usersyncs.LiveSyncCount() == 0 {
+		labels.CookieFlag = pbsmetrics.CookieFlagNo
 	} else {
-		labels.Source = pbsmetrics.DemandWeb
-		if usersyncs.LiveSyncCount() == 0 {
-			labels.CookieFlag = pbsmetrics.CookieFlagNo
-		} else {
-			labels.CookieFlag = pbsmetrics.CookieFlagYes
-		}
+		labels.CookieFlag = pbsmetrics.CookieFlagYes
 	}
+	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
+		labels.PubID = req.Site.Publisher.ID
+	} else {
+		labels.PubID = "UNKNOWN"
+	}
+
 	response, err := deps.ex.HoldAuction(ctx, req, usersyncs, labels, &deps.categories)
 	ao.AuctionResponse = response
 
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -89,7 +89,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	labels := pbsmetrics.Labels{
 		Source:        pbsmetrics.DemandUnknown,
 		RType:         pbsmetrics.ReqTypeORTB2Web,
-		PubID:         "",
+		PubID:         pbsmetrics.PublisherUnknown,
 		Browser:       pbsmetrics.BrowserOther,
 		CookieFlag:    pbsmetrics.CookieFlagUnknown,
 		RequestStatus: pbsmetrics.RequestStatusOK,
@@ -139,9 +139,6 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 			labels.PubID = req.Site.Publisher.ID
 		}
-	}
-	if labels.PubID == "" {
-		labels.PubID = "UNKNOWN"
 	}
 
 	numImps = len(req.Imp)
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -123,13 +123,15 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
-	if req.Site != nil && req.Site.Publisher != nil {
+	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 		labels.PubID = req.Site.Publisher.ID
+	} else {
+		labels.PubID = "UNKNOWN"
 	}
 	if req.App != nil {
 		labels.Source = pbsmetrics.DemandApp
 		labels.RType = pbsmetrics.ReqTypeORTB2App
-		if req.App.Publisher != nil {
+		if labels.PubID == "UNKNOWN" && req.App.Publisher != nil && req.App.Publisher.ID != "" {
 			labels.PubID = req.App.Publisher.ID
 		}
 	} else {
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -114,16 +114,6 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		return
 	}
 
-	if req.Site != nil && req.Site.Publisher != nil {
-		labels.PubID = req.Site.Publisher.ID
-	}
-	if req.App != nil {
-		labels.RType = pbsmetrics.ReqTypeORTB2App
-		if req.App.Publisher != nil {
-			labels.PubID = req.App.Publisher.ID
-		}
-	}
-
 	ctx := context.Background()
 	cancel := func() {}
 	timeout := deps.cfg.AuctionTimeouts.LimitAuctionTimeout(time.Duration(req.TMax) * time.Millisecond)
@@ -133,8 +123,15 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
+	if req.Site != nil && req.Site.Publisher != nil {
+		labels.PubID = req.Site.Publisher.ID
+	}
 	if req.App != nil {
 		labels.Source = pbsmetrics.DemandApp
+		labels.RType = pbsmetrics.ReqTypeORTB2App
+		if req.App.Publisher != nil {
+			labels.PubID = req.App.Publisher.ID
+		}
 	} else {
 		labels.Source = pbsmetrics.DemandWeb
 		if usersyncs.LiveSyncCount() == 0 {
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -123,15 +123,10 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
-	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
-		labels.PubID = req.Site.Publisher.ID
-	} else {
-		labels.PubID = "UNKNOWN"
-	}
 	if req.App != nil {
 		labels.Source = pbsmetrics.DemandApp
 		labels.RType = pbsmetrics.ReqTypeORTB2App
-		if labels.PubID == "UNKNOWN" && req.App.Publisher != nil && req.App.Publisher.ID != "" {
+		if req.App.Publisher != nil && req.App.Publisher.ID != "" {
 			labels.PubID = req.App.Publisher.ID
 		}
 	} else {
@@ -141,6 +136,12 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		} else {
 			labels.CookieFlag = pbsmetrics.CookieFlagYes
 		}
+		if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
+			labels.PubID = req.Site.Publisher.ID
+		}
+	}
+	if labels.PubID == "" {
+		labels.PubID = "UNKNOWN"
 	}
 
 	numImps = len(req.Imp)
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -129,14 +129,14 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		if req.App.Publisher != nil && req.App.Publisher.ID != "" {
 			labels.PubID = req.App.Publisher.ID
 		}
-	} else {
+	} else { //req.Site != nil
 		labels.Source = pbsmetrics.DemandWeb
 		if usersyncs.LiveSyncCount() == 0 {
 			labels.CookieFlag = pbsmetrics.CookieFlagNo
 		} else {
 			labels.CookieFlag = pbsmetrics.CookieFlagYes
 		}
-		if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
+		if req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 			labels.PubID = req.Site.Publisher.ID
 		}
 	}
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -142,29 +142,6 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 			errL = []error{err}
 			return
 		}
-		if bidReq.App == nil && bidReq.Site == nil {
-			err = fmt.Errorf("Invalid JSON in Default Request Settings: %s", err)
-			errL = []error{err}
-			return
-		} else if bidReq.App.Publisher == nil && bidReq.Site.Publisher == nil {
-			err = fmt.Errorf("Invalid JSON in Default Request Settings: %s", err)
-			errL = []error{err}
-			return
-		} else if bidReq.App != nil && bidReq.Site != nil {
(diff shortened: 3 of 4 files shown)


endpoints/openrtb2/amp_auction.go
```diff
@@ -143,16 +143,18 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
-	if req.App != nil {
-		labels.Source = pbsmetrics.DemandApp
+	labels.Source = pbsmetrics.DemandWeb
```
mansinahar: Is there a reason why we're assuming the request to be of type `web` always and not considering the possibility of it being an `app`?

endpoints/openrtb2/video_auction.go
```diff
@@ -198,6 +201,12 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 			labels.CookieFlag = pbsmetrics.CookieFlagYes
 		}
 	}
+	if labels.PubID == "" && bidReq.Site != nil && bidReq.Site.Publisher != nil {
```
mansinahar: I think we can get away with skipping the `labels.PubID == ""` check in this if statement because the next if statement on L207 checks that and assigns `labels.PubID` to `UNKNOWN` if true.

endpoints/openrtb2/amp_auction.go
guscarreon: This endpoint does not support Apps, therefore we always expect `req.App` to be `NULL`

endpoints/openrtb2/video_auction.go
guscarreon: Haven't noticed that. The logic is valid though because in this type of request we'll either have an `App` or a `Site`, therefore, if `labels.PubID == ""` is true, we can only find the `Publisher.ID` in the `Site` field or it is just `UNKNOWN`
```
 || labels.PubID == "" | bidReq.Site != nil | bidReq.Site.Publisher != nil  ||      A & B & C
1||     true           |      true          |         true                  ||  labels.PubID = bidReq.Site.Publisher.ID
2||     true           |      true          |         false                 ||  labels.PubID = "UNKNOWN"
3||     true           |      false         |         true                  ||  labels.PubID = "UNKNOWN"
4||     true           |      false         |         false                 ||  labels.PubID = "UNKNOWN"
5||     false          |      true          |         true                  ||  labels.PubID = bidReq.App.Publisher.ID
6||     false          |      true          |         false                 ||  labels.PubID = bidReq.App.Publisher.ID
7||     false          |      false         |         true                  ||  labels.PubID = bidReq.App.Publisher.ID
8||     false          |      false         |         false                 ||  labels.PubID = bidReq.App.Publisher.ID
```

guscarreon: Tell you what, in order for the code to look cleaner, maybe the following rearrangement looks better:
```
	if labels.PubID == "" {
	    if bidReq.Site != nil && bidReq.Site.Publisher != nil {
            labels.PubID = bidReq.Site.Publisher.ID
        } else {
		    labels.PubID = "UNKNOWN"
        }
	}
```
Do you agree with this approach?

mansinahar: I think I misread this. You're right. It's fine as is. Sorry about that!

hhhjort:
Is there another PR someplace that does this change for the standard openrtb endpoint? (`auction.go`)

guscarreon:
@hhhjort The current version of `endpoints/openrtb2/auction.go` seems to be doing it already from line 117 to 125:
Current `endpoints/openrtb2/auction.go`
```
  75 func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
  76
  77     ao := analytics.AuctionObject{
  78 +-- 38 lines: Status: http.StatusOK,------------------------------------------------------------
 116
 117     if req.Site != nil && req.Site.Publisher != nil {
 118         labels.PubID = req.Site.Publisher.ID
 119     }
 120     if req.App != nil {
 121         labels.RType = pbsmetrics.ReqTypeORTB2App
 122         if req.App.Publisher != nil {
 123             labels.PubID = req.App.Publisher.ID
 124         }
 125     }
 126
 127     ctx := context.Background()
 128     cancel := func() {}
 129     timeout := deps.cfg.AuctionTimeouts.LimitAuctionTimeout(time.Duration(req.TMax) * time.Millisecond)
 130     if timeout > 0 {
 131         ctx, cancel = context.WithDeadline(ctx, start.Add(timeout))
 132     }
 133     defer cancel()
 134
 135     usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
 136     if req.App != nil {
 137         labels.Source = pbsmetrics.DemandApp
 138     } else {
 139         labels.Source = pbsmetrics.DemandWeb
 140         if usersyncs.LiveSyncCount() == 0 {
 141             labels.CookieFlag = pbsmetrics.CookieFlagNo
 142         } else {
 143             labels.CookieFlag = pbsmetrics.CookieFlagYes
 144         }
 145     }
 146
 147     numImps = len(req.Imp)
```
After looking at it though, i believe the if statments in lines 120 and 136 could be merged into one. I will try this approach, test and update the PR if needed.

endpoints/openrtb2/video_auction.go
josephveach: The proposed change above is also slightly more efficient than the code currently in review:  it avoids repeating the check of labels.PubID.

endpoints/openrtb2/auction.go
```diff
@@ -133,8 +123,15 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
+	if req.Site != nil && req.Site.Publisher != nil {
+		labels.PubID = req.Site.Publisher.ID
+	}
 	if req.App != nil {
 		labels.Source = pbsmetrics.DemandApp
+		labels.RType = pbsmetrics.ReqTypeORTB2App
+		if req.App.Publisher != nil {
```
mansinahar: Shouldn't we also check if `req.App.Publisher.ID != ""` ? 

Also, after this if condition is executed, do we need to check `labels.PubID = ""` and set it to `UNKNOWN` if it is?

endpoints/openrtb2/auction.go
```diff
@@ -133,8 +123,15 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
+	if req.Site != nil && req.Site.Publisher != nil {
+		labels.PubID = req.Site.Publisher.ID
+	}
 	if req.App != nil {
```
mansinahar: I am assuming that the request will be either from an `app` or a `site` so wouldn't it be good to have a sanity check that will return an error if both `req.Site` and `req.App` are not nil for some reason unless that is already done at a higher level in the code?

endpoints/openrtb2/video_auction.go
mansinahar: I think we should move this if condition in the above else condition on L196 because say if `req.App` wasn't nil then it's unlikely that `req.Site` will also not be nil. So then in that case we shouldn't need to check again for `req.Site` too. After the if/else, we should just check `if labels.PubID == ""` and set it to `UNKNOWN`

endpoints/openrtb2/auction.go
guscarreon: Fair point. Changes made

endpoints/openrtb2/video_auction.go
```diff
@@ -198,6 +203,11 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 			labels.CookieFlag = pbsmetrics.CookieFlagYes
 		}
 	}
+	if labels.PubID == "UNKNOWN" && bidReq.Site != nil && bidReq.Site.Publisher != nil && bidReq.Site.Publisher.ID != "" {
+		labels.PubID = bidReq.Site.Publisher.ID
+	} else {
+		labels.PubID = "UNKNOWN"
```
josephveach: This is not right!  Line 209 can reset PubID to UNKNOWN when it already has a value.

endpoints/openrtb2/auction.go
mansinahar: I am sorry Gus but I don't see that check in the update. Can you point me to it?

endpoints/openrtb2/auction.go
```diff
@@ -133,8 +123,17 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	defer cancel()
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
+	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
+		labels.PubID = req.Site.Publisher.ID
+	} else {
+		labels.PubID = "UNKNOWN"
```
mansinahar: Why are we setting `labels.PubID = "UNKNOWN"` over here? My suggestion is that in the if block on L131, we check if `req.app.publisher.id != ""` and set it to that if it's true and in the else block check if `req.site.publisher.id != ""` and set it to that if it's true and then after the if/else block check if `labels.PubID = ""` and if it is then set it to `UNKNOWN`. According to the current code, if it's an app request then you're setting it to `UNKNOWN` first always and then resetting it to what `req.app.publisher.id` has.

Same comment goes for `video_auction.go` too

endpoints/openrtb2/video_auction.go
guscarreon: Copy/Paste error. Corrected

endpoints/openrtb2/video_auction.go
guscarreon: Done

endpoints/openrtb2/auction.go
guscarreon: Yep, the code changed

endpoints/openrtb2/auction.go
guscarreon: I didn't quite understand Mansi, I'm sorry

guscarreon: Do you mean something like?
**BEFORE**
```
 125     usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
 126     if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 127         labels.PubID = req.Site.Publisher.ID
 128     } else {
 129         labels.PubID = "UNKNOWN"
 130     }
 131     if req.App != nil {
 132         labels.Source = pbsmetrics.DemandApp
 133         labels.RType = pbsmetrics.ReqTypeORTB2App
 134         if labels.PubID == "UNKNOWN" && req.App.Publisher != nil && req.App.Publisher.ID != "" {
 135             labels.PubID = req.App.Publisher.ID
 136         }
 137     } else {
 138         labels.Source = pbsmetrics.DemandWeb
 139         if usersyncs.LiveSyncCount() == 0 {
 140             labels.CookieFlag = pbsmetrics.CookieFlagNo
 141         } else {
 142             labels.CookieFlag = pbsmetrics.CookieFlagYes
 143         }
 144     }
auction.go
```
**AFTER**
```
 125     usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
 126     if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
 127         labels.PubID = req.Site.Publisher.ID
 128     }
 129     if req.App != nil {
 130         labels.Source = pbsmetrics.DemandApp
 131         labels.RType = pbsmetrics.ReqTypeORTB2App
 132         if req.App.Publisher != nil && req.App.Publisher.ID != "" {
 133             labels.PubID = req.App.Publisher.ID
 134         } else {
 135             labels.PubID = "UNKNOWN"
 136         }
 137     } else {
 138         labels.Source = pbsmetrics.DemandWeb
 139         if usersyncs.LiveSyncCount() == 0 {
 140             labels.CookieFlag = pbsmetrics.CookieFlagNo
 141         } else {
 142             labels.CookieFlag = pbsmetrics.CookieFlagYes
 143         }
 144     }
auction.go [+]
```

guscarreon: But apparently a request will have one value or the other (Site or App) so if we find a site publisher value in the if statement in line 126, we'll most likely won't find a value for App.Publisher latter. What do you think?

mansinahar:
```
	if req.App != nil {
		if req.App.Publisher != nil && req.App.Publisher.ID != "" {
			labels.PubID = req.App.Publisher.ID
		}
	} else {
		labels.Source = pbsmetrics.DemandWeb
		if usersyncs.LiveSyncCount() == 0 {
			labels.CookieFlag = pbsmetrics.CookieFlagNo
		} else {
			labels.CookieFlag = pbsmetrics.CookieFlagYes
		}
                if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
		        labels.PubID = req.Site.Publisher.ID
	        } 
	}
        if labels.PubID == "" {
                labels.PubID = "UNKNOWN"
       }
```

That's what I am suggesting ^^

endpoints/openrtb2/video_auction.go
```diff
@@ -198,6 +203,9 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 			labels.CookieFlag = pbsmetrics.CookieFlagYes
 		}
 	}
+	if labels.PubID == "UNKNOWN" && bidReq.Site != nil && bidReq.Site.Publisher != nil && bidReq.Site.Publisher.ID != "" {
```
josephveach: PubID can have three values at this point:  (1) "" from line 77, (2) bidReq.App.Publisher.ID from line 194, or (3) "UNKNOWN" from line 196.  The logic from this line suggests we want to set PubID to bidReq.Site.Publisher.ID only if bidReq.App was not nil but there was no bidReq.App.Publisher.ID.  If so, this if-statement should be modified and then used to replace line 196. If not, the logic here is incorrect.

endpoints/openrtb2/auction.go
guscarreon: Ohh the error check you mean? but if there's no `req.Site` nor `req.App` we just set to "UNKNOWN" isn't it? Do you mean that we should implement an additional error message?

endpoints/openrtb2/video_auction.go
guscarreon: Yep, I knew there was a reason for me to have my `labels.PubID == "" &&` check here. I don't know when did I decide otherwise. Corrected

endpoints/openrtb2/auction.go
hhhjort: A request is not guaranteed to have either `app` or `site`. It is an error to have both,but having neither is not strictly an error.

guscarreon: @hhhjort do you want me to implement an error check to catch an instance were we have both?

guscarreon: If not, errors here have been corrected and I we can probably merge (if I didn't miss any further issues that is)

hhhjort: The error check should already exist elsewhere, but I am not 100% sure of it. Just pointing out you should be able to assume not both exist, but you can't assume at least one exists when constructing your logic.

endpoints/openrtb2/auction.go
mansinahar: I meant that we check if both `req` and `app` are set and if they are then return an error.

guscarreon:
> The UNKNOWN string value used for PubID should be a constant defined in metrics.go, similar to pbsmetrics.DemandUnknown, pbsmetrics.CookieFlagUnknown, etc. That constant should be the default value assigned when initializing PubID, which will eliminate the need for checking if the string is empty in each of the updated files. Looking at other constants in that file, I'd suggest using lower case for the new value to be consistent.

Is that important enough to hold the merge?

josephveach:
Yes, the UNKNOWN string should be corrected now.  Not fixing it now would introduce one of the exact type of issues that https://github.com/prebid/prebid-server/pull/922 was just trying to eliminate.  There is no urgency for this ticket that would justify creating an issue that would need to be addressed later at greater cost.

endpoints/openrtb2/video_auction.go
```diff
@@ -142,6 +142,29 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+			labels.Source = pbsmetrics.DemandApp
+			if bidReq.App.Publisher != nil && bidReq.App.Publisher.ID != "" {
+				labels.PubID = bidReq.App.Publisher.ID
+			}
+		} else { // bidReq.Site != nil
+			labels.Source = pbsmetrics.DemandWeb
+			if bidReq.Site.Publisher != nil && bidReq.Site.Publisher.ID != "" {
+				labels.PubID = bidReq.Site.Publisher.ID
+			}
+		}
```
josephveach: 1) this code block shouldn't be run until after videoBidReq is merged into bidReq; note that labels.Source and labels.PubID will never be populated if no deps.default.Request
2) this code is very inefficient:
- three tests for nil-ness of bidReq.App will be run in every non-error case
- two tests for nil-ness of bidReq.Site will be run in every non-error case
3) line 146:  err passed as arg to fmt.Errorf has no value, wrong error message
4) line 149:  will always panic for correct input given that at least one of bidReq.App and bidReq.Site must be nil
5) line 149:  not an error if neither Publisher value is set
6) line 150:  err passed as arg to fmt.Errorf has no value, though not an error
7) line 154:  err passed as arg to fmt.Errorf has no value

This is roughly how I would recommend writing the above block of code:

if bidReq.App != nil {
        if bidReq.Site != nil {
                // set error1 here -- both are set, probably via errors.New()
                return
        }
        if bidReq.App.Publisher != nil && bidReq.App.Publisher.ID != "" {
                labels.PubID = bidReq.App.Publisher.ID
        }
} else {
        if bidReq.Site == nil {
                // set error2 here -- neither is set, probably via errors.New()
                return
        }
        labels.Source = pbsmetrics.DemandWeb
        if bidReq.Site.Publisher != nil && bidReq.Site.Publisher.ID != "" {
                labels.PubID = bidReq.Site.Publisher.ID
        }
}

guscarreon: 1) three tests for nil-ness of bidReq.App will be run in every non-error case
       Lines 145 to 152 were left as is. Their modification is out of the scope of this PR and I personally don't mind the current if/else layout
       because it delimits every scenario in a more readable way.
    2) two tests for nil-ness of bidReq.Site will be run in every non-error case
       Same as above.
    3) line 146: err passed as arg to fmt.Errorf has no value, wrong error message
       I didn't write this part of the code, correcting those errors should be handled in a separate issue unless otherwise instructed
    4) line 149: will always panic for correct input given that at least one of bidReq.App and bidReq.Site must be nil
       Line was already sitting there. I added lines 153 to 167.
    5) line 149: not an error if neither Publisher value is set
       Line was already sitting there. Since this part of the code has been code reviewed, approved and in production already, I
       assumed it to be correct.
    6) line 150: err passed as arg to fmt.Errorf has no value, though not an error
       Doesn't affect anything though
    7) line 154: err passed as arg to fmt.Errorf has no value
       You are right, forgot to add the '%s'. No errors were thrown though.
    8) This is roughly how I would recommend writing the above block of code:
       So, it's basically the same approach I used from line 157 on, but I didn't do it because it omits the check in line 149 which, given this
       code is sitting in production and has been previously reviewed and code approved, I didn't touch because I assumed it behaved as desired

guscarreon: Joe I confused this PR with another one. Went back to a previous approach that corrects the issues you listed. I'll wait for any feedback

endpoints/openrtb2/amp_auction.go
```diff
@@ -143,16 +143,16 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
```
mansinahar: Why is the `Source` label changing from `DemandApp` to `DemandWeb`? Shouldn't it be `DemandApp` as it's an amp request?

endpoints/openrtb2/amp_auction.go
```diff
@@ -143,16 +143,16 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
 	} else {
-		labels.Source = pbsmetrics.DemandWeb
-		if usersyncs.LiveSyncCount() == 0 {
-			labels.CookieFlag = pbsmetrics.CookieFlagNo
-		} else {
-			labels.CookieFlag = pbsmetrics.CookieFlagYes
-		}
+		labels.CookieFlag = pbsmetrics.CookieFlagYes
 	}
+	if req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
```
mansinahar: Do we expect Site to be set in an amp request?

pbsmetrics/metrics.go
```diff
@@ -61,6 +61,8 @@ const (
 	DemandUnknown DemandSource = "unknown"
 )
 
+const PublisherUnknown = "unknown"
```
mansinahar: This is just a minor thing but I'd suggest moving this too either before the demand source related definitions on L56 or after on L73. This is just to keep related definitions together so it's easy to read.

endpoints/openrtb2/auction.go
```diff
@@ -135,13 +125,20 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+			labels.PubID = req.App.Publisher.ID
+		}
+	} else { //req.Site != nil
 		labels.Source = pbsmetrics.DemandWeb
 		if usersyncs.LiveSyncCount() == 0 {
 			labels.CookieFlag = pbsmetrics.CookieFlagNo
 		} else {
 			labels.CookieFlag = pbsmetrics.CookieFlagYes
 		}
+		if req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
```
mansinahar: I think we should keep the `req.Site != nil` check in this if condition unless we are sure that it will always be non-nil in this flow of the code. Similar for video_auction too.

endpoints/openrtb2/amp_auction.go
guscarreon: in amp_auction we don't expect to ever have an App field so the entire `req.App` conditional can be discarded because we won't be using that value at all

endpoints/openrtb2/amp_auction.go
guscarreon: Yes. The parseAmpRequest function calls `(deps *endpointDeps) validateRequest(req *openrtb.BidRequest)` in line 262. The check in line 303 of the `validateRequest` function, guarantees `req.Site` and `req.App` are not both NULL nor not NULL

endpoints/openrtb2/amp_auction.go
mansinahar: I don't think it hurts to check that but the question is shouldn't `labels.Source` be assigned `pbsmetrics.DemandApp`? Why is that changing to `pbsmetrics.DemandWeb` ?

endpoints/openrtb2/auction.go
guscarreon: In this instance we are because the check in line 303 of the `validateRequest` function, guarantees `req.Site` and `req.App` are not both NULL nor not NULL

endpoints/openrtb2/amp_auction.go
guscarreon:
```
 258 func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) []error {
 259     errL := []error{}
 260 +-- 42 lines: if req.ID == "" {-----------------------------------------------
 302
 303     if (req.Site == nil && req.App == nil) || (req.Site != nil && req.App != nil) {
 304         errL = append(errL, errors.New("request.site or request.app must be defined, but not both."))
 305         return errL
 306     }
 307
```

endpoints/openrtb2/amp_auction.go
guscarreon: As my understanding, it should be a Site.

endpoints/openrtb2/amp_auction.go
guscarreon: This is a fair point. The check in line 257, `errL := deps.validateRequest(req)` can return true even if `req.Site` is NULL. Let me put it back in.

endpoints/openrtb2/amp_auction.go
guscarreon: so we should use `DemmandWeb`

endpoints/openrtb2/auction.go
guscarreon: We are sure Mansi. The `parseRequest(r)` function gets executed in line 110 just before our if statements here this function in turn calls `validateRequest(r)` that makes the following check:
```
 296     if (req.Site == nil && req.App == nil) || (req.Site != nil && req.App != nil) {
 297         errL = append(errL, errors.New("request.site or request.app must be defined, but not both."))
 298         return errL
 299     }
 300
 301 +-- 19 lines: if err := deps.validateSite(req.Site); err != nil {-----------------------------------
 320
 321     return errL
 322 }
endpoints/openrtb2/auction.go [RO]
```

pbsmetrics/metrics.go
guscarreon: Done

Issue #920, opened 2019-05-28
[doc] Update PBS documentation go version


benjaminch:
Following #910 we might want to update the documentation in order to deprecate Go 1.9.
Documentation to be updated: https://github.com/prebid/prebid-server#installation

Go 1.9 should then be removed from the CI in order to support only last two versions.

PR #921, opened 2019-05-28
[Go 1.9] deprecate 1.9 & 1.10 versions and introduce 1.12

benjaminch:
This CL deprecates Go 1.9 version removing it from the CI along updating
the installation instructions.

This PR probably needs to be flagged as important change so people using Go 1.9 will be aware of this change from now on.

Issue: #920

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: go
 
 go:
-  - '1.9'
-  - '1.10'
   - '1.11.1'
+  - '1.12'
 
 go_import_path: github.com/prebid/prebid-server
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ For more information, see:
 
 ## Installation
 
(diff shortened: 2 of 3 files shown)


benjaminch:
> Since [Go only supports the last two major releases](https://golang.org/doc/devel/release.html#policy), versions 1.11 and 1.12 are the only ones currently being supported. Do you think we should try to deprecate 1.10.x too?

Sure ! :)

benjaminch:
Hey @guscarreon seems that 1.12 requires a bit more work especially around `validate.sh` for `go vet`. I updated it let me know if it works for you.

Cheers!

guscarreon:
> Hey @guscarreon seems that 1.12 requires a bit more work especially around `validate.sh` for `go vet`. I updated it let me know if it works for you.
> 
> Cheers!

Worked perfeclty. LGTM

benjaminch:
Thanks @guscarreon :)

PR #922, opened 2019-05-29
Address Sonar analysis comments (Severity:Critical)

mansinahar:


stored_requests/caches/cachestest/reliable.go
```diff
@@ -32,62 +39,63 @@ func cacheMissTester(cache stored_requests.Cache) func(*testing.T) {
...
-		assertHasValue(t, impData, "id", `{"imp":true}`)
+		assertHasValue(t, impData, commonKey, impCacheVal)
 	}
 }
 
 func cacheSaveInvalidateTester(cache stored_requests.Cache) func(*testing.T) {
 	return func(t *testing.T) {
 		cache.Save(context.Background(), map[string]json.RawMessage{
-			"known": json.RawMessage(`{}`),
+			reqCacheKey: json.RawMessage(reqCacheVal),
```
josephveach: Does having different values here compared to before affect this test?

endpoints/openrtb2/auction.go
```diff
@@ -456,14 +457,14 @@ func validateNativeContextTypes(cType native.ContextType, cSubtype native.Contex
...
-		return fmt.Errorf("request.imp[%d].native.request.contextsubtype is invalid. See https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=39", impIndex)
+		return fmt.Errorf("request.imp[%d].native.request.contextsubtype value can't be less than 0. See https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=39", impIndex)
 	}
 	if cSubtype == 0 {
 		return nil
 	}
 
 	if cSubtype >= 100 {
-		return fmt.Errorf("request.imp[%d].native.request.contextsubtype is invalid. See https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=39", impIndex)
+		return fmt.Errorf("request.imp[%d].native.request.contextsubtype can't be greater than or equal to 100. See https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=39", impIndex)
```
josephveach: The linked page suggests that there are some valid values over 100, though I suppose not currently in our implementation.  Maybe the old message is safer here!?

exchange/exchange.go
```diff
@@ -157,7 +157,7 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	return e.buildBidResponse(ctx, liveAdapters, adapterBids, bidRequest, resolvedRequest, adapterExtra, errs)
 }
 
-func (e *exchange) makeAuctionContext(ctx context.Context, needsCache bool) (auctionCtx context.Context, cancel func()) {
+func (e *exchange) makeAuctionContext(ctx context.Context, needsCache bool) (auctionCtx context.Context, cancel context.CancelFunc) {
 	auctionCtx = ctx
 	cancel = func() {}
```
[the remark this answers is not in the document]
mansinahar: We do! The `cancel` variable is sort of getting declared in the function definition but we're initializing it here.

stored_requests/caches/cachestest/reliable.go
mansinahar: I don't think it affects the purpose of the test per se. This was done more to define one set of req and imp cache key and value to be used by all tests rather than defining new ones for each one.

endpoints/openrtb2/auction.go
mansinahar: I changed it to that to indicate that we expect the values to be only below 100. Consider, if some user uses a value above 100 and gets back the old error message, they wouldn't know why exactly are they getting back an error. According to them it should be valid, given the 
IAB spec but with the new error message it will be clear to them that the reason for the error is because we don't support a value that is greater than or equal to 100. Hope that makes sense!

adapters/appnexus/appnexus.go
```diff
@@ -285,8 +285,8 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters
 	// AppNexus openrtb2 endpoint expects imp.displaymanagerver to be populated, but some SDKs will put it in imp.ext.prebid instead
 	var defaultDisplayManagerVer string
 	if request.App != nil {
-		source, err1 := jsonparser.GetString(request.App.Ext, "prebid", "source")
-		version, err2 := jsonparser.GetString(request.App.Ext, "prebid", "version")
+		source, err1 := jsonparser.GetString(request.App.Ext, openrtb_ext.PrebidExtKey, "source")
+		version, err2 := jsonparser.GetString(request.App.Ext, openrtb_ext.PrebidExtKey, "version")
 		if (err1 == nil) && (err2 == nil) {
 			defaultDisplayManagerVer = fmt.Sprintf("%s-%s", source, version)
 		}
```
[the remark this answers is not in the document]
mansinahar: I am not sure if the value of `defaultDisplayManagerVer` is important enough for us to return an error but maybe @hhhjort can weigh in on this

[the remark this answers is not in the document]
hhhjort: This was added to help impbus, it is not an error if we can't build `defaultDisplayManagerVer`. I am not sure if it is useful to propagate these errors back to the caller.

endpoints/openrtb2/auction.go
```diff
@@ -527,34 +529,45 @@ func validateNativeAsset(asset nativeRequests.Asset, impIndex int, assetIndex in
...
 		}
 	}
 
+	if multipleAssets {
+		return fmt.Errorf("request.imp[%d].native.request.assets[%d] must define at most one of {title, img, video, data}", impIndex, assetIndex)
+	}
+
+	if !foundType {
+		return fmt.Errorf("request.imp[%d].native.request.assets[%d] must define at least one of {title, img, video, data}", impIndex, assetIndex)
+	}
```
josephveach: The introduction of multipleAssets in this function extends execution of the function beyond when it might have returned in the function's original form with no change to the result.  Do we really need to do things like this just to satisfy a coding tool?

The final check for foundType changes the behavior of this function from before.  Does that mean the original code has a bug?

endpoints/openrtb2/auction.go
josephveach: It does.  My point was that the page mentioned isn't consistent with the new statement:  the page says there are possible valid values above 100.  Seems potentially confusing, though not a big deal.

mansinahar: Yeah, your point is totally valid but I just think that the old message would be much more confusing than the new one as the new one at least indicates that we don't support a value greater than or equal 100.

What do you think about the error message to say: `request.imp[%d].native.request.contextsubtype value greater than or equal to 100 isn't supported`? Do you think that would be more useful? If not, I am happy to consider any other suggestions that you think would be easier for the user to understand what went wrong.

endpoints/openrtb2/auction.go
mansinahar: Do you mean that with the introduction of `multipleAssets` we're unnecessarily checking for each asset and then return when we could have just returned when we saw the first multipleAsset? If so, yeah, you make a fair point and well, this is not just to satisfy the tool but to make the code simpler. I guess a better alternative would be to define the error string once and then reuse that rather than checking `multipleAssets` or using the same error string again and again.

I will update the PR. Thanks for pointing it out

endpoints/openrtb2/auction.go
josephveach: I don't feel very strongly about any particular solution.  My slight preference would be to use your updated text but to remove the reference to the page from the error message.

endpoints/openrtb2/auction.go
josephveach: Yes, that's what I meant.  Probably not a big deal since it's an error case and, hopefully will happen very rarely.  My bigger concern was related to my second comment, as that introduces behavior that wasn't present before.  Why is that being done?

mansinahar: Yeah, the other check is basically making sure that there's at least one native asset present of {title, image, video, data}. With none of them present, the asset object should be considered as invalid. IAB Spec reference: https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=13.

endpoints/openrtb2/auction.go
mansinahar: Actually, I pondered a little more over this after you pointed it out and wondered if that check should be actually checking for a value greater than 500 rather than greater than 100 and confirmed with @hhhjort that it should actually check for value greater than 500. Will update the PR accordingly.

Thanks for pointing it out @josephveach!

adapters/appnexus/appnexus.go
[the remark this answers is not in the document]
mansinahar: Cool. Thanks for the clarification @hhhjort

endpoints/openrtb2/auction.go
josephveach: I was a little concerned that, IMHO, the wording in the above spec was ambiguous, especially given the hitherto code support of no value being present and the wording of the PBS error messages.   I was able to confirm that other company products also expect exactly one of the four values to be present.  This leads to one final suggestion:  The new multipleAssetErr error message mentions "at most one of" but the new foundType message mentions "at least one of".  I propose that both should say "exactly one of" (which is what I would have liked to see in the spec, as well).

mansinahar: Updated @josephveach

mansinahar:
@guscarreon I have addressed your comments and they didn't seem to require any code change. Can you please take a loot at this PR again and comment/approve it? Thanks!

PR #923, opened 2019-05-29
Remove hard coded targeting keys

mansinahar:
This is a clean up PR that:
- Removes hardcoded targeting keys 
- Removes targeting key constants defined in multiple places
- Rename `VideoTargeting` struct fields to adhere to GoLang naming convention

### DIFF
--- a/endpoints/auction.go
+++ b/endpoints/auction.go
@@ -440,13 +440,13 @@ func sortBidsAddKeywordsMobile(bids pbs.PBSBidSlice, pbs_req *pbs.PBSRequest, pr
 
 			hbPbBidderKey := string(openrtb_ext.HbpbConstantKey) + "_" + bid.BidderCode
 			hbBidderBidderKey := string(openrtb_ext.HbBidderConstantKey) + "_" + bid.BidderCode
-			HbCacheIDBidderKey := string(openrtb_ext.HbCacheKey) + "_" + bid.BidderCode
+			hbCacheIDBidderKey := string(openrtb_ext.HbCacheKey) + "_" + bid.BidderCode
 			hbDealIDBidderKey := string(openrtb_ext.HbDealIDConstantKey) + "_" + bid.BidderCode
 			hbSizeBidderKey := string(openrtb_ext.HbSizeConstantKey) + "_" + bid.BidderCode
 			if pbs_req.MaxKeyLength != 0 {
 				hbPbBidderKey = hbPbBidderKey[:min(len(hbPbBidderKey), int(pbs_req.MaxKeyLength))]
 				hbBidderBidderKey = hbBidderBidderKey[:min(len(hbBidderBidderKey), int(pbs_req.MaxKeyLength))]
-				HbCacheIDBidderKey = HbCacheIDBidderKey[:min(len(HbCacheIDBidderKey), int(pbs_req.MaxKeyLength))]
+				hbCacheIDBidderKey = hbCacheIDBidderKey[:min(len(hbCacheIDBidderKey), int(pbs_req.MaxKeyLength))]
 				hbDealIDBidderKey = hbDealIDBidderKey[:min(len(hbDealIDBidderKey), int(pbs_req.MaxKeyLength))]
 				hbSizeBidderKey = hbSizeBidderKey[:min(len(hbSizeBidderKey), int(pbs_req.MaxKeyLength))]
 			}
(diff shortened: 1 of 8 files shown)


endpoints/auction.go
```diff
@@ -34,11 +34,6 @@ type bidResult struct {
 const defaultPriceGranularity = "med"
 
 // Constant keys for ad server targeting for responses to Prebid Mobile
```
guscarreon: Can we also remove this comment?

endpoints/auction.go
```diff
@@ -445,16 +440,16 @@ func sortBidsAddKeywordsMobile(bids pbs.PBSBidSlice, pbs_req *pbs.PBSRequest, pr
...
 
-			hbPbBidderKey := hbpbConstantKey + "_" + bid.BidderCode
-			hbBidderBidderKey := hbBidderConstantKey + "_" + bid.BidderCode
-			hbCacheIdBidderKey := hbCacheIdConstantKey + "_" + bid.BidderCode
-			hbDealIdBidderKey := hbDealIdConstantKey + "_" + bid.BidderCode
-			hbSizeBidderKey := hbSizeConstantKey + "_" + bid.BidderCode
+			hbPbBidderKey := string(openrtb_ext.HbpbConstantKey) + "_" + bid.BidderCode
+			hbBidderBidderKey := string(openrtb_ext.HbBidderConstantKey) + "_" + bid.BidderCode
+			hbCacheIDBidderKey := string(openrtb_ext.HbCacheKey) + "_" + bid.BidderCode
+			hbDealIDBidderKey := string(openrtb_ext.HbDealIdConstantKey) + "_" + bid.BidderCode
```
guscarreon: Do we need the capital 'D' in the new variable names for `hbCacheIdBidderKey` and `hbDealIdBidderKey`?  This sole letter accounts for many changes throughout the document and the change is so little that I find it a bit unnecessary.

guscarreon:
Aside from a couple of comments, LGTM

endpoints/auction.go
mansinahar: Honestly, I don't have a very strong opinion on this but I know that this is one of the things that GoLint complains about and it would be great to get this code base in a place where we adhere to GoLint's code style rules which is why I thought I will make this change since I am anyway touching that code

endpoints/auction.go
mansinahar: Good catch. Thanks! I will remove that

mansinahar: Updated!

endpoints/auction.go
guscarreon: Ohh I see. I agree that complying with the GoLint standard is overall a better practice.
Everything else LGTM

openrtb_ext/bid_response_video.go
```diff
@@ -14,7 +14,7 @@ type AdPod struct {
 }
 
 type VideoTargeting struct {
-	Hb_pb         string `json:"hb_pb"`
-	Hb_pb_cat_dur string `json:"hb_pb_cat_dur"`
-	Hb_cache_id   string `json:"hb_cache_id"`
+	HbPb       string `json:"hb_pb"`
+	HbPbCatDur string `json:"hb_pb_cat_dur"`
+	HbCacheId  string `json:"hb_cache_id"`
```
josephveach: In view of the earlier discussion of GoLint, why HbCacheId instead of HbCacheID?

Related to this, I see other "Id" instances in the above code (minimally, HbDealIdConstantKey).  Are they outside the scope of what's being addressed here?

mansinahar: That should be `HbCacheID`. My bad! 

And about the "Id" instances, there are a lot of violations of those and other GoLint issues in our code which we should address. We will probably not be able to address all of them in one PR but I agree, refactoring those targeting keys in `openrtb_ext/bid.go` should be minimal and can probably be done as part of this PR. Thanks for the catch Joe. I will update the PR

exchange/auction.go
```diff
@@ -73,11 +73,11 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
 	toCache := make([]prebid_cache_client.Cacheable, 0, expectNumBids+expectNumVast)
 	expByImp := make(map[string]int64)
 	competitiveExclusion := false
-	var hbCacheID string
+	var HbCacheID string
```
hhhjort: Why capitalize `HbCacheID`? Normally that would indicate that `HbCacheID` is being exported, but that is not possible in this scope.

mansinahar: Ah! Good catch @hhhjort. This wasn't on purpose but an effect of rename all :p I will change it back. Thanks for pointing it out.

mansinahar: Updated!

PR #928, opened 2019-06-07
Vrtcal Markets, Inc Adapter Integration

vrtcal-dev:
Vrtcal Markets, Inc Adapter Integration

### DIFF
--- a/adapters/vrtcal/params_test.go
+++ b/adapters/vrtcal/params_test.go
@@ -1,0 +1,46 @@
+package vrtcal
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+//Vrtcal doesn't currently require any custom fields. This file is included for conformity only
+//We do include an unused, non-required custom param in static/bidder-params/vrtcal.json, but only to hinder the prebid server from crashing by looking for at least 1 custom param
+
+// This file actually intends to test static/bidder-params/vrtcal.json
+//
+// These also validate the format of the external API: request.imp[i].ext.vrtcal
+// TestValidParams makes sure that the Vrtcal schema accepts all imp.ext fields which we intend to support.
+func TestValidParams(t *testing.T) {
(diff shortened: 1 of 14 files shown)


vrtcal-dev:
Thanks hhhjort.  The change to Gopkg.lock took place by the 'dep ensure' process. We noticed that it had changed but didn't think it mattered, not being part of the core code. We'll be happy to replace with the one from the master if necessary for your approval. Please let us know if this step is necessary...Thanks again

hhhjort:
Yes, please update your PR so that Gopkg.lock is not included in it.

vrtcal-dev:
We've updated to have the matching version of Gopkg.lock as the original so that it not included as a changed file.

hhhjort:
Seems merge conflicts are happening now. :(

hhhjort:
Still seeing merge conflicts for some reason

vrtcal-dev:
Ok, complete repush again....no conflicts now.

vrtcal-dev:
The last time we were at this point, another adapter was added to the master making our code have conflicts with the master. Is it possible for us to get this merged before this happens again?...Thanks..

mansinahar:
@vrtcal-dev @hhhjort I think we should update the Gopkg.lock file if it is updating any packages as part of doing a `dep ensure` because that means that there's an update to one of the dependencies that we're using and it's good to have updated dependencies unless the update is causing issues for us. In that case, we should fix the issue or pin that dependency to the specific version we want.

vrtcal-dev:
We no longer have a variation on the Gopkg.lock file. You can proceed with our integration as this is no longer an issue for us...Our version matches the master.

vrtcal-dev:
Thanks hhhjort. When can we expect our adapter to be commited to the master?

hhhjort:
We prefer a second review before merging to master. I have flagged this review as urgent so we should hopefully not get another merge conflict from something else merging first again.

vrtcal-dev:
Thanks all. Will we see it added to the master today?

PR #932, opened 2019-06-12
Adform bidder: keyvalues and keywords support

DzmitryKozel:


adapters/adform/adform.go
```diff
@@ -57,6 +58,8 @@ type adformDigitrustPrivacy struct {
 type adformAdUnit struct {
 	MasterTagId json.Number `json:"mid"`
 	PriceType   string      `json:"priceType,omitempty"`
+	KeyValues   string      `json:"mkv,omitempty"`
+	Keywords    string      `json:"mkw,omitempty"`
```
josephveach: Can you please use the mixed caps naming convention here, i.e., KeyWords instead of Keywords?  This change will affect some of the code below, too,

openrtb_ext/imp_adform.go
```diff
@@ -3,4 +3,6 @@ package openrtb_ext
 type ExtImpAdform struct {
 	MasterTagId string `json:"mid"`
 	PriceType   string `json:"priceType,omitempty"`
+	KeyValues   string `json:"mkv,omitempty"`
+	Keywords    string `json:"mkw,omitempty"`
```
josephveach: Please use KeyWords in place of Keywords here.

mansinahar:
LGTM in general. Could you please also update: https://github.com/prebid/prebid-server/blob/master/adapters/adform/adformtest/params/race/banner.json? Thanks!

DzmitryKozel:
@mansinahar why do we need to extend that file,  for me it's enough there to have request with minimum fields.

DzmitryKozel:
> LGTM in general. Could you please also update: https://github.com/prebid/prebid-server/blob/master/adapters/adform/adformtest/params/race/banner.json? Thanks!

@mansinahar why do we need to extend that file, for me it's enough there to have request with minimum fields.

BZSky:
Hey mates! Can I kindly ping you to merge it, as it passed approval already 😊

PR #937, opened 2019-06-14
Updating extensions with recent features

bretg:


hhhjort:
Seems fine to me, but we should definitely make sure we update the prebid.org copy. I will have to poke Mike about his PR to remove the docs from this repo.

PR #938, opened 2019-06-14
[Sharethrough] Pass domain from Site.Page to ad server

maphe:
This extracts the domain from the `Site.Page` url and pass it to the ad server as an additional header.


adapters/sharethrough/butler.go
```diff
@@ -50,10 +50,11 @@ type StrOpenRTBTranslator struct {
 	UserAgentParsers UserAgentParsers
 }
 
-func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openrtb.BidRequest) (*adapters.RequestData, error) {
+func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openrtb.BidRequest, domain string) (*adapters.RequestData, error) {
 	headers := http.Header{}
 	headers.Add("Content-Type", "text/plain;charset=utf-8")
 	headers.Add("Accept", "application/json")
+	headers.Add("Origin", domain)
```
guscarreon: @maphe, we usually pass Site.Page as `headers.Add("Referer", domain)` and not `headers.Add("Origin", domain)`. Any specific reason why are we passing it as an `Origin` header and not a `Referer`?

guscarreon: Please be advised that if your decision to fix this adapter to pull the page info from `Origin` is the correct to your particular context, we won't deny your PR for it. I'm just wondering about it because `Referer` is usually the “standard” way, but if it makes perfect sense for your adapter, we don't have a problem with this approach.

maphe: @guscarreon the main reason is that we don't send the raw `Site.Page`, we parse it to extract the domain. `Referer` is meant to have the full URL whereas `Origin` is a more appropriate header for just a domain.

adapters/sharethrough/sharethrough.go
```diff
@@ -34,9 +34,10 @@ type SharethroughAdapter struct {
 
 func (a SharethroughAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
 	var reqs []*adapters.RequestData
+	var domain = Util{}.parseDomain(request.Site.Page)
```
guscarreon: I see the `TestSuccessMakeRequests` inside `adapters/sharethrough/sharethrough_test.go` was modified to work with `openrtb.Site` instead of `openrtb.App`. Even under this new conditions, are we sure we don't want to add a NULL check on `request.Site` here just to make sure we don't ever get a segmentation fault? If we comment out the `openrtb.BidRequest.Site` field declaration in `TestSuccessMakeRequests`  like this:
```
 43 func TestSuccessMakeRequests(t *testing.T) {
 44     stubReq := &adapters.RequestData{
 45 +-- 12 lines: Method: "POST",-----------------------------------------------------
 57     }{
 58         "Generates expected Request": {
 59             input: &openrtb.BidRequest{
 60                 //Site: &openrtb.Site{
 61                 //  Page: "test.com",
 62                 //},
 63                 Device: &openrtb.Device{
 64                     UA: "Android Chome/60",
 65 +-- 20 lines: },------------------------------------------------------------------
 85     for testName, test := range tests {
 86         t.Logf("Test case: %s\n", testName)
 87
 88         output, actualErrors := adapter.MakeRequests(test.input)
adapters/sharethrough/sharethrough_test.go
```
We get the a segmentation fault exactly here (`adapters/sharethrough/sharethrough.go` line 37). Is this a realistic scenario? Complete message:
```
╰─➤  go test
--- FAIL: TestSuccessMakeRequests (0.00s)
    sharethrough_test.go:86: Test case: Generates expected Request
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x80 pc=0x1474d7e]

goroutine 37 [running]:
testing.tRunner.func1(0xc00017d600)
        /usr/local/opt/go/libexec/src/testing/testing.go:792 +0x387
panic(0x15168a0, 0x19e6980)
        /usr/local/opt/go/libexec/src/runtime/panic.go:513 +0x1b9
github.com/prebid/prebid-server/adapters/sharethrough.SharethroughAdapter.MakeRequests(0x16454e0, 0xc0003de660, 0xc00037a8c0, 0xc000065d48, 0x1, 0x1, 0xc0003ea498, 0x174ff30, 0x19b384b)
        /Users/gcarreongutierrez/go/src/github.com/prebid/prebid-server/adapters/sharethrough/sharethrough.go:37 +0x3e
github.com/prebid/prebid-server/adapters/sharethrough.TestSuccessMakeRequests(0xc00017d600)
        /Users/gcarreongutierrez/go/src/github.com/prebid/prebid-server/adapters/sharethrough/sharethrough_test.go:88 +0x6ea
testing.tRunner(0xc00017d600, 0x15d3468)
        /usr/local/opt/go/libexec/src/testing/testing.go:827 +0xbf
created by testing.(*T).Run
        /usr/local/opt/go/libexec/src/testing/testing.go:878 +0x35c
exit status 2
FAIL    github.com/prebid/prebid-server/adapters/sharethrough   0.026s
```

maphe: @guscarreon The removal of `openrtb.App` is actually not related to the addition of `openrtb.Site`, I just cleaned up the test while I was on it. All usages of `openrtb.App` had been removed in the previous PR, see this commit: https://github.com/sharethrough/prebid-server/commit/74b2f721867ed8871a8168a7e6ca057275a6d9ba

Regarding adding a null check on `request.Site`, as far as I understand it, Prebid core will always make the value accessible, feel free to correct me but I don't think a null `openrtb.Site` is a realistic scenario.

guscarreon: @maphe, prebid-server core will throw an error if both `openrtb.Site` and `openrtb.App` are NULL or not NULL. In other words something similar to:
```
if (req.Site == nil && req.App == nil) || (req.Site != nil && req.App != nil) {
    //throw ERROR
    return
} 
```
Therefore, we might risk a Segmentation Fault in processing a given request coming with an `openrtb.Site` object that is NULL and an `openrtb.App` that is not. Do you think it's worth adding this NULL check in your adapter to cover up for this scenario? I realize you are not using `openrtb.App` anymore but, according to how prebid-server core handles things, `openrtb.Site` might actually be NULL.

maphe: @guscarreon, got it! thanks for the details. We don't support in app placements so in theory we shouldn't have any of those, but still I added a null check on the `request.Site` for extra safety.

maphe:
@mansinahar do you have an ETA on when this can get merged? Thank you

adapters/sharethrough/utils_test.go
```diff
@@ -408,3 +408,30 @@ func TestGdprConsentString(t *testing.T) {
...
+		expected string
+	}{
+		"Parses domain without port": {
+			input:    "http://a.domain.com/page?param=value",
+			expected: "a.domain.com",
+		},
+		"Parses domain with port": {
+			input:    "http://a.domain.com:8000/page?param=value",
+			expected: "a.domain.com",
+		},
```
mansinahar: Might be good to have a test with an invalid URL also

adapters/sharethrough/sharethrough.go
```diff
@@ -35,8 +35,13 @@ type SharethroughAdapter struct {
 func (a SharethroughAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
 	var reqs []*adapters.RequestData
 
+	if request.Site == nil {
+		return nil, []error{fmt.Errorf("the requests does not have a Site, in App placements are not supported")}
+	}
+	var domain = Util{}.parseDomain(request.Site.Page)
```
mansinahar: Just wondering if you're okay sending an empty string as domain because currently the way `parseDomain` works, it can return an empty string

maphe: Yes, that's expected, thanks for double-checking

maphe:
Thank you @mansinahar, I added the test 👍

mansinahar:
Thanks @maphe! And sorry for the delay on this.

adapters/sharethrough/sharethrough.go
```diff
@@ -35,8 +35,13 @@ type SharethroughAdapter struct {
 func (a SharethroughAdapter) MakeRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
 	var reqs []*adapters.RequestData
 
+	if request.Site == nil {
+		return nil, []error{fmt.Errorf("the requests does not have a Site, in App placements are not supported")}
```
josephveach: Minor nit since I have another comment:  it's possible to have better grammar in the above error message.

adapters/sharethrough/sharethrough_test.go
```diff
@@ -104,7 +106,9 @@ func TestFailureMakeRequests(t *testing.T) {
 	}{
 		"Returns nil if failed to generate request": {
 			input: &openrtb.BidRequest{
-				App: &openrtb.App{Ext: []byte(`{}`)},
+				Site: &openrtb.Site{
+					Page: "test.com",
+				},
```
josephveach: Both here and above at line 60, a request with App was replaced by a request with Site.  Does this mean App is no longer supported and therefore no App test is needed?

maphe: App is not supported (actually it was never supported, the App in the test was a mistake), I improved the error message to make it more explicit

maphe:
@josephveach thanks for the review, I addressed the error message wording

PR #941, opened 2019-06-18
Revert "[Go 1.9] deprecate 1.9 & 1.10 versions and introduce 1.12"

mansinahar:
Reverts prebid/prebid-server#921


hhhjort:
SHould hopefully not be needed now

PR #942, opened 2019-06-18
Update golang version to 1.11

mansinahar:


Dockerfile
```diff
@@ -1,12 +1,20 @@
...
+WORKDIR /go/src/github.com/prebid/prebid-server/
+ENV GOROOT=/usr/local/go
+ENV GOPATH=$HOME/go
+ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+RUN apt-get install -y git go-dep && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-ENV GOPATH /go
 ENV CGO_ENABLED 0
 COPY ./ ./
+RUN echo $(go version)
```
mkendall07: probably can remove this line.

mansinahar: Ah! Good catch. That's a leftover. I will update it

PR #944, opened 2019-06-21
Gamma SSP adapter integration

gammassp:


adapters/gamma/gamma.go
```diff
@@ -0,0 +1,184 @@
...
+		if request.Device.DNT != nil {
+			addHeaderIfNonEmpty(headers, "DNT", strconv.Itoa(int(*request.Device.DNT)))
+		}
+	}
+
+	addHeaderIfNonEmpty(headers, "Connection", "keep-alive")
+	addHeaderIfNonEmpty(headers, "cache-control", "no-cache")
+	addHeaderIfNonEmpty(headers, "Accept-Encoding", "gzip, deflate")
+
+	return []*adapters.RequestData{{
```
hhhjort: I don't see where you pass along any `Imp.ID`'s to your bidder, so that you can bid against the various imps. You do pass in the `request.ID` but that is the ID of the overall request, it does not tie directly into any of the imp IDs.

adapters/gamma/gamma.go
```diff
@@ -0,0 +1,184 @@
...
+		headers.Add(headerName, headerValue)
+	}
+}
+
+// getMediaTypeForImp figures out which media type this bid is for.
+func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	mediaType := openrtb_ext.BidTypeBanner //default type
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Video != nil {
```
hhhjort: You are guaranteed to always bid video on multiformat requests?

adapters/gamma/gammatest/exemplary/simple-banner.json
```diff
@@ -0,0 +1,80 @@
...
+      },
+      "mockResponse": {
+        "status": 200,
+        "body": {
+          "id": "test-request-id",
+          "seatbid": [
+            {
+              "seat": "958",
+              "bid": [{
+                "id": "7706636740145184841",
```
hhhjort: As per my comment above, how would your bidder know how to bid against the impid "test-imp-id" given the expected request above?

hhhjort:
You also have some merge conflicts, probably from some other adapter merging since you pulled from the master branch. You probably want to fix this last, to give minimum time for another adapter to merge between fixing the merge conflicts and getting this PR reviewed. We have gotten a lot of adapter PRs lately.

adapters/gamma/gamma.go
gammassp: Ok I have passed the Imp.ID into bid request. Please double check and let me know if it's ok now.

adapters/gamma/gamma.go
gammassp: No not always, the code will return banner default (`openrtb_ext.BidTypeBanner`) if there is no `Video` detected.

adapters/gamma/gammatest/exemplary/simple-banner.json
gammassp: Resolved as above comment.

adapters/gamma/gamma.go
hhhjort: You will only ever bid on the first impression in a request?

adapters/gamma/gamma.go
hhhjort: You only allow banner and video. So the only multiformat possibility is banner+video :) Just confirming if video is allowed, then you will definitely bid video. We have had issues with other adapters being sloppy about reporting which sort of ad they are returning, causing rendering issues.

adapters/gamma/gamma.go
gammassp: Yes, our system only process one impression per request. So if request has multiformat, the first impression will be processed.

hhhjort: You are aware that you are returning a slice of request data? You could create multiple requests, one per Imp, and then the system will call `MakeBids()` for each request sent to your bidder that produced a response, and gather all the bids together. This would allow you to bid on all the impressions even though your backend system can only handle one impression at a time.

hhhjort: I would like to confirm you only want to bid on the first imp of the impression, or are you working on a solution to submit one `RequestData` per imp so you can bid on all imps?

gammassp: We are working on a solution to bid on all imps. It takes a little changes on our system.

hhhjort: Did you understand that you could bid on all imps with just a change to the adapter, and no changes to your system? (unless you wanted to include bidding on all the imps on  set, rather than bidding on each bid in the request in isolation).

adapters/gamma/gamma.go
```diff
@@ -0,0 +1,222 @@
...
+			request.Imp[i].Banner = &bannerCopy
+			validImpExists = true
+		} else if request.Imp[i].Video != nil {
+			validImpExists = true
+		} else {
+			err := &errortypes.BadInput{
+				Message: fmt.Sprintf("Gamma only supports banner and video media types. Ignoring imp id=%s", request.Imp[i].ID),
+			}
+			errs = append(errs, err)
+			request.Imp = append(request.Imp[:i], request.Imp[i+1:]...)
```
guscarreon: Since function `append(arr[:i], arr[i+1]...)` appends the rest of the elements of the array at a cost of O(n), we can assume that the time complexity is O(n^2)
```
115     // Assume: request.Imp = []Imp{"banner","other","other","other","other","other","video","banner"}
116     validImpExists := false
117     for i := 0; i < len(request.Imp); i++ {
118         if request.Imp[i].Banner != nil {
119             bannerCopy := *request.Imp[i].Banner
120             if bannerCopy.W == nil && bannerCopy.H == nil && len(bannerCopy.Format) > 0 {
121                 firstFormat := bannerCopy.Format[0]
122                 bannerCopy.W = &(firstFormat.W)
123                 bannerCopy.H = &(firstFormat.H)
124             }
125             request.Imp[i].Banner = &bannerCopy
126             validImpExists = true
127         } else if request.Imp[i].Video != nil {
128             validImpExists = true
129         } else {
130             err := &errortypes.BadInput{
131                 Message: fmt.Sprintf("Gamma only supports banner and video media types. Ignoring imp id=%s", request.Imp[i].ID),
132             }
133             errs = append(errs, err)
134             request.Imp = append(request.Imp[:i], request.Imp[i+1:]...) //we'll have a cost of O(n) every time we find an Imp of the type "other"
135             i--                                                         //yielding a cost of O(n*n) or n^2
136         }
137     }
```
Do you think this approach can be improved upon? What about an index array, a boolean array, or, even better, a bitmap?

guscarreon: Maybe something like:
```
diff --git a/adapters/gamma/gamma.go b/adapters/gamma/gamma.go
index 0d5fade..f8389bd 100644
--- a/adapters/gamma/gamma.go
+++ b/adapters/gamma/gamma.go
@@ -113,7 +113,7 @@ func (a *GammaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapte
                errs = append(errs, err)
                return nil, errs
        }
-       validImpExists := false
+       var validImpIndex = make([]int, 0, len(request.Imp)) //len(invalidImpIndex) = 0, cap(invalidImpIndex) = 0
                if request.Imp[i].Banner != nil {
                        bannerCopy := *request.Imp[i].Banner
@@ -123,33 +123,43 @@ func (a *GammaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapte
                                bannerCopy.H = &(firstFormat.H)
                        }
                        request.Imp[i].Banner = &bannerCopy
-                       validImpExists = true
-               } else if request.Imp[i].Video != nil {
-                       validImpExists = true
-               } else {
+               } else if request.Imp[i].Video == nil {
                        err := &errortypes.BadInput{
                                Message: fmt.Sprintf("Gamma only supports banner and video media types. Ignoring imp id=%s", request.Imp[i].ID),
                        }
                        errs = append(errs, err)
-                       request.Imp = append(request.Imp[:i], request.Imp[i+1:]...)
-                       i--
+                       invalidImpIndex = append(invalidImpIndex, i)
                }
        }

-       if !validImpExists {
+       var adapterRequests []*adapters.RequestData
+       var j int = 0
+       if len(invalidImpIndex) == 0 { //Proceed as usual
+               for _, imp := range request.Imp {
+                       adapterReq, errors := a.makeRequest(request, imp)
+                       if adapterReq != nil {
+                               adapterRequests = append(adapterRequests, adapterReq)
+                       }
+                       errs = append(errs, errors...)
+               }
+       } else if len(request.Imp) == len(invalidImpIndex) { //only true if every Imp was not a Banner or a Video
                err := &errortypes.BadInput{
                        Message: fmt.Sprintf("No valid impression in the bid request"),
                }
                errs = append(errs, err)
                return nil, errs
-       }
-       var adapterRequests []*adapters.RequestData
-       for _, imp := range request.Imp {
-               adapterReq, errors := a.makeRequest(request, imp)
-               if adapterReq != nil {
-                       adapterRequests = append(adapterRequests, adapterReq)
+       } else {
+               for i := 0; i < len(request.Imp); i++ {
+                       if i == invalidImpIndex[j] {
+                               j++
+                       } else {
+                               adapterReq, errors := a.makeRequest(request, request.Imp[i])
+                               if adapterReq != nil {
+                                       adapterRequests = append(adapterRequests, adapterReq)
+                               }
+                               errs = append(errs, errors...)
+                       }
                }
-               errs = append(errs, errors...)
        }

        return adapterRequests, errs
```

adapters/gamma/gamma.go
```diff
@@ -0,0 +1,222 @@
...
+		headers.Add(headerName, headerValue)
+	}
+}
+
+// getMediaTypeForImp figures out which media type this bid is for.
+func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	mediaType := openrtb_ext.BidTypeBanner //default type
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Banner != nil {
```
mansinahar: I think this if condition can be safely removed because `BidTypeBanner` is anyway the default mediatype. For example, something like:

```
if imp.Video != nil {
}
return mediaType
```

will also yield the same result.

adapters/gamma/gamma.go
```diff
@@ -0,0 +1,222 @@
...
+				mediaType = openrtb_ext.BidTypeBanner
+			} else if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			return mediaType
+		}
+	}
+	return mediaType
```
mansinahar: If this point of execution is reached then it means that no imp was found with the given `impId`. Although probably unlikely, if it does happen, shouldn't an error be returned in that case rather than the default media type?

adapters/gamma/params_test.go
```diff
@@ -0,0 +1,60 @@
+package gamma
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// This file actually intends to test static/bidder-params/brightroll.json
```
mansinahar: Just a minor thing but can you please update this comment to point to `static/bidder-params/gamma.json`? Next line needs a similar update also

openrtb_ext/imp_gamma.go
```diff
@@ -0,0 +1,27 @@
...
+type ExtImpGamma struct {
+	PartnerID string `json:"id"`
+	ZoneID    string `json:"zid"`
+	WebID     string `json:"wid"`
+	AppID     string `json:"app_id,omitempty"`
+	AppName   string `json:"app_name,omitempty"`
+	AppBundle string `json:"app_bundle,omitempty"`
+}
+
+// id=1397808490&
```
mansinahar: Are these the values that a publisher is suppose to use in their requests or they are just sample values?

gammassp: `PartnerID`, `ZoneID` and `WebID` are required, the others are unnecessary, we've removed them.

adapters/gamma/gamma.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	if request.App != nil {
+		thisURI = thisURI + "&app_id=" + request.App.ID
+		thisURI = thisURI + "&app_bundle=" + request.App.Bundle
+		thisURI = thisURI + "&app_name=" + request.App.Name
+	}
+	headers := http.Header{}
+	headers.Add("Accept", "*/*")
+	headers.Add("x-openrtb-version", "2.5")
+	if gammaExt.PartnerID == "bad-request-id" {
```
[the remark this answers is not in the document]
gammassp: Our system always return `HTTP 200`, so we add this to bypass `HTTP 400` status test.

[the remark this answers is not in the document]
mansinahar: @gammassp sorry but I don't exactly understand. Could you maybe elaborate a little?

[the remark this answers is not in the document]
hhhjort: How does this bypass an `HTTP 400` status check, and why would such a check need to be bypassed if you always return `200`? We are at the create request stage, so you are only setting this header if the publisher has set this particular ID. And why would you need to "bypass" the `HTTP 400` check, if your status is always 200, you will never get caught by a 400 check?

adapters/gamma/gammatest/exemplary/valid-full-params.json
```diff
@@ -0,0 +1,83 @@
...
+        "ext":{
+          "bidder":{
+            "id": "sample-id",
+            "zid": "sample-zone-id",
+            "wid": "sample-web-id"
+           }
+        }
+      }
+    ]
+  },
```
guscarreon: It would be nice to also include a video imp with full params here

guscarreon:
Hey @gammassp we'd really like to merge your adapter soon. Any corrections made in the last couple of weeks? Haven't heard from you in a while.

adapters/gamma/gammatest/exemplary/banner-and-video-and-audio.json
```diff
@@ -0,0 +1,136 @@
...
+        "adid": "29484110",
+        "adomain": ["sample.com"],
+        "cid": "958",
+        "crid": "29484110",
+        "w": 1024,
+        "h": 576
+      },
+      "type": "video"
+    }
+  ]
```
mansinahar: There doesn't seem to be any difference between this and the `banner-and-video.json` file. Are they both needed?

adapters/gamma/gammatest/supplemental/bad-request.json
```diff
@@ -0,0 +1,51 @@
...
+    "app":{
+    	"id":"test-app-id",	
+    	"name":"test-app-name",
+    	"bundle":"test-app-bundle"
+    },
+
+    "device":{
+    	"ua":"test-device-ua",
+    	"ip":"test-device-ip",
+      "ifa":"test-device-ifa",
```
mansinahar: Just a nit pick but since there are other changes, can you please fix the formatting in this file? The spaces seem to be off at a couple places

mansinahar:
@gammassp pinging to see if you guys have gotten a chance to address the above comments?

adapters/gamma/gamma.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+		}
+	}
+
+	return adapterRequests, errs
+}
+
+func (a *GammaAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if response.StatusCode == http.StatusNoContent {
+
```
mansinahar: In case of `http.StatusNoContent`, I'd suggest returning `nil, nil` because it's not a `BadServerResponse` if your adapter is sending back an empty response for some reason.

mansinahar:
@gammassp LGTM in general but there are still some comments that haven't been addressed. Can you please address them when you have a chance and we can help get this merged?

PR #950, opened 2019-07-01
[Sharethrough] Add URL Scheme to Origin header

maphe:
Our ad server is expecting the scheme in the Origin header


maphe:
Sorry, pushed that too early, will revisit and push a new version next week.

PR #956, opened 2019-07-08
Adds support for parentAccount property

hhhjort:
This lets a PBS host have a "parent Account" to track the entity they do business with when `publisher.id` may be too granular.

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -148,8 +148,17 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	} else {
 		labels.CookieFlag = pbsmetrics.CookieFlagYes
 	}
-	if req.Site != nil && req.Site.Publisher != nil && req.Site.Publisher.ID != "" {
-		labels.PubID = req.Site.Publisher.ID
+	if req.Site != nil && req.Site.Publisher != nil {
+		if req.Site.Publisher.ID != "" {
+			labels.PubID = req.Site.Publisher.ID
+		}
+		var pubExt openrtb_ext.ExtPublisher
+		if req.Site.Ext != nil {
+			err := json.Unmarshal(req.Site.Ext, &pubExt)
+			if err == nil && pubExt.ParentAccount != nil {
+				labels.PubID = *pubExt.ParentAccount
+			}
+		}
 	}
 
 	response, err := deps.ex.HoldAuction(ctx, req, usersyncs, labels, &deps.categories)
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -148,18 +148,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	} else {
 		labels.CookieFlag = pbsmetrics.CookieFlagYes
 	}
-	if req.Site != nil && req.Site.Publisher != nil {
-		if req.Site.Publisher.ID != "" {
-			labels.PubID = req.Site.Publisher.ID
-		}
-		var pubExt openrtb_ext.ExtPublisher
-		if req.Site.Ext != nil {
-			err := json.Unmarshal(req.Site.Ext, &pubExt)
-			if err == nil && pubExt.ParentAccount != nil {
-				labels.PubID = *pubExt.ParentAccount
-			}
-		}
-	}
+	labels.PubID = effectivePubID(req.Site.Publisher)
 
 	response, err := deps.ex.HoldAuction(ctx, req, usersyncs, labels, &deps.categories)
 	ao.AuctionResponse = response
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1184,17 +1184,10 @@ func effectivePubID(pub *openrtb.Publisher) string {
 			err := json.Unmarshal(pub.Ext, &pubExt)
 			if err == nil && pubExt.ParentAccount != nil && *pubExt.ParentAccount != "" {
 				return *pubExt.ParentAccount
-			} else {
-				if pub.ID != "" {
-					return pub.ID
-				} else {
-					return pbsmetrics.PublisherUnknown
-				}
 			}
-		} else {
-			if pub.ID != "" {
-				return pub.ID
-			}
+		}
+		if pub.ID != "" {
+			return pub.ID
 		}
 	}
 	return pbsmetrics.PublisherUnknown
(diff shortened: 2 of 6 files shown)


endpoints/openrtb2/amp_auction.go
```diff
@@ -148,8 +148,17 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
+		if req.Site.Publisher.ID != "" {
+			labels.PubID = req.Site.Publisher.ID
+		}
+		var pubExt openrtb_ext.ExtPublisher
+		if req.Site.Ext != nil {
+			err := json.Unmarshal(req.Site.Ext, &pubExt)
+			if err == nil && pubExt.ParentAccount != nil {
+				labels.PubID = *pubExt.ParentAccount
+			}
+		}
```
josephveach: It seems that in almost every changed file in this request, the ordering of checks can be changed in order to make the code more efficient.  For example, in the above  code, we go to the trouble of checking req.Site.Publisher and req.Site.Publisher.ID and assigning labels.PubID only to reassign it later if req.Site.Ext can be unmarshalled to successfully obtain ParentAccount.  It would be more efficient to first try to get PubID from req.Site.Ext and look at req.Site.Publisher only if necessary.

It's not clear to me whether we want to not use req.Site.Ext in the case that req.Site.Publisher is nil.  I would guess that's not the case  (i.e., if req.Site.Publisher is nil and we can get PubID from req.Site.Ext, I expect we want to get it), though that is what the code currently does.  If I'm wrong, that would slightly alter my above comment.

endpoints/openrtb2/auction.go
```diff
@@ -127,8 +127,17 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+			if req.App.Publisher.ID != "" {
+				labels.PubID = req.App.Publisher.ID
+			}
+			var pubExt openrtb_ext.ExtPublisher
+			if req.App.Ext != nil {
+				err := json.Unmarshal(req.App.Ext, &pubExt)
+				if err == nil && pubExt.ParentAccount != nil {
+					labels.PubID = *pubExt.ParentAccount
+				}
+			}
```
josephveach: In the above, if we can get PubID from req.App.Ext, we don't need to examine req.App.Publisher or anything beneath it at all.

As above, one thing not clear to me is whether it is intentional that req.App.Publisher not be nil before choosing to examine req.App.Ext.  As written, even if req.App.Ext can produce a value for PubID, it would not do so unless req.App.Publisher is non-nil.  My comment in the preceding paragraph assumes we would want to assign PubID from req.App.Ext even if req.App.Publisher is nil.

endpoints/openrtb2/auction.go
```diff
@@ -137,8 +146,17 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+			if req.Site.Publisher.ID != "" {
+				labels.PubID = req.Site.Publisher.ID
+			}
+			var pubExt openrtb_ext.ExtPublisher
+			if req.Site.Ext != nil {
+				err := json.Unmarshal(req.Site.Ext, &pubExt)
+				if err == nil && pubExt.ParentAccount != nil {
+					labels.PubID = *pubExt.ParentAccount
+				}
+			}
```
josephveach: Approximately the same comments here as above for lines 130-140.

endpoints/openrtb2/video_auction.go
```diff
@@ -190,8 +190,17 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+			if bidReq.App.Publisher.ID != "" {
+				labels.PubID = bidReq.App.Publisher.ID
+			}
+			var pubExt openrtb_ext.ExtPublisher
+			if bidReq.App.Ext != nil {
+				err := json.Unmarshal(bidReq.App.Ext, &pubExt)
+				if err == nil && pubExt.ParentAccount != nil {
+					labels.PubID = *pubExt.ParentAccount
+				}
+			}
```
josephveach: Same comment here as for both snippets in auction.go.

endpoints/openrtb2/video_auction.go
```diff
@@ -200,8 +209,17 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+			if bidReq.Site.Publisher.ID != "" {
+				labels.PubID = bidReq.Site.Publisher.ID
+			}
+			var pubExt openrtb_ext.ExtPublisher
+			if bidReq.Site.Ext != nil {
+				err := json.Unmarshal(bidReq.Site.Ext, &pubExt)
+				if err == nil && pubExt.ParentAccount != nil {
+					labels.PubID = *pubExt.ParentAccount
+				}
+			}
```
josephveach: Again, same comment as for lines 193-203 and for auction.go.

exchange/utils.go
```diff
@@ -46,13 +46,25 @@ func cleanOpenRTBRequests(ctx context.Context,
...
+			if bidReq.App != nil && bidReq.App.Publisher != nil && publisherID == "" {
 				publisherID = bidReq.App.Publisher.ID
-			} else {
-				publisherID = ""
+				if bidReq.App.Ext != nil {
+					err := json.Unmarshal(bidReq.App.Ext, &pubExt)
+					if err == nil && pubExt.ParentAccount != nil {
+						publisherID = *pubExt.ParentAccount
+					}
+				}
```
josephveach: Similar comment here as to those for similar changes above.

endpoints/openrtb2/amp_auction.go
hhhjort: Oh, you caught a bug. It should be `req.Site.Publisher.Ext`. I am a little up in the air if grabbing `PubID` from `pubExt` first and then checking if we need to try `Publisher.ID` will save anything. It saves an assignment from one variable to another, but adds a check that will always be needed (Did we successfully populate from `pubExt`?). I suppose we could duplicate the code to appear on both `else` branches of the decision tree to avoid the check.

endpoints/openrtb2/auction.go
```diff
@@ -1179,3 +1175,31 @@ func fatalError(errL []error) bool {
...
+				return pub.ID
+			} else {
+				return pbsmetrics.PublisherUnknown
+			}
+		}
+	} else {
+		return pbsmetrics.PublisherUnknown
+	}
+
+}
```
josephveach: Couldn't this be:

```
func effectivePubID(pub *openrtb.Publisher) string {
        if pub != nil {
                if pub.Ext != nil {
                        var pubExt openrtb_ext.ExtPublisher
                        err := json.Unmarshal(pub.Ext, &pubExt)
                        if err == nil && pubExt.ParentAccount != nil && *pubExt.ParentAccount != "" {
                                return *pubExt.ParentAccount
                        }
                }
                if pub.ID != "" {
                        return pub.ID
                }
        }
        return pbsmetrics.PublisherUnknown
}

```

hhhjort: You are correct. :)

bretg:
For the record, these are optionally utilized by adapters:

- {site,app}.publisher.ext.id.parentaccount represents the holding company. e.g. microsoft
- {site,app}.publisher.id would be the lower level company. e.g. xbox

PR #957, opened 2019-07-09
Tappx Bidder

ah-tappx:
Bidder adapter for Tappx SSP.
Since only InApp traffic is accepted for the moment, no user sync classes have been implemented


ah-tappx:
Hello,
We see that there is some issue in the Travis build. Is there something we can do to fix the error? The bidder should work well since we have run many local tests and everything was fine. Regarding the unit test coverage, we got a coverage of 82.3%, remaining tests are related to request JSON body formatting and we are not sure how we should test this exactly.

mansinahar:
Looks like Travis is not failing because of unit tests. There seems to be some Go formatting violations. Please run `go fmt` on `adapters/tappx/tappx.go` and update the PR. That should resolve the issue.

ah-tappx:
Hello again. Thank you for your answer. We see that the build is now failing because of the lack of usersync implementation. Our SSP only supports InApp traffic at this moment and we are not using cookiesync since it only applies to web traffic, so we haven't implemented it in the prebid adapter. Is it mandatory? How should we proceed?

mansinahar:
Oh! I see. That sounds like a reasonable use case. It looks like the test cares if the usersync interface is not implemented but I believe that the code itself doesn't. I have had a discussion about this just last week with @hhhjort and we concluded that although it is good to have an exception of sorts to flag a missing usersync for an adapter, it shouldn't completely fail the test. We can probably add a list of adapters that consciously decide to not implement a syncer and can therefore be ignored in that test.

Would you guys be willing to modify that test as part of this PR?

ah-tappx:
Sure, just give us some guides regarding what should we modify. Do you think it will take long to add the changes in the prebid core? The integration is quite urgent for us so if it's faster to just implement usersync our team will discuss how to proceed.

mansinahar:
So, if you look at this line: https://github.com/prebid/prebid-server/blob/master/usersync/usersyncers/syncer_test.go#L57, the for loop loops over the entire BidderMap which is essentially a map of all adapters/bidders supported by Prebid Server and this line: https://github.com/prebid/prebid-server/blob/master/usersync/usersyncers/syncer_test.go#L16 specifies all the adapters/bidders for which syncers are expected to be present. 

One possible solution would be to also check in this if condition: https://github.com/prebid/prebid-server/blob/master/usersync/usersyncers/syncer_test.go#L58 the particular bidder/adapter under check is also part of the `cfg.Adapters` map.

Essentially, something like this:

```
for _, bidderName := range openrtb_ext.BidderMap {
	_, adapterWithSyncer := cfg.Adapters[string(bidderName)]
	if _, ok := syncers[bidderName]; !ok && adapterWithSyncer {
		t.Errorf("No syncer exists for adapter: %s", bidderName)
	}
}
```

ah-tappx:
Thank you for your answer, it solved the issue with the syncer tests. Is there anything else we need to do in order to have our bidder integrated?

mansinahar:
As of now, the PR just needs to be reviewed. Once reviewed, if there's any feedback, you'd have to just address that and then we can get it merged. Our standard policy is to have at least 2 approvals before a PR can get merged.

ah-tappx:
Ok thank you. Is there an estimation of when can we expect the review? One of our partners will switch to prebid in very short time and the time window is pretty tight for us.

adapters/tappx/tappx.go
```diff
@@ -0,0 +1,207 @@
...
+			Message: "No impression in the bid request",
+		}}
+	}
+
+	errs := make([]error, 0, len(request.Imp))
+
+	var bidderExt adapters.ExtImpBidder
+	if err := json.Unmarshal(request.Imp[0].Ext, &bidderExt); err != nil { // 3
+		errs = append(errs, err)
+		return nil, errs
```
guscarreon: This is a minor thing but if we exit the function whenever we find an error as we do in lines 60 and 84, maybe we don't need the `errs` local variable declared in line 65 at all. We can exit the function, say in line 70 like:
```
 70         return nil, []error{err}
```

adapters/tappx/tappx.go
```diff
@@ -0,0 +1,207 @@
...
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	return []*adapters.RequestData{{
+		Method:  "POST",
+		Uri:     url,
+		Body:    reqJSON,
+		Headers: headers,
+	}}, errs
```
guscarreon: If we get rid of the `errs` local variable declared in line 65 maybe we could:
```
 98     return []*adapters.RequestData{{
 99         Method:  "POST",
100         Uri:     url,
101         Body:    reqJSON,
102         Headers: headers,
103     }}, []error{}
104 }
```

adapters/tappx/tappx.go
```diff
@@ -0,0 +1,207 @@
...
+func (a *TappxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	if len(request.Imp) == 0 {
+		return nil, []error{&errortypes.BadInput{
+			Message: "No impression in the bid request",
+		}}
+	}
+
+	errs := make([]error, 0, len(request.Imp))
+
+	var bidderExt adapters.ExtImpBidder
```
guscarreon: Are we sure the `request.Imp[]` array will always come with one element? If not, do we only care for the first one? Is there any possibility of a second or third `Imp[]` element to be relevant?

adapters/tappx/tappx.go
```diff
@@ -0,0 +1,207 @@
...
+
+	var test int
+	test = int(request.Test)
+
+	url, err := a.buildEndpointURL(&tappxExt, test)
+	if url == "" {
+		return nil, []error{err}
+	}
+
+	//fmt.Println(url)
```
guscarreon: Please remove

adapters/tappx/tappx.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	var bidResp openrtb.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(5)
+
+	var errs []error
```
guscarreon: This local variable is not really necessary since the array is not used in any way in the subsequent nested for loop. 
```
182     var errs []error 
```

adapters/tappx/tappx.go
ah-tappx: Most of our traffic comes with only one impression and we only validate the first Imp object. If that first element contains some error we will discard the request. If many Imp objects are sent and there is some error in the other objects we handle this directly in our SSP.

ah-tappx:
Hello again. We have fixed all requested changes. Regarding the impression array length issue, most of our traffic comes with only one impression and we only validate the first Imp object. If that first element contains some error we will discard the request. If many Imp objects are sent and there is some error in the other objects we handle this directly in our SSP.

ah-tappx:
Hello. Do you know when can we expect this to be merged into the project?
Thank you

adapters/tappx/tappx.go
```diff
@@ -0,0 +1,197 @@
...
+		reqHost = params.Host
+	}
+	if params.Endpoint != "" {
+		reqEndpoint = params.Endpoint
+	}
+	if params.TappxKey != "" {
+		reqKey = params.TappxKey
+	}
+
+	if reqHost == "" {
```
hhhjort: This could be combined with the `if params.Host != ""` as an else clause

hhhjort:
Should probably merge in today. I did leave a comment about a slight optimization you could do, but its not enough to hold up merging.

ah-tappx:
Thank you for your comment. We'll leave this optimization for the next release.
Once the code has been merged, are there any more steps we need to follow regarding the integration? Do we need to contact someone in order to appear in this list http://prebid.org/dev-docs/prebid-server-bidders.html ?

hhhjort:
I believe that URL is autoupdated once a new version is push lived to the appnexus hosted instance of PBS. So there shouldn't be anything else you need to do.

bretg:
@ah-tappx - please open a new PR with a doc in https://github.com/prebid/prebid-server/tree/master/docs/bidders -- create a tappx.md file that provides a maintenance email address and test parameters.

ah-tappx:
Hello Bret,
We have opened a new pull request https://github.com/prebid/prebid-server/pull/970
Thank you

PR #958, opened 2019-07-10
[Sharethrough] Improved Headers in ad server request

maphe:
This is forwarding more data to the Sharethrough ad server in order to bid more accurately:
- add scheme to `Origin` header
- add `X-Forwarded-For` header
- add `User-Agent` header

### DIFF
--- a/adapters/sharethrough/butler.go
+++ b/adapters/sharethrough/butler.go
@@ -55,6 +55,8 @@ func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openr
 	headers.Add("Content-Type", "text/plain;charset=utf-8")
 	headers.Add("Accept", "application/json")
 	headers.Add("Origin", domain)
+	headers.Add("X-Forwarded-For", request.Device.IP)
+	headers.Add("User-Agent", request.Device.UA)
 
 	var strImpExt adapters.ExtImpBidder
 	if err := json.Unmarshal(imp.Ext, &strImpExt); err != nil {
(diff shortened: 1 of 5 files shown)


adapters/sharethrough/butler.go
```diff
@@ -55,6 +55,8 @@ func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openr
 	headers.Add("Content-Type", "text/plain;charset=utf-8")
 	headers.Add("Accept", "application/json")
 	headers.Add("Origin", domain)
+	headers.Add("X-Forwarded-For", request.Device.IP)
```
hhhjort: What will happen here if `request.Device == nil`?

maphe: Is that possible? in which cases? 🤔 the adapter already relies on the `Device` even prior to this PR, eg. parsing the User Agent

hhhjort: Went back into the code and saw that we do try setting `Device` implicitly if it does not exist. However if we cannot deduce the IP nor the user agent from a request, and `device` is omitted from the request, then `Device` will be `nil`. I am not sure it is possible to fail to pull the IP from an http request, if all else fails we use `net.SplitHostPort(r.RemoteAddr)`, as long as this does not return `""` for the IP then `Device` will exist.

So we are probably actually good here.

maphe: Cool, thanks for sharing context too 👍

mansinahar:
@maphe Just FYI, looks like you force pushed on your branch and now this PR doesn't have any files changed but still has merge conflicts.. lol. Not sure how that happened.

maphe:
@mansinahar thanks for the heads up, I'm investigating internally to figure out what happened, I'll restore the PR as it should be as quickly as possible

maphe:
@mansinahar we're back in good shape, there was some tool that got tested internally that messed up the branch. The PR is back to normal.

PR #960, opened 2019-07-11
Feature/onemobile adapter

oath-jac:


config/config_test.go
```diff
@@ -14,6 +14,8 @@ import (
 func TestDefaults(t *testing.T) {
 	v := viper.New()
 	SetupViper(v, "")
+	v.SetConfigType("yaml")
+	v.ReadConfig(bytes.NewBuffer(requiredConfig))
 	cfg, err := New(v)
```
guscarreon: Please rollback the modifications made to this file. If you add the required  entries to `config/config.go` discussed in our comment above, these changes are unnecessary

endpoints/auction_test.go
```diff
@@ -274,6 +274,13 @@ func DummyPrebidCacheServer(w http.ResponseWriter, r *http.Request) {
 	w.Write(b)
```
guscarreon: Please rollback the modifications made to this file. If you add the required  entries to `config/config.go`discussed in our comment above, these changes are unnecessary

exchange/adapter_map.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 
 	"github.com/prebid/prebid-server/adapters"
-	ttx "github.com/prebid/prebid-server/adapters/33across"
+	"github.com/prebid/prebid-server/adapters/33across"
 	"github.com/prebid/prebid-server/adapters/adform"
```
guscarreon: Please precede with `ttx` the contents of line 9 as was before.

static/bidder-info/onemobile.yaml
```diff
@@ -0,0 +1,6 @@
+maintainer:
+  email: "hb-fe-tech@verizonmedia.com"
+capabilities:
+  site:
+    mediaTypes:
+      - banner
```
guscarreon: The yaml file is only listing `banner` in the mediatypes section but `adapters/onemobile/onemobile.go` seems to can also handle video, isn't it?
```
152 func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
153     mediaType := openrtb_ext.BidTypeBanner
154     for _, imp := range imps {
155         if imp.ID == impId {
156             if imp.Video != nil {
157                 mediaType = openrtb_ext.BidTypeVideo <==
158             }
159             return mediaType
160         }
161     }
162     return mediaType
163 }
```

config/config_test.go
radubarbos: Hi, we do not have default endpoints. All our endpoints are unique per publisher, the bidding and the user sync endpoint. I had to change this in order to pass because of our context. Do you mean that is required that we have to set some default endpoints?
If yes:
- we can set dummy urls that does nothing
- we can set a base url like https://verizonmedia.com and in the config processing to append some publisher specific URL paths. 
Which one will be accepted?

guscarreon: Good day. If there's no default endpoint we can do 
```
v.SetDefault("adapters.foo.disabled", true)
```
Instead of 
```
v.SetDefault("adapters.foo.endpoint", URL2)
```
Inside the `func SetupViper(v *viper.Viper, filename string)` function of `config/config.go`. Please let me know if this works.

hhhjort: The issue here is if the URL is dynamic, based on the content of the openrtb request, configure a base URL and modify it in your adapter to be the final URL. If the URL depends on the PBS host (the PBS hosting entity should enter something unique in the config to get your adapter working properly), then the adapter should default to disabled. The idea is that PBS should come up in a workable state even if not supplied with any configuration.

oath-jac: I see, the URL is something the publisher will have to setup in his config. I'll flag it disabled as suggested because  without this new adapter config properties the server does not start. Just as a notification I would like to say that we got through rebranding and we'll change the name of the adapter from 'onemobile' to 'verizonmedia'.

endpoints/auction_test.go
oath-jac: I'll revert the suggested files, thanks

static/bidder-info/onemobile.yaml
oath-jac: We don't have video support, at least not for now, this check was added as defense check so the media type is reported correctly based on the response instead of statically set it  to banner type. If you think that this might break the capabilities we reported we'll remove the method.

adapters/onemobile/onemobile.go
```diff
@@ -0,0 +1,189 @@
...
+	return "onemobile"
+}
+
+func (a *OneMobileAdapter) SkipNoCookies() bool {
+	return false
+}
+
+func (a *OneMobileAdapter) MakeRequests(requestIn *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+
+	request := *requestIn
```
hhhjort: The `requestIn` you are given is already a shallow copy of the original request, unique to your adapter. I don't think you need to make another copy. Along with that, the `Imp`s are also all shallow copies unique to your adapter. You still do need to copy the `Site` object as you do below.

Also seems a bit unclear and arbitrary where you use `requestIn` and `request` below, so dropping the copy might help clear that up.

adapters/onemobile/onemobile.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Bad server response: %d. ", err),
+		}}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(1)
+
+	for _, sb := range bidResp.SeatBid {
+		for _, bid := range sb.Bid {
+			impID := lookupImpID(bid.ImpID, internalRequest.Imp)
```
hhhjort: Could this be improved by having a function that just returns a boolean testing if an imp ID can be found in a slice of imps?

adapters/onemobile/onemobile.go
```diff
@@ -0,0 +1,189 @@
...
+		request.Site.ID = extension.Dcn
+	}
+}
+
+func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	mediaType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
```
hhhjort: Is your bidder guaranteed to bid a video imp (or not bid at all) for multiformat requests that include video as a format?

adapters/onemobile/onemobile.go
oath-jac: I'll replace lookImpID with an impIdExists that returns a boolean

adapters/onemobile/onemobile.go
oath-jac: We don't support video, this was a defense check that we indeed check the upstream response media type, but I'll statically set it to banner

adapters/onemobile/onemobile.go
oath-jac: OK, if is a copy already I'll clean it.

config/config_test.go
hhhjort: Just one note on terminology to help avoid confusion. "publisher" would be the actual source of ad traffic, using prebid to increase revenue. "PBS host" would be the entity running the instance of prebid server. They can both be the same thing, but won't always be the same.

static/bidder-info/onemobile.yaml
guscarreon: Prebid server supports `Banner`, `Video`, `Audio`, and `Native` as media types. If your system only supports `Banner` maybe a fair thing to do is to add a check to filter out all other media types just in case.

oath-jac:
Hi, could I get a new review check please.

adapters/verizonmedia/verizonmedia.go
```diff
@@ -0,0 +1,189 @@
...
+	errs := make([]error, 0, len(request.Imp))
+	if len(request.Imp) == 0 {
+		err := &errortypes.BadInput{
+			Message: "No impression in the bid request",
+		}
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	errors := make([]error, 0, 1)
```
guscarreon: Why declare `errs` and then `errors`? I believe we only need one.
```
func (a *OneMobileAdapter) MakeRequests(requestIn *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {

	request := *requestIn
	if len(request.Imp) == 0 {
		err := &errortypes.BadInput{
			Message: "No impression in the bid request",
		}
		errs = append(errs, err)
		return nil, errs
	}

	errors := make([]error, 0, 1)
```

adapters/verizonmedia/verizonmedia.go
```diff
@@ -0,0 +1,189 @@
...
+
+	var bidderExt adapters.ExtImpBidder
+	err := json.Unmarshal(request.Imp[0].Ext, &bidderExt)
```
guscarreon: Are we only concerned about the first `Imp`? Other adapters are "multi-imp" and loop through the `Imp` array to verify for relevant `Imp`s. Is that relevant in our context?
```
 26 func (a *VerizonMediaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.R>
 27 +-- 12 lines: errs := make([]error, 0, len(request.Imp))---------
 39     var bidderExt adapters.ExtImpBidder
 40     err := json.Unmarshal(request.Imp[0].Ext, &bidderExt)
 41     if err != nil {
 42         err = &errortypes.BadInput{
 43             Message: "ext.bidder not provided",
 44         }
 45         errors = append(errors, err)
 46         return nil, errors
 47     }
 48     var verizonMediaExt openrtb_ext.ExtImpVerizonMedia
 49     err = json.Unmarshal(bidderExt.Bidder, &verizonMediaExt)
 50     if err != nil {
```

adapters/verizonmedia/verizonmedia.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Bad server response: %d. ", err),
+		}}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(1)
+
+	for _, sb := range bidResp.SeatBid {
+		for _, bid := range sb.Bid {
+			if !impIDExists(bid.ImpID, internalRequest.Imp) {
```
guscarreon: Do we need a bigger initial size for our bids array? After all, if we expect to have more than one Bid most of the times maybe we want to allocate a buffer that makes sense in our expected bid number context so we don't have to dynamically allocate elements on the heap via the `append` function.
```
 99 func (a *VerizonMediaAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
100 +-- 17 lines: if response.StatusCode == http.StatusNoContent {----------------------------------
117
118     bidResponse := adapters.NewBidderResponseWithBidsCapacity(1)
                       --------------------------------------
                        63 func NewBidderResponseWithBidsCapacity(bidsCapacity int) *BidderResponse {
                        64     return &BidderResponse{
                        65         Currency: "USD",
                        66         Bids:     make([]*TypedBid, 0, bidsCapacity),
                        67     }
                        68 }
                       adapters/bidder.go
                       --------------------------------------
119
120     for _, sb := range bidResp.SeatBid {
121         for _, bid := range sb.Bid 
122             if !impIDExists(bid.ImpID, internalRequest.Imp) {
adapters/verizonmedia/verizonmedia.go
```

adapters/verizonmedia/verizonmedia.go
```diff
@@ -0,0 +1,189 @@
...
+				return nil, []error{&errortypes.BadServerResponse{
+					Message: fmt.Sprintf("Unknown ad unit code '%s'", bid.ImpID),
+				}}
+			}
+
+			if openrtb_ext.BidTypeBanner != getMediaTypeForImp(bid.ImpID, internalRequest.Imp) {
```
guscarreon: Both `getMediaTypeForImp` and `impIDExists` are traversing all the `Imp` array adding overhead over repetitivenes. Can we consolidate them in one function so we only traverses the `Imp` array once?
```
 99 func (a *VerizonMediaAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
100 +-- 20 lines: if response.StatusCode == http.StatusNoContent {-----------
120     for _, sb := range bidResp.SeatBid {
121         for _, bid := range sb.Bid {
122             if !impIDExists(bid.ImpID, internalRequest.Imp) {
123                 return nil, []error{&errortypes.BadServerResponse{
124                     Message: fmt.Sprintf("Unknown ad unit code '%s'", bid.ImpID),
125                 }}
126             }
127
128             if openrtb_ext.BidTypeBanner != getMediaTypeForImp(bid.ImpID, internalRequest.Imp) {
129                 //only banner is supported, anything else is ignored
130                 continue
131             }
132
 .
 .
 .
143 func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
144     var mediaType openrtb_ext.BidType
145     for _, imp := range imps {
146         if imp.ID == impId {
147             if imp.Banner != nil {
148                 mediaType = openrtb_ext.BidTypeBanner
149             }
150             return mediaType
151         }
152     }
153     return mediaType
154 }
 .
 .
 .
165 func impIDExists(impId string, imps []openrtb.Imp) bool {
166     for _, imp := range imps {
167         if imp.ID == impId {
168             return true
169         }
170     }
171     return false
172 }
```

adapters/verizonmedia/verizonmedia.go
oath-jac: Yes, one is enough, it's fixed in the next update.

adapters/verizonmedia/verizonmedia.go
oath-jac: I combined them in 1 function. I'll be available in the next PR.

adapters/verizonmedia/verizonmedia.go
hhhjort: Going with the number of Imps in the request may not be a bad idea.

adapters/verizonmedia/verizonmedia.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return bidResponse, nil
+}
+
+func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	var mediaType openrtb_ext.BidType
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Banner != nil {
+				mediaType = openrtb_ext.BidTypeBanner
```
hhhjort: You could just return `openrtb_ext.BidTypeBanner` here and avoid creating the `mediaType` variable.

adapters/verizonmedia/verizonmedia.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+			return mediaType
+		}
+	}
+	return mediaType
```
hhhjort: And then just return `nil` here, that would give the same result as you currently have.

usersync/usersyncers/syncer.go
```diff
@@ -1,11 +1,12 @@
 package usersyncers
 
 import (
+	"github.com/prebid/prebid-server/adapters/verizonmedia"
 	"strings"
 	"text/template"
 
 	"github.com/golang/glog"
-	ttx "github.com/prebid/prebid-server/adapters/33across"
```
hhhjort: This still needs to be restored

adapters/verizonmedia/verizonmedia.go
oath-jac: I improved the coverage to around 90%

adapters/verizonmedia/verizonmedia.go
oath-jac: We are concerned only about the first entry. The back-end supports only 1 entry , it processes only the first imp of the array. I've asked for a double confirmation on this.

adapters/verizonmedia/verizonmedia.go
oath-jac: I used the len of the request.

adapters/verizonmedia/verizonmedia.go
oath-jac: Not in scope anymore, after added the change suggested by guscarreon

adapters/verizonmedia/verizonmedia.go
oath-jac: Not in scope anymore as the function is changed

usersync/usersyncers/syncer.go
oath-jac: Fixed

adapters/verizonmedia/verizonmedia.go
oath-jac: I got the confirmation again that the back-end only uses the first entry of the Imp array, it supports only 1 Imp per BidRequest

oath-jac:
Could we get a new check, please

adapters/verizonmedia/verizonmedia.go
```diff
@@ -0,0 +1,181 @@
...
+
+	var bidResp openrtb.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Bad server response: %d.", err),
+		}}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(internalRequest.Imp))
+
```
guscarreon: Some other adapters throw an error if the `SeatBid` array is empty. Is this desirable in our code?
```
--- a/adapters/verizonmedia/verizonmedia.go
+++ b/adapters/verizonmedia/verizonmedia.go
@@ -115,6 +115,11 @@ func (a *VerizonMediaAdapter) MakeBids(internalRequest *openrtb.BidRequest, exte

        bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(internalRequest.Imp))

+       if len(bidResp.SeatBid) < 1 {
+               return nil, []error{
+                       newBadServerResponseError(fmt.Sprintf("Invalid SeatBids count: %d", len(bidResp.SeatBid))),
+               }
+       }
                for _, bid := range sb.Bid {
                        exists, mediaTypeId := getImpInfo(bid.ImpID, internalRequest.Imp)
```

oath-jac: Let me ask. Few months ago I was told that for a 200 response we'll always get an ad, even if is the default ad.

oath-jac: The new API responds with an empty "seatbid" array in case the inventory is not found.

oath-jac: We added this extra check. Even if the exception message is not really helpful, empty seadbid could happen because of wrong values for the ext params this way is more visible that our bank-end could not handled it.

oath-jac:
@hhhjort Could you approve ti again, thanks. I added an extra check on the SeatBid length discussed by @guscarreon

oath-jac:
Thank you for all the suggestions.

PR #962, opened 2019-07-12
Documenting the Appnexus test placement

hhhjort:
First pass at providing an example for other adapters to follow.

### DIFF
--- a/docs/bidders/appnexus.md
+++ b/docs/bidders/appnexus.md
@@ -21,8 +21,8 @@ the empty `displaymanagerver` before sending the request to AppNexus.
 ## Test Request
 
 The following test parameters can be used to verify that Prebid Server is working properly with the 
-Appnexus adapter. This example includes a sample `imp` object as the test creative must fit the size(s)
-allowed in the test impression.
+Appnexus adapter. This example includes an `imp` object with an Appnexus test placement ID and sizes
+that would match with the test creative.
 
 ```
 	"imp": [{
--- a/docs/bidders/appnexus.md
+++ b/docs/bidders/appnexus.md
@@ -20,7 +20,7 @@ the empty `displaymanagerver` before sending the request to AppNexus.
 
 ## Test Request
 
-The following test parameters can be used to verify that Prebid Server is working propely with the 
+The following test parameters can be used to verify that Prebid Server is working properly with the 
 Appnexus adapter. This example includes the `imp` object as the test creative must be valid for
 serving into the test impression.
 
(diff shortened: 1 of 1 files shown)


docs/bidders/appnexus.md
```diff
@@ -17,3 +17,29 @@ is not supplied for an `imp`, but `request.app.ext.prebid.source`
 and `request.app.ext.prebid.version` are supplied, the adapter will fill in a value for
 `diplaymanagerver`. It will concatonate the two `app` fields as `<source>-<version>` fo fill in
 the empty `displaymanagerver` before sending the request to AppNexus.
+
+## Test Request
+
+The following test parameters can be used to verify that Prebid Server is working propely with the
```
mansinahar: There's a typo in this sentence with the word `properly`. PR LGTM otherwise 👍

docs/bidders/appnexus.md
```diff
@@ -17,3 +17,29 @@ is not supplied for an `imp`, but `request.app.ext.prebid.source`
 and `request.app.ext.prebid.version` are supplied, the adapter will fill in a value for
 `diplaymanagerver`. It will concatonate the two `app` fields as `<source>-<version>` fo fill in
 the empty `displaymanagerver` before sending the request to AppNexus.
+
+## Test Request
+
+The following test parameters can be used to verify that Prebid Server is working propely with the 
+Appnexus adapter. This example includes the `imp` object as the test creative must be valid for
+serving into the test impression.
```
guscarreon: I find this sentence a bit confusing. Is it missing a coma?
```
This example includes the `imp` object as the test creative must be valid for serving into the test impression.
```
Or did we mean: 
This example includes the `imp` object as the test creative **and it** must be valid for serving into the test impression.

hhhjort: No, your sentence says the imp is the test creative, which is not so. We include the imp object so someone using this as a test can create an imp object that the creative can deliver in to.

mansinahar:
Love the typo fix commit message @hhhjort :p Also, I wouldn't completely disagree with Gus here. I had to read that sentence a couple times too before I could properly understand it. Maybe we can try to simplify it a little bit if possible?

hhhjort:
Yes, it could probably be better. Haven't come up with an improvement yet, I am open to suggestions.

mansinahar:
Thoughts on something along the lines of:

```
The following test parameters can be used to verify that Prebid Server is working properly with the 
Appnexus adapter. This example includes an `imp` object with an Appnexus test placement ID and sizes that would match with the test creative.
```

PR #965, opened 2019-07-17
Improved mixed media responses (related to Issue 862)

muncha:
Ii does not appear that beachfront specificity had a problem with returning the correct mediaTypes, but there were several problems with handling requests containing both video and banner media in a single imp. The default behavior was to return a video if there was video included in the request. There was also a problem with a cookie not getting sent with that video request which resulted in very low to no fill for video. 

Processing the multimedia request as a single request to Beachfront's servers is not possible as they have separate endpoints and contracts for banner vs. video. So what I have here is a compromise that adds the "forceBanner" flag. This along with an alias (and fixing the cookie bit) lets a request similar to the example in #862  work. This is using test AppIds for banner and video. A real request would have separate AppIds configured for banner vs. video on Beachfront's platform.

`{"id":"fe7e9b35-4ae7-495b-834f-2e4a877271d9","source":{"tid":"fe7e9b35-4ae7-495b-834f-2e4a877271d9"},"tmax":1800,"imp":[{"id":"AdThrive_Content_1_phone","ext":{"beachfront":{"appId":"3b16770b-17af-4d22-daff-9606bdf2c9c3","bidfloor":0.1,"forceBanner":true},"beachfrontVideo":{"appId":"11bc5dd5-7421-4dd8-c926-40fa653bec76","bidfloor":0.1},"brightroll":{"publisher":"adthrive"},"districtm":{"use_pmt_rule":false,"placement_id":13239211,"allow_smaller_sizes":true}},"banner":{"format":[{"w":300,"h":250},{"w":320,"h":50},{"w":336,"h":280},{"w":300,"h":50},{"w":320,"h":100},{"w":250,"h":250},{"w":120,"h":240},{"w":1,"h":1},{"w":300,"h":300}]},"video":{"context":"outstream","mimes":["video/mp4"]}}],"test":1,"site":{"publisher":{"id":"72409f4d-9858-41ac-bf45-be255b82de89"},"page":"https://damndelicious.net/2014/04/11/garlic-butter-shrimp/"},"device":{"w":425,"h":976},"ext":{"prebid":{"aliases":{"districtm":"appnexus","beachfrontVideo":"beachfront"}}}}`

I've also updated the documentation with examples of multi media requests.

### DIFF
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -402,6 +402,7 @@ func (a *BeachfrontAdapter) MakeBids(internalRequest *openrtb.BidRequest, extern
 	var bids []openrtb.Bid
 	var bidtype = getBidType(internalRequest)
 
+	println("got here.")
 	/*
 		Beachfront is now sending an empty array and 200 as their "no results" response. This should catch that.
 	*/
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/golang/glog"
 	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/adapters"
 	"github.com/prebid/prebid-server/errortypes"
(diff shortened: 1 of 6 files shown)


hhhjort:
Oh, is there a reason the adapter won't simply make two calls for a multiformat request, rather than requiring an alias and the force flag?

muncha:
@hhhjort @mansinahar 
> Oh, is there a reason the adapter won't simply make two calls for a multiformat request, rather than requiring an alias and the force flag?

I guess I just assumed that MakeRequests was only supposed to make one request, that to do otherwise was not playing fair with the project and hogging network calls from other bidders. But really - look at that - the function is even named MakeRequests, not MakeRequest. Looking through the other bidders, I see Adtelligent making arbitrary numbers of requests, as is appNexus, and that just gets me through the A's.

So, answer: my own assumptions getting in the way.

I'm going to close this and get it to handle multiformat correctly. Thank you.

Issue #967, opened 2019-07-19
Rubicon adapter sends invalid request if Target is empty


potatogopher:
Rubicon adapter is setting `Target json.RawMessage` to an empty string if it was not provided. This results in an invalid request being sent to rubicon.

potatogopher:
resolved

PR #968, opened 2019-07-19
Omit empty on 'Target json.RawMessage' to avoid sending invalid json in bid request

potatogopher:
If a target was not provided for the Rubicon adapter then it's value is set to an empty string which is invalid json. Resulting in an invalid request. Omitting the field if it's empty fixes this issue. 

#967

### DIFF
--- a/adapters/rubicon/rubicon.go
+++ b/adapters/rubicon/rubicon.go
@@ -53,7 +53,7 @@ type rubiconImpExtRPTrack struct {
 
 type rubiconImpExtRP struct {
 	ZoneID int                  `json:"zone_id"`
-	Target json.RawMessage      `json:"target"`
+	Target json.RawMessage      `json:"target,omitempty"`
 	Track  rubiconImpExtRPTrack `json:"track"`
 }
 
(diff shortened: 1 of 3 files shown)


adapters/rubicon/rubicon.go
```diff
@@ -53,7 +53,7 @@ type rubiconImpExtRPTrack struct {
```
guscarreon: Are `Inventory` and `Visitor` these two not necessary to omit if empty?
```
 40 type rubiconParams struct {
 41     AccountId int                `json:"accountId"`
 42     SiteId    int                `json:"siteId"`
 43     ZoneId    int                `json:"zoneId"`
 44     Inventory json.RawMessage    `json:"inventory"`
 45     Visitor   json.RawMessage    `json:"visitor"`
 46     Video     rubiconVideoParams `json:"video"`
 47 }
 48
adapters/rubicon/rubicon.go
```

potatogopher: Forgot a few lines. Thank you.

PR #969, opened 2019-07-22
Add an app.id blacklist to PBS and reject requests from it

guscarreon:
Still to do:
  * Get the list with app ids to include in the list
  * Configure environment variable in bitbucket


endpoints/openrtb2/auction.go
```diff
@@ -126,6 +126,9 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		labels.Source = pbsmetrics.DemandApp
 		labels.RType = pbsmetrics.ReqTypeORTB2App
 		labels.PubID = effectivePubID(req.App.Publisher)
+		if _, found := deps.cfg.BlacklistedAppMap[labels.PubID]; found {
+			return
```
hhhjort: We might want to write some sort of error back on `w` at this point before simply returning at this point. The only other places we return are in the case of a severe error, so I can kind of understand not writing something out there. Since this could be a common occurrence, I worry that simply returning will cause connections to hang until a timeout. Probably worth at least testing if the return will close the connection immediately or hang the connection.

endpoints/openrtb2/video_auction.go
```diff
@@ -189,6 +189,9 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 	if bidReq.App != nil {
 		labels.Source = pbsmetrics.DemandApp
 		labels.PubID = effectivePubID(bidReq.App.Publisher)
+		if _, found := deps.cfg.BlacklistedAppMap[labels.PubID]; found {
+			return
```
hhhjort: See above somment on simply returning at this point.

endpoints/openrtb2/auction.go
guscarreon: Corrected Hans. Updated PR now writes to `w`

endpoints/openrtb2/video_auction.go
guscarreon: Seen and done

endpoints/openrtb2/auction_test.go
```diff
@@ -253,7 +253,12 @@ func (gr *getResponseFromDirectory) assert(t *testing.T) {
 
 		expectMsg := gr.messageGetter(t, fileData)
 		if len(expectMsg) > 0 {
-			assert.Equal(t, string(expectMsg), msg, "file %s had bad response body", filename)
+			//assert.Equal(t, string(expectMsg), msg, "file %s had bad response body", filename)
```
mansinahar: Can you please remove this comment?

endpoints/openrtb2/auction_test.go
```diff
@@ -253,7 +253,12 @@ func (gr *getResponseFromDirectory) assert(t *testing.T) {
...
+			if string(expectMsg) != msg {
+				t.Errorf("[ERROR]     Test: %s \n", filename)
```
mansinahar: We should rather use testify's assert package for assertions. It prints all this information out of the box

endpoints/openrtb2/auction_test.go
guscarreon: done

endpoints/openrtb2/auction_test.go
guscarreon: rolled back the if statement. It was meant for debugging locally

config/config.go
```diff
@@ -634,6 +645,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("default_request.type", "")
 	v.SetDefault("default_request.file.name", "")
 	v.SetDefault("default_request.alias_info", false)
+	v.SetDefault("blacklisted_apps", []string{""})
```
mansinahar: @guscarreon Do we need a slice with empty string as default? I mean why can't it be just an empty slice?

endpoints/openrtb2/auction.go
```diff
@@ -1154,7 +1160,11 @@ func checkSafari(r *http.Request) (isSafari bool) {
 // Write(return) errors to the client, if any. Returns true if errors were found.
 func writeError(errs []error, w http.ResponseWriter) bool {
 	if len(errs) > 0 {
-		w.WriteHeader(http.StatusBadRequest)
+		if errortypes.DecodeError(errs[0]) == errortypes.BlacklistedAppCode {
```
mansinahar: I understand that the way the code is arranged right now, if `BlacklistedApp` error occurs, it will be the first one in the `errs` slice but if we make any changes to the code in the future then this might or might not be true. And this will also be difficult to catch at the time. For example, right now this is true because you re-arranged the imp validation block. If the imp validation block were still where it was before, the `BlacklistedApp` error might not be the first one in the `errs` slice. And so, I'd suggest we rather go through the slice and check all the errors than just the first one. Maybe we can add the `BlacklistedApp` error as a check in the `fatalError` function?

endpoints/openrtb2/video_auction.go
```diff
@@ -242,14 +242,19 @@ func cleanupVideoBidRequest(videoReq *openrtb_ext.BidRequestVideo, podErrors []P
 
 func handleError(labels pbsmetrics.Labels, w http.ResponseWriter, errL []error, ao analytics.AuctionObject) {
 	labels.RequestStatus = pbsmetrics.RequestStatusErr
-	w.WriteHeader(http.StatusInternalServerError)
+	if errortypes.DecodeError(errL[0]) == errortypes.BlacklistedAppCode {
```
mansinahar: Again, we should check the entire `errs` slice rather than just the first element.

endpoints/openrtb2/video_auction.go
```diff
@@ -617,9 +622,19 @@ func (deps *endpointDeps) validateVideoRequest(req *openrtb_ext.BidRequestVideo)
...
-		err := errors.New("request.app missing required field: id or bundle")
-		errL = append(errL, err)
+	} else if req.App != nil {
+		if req.App.ID != "" {
+			if _, found := deps.cfg.BlacklistedAppMap[req.App.ID]; found {
+				err := &errortypes.BlacklistedApp{Message: fmt.Sprintf("Prebid-server does not process requests from App ID: %s", req.App.ID)}
+				errL = append(errL, err)
+				return errL, podErrors
+			}
+		} else { // req.App.ID == ""
```
mansinahar: Just a nit pick but since there are other changes, can you please remove this comment?

config/config.go
guscarreon: Since the `flow` keyword was added, the `BlacklistedApps` field  in line 55 below expects an array of strings.
```
 21 type Configuration struct {
 22     ExternalURL string     `mapstructure:"external_url"`
 23 +-- 28 lines: Host        string     `mapstructure:"host"`---------------------------------------
 51
 52     VideoStoredRequestRequired bool `mapstructure:"video_stored_request_required"`
 53
 54     // PBS-418 array of blacklisted apps and its correpondent hash table to be instantly accessed.
 55     BlacklistedApps   []string `mapstructure:"blacklisted_apps,flow"`
 56     BlacklistedAppMap map[string]bool
 57 }
 58
```
Therefore, I figured it was better to add a slice with an empty string because other `string` fields of the `Configuration` struct add an empty string instead of a nil string as default. Take the `Host` field for example, in `config/config.go` its default is set to an empty string as well as `StatusResponse` and `RecaptchaSecret` to name a few:
```
 21 type Configuration struct {
 22     ExternalURL string     `mapstructure:"external_url"`
 23     Host        string     `mapstructure:"host"`
 24     Port        int        `mapstructure:"port"`
 25     Client      HTTPClient `mapstructure:"http_client"`
 26     AdminPort   int        `mapstructure:"admin_port"`
 27     EnableGzip  bool       `mapstructure:"enable_gzip"`
 28     // StatusResponse is the string which will be returned by the /status endpoint when things are OK.
 29     // If empty, it will return a 204 with no content.
 30     StatusResponse  string             `mapstructure:"status_response"`
 31     AuctionTimeouts AuctionTimeouts    `mapstructure:"auction_timeouts_ms"`
 32     CacheURL        Cache              `mapstructure:"cache"`
 33     RecaptchaSecret string             `mapstructure:"recaptcha_secret"`
 34     HostCookie      HostCookie         `mapstructure:"host_cookie"`
 35     Metrics         Metrics            `mapstructure:"metrics"`
 36 +--444 lines: DataCache       DataCache          `mapstructure:"datacache"`---------------------
480 func SetupViper(v *viper.Viper, filename string) {
481     if filename != "" {
482         v.SetConfigName(filename)
483         v.AddConfigPath(".")
484         v.AddConfigPath("/etc/config")
485     }
486     // Fixes #475: Some defaults will be set just so they are accessible via environment variables
487     // (basically so viper knows they exist)
488     v.SetDefault("external_url", "http://localhost:8000")
489     v.SetDefault("host", "")
490     v.SetDefault("port", 8000)
491     v.SetDefault("admin_port", 6060)
492     v.SetDefault("enable_gzip", false)
493     v.SetDefault("status_response", "")
494     v.SetDefault("auction_timeouts_ms.default", 0)
495     v.SetDefault("auction_timeouts_ms.max", 0)
496     v.SetDefault("cache.scheme", "")
497     v.SetDefault("cache.host", "")
498     v.SetDefault("cache.query", "")
499     v.SetDefault("cache.expected_millis", 10)
500     v.SetDefault("cache.default_ttl_seconds.banner", 0)
501     v.SetDefault("cache.default_ttl_seconds.video", 0)
502     v.SetDefault("cache.default_ttl_seconds.native", 0)
503     v.SetDefault("cache.default_ttl_seconds.audio", 0)
504     v.SetDefault("recaptcha_secret", "")
505     v.SetDefault("host_cookie.domain", "")
506     v.SetDefault("host_cookie.family", "")
```

endpoints/openrtb2/video_auction.go
guscarreon: corrected

endpoints/openrtb2/video_auction.go
guscarreon: done

guscarreon:
> On this line: https://github.com/prebid/prebid-server/pull/969/files#diff-09e5d415f33d74fd6df20b2584f318d1R110, the `if fatalError(errL) && writeError(errL, w)` condition will be false in case of `BlacklistedApp` errors and so we will continue processing the request. I don't think that's the expected behavior, right?

Good catch! Corrected.

endpoints/openrtb2/video_auction.go
```diff
@@ -242,14 +242,23 @@ func cleanupVideoBidRequest(videoReq *openrtb_ext.BidRequestVideo, podErrors []P
...
-		errors = fmt.Sprintf("%s %s", errors, er.Error())
+		if errortypes.DecodeError(er) == errortypes.BlacklistedAppCode {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			ao.Status = http.StatusServiceUnavailable
+			errors = fmt.Sprintf("%s %s", er.Error(), errors)
+		} else {
+			errors = fmt.Sprintf("%s %s", errors, er.Error())
+		}
+	}
+	if !foundBlacklisted {
```
mansinahar: You never set the `foundBlacklisted` anywhere in the above code, so this block will always get executed. I don't think that's what we want.

guscarreon: You are right. Corrected

PR #973, opened 2019-07-24
Add COPPA support

mansinahar:


exchange/utils.go
```diff
@@ -41,16 +42,27 @@ func cleanOpenRTBRequests(ctx context.Context,
 	// Clean PI from bidrequests if not allowed per GDPR
 	gdpr := extractGDPR(orig, usersyncIfAmbiguous)
 	consent := extractConsent(orig)
-	if gdpr == 1 {
-		for bidder, bidReq := range requestsByBidder {
-			// Fixes #820
+
+	for bidder, bidReq := range requestsByBidder {
+		var applyGDPR, applyCOPPA, isAMP bool
```
hhhjort: Shouldn't `applyCOPPA` and `isAMP` be defined and set outside of the loop, as they don't depend on the bidder in question?

exchange/utils.go
```diff
@@ -297,3 +309,105 @@ func randomizeList(list []openrtb_ext.BidderName) {
...
+		// to know if the vendor is consented or not and therefore for AMP requests
+		// we keep the BuyerUID as is
+		if !isAMP {
+			bidRequest.User.BuyerUID = ""
+		}
+		if applyCOPPA {
+			bidRequest.User.ID = ""
+			bidRequest.User.Yob = 0
+			bidRequest.User.Gender = ""
+			bidRequest.User.BuyerUID = ""
```
hhhjort: We should probably also clean out any user mappings we have defined for the user.

exchange/utils.go
mansinahar: Good point. I will change and update. Thanks!

PR #976, opened 2019-07-26
Minor fix:

Kalypsonika:
-Added null checks for Site and App
-Added required fields validation: Site: id or page, App: id or bundle

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -440,15 +440,15 @@ func getVideoStoredRequestId(request []byte) (string, error) {
 
 func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.BidRequest) error {
 
-	if videoRequest.Site.Page != "" {
-		bidRequest.Site = &videoRequest.Site
+	if videoRequest.Site != nil {
+		bidRequest.Site = videoRequest.Site
 		if &videoRequest.Content != nil {
 			bidRequest.Site.Content = &videoRequest.Content
 		}
 	}
 
-	if videoRequest.App.Domain != "" {
-		bidRequest.App = &videoRequest.App
+	if videoRequest.App != nil {
+		bidRequest.App = videoRequest.App
 		if &videoRequest.Content != nil {
 			bidRequest.App.Content = &videoRequest.Content
 		}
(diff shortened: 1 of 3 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -440,15 +440,15 @@ func getVideoStoredRequestId(request []byte) (string, error) {
 
 func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.BidRequest) error {
 
-	if videoRequest.Site.Page != "" {
-		bidRequest.Site = &videoRequest.Site
+	if videoRequest.Site != nil {
+		bidRequest.Site = videoRequest.Site
```
guscarreon: Do we still want fo check for the `Page` string content?
```
441 func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.BidRequest) error {
442
443     if videoRequest.Site != nil && videoRequest.Site.Page != "" {
444         bidRequest.Site = videoRequest.Site
445         if &videoRequest.Content != nil {
446             bidRequest.Site.Content = &videoRequest.Content
447         }
video_auction.go
```

endpoints/openrtb2/video_auction.go
```diff
@@ -440,15 +440,15 @@ func getVideoStoredRequestId(request []byte) (string, error) {
...
 		if &videoRequest.Content != nil {
 			bidRequest.Site.Content = &videoRequest.Content
 		}
 	}
 
-	if videoRequest.App.Domain != "" {
-		bidRequest.App = &videoRequest.App
+	if videoRequest.App != nil {
+		bidRequest.App = videoRequest.App
```
guscarreon: Do we still want fo check for the `Domain` string content?
```
450     if videoRequest.App != nil && videoRequest.App.Domain != "" {
451         bidRequest.App = videoRequest.App
452         if &videoRequest.Content != nil {
453             bidRequest.App.Content = &videoRequest.Content
454         }
455     }
456
video_auction.go
```

endpoints/openrtb2/video_auction.go
Kalypsonika: @guscarreon thank you for your comment. 
According to Open RTB specification this is the definition of `page` parameter: 
```
Attribute | Type    | Description
page       | string   | URL of the page where the impression will be shown.
```
I can add reg exp verification if needed. 
@hhhjort  please advice

endpoints/openrtb2/video_auction.go
Kalypsonika: For `domain` here is the definition from Open RTB spec:
```
Attribute | Type  | Description
domain  | string  |  Domain of the app (e.g., “mygame.foo.com”).
```
If we need to add reg exp verification, how it should look like? 
@hhhjort @guscarreon please share how do you guys think it should be.

PR #983, opened 2019-07-30
Shift the metrics collection rate from every 10 to 20 seconds

guscarreon:
Shift the metrics collection rate from one datapoint every 10 seconds to one every 20. Also update the graphana queries to keep the graphs reasonable


pbsmetrics/prometheus/prometheus.go
```diff
@@ -56,7 +56,7 @@ const (
 func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
 	// define the buckets for timers
 	timerBuckets := prometheus.LinearBuckets(0.05, 0.05, 20)
-	timerBuckets = append(timerBuckets, []float64{1.5, 2.0, 3.0, 5.0, 10.0, 50.0}...)
+	timerBuckets = append(timerBuckets, []float64{1.5, 2.0, 3.0, 5.0, 20.0, 50.0}...)
```
hhhjort: Why was this changed?

guscarreon: My bad, let me rollback this one.

PR #987, opened 2019-08-01
[Sharethrough] Add Unified ID support

maphe:
We are adding support for Unified ID:
- parsing `user.ext.eids` to extract id from source `adserver.org`
- sending this id to the ad server if found

### DIFF
--- a/adapters/sharethrough/utils.go
+++ b/adapters/sharethrough/utils.go
@@ -206,6 +206,10 @@ func (u Util) parseDomain(fullUrl string) string {
 		} else {
 			domain = uri.Host
 		}
+
+		if domain != "" {
+			domain = uri.Scheme + "://" + domain
+		}
 	}
 
 	return domain
--- a/adapters/sharethrough/utils.go
+++ b/adapters/sharethrough/utils.go
@@ -193,23 +193,23 @@ func (u Util) gdprApplies(request *openrtb.BidRequest) bool {
 	return gdprApplies != 0
 }
 
-func (u Util) parseUserExt(user *openrtb.User) userInfo {
+func (u Util) parseUserExt(user *openrtb.User) (ui userInfo) {
 	var userExt userExt
-	var unifiedId string
-	var gdprConsentString string
 	if user != nil && user.Ext != nil {
 		if err := json.Unmarshal(user.Ext, &userExt); err == nil {
-			gdprConsentString = userExt.Consent
+			ui.Consent = userExt.Consent
 			for i := 0; i < len(userExt.Eids); i++ {
-				if userExt.Eids[i].Source == "adserver.org" {
-					unifiedId = userExt.Eids[i].Uids[0].ID
+				if userExt.Eids[i].Source == "adserver.org" && len(userExt.Eids[i].Uids) > 0 {
+					if userExt.Eids[i].Uids[0].ID != "" {
+						ui.TtdUid = userExt.Eids[i].Uids[0].ID
+					}
 					break
 				}
 			}
 		}
 	}
 
-	return userInfo{Consent: gdprConsentString, TtdUid: unifiedId}
+	return
 }
 
 func (u Util) parseDomain(fullUrl string) string {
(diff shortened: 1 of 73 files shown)


adapters/sharethrough/utils.go
```diff
@@ -183,17 +193,23 @@ func (u Util) gdprApplies(request *openrtb.BidRequest) bool {
...
 }
 
-func (u Util) gdprConsentString(request *openrtb.BidRequest) string {
-	var consentString string
-
-	if request.User != nil {
-		if jsonExtUser, err := request.User.Ext.MarshalJSON(); err == nil {
-			// empty string is the return value if error, so no need to handle
-			consentString, _ = jsonparser.GetString(jsonExtUser, "consent")
+func (u Util) parseUserExt(user *openrtb.User) userInfo {
```
hhhjort: You could make this `func (u Util) parseUserExt(user *openrtb.User) (ui userInfo) {` and then use `ui.Consent` and `ui.TtdUid`instead of `unifiedId` and `gdprConsentString`. Then you just need to `return` rather than building a `userInfo` struct in the return statement.

maphe: Good to know, thanks for the tip. I'm just gonna leave it this way for now as I don't want to cancel your approval and wait for another round of review 😆I'll likely address it with the next feature/PR 👍

adapters/sharethrough/utils.go
```diff
@@ -183,17 +193,23 @@ func (u Util) gdprApplies(request *openrtb.BidRequest) bool {
...
+	var userExt userExt
+	var unifiedId string
+	var gdprConsentString string
+	if user != nil && user.Ext != nil {
+		if err := json.Unmarshal(user.Ext, &userExt); err == nil {
+			gdprConsentString = userExt.Consent
+			for i := 0; i < len(userExt.Eids); i++ {
+				if userExt.Eids[i].Source == "adserver.org" {
+					unifiedId = userExt.Eids[i].Uids[0].ID
```
mansinahar: Would be good to check if `Eids[i].Uids` has a length greater than 0 before trying to access `Eids[i].Uids[0]` in order to avoid potential index out of bounds error. 

Another thing would be to check if `userExt.Eids[i].Uids[0].ID` is not an empty string before assigning to `unifiedId`

Also, these will be some good test cases to add ;)

adapters/sharethrough/utils_test.go
```diff
@@ -356,55 +356,43 @@ func TestGdprApplies(t *testing.T) {
...
 	}
 
 	util := Util{}
 	for testName, test := range tests {
 		t.Logf("Test case: %s\n", testName)
 
-		output := util.gdprConsentString(test.input)
-		if output != test.expected {
-			t.Errorf("Expected: %s, got %s\n", test.expected, output)
+		output := util.parseUserExt(test.input)
```
mansinahar: Just a tip here, I would suggest using testify's assert package as it has a bunch of built in methods for asserting values and structs. Also, it has a nice user friendly output format when things fail. Might be worth checking out.

adapters/sharethrough/utils.go
epechuzal: @mansinahar thanks for the feedback! 

Are you OK to merge this PR as-is (and I can create a new PR later with that change), or would you prefer if I make the changes tomorrow and add to this PR?

mansinahar: @maphe We're planning on doing a release tomorrow so if you can make those changes by tomorrow, that's okay. I can then get it merged for it to be a part of that release.

adapters/sharethrough/utils_test.go
epechuzal: @mansinahar thank you for the tip! I changed the utils tests to use assert, now its much easier to read

adapters/sharethrough/utils.go
epechuzal: @mansinahar thank you! I've made the requested updates and pushed those out. Let me know if there is anything else we should add or change :)

epechuzal:
@mansinahar @hhhjort I've addressed the PR comments: let me know if there is anything else you would like us to change

PR #990, opened 2019-08-06
[godoc] properly format `Rev` doc

benjaminch:
Setup Field name into the comment so it's properly interpreted by godoc


benjaminch:
Thanks @mansinahar ! ;)

PR #991, opened 2019-08-06
[lint] applying go-lint suggestions

benjaminch:
Applying go-lint suggestions

### DIFF
--- a/pbs_light_test.go
+++ b/pbs_light_test.go
@@ -28,11 +28,10 @@ func forceEnv(t *testing.T, key string, val string) func() {
 				t.Fatalf("Error unsetting evnvironment %s", key)
 			}
 		}
-	} else {
-		return func() {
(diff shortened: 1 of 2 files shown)


benjaminch:
Thanks @mansinahar :)

pbs_light_test.go
```diff
@@ -28,11 +28,10 @@ func forceEnv(t *testing.T, key string, val string) func() {
 				t.Fatalf("Error unsetting evnvironment %s", key)
```
guscarreon: I realize you didn't misspell this word but, could you correct it? Unless it is a hassle to go through another round of reviews. If so, we can live with it.

benjaminch: No worries I’ll handle it !

benjaminch: Here you go: #996

PR #992, opened 2019-08-06
Adding support for hb_source field in request made to AppNexus bidder…

jmaynardxandr:
…. The value for this comes from the "adapters.appnexus.platform_id" config option, and is incremented by one for requests that come from the longform video endpoint.

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -26,6 +26,7 @@ type AppNexusAdapter struct {
 	http           *adapters.HTTPAdapter
 	URI            string
 	iabCategoryMap map[string]string
+	hbSource       int
 }
 
 // used for cookies and such
@@ -96,6 +97,7 @@ type appnexusReqExtAppnexus struct {
 	IncludeBrandCategory    *bool `json:"include_brand_category,omitempty"`
 	BrandCategoryUniqueness *bool `json:"brand_category_uniqueness,omitempty"`
 	IsAMP                   int   `json:"is_amp,omitempty"`
+	HeaderBiddingSource     int   `json:"hb_source,omitempty"`
 }
 
 // Full request extension including appnexus extension object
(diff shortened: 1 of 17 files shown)


adapters/appnexus/appnexus.go
```diff
@@ -327,43 +329,35 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *ada
...
-		} else {
+		if err := json.Unmarshal(request.Ext, &reqExt); err != nil {
 			errs = append(errs, err)
 			return nil, errs
 		}
-	} else if isAMP != 0 {
-		// if isAMP is 0, bidder doesn't need it included
-		var err error
+	}
+	if reqExt.Appnexus == nil {
```
guscarreon: In the previous version of the code, the `if reqExt.Appnexus == nil` condition in line 346 was inside the `if includeBrandCategory` if statement, which made it dependant upon this condition to be true. Whay are we changing this?
```
346     if reqExt.Appnexus == nil {
347         reqExt.Appnexus = &appnexusReqExtAppnexus{}
348     }
349     includeBrandCategory := reqExt.Prebid.Targeting != nil && reqExt.Prebid.Targeting.IncludeBrandCategory.PrimaryAdServer != 0
350     if includeBrandCategory {
351         reqExt.Appnexus.BrandCategoryUniqueness = &includeBrandCategory
352         reqExt.Appnexus.IncludeBrandCategory = &includeBrandCategory
353     }
adapters/appnexus/appnexus.go
```

jmaynardxandr: So the regExt.Appnexus was being created before only if it was needed, which was when either the includeBrandCategories or isAMP flags were on. With this hb_source field we always want to send it, so now the regExt.Appnexus object needs to be created all the time even in the case where includeBrandCategories and isAMP are both off.

guscarreon: Ohhh I see

PR #993, opened 2019-08-07
Issue862 - implemented multi media bids

muncha:
This is a correction to closed PR #965. The "forceBanner" flag is removed, and as suggested by @hhhjort, multiple bid requests are being added to the stack for multimedia.


adapters/beachfront/beachfront.go
```diff
@@ -24,41 +22,45 @@ const VideoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id="
...
+	Audio  openrtb.Audio
+	Native openrtb.Native
 }
 
 // ---------------------------------------------------
 //              Video
 // ---------------------------------------------------
 
-type BeachfrontVideoRequest struct {
+type beachfrontVideoRequest struct {
```
hhhjort: Below you marshal `beachfrontRequests.Video[i]` but `Video` is the `beachfrontVideoRequest` struct. Since this is private, the marshaling should fail. You might also want to review your tests cases that this failure was not already caught.

adapters/beachfront/beachfront.go
```diff
@@ -67,8 +69,8 @@ type BeachfrontVideoDevice struct {
 // ---------------------------------------------------
 //              Banner
 // ---------------------------------------------------
-type BeachfrontBannerRequest struct {
-	Slots          []BeachfrontSlot `json:"slots"`
+type beachfrontBannerRequest struct {
```
hhhjort: This is also marshaled, so should not be private

adapters/beachfront/beachfront.go
```diff
@@ -111,177 +113,208 @@ type BeachfrontResponseSlot struct {
...
+	r := beachfrontBannerRequest{}
+	r.AdapterName = beachfrontAdapterName
+	r.AdapterVersion = beachfrontAdapterVersion
+
+	return r
+}
+
+func newBeachfrontVideoRequest() beachfrontVideoRequest {
+	r := beachfrontVideoRequest{}
+	r.Cur = append(r.Cur, "USD")
```
hhhjort: Will it be an issue if `r.Cur` already contains a "USD" entry?

adapters/beachfront/beachfront.go
```diff
@@ -111,177 +113,208 @@ type BeachfrontResponseSlot struct {
...
-			for j := range imp.Banner.Format {
-				if j > 0 {
-					beachfrontReq.Slots[bannerImpsIndex].Sizes = append(beachfrontReq.Slots[bannerImpsIndex].Sizes, BeachfrontSize{})
-				}
-				beachfrontReq.Slots[bannerImpsIndex].Sizes[j].H = imp.Banner.Format[j].H
-				beachfrontReq.Slots[bannerImpsIndex].Sizes[j].W = imp.Banner.Format[j].W
-			}
+	for i := 0; i < len(request.Imp); i++ {
+		beachfrontReq.Slots = append(beachfrontReq.Slots, beachfrontSlot{})
+		slotIndex = len(beachfrontReq.Slots) - 1
```
hhhjort: Is it possible for `slotIndex` to diverge from `i`?

adapters/beachfront/beachfront.go
muncha: I'm not clear on what should make this fail. I mean, yes, I set that struct to private figuring limiting scope to the minimum is generally good practice, and I saw no reason to make it public, not to say that there isn't one that I'm not getting, but it does actually work.  I am testing with Prebid.js test pages on a local npm install, which are sending a variety of requests, including a multi media / multiple imp request, to a running a local prebid server debug build in intellij. I can step through the debugger and see the marshaling work. Below, I have stepped through to i = 1. beachfrontRequests.Video[i] was marshaled to bytes the first time, then again the second time, with no error. The request gets appended to reqs, sent, received, processed, and I see the test ad video play on the test page.

Screenshot from 2019-08-14 12-52-10

adapters/beachfront/beachfront.go
muncha: It would be an issue, but I don't think it's possible, as r is a spanking new beachfrontVideoRequest, but I added a test for it anyway.

adapters/beachfront/beachfront.go
hhhjort: Hmm, perhaps it works since all the fields are public, even if the struct itself is private. The json package needs know what a `beachfrontVideoRequest` is so that it can generate the json representation of it for the payload you send to your servers. But if you have verified that the request is being generated properly despite the struct name being private, then I guess it is good.

Also yes, it is best in general to default to limiting scope. :)

adapters/beachfront/beachfront.go
muncha: related - I noticed that impIndex was redundant in that loop, so eliminated it.

As for divergence, in preProcess, initial request in broken into two separate requests, one containing imps that are all guaranteed to contain a banner, the other all are guaranteed to contain a video. So the banner request is sent into getBannerRequest, it's count of imps is the the same as the count of slots that I should need in the beachfrontReq. If there is an error in getBeachfrontExtension, that slot will still in the stack, but it's a junk slot. I don't see a way that could happen, as the format of the extension would have to be wrong, and I don't see how that would get past validation, but saying it did, I would not want to include that slot, and at that point it _should_ diverge, but, no, it wasn't.

So I'm thinking that it should have the possibility of diverging in that case. I've got that where I think I want it. I'm not sure how to cause such an error in test cases, though. I'm able to manually throw an error 
`		if(i == 1) {
			err = errors.New("test error")
		}`

while running a big request on a test page with a debugging server build, I can see it diverge, and it seems to be doing so as expected - if I have 6 banner imps, and 2 are thrown a test error, I get 4 Slots in the final request.

I'm looking into how to test case this.

Or maybe I should forget about this error case that I don't think will ever happen and just index everything by "i" ?

muncha: Short answer - it wasn't but it is now, in a controlled way, but I'm not sure it should.

adapters/beachfront/beachfront.go
muncha: I'm going to go ahead and close this one and the one below, then. I'll be reading up on how the compiler handles this.  I suspect that this relates to why in go "private" and "public" are not actually the correct terms, but a package has "exported" identifiers and non-exported identifiers.  I don't quite get what the difference is besides semantics, but I suspect that this behavior is due to some specific detail of compilation that reflects that distinction. So off to RTFM.

adapters/beachfront/beachfront.go
muncha: Resolving this as above.

adapters/beachfront/beachfront.go
muncha: @hhhjort 
Do you think it is better as it is now, or better if I drop the whole diverging stack concept, drop slotIndex, make the Slots slice the same size as the Imp slice initially, and just index both by i?

muncha: Which ever is more appropriate, I should also apply to getVideoRequests, as it also has what is currently two redundant indexes,  i and videoImpsIndex,  impCount that is never used, and tests if the Ext failed.

muncha: I'm going to go with this being a good test to make, just in case. Also, if I'm doing it like that here, I should do similar, and eliminate some redundancies, in getVideoRequests.  So please don't merge this until I get that. Shouldn't take long at all.

If it is true that the params validation will _always_ catch errors in the Ext, and that the Ext unmarshaling will _never_ fail (???), then I could just drop the slotIndex, all the appending, etc, and just make the slices the same size initially and index them by "i". 

If you think that the later approach is better (Probably not as effected by increased size of Bids array), just let me know and I can get them both into that form instead.

muncha: I audited getVideoRequests and LGTM

adapters/beachfront/beachfront.go
```diff
@@ -435,43 +480,49 @@ func (a *BeachfrontAdapter) MakeBids(internalRequest *openrtb.BidRequest, extern
...
 
 	return bidResponse, errs
 }
 
-func postprocess(response *adapters.ResponseData, externalRequest *adapters.RequestData, id string, bidtype openrtb_ext.BidType) ([]openrtb.Bid, []error) {
-	var beachfrontResp []BeachfrontResponseSlot
+func postprocess(response *adapters.ResponseData, externalRequest *adapters.RequestData, id string) ([]openrtb.Bid, []error) {
+	var beachfrontResp []beachfrontResponseSlot
 	var errs = make([]error, 0)
+	// var list = json.Unmarshal()
```
mansinahar: Just a nit pick but since there are other changes can you please remove this comment?

adapters/beachfront/beachfront.go
```diff
@@ -435,43 +480,49 @@ func (a *BeachfrontAdapter) MakeBids(internalRequest *openrtb.BidRequest, extern
...
+	var openrtbResp openrtb.BidResponse
+
+	// try it as a video
+	if err := json.Unmarshal(response.Body, &openrtbResp); err != nil {
+
+		// try it as a banner
 		if err := json.Unmarshal(response.Body, &beachfrontResp); err != nil {
-			errs = append(errs, err)
+			// it's neither. I'm not appending these errors
+			// errs = append(errs, err)
```
mansinahar: Can you please remove this comment too?

adapters/beachfront/beachfront.go
muncha: Thank you for nit picking. Done locally.

adapters/beachfront/beachfront.go
muncha: Done locally

adapters/beachfront/beachfront.go
```diff
@@ -111,177 +113,229 @@ type BeachfrontResponseSlot struct {
...
+
+	if request.Imp[0].Secure != nil {
+		beachfrontReq.Secure = *request.Imp[0].Secure
 	}
 
-	if req.User != nil {
-		beachfrontReq.User.ID = req.User.ID
-		beachfrontReq.User.BuyerUID = req.User.BuyerUID
+	if request.User != nil {
+		beachfrontReq.User.ID = request.User.ID
```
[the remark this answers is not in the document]
muncha: Done locally

adapters/beachfront/beachfront.go
```diff
@@ -290,121 +344,112 @@ func getBannerRequest(req *openrtb.BidRequest) (BeachfrontBannerRequest, []error
...
+		} else {
+			r.Imp[videoIndex].Video.W = DefaultVideoWidth
+			r.Imp[videoIndex].Video.H = DefaultVideoHeight
+		}
 
-			var bidderExt adapters.ExtImpBidder
-			if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
-				errs = append(errs, err)
-				continue
+		if request.App != nil {
```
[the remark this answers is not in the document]
muncha: I'm not saying it _can't_, but it gets really tricky since such a method would need to be able to handle both beachfrontVideoRequest enums and beachfrontBannerRequest enums as arguments and return values. I think the spirit of your comment is to avoid duplicated code. 

I'm doing a push now with the two methods reordered to mostly in parallel. That's helping me see where I can get some duplication out of there. Domain is not processed in a shared method. Still looking at what else can be shared.

adapters/beachfront/beachfront.go
```diff
@@ -111,177 +113,229 @@ type BeachfrontResponseSlot struct {
...
+	}
 
-func getEndpoint(request *openrtb.BidRequest) (uri string) {
-	for i := range request.Imp {
-		if request.Imp[i].Video != nil {
-			// If there are any video imps, we will be running a video auction
-			// and dropping all of the banner actions.
-			return VideoEndpoint
+	if request.User != nil && request.User.BuyerUID != "" {
+		addHeaderIfNonEmpty(headers, "Cookie", "__io_cid="+request.User.BuyerUID)
```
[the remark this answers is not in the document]
muncha: Right. Also, I don't need to send the cookie with the banner request, so moved things around a bit to avoid that.

adapters/beachfront/beachfront.go
```diff
@@ -111,177 +113,229 @@ type BeachfrontResponseSlot struct {
...
+	}
+	reqs := make([]*adapters.RequestData, 0)
```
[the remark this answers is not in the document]
muncha: Yeah, thanks. That could slow things down a lot.

adapters/beachfront/beachfront.go
```diff
@@ -111,177 +113,229 @@ type BeachfrontResponseSlot struct {
...
-		return nil, errs
+	if weGotNothing {
+		errs = append(errs, errors.New("no valid impressions were found in the request"))
+		return
 	}
 
-	if err != nil {
-		errs = append(errs, err)
-		return nil, errs
+	request.Imp = make([]openrtb.Imp, 0)
```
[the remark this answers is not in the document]
muncha: The idea there was that I was clearing out the Imp that is sent in and replacing it with either bannerImps or videoImps, so why am I making an empty one in the first place? Looking back through it, I don't see a reason to as I'm then immediately reassigning it.   It seems like I can just skip it.

adapters/beachfront/beachfront.go
```diff
@@ -111,177 +113,229 @@ type BeachfrontResponseSlot struct {
...
+	var err error
+	var bidderExt adapters.ExtImpBidder
+	var beachfrontExt openrtb_ext.ExtImpBeachfront
 
-	if uri == BannerEndpoint {
-		beachfrontReqs.Banner, errs, imps = getBannerRequest(req)
-	} else {
-		// If there were any Video imps in the request, we have skipped to here.
-		beachfrontReqs.Video, errs, imps = getVideoRequest(req)
+	if err = json.Unmarshal(imp.Ext, &bidderExt); err != nil {
```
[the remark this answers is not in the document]
muncha: Yes, much better.

adapters/beachfront/beachfront.go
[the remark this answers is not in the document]
muncha: Worked that out. Just dealing with a Site enum instead of the whole thing.

adapters/beachfront/beachfront.go
```diff
@@ -290,121 +344,112 @@ func getBannerRequest(req *openrtb.BidRequest) (BeachfrontBannerRequest, []error
...
-				}
-			} else {
-				beachfrontReq.Site.Domain = req.Site.Domain
-			}
-			beachfrontReq.Site.Page = req.Site.Page
-		}
-	}
+	for i := 0; i < len(request.Imp); i++ {
+		r := newBeachfrontVideoRequest()
+		r.Imp = append(r.Imp, beachfrontVideoImp{})
```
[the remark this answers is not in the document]
muncha: The beachfront video endpoint takes an array of imps, but only ever handles the first in the array, so videoIndex was always 0. I eliminated the variable.

muncha:
@mansinahar LGTM. Thank you.

adapters/beachfront/beachfront.go
```diff
@@ -111,300 +113,334 @@ type BeachfrontResponseSlot struct {
...
+		beachfrontReqs.Banner, errs = getBannerRequest(request)
 	}
 
-	if err != nil {
-		errs = append(errs, err)
-		return nil, errs
+	if len(videoImps) > 0 {
+		request.Imp = videoImps
+
+		if len(errs) > 0 {
```
mansinahar: You can append a slice to another slice by doing `slice2 = append(slice2, slice1...)`. The `...` after the second argument to the `append` call lets you pass in a slice versus a single value. And so, you can replace this `if else` block with:

```
beachfrontReqs.Video, videoErrs = getVideoRequests(request)
errs = append(errs, videoErrs...)
```

This will append error returned by the `getVideoRequests` method call to the `errs` slice which should already have the errors returned by the `getBannerRequest` method call above, if any.

adapters/beachfront/beachfront.go
```diff
@@ -111,300 +113,334 @@ type BeachfrontResponseSlot struct {
...
-	/*
-		The req could contain banner,audio,native and video imps when It arrives here. I am only
-		interested in video
+	site := getSite(request)
+	bfBannerRequest.IsMobile = site.Mobile
+	bfBannerRequest.Page = site.Page
+	bfBannerRequest.Domain = site.Domain
+
+	if request.User != nil {
+		if bfBannerRequest.User.ID == "" {
```
mansinahar: Shouldn't this `if` condition rather be: `if request.User.ID != ""` and similar for the next `if` condition too?

adapters/beachfront/beachfront.go
```diff
@@ -111,300 +113,334 @@ type BeachfrontResponseSlot struct {
...
-			if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
-				errs = append(errs, err)
-				continue
-			}
+/*
+getVideoRequests, plural. One request to the endpoint can have one appId, and can return one nurl,
+so each video imp is a call to the endpoint.
+*/
+func getVideoRequests(request *openrtb.BidRequest) ([]beachfrontVideoRequest, []error) {
+	var beachfrontReqs = make([]beachfrontVideoRequest, 0)
```
mansinahar: Please change the capacity for this slice to be `len(request.Imp)` as you expect to create that many video requests

adapters/beachfront/beachfront.go
```diff
@@ -481,53 +520,52 @@ func postprocessBanner(beachfrontResp []BeachfrontResponseSlot, id string) ([]op
...
+	return ""
 }
 
 func extractVideoCrid(nurl string) string {
 	chunky := strings.SplitAfter(nurl, ":")
 	return strings.TrimSuffix(chunky[2], ":")
 }
 
-// Thank you, brightroll.
+func removeSlot(s []beachfrontSlot, i int) []beachfrontSlot {
```
mansinahar: I believe you can remove this method now since it's not being used anywhere

adapters/beachfront/beachfront.go
```diff
@@ -24,41 +22,45 @@ const VideoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id="
...
 type BeachfrontAdapter struct {
 }
 
-type BeachfrontRequests struct {
-	Banner BeachfrontBannerRequest
-	Video  BeachfrontVideoRequest
+type beachfrontRequests struct {
+	Banner beachfrontBannerRequest
+	Video  []beachfrontVideoRequest
+	Audio  openrtb.Audio
```
mansinahar: Out of curiosity, why is `Audio` and `Native` added here when your adapter only supports banner and video ads?

mansinahar:
Looks like the `prebid_server` binary got added to this PR by mistake. Can you please remove that? 

Also, can I ask the reasoning behind removing the test case `multi-mix.json`?

adapters/beachfront/beachfront.go
```diff
@@ -111,300 +113,336 @@ type BeachfrontResponseSlot struct {
...
+		bytes, err := json.Marshal(beachfrontRequests.Video[i])
+
+		if err == nil {
+			reqs = append(reqs, &adapters.RequestData{
+				Method:  "POST",
+				Uri:     VideoEndpoint + beachfrontRequests.Video[i].AppId + VideoEndpointSuffix,
+				Body:    bytes,
+				Headers: headers,
+			})
+		} else {
```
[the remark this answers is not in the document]
muncha: yep. Left behind from when I moved that block.

adapters/beachfront/beachfront.go
muncha: Unfounded optimism.

adapters/beachfront/beachfront.go
```diff
@@ -348,12 +340,12 @@ func getBannerRequest(request *openrtb.BidRequest) (beachfrontBannerRequest, []e
 	bfBannerRequest.Domain = site.Domain
 
 	if request.User != nil {
-		if bfBannerRequest.User.ID == "" {
-			bfBannerRequest.User.ID = request.User.ID
+		if request.User.ID == "" {
+			request.User.ID = request.User.ID
```
mansinahar: I believe you still want this line to be: `bfBannerRequest.User.ID = request.User.ID` and similar for L348

mansinahar:
Looks like gofmt is unhappy. Can you please run it on your branch and update the PR?

adapters/beachfront/beachfront.go
```diff
@@ -111,300 +113,336 @@ type BeachfrontResponseSlot struct {
 	Adm   string  `json:"adm"`
 }
 
-func NewBeachfrontBannerRequest() BeachfrontBannerRequest {
-	r := BeachfrontBannerRequest{}
+func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var beachfrontRequests beachfrontRequests
+	var errs = make([]error, 0, len(request.Imp))
+	var reqs = make([]*adapters.RequestData, len(beachfrontRequests.Video))
```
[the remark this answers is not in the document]
muncha: That sure does make sense, and I moved the reqs = make... line just below preprocess. I got  a run time panic: "runtime error: invalid memory address or nil pointer dereference"

I've moved it back for now, but i'll look in to this once I have the rest cleared up, be cause it sure does seem like it would be an improvement if I can work out where that is happening.

adapters/beachfront/beachfront.go
muncha: Yes. I got in a hurry and made a dumb mistake.

adapters/beachfront/beachfront.go
muncha: I was messing around with them a bit and didn't catch that they were still in the enum declaration.

muncha:
As for multi-mix.json (and simple-mix.json) see the description [ea6ee2a](url)

I'm looking into that now.

muncha:
@hhhjort @mansinahar Thank you for all of your help on this. I hope that your very helpful advice was not the result of too much frustration on your part. I'm wondering where we're at on this, considering:  
- Since getting to where I had incorporated all of @mansinahar 's comments  (~28th), Beachfront has gotten a real RTB video endpoint set up for this. In another branch, off of this one, I have everything set up to handle that endpoint, as well as the one here.  A lot of the streamlining and optimization changes in this PR made that a pretty straight forward task.
- There are test cases that really should have passed but were instead giving me weird errors. Removing them got validation.sh to pass, but obviously either something's wrong with the test cases or with the bidder.

Should I put in issues for those? Is there anything else with this PR? Should it be merged, or should it be closed in favor of a PR that covers the points above?

mansinahar:
@muncha This just got delayed due to the long holiday weekend. I will find some time to get a review on this tomorrow.

muncha:
@mansinahar no worries - I've got one of these problematic test cases worked out (I wasn't including the status in the mockResponse - doh) and close on the other, I hope. So I'll push when I have both of those, and I'll save my RTB video implementation for another PR.

mansinahar:
@muncha Awesome! Just lemme know when it is ready to review again and I will be happy to take a look. Thanks :)

muncha:
@mansinahar LGTM

hhhjort:
Is this ready for review now?

muncha:
Yes, please.

muncha:
@hhhjort - I'm not sure if this should be merged to master of if it should wait for my next PR, which includes all of this. It's no rush as far as I'm concerned. Unless there is a reason to merge as far as repo history sanity or similar, I'd think wait.  My next PR enables options for ADM vs. nurl responses on video. I should have that pushed in the next day or two.  Assuming waiting, should I go ahead and close this?

hhhjort:
i am good with waiting

mansinahar:
@muncha I am good with waiting too. Let's close this once you have the other PR up

muncha:
I'm closing this and submitting a new pull request which includes these changes and adds the ability to return AdM or nurl responses for video.

PR #1000, opened 2019-08-08
Tweaking stalebot

hhhjort:


mansinahar:
Just curious, would this update the Issues that have already been labeled `stale`?

hhhjort:
I have been manually un-staling the affected issues.

PR #1001, opened 2019-08-09
AMP passes gdpr consent string through endpoint

guscarreon:
Now that AMP project supports macros for passing GDPR consent state and consent string, PBS needs to be able to process this information. See [https://github.com/ampproject/amphtml/issues/22197](https://github.com/ampproject/amphtml/issues/22197) (also linked from [https://github.com/prebid/prebid.github.io/issues/1303](https://github.com/prebid/prebid.github.io/issues/1303)) for the consent string solution on the AMP side

[https://github.com/ampproject/amphtml/pull/22299](https://github.com/ampproject/amphtml/pull/22299) introduces the `CONSENT_STRING` macro. `CONSENT_STRING` and the pre-existing `CONSENT_STATE` appear to provide, in the AMP query string, values that correspond to request.user.ext.consent and `request.regs.ext.gdpr`, respectively.

It seems we can now parse the AMP request's query string to obtain these two values to add to the request that is sent to bidders.


endpoints/openrtb2/amp_auction.go
```diff
@@ -363,9 +366,25 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
...
+	if queryConsentString != "" {
+		// If nil, initialize
+		if req.User == nil {
+			req.User = &openrtb.User{}
+		}
+		jsonMsg, err := json.Marshal(&openrtb_ext.ExtUser{Consent: queryConsentString})
+		if err != nil {
+			return err
+		}
+		req.User.Ext = jsonMsg
```
hhhjort: This will overwrite anything else that exists in `user.ext` One possibility is to edit the raw json with jsonparser, which is already part of the project. Another option might be to use json-patch, which we already use for resolving stored requests.

guscarreon: Done but used the `regexp` library instead.

endpoints/openrtb2/amp_auction.go
```diff
@@ -369,15 +370,15 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
...
 	if queryConsentString != "" {
+		jsonMsg := json.RawMessage(`{"consent":"` + queryConsentString + `"}`)
 		// If nil, initialize
 		if req.User == nil {
-			req.User = &openrtb.User{}
+			req.User = &openrtb.User{Ext: jsonMsg}
+		} else { // else, keep whatever is in there and only substitute the consent string
+			re := regexp.MustCompile("^(.*consent\":\"?)[^,]*(\",.*)$")
+			repStr := "${1}" + queryConsentString + "$2"
+			req.User.Ext = json.RawMessage(re.ReplaceAllString(string(req.User.Ext), repStr))
```
hhhjort: Won't this only work if the consent object already exists in the `ext`?

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -127,9 +128,61 @@ func TestConsentThroughEndpoint(t *testing.T) {
...
+		},
+	}
+	fullMarshaledBidRequest, err := json.Marshal(bidRequest)
+	if err != nil {
+		t.Fatalf("Failed to marshal the complete openrtb.BidRequest object %v", err)
+	}
+
 	stored := map[string]json.RawMessage{
-		"1": json.RawMessage(validRequest(t, "gdpr.json")),
+		"1": json.RawMessage(json.RawMessage(fullMarshaledBidRequest)),
```
hhhjort: what does the double `json.RawMessage` do?

endpoints/openrtb2/amp_auction.go
guscarreon: You are right, adding an else clause there in case `req.User` is not null but `req.User.Ext` is.

endpoints/openrtb2/amp_auction_test.go
guscarreon: Corrected

endpoints/openrtb2/amp_auction.go
hhhjort: I don't think the replace will work if the `consent` object doesn't already exist.

guscarreon: Corrected Hans

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,6 +123,131 @@ func TestAMPPageInfo(t *testing.T) {
...
+	// Assert our bidRequest had a valid "User" field
+	if !assert.NotNil(t, exchange.lastRequest.User) {
+		return
+	}
+	// Assert our bidRequest had a valid "User.Ext" field
+	if !assert.NotNil(t, exchange.lastRequest.User.Ext) {
+		return
+	}
+
+	re := regexp.MustCompile(`\"consent\":\".*\"}$`)
```
mansinahar: I think we can get away with doing the regexp matching for the consent string here since you're already unmarshaling the JSON for `exchange.lastRequest.User.Ext` on L181 and checking if the expected consent string is present.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,6 +123,131 @@ func TestAMPPageInfo(t *testing.T) {
...
+	var ue openrtb_ext.ExtUser
+	err = json.Unmarshal(exchange.lastRequest.User.Ext, &ue)
+	assert.NoError(t, err)
+
+	// Assert consent string found in `http.Request` was passed correctly to the `User.Ext` object
+	assert.Contains(t, string(request.URL.RawQuery), consentString)
+	assert.Equal(t, consentString, ue.Consent)
+
+	// Assert other user properties found originally in our bid request such as `DigiTrust` were not overwritten
+	assert.Equal(t, "digitrustId", ue.DigiTrust.ID)
```
mansinahar: A better approach here would be to do `assert.Equal(t, fullMarshaledBidRequest.User.Ext.DigiTrust.ID, ue.DigiTrust.ID)` rather than using the hardcoded string. Also, would be good to add some error messages for all these asserts.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -169,13 +169,71 @@ func TestConsentThroughEndpoint(t *testing.T) {
...
 	}
 
-	re := regexp.MustCompile(`\"consent\":\".*\"}$`)
+	// Assert string `consent` is found in the User.Ext at all
+	assert.NotContainsf(t, fullMarshaledBidRequest, "consent:"+consentString, fmt.Sprintf("Expected bid request to contain consent string %s \n", consentString))
+
+	// Assert the last request has a valid User object with a consent string equal to that on the URL query
+	var ue openrtb_ext.ExtUser
+	err = json.Unmarshal(exchange.lastRequest.User.Ext, &ue)
+	assert.NoError(t, err)
```
hhhjort: You still have some asserts with no error messages.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -476,3 +718,75 @@ func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.B
 
 	return response, nil
 }
+
+func getTestBidRequest(nil_user bool, nil_ext bool, consent_string string, digitrust_ID string) ([]byte, error) {
```
mansinahar: Please change the parameter names to be in camel case per Go's naming conventions

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,6 +122,248 @@ func TestAMPPageInfo(t *testing.T) {
...
+	var ue openrtb_ext.ExtUser
+	err = json.Unmarshal(exchange.lastRequest.User.Ext, &ue)
+	assert.NoError(t, err, "Error unmarshalling last processed request")
+
+	// Assert consent string found in `http.Request` was passed correctly to the `User.Ext` object
+	assert.Contains(t, string(request.URL.RawQuery), consentString, "http.Request should come with a consent string in its query")
+	assert.Equal(t, consentString, ue.Consent, "Consent string unsuccessfully passed to bid request through AMP endpoint")
+}
+
+func TestSubstituteRequestConsentWithEndpointConsent(t *testing.T) {
```
mansinahar: I think it would be better to convert all these tests into table test format in order to reduce the redundant code for setup and assertions.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,6 +122,248 @@ func TestAMPPageInfo(t *testing.T) {
...
+	if !assert.NotNil(t, exchange.lastRequest.User, "Resulting bid request should have a valid User field after passing consent string through endpoint") {
+		return
+	}
+	// Assert our bidRequest had a valid "User.Ext" field
+	if !assert.NotNil(t, exchange.lastRequest.User.Ext, "Resulting bid request should have a valid Ext field after passing consent string through endpoint") {
+		return
+	}
+
+	// Assert string `consent` is found in the User.Ext at all
+	assert.NotContainsf(t, fullMarshaledBidRequest, "consent:"+consentString, fmt.Sprintf("This bid request should not contain a consent string. It will be passed the one in the http.Request endpoint\n"))
```
mansinahar: How is `NotContainsf` different from `NotContains`?

Also, `fmt.Sprintf` isn't needed here for the error message as there's no formatting being done.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,6 +122,248 @@ func TestAMPPageInfo(t *testing.T) {
...
+	if !assert.NotNil(t, exchange.lastRequest.User, "Resulting bid request should have a valid User field after passing consent string through endpoint") {
+		return
+	}
+	// Assert our bidRequest had a valid "User.Ext" field
+	if !assert.NotNil(t, exchange.lastRequest.User.Ext, "Resulting bid request should have a valid Ext field after passing consent string through endpoint") {
+		return
+	}
+
+	// Assert string `consent` is found in the User.Ext at all
+	assert.NotContainsf(t, fullMarshaledBidRequest, "consent:"+consentString, fmt.Sprintf("Expected bid request to contain consent string %s \n", consentString))
```
mansinahar: You don't need to use `fmt.Sprintf` here. You can just pass the error message and the `consentString` as arguments to the `NotContains` function and it will automatically format the error message for you by substituting the `consentString` value in place of `%s`

mansinahar:
LGTM except the minor comments mentioned above and also I'd suggest we add one more test case where the URL doesn't contain the `gdpr_consent` and assert that nothing is changed in the request as a result of that.

endpoints/openrtb2/amp_auction_test.go
guscarreon: done

endpoints/openrtb2/amp_auction_test.go
guscarreon: `NotContainsf()` accepts formating the error message like:

endpoints/openrtb2/amp_auction_test.go
guscarreon: I agree `fmt.Sprintf` is not needed, removed

endpoints/openrtb2/amp_auction_test.go
guscarreon: In this specific instance, I agree that `NotContains()` suffices. Using `NotContains()` over `NotContainsf()`

Issue #351, opened 2018-02-16
Cache host not supplied in current targeting parameters


hhhjort:
Currently when doing server side caching, we only send the `hb_cache_id` in the targeting parameters, and don't send any other information about the cache server in the response. That means any prebid server consumer needs to be hardcoded for the particular prebid server/cache that it is using. Depending on how the consumers using cache are implemented, (prebid mobile/AMP/?) this may limit code reuse and mobility. Logging an issue here to be resolved once we have a better understanding of the use cases that would benefit from a portable cache implementation.

Xoben:
Anyone know how to setup an amp config on prebid server?
When I test the placement via this link https://prebid.adnxs.com/pbs/v1/openrtb2/amp?tag_id=11111

I get the following error message

**Invalid request format: AMP requests require Targeting and Caching to be set**

`{
    "imp": [
        {
            "ext": {
                "appnexus": {
                    "placementId": 11111
                }
            },
            "banner": {
                "format": [
                    {
                        "h": 250,
                        "w": 300
                    },
                    {
                        "h": 600,
                        "w": 300
                    }
                ]
            },
            "id": "imp-div-gpt-300x250_0"
        }
    ],
    "id": "div-gpt-300x250_0"
}`

I can't seem to find any documentation on how to setup OpenRTB bid requests for AMP

dbemiller:
@Xoben please open up a new issue for unrelated questions.

moonshells:
@hhhjort Looks like DFP only accepts hardcoded `host` when we set up VAST tag URL, we can not use any "dynamic" cache host in VAST tag URL. Is prebid mobile or AMP integrated with DFP? @dbemiller Please correct me if I'm wrong.

guscarreon:
The following proposed solution has been discussed with @hhhjort and @jaiminpanchal27 :
   - The PBS host configures a prebid-cache URL in the configure-the-server stage.
   - We read those values into our `Configuration` struct in `config/config.go`
   - We create the pair/value targets for
        "hb_cache_host" --> "example.com"
        "hb_cache_path" --> "/dirname/endpoint"
Please let me know if you agree with this proposal in order to push a PR

hhhjort:
@bretg : Making sure you see this.

PR #1002, opened 2019-08-14
Add cache host to targeting params

guscarreon:
The following solution refers to issue #351  Cache host not supplied in current targeting parameters codes pair/value targets for
 - "hb_cache_host" --> "example.com"
 - "hb_cache_path" --> "/dirname/endpoint"


guscarreon:
Wrong branch, will recommit from the right one

PR #1003, opened 2019-08-15
Add cache host to targeting params

guscarreon:
The following proposed solution creates the pair/value targets for
 - "hb_cache_host" --> "example.com"
 - "hb_cache_path" --> "/dirname/endpoint"

As discussed in issue [#351](https://github.com/prebid/prebid-server/issues/351) Cache host not supplied in current targeting parameters


exchange/targeting_test.go
```diff
@@ -54,6 +56,11 @@ func TestTargetingCache(t *testing.T) {
 
 	assertKeyExists(t, bids["losing-bid"], string(openrtb_ext.HbCacheKey), false)
 	assertKeyExists(t, bids["losing-bid"], openrtb_ext.HbCacheKey.BidderKey(openrtb_ext.BidderAppnexus, maxKeyLength), false)
+
+	//assert hb_cache_host and hb_cache_path were included
+	assert.Equal(t, true, strings.Contains(string(bids["winning-bid"].Ext), string(openrtb_ext.HbConstantCacheHostKey)) && strings.Contains(string(bids["winning-bid"].Ext), "www.pbcserver.com"))
```
mansinahar: The `assert` package has a `contains` [method](https://godoc.org/github.com/stretchr/testify/assert#Contains) that will do exactly what you want here and you don't have to assert over a bool value.

prebid_cache_client/client.go
```diff
@@ -40,6 +47,7 @@ type Cacheable struct {
 }
 
 func NewClient(conf *config.Cache) Client {
+	PrebidCacheURL = conf.GetBaseURL() + "/cache"
```
mansinahar: Shouldn't we make the `path` configurable too? @guscarreon @hhhjort

exchange/targeting_test.go
guscarreon: Done. But have not updated the PR because I don't know if the `path` needs to be configurable as you mentioned in your second comment. @hhhjort

prebid_cache_client/client.go
guscarreon: Implementing the configurable feature...

guscarreon: Implementing the configurable feature..

guscarreon: done

exchange/exchange.go
```diff
@@ -139,6 +139,7 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 			}
 		}
 	}
+	targData.cacheHost, targData.cachePath = e.cache.GetPrebidCacheSplitURL()
```
hhhjort: We only need cacheHost and cachePath if we are supplying other targData, so it can go into the if above and you don't need to precreate a targData struct

mansinahar:
Spoke with @guscarreon offline about some changes and he's going to update the PR when he has them ready

exchange/exchange.go
guscarreon: You are right Hans, this was a bug. Corrected

prebid_cache_client/client.go
```diff
@@ -47,13 +50,21 @@ func NewClient(conf *config.Cache) Client {
 				IdleConnTimeout: 65,
 			},
 		},
-		putUrl: conf.GetBaseURL() + "/cache",
+		putUrl:  conf.GetBaseURL(true, false),
+		urlHost: conf.GetHost(),
+		urlPath: conf.GetPath(),
```
mansinahar: Why do we need `urlHost` & `urlPath` here when we have the `putUrl` which is basically made of those values? I thought we agreed on just having the `putUrl` field here and changing the `GetBaseURL` method to return the constructed URL string (should probably change the name of that new method to `GetURL`). And then for the targeting data, you can alter the `GetPrebidCacheSplitURL` method to parse the string into a `url.URL` object and return the `host` and the `path` values. 

Or an even better approach would be to just change the `putUrl` type to be `url.URL` so that you don't have to switch between a `ur.URL` object and it's string representation back and forth

router/router.go
```diff
@@ -218,7 +218,7 @@ func New(cfg *config.Configuration, rateConvertor *currencies.RateConverter) (r
 	gdprPerms := gdpr.NewPermissions(context.Background(), cfg.GDPR, adapters.GDPRAwareSyncerIDs(syncers), theClient)
 
 	exchanges = newExchangeMap(cfg)
-	theExchange := exchange.NewExchange(theClient, pbc.NewClient(&cfg.CacheURL), cfg, r.MetricsEngine, bidderInfos, gdprPerms, rateConvertor)
+	theExchange := exchange.NewExchange(theClient, pbc.NewClient(&cfg.CacheURL, &cfg.ExtCacheURL), cfg, r.MetricsEngine, bidderInfos, gdprPerms, rateConvertor)
```
mansinahar: Do we need to pass the `ExtCacheURL` to the cache client? Because it doesn't seem like the client itself uses the external cache URL for anything. Cache client is only interested in the internal URL to be able to connect to the cache. So in that case, any reason why we can't just directly pass `cfg.ExtCacheURL` as it is to the `NewExchange` function and then it can use it directly rather than trying to get it from the client like this: `e.cache.GetExtCacheData()`

guscarreon: The reason we access the host and the path via `e.cache.GetExtCacheData()` is that `e.cache` is an interface so, to my knowledge we can't just do:
```
pathThatWillBeTargeted = e.cache.ExtCacheURL.Path
hostThatWillBeTargeted = e.cache.ExtCacheURL.Host
```
We must access that data via a function, hence, the need of declaring `GetExtCacheData() (string, string)`.
```
19 type Client interface {
20     // PutJson stores JSON values for the given openrtb.Bids in the cache. Null values will be
21     //
22     // The returned string slice will always have the same number of elements as the values argument. If a
23     // value could not be saved, the element will be an empty string. Implementations are responsible for
24     // logging any relevant errors to the app logs
25     PutJson(ctx context.Context, values []Cacheable) ([]string, []error)
26  +
27  +  // Serves the purpose of a getter that returns the host and the cache of the prebid-server URL
28  +  GetExtCacheData() (string, string)
29 }
prebid_cache_client/client.go
```
The `clientImpl` object implements `GetExtCacheData()` to extract those values like this
```
59 type clientImpl struct {
60     httpClient        *http.Client
61     putUrl            string
62  +  externalCacheHost string
63  +  externalCachePath string
64 }
65
66 + func (c *clientImpl) GetExtCacheData() (string, string) {
67 +     return c.externalCacheHost, c.externalCachePath
68 + }
prebid_cache_client/client.go
```
Test file `exchange/auction_test.go` defines `mockCache` to implement the `Cache` interface too but their `GetExtCacheData() (string, string)` gets implemented differently:
```
253 + func (c *mockCache) GetExtCacheData() (string, string) {
254 +     return "", ""
255 + }
256 func (c *mockCache) PutJson(ctx context.Context, values []prebid_cache_client.Cacheable) ([]string, []error) {
257     c.items = values
258     return []string{"", "", "", "", ""}, nil
259 }
exchange/auction_test.go
```

guscarreon: Do you think there's a better way to access that data?

mansinahar: Maybe it wasn't very clear from my comment but I didn't mean to say to access it via e.Cache. What I meant was that we, send `ExtCacheURL` directly to the `NewExchange` method and then store it in the exchange object and in `HoldAuction` method you can directly access it via the exchange object. Of course, you will have to modify the `Exchange` struct a little bit for that. 

But on further thought, your approach is also fine too. I just wish we called that interface `Cache` rather than `Client`.

prebid_cache_client/client_test.go
```diff
@@ -113,6 +117,45 @@ func TestEncodeValueToBuffer(t *testing.T) {
...
+`), expectedHost: "prebid-server.prebid.org", expectedPath: "pbcache/endpoint"},
+		{inConfig: []byte(`
+external_cache:
+  host: prebidcache.net
+`), expectedHost: "prebidcache.net", expectedPath: ""},
+		{inConfig: []byte(``), expectedHost: "", expectedPath: ""},
+	}
+	for i, test := range testInput {
+		//start viper
+		v := viper.New()
```
mansinahar: Only recommendation I have is that here, you don't need to setup Viper and do a `ReadConfig` to read the config and then pass it on to the `NewClient` function because we're already testing Viper's capability to be able to do that properly in `config_test.go` file. Here, we just want to test the `NewClient` function and so we can directly create a `Cache` obj and `ExtCacheURL` object and pass them to `NewClient`

mansinahar:
Other than the minor change in the test above, LGTM @guscarreon

prebid_cache_client/client_test.go
guscarreon: Removed Mansi

PR #1004, opened 2019-08-19
limit the size of the uids cookie

guscarreon:
We hit an issue where the header (containing the uids cookie) overran the limit that Ingress could handle. This pull request checks the size of the `PBSCookie` and, if found bigger than 32 KB in its `http` form, traverses the `PBSCookie.uids` field deleting the entries closest to expiration date.


usersync/cookie.go
```diff
@@ -45,26 +46,78 @@ type uidWithExpiry struct {
...
+			var pc *PBSCookie = NewPBSCookie() //The config.HostCookie cookie was found so, ignore it? Declare a new and blank PBSCookie
 			pc.SetPreference(false)
 			return pc
 		}
 	}
+	// There's no OptOutCookie defined or at least our incoming request doesn't have that OptOutName in any http.Request.Header
+	// Let's try to download the request's cookie data into a fresh PBSCookie
 	var parsed *PBSCookie
-	uidCookie, err2 := r.Cookie(UID_COOKIE_NAME)
+	uidCookie, err2 := r.Cookie("uids") // hate the refactoring Mansi did, in real code that value is: UID_COOKIE_NAME
```
hhhjort: The downside here is you now have "`uids`" defined twice. If you ever have to change it, you will need to edit two lines in this file. This can be error prone. What is the advantage of not using the `UID_COOKIE_NAME` constant?

usersync/cookie.go
```diff
@@ -162,9 +215,31 @@ func (cookie *PBSCookie) GetId(bidderName openrtb_ext.BidderName) (id string, ex
 // SetCookieOnResponse is a shortcut for "ToHTTPCookie(); cookie.setDomain(domain); setCookie(w, cookie)"
 func (cookie *PBSCookie) SetCookieOnResponse(w http.ResponseWriter, domain string, ttl time.Duration) {
 	httpCookie := cookie.ToHTTPCookie(ttl)
+	var MAX_COOKIE_SIZE int = 1 << 15 // 32 KB
```
hhhjort: We will want this as a config parameter rather than hardcoded. Also a check for a 0 "size" to disable enforcement of a maximum size may not be bad.

usersync/cookie.go
```diff
@@ -162,9 +215,31 @@ func (cookie *PBSCookie) GetId(bidderName openrtb_ext.BidderName) (id string, ex
...
 	if domain != "" {
 		httpCookie.Domain = domain
 	}
+
+	var currSize int = len([]byte(httpCookie.String()))
+	for currSize > MAX_COOKIE_SIZE {
+		var oldestElem string = ""
+		var oldestDate int64 = math.MaxInt64
+		for key, value := range cookie.uids {
+			var timeSince time.Duration = now.Sub(value.Expires)
```
hhhjort: You shouldn't need to specify `time.Duration` as that is already defined as the return type of `now.Sub()`

hhhjort: Also, since you are taking the current time, and subtracting the expirery time, this will result in a negative number. The most negative number will be the expires time that is furthest from now, so in the end we will remove the entry that has the longest time before it expires.

usersync/cookie.go
guscarreon: That was not meant to make it to the PR, I'm sorry. I don't hate the refactoring, I just found it a little confusing, let me rollback.

usersync/cookie.go
guscarreon: Do you mean get it from the `configmap.yaml` file itself? Bring it all the way here straight from the `Configuration struct` found in `config/config.go`?

guscarreon: What do you mean check for a 0 size? A 0 size `httpCookie.String()`?

usersync/cookie.go
guscarreon: Gosh, I got the `Sub()` function backwards, corrected

usersync/cookie.go
hhhjort: Yes, bring the config struct, or the relevant value, all the way down to here. It is the configured value for `MAX_COOKIE_SIZE` you should check for a 0 value, as a method of turning off this feature. If a host limits the number of bidders they support, they can limit the size the cookie can grow to.

usersync/cookie.go
```diff
@@ -30,9 +31,10 @@ var bidderToFamilyNames = map[openrtb_ext.BidderName]string{
 // To get an instance of this from a request, use ParsePBSCookieFromRequest.
 // To write an instance onto a response, use SetCookieOnResponse.
 type PBSCookie struct {
-	uids     map[string]uidWithExpiry
-	optOut   bool
-	birthday *time.Time
+	uids         map[string]uidWithExpiry
+	optOut       bool
+	birthday     *time.Time
+	maxSizeBytes int
```
mansinahar: I don't think we need `maxSizeBytes` as part of the `PBSCookie` struct. We can just access the `MaxCookieSizeBytes` in the `HostCookie` struct and pass it on in the call to `SetCookieOnResponse`. This should make the code a little simpler. 

Spoke with @guscarreon offline about this and he said he will make the suggested change.

usersync/cookie.go
```diff
@@ -162,9 +165,30 @@ func (cookie *PBSCookie) GetId(bidderName openrtb_ext.BidderName) (id string, ex
...
 	if domain != "" {
 		httpCookie.Domain = domain
 	}
+
+	var currSize int = len([]byte(httpCookie.String()))
+	for cookie.maxSizeBytes > 0 && currSize > cookie.maxSizeBytes {
+		var oldestElem string = ""
+		var oldestDate int64 = math.MaxInt64
+		for key, value := range cookie.uids {
+			timeUntilExpiration := value.Expires.Sub(now)
```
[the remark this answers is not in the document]
guscarreon: done

usersync/cookie.go
guscarreon: done

usersync/cookie_test.go
```diff
@@ -333,31 +385,387 @@ func ensureConsistency(t *testing.T, cookie *PBSCookie) {
...
+			"key7": newTempId("abcdefghijklmnopqrstuvwxyz", 2),
+		},
+		optOut:   false,
+		birthday: timestamp(),
+	}
+	return mediumSizeCookie, len(mediumSizeCookie.uids)
+}
+
+func newBigCookie() (*PBSCookie, int) {
+	var bigCookie *PBSCookie = &PBSCookie{
```
mansinahar: How about creating a test folder and adding these different cookies there in `.txt` files and then using those in the tests rather than populating this test file with a giant map?

mansinahar: Spoke with @guscarreon about this and we probably don't need this particular test case to start with and so he's gonna evaluate and remove this if not needed.

config/config.go
```diff
@@ -679,3 +690,12 @@ func setBidderDefaults(v *viper.Viper, bidder string) {
 	v.SetDefault(adapterCfgPrefix+bidder+".disabled", false)
 	v.SetDefault(adapterCfgPrefix+bidder+".partner_id", "")
 }
+
+func isValidCookieSize(maxCookieSize int) error {
+	// If a non-zero-less-than-500-byte "host_cookie.max_cookie_size_bytes" value was specified in the
+	// environment configuration of prebid-server, default to 500 bytes
+	if maxCookieSize != 0 && maxCookieSize < MIN_COOKIE_SIZE_BYTES {
+		return errors.New("Configured cookie size is less than allowed minimum size")
```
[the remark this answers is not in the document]
guscarreon: Makes perfect sense, done

usersync/cookie_test.go
```diff
@@ -262,51 +262,32 @@ func TestGetUIDsWithNilCookie(t *testing.T) {
...
 	}
 	for i := range testCases {
 		processedCookie := writeThenRead(cookieToSend, testCases[i].maxCookieSize)
 		switch testCases[i].expAction {
 		case "equal":
-			assert.Equal(t, cookieToSendLen, len(processedCookie.uids), "[Test %d] MaxCookieSizeBytes equal to zero or bigger than %d bytes should be snough to set and remain cookie unchanged \n", i+1, len(processedCookie.uids))
-			_, oldestFound := processedCookie.uids["key7"]
+			assert.Equal(t, cookieToSendLen, len(processedCookie.uids), "[Test %d] MaxCookieSizeBytes equal to zero or bigger than %d bytes should be enough to set and remain cookie unchanged \n", i+1, len(processedCookie.uids))
+			_, oldestFound := processedCookie.uids[closestToExpirationDate]
 			assert.Equal(t, oldestFound, true, "[Test %d] Oldest entry in cookie should not have been eliminated", i+1)
```
mansinahar: You can just do `assert.Contains` here on the map and skip the previous line.

usersync/cookie_test.go
```diff
@@ -262,51 +262,32 @@ func TestGetUIDsWithNilCookie(t *testing.T) {
...
 		case "trim":
 			assert.Equal(t, cookieToSendLen > len(processedCookie.uids), true, "[Test %d] MaxCookieSizeBytes of %d is smaller than %d bytes and cookie entries should have been removed\n", i+1, testCases[i].maxCookieSize, cookieToSendLen)
-			_, oldestFound := processedCookie.uids["key7"]
+			_, oldestFound := processedCookie.uids[closestToExpirationDate]
 			assert.Equal(t, oldestFound, false, "[Test %d] Oldest entry in cookie was not eliminated", i+1)
```
mansinahar: Same as above

usersync/cookie_test.go
guscarreon: Done

guscarreon: Done

PR #1006, opened 2019-08-19
Shortening the query string params, adding BIDID macro

bretg:
The goal is to save on the number of bytes we're sending to the ad server. The differences between this and the original version are:
- shortening query string params to one char
- removed the bidder param
- added BIDID as a macro to hb_winurl

Original version is at https://github.com/prebid/prebid-server/files/2946568/Prebid.Server.Event.Notifications.-.Tech.Spec.pdf


bretg:
As for GET-vs-POST. I get your point, but:
- POSTs cause the browser to do two fetches: OPTIONS and then POST
- The most common mechanism of dropping a tracking pixel is an  which is a GET.

So we shouldn't change to POST. Did made some modifications and moved it to github wiki -- https://github.com/prebid/prebid-server/wiki/Prebid-Server-Event-Notification-Tech-Spec

- changed 'view' event to 'imp' and explained
- added new column in accounts table in the DB

PR #1011, opened 2019-08-26
[Timebox: 2d] Address Sonar Analysis issues - part 1

guscarreon:
We ran Sonar's static analysis on our code and found a bunch of issues that'd be great to address. This is the first part of a batch of 76 issues to correct in 32 files in the `prebid-server` codebase.


guscarreon:
Let's remove the stale label

endpoints/cookie_sync.go
```diff
@@ -152,6 +134,32 @@ func (deps *cookieSyncDeps) Endpoint(w http.ResponseWriter, r *http.Request, _ h
 	enc.Encode(csResp)
 }
 
+func parseRequest(parsedReq *cookieSyncRequest, bodyBytes []byte, usersyncIfAmbiguous bool, co *analytics.CookieSyncObject, w http.ResponseWriter) bool {
```
SyntaxNode: Rather than returning a bool, consider returning an Error and consolidating the error handling at the call site. This will also remove 2 arguments.

endpoints/openrtb2/auction.go
```diff
@@ -1128,16 +1123,27 @@ func setUAImplicitly(httpReq *http.Request, bidReq *openrtb.BidRequest) {
 	}
 }
 
-// Check if a request comes from a Safari browser
-func checkSafari(r *http.Request) (isSafari bool) {
-	isSafari = false
+// parseUserId gets this user's ID  for the host machine, if it exists.
```
SyntaxNode: Nitpicks: Please remove extra space after `user's ID`. Please update comment to use the same casing as the func name (parseUserID  vs parseUserId).

endpoints/openrtb2/auction.go
```diff
@@ -1128,16 +1123,27 @@ func setUAImplicitly(httpReq *http.Request, bidReq *openrtb.BidRequest) {
...
+		return hostCookie.Value, true
+	} else {
+		return "", false
+	}
+}
+
+// checkBrowserName checks if a request comes from a Safari browser.
+// Returns pbsmetrics.BrowserSafari or pbsmetrics.BrowserOther
+// depending on the value of the "User-Agent" header of our http.Request
+func checkBrowserName(r *http.Request) pbsmetrics.Browser {
```
SyntaxNode: Consider renaming to something along the lines of `parseBrowser`. The "check" part of the name conveys an expectation for a boolean return, IMHO.

exchange/exchange.go
```diff
@@ -590,3 +580,29 @@ func (e *exchange) makeBid(Bids []*pbsOrtbBid, adapter openrtb_ext.BidderName) (
...
+	// Snapshot of resolved bid request for debug if test request
+	var resolvedRequest json.RawMessage = nil
+	var err error
+
+	if bidRequest.Test == 1 {
+		if r, err := json.Marshal(bidRequest); err == nil {
+			resolvedRequest = r
+		}
+	}
+	return resolvedRequest, err
```
SyntaxNode: I think this func should return an error if the call to `json.Marshal` fails, but that's not what happens. A new err variable is being declared on line 590 local to the `if` statement and the err defined on line 587 will never be set. Consider something like:

```
if bidRequest.Test == 1 {
    return json.Marshal(bidRequest);
}

return nil, nil
```

.. or something like that. Not sure if you need to do manual casting between `byte[]` and `json.RawMessage`.

cache/filecache/filecache.go
```diff
@@ -77,7 +77,6 @@ func New(filename string) (*Cache, error) {
 }
 
 // Close does nothing
```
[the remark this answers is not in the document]
guscarreon: Done

endpoints/cookie_sync.go
guscarreon: If you are fine with the new error messages, finally we could:
```
 80       parsedReq := &cookieSyncRequest{}
 81 -     if !parseRequest(parsedReq, bodyBytes, deps.gDPR.UsersyncIfAmbiguous, &co, w) {
    +     if err := parseRequest(parsedReq, bodyBytes, deps.gDPR.UsersyncIfAmbiguous); err != nil {
    +         co.Status = http.StatusBadRequest
    +         co.Errors = append(co.Errors, err)
    +         http.Error(w, co.Errors[len(co.Errors)-1].Error(), co.Status)
 82           return
 83       }
 84
 85 +-- 46 lines: if len(biddersJSON) == 0 {------------------------------------------
131       w.Header().Set("Content-Type", "application/json; charset=utf-8")
132       enc := json.NewEncoder(w)
133       enc.SetEscapeHTML(false)
134       enc.Encode(csResp)
135   }
136
137 - func parseRequest(parsedReq *cookieSyncRequest, bodyBytes []byte, usersyncIfAmbiguous bool, co *analytics.CookieSyncObject, w http.ResponseWriter) bool {
    + func parseRequest(parsedReq *cookieSyncRequest, bodyBytes []byte, usersyncIfAmbiguous bool) error {
138       if err := json.Unmarshal(bodyBytes, parsedReq); err != nil {
139 -         co.Status = http.StatusBadRequest
140 -         co.Errors = append(co.Errors, fmt.Errorf("JSON parsing failed: %s", err.Error()))
141 -         http.Error(w, co.Errors[len(co.Errors)-1].Error(), http.StatusBadRequest)
142 -         return false
    +         return fmt.Errorf("JSON parsing failed: %s", err.Error())
143       }
144
145       if parsedReq.GDPR != nil && *parsedReq.GDPR == 1 && parsedReq.Consent == "" {
146 -         co.Status = http.StatusBadRequest
147 -         co.Errors = append(co.Errors, errors.New("gdpr_consent is required if gdpr=1"))
148 -         http.Error(w, co.Errors[len(co.Errors)-1].Error(), http.StatusBadRequest)
149 -         return false
    +         return errors.New("gdpr_consent is required if gdpr=1")
150       }
151       // If GDPR is ambiguous, lets untangle it here.
152       if parsedReq.GDPR == nil {
153           var gdpr = new(int)
154           *gdpr = 1
155           if usersyncIfAmbiguous {
156               *gdpr = 0
157           }
158           parsedReq.GDPR = gdpr
159       }
160       return nil
161   }
endpoints/cookie_sync.go
```

guscarreon: Would you be ok with those changes? Did I get the idea right?

endpoints/openrtb2/auction.go
guscarreon: Done

endpoints/openrtb2/auction.go
guscarreon: Do you think `getBrowserName` works better? Or are you better with `parseBrowser`?

SyntaxNode: `getBrowserName` sounds good to me. 👍

endpoints/cookie_sync.go
SyntaxNode: Yes. That's exactly what I was thinking. I like the unification of the error messages. The only risk is if someone had coded against the slightly different error message formats. Do you have any concern over that @hhhjort?

exchange/exchange.go
guscarreon: Pretty good catch. I don't need we need to cast to `[]byte`, do we? `resolvedRequest` ends up being passed as an argument to `buildBidResponse` which expects a `json.RawMessage` as it's fifth argument.
```
func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_ext.BidderName, adapterBids map[openrtb_ext.BidderName]*pbsOrtbSeatBid, bidRequest *openrtb.BidRequest, resolvedRequest json.RawMessage, adapterExtra map[openrtb_ext.BidderName]*seatResponseExtra, errList []error) (*openrtb.BidResponse, error) {
```

guscarreon: Done

endpoints/cookie_sync.go
hhhjort: If people are scripting to read the error messages, I would hope they are first checking the http status. And have something in place to flag unrecognized error messages.

guscarreon: Done

endpoints/openrtb2/auction.go
guscarreon: Awesome. Changed

guscarreon:
Pull request updated

PR #1016, opened 2019-08-29
[Sharethrough] Implement Cookie Sync + misc improvements

maphe:
### Implement Cookie Sync
- Change Cookie Sync Sharethrough URL with an actual URL (`config.go`)
- Send `request.User.BuyerUID` to the ad server

### Switch adapter version format
- Changing from `1.0.x` format to simple integer (here from `1.0.3` to `4`) for internal reason

### Ad Payload encoding
Before:
- Unmarshal JSON based on hardcoded ad payload schema (`imp_sharethrough.go`)
- Then re-Marshal it to JSON + B64 to put it in the ad markup

Now:
- JSON schema only has fields that actually matter in the adapter code
- the ad payload is kept raw so we can just B64-encode whatever we got back from the ad server

### DIFF
--- a/adapters/sharethrough/butler.go
+++ b/adapters/sharethrough/butler.go
@@ -23,6 +23,7 @@ type StrAdSeverParams struct {
 	Height             uint64
 	Width              uint64
 	TheTradeDeskUserId string
+	SharethroughUserId string
 }
 
 type StrOpenRTBInterface interface {
@@ -91,6 +92,7 @@ func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openr
 			Width:              width,
 			InstantPlayCapable: s.Util.canAutoPlayVideo(request.Device.UA, s.UserAgentParsers),
 			TheTradeDeskUserId: userInfo.TtdUid,
+			SharethroughUserId: request.User.BuyerUID,
 		}),
 		Body:    nil,
 		Headers: headers,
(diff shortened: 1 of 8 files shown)


maphe:
@guscarreon thanks for the heads up, I added a test for it.

maphe:
@guscarreon do you see any other changes I'd need to make? We were hoping this PR can make it to this week's release... Thanks

hhhjort:
We merged this, but had to roll back our production deploy. Sharethough adapter kept generating panics, this needs to be looked in to.

maphe:
@hhhjort can you provide more data about those panics please? so that I know where to look at...

adapters/sharethrough/butler.go
```diff
@@ -91,14 +92,20 @@ func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openr
 			Width:              width,
 			InstantPlayCapable: s.Util.canAutoPlayVideo(request.Device.UA, s.UserAgentParsers),
 			TheTradeDeskUserId: userInfo.TtdUid,
+			SharethroughUserId: request.User.BuyerUID,
```
guscarreon: Mathieu, your adapter went live but kept panic'ing here because of a NULL pointer de reference. Please implement a workaround. We also think a test case in `TestSuccessResponseToOpenRTB(t *testing.T)` (or anywhere in `butler_test.go` for that matter) to test for a `nil` value would be great.

maphe:
@hhhjort @guscarreon fix provided in #1023

PR #1018, opened 2019-08-29
Implement ability to require a publisher/account ID

guscarreon:
Adds feature to create a config option (boolean) to turn on/off enforcement of publisher/account ID in the resolved request for openrtb2 endpoint.

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -18,7 +18,6 @@ import (
 	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/analytics"
 	"github.com/prebid/prebid-server/config"
-	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/exchange"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
@@ -150,8 +149,8 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	}
 	labels.PubID = effectivePubID(req.Site.Publisher)
 	// Blacklist account now that we have resolved the value
-	if _, found := deps.cfg.BlacklistedAcctMap[labels.PubID]; found {
-		errL = append(errL, &errortypes.BlacklistedAcct{Message: fmt.Sprintf("Prebid-server has blacklisted Account ID: %s, pleaase reach out to the prebid server host.", labels.PubID)})
+	if acctIdErr := validateAccount(deps.cfg, labels.PubID); acctIdErr != nil {
+		errL = append(errL, acctIdErr)
 		w.WriteHeader(http.StatusBadRequest)
 		for _, err := range errL {
 			w.Write([]byte(fmt.Sprintf("Invalid request format: %s\n", err.Error())))
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -18,6 +18,7 @@ import (
 	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/analytics"
 	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/exchange"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
@@ -151,12 +152,17 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	// Blacklist account now that we have resolved the value
 	if acctIdErr := validateAccount(deps.cfg, labels.PubID); acctIdErr != nil {
 		errL = append(errL, acctIdErr)
-		w.WriteHeader(http.StatusBadRequest)
+		erVal := errortypes.DecodeError(acctIdErr)
+		if erVal == errortypes.BlacklistedAppCode || erVal == errortypes.BlacklistedAcctCode {
+			w.WriteHeader(http.StatusServiceUnavailable)
+		} else { //erVal == errortypes.AcctRequiredCode
+			w.WriteHeader(http.StatusBadRequest)
+		}
+		labels.RequestStatus = pbsmetrics.RequestStatusBadInput
 		for _, err := range errL {
 			w.Write([]byte(fmt.Sprintf("Invalid request format: %s\n", err.Error())))
 		}
 		ao.Errors = append(ao.Errors, errL...)
-		labels.RequestStatus = pbsmetrics.RequestStatusBadInput
 		return
 	}
 
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1184,13 +1184,11 @@ func validateAccount(cfg *config.Configuration, pubID string) error {
 	var err error = nil
 	// If specified in the configuration, discard requests that don't come with an account ID.
 	if cfg.AccountRequired && pubID == pbsmetrics.PublisherUnknown {
-		//return errortypes.AcctRequired{Message: fmt.Sprintf("Prebid-server has been configured to discard requests that don't come with an Account ID. Please reach out to the prebid server host.")}
 		err = error(&errortypes.AcctRequired{Message: fmt.Sprintf("Prebid-server has been configured to discard requests that don't come with an Account ID. Please reach out to the prebid server host.")})
 	}
 
 	// Blacklist account now that we have resolved the value
 	if _, found := cfg.BlacklistedAcctMap[pubID]; found {
-		//return errortypes.BlacklistedAcct{Message: fmt.Sprintf("Prebid-server has blacklisted Account ID: %s, please reach out to the prebid server host.", labels.PubID)}
 		err = error(&errortypes.BlacklistedAcct{Message: fmt.Sprintf("Prebid-server has blacklisted Account ID: %s, please reach out to the prebid server host.", pubID)})
 	}
 	return err
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -135,9 +135,9 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		}
 		labels.PubID = effectivePubID(req.Site.Publisher)
 	}
-	// Blacklist account now that we have resolved the value
-	if _, found := deps.cfg.BlacklistedAcctMap[labels.PubID]; found {
-		errL = append(errL, &errortypes.BlacklistedAcct{Message: fmt.Sprintf("Prebid-server has blacklisted Account ID: %s, pleaase reach out to the prebid server host.", labels.PubID)})
+
+	if acctIdErr := validateAccount(deps.cfg, labels.PubID); acctIdErr != nil {
+		errL = append(errL, acctIdErr)
 		writeError(errL, w)
 		labels.RequestStatus = pbsmetrics.RequestStatusBadInput
 		return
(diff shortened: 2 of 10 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -1179,3 +1179,19 @@ func effectivePubID(pub *openrtb.Publisher) string {
 	}
 	return pbsmetrics.PublisherUnknown
 }
+
+func validateAccount(cfg *config.Configuration, pubID string) error {
+	var err error = nil
+	// If specified in the configuration, discard requests that don't come with an account ID.
+	if cfg.AccountRequired && pubID == pbsmetrics.PublisherUnknown {
+		//return errortypes.AcctRequired{Message: fmt.Sprintf("Prebid-server has been configured to discard requests that don't come with an Account ID. Please reach out to the prebid server host.")}
```
mansinahar: Can you please remove this comment?

endpoints/openrtb2/auction.go
```diff
@@ -1179,3 +1179,19 @@ func effectivePubID(pub *openrtb.Publisher) string {
...
+		err = error(&errortypes.AcctRequired{Message: fmt.Sprintf("Prebid-server has been configured to discard requests that don't come with an Account ID. Please reach out to the prebid server host.")})
+	}
+
+	// Blacklist account now that we have resolved the value
+	if _, found := cfg.BlacklistedAcctMap[pubID]; found {
+		//return errortypes.BlacklistedAcct{Message: fmt.Sprintf("Prebid-server has blacklisted Account ID: %s, please reach out to the prebid server host.", labels.PubID)}
```
mansinahar: This one too

endpoints/openrtb2/video_auction.go
```diff
@@ -198,10 +198,12 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
 		labels.PubID = effectivePubID(bidReq.Site.Publisher)
 	}
-	// Blacklist account now that we have resolved the value
-	if _, found := deps.cfg.BlacklistedAcctMap[labels.PubID]; found {
-		errL := []error{&errortypes.BlacklistedAcct{Message: fmt.Sprintf("Prebid-server has blacklisted Account ID: %s, pleaase reach out to the prebid server host.", labels.PubID)}}
-		handleError(labels, w, errL, ao)
+
+	if acctIdErr := validateAccount(deps.cfg, labels.PubID); acctIdErr != nil {
+		errL = append(errL, acctIdErr)
+		writeError(errL, w)
```
mansinahar: Why aren't we just calling `handleError` here as it was before? `handleError` already does everything needed to handle an error before returning

endpoints/openrtb2/auction_test.go
```diff
@@ -227,6 +228,54 @@ func TestBlacklistRequests(t *testing.T) {
 	tests.assert(t)
 }
 
+// TestRejectAccountRequired asserts we return a 400 code on a request that comes with no user id nor app id
+// if the `AccountRequired` field in the `config.Configuration` structure is set to true
+func TestRejectAccountRequired(t *testing.T) {
+	tests := &getResponseFromDirectory{
```
mansinahar: Why not just add this test request to `endpoints/openrtb2/sample-requests`  and then create a new `getResponseFromDirectory` for those tests?

Would also suggest adding some more test cases like when account ID is not required and not provided in the request, required and provided, required and provided but blacklisted, etc.

endpoints/openrtb2/auction.go
guscarreon: removed

endpoints/openrtb2/auction.go
guscarreon: also removed

endpoints/openrtb2/auction_test.go
guscarreon: Four new test cases where added under `TestRejectAccountRequired` in `endpoints/openrtb2/auction_test.go`. With said additions, we are now covering all the blow scenarios
```
   | required | provided | blacklisted ||   expectedCode            |  test
---------------------------------------------------------------------------
 1 |   FALSE  |   FALSE  |   FALSE     || StatusOK                  |  8.1
 2 |   FALSE  |   FALSE  |    TRUE     || StatusOK                  |  8.1
 3 |   FALSE  |    TRUE  |   FALSE     || StatusOK                  |  TestGoodRequest
 4 |   FALSE  |    TRUE  |    TRUE     || StatusBadRequest          |  TestBlacklistRequests
 5 |    TRUE  |   FALSE  |   FALSE     || StatusOK                  |  8.2
 6 |    TRUE  |   FALSE  |    TRUE     || StatusOK                  |  8.2
 7 |    TRUE  |    TRUE  |   FALSE     || StatusOK                  |  8.3
 8 |    TRUE  |    TRUE  |    TRUE     || StatusServiceUnavailable  |  8.4
```

endpoints/openrtb2/sample-requests/account-required/with-acct.json
```diff
@@ -0,0 +1,67 @@
...
+      "id": "some-request-id",
+      "site": {
+        "publisher": { "id": "not_bad_acct"},
+      	"page": "test.somepage.com"
+      },
+      "user": { },
+      "imp": [
+        {
+        	"id": "my-imp-id",
+        	"banner": {
```
mansinahar: Can you please fix the formatting on this one? Thanks!

endpoints/openrtb2/sample-requests/account-required/no-acct.json
```diff
@@ -0,0 +1,66 @@
...
+      "user": { },
+      "imp": [
+        {
+        	"id": "my-imp-id",
+        	"banner": {
+        		"format": [
+        			{
+        				"w": 300,
+        				"h": 600
+        			}
```
mansinahar: On this one as well please.

endpoints/openrtb2/auction_test.go
```diff
@@ -227,16 +229,74 @@ func TestBlacklistRequests(t *testing.T) {
 	tests.assert(t)
 }
 
+// TestRejectAccountRequired asserts we return a 400 code on a request that comes with no user id nor app id
+// if the `AccountRequired` field in the `config.Configuration` structure is set to true
+func TestRejectAccountRequired(t *testing.T) {
+	tests := map[string]*getResponseFromDirectory{
+		"8.1) Not required and not provided. Since not provided, not blacklisted": {
```
mansinahar: What does these numbers `8.1`, `8.2`, etc mean? Also, why are these test cases part of a map? Doesn't look like we're using the key anywhere. Would recommend just making it a slice and you can comment out the description of each test case so its easier to understand for anyone else. Ideally with table tests we should have a `description` field that is printed in case of failed assertions.

endpoints/openrtb2/auction_test.go
```diff
@@ -227,16 +229,74 @@ func TestBlacklistRequests(t *testing.T) {
...
 	t.Helper()
-	for _, fileInfo := range fetchFiles(t, gr.dir) {
-		filename := gr.dir + "/" + fileInfo.Name()
-		fileData := readFile(t, filename)
+	var filename string
+	var fileData []byte
+	if gr.file == "" {
+		for _, fileInfo := range fetchFiles(t, gr.dir) {
+			filename = gr.dir + "/" + fileInfo.Name()
+			fileData = readFile(t, filename)
```
mansinahar: I'd recommend extracting the code from L286-L293 into a different function rather than having redundant code or another approach would be to create a slice of all filenames to be processed and then have the for loop range over that.

endpoints/openrtb2/auction_test.go
```diff
@@ -227,16 +229,74 @@ func TestBlacklistRequests(t *testing.T) {
...
+			expectedCode:  http.StatusOK,
+			aliased:       true,
+			accountReq:    true,
+		},
+		"8.4) Required, provided and blacklisted": {
+			dir:           "sample-requests/blacklisted",
+			file:          "blacklisted-acct.json",
+			payloadGetter: getRequestPayload,
+			messageGetter: getMessage,
+			expectedCode:  http.StatusBadRequest,
```
mansinahar: Shouldn't the expected response code here be `http. StatusServiceUnavailable`?

endpoints/openrtb2/sample-requests/account-required/with-acct.json
guscarreon: Done

endpoints/openrtb2/sample-requests/account-required/no-acct.json
guscarreon: Done

endpoints/openrtb2/auction_test.go
guscarreon: Done. Implemented the array that we discussed, rather than the function.

endpoints/openrtb2/auction_test.go
guscarreon: Yes, there was a bug in the code. Corrected

endpoints/openrtb2/auction_test.go
guscarreon: Mansi I converted the map into a slice and just left the description commented out since the errors themselves come with an error message. Let me knwo what you think

endpoints/openrtb2/auction.go
```diff
@@ -1141,9 +1139,18 @@ func checkSafari(r *http.Request) (isSafari bool) {
...
+		httpStatus := http.StatusBadRequest
+		metricsStatus := pbsmetrics.RequestStatusBadInput
+		for _, err := range errs {
+			erVal := errortypes.DecodeError(err)
+			if erVal == errortypes.BlacklistedAppCode || erVal == errortypes.BlacklistedAcctCode {
+				httpStatus = http.StatusServiceUnavailable
+			}
+		}
+		w.WriteHeader(httpStatus)
+		labels.RequestStatus = metricsStatus
```
hhhjort: Can anything change `metricsStatus` after line 1144? If not, why not combine the two lines and get rid of `metricsStatus`?

guscarreon: I left it like that because in PR#1030 the blacklist metrics get implemented so the for loop changes to:
```
func writeError(errs []error, w http.ResponseWriter, labels *pbsmetrics.Labels) bool {
	if len(errs) > 0 {
		httpStatus := http.StatusBadRequest
		metricsStatus := pbsmetrics.RequestStatusBadInput
		for _, err := range errs {
			erVal := errortypes.DecodeError(err)
			if erVal == errortypes.BlacklistedAppCode || erVal == errortypes.BlacklistedAcctCode {
				httpStatus = http.StatusServiceUnavailable
   +	        metricsStatus := pbsmetrics.RequestBlacklisted
			}
		}
		w.WriteHeader(httpStatus)
		labels.RequestStatus = metricsStatus
		for _, err := range errs {
			w.Write([]byte(fmt.Sprintf("Invalid request: %s\n", err.Error())))
		}
		return true
	}
	return false
}
```
But I can change it real quick if you want

guscarreon: Changed

guscarreon:
```
func writeError(errs []error, w http.ResponseWriter, labels *pbsmetrics.Labels) bool {
	if len(errs) > 0 {
		httpStatus := http.StatusBadRequest
		for _, err := range errs {
			erVal := errortypes.DecodeError(err)
			if erVal == errortypes.BlacklistedAppCode || erVal == errortypes.BlacklistedAcctCode {
				httpStatus = http.StatusServiceUnavailable
			}
		}
		w.WriteHeader(httpStatus)
		labels.RequestStatus = pbsmetrics.RequestStatusBadInput
		for _, err := range errs {
			w.Write([]byte(fmt.Sprintf("Invalid request: %s\n", err.Error())))
		}
		return true
	}
	return false
}
```

endpoints/openrtb2/amp_auction.go
```diff
@@ -150,14 +150,19 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
-		w.WriteHeader(http.StatusBadRequest)
+	if acctIdErr := validateAccount(deps.cfg, labels.PubID); acctIdErr != nil {
+		errL = append(errL, acctIdErr)
+		erVal := errortypes.DecodeError(acctIdErr)
+		if erVal == errortypes.BlacklistedAppCode || erVal == errortypes.BlacklistedAcctCode {
+			w.WriteHeader(http.StatusServiceUnavailable)
+		} else { //erVal == errortypes.AcctRequiredCode
+			w.WriteHeader(http.StatusBadRequest)
+		}
+		labels.RequestStatus = pbsmetrics.RequestStatusBadInput
```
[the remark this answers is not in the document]
guscarreon: Mansi these metrics are implemented in PR #1030

PR #1021, opened 2019-09-05
fix for using bid.ext.bidType field to set bid type in PubMatic adapter response

PubMatic-OpenWrap:
This PR has the fix for issue #862 .
The changes here include using the bid.ext.bidType field to set the bid type in PubMatic adapter response. Default value is "banner".

### DIFF
--- a/adapters/pubmatic/pubmatic.go
+++ b/adapters/pubmatic/pubmatic.go
@@ -21,7 +21,7 @@ import (
 )
 
 const MAX_IMPRESSIONS_PUBMATIC = 30
-const bidTypeExtKey = "bidType"
+const bidTypeExtKey = "BidType"
 
 type PubmaticAdapter struct {
 	http *adapters.HTTPAdapter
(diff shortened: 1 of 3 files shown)


hhhjort:
Travis is failing:
--- FAIL: TestJsonSamples (0.00s)
    test_json.go:269: pubmatictest/exemplary/video.json:  typedBid[0].type "banner" does not match expected "video."
FAIL
coverage: 87.3% of statements
FAIL	github.com/prebid/prebid-server/adapters/pubmatic	0.040s

hhhjort:
My guess is that the `video.json` test isn't specifying a bid with the proper `ext` for the bid type to be set as `video`.

adapters/pubmatic/pubmatic.go
```diff
@@ -569,8 +569,18 @@ func getBidType(bidExt json.RawMessage) openrtb_ext.BidType {
 		if err == nil {
 			err = json.Unmarshal(extbyte, &bidExtMap)
 			if err == nil && bidExtMap[bidTypeExtKey] != nil {
-				bidTypeStr := bidExtMap[bidTypeExtKey].(string)
-				bidType = openrtb_ext.BidType(bidTypeStr)
+				bidTypeVal := int(bidExtMap[bidTypeExtKey].(float64))
```
hhhjort: Converting to a float and then an int mean that you are safe on the conversions if someone entered a float, as opposed to casting it directly to an int?

adapters/pubmatic/pubmatic.go
```diff
@@ -550,30 +551,40 @@ func (a *PubmaticAdapter) MakeBids(internalRequest *openrtb.BidRequest, external
...
+			err = json.Unmarshal(extbyte, &bidExtMap)
+			if err == nil && bidExtMap[bidTypeExtKey] != nil {
+				bidTypeVal := int(bidExtMap[bidTypeExtKey].(float64))
+				switch bidTypeVal {
+				case 0:
+					bidType = openrtb_ext.BidTypeBanner
+				case 1:
+					bidType = openrtb_ext.BidTypeVideo
+				case 2:
+					bidType = openrtb_ext.BidTypeNative
```
guscarreon: The `parseImpressionObject(imp *openrtb.Imp, wrapExt *string, pubID *string)` function will filter out imps from `request.Imp[]` that are not `Video` nor `Banner`. Therefore, are we sure the `case:2` branch is reachable? Is `BidTypeNative` possible if we make it all the way here?.

PubMatic-OpenWrap:
Thanks @guscarreon. I've added test cases for the new function.

PR #1025, opened 2019-09-06
Optional include brand category

VeronikaSolovei9:


endpoints/openrtb2/video_auction.go
```diff
@@ -510,10 +510,14 @@ func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.Bi
 
 func createBidExtension(videoRequest *openrtb_ext.BidRequestVideo) ([]byte, error) {
 
-	inclBrandCat := openrtb_ext.ExtIncludeBrandCategory{
-		PrimaryAdServer: videoRequest.IncludeBrandCategory.PrimaryAdserver,
-		Publisher:       videoRequest.IncludeBrandCategory.Publisher,
+	var inclBrandCat *openrtb_ext.ExtIncludeBrandCategory
```
hhhjort: I assume so, but good to verify that it is safe that `inclBrandCat` can be `nil`.

Kalypsonika: @hhhjort this works with Impbus, we have ``` "includebrandcategory":null ``` in extention now. 
Before we had ``` "includebrandcategory": {"primaryadserver": 0, "publisher": ""}```. 
Please let me know if we need to confirm it and who to ask.

exchange/targeting_test.go
```diff
@@ -25,18 +25,18 @@ var mockBids = map[openrtb_ext.BidderName][]*openrtb.Bid{
...
-		Price: 0.7,
+		Price: 0.9,
 		CrID:  "2",
 	}},
 	openrtb_ext.BidderRubicon: {{
 		ID:    "contending-bid",
 		ImpID: "some-imp",
-		Price: 0.6,
+		Price: 0.5,
 		CrID:  "3",
```
guscarreon: Why did this values have to change? I noticed that if we roll the price values back to 0.5, 0.7, and 0.6, `TestTargetingCache` found in `exchange/targeting_test.go` fails.

Kalypsonika: @guscarreon  we now can build `hb_pb_cat_dur` values without category and as a result we have less bids back because of deduplication logic. In case of this unit test we have 3 bids: appnexus -> "losing-bid": price 0.5, "winning-bid" 2: price: 0.7, rubicon -> "contending-bid": price 0.6. 
In price_granularity.go -> func getCpmValue(), line 40: 
`roundedCPM := math.Floor(cpm/increment) * increment`
makes price rounding and as result prices turn into these values:
0.6 -> 0.5
0.5 -> 0.5
0.7 -> 0.6
So we have 2 similar bids and after deduplication one of them should be removed. 
I changed values in unit test to make sure all 3 bids are unique and non of them will be removed by deduplication logic.

jmaynardxandr: We were discussing this and it does bring to light an issue that having the includeBrandCategories option omitted is changing the default behavior due to the de-duplication logic @Kalypsonika mentioned still being run. We will work on some changes that will allow the test to pass with the original price values, thanks for pointing this out @guscarreon .

guscarreon: No problem. Thank you both for addressing my comments so promptly.

Kalypsonika: @guscarreon @jmaynardxandr thank you for your help guys, it's now fixed

exchange/exchange.go
```diff
@@ -151,9 +151,15 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
 
 	if anyBidsReturned {
-		bidCategory, adapterBids, err := applyCategoryMapping(ctx, requestExt, adapterBids, *categoriesFetcher, targData)
-		if err != nil {
-			return nil, fmt.Errorf("Error in category mapping : %s", err.Error())
+
+		var bidCategory map[string]string
+		//If includebrandcategory is present in ext then CE feature is on.
+		if requestExt.Prebid.Targeting != nil && requestExt.Prebid.Targeting.IncludeBrandCategory != nil {
+			var err error
```
guscarreon: Veronica apparently there's no test that enters your newly written if statement. The following screenshot was taken from `coverage.html` a file generated using golang's default coverage tool.
```
╰─➤  go tool cover -html=c.out -o coverage.html
```

guscarreon: image

VeronikaSolovei9: @guscarreon unit test added. Can you please check if coverage changed?

PR #1030, opened 2019-09-09
Metric for blacklisted accounts

guscarreon:
Created a metric for rejected requests due to blacklisted app or account as discussed with @hhhjort  and @mansinahar.


hhhjort:
I think this PR might record the new result in Prometheus, however I don't see anything to add it to InfluxDB stats.

pbsmetrics/config/metrics_test.go
```diff
@@ -123,8 +147,8 @@ func TestMultiMetricsEngine(t *testing.T) {
 	VerifyMetrics(t, "AdapterMetrics.AppNexus.NoBidMeter", goEngine.AdapterMetrics[openrtb_ext.BidderAppnexus].NoBidMeter.Count(), 5)
 }
 
-func VerifyMetrics(t *testing.T, name string, expected int64, actual int64) {
+func VerifyMetrics(t *testing.T, name string, actual int64, expected int64) {
 	if expected != actual {
-		t.Errorf("Error in metric %s: expected %d, got %d.", name, expected, actual)
+		t.Errorf("Error in metric %s: got %d, expected %d.", name, expected, actual)
```
mansinahar: i believe you want this to be: `t.Errorf("Error in metric %s: got %d, expected %d.", name, actual, expected)`

pbsmetrics/config/metrics_test.go
```diff
@@ -86,6 +86,27 @@ func TestMultiMetricsEngine(t *testing.T) {
 		metricsEngine.RecordAdapterBidReceived(pubLabels, openrtb_ext.BidTypeBanner, true)
 		metricsEngine.RecordAdapterTime(pubLabels, time.Millisecond*20)
 	}
+	labels_blacklist := []pbsmetrics.Labels{
```
mansinahar: Can you please change the `labels_blacklist` variable to be in camel case rather than snake case as per Go standards?

pbsmetrics/config/metrics_test.go
```diff
@@ -86,6 +86,27 @@ func TestMultiMetricsEngine(t *testing.T) {
...
+		{
+			Source:        pbsmetrics.DemandWeb,
+			RType:         pbsmetrics.ReqTypeVideo,
+			PubID:         "test2",
+			Browser:       pbsmetrics.BrowserOther,
+			CookieFlag:    pbsmetrics.CookieFlagYes,
+			RequestStatus: pbsmetrics.RequestStatusBlacklisted,
+		},
+	}
+	for i := 0; i < 2; i++ {
```
mansinahar: Rather than hard coding `2` as the end of the for loop, we should rather just do:
```
for _, label := range blacklistLabels {
    ...
}
```

pbsmetrics/config/metrics_test.go
guscarreon: Good catch! Corrected

pbsmetrics/config/metrics_test.go
guscarreon: LOL. I'm the worse when it comes to naming standards. Corrected

pbsmetrics/config/metrics_test.go
guscarreon: Done

PR #1033, opened 2019-09-17
Add Adpone adapter

seergiioo6:


exchange/adapter_map.go
```diff
@@ -2,6 +2,7 @@ package exchange
 
 import (
 	"fmt"
+	"github.com/prebid/prebid-server/adapters/adpone"
```
guscarreon: Can we put this include in the next paragraph where all the adapter imports are?
```
  3 import (
  4     "fmt"
  5  -  "github.com/prebid/prebid-server/adapters/adpone"
  6     "net/http"
  7     "strings"
  8
  9     "github.com/prebid/prebid-server/adapters"
 10     ttx "github.com/prebid/prebid-server/adapters/33across"
 11     "github.com/prebid/prebid-server/adapters/adform"
 12     "github.com/prebid/prebid-server/adapters/adkernel"
 13     "github.com/prebid/prebid-server/adapters/adkernelAdn"
 14  +  "github.com/prebid/prebid-server/adapters/adpone"
 15     "github.com/prebid/prebid-server/adapters/adtelligent"
 16     "github.com/prebid/prebid-server/adapters/advangelists"
exchange/adapter_map.go
```

adapters/adpone/adpone.go
```diff
@@ -0,0 +1,93 @@
...
+) {
+	openRTBRequestJSON, err := json.Marshal(openRTBRequest)
+	if err != nil {
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	requestToBidder := &adapters.RequestData{
```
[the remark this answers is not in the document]
seergiioo6: Absolutely yes; I will add more headers. Thanks a lot

adapters/adpone/adpone.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (adapter *adponeAdapter) MakeRequests(
+	openRTBRequest *openrtb.BidRequest,
+	reqInfo *adapters.ExtraRequestInfo,
+) (
+	requestsToBidder []*adapters.RequestData,
+	errs []error,
+) {
+	openRTBRequestJSON, err := json.Marshal(openRTBRequest)
```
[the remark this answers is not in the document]
seergiioo6: You're right... 

We need at least 1 Impression to return a valid response.

Thanks again.

static/bidder-info/adpone.yaml
```diff
@@ -0,0 +1,9 @@
+maintainer:
+  email: "tech@adpone.com"
+capabilities:
+  app:
+    mediaTypes:
+      - banner
+  site:
+    mediaTypes:
+      - banner
```
guscarreon: Can you add a newline in here?

static/bidder-params/adpone.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Adpone Adapter Params",
+  "description": "A schema which validates params accepted by the adpone adapter",
+  "type": "object",
+  "properties": {},
```
guscarreon: Are we sure we don't have any parameters we want to check for? If not, do you think you'll need them in the future?

seergiioo6: Probably in the future we will need some parameters so I have added one (placementId). Thanks!

adapters/adpone/adpone.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (adapter *adponeAdapter) MakeRequests(
+	openRTBRequest *openrtb.BidRequest,
+	reqInfo *adapters.ExtraRequestInfo,
+) (
+	requestsToBidder []*adapters.RequestData,
+	errs []error,
+) {
+	var imp = &openRTBRequest.Imp[0]
```
guscarreon: When a request with an empty `openRTBRequest.Imp` array comes in, your adapter fails with an "index out of range" error in line 29. Can we check for `nil` before dereferencing the `Imp` array and check its length before accessing any of its elements?

PR #1044, opened 2019-09-24
publisher.ext.parentAccount to publisher.ext.prebid.parentAccount

guscarreon:
Change parentAccount from `publisher.ext.parentAccount` to `publisher.ext.prebid.parentAccount`. This PR created a new `Prebid` field in the `ExtPublisher` struct found in `openrtb_ext/publisher.go` so JSON from newly created `prebid` object can be unmarshaled.


openrtb_ext/publisher_test.go
```diff
@@ -0,0 +1,60 @@
+package openrtb_ext
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/stretchr/testify/assert"
+)
```
mansinahar: Do we really need these tests? We're basically testing json library's `Unmarshal` method here which isn't needed. We can trust the json library

guscarreon: Mansi, I can remove the `openrtb_ext/publisher_test.go` test case file if you want, but similar test case files that pretty much only test the `json.Unmarshal()` function also exist in the `openrtb_ext/` directory such as `openrtb_ext/request_test.go` and `openrtb_ext/site_test.go`. I noticed those two when I first coded this PR and figured I had to create similar tests to assert the proper unmarshaling of the `ExtPublisherPrebid` struct that this PR implements. Please let me know if you want me to get rid of `openrtb_ext/publisher_test.go`.

mansinahar: Yeah, I don't think we need to test the `json.Unmarshal` method as part of our code so if you don't mind removing it then that will be great. Thanks!

guscarreon: Done

PR #1045, opened 2019-09-24
Dfp category mapping

VeronikaSolovei9:


mansinahar:
@VeronikaSolovei9 LGTM. Just one QQ: Any reason to name the file `dfp_dfp.json` ?

VeronikaSolovei9:
@mansinahar we have this name format due to data structure format:
` "includebrandcategory": {
            "primaryadserver": 2,
            "publisher":"dfp"
          }
`
this mapping is default for dfp(primary ad server) where publisher is also dfp.

Issue #998, opened 2019-08-08
Move to Go modules?


benjaminch:
Hello there !

Now that prebid-server supports Go 1.11 and later, what do you think about introducing Go Modules support?
It has clear advantages including the Go setup.
I think it's not really a small dev but worth it IMO.
Here's the reference for it: https://blog.golang.org/using-go-modules

As far as I know, impacted parts of the projects will be:
- Documentation
- Docker setup
- CI setup
- Maybe tests script

What do you guys think about it?

I can definitely start working on it if you want :)

Cheers

mansinahar:
@benjaminch That'd definitely be good to have and if you're interested, sure go ahead and work on it. I will assign this issue to you. Thanks so much :)

benjaminch:
Hey @mansinahar, yes please assign it to me I will be happy to tackle this one ;)
Cheers !

SyntaxNode:
Would you consider this completed now since https://github.com/prebid/prebid-server/pull/1052 was merged?

benjaminch:
Yes ! Thanks a lot !

PR #1048, opened 2019-09-26
Added option to return nurl vs. AdM video.

muncha:
This is on top of the changes from #993, which has not been merged, and this pretty much wraps up the major changes that we are doing for now. In addition to giving the titular choice, this makes a simple guess at device type and security.


adapters/beachfront/beachfront.go
```diff
@@ -111,300 +96,388 @@ type BeachfrontResponseSlot struct {
...
+	return bfr, errs
 }
 
-func getVideoRequest(req *openrtb.BidRequest) (BeachfrontVideoRequest, []error, int) {
-	var videoImpsIndex = 0
-	var beachfrontReq = NewBeachfrontVideoRequest()
-	var errs = make([]error, 0, len(req.Imp))
-	var imps = 0
+func guessDeviceType(request *openrtb.BidRequest) openrtb.DeviceType {
+	if request.Site != nil {
```
hhhjort: Would it be better to base the guess on the device screen size? This really just breaks down to app or site.

muncha: That would be a good test, as well as UA (see 
a6d761d). I think a full implementation would try by UA, then screen size, and App/Site as a fallback, but now I'm being asked to just do this much and just consider types 1 and 2. I think it's a matter of beachfront wanting to audit how they are implementing deviceTypes on their back end. So semantically it would make sense for me to rename this function as it is not a "guess".

adapters/beachfront/beachfront.go
```diff
@@ -4,71 +4,56 @@ import (
...
-	"github.com/mxmCherry/openrtb"
 )
 
 const Seat = "beachfront"
 const BidCapacity = 5
 
-const BannerEndpoint = "https://display.bfmio.com/prebid_display"
-const VideoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id="
+const bannerEndpoint = "https://display.bfmio.com/prebid_display"
+const videoEndpoint = "https://reachms.bfmio.com/bid.json?exchange_id"
```
mansinahar: @muncha Looks like you guys have different endpoints that you'd wanna use based on the format. We're working on our end on add a more configurable way to be able to do this. That will let you remove these constants and rather let the Prebid Server host specify them as part of the configuration. We're planning to have that work done by either end of this week or early next week. Would you mind waiting till then and updating this PR after that has been merged in or do you need this to go out sooner than later?

muncha: Sure, hold off on this. I've got a branch with improvements to this any way. I'll just add the endpoint changes, merge that branch and update this PR. 

Can you point me to an issue or branch to watch for those changes?

muncha: Is this related to #998?

muncha:
@hhjort @mansinahar  - Can you go ahead and merge this in to master if it looks OK?  I've got a lot piling up here, and some people that want to do some live testing. I'll proceed with changing how the URLs are set (mansinahar's comment above) once that is wrapped up.

adapters/beachfront/beachfront.go
SyntaxNode: Not that one. It's related to https://github.com/prebid/prebid-server/pull/1070.

adapters/beachfront/beachfront.go
```diff
@@ -111,427 +96,588 @@ type BeachfrontResponseSlot struct {
...
+	for j := 0; j < len(beachfrontRequests.NurlVideo); j++ {
+		bytes, err := json.Marshal(beachfrontRequests.NurlVideo[j].Request)
 
-	if request.Device != nil {
-		addHeaderIfNonEmpty(headers, "User-Agent", request.Device.UA)
-		addHeaderIfNonEmpty(headers, "X-Forwarded-For", request.Device.IP)
-		addHeaderIfNonEmpty(headers, "Accept-Language", request.Device.Language)
-		if request.Device.DNT != nil {
-			addHeaderIfNonEmpty(headers, "DNT", strconv.Itoa(int(*request.Device.DNT)))
+		bytes = append([]byte(`{"isPrebid":true,`), bytes[1:]...)
```
SyntaxNode: You might want to check the err first, otherwise this may panic on a nil bytes response. Consider moving this into the `if err == nil` clause.

adapters/beachfront/beachfront.go
```diff
@@ -111,427 +96,588 @@ type BeachfrontResponseSlot struct {
...
 
-func getVideoRequest(req *openrtb.BidRequest) (BeachfrontVideoRequest, []error, int) {
-	var videoImpsIndex = 0
-	var beachfrontReq = NewBeachfrontVideoRequest()
-	var errs = make([]error, 0, len(req.Imp))
-	var imps = 0
+func getVideoRequests(request *openrtb.BidRequest) ([]beachfrontVideoRequest, []error) {
+	var bfReqs = make([]beachfrontVideoRequest, len(request.Imp))
+	var errs = make([]error, 0, len(request.Imp))
+	var bad = make([]int, 0)
```
SyntaxNode: Based on usage, consider renaming to something along the lines of `failedRequesIndices`.

SyntaxNode:
Sorry for the delay. I'm not familiar with the Beachfront adapter. Largely looks good to me. I left a few comments, the only one I'm concerned over would cause this adapter to panic if there is a marshalling error.

adapters/beachfront/beachfront.go
muncha: Cool. That is defiantly on the table for a later update.

muncha:
@hhjort @mansinahar  - Can you go ahead and merge this in to master if it looks OK?  I've got a lot piling up here, and some people that want to do some live testing. I'll proceed with changing how the URLs are set (mansinahar's comment above) once that is wrapped up.

PR #1050, opened 2019-09-27
verizonmedia: override tag ID and site ID + split up multi-imp requests

Aadeshp:


adapters/verizonmedia/verizonmedia.go
```diff
@@ -161,12 +174,9 @@ func getImpInfo(impId string, imps []openrtb.Imp) (bool, openrtb_ext.BidType) {
...
 func changeRequestForBidService(request *openrtb.BidRequest, extension *openrtb_ext.ExtImpVerizonMedia) {
-	if request.Imp[0].TagID == "" {
-		request.Imp[0].TagID = extension.Pos
-	}
-	if request.Site.ID == "" {
-		request.Site.ID = extension.Dcn
-	}
+	/* Always override the tag ID and site ID of the request */
+	request.Imp[0].TagID = extension.Pos
+	request.Site.ID = extension.Dcn
```
oath-jac: you are missing the banner size properties latest changes that we added.

adapters/verizonmedia/verizonmedia.go
```diff
@@ -34,64 +35,76 @@ func (a *VerizonMediaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo
...
+	headers.Add("Accept", "application/json")
+	headers.Set("User-Agent", request.Device.UA)
+	headers.Add("x-openrtb-version", "2.5")
+
+	for _, imp := range request.Imp {
+		var bidderExt adapters.ExtImpBidder
+		err := json.Unmarshal(imp.Ext, &bidderExt)
+		if err != nil {
+			err = &errortypes.BadInput{
+				Message: "ext.bidder not provided",
```
oath-jac: the Imp range index can be added in the error message

adapters/verizonmedia/verizonmedia.go
```diff
@@ -34,64 +35,76 @@ func (a *VerizonMediaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo
...
-		errors = append(errors, err)
-		return nil, errors
-	}
 
-	if verizonMediaExt.Dcn == "" {
-		err = &errortypes.BadInput{
-			Message: "Missing param dcn",
+		if verizonMediaExt.Dcn == "" {
+			err = &errortypes.BadInput{
+				Message: "Missing param dcn",
```
oath-jac: the Imp range index can be added in the error message

adapters/verizonmedia/verizonmedia.go
```diff
@@ -34,64 +35,76 @@ func (a *VerizonMediaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo
...
-		errors = append(errors, err)
-		return nil, errors
-	}
 
-	if verizonMediaExt.Pos == "" {
-		err = &errortypes.BadInput{
-			Message: "Missing param pos",
+		if verizonMediaExt.Pos == "" {
+			err = &errortypes.BadInput{
+				Message: "Missing param pos",
```
oath-jac: the same as above

Aadeshp:
@oath-jac addressed all the comments

adapters/verizonmedia/verizonmedia.go
```diff
@@ -35,68 +36,82 @@ func (a *VerizonMediaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo
...
+		reqCopy.Site = &siteCopy
 
-	headers := http.Header{}
-	headers.Add("Content-Type", "application/json;charset=utf-8")
-	headers.Add("Accept", "application/json")
-	headers.Set("User-Agent", request.Device.UA)
-	headers.Add("x-openrtb-version", "2.5")
+		if err := changeRequestForBidService(&reqCopy, &verizonMediaExt); err != nil {
+			errors = append(errors, err)
+			return nil, errors
```
guscarreon: @Aadeshp , are we sure we want to return because this particular Imp does not pass the checks found in `changeRequestForBidService()`? Don't we want to `continue` instead? The `changeRequestForBidService()` function pretty much makes sure this `Imp` is a banner and check for its dimensions. Are sure we want to exit at the first non `banner` or non-properly dimensioned `banner` we find?

Aadeshp: ah yes good find!

Aadeshp: fixed!

adapters/verizonmedia/verizonmedia.go
```diff
@@ -35,68 +36,82 @@ func (a *VerizonMediaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo
...
-		return nil, errors
+	reqs := make([]*adapters.RequestData, 0, len(request.Imp))
+	headers := http.Header{}
+
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	headers.Add("x-openrtb-version", "2.5")
+
+	if request.Device != nil {
+		headers.Set("User-Agent", request.Device.UA)
```
mansinahar: Probably wanna check the value of `Device.UA` before setting the header to make sure it's not empty

Aadeshp: fixed!

PR #1051, opened 2019-10-01
Return constant rate converter if not able to get rate conversions fr…

mansinahar:
…om fetch URL

### DIFF
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -78,7 +78,6 @@ func NewRateConverterWithNotifier(
 		glog.Errorf("Error fetching currency rates from %s. Currency conversions won't be supported. %s",
 			syncSourceURL, err.Error())
 		rc.constantRates = NewConstantRates()
-		return rc
 	}
 
 	go rc.startPeriodicFetching() // Start periodic ticking
@@ -139,7 +138,10 @@ func (rc *RateConverter) startPeriodicFetching() {
 		select {
 		case <-ticker.C:
 			// Retries are handled by clients directly.
-			rc.Update()
+			err := rc.Update()
+			if err == nil && rc.constantRates != nil {
+				rc.constantRates = nil
+			}
 			updatesTicksCount++
 			if rc.updateNotifier != nil {
 				rc.updateNotifier <- updatesTicksCount
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -70,7 +70,17 @@ func NewRateConverterWithNotifier(
 		return rc
 	}
 
-	rc.Update()                   // Make sure to populate data before to return the converter
+	// Make sure to populate data before to return the converter
+	err := rc.Update()
+	// If unable to fetch currency rates from "syncSourceURL" then log the error
+	// and move on without supporting currency conversions.
+	if err != nil {
+		glog.Errorf("Error fetching currency rates from %s. Currency conversions won't be supported. %s",
+			syncSourceURL, err.Error())
+		rc.constantRates = NewConstantRates()
+		return rc
+	}
+
 	go rc.startPeriodicFetching() // Start periodic ticking
 
 	return rc
@@ -172,6 +182,11 @@ func (rc *RateConverter) Rates() Conversions {
 
 // GetInfo returns setup information about the converter
 func (rc *RateConverter) GetInfo() ConverterInfo {
+	currencyRates := rc.Rates()
+	if currencyRates == nil {
+		currencyRates = NewConstantRates()
+	}
+
 	return converterInfo{
 		source:           rc.syncSourceURL,
 		fetchingInterval: rc.fetchingInterval,
--- a/currencies/rate_converter_test.go
+++ b/currencies/rate_converter_test.go
@@ -1,10 +1,10 @@
 package currencies_test
 
 import (
-	"io/ioutil"
+	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


mansinahar:
Thanks @benjaminch :)

currencies/rate_converter.go
```diff
@@ -70,7 +70,17 @@ func NewRateConverterWithNotifier(
...
-	rc.Update()                   // Make sure to populate data before to return the converter
+	// Make sure to populate data before to return the converter
+	err := rc.Update()
+	// If unable to fetch currency rates from "syncSourceURL" then log the error
+	// and move on without supporting currency conversions.
+	if err != nil {
+		glog.Errorf("Error fetching currency rates from %s. Currency conversions won't be supported. %s",
+			syncSourceURL, err.Error())
+		rc.constantRates = NewConstantRates()
+		return rc
```
SyntaxNode: If the intent is to allow PBS to startup successfully when the rate server is unavailable, I think we should continue to try and hit the rates server during periodic updates for when the server is once again reachable. In that case, the error message makes sense but I would like to see the `startPeriodicFetching` routine still running. Ideally IMHO, Rates would return constant rates when no rate refresh should ever be attempted or when all previous attempts have failed.

benjaminch: That's a good point ! +1 to let the routine try to fetch currency rates !

mansinahar: @syntaxnode Yep, I like that approach better. Will update the PR. Thanks!

currencies/rate_converter_test.go
```diff
@@ -13,354 +13,273 @@ import (
...
+		beforeExecution := time.Now()
+		currencyConverter := currencies.NewRateConverter(
+			&http.Client{},
+			test.fetchURL,
+			time.Duration(test.updateInterval)*time.Second,
+		)
+
+		// Verify:
+		if test.expectedTicks == 0 {
+			// Check for the next 200 milliseconds that no fetch was triggered
```
SyntaxNode: Why 200ms?

Be careful here, time based unit tests are notoriously fragile. I understand the code was hard to instrument for testing. Ideally we would factor out the scheduling mechanism for the unit tests. Let's see if we can find a more resilient trigger.. but if not, I know you wanted to refactor this code in the future to fix the testability issues.

currencies/rate_converter_test.go
```diff
@@ -13,354 +13,273 @@ import (
...
+			mockResponseCode: 200,
+			updateInterval:   1,
+			expectedTicks:    1,
+			description:      "Fetching currency rates successfully",
+		},
+		{
+			mockResponse:     []byte(""),
+			mockResponseCode: 200,
+			updateInterval:   100,
+			description:      "Currency rates endpoint returns empty response",
```
SyntaxNode: What's the expectation for the test? If it's 0 ticks, consider still including that for readability.. even though it's not necessary per the compiler.

currencies/rate_converter_test.go
```diff
@@ -13,354 +13,273 @@ import (
...
-		&http.Client{},
-		mockedHttpServer.URL,
-		time.Duration(100)*time.Millisecond,
-		ticks,
-	)
+	rates := &currencies.Rates{}
+	err := json.Unmarshal(getMockRates(), rates)
+	assert.NoError(t, err, "JSON unmarshalling of conversions failed")
+
+	mockServerHandler := func(mockResponse []byte, code int, initialFailCount int,
```
SyntaxNode: **Style**: The type for `code` and for `initialFailCount` does not need to be specified since they're the same as `failRespCode`.

currencies/rate_converter_test.go
```diff
@@ -13,354 +13,273 @@ import (
...
+		fetchURL         string
+		expectedTicks    int
+		description      string
+	}{
+		{
+			mockResponse:     getMockRates(),
+			mockResponseCode: 200,
+			updateInterval:   100,
+			expectedTicks:    2,
+			initialFailCount: 2,
```
SyntaxNode: How does the initialFailCount affect these tests? Consider adding a comment in the setup or to the description.

currencies/rate_converter_test.go
```diff
@@ -13,354 +13,273 @@ import (
...
+			failRespCode:     400,
+			failRespMessage:  []byte("Bad request"),
+			description:      "Fetching currency rates successfully",
```
SyntaxNode: Looks like this (and the test below) return a 400 but are described as `Fetching currency rates successfully`. Copy/paste error?

currencies/rate_converter_test.go
```diff
@@ -13,354 +13,273 @@ import (
...
+		currencyConverter := currencies.NewRateConverterWithNotifier(
+			&http.Client{},
+			test.fetchURL,
+			updateInterval,
+			ticks,
+		)
+
+		// Verify:
+
+		// If test.expectedTicks == 0 meaning we expect back constantRates
```
SyntaxNode: Why do we expect back constantRates when expectedTicks is 0?

currencies/rate_converter_test.go
```diff
@@ -13,354 +13,273 @@ import (
...
-	}
-	lastFetched := time.Now()
-
-	// Verify:
-	// Check for the next 1 second that no fetch was triggered
-	time.Sleep(1 * time.Second)
 
-	assert.False(t, currencyConverter.LastUpdated().After(lastFetched), "LastUpdated() shouldn't be after `lastFetched` since the periodic fetching is stopped")
-}
+		errorMargin := 0.1 // 10% error margin
```
SyntaxNode: Why 10%?

SyntaxNode:
Silly stalebot. `mansinahar` is just on vacation.

SyntaxNode:
We'll get back to this soon.

PR #1052, opened 2019-10-01
[go modules] migrating to go modules

benjaminch:
This CL migrates prebid-server to Go Modules getting rid of `dep`.

Issue: #998


.travis.yml
```diff
@@ -1,20 +1,13 @@
 language: go
 
 go:
-  - '1.11.1'
+  - '1.11.13'
```
SyntaxNode: Why did you change the Go 1.11.x version?

Makefile
```diff
@@ -2,24 +2,17 @@
...
 	@echo ""
 
-.PHONY: install deps test build image
-
-# install dep https://golang.github.io/dep/ (assumes go is already installed)
-install:
-	export DEP_RELEASE_TAG=v0.4.1
-	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-
 # deps will clean out the vendor directory and use dep for a fresh install
```
SyntaxNode: please update comment to reference go modules

Dockerfile
```diff
@@ -6,27 +6,20 @@ RUN cd /tmp && \
...
-ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH
-RUN apt-get update && \
-    apt-get install -y git go-dep && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+ENV PATH=$GOROOT/bin:$PATH
 ENV CGO_ENABLED 0
 COPY ./ ./
-RUN dep ensure && \
-    go build .
+RUN go build -mod=vendor .
```
SyntaxNode: Why do you recommend using the vendor directory versus the default $GOPATH\pkg\mod directory?

.travis.yml
benjaminch: I didn't commented it, but because of https://github.com/golang/go/issues/30446 
1.11.4 would probably have fixed it but well, latest version cannot hurt.

Dockerfile
benjaminch: No preferences, but IMO modules allowing to work outside of the GOPATH is one of the best features, so just wanted to take this opportunity :)

Makefile
benjaminch: Oups, Sure !

.travis.yml
SyntaxNode: Gotcha. I was taking this change to mean that the lowest version of Go now supported is 1.11.13.

What do you think of including '1.11.4' as the first build target and mentioning in the docs that's it the minimum supported version? Then adding '1.11.x' and '1.12.x' for the latest version of both of those?

Dockerfile
SyntaxNode: I am not opposed :), was just curious if you ran into an issue specific to building the Docker container. I like that it allows for a more targeted cache clean by the deps command.

.travis.yml
benjaminch: Ok, let's use > 1.11.4.
1.11 will be removed soon I guess since PBS supports only the last 2 Go versions (your PR is on its way for 1.13 #1058 :))

Dockerfile
benjaminch: So it's ok for you this way? Also, I might introduce in another PR later a GOPROXY for modules to speed up the build.

.travis.yml
SyntaxNode: Ah, that's a very good point.

Dockerfile
SyntaxNode: Yes. I don't see a problem with it. 

The proxy sounds interesting. I'm curious to see what that looks like.

benjaminch:
Thanks @SyntaxNode :)

README.md
```diff
@@ -18,20 +18,15 @@ For more information, see:
 
 ## Installation
 
-First install [Go 1.11](https://golang.org/doc/install) or later and [dep](https://golang.github.io/dep/docs/installation.html). Note that dep requires an explicit GOPATH to be set.
+First install [Go 1.11](https://golang.org/doc/install) latest version (1.11.4) or later.
+Note that prebid-server is using [Go modules](https://blog.golang.org/using-go-modules).
```
mansinahar: Should probably add a note here to set `GO111MODULE=on` if you're using Go version <1.13 and are inside `GOPATH`. We can remove that after we stop supporting Go versions <1.13

Makefile
```diff
@@ -2,24 +2,17 @@
...
 	@echo ""
-	@echo "  install: install dep (assumes go is installed)"
-	@echo "  deps: grab dependencies using dep"
+	@echo "  deps: grab dependencies using go modules"
 	@echo "  test: test prebid-server (via validate.sh)"
 	@echo "  build: build prebid-server"
 	@echo "  image: build docker image"
 	@echo ""
 
-.PHONY: install deps test build image
```
mansinahar: Any reason for removing the `PHONY` targets? Generally, it is recommended that any target in the `makefile` that doesn't result into an actual file with the same name should be listed as a `PHONY` target.

More info about `PHONY` targets: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

README.md
benjaminch: Yes ! I'll do :)

Makefile
benjaminch: Oh, you are right, I just wanted to remove the deps part :)

Dockerfile
```diff
@@ -6,27 +6,20 @@ RUN cd /tmp && \
...
 FROM ubuntu:18.04 AS release
 LABEL maintainer="hans.hjort@xandr.com" 
 WORKDIR /usr/local/bin/
-COPY --from=build /go/src/github.com/prebid/prebid-server/prebid-server .
+COPY --from=build /app/prebid-server/ .
 COPY static static/
 COPY stored_requests/data stored_requests/data
-RUN apt-get update && \
-    apt-get install -y ca-certificates mtr && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
mansinahar: Sorry, I didn't catch this earlier but any reason to remove the `ca-certificates` and `mtr` package installation? I am not a 100% sure if we even need it in the first place. Likely not and therefore this should be fine but I just wanna confirm. Also, we should still at least keep the  `apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*` clean up command.

mansinahar: LGTM otherwise. Thanks @benjaminch :)

benjaminch: No worries, let's keep it all then, no real reasons here :)

benjaminch:
Thanks a lot @mansinahar and @SyntaxNode :)

Dockerfile
benjaminch: Hey @SyntaxNode , here you go for the proxy: #1079 :)

PR #1055, opened 2019-10-03
Disable account-specific metrics with boolean value in config

guscarreon:
This PR improves upon and substitutes PR #1053 by implementing the ability to disable the collection of account-specific metrics in prebid-server by setting the boolean variable `account_adapter_details` to `true` in the configuration environment of prebid-server

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -59,6 +59,8 @@ type Configuration struct {
 	BlacklistedAcctMap map[string]bool
 	// Is publisher/account ID required to be submitted in the OpenRTB2 request
 	AccountRequired bool `mapstructure:"account_required"`
+	// True if we want to stop collecting account-specific metrics
+	AccountAdapterDetails bool `mapstructure:"account_adapter_details"`
 }
 
 const MIN_COOKIE_SIZE_BYTES = 500
--- a/config/config.go
+++ b/config/config.go
@@ -59,8 +59,6 @@ type Configuration struct {
 	BlacklistedAcctMap map[string]bool
 	// Is publisher/account ID required to be submitted in the OpenRTB2 request
 	AccountRequired bool `mapstructure:"account_required"`
-	// True if we want to stop collecting account-specific metrics
-	AccountAdapterDetails bool `mapstructure:"account_adapter_details"`
 }
 
 const MIN_COOKIE_SIZE_BYTES = 500
@@ -306,6 +304,12 @@ func validateAdapters(adapterMap map[string]Adapter, errs configErrors) configEr
 type Metrics struct {
 	Influxdb   InfluxMetrics     `mapstructure:"influxdb"`
 	Prometheus PrometheusMetrics `mapstructure:"prometheus"`
+	Disabled   *DisabledMetrics  `mapstructure:"disabled_metrics"`
+}
+
+type DisabledMetrics struct {
+	// True if we want to stop collecting account-specific metrics
+	AccountAdapterDetails bool `mapstructure:"account_adapter_details"`
 }
 
 func (cfg *Metrics) validate(errs configErrors) configErrors {
--- a/config/config.go
+++ b/config/config.go
@@ -304,7 +304,7 @@ func validateAdapters(adapterMap map[string]Adapter, errs configErrors) configEr
 type Metrics struct {
 	Influxdb   InfluxMetrics     `mapstructure:"influxdb"`
 	Prometheus PrometheusMetrics `mapstructure:"prometheus"`
-	Disabled   *DisabledMetrics  `mapstructure:"disabled_metrics"`
+	Disabled   DisabledMetrics   `mapstructure:"disabled_metrics"`
 }
 
 type DisabledMetrics struct {
--- a/config/config.go
+++ b/config/config.go
@@ -308,7 +308,7 @@ type Metrics struct {
 }
 
 type DisabledMetrics struct {
-	// True if we want to stop collecting account-specific metrics
+	// True if we want to stop collecting account-to-adapter metrics
 	AccountAdapterDetails bool `mapstructure:"account_adapter_details"`
 }
 
--- a/pbsmetrics/go_metrics.go
+++ b/pbsmetrics/go_metrics.go
@@ -89,7 +89,7 @@ const unknownBidder openrtb_ext.BidderName = "unknown"
 // rather than loading legacy metrics that never get filled.
 // This will also eventually let us configure metrics, such as setting a limited set of metrics
 // for a production instance, and then expanding again when we need more debugging.
-func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName) *Metrics {
+func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName, disableAccountMetrics bool) *Metrics {
 	blankMeter := &metrics.NilMeter{}
 	newMetrics := &Metrics{
 		MetricsRegistry:            registry,
@@ -121,9 +121,12 @@ func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderNa
 		ImpsTypeNative: blankMeter,
 
 		AdapterMetrics: make(map[openrtb_ext.BidderName]*AdapterMetrics, len(exchanges)),
-		accountMetrics: make(map[string]*accountMetrics),
+		accountMetrics: nil,
 
 		exchanges: exchanges,
+	}
+	if !disableAccountMetrics {
+		newMetrics.accountMetrics = make(map[string]*accountMetrics)
 	}
 	for _, a := range exchanges {
 		newMetrics.AdapterMetrics[a] = makeBlankAdapterMetrics()
@@ -144,8 +147,9 @@ func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderNa
 // metrics object to contain only the metrics we are interested in. This would allow for debug
 // mode metrics. The code would allways try to record the metrics, but effectively noop if we are
 // using a blank meter/timer.
-func NewMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName) *Metrics {
-	newMetrics := NewBlankMetrics(registry, exchanges)
+//Account_adapter_details
+func NewMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName, disableAccountMetrics bool) *Metrics {
+	newMetrics := NewBlankMetrics(registry, exchanges, disableAccountMetrics)
 	newMetrics.ConnectionCounter = metrics.GetOrRegisterCounter("active_connections", registry)
 	newMetrics.ConnectionAcceptErrorMeter = metrics.GetOrRegisterMeter("connection_accept_errors", registry)
 	newMetrics.ConnectionCloseErrorMeter = metrics.GetOrRegisterMeter("connection_close_errors", registry)
@@ -256,6 +260,14 @@ func makeDeliveryMetrics(registry metrics.Registry, prefix string, bidType openr
 // There is no getBlankAccountMetrics() as all metrics are generated dynamically.
 func (me *Metrics) getAccountMetrics(id string) *accountMetrics {
 	var am *accountMetrics
+	/* type accountMetrics struct {
+		requestMeter      metrics.Meter
+		bidsReceivedMeter metrics.Meter
+		priceHistogram    metrics.Histogram
+		// store account by adapter metrics. Type is map[PBSBidder.BidderCode]
+		adapterMetrics map[openrtb_ext.BidderName]*AdapterMetrics
+	}
+	*/
 	var ok bool
 
 	me.accountMetricsRWMutex.RLock()
@@ -313,8 +325,10 @@ func (me *Metrics) RecordRequest(labels Labels) {
 	}
 
 	// Handle the account metrics now.
-	am := me.getAccountMetrics(labels.PubID)
-	am.requestMeter.Mark(1)
+	if me.accountMetrics != nil {
+		am := me.getAccountMetrics(labels.PubID)
+		am.requestMeter.Mark(1)
+	}
 }
 
 func (me *Metrics) RecordImps(labels ImpLabels) {
@@ -375,19 +389,26 @@ func (me *Metrics) RecordAdapterPanic(labels AdapterLabels) {
 // RecordAdapterRequest implements a part of the MetricsEngine interface
 func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
 	am, ok := me.AdapterMetrics[labels.Adapter]
+	var aam *AdapterMetrics
 	if !ok {
 		glog.Errorf("Trying to run adapter metrics on %s: adapter metrics not found", string(labels.Adapter))
 		return
 	}
 
-	aam := me.getAccountMetrics(labels.PubID).adapterMetrics[labels.Adapter]
+	if me.accountMetrics != nil {
+		aam = me.getAccountMetrics(labels.PubID).adapterMetrics[labels.Adapter]
+	}
 	switch labels.AdapterBids {
 	case AdapterBidNone:
 		am.NoBidMeter.Mark(1)
-		aam.NoBidMeter.Mark(1)
+		if me.accountMetrics != nil {
+			aam.NoBidMeter.Mark(1)
+		}
 	case AdapterBidPresent:
 		am.GotBidsMeter.Mark(1)
-		aam.GotBidsMeter.Mark(1)
+		if me.accountMetrics != nil {
+			aam.GotBidsMeter.Mark(1)
+		}
 	default:
 		glog.Warningf("No go-metrics logged for AdapterBids value: %s", labels.AdapterBids)
 	}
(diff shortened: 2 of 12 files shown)


config/config.go
```diff
@@ -59,6 +59,8 @@ type Configuration struct {
 	BlacklistedAcctMap map[string]bool
 	// Is publisher/account ID required to be submitted in the OpenRTB2 request
 	AccountRequired bool `mapstructure:"account_required"`
+	// True if we want to stop collecting account-specific metrics
```
hhhjort: I think we want to put this in with the metrics config someplace. Probably in its own subobject, so we can add additional flags/metrics groups as needed.

pbsmetrics/go_metrics.go
```diff
@@ -89,7 +89,7 @@ const unknownBidder openrtb_ext.BidderName = "unknown"
 // rather than loading legacy metrics that never get filled.
 // This will also eventually let us configure metrics, such as setting a limited set of metrics
 // for a production instance, and then expanding again when we need more debugging.
-func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName) *Metrics {
+func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName, disableAccountMetrics bool) *Metrics {
```
hhhjort: I mentioned earlier that we should shove this boolean into its own struct/object. We should then pass this object in rather than just the boolean value, so if we expand the list, we won't have to change the function signature every time.

pbsmetrics/go_metrics.go
```diff
@@ -121,10 +121,13 @@ func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderNa
 		ImpsTypeNative: blankMeter,
 
 		AdapterMetrics: make(map[openrtb_ext.BidderName]*AdapterMetrics, len(exchanges)),
-		accountMetrics: make(map[string]*accountMetrics),
+		accountMetrics: nil,
 
 		exchanges: exchanges,
 	}
+	if !disableAccountMetrics {
```
hhhjort: We want to keep account metrics, just disable Account X Adapter metrics. So we need this map in all cases.

pbsmetrics/go_metrics.go
```diff
@@ -256,6 +260,14 @@ func makeDeliveryMetrics(registry metrics.Registry, prefix string, bidType openr
 // There is no getBlankAccountMetrics() as all metrics are generated dynamically.
 func (me *Metrics) getAccountMetrics(id string) *accountMetrics {
 	var am *accountMetrics
+	/* type accountMetrics struct {
```
hhhjort: Was this supposed to be a temporary note, or do you think we should document the the struct's definition in comments here?

pbsmetrics/go_metrics.go
```diff
@@ -313,8 +325,10 @@ func (me *Metrics) RecordRequest(labels Labels) {
 	}
 
 	// Handle the account metrics now.
-	am := me.getAccountMetrics(labels.PubID)
-	am.requestMeter.Mark(1)
+	if me.accountMetrics != nil {
+		am := me.getAccountMetrics(labels.PubID)
```
hhhjort: We want plain account metrics, just not account*adapter metrics.

pbsmetrics/go_metrics.go
```diff
@@ -375,19 +389,26 @@ func (me *Metrics) RecordAdapterPanic(labels AdapterLabels) {
...
 func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
 	am, ok := me.AdapterMetrics[labels.Adapter]
+	var aam *AdapterMetrics
 	if !ok {
 		glog.Errorf("Trying to run adapter metrics on %s: adapter metrics not found", string(labels.Adapter))
 		return
 	}
 
-	aam := me.getAccountMetrics(labels.PubID).adapterMetrics[labels.Adapter]
+	if me.accountMetrics != nil {
```
hhhjort: `me.accountMetrics` should always be non-nil. We should be creating new account metrics the first time we see an account.

pbsmetrics/go_metrics.go
```diff
@@ -256,6 +260,14 @@ func makeDeliveryMetrics(registry metrics.Registry, prefix string, bidType openr
...
 func (me *Metrics) getAccountMetrics(id string) *accountMetrics {
 	var am *accountMetrics
+	/* type accountMetrics struct {
+		requestMeter      metrics.Meter
+		bidsReceivedMeter metrics.Meter
+		priceHistogram    metrics.Histogram
+		// store account by adapter metrics. Type is map[PBSBidder.BidderCode]
+		adapterMetrics map[openrtb_ext.BidderName]*AdapterMetrics
+	}
+	*/
```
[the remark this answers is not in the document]
guscarreon: Yes it is.... sorry.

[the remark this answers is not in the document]
guscarreon: Gone

pbsmetrics/go_metrics.go
guscarreon: Forgot to get rid of this comment. Gone.

pbsmetrics/go_metrics.go
guscarreon: You are absolutely right.

pbsmetrics/go_metrics.go
guscarreon: Rollbacked

pbsmetrics/go_metrics.go
guscarreon: Got confused, I thought we wanted to disable all of the metrics this function works with. Which ones do you want us to keep?
```
255 // getAccountMetrics gets or registers the account metrics for account "id".
256 // There is no getBlankAccountMetrics() as all metrics are generated dynamically.
257 func (me *Metrics) getAccountMetrics(id string) *accountMetrics {
258     var am *accountMetrics
259     var ok bool
260
261     me.accountMetricsRWMutex.RLock()
262     am, ok = me.accountMetrics[id]
263     me.accountMetricsRWMutex.RUnlock()
264
265     if ok {
266         return am
267     }
268
269     me.accountMetricsRWMutex.Lock()
270     // Made sure to use defer as we have two exit points: we want to unlock the mutex as quickly as possible.
271     defer me.accountMetricsRWMutex.Unlock()
272
273     am, ok = me.accountMetrics[id]
274     if ok {
275         // Unlock and return as quickly as possible
276         return am
277     }
278     am = &accountMetrics{}
279     am.requestMeter = metrics.GetOrRegisterMeter(fmt.Sprintf("account.%s.requests", id), me.MetricsRegistry)
280     am.bidsReceivedMeter = metrics.GetOrRegisterMeter(fmt.Sprintf("account.%s.bids_received", id), me.MetricsRegistry)
281     am.priceHistogram = metrics.GetOrRegisterHistogram(fmt.Sprintf("account.%s.prices", id), me.MetricsRegistry, metrics.NewExpDecaySample(1028, 0.015))
282     am.adapterMetrics = make(map[openrtb_ext.BidderName]*AdapterMetrics, len(me.exchanges))
283     for _, a := range me.exchanges {
284         am.adapterMetrics[a] = makeBlankAdapterMetrics()
285         registerAdapterMetrics(me.MetricsRegistry, fmt.Sprintf("account.%s", id), string(a), am.adapterMetrics[a])
286     }
287
288     me.accountMetrics[id] = am
289
290     return am
291 }
pbsmetrics/go_metrics.go [+]
```

guscarreon: the ones in the for loop in line 283? Like this:
```
     +  if !AdapterMetricsDisabled {
283         for _, a := range me.exchanges {
284             am.adapterMetrics[a] = makeBlankAdapterMetrics()
285             registerAdapterMetrics(me.MetricsRegistry, fmt.Sprintf("account.%s", id), string(a), am.adapterMetrics[a])
286         }
     +  }
pbsmetrics/go_metrics.go [+]
```

hhhjort: Yes, we don't want to do anything with `adapterMetrics` inside the account metrics context. There are adapter metrics elsewhere that do not reference account.

pbsmetrics/go_metrics.go
```diff
@@ -412,8 +433,10 @@ func (me *Metrics) RecordAdapterBidReceived(labels AdapterLabels, bidType openrt
 	// Adapter metrics
 	am.BidsReceivedMeter.Mark(1)
 	// Account-Adapter metrics
-	aam := me.getAccountMetrics(labels.PubID).adapterMetrics[labels.Adapter]
-	aam.BidsReceivedMeter.Mark(1)
+	if me.accountMetrics != nil {
```
hhhjort: This is correct, anything that has both `labels.PubID` and `labels.Adapter` we want to get rid of. We want to keep it when it is just one or the other.

config/config.go
guscarreon: Done

pbsmetrics/go_metrics.go
guscarreon: Got it. Update corrects this issue

pbsmetrics/go_metrics.go
guscarreon: Corrected

config/config.go
```diff
@@ -59,6 +59,8 @@ type Configuration struct {
...
+	AccountAdapterDetails bool `mapstructure:"account_adapter_details"`
```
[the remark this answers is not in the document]
guscarreon: Hans chose the name himself. I guess we can decide on its name on our next meeting

pbsmetrics/go_metrics.go
guscarreon: With the check implemented inside of `getAccountMetrics(id string)` we are only not setting metrics named like: `account.<labels.PubID>.<ADAPTER_NAME>.<SOME_METIRC>` but in the newest version of the code, we still need a null check on whether or not  `labels.Adapter` is found in the `adapterMetrics` hash table.

config/config.go
```diff
@@ -304,6 +304,12 @@ func validateAdapters(adapterMap map[string]Adapter, errs configErrors) configEr
 type Metrics struct {
 	Influxdb   InfluxMetrics     `mapstructure:"influxdb"`
 	Prometheus PrometheusMetrics `mapstructure:"prometheus"`
+	Disabled   *DisabledMetrics  `mapstructure:"disabled_metrics"`
```
hhhjort: Why is this a pointer?

pbsmetrics/go_metrics.go
```diff
@@ -120,11 +123,15 @@ func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderNa
 		ImpsTypeAudio:  blankMeter,
 		ImpsTypeNative: blankMeter,
 
-		AdapterMetrics: make(map[openrtb_ext.BidderName]*AdapterMetrics, len(exchanges)),
-		accountMetrics: make(map[string]*accountMetrics),
+		AdapterMetrics:  make(map[openrtb_ext.BidderName]*AdapterMetrics, len(exchanges)),
+		accountMetrics:  make(map[string]*accountMetrics),
+		MetricsDisabled: config.DisabledMetrics{},
```
hhhjort: If the DisabledMetrics was not a pointer, we would not need any nil checking logic below.

pbsmetrics/go_metrics.go
```diff
@@ -280,9 +287,9 @@ func (me *Metrics) getAccountMetrics(id string) *accountMetrics {
 	am.bidsReceivedMeter = metrics.GetOrRegisterMeter(fmt.Sprintf("account.%s.bids_received", id), me.MetricsRegistry)
 	am.priceHistogram = metrics.GetOrRegisterHistogram(fmt.Sprintf("account.%s.prices", id), me.MetricsRegistry, metrics.NewExpDecaySample(1028, 0.015))
 	am.adapterMetrics = make(map[openrtb_ext.BidderName]*AdapterMetrics, len(me.exchanges))
-	for _, a := range me.exchanges {
-		am.adapterMetrics[a] = makeBlankAdapterMetrics()
-		registerAdapterMetrics(me.MetricsRegistry, fmt.Sprintf("account.%s", id), string(a), am.adapterMetrics[a])
+	for i := 0; i < len(me.exchanges) && me.MetricsDisabled.AccountAdapterDetails; i++ {
```
hhhjort: not sure if putting the `me.MetricsDisabled.AccountAdapterDetails` check in an if outside the for would be marginally better, as you wouldn't have to recheck the value each loop. Then again this is startup code, so won't impact running performance.

config/config.go
guscarreon: I thought that, if we are going to implelment more metric disabling boolean values inside `config.DisabledMetrics` than just `config.DisabledMetrics.AccountAdapterDetails`, we might as well initialize them all to false (meaning we are not disabling any metrics, all are enable by default), and if the `config.DisabledMetrics` object is not null, then just copy the entire thing with the pointer dereference notation as we do in line 133 of `pbsmetrics/go_metrics.go` shown below. But I realize there are other ways to do this so maybe I'm overthinking things or overcomplicating them using the pointer.
```
 95 func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName, disableMetrics *config.DisabledMetrics) *Metrics {
 96     blankMeter := &metrics.NilMeter{}
 97     newMetrics := &Metrics{
 98         MetricsRegistry:            registry,
 99         RequestStatuses:            make(map[RequestType]map[RequestStatus]metrics.Meter),
100 +-- 23 lines: ConnectionCounter:          metrics.NilCounter{},-------------------------------
123         ImpsTypeAudio:  blankMeter,
124         ImpsTypeNative: blankMeter,
125
126         AdapterMetrics:  make(map[openrtb_ext.BidderName]*AdapterMetrics, len(exchanges)),
127         accountMetrics:  make(map[string]*accountMetrics),
128         MetricsDisabled: config.DisabledMetrics{},
129
130         exchanges: exchanges,
131     }
132     if disableMetrics != nil && disableMetrics.AccountAdapterDetails {
133         newMetrics.MetricsDisabled = *disableMetrics //<-- copy the entire object in one line using the pointer notation
134     }
135     for _, a := range exchanges {
136         newMetrics.AdapterMetrics[a] = makeBlankAdapterMetrics()
137     }
138
139     for _, t := range RequestTypes() {
140         newMetrics.RequestStatuses[t] = make(map[RequestStatus]metrics.Meter)
141         for _, s := range RequestStatuses() {
142             newMetrics.RequestStatuses[t][s] = blankMeter
143         }
144     }
145
146     return newMetrics
147 }
pbsmetrics/go_metrics.go
```

guscarreon: If you want, I can get rid of the pointer and adjust accordingly

pbsmetrics/go_metrics.go
guscarreon: It does make sense. I can get rid of it if you want.

pbsmetrics/go_metrics.go
guscarreon: You are absolutely right. Wow I'm missing a test case then, let me double check them

pbsmetrics/go_metrics.go
hhhjort: I think it will be cleaner to not have a pointer. Copying a handfull of booleans on startup should not be a big performance issue.

config/config.go
```diff
@@ -304,6 +304,12 @@ func validateAdapters(adapterMap map[string]Adapter, errs configErrors) configEr
 type Metrics struct {
 	Influxdb   InfluxMetrics     `mapstructure:"influxdb"`
 	Prometheus PrometheusMetrics `mapstructure:"prometheus"`
+	Disabled   DisabledMetrics   `mapstructure:"disabled_metrics"`
+}
+
+type DisabledMetrics struct {
+	// True if we want to stop collecting account-specific metrics
```
SyntaxNode: This comment may be misleading. If I understand the PR correctly, we're still collecting account-specific metrics, but no longer account-adapter specific metrics - that is, metrics for each adapter per account.

guscarreon: That's true, let me change that

pbsmetrics/go_metrics_test.go
```diff
@@ -99,6 +102,29 @@ func ensureContainsAdapterMetrics(t *testing.T, registry metrics.Registry, name
...
+	m.RecordAdapterBidReceived(AdapterLabels{
+		Adapter: openrtb_ext.BidderAppnexus,
+		PubID:   accountId,
+	}, openrtb_ext.BidTypeBanner, true)
+
+	// Assert no adapter metrics are found in the accountMetrics map
+	assert.Equalf(t, len(m.accountMetrics[accountId].adapterMetrics) == 0, true, "Test failed. Account metrics that contain adapter information are disabled, therefore we expect no entries in m.accountMetrics[accountId].adapterMetrics, we have %d \n", len(m.accountMetrics[accountId].adapterMetrics))
+
+	// Assert no account metrics exist in the registry
+	registryMap := registry.GetAll()
```
mansinahar: We should break this up into two different tests because right now we're testing two things here: 1. When we call `NewMetrics` with a account-adapter metrics being disabled, the registry shouldn't have any account-adapter metrics (L119-125) and 2. When we call `RecordAdapterBidReceived`, there shouldn't be any account-adapter metrics recorded (because they shouldn't exist to start with) (L110 - L116)

Also, this file has a `TestRecordBidType` method that tests `RecordAdapterBidReceived`, so we should just add `#2` from above to that method.

guscarreon: Done

PR #1056, opened 2019-10-03
[Go 1.13] Testing Flags Refactor

SyntaxNode:
Go 1.13 includes the following change:

> Testing flags are now registered in the new Init function, which is invoked by the generated main function for the test. As a result, testing flags are now only registered when running a test binary, and packages that call flag.Parse during package initialization may cause tests to fail.

This is partially responsible for introducing an init flow bug when compiling against Go 1.13, causing tests to fail. It doesn't look like `flag.Parse()` needs to be performed during init and everything seems to work fine when its moved to the start of the main method.

### DIFF
--- a/pbs_light.go
+++ b/pbs_light.go
@@ -26,10 +26,11 @@ var Rev string
 
 func init() {
 	rand.Seed(time.Now().UnixNano())
-	flag.Parse() // read glog settings from cmd line
 }
 
 func main() {
+	flag.Parse() // required for glog flags and testing package flags
+
 	v := viper.New()
 	config.SetupViper(v, "pbs")
 	cfg, err := config.New(v)


pbs_light.go
```diff
@@ -26,10 +26,11 @@ var Rev string
 
 func init() {
 	rand.Seed(time.Now().UnixNano())
```
mansinahar: Wondering if we need the seed initialization for the random generator in `init` or can we move that to the `main` method too

SyntaxNode: I don't know what benefit we get from setting the seed in `init()`, but that does seem to common practice from what I could find. I can investigate further. I don't think this is related to Go 1.13.

PR #1058, opened 2019-10-04
[Go 1.13] Testing Expected Errors Refactor

SyntaxNode:
Go 1.13 includes a change to the `encoding/json.UnmarshalTypeError` where the `Field` now includes the full object path. This affects the deserialization error message and breaks two adapter tests which rely on error string comparison.

See:
https://github.com/golang/go/commit/29bc4f12581d836a96139c924f16a4987324edd1

To avoid requiring Go 1.13 as a minimum version due to this change, those tests should accept either expected error message. To accomplish this, I propose to enhance the expected error string comparison to support regex, which gives us enough flexibility to express both variants. This PR changes the error strings to a complex type for which a comparison mode is specified. This also gives us a foundation to build upon for potential other comparison modes.

I experimented with changing all string comparisons to regex, but it became complicated and non-obvious as regex special characters like . and [ would need to be escaped.

.. or, the affected adapters `eplanning` and `emx_digital` might be altered to produce uniform (but likely less detailed) error messages for json deserialization issues.

Thoughts?

### DIFF
--- a/adapters/adapterstest/test_json.go
+++ b/adapters/adapterstest/test_json.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"regexp"
 	"testing"
 
 	"github.com/mxmCherry/openrtb"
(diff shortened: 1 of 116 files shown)


adapters/adapterstest/test_json.go
```diff
@@ -198,15 +204,23 @@ func diffHttpRequestLists(t *testing.T, filename string, actual []*adapters.Requ
...
+func diffErrorLists(t *testing.T, description string, actual []error, expected []testSpecExpectedError) {
 	t.Helper()
 
 	if len(expected) != len(actual) {
 		t.Fatalf("%s had wrong error count. Expected %d, got %d", description, len(expected), len(actual))
 	}
 	for i := 0; i < len(actual); i++ {
-		if expected[i] != actual[i].Error() {
-			t.Errorf(`%s error[%d] had wrong message. Expected "%s", got "%s"`, description, i, expected[i], actual[i].Error())
+		if expected[i].Comparison == "literal" {
```
SyntaxNode: Is there a better way to do this in Go?

PR #1064, opened 2019-10-09
facebook adapter refactor

Aadeshp:


adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
-		err = e
-		return
+
+	reqs, errs := this.buildRequests(request)
+	return reqs, errs
+}
+
+func (this *FacebookAdapter) buildRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
+	if len(request.Imp) == 0 {
+		panic("invalid number of impressions")
```
hhhjort: I don't think the adapter should panic like this. It should return an error so that the error can be passed back in the response and the incorrect config/request can be fixed. A panic won't go back in the response, and can trigger our alerts that an adapter is panicing and needs to be fixed.

Aadeshp: Agreed, but this panic is there because this should never happen unless the code is wrong. I added a len(request.Imp) == 0 earlier in `MakeRequests` that returns an error, so the panic on L72 should never happen, its just there to safeguard from future iterations

adapters/audienceNetwork/facebook.go
```diff
@@ -1,34 +1,38 @@
...
-	"golang.org/x/net/context/ctxhttp"
+	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
 type FacebookAdapter struct {
 	http         *adapters.HTTPAdapter
 	URI          string
 	nonSecureUri string
-	platformJSON json.RawMessage
+	platformId   string
```
SyntaxNode: The Go style guide asks for `Id` to be `ID`. Same a few other structs in this PR.

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
+	AuthId     string `json:"authentication_id"`
 }
 
-func coinFlip() bool {
-	return rand.Intn(2) != 0
-}
+func (this *FacebookAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	if len(request.Imp) == 0 {
+		return nil, []error{&errortypes.BadInput{
+			Message: "no impressions provided",
```
SyntaxNode: Consider starting with a capital letter to match the style of the other `BadInput` error.

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
-	httpReq.Header.Add("Content-Type", "application/json")
-	httpReq.Header.Add("Accept", "application/json")
-
-	anResp, e := ctxhttp.Do(ctx, a.http.Client, httpReq)
-	if e != nil {
-		err = e
-		return
+
+	reqs, errs := this.buildRequests(request)
+	return reqs, errs
```
SyntaxNode: You can return the result of the `buildRequests` call directly. No need to assign local variables.

adapters/audienceNetwork/facebook.go
SyntaxNode: I see what you're saying. IMHO it would be easier to understand if you move the check which produces an error from `MakeRequests` right here, or omit the check here since it covered earlier by the only usage of this method. I don't think you need to `panic`. Future devs will understand and the panic is raising a few eyebrows.

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
 
-	if anResp.StatusCode != http.StatusOK {
-		err = &errortypes.BadServerResponse{
-			Message: fmt.Sprintf("HTTP status %d; body: %s", anResp.StatusCode, result.ResponseBody),
-		}
-		return
+	return reqs, errs
+}
+
+// The authentication ID is a sha256 hmac hash encoded as a he string, based on
```
SyntaxNode: Typo: he -> hex

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
+func (this *FacebookAdapter) makeAuthId(req *openrtb.BidRequest) string {
+	h := hmac.New(sha256.New, []byte(this.appSecret))
+	h.Write([]byte(req.ID))
+
+	return hex.EncodeToString(h.Sum(nil))
+}
+
+func (this *FacebookAdapter) modifyRequest(out *openrtb.BidRequest) error {
+	if len(out.Imp) != 1 {
+		panic("each bid request to facebook should only have a single impression")
```
SyntaxNode: Why panic here instead of outputting an error?

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
+			return nil
+		}
+
+		if _, ok := supportedBannerHeights[*out.Banner.H]; !ok {
+			return &errortypes.BadInput{
+				Message: fmt.Sprintf("imp #%s: only banner heights 50 and 250 are supported", out.ID),
+			}
+		}
+
+		/* This will get overwritten post-serialization */
```
SyntaxNode: Good to know. Thnx. 👍

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
+		out.Banner.W = openrtb.Uint64Ptr(0)
+		out.Banner.Format = nil
+		break
 
-	if fbReq.Imp != nil && len(fbReq.Imp) > 0 {
-		// only returns 1 imp for requested ad unit
-		fbReq.Imp = fbReq.Imp[unitInd : unitInd+1]
+	default:
```
SyntaxNode: I don't believe Go requires a default case. Since this does nothing, consider omitting it.

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
+	}
+
+	return nil
+}
 
-		if fbReq.Site != nil {
-			siteCopy := *fbReq.Site
-			siteCopy.Publisher = &openrtb.Publisher{ID: pubId}
-			fbReq.Site = &siteCopy
+func (this *FacebookAdapter) extractPlmtAndPub(out *openrtb.Imp) (string, string, error) {
```
SyntaxNode: it took me a while to figure out what plmt was referring to. Consider renaming to `extractPlacementAndPub` to make it easier to read.

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
+
+	placementId := fbExt.PlacementId
+	publisherId := fbExt.PublisherId
+
+	// Support the legacy path with the caller was expected to pass in just placementId
+	// which was an underscore concantenated string with the publisherId and placementId.
+	// The new path for callers is to pass in the placementId and publisherId independently
+	// and the below code will prefix the placementId that we pass to FAN with the publsiherId
+	// so that we can abstract the implementation details from the caller
+	toks := strings.Split(placementId, "_")
```
SyntaxNode: Does toks refer to tokens?

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,416 @@ func (a *FacebookAdapter) Name() string {
...
+				return typ
+			}
+
+			panic("processing an invalid impression")
 		}
 	}
 
-	if len(bids) == 0 {
-		return nil, err
+	panic(fmt.Sprintf("invalid bid imp ID %s", bid.ImpID))
```
SyntaxNode: Please add a bit more description to why the impression is invalid. A future dev debugging this thanks you in advance. :)

adapters/audienceNetwork/facebook.go
Aadeshp: Thats fair, i'll remove it

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/facebook.go
Aadeshp: yup

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/facebook.go
Aadeshp: done

adapters/audienceNetwork/audienceNetworktest/exemplary/native-1.1.json
```diff
@@ -0,0 +1,135 @@
...
+    },
+    "httpcalls": [
+        {
+            "expectedRequest": {
+                "uri": "https://an.facebook.com/placementbid.ortb",
+                "headers": {
+                    "Accept": [
+                        "application/json"
+                    ],
+                    "Auth-Token": [
```
annahc: We don't need this anymore with updated secure bidding spec

adapters/audienceNetwork/audienceNetworktest/exemplary/native-1.1.json
```diff
@@ -0,0 +1,135 @@
...
+                        }
+                    },
+                    "device": {
+                        "ip": "152.193.6.74"
+                    },
+                    "user": {
+                        "id": "db089de9-a62e-4861-a881-0ff15e052516",
+                        "buyeruid": "v4_bidder_token"
+                    },
+                    "tmax": 500,
```
annahc: Can we start with 800?

Aadeshp: this is just a test response, its based on the test request

SyntaxNode: That's part of the internal tests. The exact time has no affect on production code.

adapters/audienceNetwork/audienceNetworktest/exemplary/native-1.1.json
Aadeshp: So do we not need App ID anymore?

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,405 @@ func (a *FacebookAdapter) Name() string {
...
+func (this *FacebookAdapter) buildRequests(request *openrtb.BidRequest) ([]*adapters.RequestData, []error) {
+	// Documentation suggests bid request splitting by impression so that each
+	// request only represents a single impression
+	reqs := make([]*adapters.RequestData, 0, len(request.Imp))
+	headers := http.Header{}
+	var errs []error
+
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	headers.Add("Auth-Token", this.appID+"|"+this.appSecret)
```
annahc: We don't need this Auth-Token anymore. Sorry to miss this point.

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,405 @@ func (a *FacebookAdapter) Name() string {
...
-		fbReqV, err := a.MakeOpenRtbBidRequest(req, bidder, params.PlacementId, pbs.MEDIA_TYPE_VIDEO, pubId, i)
-		if err == nil {
-			requests[reqIndex] = &fbReqV
-			reqIndex = reqIndex + 1
+		break
 
+	case openrtb_ext.BidTypeNative:
+		// Set w/h to -1 for native impressions based on the facebook native spec.
+		// We have to set this post-serialization since the OpenRTB protocol doesn't
+		// actaully support w/h in the native object
```
annahc: Thanks for adding this

adapters/audienceNetwork/facebook.go
```diff
@@ -37,267 +41,405 @@ func (a *FacebookAdapter) Name() string {
...
-			Message: err.Error(),
-		}
-		return
+		return err
 	}
-	if len(bidResp.SeatBid) == 0 {
-		return
+
+	reqExt := facebookReqExt{
+		PlatformID: this.platformID,
```
annahc: I couldn't find where this platformID is defined, think this can be defined inside server since it will be identical across your bid requests.

PR #1070, opened 2019-10-10
Support configurable endpoints for adapters

guscarreon:
In order for adapters to be able to switch between endpoints based on the media type of the request, we are adding an extra field for adapters to pass additional info in JSON format. This allows for adapters to implement not only endpoint switching but also added functionality if needed.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -9,6 +9,7 @@ import (
 	"text/template"
 	"time"
 
+	"encoding/json"
 	"github.com/golang/glog"
 	"github.com/prebid/prebid-server/macros"
 	"github.com/prebid/prebid-server/openrtb_ext"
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -225,7 +226,8 @@ type Adapter struct {
 		Password string `mapstructure:"password"`
 		Tracker  string `mapstructure:"tracker"`
 	} `mapstructure:"xapi"` // needed for Rubicon
-	Disabled bool `mapstructure:"disabled"`
+	Disabled         bool            `mapstructure:"disabled"`
+	ExtraAdapterInfo json.RawMessage `json:"extra_info,omitempty"`
```
SyntaxNode: Can you add a test of this field to the existing `TestFullConfig` please?

hhhjort: You will also need to add this to `setBidderDefaults()` so that this new field can be set from an environment variable.

guscarreon: Done

config/config_test.go
```diff
@@ -7,9 +7,12 @@ import (
 	"testing"
 	"time"
 
+	"encoding/json"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
+	//"regexp"
```
SyntaxNode: Why are these imports commented out?

guscarreon: I forgot to remove them. Done

PR #1072, opened 2019-10-11
Add/enable account metrics for Prometheus

hhhjort:
This will allow Prometheus to record account level metrics. Account-Adapter metrics are still completely unsupported.

### DIFF
--- a/pbsmetrics/prometheus/prometheus.go
+++ b/pbsmetrics/prometheus/prometheus.go
@@ -49,6 +49,7 @@ const (
 	videoLabel          = "video"
 	audioLabel          = "audio"
 	nativeLabel         = "native"
+	accountLabel        = "account"
 )
 
 // NewMetrics constructs the appropriate options for the Prometheus metrics. Needs to be fed the promethus config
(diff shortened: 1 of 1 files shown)


pbsmetrics/prometheus/prometheus.go
```diff
@@ -294,9 +295,9 @@ func (me *Metrics) RecordUserIDSet(userLabels pbsmetrics.UserLabels) {
 
 func resolveLabels(labels pbsmetrics.Labels) prometheus.Labels {
 	return prometheus.Labels{
-		demandSourceLabel: string(labels.Source),
-		requestTypeLabel:  string(labels.RType),
-		// "pubid":   labels.PubID,
+		demandSourceLabel:   string(labels.Source),
+		requestTypeLabel:    string(labels.RType),
+		accountLabel:        string(labels.PubID),
```
guscarreon: There other 3 instances of the commented `// "pubid":   labels.PubID,` line in functions `resolveAdapterLabels()`, `resolveBidLabels()`, and `resolveAdapterErrorLabels()`. Shouild we also add `accountLabel` there? After all, `demandSourceLabel`, `requestTypeLabel`, `browserLabel`, `cookieLabel`, and `responseStatusLabel` do get repeated in all of them.

SyntaxNode:
Would always mapping the account/pubid label put us in a situation of logging account x adapter metrics? Should this code look at the `DisabledMetrics` config?

pbsmetrics/prometheus/prometheus.go
guscarreon: Discussed in a meeting. LGMT

PR #1073, opened 2019-10-15
[Go 1.13] Add Travis Build Target

SyntaxNode:


.travis.yml
```diff
@@ -3,6 +3,7 @@ language: go
 go:
   - '1.11.1'
```
benjaminch: Maybe remove Go 1.11 since as far as I know PBS wants to support last 2 Go major versions right?

SyntaxNode: Yes, I do believe PBS wants to support the latest two major versions. Just being extra cautious before removing 1.11 since there are a few older PRs in the pipeline which may unknowingly (although unlikely) introduce an issue specific to 1.13.

PR #1074, opened 2019-10-15
Rename pbs_light.go To main.go

SyntaxNode:
It is common for the main func to be contained within main.go. I understand the pbs_light name is an artifact of an earlier phase of the project. I suggest renaming the app entry point for clarity.


main.go
```diff
@@ -31,28 +31,37 @@ func init() {
...
 	}
-	if err := serve(Rev, cfg); err != nil {
+
+	err = serve(Rev, cfg)
+	if err != nil {
 		glog.Errorf("prebid-server failed: %v", err)
 	}
 }
 
+func loadConfig() (*config.Configuration, error) {
```
SyntaxNode: Moved the Viper setup logic into its own method to mirror the encapsulation of the server setup. IMHO, this seems easier to read.

main.go
```diff
@@ -31,28 +31,37 @@ func init() {
...
-	config.SetupViper(v, "pbs")
-	cfg, err := config.New(v)
+	cfg, err := loadConfig()
 	if err != nil {
 		glog.Fatalf("Configuration could not be loaded or did not pass validation: %v", err)
 	}
-	if err := serve(Rev, cfg); err != nil {
+
+	err = serve(Rev, cfg)
+	if err != nil {
```
SyntaxNode: The call to loadConfig cannot be inlined within the `if` block since the `cfg` is needed outside the block. Changing the usage of the server call to match for (hopefully you'll agree) improved readability.

main.go
```diff
@@ -31,28 +31,37 @@ func init() {
...
 }
 
+func loadConfig() (*config.Configuration, error) {
+	v := viper.New()
+	config.SetupViper(v, "pbs") // filke = filename
+	return config.New(v)
+}
+
 func serve(revision string, cfg *config.Configuration) error {
-	currencyConverter := currencies.NewRateConverter(&http.Client{}, cfg.CurrencyConverter.FetchURL, time.Duration(cfg.CurrencyConverter.FetchIntervalSeconds)*time.Second)
```
SyntaxNode: This line was quite long. Pulled out the time conversion.

PR #1078, opened 2019-10-17
Add Timing Metrics For Prebid Cache

SyntaxNode:


exchange/exchange.go
```diff
@@ -130,12 +130,8 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 				priceGranularity:  requestExt.Prebid.Targeting.PriceGranularity,
 				includeWinners:    requestExt.Prebid.Targeting.IncludeWinners,
 				includeBidderKeys: requestExt.Prebid.Targeting.IncludeBidderKeys,
-			}
-			if shouldCacheBids {
-				targData.includeCacheBids = true
-			}
-			if shouldCacheVAST {
-				targData.includeCacheVast = true
+				includeCacheBids:  shouldCacheBids,
```
SyntaxNode: Unrelated to PR title. A simple refactor opportunity I stumbled upon.

pbsmetrics/metrics.go
```diff
@@ -256,4 +256,5 @@ type MetricsEngine interface {
 	RecordUserIDSet(userLabels UserLabels) // Function should verify bidder values
 	RecordStoredReqCacheResult(cacheResult CacheResult, inc int)
 	RecordStoredImpCacheResult(cacheResult CacheResult, inc int)
+	RecordPrebidCacheRequestTime(length time.Duration)
```
SyntaxNode: The intent is to capture the duration of the HTTP call out to Prebid Cache for alerts and to compare against the Prebid Cache duration metrics. I don't personally think labels are warranted here, but please let me know if you feel differently.

pbsmetrics/metrics_mock.go
```diff
@@ -15,95 +15,84 @@ type MetricsEngineMock struct {
 // RecordRequest mock
 func (me *MetricsEngineMock) RecordRequest(labels Labels) {
 	me.Called(labels)
-	return
```
SyntaxNode: These return statements are not needed. The methods have no return parameters.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -58,6 +59,9 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
 	timerBuckets := prometheus.LinearBuckets(0.05, 0.05, 20)
 	timerBuckets = append(timerBuckets, []float64{1.5, 2.0, 3.0, 5.0, 10.0, 50.0}...)
 
+	timerBucketsQuickTasks := prometheus.LinearBuckets(0.005, 0.005, 20)
```
SyntaxNode: Prebid Cache times may be very fast. I defined a new time bucket for measuring tasks which are expected to be quick. The series looks like: 1ms, 1.5ms, 3ms, 5ms, 10ms, 15ms, 20ms, 25ms, 30ms, 35ms, 40ms, 45ms, 50ms, 55ms, 60ms, 65ms, 70ms, 75ms, 80ms, 85ms, 90ms, 95ms, 100ms.

pbsmetrics/metrics.go
hhhjort: No, I don't think labels will be needed for this.

hhhjort: Actually, we might want a success/failure flag, so we can see what the failure population looks like. For example, if we have significant population of cache requests that go 50ms+, but are ok because the bid requests all completed well under the TTL, yet have a bunch of quick cache requests that failed for some other reason, we could wrongly deduce that the failures are the fault of the 50ms+ population.

SyntaxNode: Pushed.

exchange/exchange.go
guscarreon: Isn't the previous code cleaner? Why do you think the `if` statements are preferable? I believe they add unnecessary identation and statements to the code and make it a bit less readable. Plus, they add a couplr of conditionals to the code and according to SonarCloud, it is preferable to have at most 15 of those.

PR #1079, opened 2019-10-18
[Go Modules] Add proxy

benjaminch:
This CL adds a proxy to ease Go modules usage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,8 @@ all:
 # deps will clean out the vendor directory and use go mod for a fresh install
 deps:
 	rm -rf vendor
-	go mod vendor
-	go mod tidy
-
+	GOPROXY="https://proxy.golang.org" go mod vendor -v && go mod tidy -v
+	
 # test will ensure that all of our dependencies are available and run validate.sh
 test: deps
 	./validate.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN mkdir -p /app/prebid-server/
 WORKDIR /app/prebid-server/
 ENV GOROOT=/usr/local/go
 ENV PATH=$GOROOT/bin:$PATH
+ENV GOPROXY="https://proxy.golang.org"
 RUN apt-get update && \
     apt-get install -y git && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*


SyntaxNode:
Interesting. Do you have an estimate for how much faster the build becomes?

SyntaxNode:
The desire is to use the cache to speed up builds, but it also looks like it may provide protection against un-tagged changes. Looking into this a bit, it seems the default proxy is blocked in China - so I just want to confirm this will not affect the user's environment such that `go build` would respect their own local proxy choice.

SyntaxNode:
One more thing to consider. The proxy added in this PR becomes default in Go 1.13. We would get largely the same affect changing the Go version in the Docker image. I'm curious if that affects you preferences to explicitly set it.

benjaminch:
> Interesting. Do you have an estimate for how much faster the build becomes?

I don't have any figures in mind nor tested but from what I can see from other proxies documentation, it's significantly faster: 

> Performance
Downloading stored dependencies from Athens is significantly faster than downloading dependencies from Version Control Systems. This is because go get by default uses VCS to download modules such as git clone, while go get with GOPROXY enabled will use HTTP to download zip archives. Therefore, depending on your computer and internet connection speed, it takes 10 seconds to download the CockroachDB source tree as a zip file from GitHub but almost four minutes to git clone it.
https://docs.gomods.io/intro/why/#performance

> The desire is to use the cache to speed up builds, but it also looks like it may provide protection against un-tagged changes. Looking into this a bit, it seems the default proxy is blocked in China - so I just want to confirm this will not affect the user's environment such that go build would respect their own local proxy choice.

I didn't tink about China, indeed, this should be checked / documented so users in China can change the proxy as they please.

> One more thing to consider. The proxy added in this PR becomes default in Go 1.13. We would get largely the same affect changing the Go version in the Docker image. I'm curious if that affects you preferences to explicitly set it.

I kind of like to sometimes explicitely set settings even if it's the default one. It allows people to be aware that this setting exists along with making it explicit.
Once dropping support to Go 1.12, then we can remove this setting but I think it's always a good thing to have it. From example, the China issue, it would be easy to track it down for a given user seeing that this URL is not reachable, then look for it in the project and then see the param and change it :)

Makefile
```diff
@@ -29,7 +29,7 @@ test: deps
 
 # build will ensure all of our tests pass and then build the go binary
 build: test
-	go build -mod=vendor ./...
+	GOPROXY="https://proxy.golang.org" go build -mod=vendor ./...
```
guscarreon: Do we have to redefine `GOPROXY` even when it was already set in the Dockerfile?

guscarreon: I tried the following and it also seemed to have worked:
```
╰─➤  $GOPROXY go build .
```
I'm not saying we should change it I'm just wondering why this specific syntax?

Makefile
```diff
@@ -13,7 +13,7 @@ all:
 # deps will clean out the vendor directory and use go mod for a fresh install
 deps:
 	rm -rf vendor
-	go mod vendor
+	go mod vendor -v
 	go mod tidy
```
guscarreon: Why do we need to print them to standard error? If the reason is to log possible package download issues, does it make sense to also add the `-v` flag to `go mod tidy` ? Again, just wondering

benjaminch: You are right, for consistency, better to add the `-v` on `tidy` if we keep it on `vendor`. i kind of like it so we know if anything goes bad. Let me know if you prefer not to have the verbose mode.

Makefile
benjaminch: Sure, let's use `$GOPROXY` instead :)
As to why redefining it in Makefile and in Dockerfileis just because not sure everyone use Docker directly to launch (I don't locally for example). I can remove it if you want though.

Makefile
```diff
@@ -29,8 +29,8 @@ test: deps
 
 # build will ensure all of our tests pass and then build the go binary
 build: test
-	go build -mod=vendor ./...
+	$GOPROXY go build -mod=vendor ./...
```
mansinahar: I am not sure what does it means to just add `$GOPROXY` at the start of the `go build` command without actually setting a value for it.

**Note:** If GOPROXY is unset, empty or set to direct then go get will use a direct connection to the VCS (e.g github.com) Ref: https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy/

Also, `go build  mod=vendor ./...` doesn't download any dependencies, it just looks for dependencies to build the binary, in the `vendor` folder. If the dependencies don't exist, it will fail. So, setting the `GOPROXY` var for this command doesn't really make sense.

I believe the right approach here is to set the `GOPROXY` env var to `"https://proxy.golang.org"` before running `go mod vendor` which is the command that will actually install all the dependencies.

benjaminch: Hey @mansinahar , indeed, you are right, I just pushed the changes ! Thanks :)

Makefile
```diff
@@ -13,8 +13,8 @@ all:
 # deps will clean out the vendor directory and use go mod for a fresh install
 deps:
 	rm -rf vendor
-	go mod vendor
-	go mod tidy
+	GOPROXY="https://proxy.golang.org" go mod vendor -v
+	go mod tidy -v
```
mansinahar: Thanks @benjaminch! Just one more thing though, what do you think about combining these two commands into one like:

```
GOPROXY="https://proxy.golang.org" go mod vendor -v && go mod tidy -v
```

The reason being, that `go mod tidy` can also potentially install any missing dependencies and so it will be good to supply it with the `GOPROXY` variable as well.

Ref (https://github.com/golang/go/wiki/Modules):

go mod tidy — Prune any no-longer-needed dependencies from go.mod and add any dependencies needed for other combinations of OS, architecture, and build tags

benjaminch: No worries, just updated :) Thanks !

mansinahar:
Thank you again for the PR @benjaminch :)

Issue #1041, opened 2019-09-23
I2I: Prebid server native preprocessor for Prebid mobile banner native


jaiminpanchal27:
## Problem

Prebid Universal creative needs to know the type of asset to render Prebid Mobile Banner Native. PUC fetches the content from Prebid cache to render the creative. In case of Prebid mobile native, the object returned looks like below. In this response, there is no way to identify asset with id 0 that it it main image or icon. As per native spec 1.2 https://www.iab.com/wp-content/uploads/2017/04/OpenRTB-Native-Ads-Specification-Draft_1.2_2017-04.pdf, Image Response Object has optional type attribute which helps in identifying whether it is main image or icon. Same goes for other assets as well. 
```
// Sample Native Bid Response
{
  "ver": "1.2",
  "assets": [
    {
      "id": 0,
      "img": {
        "url": "https://vcdn.adnxs.com/p/creative-image/7e/71/90/27/7e719027-80ef-4664-9b6d-a763da4cea4e.png",
        "w": 300,
        "h": 250,
        "ext": {
          "appnexus": {
            "prevent_crop": 0
          }
        }
      }
    },
    {
      "id": 1,
      "title": {
        "text": "This is an RTB ad"
      }
    },
    {
      "id": 2,
      "data": {
        "value": "AppNexus"
      }
    }
  ],
  "link": {
    "url": "https:/..."
  },
  "eventtrackers": [
    {
      "event": 1,
      "method": 1,
      "url": "https://ny..."
    }
  ]
}
```

### Option 1
Prebid server needs to preprocess the native response before storing in Prebid cache. Native Preprocessor will check whether type is returned or not, if not it will get the type from original bid request by comparing asset id's and then store the response in Prebid cache.

This preprocessing will only be done for Prebid mobile banner native use case. Prebid mobile native bid request will have a flag to enable preprocess

While rendering creative in Mobile device, Prebid universal creative will have a mapping of assets as per IAB native spec 1.2 to identify type and render the creative.

```
// IAB Native spec 1.2 asset mapping
const nativeAssets = {
  'image': {
    1: 'image',
    3: 'icon'
  },
  'body': {
    ...
  }
}

const nativeMapping = {
  "title": "hb_native_title",
  "body": "hb_native_body",
  "body2": "hb_native_body2",
  "privacyLink": "hb_native_privacy",
  "privacyIcon": "hb_native_privicon",
  "sponsoredBy": "hb_native_brand",
  "image": "hb_native_image",
[...]

PR #1081, opened 2019-10-18
Adding type to native markup for Prebid mobile

jaiminpanchal27:
This PR implements https://github.com/prebid/prebid-server/issues/1041


exchange/bidder.go
```diff
@@ -178,6 +225,26 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+	}
+	return nil, errors.New("Could not find native imp")
+}
+
+func getAssetByID(id int64, assets []nativeRequests.Asset) nativeRequests.Asset {
+	var asset nativeRequests.Asset
+	for i := 0; i < len(assets); i++ {
+		if id == assets[i].ID {
+			asset = assets[i]
+			break
```
SyntaxNode: Consider returning asset here and returning nil at the end.

exchange/bidder.go
```diff
@@ -178,6 +225,26 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+		if impInRequest.ID == impID && impInRequest.Native != nil {
+			return impInRequest.Native, nil
+		}
+	}
+	return nil, errors.New("Could not find native imp")
+}
+
+func getAssetByID(id int64, assets []nativeRequests.Asset) nativeRequests.Asset {
+	var asset nativeRequests.Asset
+	for i := 0; i < len(assets); i++ {
```
SyntaxNode: Consider using the `range` syntax of `for`.

exchange/bidder.go
```diff
@@ -153,6 +156,50 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+							nativeImp, err := getNativeImpByImpID(bidResponse.Bids[i].Bid.ImpID, request)
+							if err != nil {
+								errs = append(errs, err)
+							}
+							var nativePayload nativeRequests.Request
+							if err := json.Unmarshal(json.RawMessage((*nativeImp).Request), &nativePayload); err != nil {
+								errs = append(errs, err)
+							}
+
+							for j := 0; j < len(nativeMarkup.Assets); j++ {
```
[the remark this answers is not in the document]
jaiminpanchal27: Done

exchange/bidder.go
jaiminpanchal27: Done

exchange/bidder.go
jaiminpanchal27: Done

exchange/bidder.go
```diff
@@ -153,6 +156,50 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
 					}
 				}
 
+				// Only do this for request from mobile app
+				if request.App != nil {
+					for i := 0; i < len(bidResponse.Bids); i++ {
+						if bidResponse.Bids[i].BidType == openrtb_ext.BidTypeNative {
+							var nativeMarkup nativeResponse.Response
+							if err := json.Unmarshal(json.RawMessage(bidResponse.Bids[i].Bid.AdM), &nativeMarkup); err != nil {
```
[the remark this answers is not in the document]
jaiminpanchal27: json.Unmarshal expects a slice of byte. Here AdM is of type string and hence I used json.RawMessage

exchange/bidder.go
```diff
@@ -153,6 +156,50 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+								// Some bidders are not following IAB spec when doing native. One such example is Facebook. We are not ignoring such bids
```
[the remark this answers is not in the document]
jaiminpanchal27: Updated comment. I hope it makes more sense now

exchange/bidder.go
```diff
@@ -153,6 +156,50 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
 					}
 				}
 
+				// Only do this for request from mobile app
+				if request.App != nil {
+					for i := 0; i < len(bidResponse.Bids); i++ {
```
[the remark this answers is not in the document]
jaiminpanchal27: Done

jaiminpanchal27:
@SyntaxNode Thanks for feedback. I pushed an update

exchange/bidder.go
```diff
@@ -178,6 +199,64 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
 	return seatBid, errs
 }
 
+func addNativeTypes(bid *openrtb.Bid, request *openrtb.BidRequest) (*nativeResponse.Response, []error) {
+	var errs []error
+	var nativeMarkup *nativeResponse.Response
+	if err := json.Unmarshal(json.RawMessage(bid.AdM), &nativeMarkup); err != nil {
```
SyntaxNode: Why is it necessary to cast the message to `json.RawMessage` here and below on line 215?

SyntaxNode: ... I didn't see your answer above. Doh!

exchange/bidder.go
```diff
@@ -153,6 +156,24 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+						if bidResponse.Bids[i].BidType == openrtb_ext.BidTypeNative {
+							nativeMarkup, moreErrs := addNativeTypes(bidResponse.Bids[i].Bid, request)
+							errs = append(errs, moreErrs...)
+
+							if nativeMarkup != nil {
+								markup, err := json.Marshal(*nativeMarkup)
+								if err != nil {
+									errs = append(errs, err)
+								}
+								bidResponse.Bids[i].Bid.AdM = string(markup)
```
hhhjort: If there were errors in the marshalling, do we still want to we still want to try overwriting the AdM with whatever `json.Marshal()` managed to spit out?

jaiminpanchal27:
@anwzhang Correct. If `request.App` flag is only set when the request is from mobile device, we are all set

exchange/bidder.go
SyntaxNode: Good point. `markup` very well may be `nil`, of which `string(markup)` would succeed and result in an empty string.

exchange/bidder.go
```diff
@@ -178,6 +199,64 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+		// Some bidders are returning non-IAB complaiant native markup. In this case Prebid server will not be able to add types. E.g Facebook
+		return nil, errs
+	}
+
+	nativeImp, err := getNativeImpByImpID(bid.ImpID, request)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	var nativePayload nativeRequests.Request
+	if err := json.Unmarshal(json.RawMessage((*nativeImp).Request), &nativePayload); err != nil {
```
hhhjort: We have a potential panic here. If `getNativeImpByImpID()` fails to find the imp, then `nativeImp` is nil. Trying to access `nil.Request` will then panic, as you cannot dereference a `nil`. You should probably abort above if `nativeImp` is `nil`. I don't think we need to abort here on `Unmarshal` errors as the worst that should happen is we fail asset lookups and don't do anything.

exchange/exchangetest/mobile-native-types.json
```diff
@@ -0,0 +1,116 @@
...
+      "mockResponse": {
+        "pbsSeatBid": {
+          "pbsBids": [
+            {
+              "ortbBid": {
+                "id": "apn-bid",
+                "impid": "my-imp-id",
+                "price": 10,
+                "adid": "1234",
+                "adm": "{\"assets\":[{\"id\":1,\"img\":{\"type\":3,\"url\":\"http://vcdn.adnxs.com/p/creative-image/f8/7f/0f/13/f87f0f13-230c-4f05-8087-db9216e393de.jpg\",\"w\":989,\"h\":742,\"ext\":{\"appnexus\":{\"prevent_crop\":0}}}},{\"title\":{\"text\":\"This is a Prebid Native Creative\"}},{\"id\":2,\"data\":{\"type\":1,\"value\":\"Prebid.org\"}},{\"id\":3,\"data\":{\"type\":2,\"value\":\"This is a Prebid Native Creative.  There are many like it, but this one is mine.\"}}],\"link\":{\"url\":\"http://some-url.com\"},\"imptrackers\":[\"http://someimptracker.com\"],\"jstracker\":\"some-js-tracker\"}",
```
hhhjort: In the mock response, don't you want the adapter to return native markup without types being set, so you can verify the code does actually set the types?

exchange/bidder.go
jaiminpanchal27: Agreed. Updated

exchange/bidder.go
jaiminpanchal27: Done

exchange/exchangetest/mobile-native-types.json
jaiminpanchal27: I misunderstood the exchange_test.go Now I moved my test to bidder_test.go

jaiminpanchal27:
@hhhjort Pushed an update. Addressed all your feedback

PR #1082, opened 2019-10-18
[Sharethrough] Adapter fixes

maphe:
Couple of fixes:
- get `bidfloor` from custom impression params
- remove `gzip` header

### DIFF
--- a/adapters/sharethrough/butler.go
+++ b/adapters/sharethrough/butler.go
@@ -71,7 +71,6 @@ func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openr
 	headers := http.Header{}
 	headers.Add("Content-Type", "application/json;charset=utf-8")
 	headers.Add("Accept", "application/json")
-	headers.Add("Accept-Encoding", "gzip")
 	headers.Add("Origin", domain)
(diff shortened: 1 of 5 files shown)


adapters/sharethrough/butler.go
```diff
@@ -71,7 +71,6 @@ func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb.Imp, request *openr
 	headers := http.Header{}
 	headers.Add("Content-Type", "application/json;charset=utf-8")
 	headers.Add("Accept", "application/json")
-	headers.Add("Accept-Encoding", "gzip")
```
SyntaxNode: Curious, why is gzip no longer accepted?

maphe: @SyntaxNode it actually broke our adapter, our ad server now returns gzip encoded responses and it looks like the adapter needs to handle unzipping itself. We'll push a fix later to decode gzipped responses but for now we just need this fix asap in production as it can't process any bid in its current state.

SyntaxNode: Gotcha.

maphe:
Hi @guscarreon, just a heads up we would really appreciate it if this makes it into this week's release as our adapter is broken in its current state. Thank you very much.

PR #1084, opened 2019-10-21
Restrict /setuid for Prebid supported bidders only

### LINKED ISSUE #1054: Setuid sets unknown bidder in UIDS cookie
The `/setuid` call sets any bidder to UIDS cookie from incoming request without checking whether valid or active that bidder.

Call to `/setuid?bidder=I_DONT_KNOW_WHO_I_AM&gdpr=0&gdpr_consent=&uid=test-uid` returns this UIDS value:
```
{"tempUIDs":{"I_DONT_KNOW_WHO_I_AM":{"uid":"test-uid",
"expires":"2019-10-17T05:59:43.584405534Z"}},"bday":"2019-10-03T05:59:43.584398849Z"}
```

This can leads to unnecessary increasing of UIDS cookie.
So, it is better to respond with `HTTP 400` for invalid/inactive bidder.

mansinahar:
Fixes #1054 

Also, refactors some existing tests for better readability and organization of test cases.

### DIFF
--- a/endpoints/setuid_test.go
+++ b/endpoints/setuid_test.go
@@ -39,7 +39,7 @@ func TestSetUIDEndpoint(t *testing.T) {
 			description:           "Set uid for valid bidder",
 		},
 		{
-			uri:                   "/setuid?bidder=some-bidder&uid=123",
+			uri:                   "/setuid?bidder=unsupported-bidder&uid=123",
 			existingSyncs:         nil,
 			gdprAllowsHostCookies: true,
 			expectedSyncs:         nil,
@@ -55,7 +55,7 @@ func TestSetUIDEndpoint(t *testing.T) {
 			description:           "Don't set uid for an empty bidder",
 		},
 		{
-			uri:                   "/setuid?bidder=some-bidder&uid=123",
+			uri:                   "/setuid?bidder=unsupported-bidder&uid=123",
 			existingSyncs:         map[string]string{"pubmatic": "1234"},
 			gdprAllowsHostCookies: true,
 			expectedSyncs:         nil,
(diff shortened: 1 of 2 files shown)


endpoints/setuid_test.go
```diff
@@ -18,67 +19,163 @@ import (
...
+		existingSyncs         map[string]string
+		gdprAllowsHostCookies bool
+		gdprReturnsError      bool
+		expectedSyncs         map[string]string
+		expectedRespMessage   string
+		expectedResponseCode  int
+		description           string
+	}{
+		{
+			uri:                   "/setuid?bidder=pubmatic&uid=123",
```
SyntaxNode: Is it common in Go to write mini test runners versus breaking them out into individual tests?

SyntaxNode: Talked offline. This is a common design pattern for tests in Go. See:
- https://github.com/golang/go/wiki/TableDrivenTests
- https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

endpoints/setuid_test.go
```diff
@@ -18,67 +19,163 @@ import (
...
+			existingSyncs:         map[string]string{"rubicon": "def"},
+			gdprAllowsHostCookies: true,
+			expectedSyncs:         map[string]string{"pubmatic": "123", "rubicon": "def"},
+			expectedResponseCode:  http.StatusOK,
+			description:           "Add the uid for the requested bidder to the list of existing syncs",
+		},
+		{
+			uri:                  "/setuid?bidder=pubmatic&uid=123&gdpr=0",
```
SyntaxNode: Do these tests cases rely on the assumption that pubmatic enables GDPR?

endpoints/setuid_test.go
```diff
@@ -18,67 +19,163 @@ import (
...
+			expectedSyncs:        nil,
+			gdprReturnsError:     true,
+			expectedResponseCode: http.StatusBadRequest,
+			expectedRespMessage: "No global vendor list was available to interpret this consent string. " +
+				"If this is a new, valid version, it should become available soon.",
+			description: "Return an error if the GDPR string is either malformed or using a newer version that isn't yet supported",
+		},
+		{
+			uri: "/setuid?bidder=pubmatic&uid=123&gdpr=1&gdpr_consent" +
+				"=BONciguONcjGKADACHENAOLS1rAHDAFAAEAASABQAMwAeACEAFw",
```
SyntaxNode: Nitpick: The `=` is on the line with the consent string in this test, but on the above test it's on the previous line. For a moment, I thought the content of the consent string was different due to the inconsistency.

mansinahar: Ah! Thanks for pointing that out

endpoints/setuid_test.go
mansinahar: Well, the tests actually specify if the given host/vendor (in this case `pubmatic`) is allowed under GDPR or not with the `gdprAllowsHostCookies` field in the testCase. We're using the `mockPermsSetUID` implementation of the `Permissions` interface which allows or disallows a vendor based on it's `allowHost` field rather than actually looking at the consent string

PR #1085, opened 2019-10-22
Bid floor

ah-tappx:
Adding bid floor parameter


adapters/tappx/tappx.go
```diff
@@ -143,7 +148,7 @@ func (a *TappxAdapter) buildEndpointURL(params *openrtb_ext.ExtImpTappx, test in
 		thisURI = thisURI + "&ts=" + strconv.Itoa(int(t))
 	}
 
-	thisURI = thisURI + "&v=" + TAPPX_BIDDER_VERSION
+	thisURI = thisURI + "&v=" + TAPPX_BIDDER_VERSION + "&type_cnn=prebid"
```
guscarreon: I know there are test cases that assert the `thisURI` expected values but, do you think it'd be more reliable or maintainable to use golangs URL package to manage queries and other URL related operations? This is entirely up to you.

adapters/tappx/tappx.go
```diff
@@ -39,6 +39,7 @@ type tappxParams struct {
 	Host     string `json:"host"`
 	TappxKey string `json:"tappxkey"`
 	Endpoint string `json:"endpoint"`
+	BidFloor string `json:"bidfloor"`
 }
```
guscarreon: Unless I'm wrong, the `tappxParams struct` doesn't seem to be used anywhere in your code. Are we sure we need it at all?

adapters/tappx/tappx.go
```diff
@@ -143,7 +148,7 @@ func (a *TappxAdapter) buildEndpointURL(params *openrtb_ext.ExtImpTappx, test in
 		thisURI = thisURI + "&ts=" + strconv.Itoa(int(t))
```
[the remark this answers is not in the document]
ah-tappx: We noticed this when writing the tests. Since the parameter added to `thisURI` is a timestamp we cannot predict its value and we don't know how to include this in a written test case. Could you give us some help?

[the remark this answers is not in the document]
guscarreon: The `strconv.Itoa(int(t))` in line 147 seems to return a 13 digit number
```
                //Some random code I wrote to watch the contents of strconv.Itoa(int(t)) with the debugger
   151:         stringWithTime := "=" + strconv.Itoa(int(t))
=> 152:         if stringWithTime == "" {
   153:                 fmt.Printf("ok \n")
   154:         }
(dlv) p stringWithTime
"=1571940732071"
//1234567890123
```
To create the test case withouth knowing the exact digits that `strconv.Itoa(int(t))` maybe we can assert with a regex match:
```
func TestNewTestToAssertEndpointURL() {    

    // build endpoint URL here

    assert.True(t,regexp.MatchString(`&ts=[0-9]\{13\}`,endpointURL)
}
```
Or maybe we could use the `time` library's `Parse` function [https://golang.org/pkg/time/#Parse|https://golang.org/pkg/time/#Parse] to turn the string unto a valid `Time`object and `assert.NoError` to make sure the test passes
```
func TestNewTestToAssertEndpointURL() {    

    // build endpoint URL here

    lastDigitsOfEndpoint := //String function to strip the last n characters from the endpoint

    _, err := Parse(layout, lastDigitsOfEndpoint)

    assert.NoError(t,err,"Error: `lastDigitsOfEndpoint` could not be parsed to a valid Time object")
}
```

ah-tappx:
added new commit with changes addressing mentioned issues

ah-tappx:
Hello,
Do we have an estimate time on when will this be approved and deployed?

Thank you

adapters/tappx/tappx_test.go
```diff
@@ -2,10 +2,31 @@ package tappx
...
+	var tappxExt openrtb_ext.ExtImpTappx
+	tappxExt.Host = "example.host.tappx.com"
+	tappxExt.Endpoint = "DUMMYENDPOINT"
+	tappxExt.TappxKey = "dummy-tappx-key"
+
+	url, err := adapter.buildEndpointURL(&tappxExt, test)
+
+	match, err := regexp.MatchString(`&ts=[0-9]{13}`, url)
+	if err != nil {
+		//something happened during regex validation
```
SyntaxNode: Should that fail the test?

SyntaxNode:
Soon. :)
I'll track this tomorrow and I'll request a review from @guscarreon. We should be able to release soon after it merges in.

adapters/tappx/tappx.go
```diff
@@ -136,16 +138,30 @@ func (a *TappxAdapter) buildEndpointURL(params *openrtb_ext.ExtImpTappx, test in
 		}
 	}
 
-	thisURI := host + params.Endpoint + "?tappxkey=" + params.TappxKey
+	thisURI.Path += params.Endpoint
+
+	queryParams := url.Values{}
+
+	if reqKey == "" {
+		return "", &errortypes.BadInput{
```
mansinahar: It would be easier to just change the `if` condition on L108 to be:
```
if params.TappxKey == "" {
  return "", &errortypes.BadInput{
    Message: "Tappx key undefined",
  }
}
```

and fail early

adapters/tappx/tappx_test.go
```diff
@@ -2,10 +2,31 @@ package tappx
...
+	var test int
+	test = 0
+	var tappxExt openrtb_ext.ExtImpTappx
+	tappxExt.Host = "example.host.tappx.com"
+	tappxExt.Endpoint = "DUMMYENDPOINT"
+	tappxExt.TappxKey = "dummy-tappx-key"
+
+	url, err := adapter.buildEndpointURL(&tappxExt, test)
+
+	match, err := regexp.MatchString(`&ts=[0-9]{13}`, url)
```
mansinahar: I'd say assert the url to the actual expected string rather than doing a regexp match

adapters/tappx/tappx_test.go
```diff
@@ -2,10 +2,31 @@ package tappx
...
+	"github.com/stretchr/testify/assert"
 	"net/http"
+	"regexp"
 	"testing"
 )
 
 func TestJsonSamples(t *testing.T) {
 	adapterstest.RunJSONBidderTest(t, "tappxtest", NewTappxBidder(new(http.Client), "https://{{.Host}}"))
 }
+
```
mansinahar: Would be good to add some more test cases for the `buildEndpointURL` method like passing an invalid `host` which will lead to `url.Parse(host)` throwing an error, passing empty `tappxkey`, etc

Tip: You can look at table driven test cases to make it easier to add new test cases and reduce redundant code with multiple test cases. Ref: https://github.com/golang/go/wiki/TableDrivenTests

ah-tappx:
Hello again,
Just asking about the timing of this pull request to be merged and deployed since one of our publishers is pushing us about it.

Thank you

guscarreon:
> Hello again,
> Just asking about the timing of this pull request to be merged and deployed since one of our publishers is pushing us about it.
> 
> Thank you

@ah-tappx my teammates @mansinahar and @SyntaxNode left some comments above. Can they be addressed? The PR needs two approvals. So there's only one approval.

adapters/tappx/tappx_test.go
ah-tappx: The regex match was purposely added in order to validate the ts parameter. You can read more about this in one of the first requested changes in this pull request by @guscarreon .
The full URL cannot be asserted since ts value is dynamic. I will add the rest of the URL to the regexp match to make it more robust but I think this cannot be  skipped.

adapters/tappx/tappx_test.go
ah-tappx: added test case for url.Parse(host) throwing error.
empty parameter tests should be already covered through JSONBidder tests
there are still some errors that aren't covered like resolving macros or JSON parsing errors, if those are necessary please send us a full list and we will try to write them all

adapters/tappx/tappx_test.go
mansinahar: Ah! I missed the `ts` part before. Makes sense. Thanks for the clarification :)

adapters/tappx/tappx_test.go
mansinahar: Sounds good. Yeah, that should cover most cases. 

The only reason behind that suggestion was that `JSONBidderTests` are more testing the `MakeRequest` and `MakeBids` function and even though `buildEndpointURL` is used as part of the `MakeRequest` function,  breaking up the unit tests by each function helps to get better coverage.  For example, if we have good unit tests written for `buildEndpointURL` and all the other functions that `MakeRequest` and `MakeBids` call then we don't have to worry about the different places those functions can fail when writing unit tests for `MakeRequest` and `MakeBids` themselves. It just makes the tests more readable and manageable.

I understand that doing that in this PR can make it a little bloated but just something to consider for next time :)

PR #1087, opened 2019-10-22
Add SSL cert for accessing stored request API

guscarreon:
The stored request SSL certificate must be stored in our private repository of files. Core PBS must be capable of reading an SSL certificate from a path specified in the `Configuration` struct.


router/router.go
```diff
@@ -175,12 +175,22 @@ func New(cfg *config.Configuration, rateConvertor *currencies.RateConverter) (r
 	r = &Router{
 		Router: httprouter.New(),
 	}
+
+	// For bid processing, we need both the hardcoded certificates and the certificates found in container's
+	// local file system
+	certPool := ssl.GetRootCAPool()
+	var readCertErr Error
+	certPool, readCertErr = ssl.AppendPEMFileToRootCAPool(certPool, cfg.PemCertsFile)
```
hhhjort: Do we need a check if the string is empty?

ssl/ssl.go
```diff
@@ -16,6 +18,28 @@ func GetRootCAPool() *x509.CertPool {
...
+		//read file and place it's contents in `pemCerts`
+		pemCerts, err := ioutil.ReadFile(pemFileName)
+		if err != nil {
+			return certPool, fmt.Errorf("Failed to read file %s: %v", pemFileName, err)
+		}
+
+		//`pemCerts` has been obtained, append to certPool
+		certPool.AppendCertsFromPEM(pemCerts)
+	} else {
+		return certPool, fmt.Errorf("Path to local PEM file containing SSL certificates is\n")
```
hhhjort: This error message is confusing ... is it really an error to fail to read a cert file when none was given?

guscarreon: Removed

router/router.go
guscarreon: The `AppendPEMFileToRootCAPool(certPool *x509.CertPool, pemFileName string)` function checks for an empty string but I removed the error it was returning when the string is empty because as of right now, we have no certificates file and it seems odd to return an error if the certificates file has not been defined in the configuration.

Do you prefer the calling function to do the empty string check?

ssl/ssl_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCertsFromFilePoolExists(t *testing.T) {
+	// Load hardcoded certificates found in ssl.go
+	certPool := GetRootCAPool()
+
+	// Assert loaded certificates by looking at the lenght of the subjects array of strings
```
SyntaxNode: Typo: lenght

ssl/ssl_test.go
```diff
@@ -0,0 +1,54 @@
...
+	subjects := certPool.Subjects()
+	hardCodedSubNum := len(subjects)
+	assert.True(t, hardCodedSubNum > 0)
+
+	// Load certificates from file
+	certificatesFile := "mockcertificates/mock-certs.pem"
+	certPool, err := AppendPEMFileToRootCAPool(certPool, certificatesFile)
+
+	// Assert loaded certificates by looking at the lenght of the subjects array of strings
```
SyntaxNode: Typo: lenght

ssl/ssl_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func TestCertsFromFilePoolDontExist(t *testing.T) {
+	// Empty certpool
+	var certPool *x509.CertPool = nil
+
+	// Load certificates from file
+	certificatesFile := "mockcertificates/mock-certs.pem"
+	certPool, err := AppendPEMFileToRootCAPool(certPool, certificatesFile)
+
+	// Assert loaded certificates by looking at the lenght of the subjects array of strings
```
SyntaxNode: Typo: lenght

ssl/ssl_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func TestAppendPEMFileToRootCAPoolFail(t *testing.T) {
+	// Empty certpool
+	var certPool *x509.CertPool
+
+	// Load certificates from file
+	fakeCertificatesFile := "mockcertificates/NO-FILE.pem"
+	certPool, err := AppendPEMFileToRootCAPool(certPool, fakeCertificatesFile)
+
+	// Assert loaded certificates by looking at the lenght of the subjects array of strings
```
SyntaxNode: This comment is likely a copy/paste from the other tests, but doesn't make sense in this instance because an error is expected and the length is never checked.

PR #1097, opened 2019-11-01
Currency handling fix

### LINKED ISSUE #1067: Port currency fix from PBS-Java
There's a currency conversion issue in PBS. We've fixed this in PBS-Java with https://github.com/rubicon-project/prebid-server-java/pull/457

Here are the requirements and some test cases:

1) the client only ever specifies one entry in the `cur` array: the one expected by their adServer
If there are more than one cur specified, use the first and warn.

2) This cur value must be sent to each adapter

3) PBS must try to convert bid responses to the first currency in the array

4) if PBS can't convert a bid, then it should error with "unable to covert bid currency XYZ to desired ad server currency ABC."

Test cases:

1) ad-server-currency is USD. (whether defined in the `cur` field or in the config is the same behavior)
2) bidderA bids are in EUR. they get converted to USD
3) bidderB bids are in GBP. they get converted to USD
4) bidderC bids are in USD. no conversion needed

hhhjort:
Resolves #1067

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -256,6 +256,11 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) []error {
 		return []error{errors.New("request.imp must contain at least one element.")}
 	}
 
+	if len(req.Cur) > 1 {
+		req.Cur = req.Cur[0:1]
+		errL = append(errL, &errortypes.Warning{Message: fmt.Sprintf("A prebid request can only process one currency. Taking the first currency in the list, %s, as the active currency", req.Cur[0])})
+	}
+
 	var aliases map[string]string
 	if bidExt, err := deps.parseBidExt(req.Ext); err != nil {
 		return []error{err}
@@ -1172,7 +1177,7 @@ func writeError(errs []error, w http.ResponseWriter, labels *pbsmetrics.Labels) 
 func fatalError(errL []error) bool {
 	for _, err := range errL {
 		errCode := errortypes.DecodeError(err)
-		if errCode != errortypes.BidderTemporarilyDisabledCode || errCode == errortypes.BlacklistedAppCode || errCode == errortypes.BlacklistedAcctCode {
+		if errCode != errortypes.BidderTemporarilyDisabledCode && errCode != errortypes.WarningCode {
 			return true
 		}
 	}
--- a/endpoints/openrtb2/auction_test.go
+++ b/endpoints/openrtb2/auction_test.go
@@ -865,6 +865,52 @@ func validRequest(t *testing.T, filename string) string {
 	return string(requestData)
 }
 
+func TestCurrencyTrunc(t *testing.T) {
+	deps := &endpointDeps{
+		&nobidExchange{},
+		newParamsValidator(t),
+		&mockStoredReqFetcher{},
+		empty_fetcher.EmptyFetcher{},
+		empty_fetcher.EmptyFetcher{},
+		&config.Configuration{},
+		pbsmetrics.NewMetrics(metrics.NewRegistry(), openrtb_ext.BidderList(), config.DisabledMetrics{}),
+		analyticsConf.NewPBSAnalytics(&config.Analytics{}),
+		map[string]string{},
+		false,
+		[]byte{},
+		openrtb_ext.BidderMap,
+	}
+
+	ui := uint64(1)
+	req := openrtb.BidRequest{
+		ID: "someID",
+		Imp: []openrtb.Imp{
+			{
+				ID: "imp-ID",
+				Banner: &openrtb.Banner{
+					W: &ui,
+					H: &ui,
+				},
+				Ext: json.RawMessage("{\"appnexus\": {\"placementId\": 5667}}"),
+			},
+		},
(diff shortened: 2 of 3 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -256,6 +256,11 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) []error {
 		return []error{errors.New("request.imp must contain at least one element.")}
 	}
 
+	if len(req.Cur) > 1 {
```
SyntaxNode: Please add a test for this new case.

errortypes/errortypes.go
```diff
@@ -155,6 +156,20 @@ func (err *BidderTemporarilyDisabled) Code() int {
 	return BidderTemporarilyDisabledCode
 }
 
+// Warning is a genric warning type, not a serious error
```
benjaminch: Maybe a typo ? Should be `generic`

errortypes/errortypes.go
```diff
@@ -12,6 +12,7 @@ const (
 	BidderTemporarilyDisabledCode
 	BlacklistedAcctCode
 	AcctRequiredCode
+	WarningCode
```
benjaminch: This is just a question I have: Warning seems to be more generic than other code listed here. Should it rather be scoped to the currency warning? Something like `OneCurrencySupportedCode` or anything like that?

endpoints/openrtb2/auction_test.go
```diff
@@ -865,6 +865,52 @@ func validRequest(t *testing.T, filename string) string {
...
+			},
+		},
+		Site: &openrtb.Site{
+			ID: "myID",
+		},
+		Cur: []string{"USD", "EUR"},
+	}
+
+	errL := deps.validateRequest(&req)
+	if len(errL) > 0 {
```
mansinahar: We should probably assert here for `errL` len to be greater than 1 because we will always expect at least one err to be returned here, the multiple currency warning one

endpoints/openrtb2/auction_test.go
```diff
@@ -865,6 +865,52 @@ func validRequest(t *testing.T, filename string) string {
...
+		for _, err := range errL {
+			if err.Error() != "A prebid request can only process one currency. Taking the first currency in the list, USD, as the active currency" {
+				t.Errorf(err.Error())
+			}
```
mansinahar: Would be good to assert the error code and the error type as well for the multiple currency warning

mansinahar:
@hhhjort what do you think about moving https://github.com/prebid/prebid-server/blob/master/exchange/bidder.go#L143-L145 to the `validateRequest` method as well? I am not sure if there's any effect of sending the default currency to an adapter rather than no currencies. I can only assume it to be a good thing :)

Also, if we do this, might be just good to just create a new `validateCurrency` method

endpoints/openrtb2/auction_test.go
hhhjort: That may be a bit overkill. A lot would have to go wrong to get an error with that message, but somehow not have the correct code/type.

hhhjort:
Upon further consideration, I think it might be best to leave the default currency code where it is. We might have a use case come up where multiple currencies in the response is allowed, or we want to attach some special meaning to the zero length request currency. This leaves the default question where it is needed, and not have to shift code around.

I am open to moving it if there is a good reason, but currently I don't believe it is worth the effort and reduced flexibility it will entail.

errortypes/errortypes.go
hhhjort: Missed this, thanks

mansinahar:
The reason behind my suggestion was so that we have the validation and default setting of the currency in one place, if the currency is empty or not, i.e. in the `validateRequest` method and also, modifying the request at the time of parsing the response just doesn't seem like a right thing to do.

That being said, I understand that this change is not the scope of this PR but I came across it and thought maybe we should change it if it's not that difficult to do. So, I will leave this up to you.

endpoints/openrtb2/auction_test.go
mansinahar: Well, I was suggesting for those assertions more to safeguard us from situations where somebody touching that code and potentially changing the error type for it, from `warning` to something else, maybe a fatal error type, then the tests can help catch that.

hhhjort: If someone just randomly changes the error type, that is a pretty blatant and silly change. If it something more subtle like the error code changing, then the test will probably target the new code, unless something really subtle is going on. In any case, I had already beefed up the test since it wasn't much work.

errortypes/errortypes.go
hhhjort: The most generic error code is not having a code at all. The purpose of the error codes is to aid in automation, that is a client making the requests and receiving the responses can just look at the code rather than parse out the error message and behave accordingly. I would not have added the code at all except we have a piece of code that whitelists which error codes are not fatal codes, and the request can continue. So I felt a generic warning code would cover all the use cases and keep us from generating more codes in the future that will not be used for anything beyond adding to the non-fatal white list.

benjaminch: Alright :)

PR #1098, opened 2019-11-01
Adding translatecategories flag to includebrandcategory

camrice:


exchange/exchange.go
```diff
@@ -355,14 +355,22 @@ func applyCategoryMapping(ctx context.Context, requestExt openrtb_ext.ExtRequest
...
 	var publisher string
 	var err error
+	var translateCategories = true
 
 	if includeBrandCategory && brandCatExt.WithCategory {
-		//if ext.prebid.targeting.includebrandcategory present but primaryadserver/publisher not present then error out the request right away.
-		primaryAdServer, err = getPrimaryAdServer(brandCatExt.PrimaryAdServer) //1-Freewheel 2-DFP
-		if err != nil {
-			return res, seatBids, err
+		//if TranslateCategories is nil, default category translation to enabled
```
SyntaxNode: You might be fine without this comment, especially since the the following code line doesn't set the default to enabled... that happens a few lines above.

exchange/exchange_test.go
```diff
@@ -526,7 +526,8 @@ func newExtRequest() openrtb_ext.ExtRequest {
 		},
 	}
 
-	brandCat := openrtb_ext.ExtIncludeBrandCategory{PrimaryAdServer: 1, WithCategory: true}
+	translateCat := true
```
SyntaxNode: Consider renaming to `translateCategories` to match the naming of exchange.go.

exchange/exchange_test.go
```diff
@@ -568,6 +569,61 @@ func newExtRequestNoBrandCat() openrtb_ext.ExtRequest {
 	}
 }
 
+func newExtRequestTranslateCatNil() openrtb_ext.ExtRequest {
```
SyntaxNode: Please move this helper method below the TestCategoryMappingTranslateCategoriesNil (the first place it's called) for better readability. Consider accepting a TranslateCategories argument so you don't need two helper methods.

endpoints/openrtb2/video_auction_test.go
```diff
@@ -106,10 +106,12 @@ func TestCreateBidExtension(t *testing.T) {
 		Increment: 0.1,
 	})
 
+	translateCat := true
```
SyntaxNode: Consider renaming to translateCategories to match the naming of video_auction.go.

exchange/exchange.go
camrice: Agreed, this isn't the right place for this comment. I removed it as translateCategories is declared as being true above, so the comment is probably not necessary.

exchange/exchange_test.go
camrice: Updated to `translateCategories`

exchange/exchange_test.go
camrice: Moved this function to be below the first test that uses it. Also updated it to take a `translateCategories` argument so that there is only one helper function for both tests.

endpoints/openrtb2/video_auction_test.go
camrice: Updated to `translateCategories`

PR #1099, opened 2019-11-05
Update Dockerfile and Makefile 

mansinahar:


Dockerfile
```diff
@@ -16,12 +16,16 @@ RUN apt-get update && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 ENV CGO_ENABLED 0
 COPY ./ ./
+RUN go mod vendor
+RUN go mod tidy
+ARG TEST="true"
+RUN if [ "$TEST" != "false" ]; then ./validate.sh ; fi
 RUN go build -mod=vendor .
```
guscarreon: Is `ARG` defining the environment variable `"true"`? if it is going to be hardcoded, why not just running `validate.sh`?

mansinahar: That's actually the default value. I understand it can be a little confusing but the `ARG` value that you pass in the command line will override that. And the reason for the `true` default value is we wanna always run the tests as part of building the image unless specified otherwise.

guscarreon: Awesome. I already approved Mansi

PR #1100, opened 2019-11-05
update static bidder params for rubicon video to follow the json marshaling names

Aadeshp:
Want the field names in /bidder/params to follow the exactly what the code expects when deserializing the JSON request. Its misleading the the /bidder/params say that they key is called VideoSizeID when it really needs to be passed as size_id in the request


Aadeshp:
cc: @bretg

PR #1101, opened 2019-11-05
enforce video.size_id for video imps in rubicon adapter

Aadeshp:
cc: @bretg 

Based on conversations with Michael Moschovas over at Rubicon, it seems like video.size_id needs to be defined in every video request sent to rubicon

### DIFF
--- a/adapters/rubicon/rubicon.go
+++ b/adapters/rubicon/rubicon.go
@@ -5,12 +5,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/golang/glog"
-	"github.com/prebid/prebid-server/pbs"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
+
+	"github.com/golang/glog"
+	"github.com/prebid/prebid-server/pbs"
 
 	"golang.org/x/net/context/ctxhttp"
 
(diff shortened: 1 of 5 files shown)


mansinahar:
@Aadeshp Can you please also update the existing tests to account for this change? It is leading to failed tests. Also, it'd be great if you have a chance and can add a new test case for this that asserts a bad input error when the video size is not specified.

mansinahar:
@rpanchyk @bretg Can you guys please take a look at this PR and make sure this the expected behavior for the Rubicon adapter. Once you give a thumbs up, we can fix the failing test.

bretg:
Sorry - I get a lot of Prebid email and this one fell between the cracks. Please assign things to me or slack me if you want to get my attention.

Anyhow, I don't know where this one's coming from or that this is the right solution. Will work the team internally to figure it out.

bretg:
Ok, worked through this with Michael. I understand and approve this change. 

Really the problem is that Prebid Server doesn't properly support outstream -- but that's a story for another day.

SyntaxNode:
> Please assign things to me or slack me if you want to get my attention.

Will do next time. Sorry about that.

> Ok, worked through this with Michael. I understand and approve this change.

Thank you.

> Really the problem is that Prebid Server doesn't properly support outstream -- but that's a story for another day.

Sounds like a good item for the next agenda. 👍

bretg:
I've asked Ruslan to review, but do we really need this change now? My understanding is that it was for a transition period that may have come and gone. (?)

Aadeshp:
@bretg from my discussions with Michael, it sounded like this is a requirement for all video impressions, didn't mention anything about a transition period

bretg:
Very well. Approved.

SyntaxNode:
The test in this PR still needs to be fixed.

PR #1104, opened 2019-11-06
JSON block that has a full data-center specific URL cache info

guscarreon:


exchange/exchange.go
```diff
@@ -587,6 +590,28 @@ func (e *exchange) makeBid(Bids []*pbsOrtbBid, adapter openrtb_ext.BidderName) (
...
 
+// If bid got cached inside `(a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client, targData *targetData, bidRequest *openrtb.BidRequest, ttlBuffer int64, defaultTTLs *config.DefaultTTLs, bidCategory map[string]string)`,
+// a UUID should be found inside `a.cacheIds` or `a.vastCacheIds`. This function returns the UUID along with the internal cache URL
+func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auc *auction) (openrtb_ext.ExtBidPrebidCacheBids, bool) {
+	var cacheInfo openrtb_ext.ExtBidPrebidCacheBids
+	var cacheUUID string
+	var found bool = false
+
+	if auc != nil {
+		if bid.bidType == openrtb_ext.BidTypeVideo {
```
hhhjort: There may be use cases in the future for video bids to get a full bid cached, or non-video bids to get some sort of vast cache. Also we cannot 100% trust adapters to report the `bidType` correctly. We should just check for a bid cache first, and if that does not exist, check for a VAST cache.

exchange/exchange.go
```diff
@@ -587,6 +590,28 @@ func (e *exchange) makeBid(Bids []*pbsOrtbBid, adapter openrtb_ext.BidderName) (
...
+			cacheUUID, found = auc.vastCacheIds[bid.bid]
+		} else {
+			cacheUUID, found = auc.cacheIds[bid.bid]
+		}
+
+		if found {
+			cacheInfo.CacheId = cacheUUID
+			cacheInfo.Url = e.cache.GetPutUrl() + "?uuid=" + cacheUUID //if there's no UUID, should we still define a url?
```
hhhjort: Why are you calling this method `GetPutUrl()`? Wouldn't be `GetGetUrl()`, as it is the URL that clients will use to issue the GET request to get the cached content?

prebid_cache_client/client.go
```diff
@@ -72,6 +75,10 @@ func (c *clientImpl) GetExtCacheData() (string, string) {
 	return c.externalCacheHost, c.externalCachePath
 }
 
+func (c *clientImpl) GetPutUrl() string {
```
hhhjort: For this PR, we want to use the URL we can assemble from `getExtCacheData()`, we don't need a new method. When we return a URL to a client, we always want to return an external URL, not an PBS internal URLs, as they will be useless to the client if the internal differs from the external.

PR #1109, opened 2019-11-12
Remove Go 1.11 Build Target

SyntaxNode:
There doesn't seem to be any issue with Go 1.13, so I think it's time to drop 1.11 from the build tasks.


mansinahar:
@SyntaxNode Can we please also update the `README.md` file to point to 1.12 as the suggested Go version to install as that's the minimum version we're supporting? 

Also, I guess we can now consider removing the note mentioned here: https://github.com/prebid/prebid-server/blob/master/docs/developers/automated-tests.md#automated-tests as that is true only for Go 1.11

SyntaxNode:
> @SyntaxNode Can we please also update the `README.md` file to point to 1.12 as the suggested Go version to install as that's the minimum version we're supporting?
> 
> Also, I guess we can now consider removing the note mentioned here: https://github.com/prebid/prebid-server/blob/master/docs/developers/automated-tests.md#automated-tests as that is true only for Go 1.11

Thank you for pointing those out. Fixed.

PR #1119, opened 2019-11-19
Set "Secure" on Same SIte cookies

### LINKED ISSUE #900: Set Prebid Server Cookie With SameSite=None and Secure

hhhjort:
Fixes #900

### DIFF
--- a/usersync/cookie.go
+++ b/usersync/cookie.go
@@ -125,7 +125,6 @@ func (cookie *PBSCookie) ToHTTPCookie(ttl time.Duration) *http.Cookie {
 		Value:   b64,
 		Expires: time.Now().Add(ttl),
 		Path:    "/",
-		Secure:  true,
 	}
 }
 
@@ -194,9 +193,11 @@ func (cookie *PBSCookie) SetCookieOnResponse(w http.ResponseWriter, setSiteCooki
 		currSize = len([]byte(httpCookie.String()))
 	}
 
-	uidsCookieStr := httpCookie.String()
+	var uidsCookieStr string
 	var sameSiteCookie *http.Cookie
 	if setSiteCookie {
+		httpCookie.Secure = true
+		uidsCookieStr := httpCookie.String()
 		uidsCookieStr += SameSiteAttribute
 		sameSiteCookie = &http.Cookie{
 			Name:    SameSiteCookieName,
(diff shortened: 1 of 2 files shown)


usersync/cookie.go
```diff
@@ -193,19 +193,24 @@ func (cookie *PBSCookie) SetCookieOnResponse(w http.ResponseWriter, setSiteCooki
 		currSize = len([]byte(httpCookie.String()))
 	}
 
-	uidsCookieStr := httpCookie.String()
+	var uidsCookieStr string
 	var sameSiteCookie *http.Cookie
 	if setSiteCookie {
+		httpCookie.Secure = true
+		uidsCookieStr := httpCookie.String()
```
guscarreon: `uidsCookieStr` has already been declared and it doesn't need the `:=` operator. The `uidsCookieStr = httpCookie.String()` assignation happens in both the `true` and `false` branches of if statement in line 198, so maybe we can simplify:
```
196 -     var uidsCookieStr string
    +     uidsCookieStr := httpCookie.String()
197       var sameSiteCookie *http.Cookie
198       if setSiteCookie {
199           httpCookie.Secure = true
200 -         uidsCookieStr = httpCookie.String()
201           uidsCookieStr += SameSiteAttribute
202           sameSiteCookie = &http.Cookie{
203               Name:    SameSiteCookieName,
204               Value:   SameSiteCookieValue,
205               Expires: time.Now().Add(ttl),
206               Path:    "/",
207               Secure:  true,
208           }
209           sameSiteCookieStr := sameSiteCookie.String()
210           sameSiteCookieStr += SameSiteAttribute
211           w.Header().Add("Set-Cookie", sameSiteCookieStr)
212 -     } else {
213 -         uidsCookieStr = httpCookie.String()
214       }
215       w.Header().Add("Set-Cookie", uidsCookieStr)
216   }
usersync/cookie.go
```

hhhjort: Good catch, that is actually a bug. If it was simply it didn't need the `:=` operator, then Go would flag that as a compiler error. What Go actually did is see that I used `uidsCookieStr :=` inside of a scope `{}`, so assumed I was declaring a new variable of the same name, that would only exist inside that code block. That is valid, but not the intended behavior.

guscarreon: Wow

SyntaxNode: Yeah, variable shadowing can be tricky in Go. Other languages ban it outright, like C#.

PR #1120, opened 2019-11-22
Update sync url for BidderGrid adapter

TheMediaGrid:


config/config.go
```diff
@@ -490,7 +490,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderEPlanning, "https://ads.us.e-planning.net/uspd/1/?du=https%3A%2F%2Fads.us.e-planning.net%2Fgetuid%2F1%2F5a1ad71d2d53a0f5%3F"+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Deplanning%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
 	// openrtb_ext.BidderFacebook doesn't have a good default.
 	// openrtb_ext.BidderGamma doesn't have a good default.
-	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderGrid, "https://grid.bidswitch.net/sp_sync?sp_id=prebid&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dgrid%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderGrid, "https://x.bidswitch.net/check_uuid/"+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dgrid%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24%7BBSW_UUID%7D?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}")
```
SyntaxNode: I think this may the first instance of including redirect in the url path instead of in a query parameter. Have you tested this locally? I'm concerned about using `url.QueryEscape` for path insertion.

TheMediaGrid: Hello, yes, we have tested it locally - everything works fine

SyntaxNode: Gotcha. I guess your `check_uuid` endpoint is using a query unescape on the last segment.

TheMediaGrid:
Hey guys, @SyntaxNode , @guscarreon 
Could you please tell us, is there a way to understand that a partner has updated its prebid-server version, in case it is a third-party hosted solution (like appnexus or rubicon hosted) so we know that the updates to our adapter are in prod?
Do you plan to pass somehow the prebid-server version?
Thanks in advance

SyntaxNode:
We've talked it before and I personally agree there's value in exposing the version number from the service itself. I'll bring it up to the team. We don't have that available today.

The compiled Prebid Server app isn't aware of its version number. Internally, we use the version number in the image name for tracking.

TheMediaGrid:
Hi guys @SyntaxNode , @guscarreon ,
we’ll be very thankful if you could tell us about the APN-hosted prebid-server updates frequency (once/twice a week/month) so we could approximately know about the version is in production right now
And one more thing is about the update process itself. Is it a kind of ab testing update or the full update a time?
We’ll also appreciate it if you could tell about where we could gather all this info if such a source exists.

SyntaxNode:
This PR is included in the 0.91.0 version and is released to Xandr/AppNexus hosted Prebid Server instances.

> Hi guys @SyntaxNode , @guscarreon ,
> we’ll be very thankful if you could tell us about the APN-hosted prebid-server updates frequency (once/twice a week/month) so we could approximately know about the version is in production right now

We generally release to production the same day as a Prebid Server tagged release.

> And one more thing is about the update process itself. Is it a kind of ab testing update or the full update a time?

There is short overlap of versions during a release in which requests may be serviced by either the current version or the previous version.

> We’ll also appreciate it if you could tell about where we could gather all this info if such a source exists.

I am not aware of documentation from Xandr specific to the Prebid Server release schedule or process. Do you have an existing contact within Xandr? They would likely be able to assist better on this concern.

I agree it would be beneficial to know the version number of the Prebid Server instance handling your request. I'll be discussing with the Prebid.org team to explore ways of making that accessible. Community proposals and contributions are always welcomed.

TheMediaGrid:
Hi @SyntaxNode , @guscarreon ,
Thanks for the explicit explanation and assistance here,
let us ask you one more portion of questions,
could you confirm that all the AppNexus hosted Prebid Server instances now updated at leaset to the 0.91 prebid-server version?
We ask these questions as we see that cookie_sync config still returns the old sync url for the grid bidder in Safari and Firefox
Снимок экрана 2019-12-13 в 16 39 36
Снимок экрана 2019-12-13 в 17 05 32
and returns nothing in Chrome:
Снимок экрана 2019-12-13 в 16 33 25
Maybe you have an answer, was it expected or not?
The domain where it was observed is https://cafedelites.com/
Thanks in advance

SyntaxNode:
Thank you for providing in-depth details on what you're seeing.

I can confirm all Xandr/AppNexus Prebid Sever instances are running at least 0.91.0. I was very surprised to see the old sync url being used. I investigated and there is a conflict with an internal config service. I'm fixing that and you will see the change on Monday.

TheMediaGrid:
Hi @SyntaxNode,

Thanks for fixing the issue, now we see that the new sync URL is called.
However, we have a couple of new questions regarding syncs and we'll appreciate your assistance here very much.

The thing is that we see a syncs being called properly under the US location in Safari and Mozilla browsers:
Mozila US
safari US

At the same time sync, we’ve never had a chance to catch syncs from Chrome browser, we always see the same picture:
Chrome US + EEA
(this behaviour doesn’t depends on site, but just to the ref we were observing all this on https://sallysbakingaddiction.com/)

Then we tried to visit the same site under the EEA country using the VPN. The first thing that you see when visiting some site in the EEA country is consent banner, which in this case looks like on the picture:
consent banner

We pushed “Accept All” button, then refreshed the page, and tried to find sync, so this is the result:
In chrome:
Chrome US + EEA
chorme params

In Safari:
safai EEA

And finally in Mozilla:
Mozila EEA
Mozilla params


While we see that there are 3 bidders in the configuration in all the browsers, the syncs in the US are called for all 3 bidders only in Safari and Mozilla and sync in EEA are also called only in Safari and Mozilla with the only exclusion of Grid syncs.
We haven’t got a chance to catch syncs in Chrome browser whatever location to use

Our question here, is that the expected behaviour?
Please let us know if something was wrong during our observation or what could be done on our side to improve the situation

Thanks in advance

SyntaxNode:
This is expected behavior.

The cookie sync endpoint will respond with an empty `bidder_status` array if the server already has existing syncs for the bidders. The root cause here for Firefox (and likely Safari as well) is that the browser is intentionally blocking the cookies as part of the browser's built in tracker blocking. Firefox is not sending the user sync cookie (you can see this for yourself in the dev tools cookie tab) and therefore the cookie sync endpoint is always responding with a status of `no_cookie` and requesting the client to sync with all bidders. Firefox will not allow the cookies to be written and so all requests look like this. If you disable the Firefox privacy feature (called "Enhanced Tracking Protection") you will see one more `no_cookie` request followed by the same response you're seeing in Chrome. Conversely, if you clear cookies in Chrome, you'll see that the next call returns a full bidder list to request the browser sync with those bidders. 

If you would like to chat further about this, please open an issue. That'll be easier to track instead of continuing on this merged PR.

PR #1121, opened 2019-11-22
Setuid Fix

SyntaxNode:
The great majority, but not all (cough... AppNexus... cough) adapters use the same cookie name as their adapter name. Fixing the validation with respect to that.


endpoints/setuid.go
```diff
@@ -25,8 +26,14 @@ const (
 	chromeiOSStrLen = len(chromeiOSStr)
 )
 
-func NewSetUIDEndpoint(cfg config.HostCookie, perms gdpr.Permissions, pbsanalytics analytics.PBSAnalyticsModule, metrics pbsmetrics.MetricsEngine) httprouter.Handle {
+func NewSetUIDEndpoint(cfg config.HostCookie, syncers map[openrtb_ext.BidderName]usersync.Usersyncer, perms gdpr.Permissions, pbsanalytics analytics.PBSAnalyticsModule, metrics pbsmetrics.MetricsEngine) httprouter.Handle {
 	cookieTTL := time.Duration(cfg.TTL) * 24 * time.Hour
+
+	validFamilyNameMap := make(map[string]struct{})
```
SyntaxNode: The other methods which need syncer information take the entire map. I thought it best to keep the signature for this endpoint similar and pull out an optimized name hashset. The hashset creation only runs once on startup.

endpoints/setuid.go
```diff
@@ -44,64 +51,64 @@ func NewSetUIDEndpoint(cfg config.HostCookie, perms gdpr.Permissions, pbsanalyti
...
-		}
 
-		if err := validateBidder(bidder); err != nil {
+		familyName, err := getFamilyName(query, validFamilyNameMap)
+		if err != nil {
 			w.WriteHeader(http.StatusBadRequest)
 			w.Write([]byte(err.Error()))
 			metrics.RecordUserIDSet(pbsmetrics.UserLabels{
 				Action: pbsmetrics.RequestActionErr,
-				Bidder: openrtb_ext.BidderName(bidder),
```
SyntaxNode: Family name is not necessarily known, so omitting from metrics.

endpoints/setuid.go
```diff
@@ -44,64 +51,64 @@ func NewSetUIDEndpoint(cfg config.HostCookie, perms gdpr.Permissions, pbsanalyti
...
 			})
 			so.Status = http.StatusBadRequest
 			return
 		}
-		so.Bidder = bidder
+		so.Bidder = familyName
```
SyntaxNode: Does this assignment look strange? It's the same as we did before, but the names weren't clear. We'll need to address this in future work. Keeping the same behavior for now.

endpoints/setuid.go
```diff
@@ -44,64 +51,64 @@ func NewSetUIDEndpoint(cfg config.HostCookie, perms gdpr.Permissions, pbsanalyti
...
 		pc.SetCookieOnResponse(w, setSiteCookie, &cfg, cookieTTL)
 	})
 }
 
-func validateBidder(bidderName string) error {
-	if bidderName == "" {
-		return errors.New(`"bidder" query param is required`)
+func getFamilyName(query url.Values, validFamilyNameMap map[string]struct{}) (string, error) {
+	// The family name is bound to the 'bidder' query param. In most cases, these values are the same.
+	familyName := query.Get("bidder")
```
SyntaxNode: Moved the `query.Get` down here to keep the concept of bidder->familyName contained within this method.

endpoints/setuid_test.go
```diff
@@ -211,40 +229,53 @@ func TestSiteCookieCheck(t *testing.T) {
...
+
+		name, err := getFamilyName(test.urlValues, validFamilyNameMap)
+
+		assert.Equal(t, test.expectedName, name, test.description)
+
+		if test.expectedError != "" {
+			assert.EqualError(t, err, test.expectedError, test.description)
 		} else {
-			assert.Nil(t, err, test.description)
+			assert.NoError(t, err, test.description)
```
SyntaxNode: Thank you to @guscarreon for pointing this one out in an earlier PR.

openrtb_ext/bidders.go
```diff
@@ -21,11 +21,11 @@ type BidderName string
 // Please keep these (and the BidderMap) alphabetized to minimize merge conflicts among adapter submissions.
 const (
 	Bidder33Across         BidderName = "33across"
```
SyntaxNode: Same values, just alphabetized.

endpoints/setuid.go
hhhjort: You could have the value of the map be the `FamilyName`, so that down below you will have it handy for metrics.

endpoints/setuid.go
```diff
@@ -44,64 +51,64 @@ func NewSetUIDEndpoint(cfg config.HostCookie, perms gdpr.Permissions, pbsanalyti
...
+			pc.Unsync(familyName)
 		} else {
-			err = pc.TrySync(bidder, uid)
-		}
-
-		if err == nil {
-			labels := pbsmetrics.UserLabels{
-				Action: pbsmetrics.RequestActionSet,
-				Bidder: openrtb_ext.BidderName(bidder),
+			if err = pc.TrySync(familyName, uid); err != nil {
```
hhhjort: Should this be `err == nil`? ... it seems you are coding for success here.

endpoints/setuid.go
SyntaxNode: Family Name to Bidder Name is not a 1:1 mapping.

endpoints/setuid.go
SyntaxNode: Correct. I'll fix and add a test.

endpoints/setuid.go
```diff
@@ -44,64 +51,64 @@ func NewSetUIDEndpoint(cfg config.HostCookie, perms gdpr.Permissions, pbsanalyti
...
 		pc.SetCookieOnResponse(w, setSiteCookie, &cfg, cookieTTL)
 	})
 }
 
-func validateBidder(bidderName string) error {
-	if bidderName == "" {
-		return errors.New(`"bidder" query param is required`)
+func getFamilyName(query url.Values, validFamilyNameMap map[string]struct{}) (string, error) {
+	// The family name is bound to the 'bidder' query param. In most cases, these values are the same.
+	familyName := strings.ToLower(query.Get("bidder"))
```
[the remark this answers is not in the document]
SyntaxNode: I'll remove the `ToLower`.

endpoints/setuid.go
hhhjort: Nope, usersyncer doesn't have it .... there is an old confusion in the code, `FamilyName` is supposed to be the bidder name, after resolving aliases. But usersync has given it its own definition ... since 90% of the time it is the same. We would have to refactor `Usersyncer` to be able to access this information.

endpoints/setuid.go
SyntaxNode: We'll change the metrics to be based on the cookie name in the future.

PR #1123, opened 2019-11-25
require the caller to define native assets[...].ID

Aadeshp:


SyntaxNode:
I'm worried about the impact of this change on users who might've come to depend on the behavior of Prebid Server to set these values. Thoughts on providing a warning if the ids are missing and still filling them in?

endpoints/openrtb2/auction.go
```diff
@@ -491,23 +491,35 @@ func fillAndValidateNativeAssets(assets []nativeRequests.Asset, impIndex int) er
 		return fmt.Errorf("request.imp[%d].native.request.assets must be an array containing at least one object", impIndex)
 	}
 
+	assetIDs := make(map[int64]bool, len(assets))
+	assignAssetIDs := (assets[0].ID == 0)
```
SyntaxNode: The assumption here is that `0` signifies an unset ID and if the first asset has an unset ID then Prebid Server should assign ids. The problem is an ID of 0 for the 0th item makes sense if you're following the recommendation of using array index values as IDs. Instead, I propose we set `assignAssetIDs` to `true` when every asset has an ID if 0.

Aadeshp: good point, ill update

endpoints/openrtb2/auction.go
```diff
@@ -491,23 +491,35 @@ func fillAndValidateNativeAssets(assets []nativeRequests.Asset, impIndex int) er
 		return fmt.Errorf("request.imp[%d].native.request.assets must be an array containing at least one object", impIndex)
 	}
 
+	assetIDs := make(map[int64]bool, len(assets))
```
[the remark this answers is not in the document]
Aadeshp: oh reallly? whys that, curious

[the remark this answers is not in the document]
SyntaxNode: Good question. :)

The empty struct does not use memory, unlike the boolean. Here's some more detail:
https://dave.cheney.net/2014/03/25/the-empty-struct

endpoints/openrtb2/auction.go
```diff
@@ -491,23 +491,41 @@ func fillAndValidateNativeAssets(assets []nativeRequests.Asset, impIndex int) er
...
-		// Per the OpenRTB spec docs, this is a "unique asset ID, assigned by exchange. Typically a counter for the array"
-		// To avoid conflict with the Request, we'll return a 400 if the Request _did_ define this ID,
-		// and then populate it as the spec suggests.
 		if err := validateNativeAsset(assets[i], impIndex, i); err != nil {
 			return err
 		}
-		assets[i].ID = int64(i)
+
+		if assignAssetIDs {
+			if assets[i].ID != 0 {
```
SyntaxNode: This situation (`assets[i].ID != 0`) is no longer possible with the new check.

Aadeshp: still is in case a partial number of assets have IDs defined, should be all or nothing

endpoints/openrtb2/auction.go
```diff
@@ -491,23 +491,41 @@ func fillAndValidateNativeAssets(assets []nativeRequests.Asset, impIndex int) er
 		return fmt.Errorf("request.imp[%d].native.request.assets must be an array containing at least one object", impIndex)
 	}
 
+	assetIDs := make(map[int64]struct{}, len(assets))
+
+	// If none of the asset IDs are defined by the caller, then prebid server should assign its own unique IDs. But
+	// if the caller did assign its own asset IDs, then prebid server will respect those IDs
+	assignAssetIDs := true
+	for i := 0; i < len(assets); i++ {
+		assignAssetIDs = assignAssetIDs && (assets[0].ID == 0)
```
mansinahar: I believe this was meant to be:
```
assignAssetIDs = assignAssetIDs && (assets[i].ID == 0)
```
Right?

endpoints/openrtb2/auction.go
```diff
@@ -491,23 +491,41 @@ func fillAndValidateNativeAssets(assets []nativeRequests.Asset, impIndex int) er
...
-		// Per the OpenRTB spec docs, this is a "unique asset ID, assigned by exchange. Typically a counter for the array"
-		// To avoid conflict with the Request, we'll return a 400 if the Request _did_ define this ID,
-		// and then populate it as the spec suggests.
 		if err := validateNativeAsset(assets[i], impIndex, i); err != nil {
 			return err
 		}
-		assets[i].ID = int64(i)
+
+		if assignAssetIDs {
+			assets[i].ID = int64(i)
```
mansinahar: From the code it seems like we expect that asset IDs are either specified or not specified for all assets i.e it can't be that they are specified for some and not for others, right? 

I am not a 100% sure if we can make that assumption or not so just wanted to call it out and make sure that indeed is the case.

endpoints/openrtb2/auction.go
```diff
@@ -491,23 +491,41 @@ func fillAndValidateNativeAssets(assets []nativeRequests.Asset, impIndex int) er
...
+			continue
+		}
+
+		// Each asset should have a unique ID thats assigned by the caller
+		if assets[i].ID == 0 {
```
mansinahar: In case of us assigning the asset IDs, for the first asset we will always end up assigning the asset ID `0` as the value of `i` is going to be 0. We need to change: 
```
assets[i].ID = int64(i)
```
to

```
assets[i].ID = int64(i + 1)
```

So that we start assigning asset IDs from 1 and not 0.

endpoints/openrtb2/auction.go
Aadeshp: removed

endpoints/openrtb2/auction.go
```diff
@@ -491,23 +491,41 @@ func fillAndValidateNativeAssets(assets []nativeRequests.Asset, impIndex int) er
...
+			return fmt.Errorf("request.imp[%d].native.request.assets[%d].id must not be defined. Prebid Server will set this automatically, using the index of the asset in the array as the ID", impIndex, i)
```
mansinahar: The only time when assetID will could be 0 (based on the above assumption) is that the request has asset IDs specified but with one or more of the asset IDs set as 0. In that case the error message here should be more on the lines of:
```
Asset ID can't be 0.
```

endpoints/openrtb2/sample-requests/valid-native/asset-with-no-id.json
```diff
@@ -0,0 +1,11 @@
+{
```
mansinahar: Can we please add some more test cases like for example, one with multiple assets with unique asset IDs, one with multiple assets but with non-unique asset IDs, one with multiple assets with unique asset IDs but one of the assets having asset ID `0`.

endpoints/openrtb2/auction.go
mansinahar: I just realized reading one of @SyntaxNode 's comment above that if the asset IDs are not specified we want to assign them as array indices and in that case `0` makes sense. Based on that we can leave this as is but I guess we can get rid of the next `if` check on L514, no? As an asset ID at that point of execution will be `0` only in two cases:
1. All asset IDs were `0` and we assigned the asset ID `0` for the first asset in which case it is valid
2. All asset IDs weren't `0`, meaning asset IDs were specified in the request itself and one of the assets have ID `0`. This should also be valid.

endpoints/openrtb2/auction.go
Aadeshp: derp 🤦‍♂  nice catch

endpoints/openrtb2/auction.go
Aadeshp: thats indeed the case

endpoints/openrtb2/auction.go
Aadeshp: we still need L514 imo, for partially defined asset IDs. L514 also shouldnt affect L509 since L509 is followed by a cont op

Aadeshp: e.g. if asset[0] has an ID but asset[1] doesnt

endpoints/openrtb2/auction.go
Aadeshp: thats not necessarily true, zero also means not defined

endpoints/openrtb2/auction.go
hhhjort: You can technically break out of the loop as soon as assignAssetIDs is false. Not a major concern though.

endpoints/openrtb2/auction.go
hhhjort: Actually Mansi has a point. One asset with asset ID of 0 is valid. Multiple zeros are bad, but would be caught in the uniqueness check you have.

Aadeshp: well the issue is really that you can't tell if zero is defined by the caller or if its undefined, asset.ID isn't a pointer so it could be zero cuz of the caller or cuz its left out of the object

Aadeshp: I guess if we allow zero, its hard to figure out the appropriate error message

Aadeshp: I see what y'all are saying tho, its counter-intuitive to assign zero asset ID and then on the flip side not allow zero asset ID from the caller

Aadeshp: I'll update this by removing this check, so if multiple assets don't have an ID defined _or_ if multiple assets have zero as ID then the hashset check will catch it. I'll update the error message to appropriately explain that it could be either of two issues

hhhjort: Sounds good to me.

SyntaxNode:
Travis is reporting test failures. :(

endpoints/openrtb2/auction.go
mansinahar: @Aadeshp Yeah you're right. I missed the `continue` statement. My bad. Even with that, as Hans mentioned one asset ID of `0` is still valid and I see you've already updated the PR for that. Thank you!

endpoints/openrtb2/sample-requests/invalid-native/assets-with-partial-ids.json
```diff
@@ -0,0 +1,17 @@
+{
+    "context": 1,
+    "plcmttype": 1,
+    "assets": [
+        {
+            "id": 1,
+            "img": {
```
mansinahar: I am wondering.. is this test case actually for an _invalid_ native request? Based on our assumption that there's gonna be asset IDs either specified or not specified for _all_ assets and that we don't expect partial IDs to be specified, this would actually be a _valid-native_ test case, no? 

I guess it gets a little tricky when we consider assets with partial IDs set. We will treat them as if all asset IDs are set and so the assets that don't have asset IDs set will have zero values and we will therefore consider them not being unique asset IDs and return an error which is a fair behavior. But, if only one of the assets doesn't have an ID set, like in this case, then we won't return an error because one asset with a `0` asset ID is still valid. Therefore, this example shouldn't really return an error, correct?

mansinahar: And looks like that is actually one of the tests that's failing:

```
--- FAIL: TestBadNativeRequests (0.38s)
    auction_test.go:175: Expected a 400 response from sample-requests/invalid-native/assets-with-partial-ids.json. Got 200: {"id":"req-id","bidid":"test bid id","nbr":0}
```

Issue #1107, opened 2019-11-06
Prebid Server and CCPA - phase 1


bretg:
The proposal for Prebid.js support for CCPA has been posted at https://github.com/prebid/Prebid.js/issues/4425

The interface to Prebid Server is straightforward: a single value is placed on regs.exp.us_privacy
```
{  
    "regs": {
        "ext": {
            "us_privacy": "1YN"
        }
    }
}
```

We propose this two-phase support plan for Prebid Server:

- Phase 1 - Simple pass-through: Prebid Server just passes regs.exp.us_privacy through to each bidder adapter, each of which is responsible for passing the signal through to their endpoints. The value may also be passed through cookie-sync/setuid calls. No validation is done on the value. This phase will be done in time for the early January deadline. It's assumed that each SSP/DSP will be implementing CCPA enforcement according to their interpretation of the guidelines.
- Phase 2 - Enforcement support: similar to the way that Prebid Server allows host companies to turn on GDPR enforcement, we will design and build CCPA/US-Privacy enforcement capability. This phase would not be done until the end of 1Q2020.

## Phase 1 Changes

1) Passing regs.exp.us_privacy through the system untouched likely requires no work other than testing.

2) AMP requests can pass a `us_privacy` parameter through the GET URI, which is added to the OpenRTB as regs.ext.us_privacy

3) The /cookie_sync endpoint will accept an additional `us_privacy` parameter
```
{
    "bidders": ["appnexus", "rubicon"],
    "gdpr": 1,
    "gdpr_consent": "BONV8oqONXwgmADACHENAO7pqzAAppY",
    "us_privacy": "1YN",
    "limit": 2
}
```
4) Every bidder's sync URL will be modified to pass the URL-encoded privacy value through to the sync endpoints. e.g.

```
usersync:
   url: https://pixel.rubiconproject.com/exchange/sync.php?p=prebid&gdpr={{gdpr}}&gdpr_consent={{gdpr_consent}}&us_privacy={{us_privacy}}

usersync:
      url: //ib.adnxs.com/getuid?
      redirect-url: /setuid?bidder=adnxs&gdpr={{gdpr}}&gdpr_consent={{gdpr_consent}}&us_privacy={{us_privacy}}&uid=$UID
```
5) The {{us_privacy}} macro will get resolved to the value passed through /cookie_sync or empty.

## Phase 2 Design

A separate issue will be opened to propose and discuss the more difficult enforcement feature.

hhhjort:
[...]

PR #1125, opened 2019-11-27
CCPA Phase 1: AMP Endpoint

SyntaxNode:
Implements: prebid/prebid-server#1107

Parses the CCPA 'us_privacy' signal from the AMP request URL and inserts it into the OpenRTB request.

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -382,16 +382,14 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
 		req.Imp[0].TagID = slot
 	}
 
-	gdprConsent, gdprOK := getQueryParam(httpRequest, "gdpr_consent")
-	ccpaValue, ccpaOK := getQueryParam(httpRequest, "us_privacy")
+	gdprConsent := getQueryParam(httpRequest, "gdpr_consent")
+	ccpaValue := getQueryParam(httpRequest, "us_privacy")
 	privacyPolicies := privacy.Policies{
 		GDPR: gdpr.Policy{
-			Enabled: gdprOK && len(gdprConsent) > 0,
 			Consent: gdprConsent,
 		},
 		CCPA: ccpa.Policy{
-			Enabled: ccpaOK,
-			Value:   ccpaValue,
+			Value: ccpaValue,
 		},
 	}
 
@@ -406,15 +404,15 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
 	return nil
 }
 
-func getQueryParam(httpRequest *http.Request, name string) (string, bool) {
+func getQueryParam(httpRequest *http.Request, name string) string {
 	values, ok := httpRequest.URL.Query()[name]
 
 	if !ok || len(values) == 0 {
-		return "", false
+		return ""
 	}
 
 	// return first value of the query param, matching the behavior of httpRequest.FormValue
-	return values[0], true
+	return values[0]
 }
 
 func makeFormatReplacement(overrideWidth uint64, overrideHeight uint64, width uint64, height uint64, multisize string) []openrtb.Format {
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -22,6 +22,9 @@ import (
 	"github.com/prebid/prebid-server/exchange"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
+	"github.com/prebid/prebid-server/privacy"
+	"github.com/prebid/prebid-server/privacy/ccpa"
+	"github.com/prebid/prebid-server/privacy/gdpr"
 	"github.com/prebid/prebid-server/stored_requests"
 	"github.com/prebid/prebid-server/stored_requests/backends/empty_fetcher"
 	"github.com/prebid/prebid-server/usersync"
@@ -379,22 +382,21 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
 		req.Imp[0].TagID = slot
 	}
 
-	//In the AMP endpoint the consent string found in the http.Request query overrides that of the prebid query
-	queryConsentString := httpRequest.FormValue("gdpr_consent")
-	if queryConsentString != "" {
-		jsonMsg := json.RawMessage(`{"consent":"` + queryConsentString + `"}`)
-		// If nil, initialize
-		if req.User == nil {
-			req.User = &openrtb.User{Ext: jsonMsg}
-		} else if req.User.Ext == nil {
-			req.User.Ext = jsonMsg
-		} else { // req.User.Ext != nil, keep whatever is in there and only substitute the consent string
-			var parserErr error
-			req.User.Ext, parserErr = jsonparser.Set(req.User.Ext, []byte(`"`+queryConsentString+`"`), "consent")
-			if parserErr != nil {
-				return parserErr
-			}
-		}
+	gdprConsent, gdprOK := getQueryParam(httpRequest, "gdpr_consent")
+	ccpaValue, ccpaOK := getQueryParam(httpRequest, "us_privacy")
+	privacyPolicies := privacy.Policies{
+		GDPR: gdpr.Policy{
+			Enabled: gdprOK && len(gdprConsent) > 0,
+			Consent: gdprConsent,
+		},
+		CCPA: ccpa.Policy{
+			Enabled: ccpaOK,
+			Value:   ccpaValue,
+		},
+	}
+
+	if err := privacyPolicies.Write(req); err != nil {
+		return err
 	}
 
 	if timeout, err := strconv.ParseInt(httpRequest.FormValue("timeout"), 10, 64); err == nil {
@@ -402,6 +404,17 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
 	}
 
 	return nil
+}
+
+func getQueryParam(httpRequest *http.Request, name string) (string, bool) {
+	values, ok := httpRequest.URL.Query()[name]
+
+	if !ok || len(values) == 0 {
+		return "", false
+	}
+
+	// return first value of the query param, matching the behavior of httpRequest.FormValue
+	return values[0], true
 }
 
 func makeFormatReplacement(overrideWidth uint64, overrideHeight uint64, width uint64, height uint64, multisize string) []openrtb.Format {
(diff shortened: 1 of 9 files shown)


endpoints/openrtb2/amp_auction.go
```diff
@@ -379,22 +382,21 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
 		req.Imp[0].TagID = slot
 	}
 
-	//In the AMP endpoint the consent string found in the http.Request query overrides that of the prebid query
```
SyntaxNode: The GDPR implementation is scattered throughout the code. I'm beginning to consolidate the logic to a privacy package. I did a full refactor locally, but it was way too much change... so I'll be phasing it in.

endpoints/openrtb2/amp_auction.go
```diff
@@ -404,6 +406,17 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
 	return nil
 }
 
+func getQueryParam(httpRequest *http.Request, name string) (string, bool) {
```
SyntaxNode: Switched from using `httprequest.FormValue` since that also looks in the request body and can't tell the difference between a param being omitted vs set to an empty string.

endpoints/openrtb2/amp_auction.go
```diff
@@ -379,22 +382,21 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
...
-			req.User.Ext, parserErr = jsonparser.Set(req.User.Ext, []byte(`"`+queryConsentString+`"`), "consent")
-			if parserErr != nil {
-				return parserErr
-			}
-		}
+	gdprConsent, gdprOK := getQueryParam(httpRequest, "gdpr_consent")
+	ccpaValue, ccpaOK := getQueryParam(httpRequest, "us_privacy")
+	privacyPolicies := privacy.Policies{
+		GDPR: gdpr.Policy{
+			Enabled: gdprOK && len(gdprConsent) > 0,
```
SyntaxNode: For compatibility with existing code, an empty GDPR consent value is interpreted as not enabled. For CCPA, I'd like to be less opinionated and pass through exactly the string provided, even if it's empty. Thoughts?

privacy/ccpa/ccpa.go
```diff
@@ -0,0 +1,34 @@
...
+	if p.Enabled == false {
+		return nil
+	}
+
+	if req.Regs == nil {
+		req.Regs = &openrtb.Regs{}
+	}
+
+	if req.Regs.Ext == nil {
+		req.Regs.Ext = json.RawMessage(`{"us_privacy":"` + p.Value + `"}`)
```
SyntaxNode: It's a bit strange how the CCPA regulation maps to the `us_privacy` symbol. Should we call it USPrivacy in all of the models or keep the CCPA terminology?

privacy/gdpr/gdpr.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"encoding/json"
+
+	"github.com/buger/jsonparser"
+	"github.com/mxmCherry/openrtb"
+)
+
+// Policy represents the GDPR regulation of an OpenRTB bid request.
+type Policy struct {
+	Enabled bool
```
SyntaxNode: Is 'Enabled' the right term here? .. or perhaps 'Present' would be better?

guscarreon: In order to make it even more obvious that it is a boolean value, why not `isPresent` or `isDefined`?

SyntaxNode: I like `IsPresent`. I've been back and forth on if it's really needed. I'm heaving leaning towards including it so the information is captured and made available for other parts of the code to consume.

SyntaxNode: Eh. I think I'll just use pointers for now. That's my best guess on how this sort of thing is usually done with Go, since these values are a 1:1 mapping to JSON structures.

privacy/policies.go
```diff
@@ -0,0 +1,27 @@
...
+// Policies represents the privacy policies and regulations of an OpenRTB bid request.
+type Policies struct {
+	GDPR gdpr.Policy
+	CCPA ccpa.Policy
+}
+
+// Write mutates an OpenRTB bid request with the context of the policies.
+func (p Policies) Write(req *openrtb.BidRequest) error {
+	if err := p.GDPR.Write(req); err != nil {
+		return err
```
guscarreon: Is it necessary to interrupt execution even if only one of them returns an error? After all, they fill two different `Ext` fields.

privacy/policies.go
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	if err := p.CCPA.Write(req); err != nil {
+		return err
+	}
+
+	return nil
+}
```
guscarreon: I realize both `privacy/ccpa/ccpa.go` and `privacy/gdpr/gdpr.go` have 100% coverage with their respective "_test.go" files but do you think a `privacy/policies_test.go` is also desired?

hhhjort:
CCPA is not being implemented on the legacy endpoint. That endpoint is deprecated, and doesn't define extension fileds to put new values in. So this would be an update to PBS and all clients of the legacy endpoint. Any clients using the legacy endpoint are doing so because they have not been touched in years, so are unlikely to be updated to send the CCPA signal even if we updated the endpoint to accept it.

endpoints/openrtb2/amp_auction.go
hhhjort: AMP requests are always GET, so body should be empty anyway. Also not set and turned off mean the same thing, so it is a distinction without a difference.

endpoints/openrtb2/amp_auction.go
hhhjort: I thought GDPR being turned on, but an empty consent string meant that GDPR applied with zero consents given.

hhhjort: Oh, yes, we are only sending one GDPR signal in the AMP query, never mind

privacy/ccpa/ccpa.go
hhhjort: I think the idea is that several states are enacting similar laws, and we will just filter based on the most strict version of the law as "us_privacy" signal.

privacy/ccpa/ccpa.go
```diff
@@ -0,0 +1,34 @@
...
+		return nil
+	}
+
```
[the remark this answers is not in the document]
hhhjort: Do we even need a string pointer, as not set and empty mean the same thing for the signal?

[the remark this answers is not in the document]
SyntaxNode: I've been going back and forth a lot on this. If we get an empty string in the AMP params then adding an empty string in the OpenRTB request makes sense. If we receive no query param at all in the AMP params then does it still make sense to add an empty string to the OpenRTB request? Or perhaps just treat nil and empty as the same and only add non-empty values in the OpenRTB request.

In GDPR land, the signal is an int but we have it parsed as an int pointer because nil and empty (0) mean different things.

privacy/policies.go
SyntaxNode: I don't think we should continue with the AMP auction if we can't construct a proper bid request. In reality, this should never occur.

endpoints/openrtb2/amp_auction.go
SyntaxNode: Distinction removed.

endpoints/openrtb2/amp_auction.go
SyntaxNode: Should look better now with nil having the same meaning as empty.

privacy/ccpa/ccpa.go
SyntaxNode: Cool. Kept CCPA in code for now.

privacy/gdpr/gdpr.go
SyntaxNode: Nope. That didn't work out nicely.

privacy/ccpa/ccpa.go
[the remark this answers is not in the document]
SyntaxNode: String pointer removed.

privacy/policies.go
SyntaxNode: Refactored and added tests.

privacy/ccpa/ccpa_test.go
```diff
@@ -0,0 +1,74 @@
...
+			expected: &openrtb.BidRequest{Regs: &openrtb.Regs{
+				Ext: json.RawMessage(`{"us_privacy":"anyValue"}`)}},
+		},
+		{
+			description: "Enabled With Existing Request Regs Ext Object - Doesn't Overwrite",
+			policy:      Policy{Value: "anyValue"},
+			request: &openrtb.BidRequest{Regs: &openrtb.Regs{
+				Ext: json.RawMessage(`{"existing":"any"}`)}},
+			expected: &openrtb.BidRequest{Regs: &openrtb.Regs{
+				Ext: json.RawMessage(`{"existing":"any","us_privacy":"anyValue"}`)}},
```
guscarreon: I've been looking in the codebase for the recipient field of the `"existing"` value and found nothing. Was it meant to be a filler value in the `json.RawMessage` depicted in the tests? What exactly is its purpose?

privacy/policies_test.go
```diff
@@ -1,60 +1,77 @@
...
-	}
-	expectedRequest := &openrtb.BidRequest{
-		Regs: &openrtb.Regs{
-			Ext: json.RawMessage(`{"us_privacy":"anyValue"}`)},
-		User: &openrtb.User{
-			Ext: json.RawMessage(`{"consent":"anyConsent"}`)},
-	}
-
+func TestWritePolicieNone(t *testing.T) {
 	request := &openrtb.BidRequest{}
```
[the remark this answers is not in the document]
SyntaxNode: Ha! Ooops.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -739,6 +739,102 @@ func TestWidthOnly(t *testing.T) {
...
+		return
+	}
+
+	// Assert CCPA Signal Not Found
+	if exchange.lastRequest.Regs != nil && exchange.lastRequest.Regs.Ext != nil {
+		var regs openrtb_ext.ExtRegs
+		err = json.Unmarshal(exchange.lastRequest.Regs.Ext, &regs)
+		assert.NoError(t, err)
+		assert.Empty(t, regs.USPrivacy)
+	}
```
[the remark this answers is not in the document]
SyntaxNode: I don't think so. I'm testing that the signal isn't present, so if there is no `regs` or no `regs.ext` or an empty `regs.ext.us_privacy` then I think the test should pass.

privacy/ccpa/ccpa_test.go
SyntaxNode: It represents an existing json value to test the overwriting (or lack thereof) behavior.

privacy/policies_test.go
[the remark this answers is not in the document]
SyntaxNode: Fixed everywhere to TestWritePoliciesNone since its testing the private writePolicies func.

PR #1126, opened 2019-11-27
[currency converter] allow to deduce reverse rate

benjaminch:
This CL allows the currency rate currency to deduce a currency rate even
if not directly defined in the table but the reverse rate is present.

E.q.

USD => EUR is 1.0897
EUR => USD is not set

Old behavior when asking rate from EUR to USD will not be found,
New behavior is using the known reverse rate to deduce the rate.

Rate for 2 USD will be 2 * (1 / 1.0897)

### DIFF
--- a/currencies/rates_test.go
+++ b/currencies/rates_test.go
@@ -163,6 +163,60 @@ func TestGetRate(t *testing.T) {
 	}
 }
 
+func TestGetRate_ReverseConversion(t *testing.T) {
+
+	// Setup:
+	rates := currencies.NewRates(time.Now(), map[string]map[string]float64{
+		"USD": {
+			"GBP": 0.77208,
+		},
+		"EUR": {
+			"USD": 0.88723,
+		},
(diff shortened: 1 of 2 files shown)


currencies/rates_test.go
```diff
@@ -163,6 +163,39 @@ func TestGetRate(t *testing.T) {
...
+			"USD": 0.88723,
+		},
+	})
+
+	testCases := []struct {
+		from         string
+		to           string
+		expectedRate float64
+	}{
+		{from: "USD", to: "GBP", expectedRate: 0.77208},
```
SyntaxNode: Makes sense to me and looks great. Could you please add a description to the test and pass it to the asserts so if a test case fails, we can know which one specifically and to explicitly call out which conversions are direct vs "reverse engineered".

benjaminch: Hey ! Yes sure, just added, thanks :)

Issue #1130, opened 2019-12-04
0.90.0 issue with hb_cache_host & hb_cache_path with universal creative


Austinb:
There is an issue with how the responses are returned from PBS regarding the values set for `hb_cache_host` and `hb_cache_path` and using the universal creative (UC.js).  In the PBS side the leading slash from `hb_cache_path` (https://github.com/prebid/prebid-server/blob/master/prebid_cache_client/client.go#L77) is now removed since 0.90.0 which causes issues on the UC.js (delivery) side of things.  Using the PBS settings of 
```yaml
external_cache:
  host: myhost.com
  path: /c/v1/cache
```

Before 0.90.0 the cache values in the `bid[i].ext.prebid` response from PBS would be something like 
```json
{
  "targeting": {
    "hb_bidder": "appnexus",
    "hb_bidder_appnexus": "appnexus",
    "hb_cache_host": "myhost.com",
    "hb_cache_host_appnex": "myhost.com",
    "hb_cache_id": "xxxxx",
    "hb_cache_id_appnexus": "xxxxxx",
    "hb_cache_path": "/c/v1/cache",
    "hb_cache_path_appnex": "/c/v1/cache",
    "hb_env": "mobile-app",
    "hb_env_appnexus": "mobile-app",
    "hb_pb": "0.00",
    "hb_pb_appnexus": "0.00",
    "hb_size": "320x100",
    "hb_size_appnexus": "320x100"
  }
}
```

Now the response comes back as
```json
{
  "targeting": {
    "hb_bidder": "appnexus",
    "hb_bidder_appnexus": "appnexus",
    "hb_cache_host": "myhost.com",
    "hb_cache_host_appnex": "myhost.com",
    "hb_cache_id": "xxxxx",
    "hb_cache_id_appnexus": "xxxxxx",
    "hb_cache_path": "c/v1/cache",
    "hb_cache_path_appnex": "c/v1/cache",
    "hb_env": "mobile-app",
    "hb_env_appnexus": "mobile-app",
    "hb_pb": "0.00",
    "hb_pb_appnexus": "0.00",
    "hb_size": "320x100",
    "hb_size_appnexus": "320x100"
  }
}
```

[...]

PR #1131, opened 2019-12-04
Revert changes to GetExtCacheData() made in #1104 (#1130)

Austinb:
As described in #1130 the changes removing the left most `/` in `external_cache.path` breaks the cache response send by PBS when delivering a line item using the universal creative js (https://github.com/prebid/prebid-universal-creative).  There is no documentation on where the slash should be placed for the `external_cache`.  Should it be
```yaml
external_cache:
  host: myhost.com
  path: /c/v1/cache
```
or 
```yaml
external_cache:
  host: myhost.com/
  path: c/v1/cache
```
or neither and the universal creative needs to be updated?

The one test that exists has no leading slash on `external_cache.path` or trailing slash on `external_cache.host` so I am unsure where it is supposed to go. See https://github.com/prebid/prebid-server/blob/master/exchange/exchange_test.go#L1293.  

Or does the current code need to be changed to add in `/` where missing to make a valid url as it relates to the universal creative at https://github.com/prebid/prebid-universal-creative/blob/master/src/nativeAssetManager.js#L66?


Austinb:
Tests should be updated as well to enforce the location of a leading/trailing slash (if any).  Will wait until a maintainer weighs in on what the expected behavior should be.

prebid_cache_client/client.go
```diff
@@ -70,14 +70,7 @@ type clientImpl struct {
 }
 
 func (c *clientImpl) GetExtCacheData() (string, string) {
-	path := c.externalCachePath
-	if path == "/" {
```
SyntaxNode: I think it makes sense to leave in the `path == "/"` check as returning an empty string instead of empty path should be good and seems preferred as a way to communicate "there is no path". Then we should revert the Index check such as if the index has length >0 and doesn't start with a "/" to prepend it so we enforce it always being present to avoid breaking UC.js.

PR #1132, opened 2019-12-04
Better native request validation

hhhjort:
We were requiring some fields that are only "recommended", and we have seen examples in the wild where they were not filled. This relaxes those requirements.


endpoints/openrtb2/auction.go
```diff
@@ -444,6 +444,10 @@ func fillAndValidateNative(n *openrtb.Native, impIndex int) error {
 }
 
 func validateNativeContextTypes(cType native.ContextType, cSubtype native.ContextSubType, impIndex int) error {
+	if cType == 0 {
```
SyntaxNode: Can we please add a unit test or json test for this case?

endpoints/openrtb2/auction.go
```diff
@@ -480,6 +484,10 @@ func validateNativeContextTypes(cType native.ContextType, cSubtype native.Contex
 }
 
 func validateNativePlacementType(pt native.PlacementType, impIndex int) error {
+	if pt == 0 {
```
SyntaxNode: Can we please add a unit test or json test for this case?

endpoints/openrtb2/auction.go
hhhjort: endpoints/openrtb2/sample-requests/valid-native/request-no-context.json

endpoints/openrtb2/auction.go
hhhjort: See the other JSON tests created/moved to reflect the new rules.

endpoints/openrtb2/auction.go
```diff
@@ -444,6 +444,10 @@ func fillAndValidateNative(n *openrtb.Native, impIndex int) error {
 }
 
 func validateNativeContextTypes(cType native.ContextType, cSubtype native.ContextSubType, impIndex int) error {
+	if cType == 0 {
+		// Context is only recommended, so none is a valid type.
+		return nil
```
guscarreon: Should we return in here without also checking `cSubtype` first?

endpoints/openrtb2/auction_test.go
```diff
@@ -282,7 +282,6 @@ func TestRejectAccountRequired(t *testing.T) {
 func (gr *getResponseFromDirectory) assert(t *testing.T) {
 	//t *testing.T, dir string, payloadGetter func(*testing.T, []byte) []byte, messageGetter func(*testing.T, []byte) []byte, expectedCode int, aliased bool) {
 	t.Helper()
-	var filename string
 	var filesToAssert []string
```
guscarreon: Great catch! `filename` was doing nothing here

endpoints/openrtb2/auction.go
SyntaxNode: Ah. Changing the directory changes the validation criteria. Missed that.

endpoints/openrtb2/auction.go
SyntaxNode: Ah. Changing the directory changes the validation criteria. Missed that.

endpoints/openrtb2/auction.go
hhhjort: If there is no context type, would subtype have any meaning?

guscarreon: I guess not. LOL

SyntaxNode: The `contextsubtype` is optional but does have defined ranges same as cType. Not sure if we want to add it in the future, but I would think that's a separate story.

PR #1133, opened 2019-12-04
Update http refresh to use url builder. Fixes #1065

### LINKED ISSUE #1065: Issue with HTTP event-based update using amp_endpoint
The url being created for the amp HTTP request when doing event-based updates is incorrect.  The code at https://github.com/prebid/prebid-server/blob/master/stored_requests/events/http/http.go#L97 just blindly appends the `?last-modified=` query string to the end of the url that is passed.  This works fine for the normal endpoint but creates an invalid url when using the default example for pulling in amp updates found at https://github.com/prebid/prebid-server/blob/master/docs/developers/stored-requests.md#caches-and-event-based-updating. 

The normal updates url used for the event based updating is `https://server/path/?last-modified=2019-10-10T01:36:16Z` which is fine. However the amp update url used ends up being `https://server/path/?amp=true?last-modified=2019-10-10T01:36:17Z` which is an invalid url. 

I am new to GO so I am unsure the best way to check for existing query params in a string and appending properly so I have not attempted to make a PR for this issue.

Austinb:
Updates to master to fix issue #1065 where the update event is blindly appending the `last-modified` query param to the `stored_requests.http_events` endpoints.  Added error handling for failed parse and an info log event when the http refresh event is called.

### DIFF
--- a/stored_requests/events/http/http.go
+++ b/stored_requests/events/http/http.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	httpCore "net/http"
+	url2 "net/url"
 	"time"
 
 	"golang.org/x/net/context/ctxhttp"
@@ -94,7 +95,27 @@ func (e *HTTPEvents) refresh(ticker <-chan time.Time) {
 		select {
 		case thisTime := <-ticker:
 			thisTimeInUTC := thisTime.UTC()
-			thisEndpoint := e.Endpoint + "?last-modified=" + e.lastUpdate.Format(time.RFC3339)
+			
+			// Parse the endpoint url defined
+			url, urlErr := url2.Parse(e.Endpoint)
+	
+			// Error with url parsing
+			if urlErr != nil {
+   				// Should do something here probably...
+			}
+
+			// Parse the url query string
+			urlQuery := url.Query()
+
+			// See the last-modified query param
+			urlQuery.Set("last-modified", e.lastUpdate.Format(time.RFC3339))
+
+			// Rebuild
+			url.RawQuery = urlQuery.Encode()
+
+			// Convert to string and save
+			thisEndpoint := url.String()
+			
 			ctx, cancel := e.ctxProducer()
 			resp, err := ctxhttp.Get(ctx, e.client, thisEndpoint)
 			if respObj, ok := e.parse(thisEndpoint, resp, err); ok {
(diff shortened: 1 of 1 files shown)


stored_requests/events/http/http.go
```diff
@@ -94,10 +95,33 @@ func (e *HTTPEvents) refresh(ticker <-chan time.Time) {
 		select {
 		case thisTime := <-ticker:
 			thisTimeInUTC := thisTime.UTC()
-			thisEndpoint := e.Endpoint + "?last-modified=" + e.lastUpdate.Format(time.RFC3339)
+
+			// Parse the endpoint url defined
+			url, urlErr := url2.Parse(e.Endpoint)
```
SyntaxNode: Instead of aliasing the package, could you please use a different name for the variable?

stored_requests/events/http/http.go
```diff
@@ -94,10 +95,33 @@ func (e *HTTPEvents) refresh(ticker <-chan time.Time) {
 		select {
 		case thisTime := <-ticker:
 			thisTimeInUTC := thisTime.UTC()
-			thisEndpoint := e.Endpoint + "?last-modified=" + e.lastUpdate.Format(time.RFC3339)
+
+			// Parse the endpoint url defined
```
SyntaxNode: Consider extracting the new code to a method, something like:

```
url.RawQuery = buildUpdateURL(e.Endpoint, e.lastUpdate);
```

stored_requests/events/http/http.go
Austinb: IDE imported it with an alias. Removed.

stored_requests/events/http/http.go
Austinb: If a new stand alone function is required it might as well be more generic then that and pass a structure with the query params to update.  This function should be moved into some kind of utils file so that other parts of the project can possibly utilize it in the future.  Currently that is 
beyond my knowledge in GoLang at this point or available time to learn. 

I would imagine something like
```golang
endpointUrl.RawQuery = updateURLQuery(url, struct_of_query_params)
```

SyntaxNode: Gotcha. Mind if we refactor this later on?

Austinb: If you are asking me that is completely up to you guys.  If you want to change the code now or later to make it better and more testable that is fine.  I just wanted to get it working so the PBS would stop making invalid url calls when trying to do amp updates.

SyntaxNode: I'd like to get the fix in ASAP as well. Wanted to first make sure you won't be offended if you see us refactor it a bit later on.

Austinb: Not a problem. I am not at all good at GoLang so I am sure it can be done better so I expect it will be changed or even removed at a later time.

PR #1134, opened 2019-12-05
Add Pubnative Adapter

fhartwig:


adapters/pubnative/pubnative.go
```diff
@@ -0,0 +1,170 @@
...
+			continue
+		}
+
+		err := convertImpression(&imp)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
+		reqJSON, err := json.Marshal(request)
```
hhhjort: Here you are looping over the imps, modifying one `Imp` in each loop, and then generating a request for each `Imp` that includes all the `Imp`s, with the current and prior `Imp`s modified, and the remaining `Imp`s untouched. You probably want to make a shallow copy of `request`, and assign its `Imp[]` as a one element slice containing the current modified `Imp` that you are looping over.

fhartwig: You're right, thanks for spotting this 👍

fhartwig:
> Seems to be a bit of confusion on handling `Imp`s. It looks like you want to make one request per `Imp`, but you are sending all `Imp`s in each request. This should be fixed unless for some reason this is intentional.

It wasn't intentional, we just didn't test this with multi-imp requests. I've fixed this now.

adapters/pubnative/pubnative.go
```diff
@@ -0,0 +1,172 @@
...
+		requestCopy.Imp = []openrtb.Imp{imp}
+		reqJSON, err := json.Marshal(&requestCopy)
+		if err != nil {
+			errs = append(errs, err)
+			return nil, errs
+		}
+
+		reqData := &adapters.RequestData{
+			Method:  "POST",
+			Uri:     fmt.Sprintf("%s?apptoken=%s&zoneid=%d", a.URI, pubnativeExt.AppAuthToken, pubnativeExt.ZoneID),
```
SyntaxNode: Consider adding a query param encoder on the AppAuthToken string for safety. The zone should be fine since it's an int.

fhartwig: Good point, I changed it to use Go's URL encoding to build the query string.

SyntaxNode:
Looks good to me as well. Happy to merge when the tests pass. Failing now with:
```
adapters/pubnative/pubnative.go
```

fhartwig:
@SyntaxNode Thanks, I fixed the formatting issue and the checks pass now.

PR #1142, opened 2019-12-12
null check banner.h

Aadeshp:


adapters/audienceNetwork/facebook.go
```diff
@@ -183,6 +183,12 @@ func (this *FacebookAdapter) modifyImp(out *openrtb.Imp) error {
 			return nil
 		}
 
+		if out.Banner.H == nil {
+			return &errortypes.BadInput{
+				Message: fmt.Sprintf("imp #%s: banner height required", out.ID),
+			}
+		}
+
```
guscarreon: Do you think we could make use of the information inside the `Format` field, if any, before we throw an error? Are supposed to ever look in there in case we have missing `W` and `H` fields?
```
162   func (this *FacebookAdapter) modifyImp(out *openrtb.Imp) error {
163       impType, ok := resolveImpType(out)
164   --- 12 lines: if !ok {----------------------------------------------------------------------
176       switch impType {
177       case openrtb_ext.BidTypeBanner:
178           if out.Instl == 1 {
179               out.Banner.W = openrtb.Uint64Ptr(0)
180               out.Banner.H = openrtb.Uint64Ptr(0)
181               out.Banner.Format = nil
182
183               return nil
184           }
185
186 +         if out.Banner.H == nil {
187 +             if out.Banner.Format != nil && len(out.Banner.Format) > 0 {
188 +                 //Assign this values to Banner.W and Banner.H so the supportedBannerHeights check that follows has something to work with
189 +                 out.Banner.W = openrtb.Uint64Ptr(out.Banner.Format[0].W)
190 +                 out.Banner.H = openrtb.Uint64Ptr(out.Banner.Format[0].H)
191 +             } else {
192 +                 return &errortypes.BadInput{
193 +                     Message: fmt.Sprintf("imp #%s: only banner heights 50 and 250 are supported", out.ID),
194 +                 }
195 +             }
196 +         }
197
198           if _, ok := supportedBannerHeights[*out.Banner.H]; !ok {
199               return &errortypes.BadInput{
200                   Message: fmt.Sprintf("imp #%s: only banner heights 50 and 250 are supported", out.ID),
201               }
202           }
203
204   ---  4 lines: This will get overwritten post-serialization */-------------------------------
208       }
209
210       return nil
211   }
adapters/audienceNetwork/facebook.go [+]
```

Aadeshp: i dont actually think the FAN request supports multiple sizes

PR #1138, opened 2019-12-11
CCPA Phase 2: Enforcement

SyntaxNode:
All good now. Full functionality implemented.

PR #1143, opened 2019-12-13
Adding the passing of CCPA value to the bid request for video endpoint

camrice:
Just to note, this will need a small update when PR #1138 is merged

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -24,8 +24,6 @@ import (
 	"github.com/prebid/prebid-server/exchange"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
-	"github.com/prebid/prebid-server/privacy"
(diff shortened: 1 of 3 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -500,9 +502,31 @@ func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.Bi
 		bidRequest.TMax = videoRequest.TMax
 	}
 
+	if videoRequest.Regs != nil {
```
SyntaxNode: I didn't realize you were going to add the OpenRTB Regs into the video model directly. I expected something more along the lines of the simplified user object. This is totally fine with me, but in that case you can just set the bidRequest.Regs directly to the videoRequest.Regs... provided nothing else in the OpenRTB transform is writing out Regs in which case this code is safer.

camrice: Updated the change here to setting the bidRequest.Regs directly to the regs from the videoRequest. Also updated unit test to have GDPR fields.

PR #1145, opened 2019-12-13
Add custom JSON req/resp data to the analytics logging…

jmaynardxandr:
… for the /openrtb2/video endpoint.

### DIFF
--- a/analytics/filesystem/file_module.go
+++ b/analytics/filesystem/file_module.go
@@ -14,6 +14,7 @@ type RequestType string
 const (
 	COOKIE_SYNC RequestType = "/cookie_sync"
 	AUCTION     RequestType = "/openrtb2/auction"
+	VIDEO       RequestType = "/openrtb2/video"
 	SETUID      RequestType = "/set_uid"
 	AMP         RequestType = "/openrtb2/amp"
 )
@@ -28,6 +29,15 @@ func (f *FileLogger) LogAuctionObject(ao *analytics.AuctionObject) {
 	//Code to parse the object and log in a way required
 	var b bytes.Buffer
 	b.WriteString(jsonifyAuctionObject(ao))
+	f.Logger.Debug(b.String())
+	f.Logger.Flush()
+}
+
+//Writes VideoObject to file
+func (f *FileLogger) LogVideoObject(vo *analytics.VideoObject) {
+	//Code to parse the object and log in a way required
+	var b bytes.Buffer
+	b.WriteString(jsonifyVideoObject(vo))
 	f.Logger.Debug(b.String())
 	f.Logger.Flush()
 }
@@ -95,6 +105,23 @@ func jsonifyAuctionObject(ao *analytics.AuctionObject) string {
 		return string(b)
 	} else {
 		return fmt.Sprintf("Transactional Logs Error: Auction object badly formed %v", err)
+	}
+}
+
+func jsonifyVideoObject(vo *analytics.VideoObject) string {
+	type alias analytics.VideoObject
+	b, err := json.Marshal(&struct {
+		Type RequestType `json:"type"`
+		*alias
+	}{
+		Type:  VIDEO,
+		alias: (*alias)(vo),
+	})
+
+	if err == nil {
+		return string(b)
+	} else {
+		return fmt.Sprintf("Transactional Logs Error: Video object badly formed %v", err)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


analytics/filesystem/file_module.go
```diff
@@ -32,6 +33,15 @@ func (f *FileLogger) LogAuctionObject(ao *analytics.AuctionObject) {
...
 
+//Writes VideoObject to file
+func (f *FileLogger) LogVideoObject(vo *analytics.VideoObject) {
+	//Code to parse the object and log in a way required
+	var b bytes.Buffer
+	b.WriteString(jsonifyVideoObject(vo))
+	f.Logger.Debug(b.String())
+	f.Logger.Flush()
+}
+
```
guscarreon: image
Again, newly created function seems to not be covered by the test cases in contrast with other functions in the same file

analytics/filesystem/file_module.go
```diff
@@ -98,6 +108,23 @@ func jsonifyAuctionObject(ao *analytics.AuctionObject) string {
...
+		alias: (*alias)(vo),
+	})
+
+	if err == nil {
+		return string(b)
+	} else {
+		return fmt.Sprintf("Transactional Logs Error: Video object badly formed %v", err)
+	}
+}
+
```
guscarreon: image
there seems to not be test cases for `jsonifyVideoObject` either. What if the parameter is `nil`? Are we ok with a JSON object with a `nil` field? Do we need to check for `nil`? I realized this is a local function but if we want to extend in the future, corner cases wouldn't hurt.

jmaynardxandr: Yes good point Gus, I added some calls to test the video endpoint logging and jsonify also in the new commit. Regarding checking for nil fields in jsonifyVideoObject I think it will be ok because if a field is nil it will show up in the JSON with a null value, which the person evaluating the log file should intuitively associate with a nil value in Go. Thanks for the feedback.

analytics/filesystem/file_module.go
```diff
@@ -32,6 +33,15 @@ func (f *FileLogger) LogAuctionObject(ao *analytics.AuctionObject) {
 	f.Logger.Flush()
 }
 
+//Writes VideoObject to file
+func (f *FileLogger) LogVideoObject(vo *analytics.VideoObject) {
+	//Code to parse the object and log in a way required
+	var b bytes.Buffer
+	b.WriteString(jsonifyVideoObject(vo))
+	f.Logger.Debug(b.String())
```
guscarreon: I'm probably not understanding the whole context but if `jsonifyVideoObject` finds an error, will just flush to the log with a `DEBUG` flag?
```
"2019/12/18 14:16:03 DEBUG Transactional Logs Error: Auction object badly formed JSON_MARSHALL_ERR_MESSAGE"
```
Is this what we want? Don't we want to log to the `Error` level?

analytics/filesystem/file_module_test.go
```diff
@@ -64,6 +73,7 @@ func TestFileLogger_LogObjects(t *testing.T) {
 	defer os.RemoveAll(TEST_DIR)
 	if fl, err := NewFileLogger(TEST_DIR + "//test"); err == nil {
 		fl.LogAuctionObject(&analytics.AuctionObject{})
+		fl.LogVideoObject(&analytics.VideoObject{})
 		fl.LogAmpObject(&analytics.AmpObject{})
```
guscarreon: In order to have a more comprehensive test do you think it won't be much of a hassle to compare the log file at the end of this test? Maybe the the `SameFile(fi1, fi2 FileInfo)` method from the OS package or even asserting against `bytes.Compare` could suffice. What do you think?

hhhjort: Comparing log files might be tricky, if logger is including time stamps. you will have to craft some regex's to test the output. Or make your own version of `NewFileLogger()` that creates a non-timestamped logger to generate data.

analytics/filesystem/file_module.go
jmaynardxandr: This function is intended to log the video object at the debug level, but if there is an error in that process then yes I agree it would be better to log that as an error instead. The point of this PR however is really just to add the custom JSON request and response used by the /video endpoint to the analytics logging, so I'd prefer to leave improvements to the existing testing/logging as a separate issue. (note while this specific function to log the video object is technically new code, I say "existing" here because it's basically just a cut-&-paste of how all the other objects are being logged).

SyntaxNode:
Does this PR have any unresolved feedback @guscarreon @hhhjort ?

PR #1146, opened 2019-12-15
Gamoshi Adapter: Update cookie sync

sa1omon:
## Type of change
- [X] Maintenance

## Changes description

Update default configuration for cookie sync


SyntaxNode:
A few questions:
1) This also updates the auction/bid url. That isn't mentioned in the PR description and I just want to verify it's intentional.
2) While we're here, does the Gamoshi cookie sync url accept a CCPA us_privacy value? That's available now with a `{{.USPrivacy}}` macro.

mmoses:
Hi thanks.

Referring to your questions
1. Yes, The bid url and the cookie syncing are both international.
2. Great thinking, we are planing to add it in the next version, but we can
add it now to our url, do you want us to do it or you can add it ?

Best
Moses

SyntaxNode:
> 1. Yes, The bid url and the cookie syncing are both international. 

Excellent. Thank you for confirming.

2. Great thinking, we are planing to add it in the next version, but we can add it now to our url, do you want us to do it or you can add it ?

I can add it in for you in a separate PR soon. No worries. I imagine this would just be adding `&us_privacy={{.USPrivacy}}` at the end of your sync url?

mmoses:
Yes just &us_privacy={{.USPrivacy}
Tnx
Moses

mmoses:
tnx

PR #1151, opened 2019-12-17
Add Applogy adapter

IncSW:


adapters/applogy/applogy.go
```diff
@@ -0,0 +1,157 @@
...
+	endpoint string
+}
+
+func (a *ApplogyAdapter) MakeRequests(request *openrtb.BidRequest, _ *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	impressions := request.Imp
+	totalImpressions := len(impressions)
+	result := make([]*adapters.RequestData, 0, totalImpressions)
```
guscarreon: no need to declare `totalImpressions` since `len` reads from memory rather than counting items and, in consequence, comes at a time complexity of O(1)

adapters/applogy/applogy.go
```diff
@@ -0,0 +1,157 @@
...
+					errs = append(errs, &errortypes.BadInput{
+						Message: "banner size information missing",
+					})
+					continue
+				}
+				format := banner.Format[0]
+				banner.W = &format.W
+				banner.H = &format.H
+			}
+		}
```
guscarreon: What happens if `impression.Banner == nil` but `impression.Video != nil` or `impression.Native != nil`? Will we perform no checks on any of them?

adapters/applogy/applogy.go
```diff
@@ -0,0 +1,157 @@
...
+		err = json.Unmarshal(bidderExt.Bidder, &impressionExt)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if impressionExt.Token == "" {
+			errs = append(errs, errors.New("Applogy token required"))
+			continue
+		}
+		request.Imp = impressions[i : i+1]
```
guscarreon: Do you think `request.Imp = []openrtb.Imp{impressions[i]}` is an easier to read assignment?

adapters/applogy/applogy.go
```diff
@@ -0,0 +1,157 @@
...
+		body, err := json.Marshal(request)
+		if err != nil {
+			errs = append(errs, err)
+			return nil, errs
```
guscarreon: Do you want to exit the function even if there could be other "healthy" elements inside the `impression` slice to check for?

adapters/applogy/applogy.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		result = append(result, &adapters.RequestData{
+			Method:  "POST",
+			Uri:     a.endpoint + "/" + impressionExt.Token,
+			Body:    body,
+			Headers: headers,
+		})
```
guscarreon: This approach will have at most one `Imp` per request. Are you ok with this?

adapters/applogy/applogy.go
```diff
@@ -0,0 +1,157 @@
...
+	}
```
guscarreon: The current logic allows `MakeRequests` to return an empty `result` array in the following two corner scenarios:
   1) `request.Imp` was empty to begin with
   2) `request.Imp` was not empty but none of its elements passed the checks inside the `for` loop and we are returning an empty `result` array and an `errs` array of lenght `len(impressions)`

Do you want to append one more error to `errs` describing this situation and setting `result` to `nil`?

adapters/applogy/applogy.go
IncSW: Only banner has format, video and native no need to check.

adapters/applogy/applogy.go
IncSW: Its ok.

adapters/applogy/applogy.go
IncSW: Yes, it's easy to read but extra allocs.

adapters/applogy/applogy.go
IncSW: Thanks, fixed.

IncSW:
@guscarreon now coverage: 96.1% of statements =)

adapters/applogy/applogy.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: err.Error(),
+		}}
+	}
+
+	response := adapters.NewBidderResponseWithBidsCapacity(len(request.Imp))
+
+	for _, seatBid := range bidResponse.SeatBid {
+		for _, bid := range seatBid.Bid {
+			bid := bid
```
SyntaxNode: Why?

IncSW: pin https://github.com/kyoh86/scopelint#whats-this

PR #1155, opened 2020-01-02
Updated PubMatic endpoint to use https

PubMatic-OpenWrap:


SyntaxNode:
There is a test failure with `TestDefaults`. For some reason, it appears the pubmatic url is used as a test value. Can you please either update the test or if you like, switch it to using the appnexus adapter as a test value.

You may also want to update `pubmatic_test.go` with the https variant to reduce confusion, although it's not strictly necessary.

PubMatic-OpenWrap:
Fixed failing tests and updated endpoint url at appropriate places.

SyntaxNode:
Changes look good. Why does it include the MarsMedia adapter as well? Perhaps something wrong with merging in from master?

SyntaxNode:
Just wanted to check in to see if you can rebase the PR to remove the MarsMedia bits.

PubMatic-OpenWrap:
Update PR with pull from prebid_master. Please verify.

PR #1159, opened 2020-01-06
Added new adapter for CPMStar ad network banners and video

JoshuaMGoldstein:
Added new adapter for CPMStar ad network banners and video


adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,158 @@
...
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+
+	return &adapters.RequestData{
+		Method:  "POST",
+		Uri:     a.endpoint,
+		Body:    jsonBody,
+		Headers: headers,
+	}, errs
```
SyntaxNode: Consider returning `nil` for errors instead of the empty slice for clarity.

adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,158 @@
...
+	var errs []error
+	var adapterRequests []*adapters.RequestData
+
+	if err := preprocess(request); err != nil {
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	adapterReq, errors := a.makeRequest(request)
+	if adapterReq != nil {
```
SyntaxNode: It's more common to check for `errors != nil` and deduce the data is valid.

adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,158 @@
...
+		var imp = request.Imp[i]
+		var bidderExt adapters.ExtImpBidder
+
+		if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
+			return &errortypes.BadInput{
+				Message: err.Error(),
+			}
+		}
+
+		if err := validateImp(&request.Imp[i]); err != nil {
```
SyntaxNode: Consider referencing the `imp` var here instead of indexing again.

adapters/cpmstar/cpmstar_test.go
```diff
@@ -0,0 +1,11 @@
+package cpmstar
+
+import (
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+)
+
+func TestJsonSamples(t *testing.T) {
+	adapterstest.RunJSONBidderTest(t, "cpmstartest", NewCpmstarBidder("https://server.cpmstar.com/openrtbbidrq.aspx"))
```
SyntaxNode: FYI: You don't need to use a real url for the tests. Consider using a simpler value like "host".

adapters/cpmstar/usersync_test.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"testing"
+	"text/template"
+
+	"github.com/prebid/prebid-server/privacy"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCpmstarSyncer(t *testing.T) {
+	temp := template.Must(template.New("sync-template").Parse("https://server.cpmstar.com/usersync.aspx?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redirectUri=http%3A%2F%2Flocalhost:8000%2Fsetuid%3Fbidder%3Dcpmstar%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26us_privacy%3D{{.USPrivacy}}%26uid%3D%24UID"))
```
SyntaxNode: We recently updated the standard style of these tests to improve readability. Please consider using the update style, ex:
https://github.com/prebid/prebid-server/blob/master/adapters/appnexus/usersync_test.go

config/config.go
```diff
@@ -496,6 +496,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderBrightroll, "https://pr-bh.ybp.yahoo.com/sync/appnexusprebidserver/?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&url="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dbrightroll%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConsumable, "https://e.serverbid.com/udb/9969/match?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconsumable%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConversant, "https://prebid-match.dotomi.com/prebid/match?rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconversant%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderCpmstar, "https://server.cpmstar.com/usersync.aspx?redirect="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dcpmstar%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26us_privacy%3D{{.USPrivacy}}%26uid%3D%24UID")
```
SyntaxNode: The tests include passing gdpr, gdpr_consent, and us_privacy to your user sync endpoint, but those are not present here. You also don't need to include the us_privacy value in the redirectUri as that is a callback to Prebid Server.

exchange/adapter_map.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/prebid/prebid-server/adapters/brightroll"
 	"github.com/prebid/prebid-server/adapters/consumable"
 	"github.com/prebid/prebid-server/adapters/conversant"
+	"github.com/prebid/prebid-server/adapters/cpmstar"
```
SyntaxNode: Thank you for maintaining the alphabetization.

adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+//Adapter :
```
SyntaxNode: I'm guessing you are adding these comments to avoid linter warnings. Please consider providing a simple comment rather than leaving it empty. If you want to omit these kinds of comments, that's fine by me.

adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,158 @@
...
+			return &errortypes.BadInput{
+				Message: err.Error(),
+			}
+		}
+
+		if err := validateImp(&request.Imp[i]); err != nil {
+			return err
+		}
+
+		var extImp openrtb_ext.ExtImpCpmstar
```
SyntaxNode: It looks like you're using `json.Unmarshal` to validate the extension content is valid json. Consider passing it through as it and letting your servers reject invalid extensions.

adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,158 @@
...
+		if err := json.Unmarshal(bidderExt.Bidder, &extImp); err != nil {
+			return &errortypes.BadInput{
+				Message: err.Error(),
+			}
+		}
+
+		extImpJSON, err := json.Marshal(extImp)
```
SyntaxNode: This is not needed because the `extImp` is not changed. JSON Marshaling/Unmarshaling is expensive in terms of runtime and memory, so it's best to call them sparingly.

adapters/cpmstar/cpmstartest/supplemental/explicit-dimensions.json
```diff
@@ -0,0 +1,58 @@
+{
```
SyntaxNode: Curious. There doesn't seem to be any logic in this adapter for dimensions, so what are you testing here?

adapters/cpmstar/params_test.go
```diff
@@ -0,0 +1,54 @@
...
+	`{"placementId": 154, "subpoolId": 123}`,
+}
+
+var invalidParams = []string{
+	`{}`,
+	`null`,
+	`true`,
+	`154`,
+	`{"placementId": "154"}`,
+	`{"placementId": 154, "subpoolId": "123"}`,
```
SyntaxNode: The invalidity of this case is a bit subtle. Consider adding a comment to explain the type mistmatch.

adapters/cpmstar/cpmstar.go
JoshuaMGoldstein: Error responses are going to be a lot more informative here if the user uses a quoted string for placementId instead of a number, for example. So we'd prefer to maintain the unmarshalling here. I've removed the re-marshalling, per your other comment.

adapters/cpmstar/cpmstartest/supplemental/explicit-dimensions.json
JoshuaMGoldstein: This is just to ensure that the normal dimensions plus formats case is shown as a potential usage method to end users and to ensure that we show it as supported.

JoshuaMGoldstein:
@SyntaxNode I've added a commit addressing all of your issues, with the exception of the two I responded to which we'd prefer to maintain the way it is if possible.

adapters/cpmstar/cpmstar.go
SyntaxNode: I didn't see a change here. Is there a reason to use `request.Imp[i]` here instead of `imp`?

JoshuaMGoldstein:
Sorry about that, you're right - I had made that change but it got lost when working through some of the other requested fixes. Thanks for spotting that! I've now pushed a new commit containing that patch as well.

adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,149 @@
...
+			if foundMatchingBid {
+				rv.Bids = append(rv.Bids, &adapters.TypedBid{
+					Bid:     &bid,
+					BidType: bidType,
+				})
+			} else {
+				errors = append(errors, &errortypes.BadServerResponse{
+					Message: fmt.Sprintf("bid id='%s' could not find valid impid='%s'", bid.ID, bid.ImpID),
+				})
+			}
```
guscarreon: It would be desirable to set a test case up cover the scenario where no matching bid was found. Maybe even when a request returning a `StatusCode` different than `StatusOK`
image

adapters/cpmstar/cpmstar.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return nil
+}
+
+// MakeBids based on cpmstar server response
+func (a *Adapter) MakeBids(bidRequest *openrtb.BidRequest, unused *adapters.RequestData, responseData *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if responseData.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
```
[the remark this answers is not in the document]
JoshuaMGoldstein: This is done on the next line where it checks ```if responseData.StatusCode != http.StatusOK {```

JoshuaMGoldstein:
I've merged and pushed a new commit that addresses all the requested changes. Test coverage is now at 93.8%.

PR #1160, opened 2020-01-07
Update Example AppNexus Placement ID

SyntaxNode:
Changes to an innocuous simple value for most tests and examples. Updated to a working test placement id for openrtb examples, since the docs reference those examples and users run them and except success.


hhhjort:
Do we know that all the placement IDs you replaced are not valid test replacements?

SyntaxNode:
> Do we know that all the placement IDs you replaced are not valid test replacements?
> 
> Nevermind, it should be fine. adtelligent using the same placement ID as appnexus is probably just a cut and paste artifact on their part.

Yeah. That's what it looked like to me too.

PR #1161, opened 2020-01-08
Update the Conversant sync pixel

johnwier:
Update the sync pixel to use the new conversant pixel


config/config.go
```diff
@@ -495,7 +495,7 @@ func (cfg *Configuration) setDerivedDefaults() {
...
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderBrightroll, "https://pr-bh.ybp.yahoo.com/sync/appnexusprebidserver/?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&url="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dbrightroll%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConsumable, "https://e.serverbid.com/udb/9969/match?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconsumable%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
-	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConversant, "https://prebid-match.dotomi.com/prebid/match?rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconversant%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConversant, "https://prebid-match.dotomi.com/match/bounce/current?rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconversant%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26us_privacy%3D{{.USPrivacy}}%26uid%3D")
```
SyntaxNode: Why did you add the us_privacy to the redirect url?

johnwier: The original intent was to get data related to CCPA but those requests may not fire the sync pixel at all so I can remove it.

SyntaxNode: Gotcha. As of now, you should not expect the sync pixel to be fired at all. I would except that if you wanted to look at the value of the us_privacy param, you would add a query param to your url alongside the url param, rather than parsing it from the embedded Prebid Server callback / redirect url.

johnwier:
Thought it was all resolved.  I’ll take care of it

PR #1165, opened 2020-01-14
[Synacormedia] Add tagId bidder parameter

coreykress:
This PR has two main goals:
1. Add a new bidder parameter called "tagId" which we will pass along with our imp object
2. Update the test values to be more consistent with internal documentation


static/bidder-params/synacormedia.json
```diff
@@ -8,8 +8,12 @@
...
       "description": "The seat id."
+    },
+    "tagId": {
+      "type": "string",
+      "description": "The tag id."
     }
   },
 
-  "required": ["seatId"]
+  "required": ["seatId", "tagId"]
```
hhhjort: Do you have any current traffic coming from PBS? When you set `tagId` as required, if a request comes in missing `tagId` once it is set to required, it will mark the whole request as failed and not send a bid request to any bidder. So adding a required field can break the entire prebid setup of a publisher.

This behavior is to help ensure publishers set up their configs correctly. Otherwise a bidder might set up a request with 5 bidders, but bidder ABC has a bad config and always errors out. If the other 4 are working normally, the publisher might not notice they have a bad config for quite a while. If some pages have a valid ABC config, they might just conclude that ABC just doesn't bid much if they don't troubleshoot carefully.

If there are publishers who currently have configs that omit `tagId`. we would recommend that initially the adapter not advertise `tagId` as required in `bidder-params`, but return an error from within the adapter if it is missing. An error from within the adapter will not invalid the whole request, allowing the other bidders to function. Then after all publishers have had a reasonable amount of time to update their configs, we can do the additional step of making it required in `bidder-params`.

adapters/synacormedia/synacormedia.go
```diff
@@ -72,11 +75,12 @@ func (a *SynacorMediaAdapter) makeRequest(request *openrtb.BidRequest) (*adapter
...
 
-	if firstExtImp == nil || firstExtImp.SeatId == "" {
+	if firstExtImp == nil || firstExtImp.SeatId == "" || firstExtImp.TagId == "" {
 		return nil, append(errs, &errortypes.BadServerResponse{
-			Message: fmt.Sprintf("Impression missing seat id"),
+			Message: fmt.Sprintf("Invalid Impression"),
 		})
 	}
+	// this is where the empty seatId is filled
 	re = &ReqExt{SeatId: firstExtImp.SeatId}
```
guscarreon: Isn't `TagId` also meant to be copied into the `ReqExt` object? I realize `ReqExt` isn't even defined with a `TagId` field, I'm just wondering if we don't want that value in there.

coreykress: TagId is a part of the Imp object, so for each Imp in request.Imp we get the TagId out of the Ext object and add it to the root of the Imp object and then add all those Imps to validImps.  Those become the new request.Imp that gets sent out.

SyntaxNode:
Looks like there's some unaddressed feedback in this PR, @coreykress.

coreykress:
@SyntaxNode currently dealing with some internal processes.  Will update shortly.

static/bidder-params/synacormedia.json
coreykress: I'm not sure who is using this tag, so I went ahead and removed tagId from the required params.

adapters/synacormedia/synacormedia.go
```diff
@@ -55,14 +56,16 @@ func (a *SynacorMediaAdapter) makeRequest(request *openrtb.BidRequest) (*adapter
...
-		validImp, err := getExtImpObj(&imp)
+		validExtImpObj, err := getExtImpObj(&imp) // getExtImpObj returns {seatId:"", tagId:""}
 		if err != nil {
 			errs = append(errs, err)
 			continue
 		}
+		// right here is where we need to take out the tagId and then add it to imp
+		imp.TagID = validExtImpObj.TagId
 		validImps = append(validImps, imp)
 		if firstExtImp == nil {
```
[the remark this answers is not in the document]
coreykress: I updated the code to note any empty tagId or seatID values, then skip over them.  There is a unit test for this now.

coreykress:
@SyntaxNode, I added a new commit to address the feedback that was left to me about a week ago.  If anyone has time to look at it I'd appreciate it.

adapters/synacormedia/synacormedia.go
SyntaxNode: You may wish to consider using a different error message here and on line 67 so the unique error message may assist in debugging.

Issue #1136, opened 2019-12-06
SIGSEGV in case of invalid currency server


moteus:
```
 E1206 13:50:19.958735    7353 rate_converter.go:114] Error updating conversion rates: Get : unsupported protocol scheme ""
 E1206 13:50:19.981677    7353 facebook.go:411] No facebook partnerID specified. Calls to the Audience Network will fail. Did you set adapters.facebook.platform_id in the app config?
 panic: runtime error: invalid memory address or nil pointer dereference
 [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x8ac7fc]
 
 goroutine 1 [running]:
 github.com/prebid/prebid-server/currencies.(*RateConverter).GetInfo(0xc0000eefc0, 0x0, 0x0)
     /tmp/prebid-server/currencies/rate_converter.go:179 +0xac
 github.com/prebid/prebid-server/endpoints.newCurrencyRatesInfo(0xdcaea0, 0xc0000eefc0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
     /tmp/prebid-server/endpoints/currency_rates.go:39 +0x96
 github.com/prebid/prebid-server/endpoints.NewCurrencyRatesEndpoint(0xdcaea0, 0xc0000eefc0, 0x12)
     /tmp/prebid-server/endpoints/currency_rates.go:61 +0x4d
 github.com/prebid/prebid-server/router.Admin(0x0, 0x0, 0xc0000eefc0, 0xc0002fc080)
     /tmp/prebid-server/router/admin.go:22 +0x17c
 main.serve(0x0, 0x0, 0xc000292a00, 0x0, 0x0)
     /tmp/prebid-server/main.go:63 +0x16a
 main.main()
     /tmp/prebid-server/main.go:39 +0x11c
```
The problem in [the line](https://github.com/prebid/prebid-server/blob/master/currencies/rate_converter.go#L179) `rc.Rates().GetRates()`.
Function `rc.Rates()` may returns nil so there need to add check for it

benjaminch:
Hello @moteus !

Thanks for reporting this one, just push a patch aiming to fix this.

Cheers !

PR #1169, opened 2020-01-19
[currencies] fix GetInfo() null ref issue

benjaminch:
This CL fixes the null ref on `RateConverter.GetInfo()` when rates
are nil.

Issue: #1136

### DIFF
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -172,11 +172,15 @@ func (rc *RateConverter) Rates() Conversions {
 
 // GetInfo returns setup information about the converter
 func (rc *RateConverter) GetInfo() ConverterInfo {
+	var rates *map[string]map[string]float64
+	if rc.Rates() != nil {
+		rates = rc.Rates().GetRates()
+	}
 	return converterInfo{
 		source:           rc.syncSourceURL,
 		fetchingInterval: rc.fetchingInterval,
 		lastUpdated:      rc.LastUpdated(),
-		rates:            rc.Rates().GetRates(),
+		rates:            rates,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


currencies/rate_converter.go
```diff
@@ -172,11 +172,15 @@ func (rc *RateConverter) Rates() Conversions {
...
+	if rc.Rates() != nil {
+		rates = rc.Rates().GetRates()
+	}
 	return converterInfo{
 		source:           rc.syncSourceURL,
 		fetchingInterval: rc.fetchingInterval,
 		lastUpdated:      rc.LastUpdated(),
-		rates:            rc.Rates().GetRates(),
+		rates:            rates,
 	}
```
guscarreon: Since `rc.Rates()` can return `nil`, the `rates` field in the `ConverterInfo` object we return from the `GetInfo()` function can be `nil` too. Do we want that? Isn't that a risk of another null pointer dereference? After all, the `(ci converterInfo) Rates()` function just returns `rates` without checking for nil either. Are we ok with letting the caller check for `nil` on `Rates()` return value?
```
// Rates returns converter's internal rates
func (ci converterInfo) Rates() *map[string]map[string]float64 {
        return ci.rates
}
endpoints/currency_rates.go
```

benjaminch: I think it's ok. This one is used in the info handler only. Having nil rates is valid but we should make sure it doesn't create a null ref. 
Before this patch, it was creating a null ref which cannot be handled by the `GetInfos` caller.

guscarreon: Makes sense. Thank you @benjaminch 
Approved

benjaminch:
Thank you ! @guscarreon & @SyntaxNode ;)

PR #1181, opened 2020-01-30
Adding bid rejection messages to debug response

camrice:


exchange/exchange.go
```diff
@@ -146,10 +146,16 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 		//If includebrandcategory is present in ext then CE feature is on.
 		if requestExt.Prebid.Targeting != nil && requestExt.Prebid.Targeting.IncludeBrandCategory != nil {
 			var err error
-			bidCategory, adapterBids, err = applyCategoryMapping(ctx, requestExt, adapterBids, *categoriesFetcher, targData)
+			var rejections []string
+			bidCategory, adapterBids, rejections, err = applyCategoryMapping(ctx, requestExt, adapterBids, *categoriesFetcher, targData)
 			if err != nil {
 				return nil, fmt.Errorf("Error in category mapping : %s", err.Error())
 			}
+			if len(rejections) > 0 {
```
SyntaxNode: The length check is not required. The `for` loop alone is enough.

exchange/exchange.go
```diff
@@ -405,6 +414,8 @@ func applyCategoryMapping(ctx context.Context, requestExt openrtb_ext.ExtRequest
 						//TODO: add metrics
 						//if mapping required but no mapping file is found then discard the bid
 						bidsToRemove = append(bidsToRemove, bidInd)
+						tempReason := fmt.Sprintf("Category mapping file for primary ad server: '%s', publisher: '%s' not found", primaryAdServer, publisher)
```
SyntaxNode: Can you just call it `reason` instead of `tempReason`?

exchange/exchange.go
```diff
@@ -447,11 +459,13 @@ func applyCategoryMapping(ctx context.Context, requestExt openrtb_ext.ExtRequest
 					if dupe.bidderName == bidderName {
 						// An older bid from the current bidder
 						bidsToRemove = append(bidsToRemove, dupe.bidIndex)
+						rejections = updateRejections(rejections, dupe.bidID, "Bid was deduplicated")
```
SyntaxNode: Consider using `bidID` for consistency with the other calls..

exchange/exchange.go
```diff
@@ -484,18 +499,17 @@ func applyCategoryMapping(ctx context.Context, requestExt openrtb_ext.ExtRequest
 
 	}
 	if len(seatBidsToRemove) > 0 {
```
SyntaxNode: This `if` statement can be removed too.

exchange/exchange.go
```diff
@@ -484,18 +499,17 @@ func applyCategoryMapping(ctx context.Context, requestExt openrtb_ext.ExtRequest
...
 		}
 	}
 
-	return res, seatBids, nil
+	return res, seatBids, rejections, nil
+}
+
+func updateRejections(rejections []string, bidID string, reason string) []string {
+	message := fmt.Sprintf("bid rejected [bid ID: %s] reason: %s", bidID, reason)
+	return append(rejections, message)
```
SyntaxNode: Does this checkout ok? Appending to the rejections would alter the slice, so wouldn't this need to be a reference?

exchange/exchange_test.go
```diff
@@ -930,9 +930,10 @@ func TestCategoryMapping(t *testing.T) {
 
 	adapterBids[bidderName1] = &seatBid
 
-	bidCategory, adapterBids, err := applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
+	bidCategory, adapterBids, rejections, err := applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
 
 	assert.Equal(t, nil, err, "Category mapping error should be empty")
+	assert.Equal(t, 1, len(rejections), "There should be 1 bid rejection message")
```
SyntaxNode: I would like the test to verify the content of the rejection message.

exchange/exchange_test.go
```diff
@@ -983,9 +984,10 @@ func TestCategoryMappingNoIncludeBrandCategory(t *testing.T) {
 
 	adapterBids[bidderName1] = &seatBid
 
-	bidCategory, adapterBids, err := applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
+	bidCategory, adapterBids, rejections, err := applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
 
 	assert.Equal(t, nil, err, "Category mapping error should be empty")
+	assert.Equal(t, 0, len(rejections), "There should be no bid rejection messages")
```
SyntaxNode: Consider using `assert.Empty` instead.

exchange/exchange_test.go
```diff
@@ -1196,6 +1201,131 @@ func TestCategoryDedupe(t *testing.T) {
...
+	assert.Equal(t, 2, len(bidCategory), "Bidders category mapping doesn't match")
+}
+
+func TestUpdateRejections(t *testing.T) {
+	rejections := []string{}
+
+	rejections = updateRejections(rejections, "bid_id1", "some reason 1")
+	rejections = updateRejections(rejections, "bid_id2", "some reason 2")
+
+	assert.Equal(t, 2, len(rejections), "Rejections should contain 2 rejection messages")
```
SyntaxNode: I don't think the order of the rejection messages matter, so consider using `assert.Contains`.

exchange/exchange_test.go
```diff
@@ -1196,6 +1201,131 @@ func TestCategoryDedupe(t *testing.T) {
...
+	innerBids := []*pbsOrtbBid{
+		&bid1_1,
+	}
+
+	seatBid := pbsOrtbSeatBid{innerBids, "USD", nil, nil}
+	bidderName := openrtb_ext.BidderName("appnexus")
+
+	adapterBids[bidderName] = &seatBid
+
+	bidCategory, adapterBids, rejections, err := applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
```
SyntaxNode: The test is confusing to read because it contains several different test cases. In this situation, it's ideal to use a table driven test approach. Look at the tests in the privacy package as examples.

exchange/exchange_test.go
```diff
@@ -1196,6 +1201,131 @@ func TestCategoryDedupe(t *testing.T) {
...
+		&bid1_1,
+		&bid1_2,
+		&bid1_3,
+	}
+
+	seatBid = pbsOrtbSeatBid{innerBids, "USD", nil, nil}
+
+	bidCategory, adapterBids, rejections, err = applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
+
+	// There is a random factor to deduplication
```
SyntaxNode: Random and Tests aren't a good match. Think of what is the cause of the randomness and how it can be factored away for the test. If it's just order dependent, consider using assert statements which ignore order.. like `assert.Contains`.

exchange/exchange.go
jmaynardxandr: This is correct as is in this case, we are removing dupe.bidID instead of bidID (these are two different bids).

exchange/exchange.go
```diff
@@ -447,11 +459,13 @@ func applyCategoryMapping(ctx context.Context, requestExt openrtb_ext.ExtRequest
 					if dupe.bidderName == bidderName {
 						// An older bid from the current bidder
 						bidsToRemove = append(bidsToRemove, dupe.bidIndex)
+						rejections = updateRejections(rejections, dupe.bidID, "Bid was deduplicated")
 					} else {
 						// An older bid from a different seatBid we've already finished with
 						oldSeatBid := (seatBids)[dupe.bidderName]
 						if len(oldSeatBid.bids) == 1 {
 							seatBidsToRemove = append(seatBidsToRemove, bidderName)
+							rejections = updateRejections(rejections, bidID, "Bid was deduplicated")
```
jmaynardxandr: dupe.bidID should also be used here as this is still a case where we are removing the bid that was stored in dupe instead of the current one (bidID).

exchange/exchange.go
jmaynardxandr: From what I've seen it's a pretty common pattern with slices to pass by value and return a modified slice rather than passing by reference, the append function itself is a good example of that. That said I do agree with you that        it's more confusing that way, but it's idoimatic. I guess what I'm saying is I don't really have an opinion after all. :)

exchange/exchange_test.go
jmaynardxandr: Maybe passing in a function that decides which bid to keep instead of hard-coding it to a 50% chance there where it picks (so test could use a function that just always picks the first)? I kind of wanted to re-work that code eventually anyway, not a fan of how the original bid would have to win many coin flips in a row to remain there if there were like 4 or 5 duplicates found (probably unlikely, but it's still one of those things that bugs me to this day).

exchange/exchange.go
hhhjort: A slice is basically 3 bits of data. A pointer to some memory, the length of the slice that extends beyond that pointer, and the max size that the slice can expand to given the current memory allocation. So effectively it is a pointer as far as the data contained. The resulting slice is returned, because if we overrun the max size, we will have to allocate a new chunk of memory for the larger slice, which will have no relation to the current chunk of memory. Thus we push up the replaced slice back to the caller so it will have the correct data there. This is how the built in `append()` function works.

exchange/exchange.go
camrice: Removed the length check here

exchange/exchange.go
camrice: Updated to be `reason`

exchange/exchange.go
camrice: Removed length check here as well

exchange/exchange_test.go
camrice: Added verification for rejection messages here and in other tests where `rejections` was being tested.

exchange/exchange_test.go
camrice: Updated to use `assert.Empty`, both here and other related tests.

exchange/exchange_test.go
camrice: Updated the test to be more readable. I used a similar approach to the policy tests. Now runs a simple set of tests in a for loop so that it is easier to follow.

exchange/exchange.go
camrice: Updated to use the correct duplicate bid ID

exchange/exchange_test.go
camrice: Updated to use `assert.Contains` instead

exchange/exchange_test.go
camrice: Updated this test here to instead use the same bid ID so that the expected messages are always the same. The removal of bids uses the index instead of the ID, so I am just using two bids that are the same and expecting 1 to be removed, and 1 to be in the resulting bidCategory. Let me know if I should use a different approach (also could use the `assert.Regex` if we want it to have different IDs).

exchange/exchange_test.go
```diff
@@ -1179,9 +1186,12 @@ func TestCategoryDedupe(t *testing.T) {
 
 		adapterBids[bidderName1] = &seatBid
 
-		bidCategory, adapterBids, err := applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
+		bidCategory, adapterBids, rejections, err := applyCategoryMapping(nil, requestExt, adapterBids, categoriesFetcher, targData)
 
 		assert.Equal(t, nil, err, "Category mapping error should be empty")
+		assert.Equal(t, 2, len(rejections), "There should be 2 bid rejection messages")
+		assert.Regexpf(t, regexp.MustCompile(`bid rejected \[bid ID: bid\_id(1|3)\] reason: Bid was deduplicated`), rejections[0], "Rejection message did not match expected")
```
jmaynardxandr: It probably doesn't hurt anything, but the `_` character shouldn't need to be escaped in the regex.

exchange/exchange_test.go
jmaynardxandr: Same bid id = same expected message, that's a super simple way to work around the randomness with duplicate bid removal, I think it's a great idea!

exchange/exchange_test.go
camrice: Removed the character escape

exchange/exchange.go
SyntaxNode: >  The removal of bids uses the index instead of the ID

That's the part I was missing.

exchange/exchange.go
```diff
@@ -146,10 +146,14 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
 		if requestExt.Prebid.Targeting != nil && requestExt.Prebid.Targeting.IncludeBrandCategory != nil {
 			var err error
-			bidCategory, adapterBids, err = applyCategoryMapping(ctx, requestExt, adapterBids, *categoriesFetcher, targData)
+			var rejections []string
+			bidCategory, adapterBids, rejections, err = applyCategoryMapping(ctx, requestExt, adapterBids, *categoriesFetcher, targData)
 			if err != nil {
 				return nil, fmt.Errorf("Error in category mapping : %s", err.Error())
 			}
+			for _, message := range rejections {
+				errs = append(errs, fmt.Errorf(message))
```
SyntaxNode: Minor: You're calling `fmt.Errorf` which allows for formatting parameters, but that's never used here. Consider switching to `errors.New`.

camrice: Updated to use `errors.New` instead

PR #1182, opened 2020-01-31
Adds timeout notifications for Facebook

hhhjort:


adapters/audienceNetwork/facebook.go
```diff
@@ -447,3 +447,18 @@ func NewFacebookBidder(client *http.Client, platformID string, appSecret string)
...
+
+func (fa *FacebookAdapter) MakeTimeoutNotification(req *adapters.RequestData) (*adapters.RequestData, []error) {
+	auction_id, err := jsonparser.GetString(req.Body, "imp", "[0]", "id")
+	if err != nil {
+		return &adapters.RequestData{}, []error{err}
+	}
+
+	timeoutReq := adapters.RequestData{
+		Method:  "GET",
+		Uri:     fmt.Sprintf("https://www.facebook.com/audiencenetwork/nurl/?partner=%s&app=%s&auction=%s&ortb_loss_code=2", fa.platformID, fa.platformID, auction_id),
```
SyntaxNode: Consider moving the uri construction up a level. IMHO the extra indentation doesn't make the string substitutions any easier to see.

adapters/bidder.go
```diff
@@ -39,6 +39,20 @@ type Bidder interface {
 	MakeBids(internalRequest *openrtb.BidRequest, externalRequest *RequestData, response *ResponseData) (*BidderResponse, []error)
 }
 
+// TimeoutBidder is used to identify bidders that support timeout notifications.
+type TimeoutBidder interface {
+	MakeRequests(request *openrtb.BidRequest, reqInfo *ExtraRequestInfo) ([]*RequestData, []error)
```
SyntaxNode: Why are the standard bidder methods duplicated here?

exchange/bidder.go
```diff
@@ -295,6 +296,10 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 	if err != nil {
 		if err == context.DeadlineExceeded {
 			err = &errortypes.Timeout{Message: err.Error()}
+			if tb, ok := bidder.Bidder.(adapters.TimeoutBidder); ok {
+				go bidder.doTimeoutNotification(tb, req)
```
SyntaxNode: This might benefit from a comment explaining the purposeful 'set it and forget it' approach. Should be also be concerned about creating too many go routines? Probably not, but wanted to seek opinions.

adapters/bidder.go
hhhjort: Not 100% sure it is needed, but it does make clear that it is a bidder rather than some other random thing, and can still be used for bidder things.

exchange/bidder.go
hhhjort: We can only do one timeout call per timeout, so it is probably not too much of an issue.

hhhjort:
We can only generate one timeout notification per timeout, and we already have metrics on timeout, so I don't see the need for an additional metric for explicit notifications

exchange/bidder.go
guscarreon: I'm split about proposing this because it does imply having an extra function to maintain and the benefits are marginal to say the least but, do you think a function to check whether or not the bidder implements `MakeTimeoutNotification` makes the code a bit easier to read?
```
283   // doRequest makes a request, handles the response, and returns the data needed by the
284   // Bidder interface.
285   func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.RequestData) *httpCallInfo {
286       httpReq, err := http.NewRequest(req.Method, req.Uri, bytes.NewBuffer(req.Body))
287   +--  9 lines: if err != nil {---------------------------------------------------------------------------
296       if err != nil {
297           if err == context.DeadlineExceeded {
298               err = &errortypes.Timeout{Message: err.Error()}
299 -             if tb, ok := bidder.Bidder.(adapters.TimeoutBidder); ok {
    +             if tb, ok := isTimeoutImplemented(bidder.Bidder); ok {
300                   go bidder.doTimeoutNotification(tb, req)
301               }
302  
303           }
304           return &httpCallInfo{
305               request: req,
306               err:     err,
307           }
308       }
309   +-- 16 lines: respBody, err := ioutil.ReadAll(httpResp.Body)--------------------------------------------
310       return &httpCallInfo{
326           request: req,
327           response: &adapters.ResponseData{
328               StatusCode: httpResp.StatusCode,
329               Body:       respBody,
330               Headers:    httpResp.Header,
331           },
332           err: err,
333       }
334   }
335  
    + func isTimeoutImplemented(adapterBidder adapters.Bidder) (adapters.TimeoutBidder, bool) {
    +     return adapterBidder.(adapters.TimeoutBidder)
    + }
     
exchange/bidder.go [+]
```

exchange/bidder.go
```diff
@@ -328,6 +333,22 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
...
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+	toReq, errL := timeoutBidder.MakeTimeoutNotification(req)
+	if toReq != nil && len(errL) == 0 {
+		httpReq, err := http.NewRequest(toReq.Method, toReq.Uri, bytes.NewBuffer(toReq.Body))
+		if err == nil {
+			httpReq.Header = req.Headers
+			httpResp, _ := ctxhttp.Do(ctx, bidder.Client, httpReq)
+			// No validation yet on sending notifications
+			_ = httpResp
```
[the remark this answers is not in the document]
hhhjort: Yes, that is probably best. I had pondered doing something with the response, but then decided against it.

adapters/audienceNetwork/facebook_test.go
```diff
@@ -40,3 +42,21 @@ type FacebookExt struct {
 func TestJsonSamples(t *testing.T) {
 	adapterstest.RunJSONBidderTest(t, "audienceNetworktest", NewFacebookBidder(nil, "test-platform-id", "test-app-secret"))
 }
+
+func TestMakeTimeoutNotice(t *testing.T) {
+	req := adapters.RequestData{
+		Body: []byte(`{"imp":[{"id":"1234"}]}}`),
```
mansinahar: Could we also add another test case for an invalid json in this test?

adapters/bidder.go
```diff
@@ -39,6 +39,20 @@ type Bidder interface {
...
+	MakeBids(internalRequest *openrtb.BidRequest, externalRequest *RequestData, response *ResponseData) (*BidderResponse, []error)
```
mansinahar: We should be able to get this same behavior of inheriting the standard `bidder` interface methods by embedding the bidder interface here. That way we won't have to manage the method signatures in two places.

adapters/audienceNetwork/facebook_test.go
hhhjort: Done

adapters/bidder.go
hhhjort: Done

exchange/bidder.go
```diff
@@ -328,6 +337,21 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 	}
 }
 
+func (bidder *bidderAdapter) doTimeoutNotification(timeoutBidder adapters.TimeoutBidder, req *adapters.RequestData) {
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
```
SyntaxNode: Why do we need to defer the cancellation?

exchange/bidder.go
```diff
@@ -328,6 +337,21 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 	}
 }
 
+func (bidder *bidderAdapter) doTimeoutNotification(timeoutBidder adapters.TimeoutBidder, req *adapters.RequestData) {
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
```
SyntaxNode: Should the 200ms timeout be configurable?

hhhjort: Pushing down config info is a non-trivial amount of work. Hard coding it at this time won't make that work any harder, and it doesn't seem likely that there will be a need. We can add in another PR as needed.

SyntaxNode: I'm cool with that.

exchange/bidder.go
SyntaxNode: The answer: memory leaks
https://stackoverflow.com/questions/44393995/what-happens-if-i-dont-cancel-a-context

Issue #1174, opened 2020-01-24
Prebid Server Long Form Video Deal Support Changes 


camrice:
Below is a work in progress write up of the intended changes.

# Deal Support

## Overview

Bids that contain deals may need to be updated with a bidder defined prefix and priority level. This will be reflected in the `hb_pb_cat_dur` value in the resulting `adPod`. The field `hb_pb_cat_dur`is usually formed with `{price bucket}_{industry code}_{duration}`. With this feature, `hb_pb_cat_dur` will instead be formed by `{bidder prefix}{deal priority}_{industry code}_{duration}` if all conditions are met.

Example:

Original:

```
"adPods": [
    {
        "podid": 1,
        "targeting": [
            {
                "hb_pb": "15.00",
                "hb_pb_cat_dur": "15.00_food_30s",
                "hb_cache_id": "420d7329-30e8-4c4e-8eaa-fe937172e4e0"
            }
        ],
        "errors": null
    }
],
```

Modified:

```
"adPods": [
    {
        "podid": 1,
        "targeting": [
            {
                "hb_pb": "15.00",
                "hb_pb_cat_dur": "tier7_food_30s",
                "hb_cache_id": "420d7329-30e8-4c4e-8eaa-fe937172e4e0"
            }
        ],
        "errors": null
    }
],
```

In order for the modification to occur, there are a few pieces that need to be setup. The configuration must contain a `dealTier` object that holds the `prefix` and `minDealTier` values. The request to Prebid Server must also enable this feature through the `SupportDeals` field. Note that in order for ImpBus to return the deal priority for resulting bids, deals must be configured for the target campaign outside of Prebid Server.

## Prebid Configuration

The prefix and minimum deal tier must be defined in the Prebid configuration in order for the key to be updated. These are set per bidder in a `dealTier` object.

| Field                | Type      | Description                                                                                                     |
| -------------------- | --------- | --------------------------------------------------------------------------------------------------------------- |
| dealTier             | 'Object'  | Contains prefix and minimum deal tier values                                                                    |
| dealTier.prefix      | 'String'  | The prefix string that is added to `hb_pb_cat_dur`. Must be non-empty.                                          |
| dealTier.minDealTier | 'Integer' | The minimum deal priority value (inclusive) that must be met for modification to occur. Must be greater than 0. |

Example:

```
[
    {
        "params": {
            "dealTier": {
                "prefix": "tier",
                "minDealTier": 5
            },
            "placementId": 12345678
        },
[...]

PR #1183, opened 2020-02-03
Adding support for deal prefixes

camrice:
Opening PR to get an early start on review comments. We have a couple questions remaining on our end. One major change that may need to be added is implementing the `pmp` object for targeting deals in the impressions. I will update this PR with any action items regarding this.

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -87,6 +87,7 @@ type appnexusBidExtAppnexus struct {
 	BrandId       int                    `json:"brand_id"`
 	BrandCategory int                    `json:"brand_category_id"`
 	CreativeInfo  appnexusBidExtCreative `json:"creative_info"`
+	DealPriority  int                    `json:"deal_priority"`
 }
 
 type appnexusBidExt struct {
@@ -543,9 +544,10 @@ func (a *AppNexusAdapter) MakeBids(internalRequest *openrtb.BidRequest, external
 					}
 
 					bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
-						Bid:      &bid,
-						BidType:  bidType,
-						BidVideo: impVideo,
+						Bid:          &bid,
+						BidType:      bidType,
+						BidVideo:     impVideo,
+						DealPriority: bidExt.Appnexus.DealPriority,
 					})
 				} else {
 					errs = append(errs, err)
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"runtime/debug"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/prebid/prebid-server/stored_requests"
@@ -167,10 +168,90 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 			}
 			targData.setTargeting(auc, bidRequest.App != nil, bidCategory)
 		}
+
+		if requestExt.Prebid.SupportDeals {
+			applyDealSupport(bidRequest, auc)
+		}
 	}
 
 	// Build the response
 	return e.buildBidResponse(ctx, liveAdapters, adapterBids, bidRequest, resolvedRequest, adapterExtra, auc, errs)
+}
+
+// TODO needs better naming
+type DealTierInfo struct {
+	Prefix      string `json:"prefix"`
+	MinDealTier int    `json:"minDealTier"`
+}
+
+type DealTier struct {
+	Info *DealTierInfo `json:"dealTier,omitempty"`
+}
+
+type BidderDealTier struct {
+	DealInfo map[string]*DealTier
+}
+
+// Updates targeting keys with deal prefixes if minimum deal tier exceeded
+func applyDealSupport(bidRequest *openrtb.BidRequest, auc *auction) {
+	impDealMap := getDealTiers(bidRequest)
+
+	for impId, topBidsPerImp := range auc.winningBidsByBidder {
+		impDeal := impDealMap[impId].DealInfo
+		for bidderName, topBidPerBidder := range topBidsPerImp {
+			if validateDealTier(impDeal[bidderName.String()]) {
+				updateCatDur(topBidPerBidder, impDeal[bidderName.String()].Info)
+			}
+		}
+	}
+}
+
+func getDealTiers(bidRequest *openrtb.BidRequest) map[string]*BidderDealTier {
+	impDealMap := make(map[string]*BidderDealTier)
+
+	for _, imp := range bidRequest.Imp {
+		var bidderDealTier BidderDealTier
+		err := json.Unmarshal(imp.Ext, &bidderDealTier.DealInfo)
+		if err != nil {
+			continue
+		}
+		impDealMap[imp.ID] = &bidderDealTier
+	}
+
+	return impDealMap
+}
+
+func validateDealTier(impDeal *DealTier) bool {
+	return impDeal.Info != nil && len(impDeal.Info.Prefix) > 0
+}
+
+func updateCatDur(topBidPerBidder *pbsOrtbBid, dealTierInfo *DealTierInfo) {
+	var anExt AppNexusExt
+	err := json.Unmarshal(topBidPerBidder.bid.Ext, &anExt)
+	if err != nil {
+		return
+	}
+
+	if anExt.Info.Priority >= dealTierInfo.MinDealTier {
+		newPb := fmt.Sprintf("%s%d", dealTierInfo.Prefix, anExt.Info.Priority)
+
+		if oldCatDur, ok := topBidPerBidder.bidTargets["hb_pb_cat_dur"]; ok {
+			oldCatDurSplit := strings.SplitAfterN(oldCatDur, "_", 2)
+			oldCatDurSplit[0] = newPb + "_"
+
+			newCatDur := strings.Join(oldCatDurSplit, "")
+			topBidPerBidder.bidTargets["hb_pb_cat_dur"] = newCatDur
+		}
+	}
+}
+
+type DealPriority struct {
+	Priority int `json:"deal_priority"`
+	Extra    json.RawMessage
+}
+
+type AppNexusExt struct {
+	Info DealPriority `json:"appnexus"`
 }
 
 func (e *exchange) makeAuctionContext(ctx context.Context, needsCache bool) (auctionCtx context.Context, cancel context.CancelFunc) {
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -205,10 +205,12 @@ func applyDealSupport(bidRequest *openrtb.BidRequest, auc *auction) []error {
 		for bidderName, topBidPerBidder := range topBidsPerImp {
 			bidderString := bidderName.String()
 
-			if validateDealTier(impDeal[bidderString]) {
-				updateCatDur(topBidPerBidder, impDeal[bidderString].Info)
-			} else {
-				errs = append(errs, fmt.Errorf("dealTier configuration invalid for bidder '%s', imp ID '%s'", bidderString, impID))
+			if topBidPerBidder.dealPriority > 0 {
+				if validateDealTier(impDeal[bidderString]) {
+					updateCatDur(topBidPerBidder, impDeal[bidderString].Info)
+				} else {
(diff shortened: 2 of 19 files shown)


camrice:
Relevant issue for this PR with general documentation: #1174

camrice:
Update to my initial comment:

The `pmp` object is not required for supporting deals, so this change is ready for review. There aren't any other major additions needed at this time.

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+
+// Updates targeting keys with deal prefixes if minimum deal tier exceeded
+func applyDealSupport(bidRequest *openrtb.BidRequest, auc *auction) []error {
+	errs := []error{}
+	impDealMap := getDealTiers(bidRequest)
+
+	for impID, topBidsPerImp := range auc.winningBidsByBidder {
+		impDeal := impDealMap[impID].DealInfo
+		for bidderName, topBidPerBidder := range topBidsPerImp {
+			bidderString := bidderName.String()
```
SyntaxNode: Nitpick: Including the type of the variable in the name is a bit of a 'smell' IMHO. Can we simply name it `bidder`?

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+					errs = append(errs, fmt.Errorf("dealTier configuration invalid for bidder '%s', imp ID '%s'", bidderString, impID))
+				}
+			}
+		}
+	}
+
+	return errs
+}
+
+// Creates map of impression to bidder deal tier configuration
```
SyntaxNode: Should this follow the go doc convention of `getDealTiers creates a map of...`?

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+		impDealMap[imp.ID] = &bidderDealTier
+	}
+
+	return impDealMap
+}
+
+func validateDealTier(impDeal *DealTier) bool {
+	if impDeal.Info != nil {
+		// Remove whitespace from prefix before checking if it can be used
+		impDeal.Info.Prefix = regexp.MustCompile(`\s+`).ReplaceAllString(impDeal.Info.Prefix, "")
```
SyntaxNode: Is it common for a prefix to include whitespace? aka... why is this needed?

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+		return len(impDeal.Info.Prefix) > 0 && impDeal.Info.MinDealTier > 0
+	}
+	return false
+}
+
+func updateCatDur(bid *pbsOrtbBid, dealTierInfo *DealTierInfo) {
```
SyntaxNode: This is likely industry knowledge I'm lacking, but what is a 'CatDur'?

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+	if bid.dealPriority >= dealTierInfo.MinDealTier {
+		newPb := fmt.Sprintf("%s%d_", dealTierInfo.Prefix, bid.dealPriority)
```
SyntaxNode: Could `newPb` have a more descriptive name?

adapters/appnexus/appnexus.go
```diff
@@ -87,6 +87,7 @@ type appnexusBidExtAppnexus struct {
 	BrandId       int                    `json:"brand_id"`
 	BrandCategory int                    `json:"brand_category_id"`
 	CreativeInfo  appnexusBidExtCreative `json:"creative_info"`
+	DealPriority  int                    `json:"deal_priority"`
```
SyntaxNode: To be clear, is this just modeling the response from AppNexus? I'm really asking if an int is an appropriate type or if it should be an int pointer or if an omit empty attribute should be added to the field.

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 			}
 			targData.setTargeting(auc, bidRequest.App != nil, bidCategory)
 		}
+
+		if requestExt.Prebid.SupportDeals {
+			dealErrs := applyDealSupport(bidRequest, auc)
+			if len(dealErrs) > 0 {
```
SyntaxNode: Is the length check needed? Go's builtin methods are kind to nil or empty collections, so `errs = append(errs, dealErrs...)` might work just fine.

exchange/exchange_test.go
```diff
@@ -1325,6 +1325,267 @@ func TestUpdateRejections(t *testing.T) {
...
+	bidderName2 := openrtb_ext.BidderName("validmultiple2")
+	bidderName3 := openrtb_ext.BidderName("minnotmet")
+	bidderName4 := openrtb_ext.BidderName("invalidconfig")
+	bidderName5 := openrtb_ext.BidderName("nodeal")
+	bidderName6 := openrtb_ext.BidderName("invalidconfig2")
+
+	bidRequest := &openrtb.BidRequest{
+		ID: "some-request-id",
+		Imp: []openrtb.Imp{
+			{
```
SyntaxNode: Why does this test need 6 imps? If each one is intended to be a unique test case, consider breaking it down using table driven tests. Same feedback for the other tests in this file.

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+		}
+
+		impDealMap[imp.ID] = &bidderDealTier
+	}
+
+	return impDealMap
+}
+
+func validateDealTier(impDeal *DealTier) bool {
+	if impDeal.Info != nil {
```
SyntaxNode: Consider switching the logic and including the `return false` up here for a quick exit, leaving the rest of the code not indented down a level.

exchange/exchange.go
```diff
@@ -167,12 +169,95 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+	return errs
+}
+
+// Creates map of impression to bidder deal tier configuration
+func getDealTiers(bidRequest *openrtb.BidRequest) map[string]*BidderDealTier {
+	impDealMap := make(map[string]*BidderDealTier)
+
+	for _, imp := range bidRequest.Imp {
+		var bidderDealTier BidderDealTier
+		err := json.Unmarshal(imp.Ext, &bidderDealTier.DealInfo)
```
SyntaxNode: Would this information have already been parsed previously somewhere else in code? Is there any way to avoid this unmarshal call?

exchange/exchange.go
camrice: The prefixes shouldn't contain whitespace, but currently the UI for the prebid configs doesn't have any validation or blocking for this case. This is used to just guard against those cases until the UI is able to validate it. Would it be better to drop the prefix modification if it does have whitespace?

exchange/exchange.go
camrice: Ah yeah the variable name is unnecessary, updated to `bidder`

exchange/exchange.go
camrice: Updated comment to follow doc convention

exchange/exchange.go
camrice: 'CatDur' is a shortening of the key value `hb_pb_cat_dur` but I don't think it is a standard shortening. I will update this function to be `updateHbPbCatDur` so that it is more specific.

exchange/exchange.go
SyntaxNode: Gotcha. I'm asking because the use of the regex raises a performance red flag and was going to see if a string trim method could be used instead.

exchange/exchange.go
SyntaxNode: Thank you. That makes sense.

exchange/exchange.go
camrice: Understood! Ideally this kind of check wouldn't be needed. I mainly used the regex to remove newlines and tabs, though that seems excessive now. I could use a string replace or trim instead to avoid this.

exchange/exchange.go
camrice: Ah yes this isn't a price bucket anymore. Updated to be `prefixTier`.

exchange/exchange.go
camrice: Length check is indeed not needed, removed.

exchange/exchange.go
camrice: Flipped the logic here so that the nil check will lead to false first

exchange/exchange_test.go
camrice: Refactored all added unit tests to use the table driven test approach. Also removed a couple test cases that were redundant.

exchange/exchange.go
camrice: I don't think the impression ext is unmarshalled elsewhere, at least not with the deal tier info. It seems the extension is used in the auction endpoint, but the map created there is bidder to raw message.

adapters/appnexus/appnexus.go
camrice: I think here the int is the appropriate type since it will initialize to 0 from the `TypedBid` struct. A value of 0 here would be considered a bid without a deal.

exchange/exchange.go
camrice: Updated to use `strings.ReplaceAll` instead of regex

adapters/appnexus/appnexus.go
SyntaxNode: Ok, cool.

exchange/exchange.go
```diff
@@ -167,12 +168,93 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+type DealTier struct {
+	Info *DealTierInfo `json:"dealTier,omitempty"`
+}
+
+type BidderDealTier struct {
+	DealInfo map[string]*DealTier
+}
+
+// Updates targeting keys with deal prefixes if minimum deal tier exceeded
+func applyDealSupport(bidRequest *openrtb.BidRequest, auc *auction) []error {
```
SyntaxNode: Please update comment to follow Go conventions:
```
applyDealSupport updates targeting keys....
```

exchange/exchange.go
```diff
@@ -167,12 +168,93 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+		}
+
+		impDealMap[imp.ID] = &bidderDealTier
+	}
+
+	return impDealMap
+}
+
+func validateDealTier(impDeal *DealTier) bool {
+	if impDeal.Info == nil {
```
SyntaxNode: Please add check for `impDeal == nil`. If it's not found in the DealInfo map, the impDeal will be nil here.

exchange/exchange.go
```diff
@@ -167,12 +168,93 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+			continue
+		}
+
+		impDealMap[imp.ID] = &bidderDealTier
+	}
+
+	return impDealMap
+}
+
+func validateDealTier(impDeal *DealTier) bool {
```
SyntaxNode: Please rename to `validateAndNormalizeDealTier` or `validateAndCleanDealTier` to make the mutation on line 240 obvious.

exchange/exchange.go
camrice: Updated comment to follow conventions

exchange/exchange.go
camrice: Added check for this case

exchange/exchange.go
camrice: Renamed to `validateAndNormalizeDealTier` to make usage more clear

ShriprasadM:
@SyntaxNode and @hhhjort : As per prebid documentation http://prebid.org/prebid-video/video-long-form.html#deal-support, `prioritizeDeals` parameter is used to convey if deals should be given a higher preference. But in this case, the parameter name we are expecting is `SupportDeals`. Can't we make it same as of Prebid i.e. `prioritizeDeals`? It will keep the nomenclature consistent.

Issue #1176, opened 2020-01-27
Prebid Server, AMP, and CCPA


bretg:
# Prebid Server, AMP, and CCPA

Researching how AMP actually supports CCPA consent, found that there's not a clear way to determine whether a given CONSENT_STRING is GDPR, CCPA, or other format. See https://github.com/ampproject/amphtml/pull/26220

In other words, I don't think the enhancement in https://github.com/prebid/prebid-server/pull/1125 is where we want to be. It assumes that

a) we're going to be getting a `us_privacy` parameter from AMP.
b) the value, if provided, will always be formatted as a CCPA string

Unfortunately, neither of these are true.

Instead, we're stuck in a situation where we're going to be receiving a CONSENT_STRING that's formatted in an unknown way. i.e. we have to parse the string to determine how to process it.

Currently the AMP `CONSENT_STRING` comes in on gdpr_consent, but in reality, that string isn't guaranteed to be formatted as an IAB TCF1.0 GDPR string. It could be IAB US_PRIVACY formatted.

Proposed path-forward has been reviewed with the committee and approved:

## Phase 1 - Combined GDPR+CCPA Support on two synonym keys

1) PBS should start looking for AMP parameters `gdpr_consent` and `consent_string`, treating them as the same. (consent_string won't come in initially, but after a transition phase, we'll convert to it so we don't get confused that the contents of gdpr_consent can actually be CCPA.)

2) PBS introspects `consent_string`/`gdpr_consent`:
2a) If consent_string is exactly 4 chars, starts with "1", and the next 3 chars are in the set [yYnN-] then it's a CCPA-formatted string. Pass it through OpenRTB as regs.ext.us_privacy.
2b) Pass consent_string through the TCF1.0 parsing function. If it successfully parses, send it through OpenRTB as user.ext.consent.
2c) Anything other format causes PBS to log a warning into the response and ignore it. Yes, it's understood that warnings in an AMP response are only seen in test mode.

3) Accepting/denying the AMP consent string is completely separate from Prebid Server "enforce" mode. i.e. if PBS determines that the user is in-scope for GDPR or CCPA, then a missing AMP consent_string will be a problem.

(Note that it doesn't make much sense to add `us_privacy` to the AMP RTC string because then we'll have two copies of the very long TCF1.0 string and still have to parse it to figure out what it really is.)

## Phase 2 - Update AMP RTC strings to use consent_string instead of gdpr_consent

[...]

PR #1187, opened 2020-02-10
AMP CCPA Fix

SyntaxNode:
Implements: https://github.com/prebid/prebid-server/issues/1176


endpoints/openrtb2/amp_auction.go
```diff
@@ -382,16 +380,15 @@ func (deps *endpointDeps) overrideWithParams(httpRequest *http.Request, req *ope
...
-	if err := privacyPolicies.Write(req); err != nil {
-		return err
+	consent := readConsent(httpRequest.URL)
+	if consent != "" {
+		if policies, ok := privacy.ReadPoliciesFromConsent(consent); ok {
+			if err := policies.Write(req); err != nil {
+				return err
+			}
+		} else {
+			// TODO: Send warning to caller.
```
SyntaxNode: Not sure where this warning should go. Asked for clarification in linked issue.

privacy/ccpa/policy.go
```diff
@@ -49,35 +50,44 @@ func (p Policy) Write(req *openrtb.BidRequest) error {
...
+		return fmt.Errorf("request.regs.ext.us_privacy %s", err.Error())
 	}
 
-	if p.Value[0] != '1' {
-		return errors.New("request.regs.ext.us_privacy must specify version 1")
+	return nil
+}
+
+// ValidateConsent returns an error if the CCPA consent string does not adhere to the IAB spec.
+func ValidateConsent(consent string) error {
```
SyntaxNode: I separated the functionality of validating the consent string from validating the entire policy. Allows the simpler function to be used for the "sensing" required of AMP.

privacy/gdpr/policy_test.go
```diff
@@ -72,3 +72,32 @@ func TestWrite(t *testing.T) {
...
+			description: "Invalid",
+			consent:     "<any invalid>",
+			expectError: true,
+		},
+		{
+			description: "Invalid",
+			consent:     "BONV8oqONXwgmADACHENAO7pqzAAppY",
+			expectError: false,
+		},
+	}
```
SyntaxNode: I don't care to test any other scenarios, since I don't want to focus on retesting the go-gdpr library here.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,357 +122,291 @@ func TestAMPPageInfo(t *testing.T) {
 	assert.Equal(t, "test.somepage.co.uk", exchange.lastRequest.Site.Domain)
 }
 
-func TestConsentThroughEndpoint(t *testing.T) {
```
SyntaxNode: The existing GDPR consent tests were difficult to work with. I converted them into a table driven test and added CCPA tests as well.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,357 +122,291 @@ func TestAMPPageInfo(t *testing.T) {
...
+				DigiTrust: digitrust,
+			},
+			expectedUserExt: openrtb_ext.ExtUser{
+				Consent:   validConsent,
+				DigiTrust: digitrust,
+			},
+		},
+		{
+			description: "Does Not Override Existing Consent If Empty",
+			consent:     "",
```
hhhjort: Should we have a test on the behavior if a new and an existing consent both exist?

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,357 +122,291 @@ func TestAMPPageInfo(t *testing.T) {
...
-
-	// Assert string `consent` is found in the User.Ext at all
-	assert.NotContainsf(t, fullMarshaledBidRequest, "consent:"+consentString, "Expected bid request to contain consent string %s \n", consentString)
-
-	// Assert the last request has a valid User object with a consent string equal to that on the URL query
-	var ue openrtb_ext.ExtUser
-	err = json.Unmarshal(exchange.lastRequest.User.Ext, &ue)
-	assert.NoError(t, err, "Error unmarshalling last processed request")
-
-	// Assert consent string found in `http.Request` was passed correctly to the `User.Ext` object
```
guscarreon: I see the following assertions were removed. Do you think they are worth checking?
```
    // Assert consent string found in `http.Request` was passed correctly to the `User.Ext` object
    assert.Contains(t, string(request.URL.RawQuery), consentString, "http.Request should come with a consent string in its query")
    assert.Equal(t, consentString, ue.Consent, "Consent string unsuccessfully passed to bid request through AMP endpoint")

    // Assert other user properties found originally in our bid request such as `DigiTrust` were not overwritten
    assert.Equal(t, DigiTurstID, ue.DigiTrust.ID, "Passing GDPR consent through endpoint should not override http.Request ExtUser fields other than consent")
amp_auction_test.go
```

endpoints/openrtb2/amp_auction.go
SyntaxNode: > We could maybe make privacy.ReadPoliciesFromConsent return an error instead of a bool so we can channel its message into our potential warning system:

I'm afraid that won't work in this case. `ReadPoliciesFromConsent` attempts to parse the consent string as both CCPA and TCF, since AMP doesn't tell us which one to expect. The detailed error messages are of little use if we aren't sure which format we're supposed to be targeting. The best I think we can do here is to fill in the TODO with a warning that the format was unrecognized. Other parts of code either currently use or may in the future use the message of the error.

privacy/ccpa/policy.go
```diff
@@ -49,35 +50,44 @@ func (p Policy) Write(req *openrtb.BidRequest) error {
...
+	if len(consent) != 4 {
+		return errors.New("must contain 4 characters")
+	}
+
+	if consent[0] != '1' {
+		return errors.New("must specify version 1")
 	}
 
 	var c byte
```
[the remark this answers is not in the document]
SyntaxNode: It's certainly possible. I don't think it looks any cleaner though and adding error message formatting is slightly more overhead (along with the loop counter). Do you think extracting the charset check of `c != 'N' && c != 'Y' && c != '-'` to a helper method would be useful?

endpoints/openrtb2/amp_auction_test.go
SyntaxNode: Those checks are (mostly) still present, just in different forms. The new tests assert the entire User.Ext object instead of breaking it down into individual assertions. That just leaves...

> assert.Contains(t, string(request.URL.RawQuery), consentString, "http.Request should come with a consent string in its query")

Which looks to me that it's asserting the test setup and is therefore not useful.

endpoints/openrtb2/amp_auction_test.go
SyntaxNode: Not a bad idea for documentation purposes. Added.

SyntaxNode:
I'll commit an update soon which implements the approach for exposing warnings that was discussed at the Prebid.org 2/21/2020 meeting.

errortypes/code.go
```diff
@@ -0,0 +1,29 @@
...
+	NoErrorCode = iota
+	TimeoutCode
+	BadInputCode
+	BlacklistedAppCode
+	BadServerResponseCode
+	FailedToRequestBidsCode
+	BidderTemporarilyDisabledCode
+	BlacklistedAcctCode
+	AcctRequiredCode
+	InvalidPrivacyConsentCode
```
SyntaxNode: `InvalidPrivacyConsentCode` was formerly the `WarningCode`. I think it's safe to remove since the `WarningCode` was never documented or written out in the response. Do you think it's ok to continue counting here and mix error and warning codes? Or maybe add, say, a 10,000 gap between errors and warnings? Or keep them entirely separate?

errortypes/code.go
```diff
@@ -0,0 +1,29 @@
...
+	UnknownErrorCode = 999
```
SyntaxNode: Should we also have an `UnknownWarningCode`? They're in their own collections so the same unknown number being used makes a bit of sense to me.

errortypes/code.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// Coder provides an error code.
+type Coder interface {
+	Code() int
+}
+
+// ReadErrorCode returns the error code, or UnknownErrorCode if unavailable.
+func ReadErrorCode(err error) int {
```
SyntaxNode: The rest of the file is unchanged, but copied out of errortypes.go.

errortypes/severitylevel.go
```diff
@@ -0,0 +1,45 @@
+package errortypes
```
SyntaxNode: A new way to determine error vs warning, so warnings can have their own codes without using a long hardcoded list of which codes are warning vs error (like we had before to a degree).

errortypes/severitylevel.go
```diff
@@ -0,0 +1,68 @@
...
+		if isFatal(err) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// FatalOnly returns a new error list with only the fatal severity errors.
+func FatalOnly(errs []error) []error {
```
SyntaxNode: Is this a good name? I was also considering `FilterFatalOnly`, `FilterFatal`,`FilterFatals`...

errortypes/code.go
hhhjort: I lean towards the gap idea to keep things simple and organized, but I am flexible on this idea.

errortypes/code.go
hhhjort: Makes sense, especially if we go with the gap idea

errortypes/code.go
SyntaxNode: I like that too. I added a gap. Error codes start at 1. Warning codes start at 10,001 (or would this better be 1,001?).

errortypes/code.go
SyntaxNode: I thought this would make sense, but it got messy when I tried to code it because `ReadErrorCode` is hardcoded to `UnknownErrorCode`. I would need to also add a detection of severity to that method to distinguish between known errors and warnings.. and at that point, do you think it's worth it? We don't call the function very often so perhaps the impact isn't bad. Thoughts?

errortypes/code.go
```diff
@@ -0,0 +1,36 @@
...
+type Coder interface {
+	Code() int
+}
+
+// ReadErrorCode returns the error or warning code, or UnknownCode if unavailable.
+func ReadErrorCode(err error) int {
+	if e, ok := err.(Coder); ok {
+		return e.Code()
+	}
+	return UnknownCode
```
SyntaxNode: I would need to add a Severity check here as well, or just combine the ErrorCode and SeverityLevel methods into a single CodedError interface.. which might honestly make sense.

hhhjort: I could see having a method that just returns a touple to deliver both values.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -122,357 +121,472 @@ func TestAMPPageInfo(t *testing.T) {
...
+			[]byte{},
+			openrtb_ext.BidderMap,
+		)
+
+		// Invoke Endpoint
+		request := httptest.NewRequest("GET", fmt.Sprintf("/openrtb2/auction/amp?tag_id=1&consent_string=%s", test.consent), nil)
+		responseRecorder := httptest.NewRecorder()
+		endpoint(responseRecorder, request, nil)
+
+		// Parse Resonse
```
guscarreon: Nitpick "Response"

privacy/ccpa/policy.go
guscarreon: I sort of disagree on this because I'd like to avoid an additional function on top of the stack at execution time and an additional function to maintain, but this doesn't diminish its functionality and the hard work behind its design.

PR #1188, opened 2020-02-10
Update required params for TheMediaGrid adapter

TheMediaGrid:
Update info about required params for TheMediaGrid adapter


static/bidder-params/grid.json
```diff
@@ -3,6 +3,11 @@
...
   "type": "object",
-  "properties": {},
-  "required": []
+  "properties": {
+    "uid": {
+      "type": "integer",
+      "description": "An ID which identifies this placement of the impression"
+    }
+  },
+  "required": ["uid"]
```
SyntaxNode: I'm stealing this warning from @hhhjort:

> Do you have any current traffic coming from PBS? When you set uid as required, if a request comes in missing uid once it is set to required, it will mark the whole request as failed and not send a bid request to any bidder. So adding a required field can break the entire prebid setup of a publisher.
> 
> This behavior is to help ensure publishers set up their configs correctly. Otherwise a bidder might set up a request with 5 bidders, but bidder ABC has a bad config and always errors out. If the other 4 are working normally, the publisher might not notice they have a bad config for quite a while. If some pages have a valid ABC config, they might just conclude that ABC just doesn't bid much if they don't troubleshoot carefully.
> 
> If there are publishers who currently have configs that omit uid. we would recommend that initially the adapter not advertise uid as required in bidder-params, but return an error from within the adapter if it is missing. An error from within the adapter will not invalid the whole request, allowing the other bidders to function. Then after all publishers have had a reasonable amount of time to update their configs, we can do the additional step of making it required in bidder-params.

SyntaxNode:
@TheMediaGrid Reminder that there is feedback on this PR to be addressed. Thank you.

static/bidder-params/grid.json
TheMediaGrid: Hi guys
Answering your question - yes, we have traffic from pbs now. Daily amount of incoming traffic is hundreds of millions of requests. All our partners who use pbs send us uid in requests. This is the only param we expect to see and it is not a new required filed then. Our goal was to update the doc so partners could see the required params in the right place.
I'm not sure that we interpret this statement correctly
'if a request comes in missing uid once it is set to required, it will mark the whole request as failed and not send a bid request to any bidder.'
, so we'd appreciate it if you could clarify it a bit.

hhhjort: One of the first things PBS does on receiving a request is to validate all the bidder parameters against the JSON schemas that have been set up. If any one of the bidder parameters for any bidder fails this validation, the entire request is marked bad and an error message is returned. This was done under the assumption that these definitions would rarely change, and it is easy to miss that a configuration has one bad bidder if the valid bidders are returning bids. Returning nothing will quickly flag the publisher that they have an incorrect configuration.

The recommended path forward here is to update your schema, but not flag it as required. Put a check in your adapter that will return an error if the uid is missing. An error from within the adapter code will not trigger PBS to fail the entire request. Then some time later, when publishers have had time to validate/update their configs, a second PR can make it required.

TheMediaGrid: Hello,
We made the changes you proposed. 
@hhhjort @guscarreon , review please our updates

adapters/grid/grid.go
```diff
@@ -15,11 +15,50 @@ type GridAdapter struct {
...
+	// pre-process the imps
+	for _, imp := range requestCopy.Imp {
+		if err := processImp(&imp); err == nil {
+			validImps = append(validImps, imp)
+		} else {
+			errors = append(errors, err)
+		}
+	}
+	if len(validImps) == 0 {
+		err := fmt.Errorf("No valid impressions for grid")
```
SyntaxNode: Please add a test (_test.go or json based) for this condition and verify the `uid is empty` input error.

TheMediaGrid: Hello, we made the necessary updates, please review

adapters/grid/grid.go
```diff
@@ -15,11 +15,53 @@ type GridAdapter struct {
...
+		err := &errortypes.BadInput{
+			Message: "uid is empty",
+		}
+		return err
+	}
+	// no error
+	return nil
+}
+
 // MakeRequests makes the HTTP requests which should be made to fetch bids.
```
guscarreon: Can we add a couple more tests in `supplemental` to cover the if statements shown in red below?
image

Issue #1185, opened 2020-02-09
VideoBidRequest user.Ext does  not copy consent to OpenRtb.user.ext


neririchter:
It seems that We don't pass GDPR consent from the videoBidRequest to openRtb request

 In file uilts.go inside  function "cleanOpenRTBRequests" it tries to extract the GDPR consent from function "extractConsent" inside file gdpr.go while it tried to pull the consent from the user.ext of the open rtb request

but when you look at function mergeData at the video_auction class we do not construct a user.ext in the openrtb bid request.

SyntaxNode:
@camrice @jmaynardxandr Can you please take a look?

camrice:
There will be a PR for this today. This will update it so that in the video request, if `User.GDPR.ConsentRequired` is set to `true`, then the value of `User.GDPR.ConsentString` is copied to the OpenRTB Bid Request under `User.Ext.Consent`.

camrice:
PR for this issue: https://github.com/prebid/prebid-server/pull/1189

SyntaxNode:
PR is under review. Removed stale label.

SyntaxNode:
PR is still under review. We waited to update the docs and verifying the fix will not break current usages.

SyntaxNode:
The PR has been merged and released on the AppNexus servers. Does that resolve the issue for you, @neririchter?

neririchter:
Yes, thanks!

PR #1189, opened 2020-02-10
Adding copying of gdpr consent string to openrtb bid request

camrice:
addresses https://github.com/prebid/prebid-server/issues/1185

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -478,6 +478,11 @@ func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.Bi
 			Gender:   videoRequest.User.Gender,
 			Keywords: videoRequest.User.Keywords,
 		}
+		if videoRequest.User.Gdpr.ConsentRequired {
+			consentString := fmt.Sprintf(`{"consent":"%s"}`, videoRequest.User.Gdpr.ConsentString)
+			userExt := json.RawMessage([]byte(consentString))
+			bidRequest.User.Ext = userExt
+		}
 	}
 
 	if len(videoRequest.BCat) != 0 {
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -298,7 +298,7 @@ func (deps *endpointDeps) createImpressions(videoReq *openrtb_ext.BidRequestVide
 
 		impsArray := make([]openrtb.Imp, numImps)
 		for impInd := range impsArray {
-			newImp := createImpressionTemplate(storedImp, videoData)
+			newImp := createImpressionTemplate(storedImp, *videoData)
 			impsArray[impInd] = newImp
 			if reqExactDur {
 				//floor := int(math.Floor(ind/impDivNumber))
@@ -328,12 +328,12 @@ func max(a, b int) int {
 	return b
 }
 
-func createImpressionTemplate(imp openrtb.Imp, video openrtb_ext.SimplifiedVideo) openrtb.Imp {
+func createImpressionTemplate(imp openrtb.Imp, video openrtb.Video) openrtb.Imp {
 	imp.Video = &openrtb.Video{}
 	imp.Video.W = video.W
 	imp.Video.H = video.H
 	imp.Video.Protocols = video.Protocols
-	imp.Video.MIMEs = video.Mimes
+	imp.Video.MIMEs = video.MIMEs
 	return imp
 }
 
@@ -472,17 +472,7 @@ func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.Bi
 	}
 
 	if &videoRequest.User != nil {
-		bidRequest.User = &openrtb.User{
-			BuyerUID: videoRequest.User.Buyeruids["appnexus"], //TODO: map to string merging
-			Yob:      videoRequest.User.Yob,
-			Gender:   videoRequest.User.Gender,
-			Keywords: videoRequest.User.Keywords,
-		}
-		if videoRequest.User.Gdpr.ConsentRequired {
-			consentString := fmt.Sprintf(`{"consent":"%s"}`, videoRequest.User.Gdpr.ConsentString)
-			userExt := json.RawMessage([]byte(consentString))
-			bidRequest.User.Ext = userExt
-		}
+		bidRequest.User = videoRequest.User
 	}
 
 	if len(videoRequest.BCat) != 0 {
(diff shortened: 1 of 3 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -478,6 +478,11 @@ func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.Bi
 			Gender:   videoRequest.User.Gender,
 			Keywords: videoRequest.User.Keywords,
 		}
+		if videoRequest.User.Gdpr.ConsentRequired {
```
SyntaxNode: What does `User.Gdpr.ConsentRequired` mean?

There is a boolean GDPR setting in the IAB spec at `Regs.ext.gdpr` which in which indicates if the request is subject to GDPR regulations. If that's the same meaning here, consider mapping `User.Gdpr.ConsentRequired` to `Regs.ext.gdpr` and `videoRequest.User.Gdpr.ConsentString` to `User.ext.consent`.

camrice: Updated per our discussion. The video request has been updated to use OpenRTB objects instead of custom ones for User and Video. The User here will instead by copied directly from the request. Handling of `User.Ext.Consent` will be considered downstream.

camrice:
Further discussion regarding this PR will occur tomorrow. Please do not merge or review yet.

camrice:
This change is good to review now. Awaiting confirmation of documentation changes.

SyntaxNode:
This PR introduces a breaking change to callers of the video endpoint. We are aware of anyone using this right now and feel it's safe to make this change. By all means, please speak up if you are actively using this endpoint and will be affected.

endpoints/openrtb2/video_auction.go
```diff
@@ -298,7 +298,7 @@ func (deps *endpointDeps) createImpressions(videoReq *openrtb_ext.BidRequestVide
 
 		impsArray := make([]openrtb.Imp, numImps)
 		for impInd := range impsArray {
-			newImp := createImpressionTemplate(storedImp, videoData)
+			newImp := createImpressionTemplate(storedImp, *videoData)
```
SyntaxNode: Would it be better just pass down the pointer?

endpoints/openrtb2/video_auction.go
```diff
@@ -472,12 +472,7 @@ func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.Bi
 	}
 
 	if &videoRequest.User != nil {
-		bidRequest.User = &openrtb.User{
-			BuyerUID: videoRequest.User.Buyeruids["appnexus"], //TODO: map to string merging
-			Yob:      videoRequest.User.Yob,
-			Gender:   videoRequest.User.Gender,
-			Keywords: videoRequest.User.Keywords,
-		}
+		bidRequest.User = videoRequest.User
```
SyntaxNode: Is the `nil` check necessary since it's now just wrapping an assignment?

endpoints/openrtb2/video_auction.go
```diff
@@ -659,27 +654,32 @@ func (deps *endpointDeps) validateVideoRequest(req *openrtb_ext.BidRequestVideo)
...
-		mimes := make([]string, 0, 0)
-		for _, mime := range req.Video.Mimes {
-			if mime != "" {
-				mimes = append(mimes, mime)
+	if req.Video != nil {
+		if len(req.Video.MIMEs) == 0 {
+			err := errors.New("request missing required field: Video.Mimes")
+			errL = append(errL, err)
+		} else {
+			mimes := make([]string, 0, 0)
```
SyntaxNode: I know this is copied code, but consider initializing to the size of `req.Video.MIMEs`.

endpoints/openrtb2/video_auction.go
```diff
@@ -659,27 +654,32 @@ func (deps *endpointDeps) validateVideoRequest(req *openrtb_ext.BidRequestVideo)
...
+			for _, mime := range req.Video.MIMEs {
+				if mime != "" {
+					mimes = append(mimes, mime)
+				}
+			}
+			if len(mimes) == 0 {
+				err := errors.New("request missing required field: Video.Mimes, mime types contains empty strings only")
+				errL = append(errL, err)
+			}
+			if len(mimes) > 0 {
```
SyntaxNode: This check doesn't seem necessary. I don't the harm of an assignment of `req.Video.MIMEs` to an empty slice.

endpoints/openrtb2/video_auction_test.go
```diff
@@ -484,6 +484,58 @@ func TestVideoEndpointValidationsSiteMissingRequiredField(t *testing.T) {
...
+func TestVideoEndpointValidationsMissingVideo(t *testing.T) {
+	ex := &mockExchangeVideo{}
+	deps := mockDeps(t, ex)
+	deps.cfg.VideoStoredRequestRequired = true
+
+	durationRange := make([]int, 0)
+	durationRange = append(durationRange, 15)
+	durationRange = append(durationRange, 30)
+
+	pods := make([]openrtb_ext.Pod, 0)
```
SyntaxNode: `pod1` and `pod2` are not used beyond creation of the slice. Consider replacing lines 496-508 with a literal initialization.

.. and actually, same goes for `durationRange`, `mimes`, and `videoProtocols`.

endpoints/openrtb2/video_auction.go
camrice: Updated to use a pointer argument

endpoints/openrtb2/video_auction.go
camrice: Removed nil check, it was not needed

endpoints/openrtb2/video_auction_test.go
camrice: Updated to be literal initializations. Removed `mimes` and `videoProtocols` as those weren't needed either.

endpoints/openrtb2/video_auction.go
camrice: After playing around with this section, I think it may still be necessary to initialize with an empty slice. This check here filters out empty strings from `MIMEs`, so it is a different error than having no `MIMEs` in the request. Let me know if there is a better way to do this however.

endpoints/openrtb2/video_auction.go
camrice: Removed check here. The error of having all empty strings will still be caught regardless of if the value is copied over empty.

endpoints/openrtb2/video_auction.go
```diff
@@ -471,14 +471,7 @@ func mergeData(videoRequest *openrtb_ext.BidRequestVideo, bidRequest *openrtb.Bi
...
 
-	if &videoRequest.User != nil {
-		bidRequest.User = &openrtb.User{
-			BuyerUID: videoRequest.User.Buyeruids["appnexus"], //TODO: map to string merging
-			Yob:      videoRequest.User.Yob,
-			Gender:   videoRequest.User.Gender,
-			Keywords: videoRequest.User.Keywords,
-		}
-	}
+	bidRequest.User = videoRequest.User
```
SyntaxNode: @hhhjort Is a direct assignment here good with you or do you recommend a shallow copy?

endpoints/openrtb2/video_auction.go
SyntaxNode: The slice would still be empty, but setting an initial capacity to `len(req.Video.MIMEs)` is more efficient if there is a high chance that all MIMEs are good. If you expect most of the MIME values to be empty, then leaving it this way is fine.

endpoints/openrtb2/video_auction_test.go
```diff
@@ -484,6 +484,50 @@ func TestVideoEndpointValidationsSiteMissingRequiredField(t *testing.T) {
...
+			PodId:            2,
+			AdPodDurationSec: 30,
+			ConfigId:         "qwerty",
+		},
+	}
+
+	req := openrtb_ext.BidRequestVideo{
+		StoredRequestId: "123",
+		PodConfig: openrtb_ext.PodConfig{
+			DurationRangeSec:     durationRange,
```
SyntaxNode: Nitpick: Last minor nitpick here. Since durationRange is only used to construct the config, consider declaring it inline like:
```
PodConfig: openrtb_ext.PodConfig{
			DurationRangeSec:     []int{15,30},
```

endpoints/openrtb2/video_auction.go
hhhjort: That should be fine. Nothing should be touching the `User` through the `videoRequest` pointer.

endpoints/openrtb2/video_auction_test.go
camrice: Updated to be inline. Also moved pods declaration to be inline for consistency.

endpoints/openrtb2/video_auction.go
camrice: Ah I just realized I set the initial length instead of the initial capacity when I originally tried to change this. You're right, working as expected. Updated so that initial cap is `len(req.Video.MIMEs)`.

PR #1195, opened 2020-02-17
add admixer adapter

DmitryStashkevich:


DmitryStashkevich:
Hello team,
guscarreon, thank you very much for such detailed code review. I've added some additional test-cases to make test coverage more than 90% and fixed issues

adapters/admixer/admixer_test.go
```diff
@@ -0,0 +1,10 @@
+package admixer
+
+import (
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"testing"
+)
+
+func TestJsonSamples(t *testing.T) {
+	adapterstest.RunJSONBidderTest(t, "admixertest", NewAdmixerBidder("http://inv-nets.admixer.net/pbs.aspx"))
```
SyntaxNode: The url does not need to be real here. It just needs to match the json examples. Consider using an obviously fake stub so you don't have to worry about this address becoming outdated.

PR #1200, opened 2020-02-22
API doc: added Rewarded Video section

bretg:
also edited all examples so they include the full openRTB context


SyntaxNode:
This is waiting on a merge conflict fix.

PR #1201, opened 2020-02-25
updating default hard-coded list of certs

PubMatic-OpenWrap:


SyntaxNode:
Which version of Ubuntu is the list from? I'd like to independently compare before merging in.

PubMatic-OpenWrap:
@SyntaxNode , the certs are from Ubuntu 18.04 (I took the this OS version from prebid-server's Dockerfile)

hhhjort:
It wouldn't surprise me ... probably getting pretty old by now, I don't recall when they were last updated.

PR #1204, opened 2020-02-26
Add ValueImpression Adapter

thuyhq:
Opening pull request to add a prebid server adapter for Valueimpression


adapters/valueimpression/usersync.go
```diff
@@ -0,0 +1,13 @@
+package valueimpression
+
+import (
+	"text/template"
+
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/usersync"
+)
+
+//NewValueImpressionSyncer :
```
SyntaxNode: Please provide an proper description or omit the comment.

adapters/valueimpression/usersync_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"testing"
+	"text/template"
+
+	"github.com/prebid/prebid-server/privacy"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValueImpressionSyncer(t *testing.T) {
+	syncURL := "https://rtb.valueimpression.com/usersync?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redirectUri=http%3A%2F%2Flocalhost:8000%2Fsetuid%3Fbidder%3Dvalueimpression%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26us_privacy%3D{{.USPrivacy}}%26uid%3D%24UID"
```
SyntaxNode: Please omit `us_privacy` from the direct url. Prebid Server does not use that value during a callback. It will, instead,  never request a cookie sync if CCPA is enforced and opted out.

adapters/valueimpression/usersync_test.go
```diff
@@ -0,0 +1,25 @@
...
+	syncURLTemplate := template.Must(
+		template.New("sync-template").Parse(syncURL),
+	)
+
+	syncer := NewValueImpressionSyncer(syncURLTemplate)
+	syncInfo, err := syncer.GetUsersyncInfo(privacy.Policies{})
```
SyntaxNode: Consider providing simple privacy policy values to test the macro works correctly.

adapters/valueimpression/valueimpression.go
```diff
@@ -0,0 +1,161 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type Adapter struct {
```
SyntaxNode: Please rename to `ValueImpressionAdapter` to follow adapter naming conventions.

adapters/valueimpression/valueimpression.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return nil
+}
+
+// MakeBids based on valueimpression server response
+func (a *Adapter) MakeBids(bidRequest *openrtb.BidRequest, unused *adapters.RequestData, responseData *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if responseData.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
```
SyntaxNode: It's also common to check for a `StatusBadRequest` return type and map it to a `errortypes.BadInput`.

adapters/valueimpression/valueimpression.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+
+		imp.Ext = bidderExt.Bidder
+	}
+
+	return nil
+}
+
+func validateImp(imp *openrtb.Imp) error {
```
hhhjort: This should not be necessary, PBS core will filter out types that you adapter does not support according to `static/bidder-info/valueimpression.yaml`. It doesn't hurt to have it of course.

PR #1209, opened 2020-03-02
Districtm Dmx: new adapter 

stevealliance:
DistrictM new Adapter support for prebid server.
This adapter only support mediatype banner for website and app.


adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+	var dmxImp dmxBidder
+	var imps []openrtb.Imp
+	//var userParams *dmxUser
+	if err := json.Unmarshal(request.Imp[0].Ext, &dmxImp); err != nil {
+		errs = append(errs, err)
+	}
+
+	//fmt.Println(request.User)
+	if request.User != nil {
+		if request.User.BuyerUID != "" {
```
SyntaxNode: I'm curious about this case. Is it valid / common for the `request.User.ID` to be provided but the `request.User.BuyerUID` to be empty?

usersync/usersyncers/syncer_test.go
```diff
@@ -27,6 +27,7 @@ func TestNewSyncerMap(t *testing.T) {
 			string(openrtb_ext.BidderConsumable):       syncConfig,
 			string(openrtb_ext.BidderConversant):       syncConfig,
 			string(openrtb_ext.BidderDatablocks):       syncConfig,
+			string(openrtb_ext.BidderDmx):       syncConfig,
```
SyntaxNode: Please run gofmt on all submissions.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+
+type dmxSize struct {
+	W uint64
+	H uint64
+	S uint64
+}
+
+type DmxSize []dmxSize
+
+func (a DmxSize) Len() int {
```
SyntaxNode: Consider adding comments for Len, Swap, and Less to indicate that they're implemented to support sorting.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+	Banner *openrtb.Banner `json:"banner"`
+}
+
+type dmxSize struct {
+	W uint64
+	H uint64
+	S uint64
+}
+
+type DmxSize []dmxSize
```
SyntaxNode: Consider naming this `DmxSizes` to indicate it's a collection rather than a single unit.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+
+func NewDmxBidder(endpoint string, publisher_id string) *DmxAdapter {
+	return &DmxAdapter{endpoint: endpoint, publisherId: publisher_id}
+}
+
+type dmxBidder struct {
+	Bidder dmxExt `json:"bidder"`
+}
+
+type dmxUser struct {
```
SyntaxNode: This looks unused.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+	User *openrtb.User `json:"user"`
+}
+
+type dmxExt struct {
+	PublisherId string `json:"publisher_id,omitempty"`
+}
+
+type dmxBanner struct {
```
SyntaxNode: This looks unused.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+	Banner *openrtb.Banner `json:"banner"`
+}
+
+type dmxSize struct {
+	W uint64
+	H uint64
+	S uint64
```
SyntaxNode: I'm curious, what does S refer to?

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+	{300, 50, 80},
+	{970, 250, 75},
+	{970, 90, 70},
+}
+
+func (adapter *DmxAdapter) MakeRequests(request *openrtb.BidRequest, req *adapters.ExtraRequestInfo) (reqsBidder []*adapters.RequestData, errs []error) {
+	var dmxImp dmxBidder
+	var imps []openrtb.Imp
+	//var userParams *dmxUser
+	if err := json.Unmarshal(request.Imp[0].Ext, &dmxImp); err != nil {
```
SyntaxNode: Please verify there is at least 1 Imp in the collection before indexing it via `request.Imp[0].Ext`.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+
+		}
+	}
+
+	for _, inst := range request.Imp {
+		var banner openrtb.Banner
+		var ins openrtb.Imp
+		//for _, insbanner := range inst.Banner.Format {
+		banner = openrtb.Banner{
+			W:      &inst.Banner.Format[0].W,
```
SyntaxNode: Please verify there is at least 1 Format in the collection before indexing it via `inst.Banner.Format[0]`.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+			H:      &inst.Banner.Format[0].H,
+			Format: inst.Banner.Format,
+		}
+		nSize := Remove(inst.Banner.Format, CheckTopSizes)
+		sort.Sort(DmxSize(nSize))
+		if inst.Banner.Format[0].W != 0 {
```
SyntaxNode: Is this the correct logic? If the banner format Width is specified, overwrite it with nSize[0].W?

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+	headers := http.Header{}
+	headers.Add("Content-Type", "Application/json;charset=utf-8")
+	reqBidder := &adapters.RequestData{
+		Method:  "POST",
+		Uri:     adapter.endpoint, //adapter.endpoint,
+		Body:    oJson,
+		Headers: headers,
+	}
+
+	if request.User == nil {
```
SyntaxNode: Consider simplifying the case here with a single if clause.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+
+	// This shouldnt happen. Lets handle it just incase by returning an error.
+	return "", &errortypes.BadInput{
+		Message: fmt.Sprintf("Failed to find impression \"%s\" ", impID),
+	}
+}
+
+func getCookieInfo(request *pbs.PBSRequest) {
```
SyntaxNode: This appears to be unused.

adapters/dmx/usersync_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"testing"
+	"text/template"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDmxSyncer(t *testing.T) {
+	temp := template.Must(template.New("sync-template").Parse("https://cdn.districtm.io/ids/?sellerid=10007"))
```
SyntaxNode: We've updated the template for this test. Did you get this old style from a doc we forgot to update?

config/config.go
```diff
@@ -233,6 +233,9 @@ type Adapter struct {
 	PlatformID string `mapstructure:"platform_id"`
 	AppID      string `mapstructure:"app_id"`
 	AppSecret  string `mapstructure:"app_secret"`
+
+	// needed for district m DMX
+	PublisherId string `mapstructure:"publisher_id"`
```
SyntaxNode: Please consider using `ExtraAdapterInfo` for bidder specific configs. We added that recently to avoid many one-off  config entries.

config/config.go
```diff
@@ -486,6 +489,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConsumable, "https://e.serverbid.com/udb/9969/match?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconsumable%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConversant, "https://prebid-match.dotomi.com/prebid/match?rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconversant%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderDatablocks, "https://search.v5prebid.datablocks.net/s2ssync?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&r="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Ddatablocks%26gdpr%3D%24%7Bgdpr%7D%26gdpr_consent%3D%24%7Bgdpr_consent%7D%26uid%3D%24%7Buid%7D")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderDmx, "https://cdn.districtm.io/ids/?sellerid="+cfg.Adapters[string(openrtb_ext.BidderDmx)].PublisherId+"&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}")
```
SyntaxNode: Do you need a callback / redirect url to complete the user sync?

static/bidder-info/dmx.yaml
```diff
@@ -0,0 +1,9 @@
+maintainer:
+  email: "steve.alliance@gmail.com"
```
SyntaxNode: Curious. Do you work for DistrictM?

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/pbs"
+	"net/http"
+	"sort"
+)
+
+type DmxAdapter struct {
```
SyntaxNode: Please add unit and json tests for the DmxAdapter.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (a DmxSize) Less(i, j int) bool {
+	return a[i].S > a[j].S
+}
+
+func Remove(toBeRemove []openrtb.Format, a DmxSize) (dmx DmxSize) {
+	for _, value := range toBeRemove {
+		for _, dmxValue := range a {
+			if dmxValue.H == value.H && dmxValue.W == value.W {
+				//fmt.Println("true")
```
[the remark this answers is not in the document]
stevealliance: Removing all code commented

adapters/dmx/dmx.go
stevealliance: @SyntaxNode you are right I was reading our internal docs the wrong way since I work more on the Front end Remove that part.

SyntaxNode:
Friendly ping to keep the PR open. I saw you responded to some of the feedback so I assume you're working on this when time allows. It will auto close if you can't circle back to it right now, but that's ok. We can always open it up again when you do.

config/config.go
hhhjort: Yes, best to avoid one off additions to adapter configs.

config/config.go
```diff
@@ -486,6 +489,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConsumable, "https://e.serverbid.com/udb/9969/match?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconsumable%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConversant, "https://prebid-match.dotomi.com/prebid/match?rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconversant%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderDatablocks, "https://search.v5prebid.datablocks.net/s2ssync?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&r="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Ddatablocks%26gdpr%3D%24%7Bgdpr%7D%26gdpr_consent%3D%24%7Bgdpr_consent%7D%26uid%3D%24%7Buid%7D")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderDmx, "https://dmx.districtm.io/s/v1/img/s/"+cfg.Adapters[string(openrtb_ext.BidderDmx)].SellerID+"&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}")
```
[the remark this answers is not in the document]
stevealliance: Each publisher need to have a publisher ID with us and need to be added into our database.
Once it's done we take their URL and add it to our system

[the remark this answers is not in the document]
stevealliance: Hi @SyntaxNode, do you know how can I have the publisher id pass into config/config.go 

`	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderDmx, "https://dmx.districtm.io/s/v1/img/s/"+cfg.Adapters[string(openrtb_ext.BidderDmx)].ExtraAdapterInfo+"&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}")
`

I've look at the tripleliftNative adapter, but I really need this value to be global or part of the publisher setting in prebid. The goal is to send the id sync link that belong to all specific seller/publisher.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"net/http"
+	"net/url"
+)
+
+type DmxAdapter struct {
+	endpoint    string
+	publisherId string
```
hhhjort, line 17: I don't see `publisherId` in this struct being used for anything? It should not belong there anyway, since there will be only one instance of the `DmxAdapter` across all requests, and I imagine that `publisherId` is more dynamic than that.

adapters/dmx/dmx_test.go
```diff
@@ -0,0 +1,449 @@
+package dmx
```
hhhjort, line 1: In the future, you should look at the json test machinery that has been developed for adapters. It does improve test maintainability and readability. But may not be worth the time of converting all these hard coded tests over. But do look at creating`dmxtest/params/race/banner.json`, this would allow us to include your adapter in the race condition tests.

config/config.go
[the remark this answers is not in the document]
hhhjort: The design of PBS is that a user sync ID would be global for the PBS host. We do not support the ability to say that `dmxID=123` for publisher A and `dmxID=456` for publisher B. Do you really have a use case for having multiple IDs for the same user?

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+
+	}
+
+	dmxReq.Imp = imps
+
+	oJson, err := json.Marshal(request)
+
+	if err != nil {
+		errs = append(errs, err)
```
SyntaxNode, line 123: You should exit early here. Right now, you'll be posting an empty string if the json marshalling fails.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,259 @@
...
+	headers := http.Header{}
+	headers.Add("Content-Type", "Application/json;charset=utf-8")
+	reqBidder := &adapters.RequestData{
+		Method:  "POST",
+		Uri:     adapter.endpoint + addParams(sellerId), //adapter.endpoint,
+		Body:    oJson,
+		Headers: headers,
+	}
+
+	if dmxReq.User == nil {
```
SyntaxNode, line 134: Consider moving this logic earlier in the method. A lot of work can be avoided if this is detected earlier.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,259 @@
...
+			}
+
+			const intVal int8 = 1
+			var params dmxExt
+
+			source := (*json.RawMessage)(&inst.Ext)
+			if err := json.Unmarshal(*source, &params); err != nil {
+				errs = append(errs, err)
+			}
+			if params.Bidder.PublisherId == "" && params.Bidder.MemberId == "" {
```
SyntaxNode, line 103: To be safe, please verify `params.Bidder` is not null before accessing `PublisherId` and `MemberId`.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,259 @@
...
+	if params.Bidder.DmxId != "" {
+		ins = openrtb.Imp{
+			ID:     inst.ID,
+			TagID:  params.Bidder.DmxId,
+			Banner: &banner,
+			Ext:    inst.Ext,
+			Secure: &intVal,
+		}
+	}
+	imps = append(imps, ins)
```
SyntaxNode, line 233: What if `ins` is nil? It would be if both the `TagId` and `DmxId` are empty. An nil item in an `Imps` slice can have unpredictable behavior.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,259 @@
...
+	if params.DmxId != "" {
+		ins = openrtb.Imp{
+			ID:     inst.ID,
+			TagID:  params.DmxId,
+			Banner: &banner,
+			Ext:    inst.Ext,
+			Secure: &intVal,
+		}
+	}
+	imps = append(imps, ins)
```
SyntaxNode, line 209: What if `ins` is nil? It would be if both the `TagId` and `DmxId` are empty. An nil item in an `Imps` slice can have unpredictable behavior.

config/config.go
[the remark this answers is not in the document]
SyntaxNode: @stevealliance Please address this feedback. I don't think Prebid Server will behave the way you intend in this case.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,259 @@
...
+		errs = append(errs, err)
+	}
+
+	if dmxBidderStaticInfo.Bidder.PublisherId != "" && publisherId == "" {
+		publisherId = dmxBidderStaticInfo.Bidder.PublisherId
+	}
+
+	if dmxReq.App != nil {
+		dmxReq.Site = nil
+		dmxReq.App.Publisher.ID = publisherId
```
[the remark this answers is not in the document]
hhhjort: Good catch, `Site` is safe as it is part of the shallow request copy. `dmxReq.App` is either `nil` or a pointer to the shared `App` object. `dmxReq.App.Publisher` (if `App` is not nil) again is either `nil` or a pointer to a shared object, and should also be copied. And of course `nil` checks will be needed for all the `nil`s I just highlighted.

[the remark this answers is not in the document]
stevealliance: Check was added

adapters/dmx/dmx.go
stevealliance: Check to append only when tagid or dmxid is present

stevealliance:
@SyntaxNode please review when you can!

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+	}
+
+	if dmxReq.App != nil {
+		dmxReq.Site = nil
+		if dmxReq.App.ID != "" {
+			hasId = append(hasId, true)
+		}
+		if dmxReq.App.Publisher != nil {
+			dmxReq.App.Publisher.ID = publisherId
```
SyntaxNode, line 92: The App object is shared memory among other adapters. If you want to modify it, you'll need to create a copy first. Something like...

```
appCopy := *bidRequest.App
bidRequest.App = &appCopy
bidRequest.App.Publisher = nil
```

So.. things like this is fine:
```
dmxReq.Site = nil
```
and this is not (without making a copy first)
```
dmxReq.Site.Publisher = &openrtb.Publisher{ID: publisherId}
```

In this PR, this applies to dmxReq.App, dmxReq.Site, and dmxReq.User.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,297 @@
...
+		if dmxReq.User.Ext != nil {
+			if err := json.Unmarshal(dmxReq.User.Ext, &user_ext); err == nil {
+				if len(user_ext.Eids) > 0 || user_ext.Digitrust.ID != "" {
+					hasId = append(hasId, true)
+				}
+			}
+		}
+	}
+
+	if len(hasId) == 0 {
```
SyntaxNode, line 115: If this is all `hasId` is being used for, please replace it with an `anyHasId` bool (or something like that) and just set it to true if you find any of the ids.

config/config.go
```diff
@@ -486,6 +486,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConsumable, "https://e.serverbid.com/udb/9969/match?gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconsumable%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderConversant, "https://prebid-match.dotomi.com/prebid/match?rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dconversant%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderDatablocks, "https://search.v5prebid.datablocks.net/s2ssync?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&r="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Ddatablocks%26gdpr%3D%24%7Bgdpr%7D%26gdpr_consent%3D%24%7Bgdpr_consent%7D%26uid%3D%24%7Buid%7D")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderDmx, "https://dmx.districtm.io/s/v1/img/s/10007?r="+url.QueryEscape(externalURL)+"&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}")
```
SyntaxNode, line 489: I just want to verify that 10007 is not specific to a single host, such as AppNexus/Xandr.

SyntaxNode, line 489: You need to add the path of the return url with the correct bidder name. Please look to other adapters for examples.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,297 @@
...
+	var publisherId string
+	var sellerId string
+	var user_ext dmxUserExt
+	var hasId []bool
+
+	var dmxReq = request
+
+	if dmxReq.User == nil {
+		if dmxReq.App == nil {
+			return nil, []error{errors.New("no user Id found and AppID, no request to DMX")}
```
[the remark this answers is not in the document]
stevealliance: Ok thanks for this information!

adapters/dmx/dmx.go
stevealliance: Something like this would work 

```go
if request.App != nil {
	appCopy := *request.App
	dmxReq.App = &appCopy
} else {
	dmxReq.App = nil
}
```

config/config.go
stevealliance: @SyntaxNode 
No this id represent the channel where the seller come from!

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,333 @@
...
+		dmxReq.User = nil
+	}
+
+	if dmxReq.App != nil {
+		dmxReq.Site = nil
+		if dmxReq.App.ID != "" {
+			anyHasId = true
+		}
+		if dmxReq.App.Publisher != nil {
+			dmxReq.App.Publisher.ID = publisherId
```
SyntaxNode, line 115: Thank you addressing the memory sharing issue with `.App` and `.Site`. `Publisher` is also a pointer, so it needs to be copied here before changing the ID.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,333 @@
...
+		} else {
+			dmxReq.App.Publisher = &openrtb.Publisher{ID: publisherId}
```
SyntaxNode, line 117: This is fine though, since it replaces the pointer.

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,333 @@
...
+func isDmxParams(t interface{}) bool {
+	switch t.(type) {
+	case dmxParams:
+		return true
+	default:
+		return false
+	}
+}
+
+func CloneValue(source interface{}, destin interface{}) {
```
SyntaxNode, line 322: I'm not exactly sure what this code is doing. It looks like it makes a shallow copy since it doesn't appear to be looping over the members (unless the reflect packages magically handle that?). The request object passed in to this method is already a shallow copy, so this likely isn't necessary. The code you added to copy App and Site structures before mutating them solves the shared memory problem in the recommended manner.

config/config.go
SyntaxNode: Great. Thank you. We've seen some adapters include host specific values in the past, so just wanted to make sure.

adapters/dmx/dmx.go
stevealliance: @SyntaxNode this was resolved

adapters/dmx/dmx.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+
+	if dmxReq.User != nil {
+		if dmxReq.User.ID != "" {
+			anyHasId = true
+		}
+		if dmxReq.User.Ext != nil {
+			if err := json.Unmarshal(dmxReq.User.Ext, &user_ext); err == nil {
+				if len(user_ext.Eids) > 0 || user_ext.Digitrust.ID != "" {
```
SyntaxNode, line 128: Digitrust is a reference. Please add a nil check before checking the id:
`if len(user_ext.Eids) > 0 || (user_ext.Digitrust != nil && user_ext.Digitrust.ID != "")) {`

stevealliance: @SyntaxNode  update that portion of the code

SyntaxNode: Why did you use '&user_ext.Digitrust' instead of 'user_ext.Digitrust'?

stevealliance: @SyntaxNode propose code gave me an error or warning 
```ssh
cannot convert nil to type dmxDigitrust
```

adapters/dmx/dmx.go
hhhjort: Yes, this does seem odd to have since you use it only once, and you know the type of the source/destination when you use it. What does this give you over a simple:
`reqcopy := *request`
`dmxReq = &reqcopy`
And do you need `dmxReq` to be a pointer rather than an actual struct?

adapters/dmx/dmx.go
stevealliance: @SyntaxNode just revert my code back, I've test the empty ext and I still don't get any error on my side . I believe we are good now

hhhjort: That is odd, `user_ext.Digitrust` is a pointer, and it should give an error for looking up `ID` if it is a nil pointer. Then again, I don't 100% understand what `CloneValue()` is doing, but if it is populating all empty pointers with pointers to structs for some reason, that is bad on resource usage. We don't want to do deep copies if we don't need to, and deep copies + allocating structs for otherwise empty pointers is worse.

stevealliance: I remove the deep copie function and ran some test to make sure everything is still working.
@hhhjort hope my modifications will be enough so we can be added to prebid server.

Thanks

SyntaxNode: Ah, ok. I see what's going on here. You're defining your own UserExt structure. We have this one already defined in the codebase which uses a pointer for the DigiTrust structure.
https://github.com/prebid/prebid-server/blob/ce94bf436c4853c150084031fbe0479a83bddc6b/openrtb_ext/user.go#L6

stevealliance: @hhhjort code was updated please review when you can

adapters/dmx/dmx.go
```diff
@@ -41,40 +40,21 @@ func UserSellerOrPubId(str1, str2 string) string {
...
 func (adapter *DmxAdapter) MakeRequests(request *openrtb.BidRequest, req *adapters.ExtraRequestInfo) (reqsBidder []*adapters.RequestData, errs []error) {
 	var imps []openrtb.Imp
 	var rootExtInfo dmxExt
 	var publisherId string
 	var sellerId string
-	var user_ext dmxUserExt
+	var userExt openrtb_ext.ExtUser
 	var anyHasId = false
-	var dmxReq *openrtb.BidRequest = &openrtb.BidRequest{}
+	var dmxReq *openrtb.BidRequest = request
```
hhhjort, line 50: This will create a new pointer pointing the the same struct that `request` is pointing to. I don't think this is what you want, rather you want something like `var dmxReq openrtb.BidRequest = *request`, or if you need dmxReq to definitely be a pointer:
`var reqCopy openrtb.BidRequest = *request`
`var dmxReq *openrtb.BidRequest = &reqCopy`

stevealliance: @hhhjort  thanks for the proposition

hhhjort:
You still have some merge conflicts remaining. It looks like you have resolved the conflicts between the master branch and your branch of your cloned repository, but your cloned master is currently not up to date with our master, and thus merge conflicts with our master remain. You should probably update your master branch, and then resolve the conflicts again.

stevealliance:
> You still have some merge conflicts remaining. It looks like you have resolved the conflicts between the master branch and your branch of your cloned repository, but your cloned master is currently not up to date with our master, and thus merge conflicts with our master remain. You should probably update your master branch, and then resolve the conflicts again.

When I try creating another PR no conflict between prebid-server master and my branch --- merging did work but added this warning @hhhjort what will be the next step?

hhhjort:
Apparently one piece was not updated properly to be in line with the new changes.

You can get this output by running `./validate.sh`

github.com/prebid/prebid-server/adapters/dmx [github.com/prebid/prebid-server/adapters/dmx.test]
adapters/dmx/usersync_test.go:13:41: too many arguments in call to syncer.GetUsersyncInfo
	have (string, string)
	want (privacy.Policies)

stevealliance:
@hhhjort thanks for making me realise that my local was still using some old code and was not using the new privacy struct. Can you have a look?

PR #1210, opened 2020-03-04
Kidoz adapter

rhaksi-kidoz:
```
ok  	github.com/prebid/prebid-server/adapters/kidoz	0.480s	coverage: 100.0% of statements
Success: Tests passed.
```


SyntaxNode:
Thank you for adding your adapter. We have a fair bit of new adapter PRs open at the moment, so please be patient and we'll review this as soon.

adapters/kidoz/kidoz.go
```diff
@@ -0,0 +1,188 @@
...
+	case http.StatusServiceUnavailable:
+		return nil, nil
+
+	case http.StatusBadRequest:
+		fallthrough
+	case http.StatusUnauthorized:
+		fallthrough
+	case http.StatusForbidden:
+		return nil, []error{&errortypes.BadInput{
+			Message: "unexpected status code: " + strconv.Itoa(responseData.StatusCode),
```
SyntaxNode: Consider using different text in the error messages between BadInput and BadServerResponse to make the case more obvious to callers.

validate.sh
```diff
@@ -27,11 +27,11 @@ GOGLOB="${GOGLOB/ docs/}"
 GOGLOB="${GOGLOB/ vendor/}"
 
 # Check that there are no formatting issues
-GOFMT_LINES=`gofmt -s -l $GOGLOB | wc -l | xargs`
+GOFMT_LINES=`gofmt -s -l $GOGLOB | tr '\\\\' '/' | wc -l | xargs`
```
SyntaxNode: Why did you need to change this?

analytics/filesystem/testFiles/test-20200303
```diff
@@ -0,0 +1,5 @@
+2020/03/03 13:53:18 DEBUG {"type":"/openrtb2/auction","Status":0,"Errors":null,"Request":null,"Response":null}
```
SyntaxNode: Should these testFiles be committed?

adapters/kidoz/params_test.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+}
+
+var validParams = []string{
+	`{"publisher_id":"pub-valid-0", "access_token":"token-valid-0"}`,
+}
+
+var invalidParams = []string{
+	``,
```
SyntaxNode: Nice! Thank you for the thorough testing.

adapters/kidoz/kidoz_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	t.Run("Handles response marshal failure", func(t *testing.T) {
+		request := makeBidRequest()
+		requestData := &adapters.RequestData{}
+		responseData := &adapters.ResponseData{
+			StatusCode: http.StatusOK,
+		}
+
+		resp, errs := kidoz.MakeBids(request, requestData, responseData)
+		// cant assert message its different on different versions of go
```
SyntaxNode: If you'd like to perform this test in json, we have a regex matcher to solve the problem between go < 1.13 and 1.13+.

validate.sh
rhaksi-kidoz: go tooling is not shell aware, so when you use git bash to run validate.sh, it call gofmt, and gofmt returns paths\like\this, which is incompatible with bash, so this reformats them as paths/like/this et voila.

analytics/filesystem/testFiles/test-20200303
rhaksi-kidoz: No that was a mistake, but shouldnt the gitignore file ignore the analytics dir?

adapters/kidoz/kidoz_test.go
rhaksi-kidoz: I think this tests the important thing, ie that marshalling errors are handled correctly.

analytics/filesystem/testFiles/test-20200303
SyntaxNode: I would expect it to, yes. Doesn't seem to be though. I haven't seen these accumulate locally for me. I guess the cleanup part of the test failed or was skipped during a debug session? In any event, I agree that should be in gitignore. If you can through that in here, great. If not, no worries and I'll throw it in my next PR.

openrtb_ext/bid.go
```diff
@@ -42,9 +42,9 @@ type BidType string
 
 const (
 	BidTypeBanner BidType = "banner"
-	BidTypeVideo          = "video"
-	BidTypeAudio          = "audio"
-	BidTypeNative         = "native"
+	BidTypeVideo  BidType = "video"
```
SyntaxNode: Was this change necessary? Did the other constants not pick up the correct type?

adapters/kidoz/kidoz_test.go
SyntaxNode: That's fine. The json tests are not a requirement. It's a bit difficult to understand which error you're expecting here could you do an assert.Contains() for the non-different parts?

openrtb_ext/bid.go
rhaksi-kidoz: Other than the first one, they all are implicitly being assigned the type "string" rather than BidType. (ie BidTypeVideo = "video" is a string, BidTypeVideo  BidType = "video" is the type BidType)

Definitely necessary, without this examining passing any constant other than the first will generate type errors.

PR #1211, opened 2020-03-04
openx adapter: forward bid response currency in openx adapter if set

kenan-gillet:
it was always set to the default USD before

### DIFF
--- a/adapters/openx/openx.go
+++ b/adapters/openx/openx.go
@@ -169,6 +169,11 @@ func (a *OpenxAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalReq
 
 	bidResponse := adapters.NewBidderResponseWithBidsCapacity(5)
 
+	// overrride default currency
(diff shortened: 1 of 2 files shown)


adapters/openx/openx.go
```diff
@@ -169,6 +169,11 @@ func (a *OpenxAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalReq
 
 	bidResponse := adapters.NewBidderResponseWithBidsCapacity(5)
 
+	// overrride default currency
+	if bidResp.Cur != "" && bidResp.Cur != "USD" {
```
SyntaxNode: Is the check for `bidResp.Cur != "USD"` necessary? I know it's currently the default, but is it worth to hardcode that assumption here?

kenan-gillet: not really, will remove it

kenan-gillet: done

PR #1219, opened 2020-03-11
If Device.UA is not present in request body, init it with user-agent from header

VeronikaSolovei9:
Cannot come up with unit tests for this change. Do we want it?

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -566,10 +566,7 @@ func (deps *endpointDeps) parseVideoRequest(request []byte, headers http.Header)
 
 	//if Device.UA is not present in request body, init it with user-agent from request header if it's present
 	if req.Device.UA == "" {
-		userAgent := headers.Get("User-Agent")
-		if userAgent != "" {
-			req.Device.UA = userAgent
-		}
+		req.Device.UA = headers.Get("User-Agent")
 	}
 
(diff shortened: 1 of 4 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -140,6 +140,14 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 	//create full open rtb req from full video request
 	mergeData(videoBidReq, bidReq)
 
+	//if Device.UA is not present in request body, init it with user-agent from request header if it's present
+	if videoBidReq.Device.UA == "" {
+		userAgent := r.Header.Get("User-Agent")
```
SyntaxNode: You could probably just write `videoBidReq.Device.UA = r.Header.Get("User-Agent")` as nothing really changes if the header value is empty and the device ua is already empty.

VeronikaSolovei9: Yes!

VeronikaSolovei9:
@guscarreon Gus, thank you for the comment! In initial implementation it looked exactly like you wrote, this check was outside the `parseVideoRequest` function. But I was not able to come with good unit tests for this, because we don't return video request, it's only an input parameter to other functions. In case to test it outside `parseVideoRequest` I would need to mock all functions following by the header setting. We had this discussion in Slack and decided to move header setting to `parseVideoRequest`.

PR #1220, opened 2020-03-12
Added logging to cache for video endpoint

camrice:
Redo of the previous PR. Caching now occurs during VAST caching. If the request fails, a new cache call is made. Unit tests to be added.

### DIFF
--- a/endpoints/openrtb2/amp_auction_test.go
+++ b/endpoints/openrtb2/amp_auction_test.go
@@ -903,7 +903,7 @@ type mockAmpExchange struct {
 	lastRequest *openrtb.BidRequest
 }
 
-func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, ids exchange.IdFetcher, labels pbsmetrics.Labels, categoriesFetcher *stored_requests.CategoryFetcher, videoReq []byte) (*openrtb.BidResponse, error) {
+func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, ids exchange.IdFetcher, labels pbsmetrics.Labels, categoriesFetcher *stored_requests.CategoryFetcher, debugLog string) (*openrtb.BidResponse, error) {
 	m.lastRequest = bidRequest
 
 	response := &openrtb.BidResponse{
--- a/endpoints/openrtb2/amp_auction_test.go
+++ b/endpoints/openrtb2/amp_auction_test.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/prebid/prebid-server/prebid_cache_client"
 	"github.com/prebid/prebid-server/stored_requests"
 	"github.com/prebid/prebid-server/stored_requests/backends/empty_fetcher"
 
@@ -902,7 +903,7 @@ type mockAmpExchange struct {
 	lastRequest *openrtb.BidRequest
 }
 
-func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, ids exchange.IdFetcher, labels pbsmetrics.Labels, categoriesFetcher *stored_requests.CategoryFetcher) (*openrtb.BidResponse, error) {
+func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, ids exchange.IdFetcher, labels pbsmetrics.Labels, categoriesFetcher *stored_requests.CategoryFetcher, videoReq []byte) (*openrtb.BidResponse, error) {
 	m.lastRequest = bidRequest
 
 	response := &openrtb.BidResponse{
@@ -924,6 +925,10 @@ func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.B
 	}
 
 	return response, nil
+}
+
+func (m *mockAmpExchange) GetCache() prebid_cache_client.Client {
+	return nil
 }
 
 func getTestBidRequest(nilUser bool, nilExt bool, consentString string, digitrustID string) ([]byte, error) {
--- a/endpoints/openrtb2/amp_auction_test.go
+++ b/endpoints/openrtb2/amp_auction_test.go
@@ -11,7 +11,6 @@ import (
 	"strconv"
 	"testing"
 
-	"github.com/prebid/prebid-server/prebid_cache_client"
 	"github.com/prebid/prebid-server/stored_requests"
 	"github.com/prebid/prebid-server/stored_requests/backends/empty_fetcher"
 
@@ -903,7 +902,7 @@ type mockAmpExchange struct {
 	lastRequest *openrtb.BidRequest
 }
 
-func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, ids exchange.IdFetcher, labels pbsmetrics.Labels, categoriesFetcher *stored_requests.CategoryFetcher, debugLog string) (*openrtb.BidResponse, error) {
+func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, ids exchange.IdFetcher, labels pbsmetrics.Labels, categoriesFetcher *stored_requests.CategoryFetcher, debugLog *exchange.DebugLog) (*openrtb.BidResponse, error) {
 	m.lastRequest = bidRequest
 
 	response := &openrtb.BidResponse{
@@ -925,10 +924,6 @@ func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.B
 	}
 
 	return response, nil
-}
-
-func (m *mockAmpExchange) GetCache() prebid_cache_client.Client {
-	return nil
 }
 
 func getTestBidRequest(nilUser bool, nilExt bool, consentString string, digitrustID string) ([]byte, error) {
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -289,7 +289,7 @@ func cleanupVideoBidRequest(videoReq *openrtb_ext.BidRequestVideo, podErrors []P
 
 func handleError(labels *pbsmetrics.Labels, w http.ResponseWriter, errL []error, vo *analytics.VideoObject, debugLog *exchange.DebugLog) {
 	if debugLog != nil && debugLog.EnableDebug {
-		if rawUUID, _ := uuid.NewV4(); len(rawUUID) > 0 {
+		if rawUUID, err := uuid.NewV4(); err == nil {
 			debugLog.CacheKey = rawUUID.String()
 		}
 		errL = append(errL, fmt.Errorf("[Debug cache ID: %s]", debugLog.CacheKey))
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/buger/jsonparser"
 	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/gofrs/uuid"
 	"github.com/prebid/prebid-server/errortypes"
 
 	"github.com/golang/glog"
@@ -24,6 +25,7 @@ import (
 	"github.com/prebid/prebid-server/exchange"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/pbsmetrics"
+	"github.com/prebid/prebid-server/prebid_cache_client"
 	"github.com/prebid/prebid-server/stored_requests"
 	"github.com/prebid/prebid-server/usersync"
 )
@@ -79,7 +81,24 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		CookieFlag:    pbsmetrics.CookieFlagUnknown,
 		RequestStatus: pbsmetrics.RequestStatusOK,
 	}
+	var videoReq []byte
+	debugQuery := r.URL.Query().Get("debug")
+	enableDebug := debugQuery == "true"
+	errorCacheID := ""
 	defer func() {
+		if enableDebug && vo.VideoResponse == nil {
+			cache := deps.ex.GetCache()
+			data, _ := json.Marshal(string(videoReq))
+			toCache := []prebid_cache_client.Cacheable{
+				{
+					Type:       prebid_cache_client.TypeXML,
+					Data:       data,
+					TTLSeconds: int64(3600),
+					Key:        errorCacheID,
+				},
+			}
+			cache.PutJson(context.Background(), toCache)
+		}
 		deps.metricsEngine.RecordRequest(labels)
 		deps.metricsEngine.RecordRequestTime(labels, time.Since(start))
 		deps.analytics.LogVideoObject(&vo)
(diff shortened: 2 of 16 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,24 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+	errorCacheID := ""
 	defer func() {
+		if enableDebug && vo.VideoResponse == nil {
+			cache := deps.ex.GetCache()
+			data, _ := json.Marshal(string(videoReq))
+			toCache := []prebid_cache_client.Cacheable{
+				{
+					Type:       prebid_cache_client.TypeXML,
+					Data:       data,
+					TTLSeconds: int64(3600),
```
camrice: This will be updated to not use a hard coded value

exchange/exchange.go
```diff
@@ -162,7 +163,13 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 
 		if targData != nil {
 			auc.setRoundedPrices(targData.priceGranularity)
-			cacheErrs := auc.doCache(ctx, e.cache, targData, bidRequest, 60, &e.defaultTTLs, bidCategory)
+			if len(videoReq) > 0 {
+				// Build the response so far
+				debugBidResp, _ := e.buildBidResponse(ctx, liveAdapters, adapterBids, bidRequest, resolvedRequest, adapterExtra, auc, errs)
```
camrice: This is something I am looking to remove. However, the ordering of the calls to get the final response seem to be fixed. If the `doCache` call is made after the bid response is built, then the cache keys are not set and the response will fail. I will keep investigating to see if there's a way to avoid this duplicate call.

endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,27 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+		if enableDebug && vo.VideoResponse == nil {
+			cache := deps.ex.GetCache()
+			debugString = fmt.Sprintf("<!--\n%s\n-->", debugString)
+			data, _ := json.Marshal(debugString)
+			toCache := []prebid_cache_client.Cacheable{
+				{
+					Type:       prebid_cache_client.TypeXML,
+					Data:       data,
+					TTLSeconds: int64(deps.cfg.CacheURL.DefaultTTLs.Video),
+					Key:        fmt.Sprintf("log_%s", errorCacheID),
```
maymer75: Is errorCacheID guaranteed to be set to something here? In other words we're sure handleError is called in every case where vo.VideoResponse doesn't get set?

Actually this gives me an idea about how we might be able to avoid having errorCacheID be returned by the handleError() function but still have it in the response, remind me tomorrow and we can discuss this either in person or via slack.

camrice: Ah I should be checking if it's set to something here. I could use a string pointer instead and modify in the `handleError`. Then I can check if it is empty. Changing this is quick though so we can discuss it tomorrow.

exchange/exchange.go
```diff
@@ -179,6 +200,10 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	return e.buildBidResponse(ctx, liveAdapters, adapterBids, bidRequest, resolvedRequest, adapterExtra, auc, errs)
 }
 
+func (e *exchange) GetCache() prebid_cache_client.Client {
```
SyntaxNode: I don't think you need this. Please see my other comment.

endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,27 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		CookieFlag:    pbsmetrics.CookieFlagUnknown,
 		RequestStatus: pbsmetrics.RequestStatusOK,
 	}
+
+	debugQuery := r.URL.Query().Get("debug")
+	enableDebug := debugQuery == "true"
```
SyntaxNode: Consider being 'nice' and making this a case insensitive compare.

endpoints/openrtb2/video_auction.go
SyntaxNode: I'm good with the cache put here, since it will only happen if the auction didn't write to the cache ensuring only a single cache put per request.

Please add checks in this code to defend against nils and errors. Specifically, ensure that `cache` is not nil before calling `PutJson`, do not ignore the error from json.Marshal. The TTLSeconds conversion also looks a bit risky to me, is it guaranteed here to succeed? 

For the key formatting, it might be better to use simple string concatenation in this instance.

endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,27 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+	var debugString string
+	var errorCacheID string
+
 	defer func() {
+		if len(errorCacheID) > 0 && vo.VideoResponse == nil {
+			cache := deps.ex.GetCache()
```
SyntaxNode: Asking the exchange for an instance of the cache doesn't make sense to me. I recommend making the cache available directly to the video auction endpoint handler. We use a simple DI construct in router.go. Then you don't need to add a method to the exchange interface... and you can then also pre-check that it isn't nil.

endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,27 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+			data, _ := json.Marshal(debugString)
+			toCache := []prebid_cache_client.Cacheable{
+				{
+					Type:       prebid_cache_client.TypeXML,
+					Data:       data,
+					TTLSeconds: int64(deps.cfg.CacheURL.DefaultTTLs.Video),
+					Key:        fmt.Sprintf("log_%s", errorCacheID),
+				},
+			}
+			cache.PutJson(context.Background(), toCache)
```
SyntaxNode: Please use a timed context here and consider exposing the duration as a config option, or perhaps since this is temporary code until a better logging mechanism can be built just use the same 50ms hardcoded value as the auction reserve time.

exchange/auction.go
```diff
@@ -147,6 +147,17 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
 		}
 	}
 
+	if len(debugLog) > 0 {
+		if jsonBytes, err := json.Marshal(debugLog); err == nil {
+			toCache = append(toCache, prebid_cache_client.Cacheable{
+				Type:       prebid_cache_client.TypeXML,
```
SyntaxNode: Consider changing debugLog to a struct which includes the same string value you have right now and the type. That'll make it reusable for, say, the AMP endpoint if it needs to do something similar in the future for json logs. Same for the TTL.

exchange/exchange.go
```diff
@@ -162,7 +163,27 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+
+				// Remove the VAST object from the request
+				for i := 0; i < len(debugBidResp.SeatBid); i++ {
+					for j := 0; j < len(debugBidResp.SeatBid[i].Bid); j++ {
+						debugBidResp.SeatBid[i].Bid[j].AdM = ""
+					}
+				}
+
+				// Remove the resolved request due to including duplicate info
+				var respExt openrtb_ext.ExtBidResponse
```
SyntaxNode: The unmarshalling and remarshalling is expensive. Consider writing a `buildBidDebugResponse` which can omit the resolved request entirely... and it can also process the AdM removal.

exchange/exchange.go
SyntaxNode: Easy if you just want to include the debug info. You can pull up creation of the debug info and pass it down into the existing buildBidResponse below. If you want to include the seat bids, well.. there's no avoiding building those twice since the ones in the actual response will need to include the cache ids.

endpoints/openrtb2/video_auction.go
```diff
@@ -242,7 +273,13 @@ func cleanupVideoBidRequest(videoReq *openrtb_ext.BidRequestVideo, podErrors []P
 	return videoReq
 }
 
-func handleError(labels *pbsmetrics.Labels, w http.ResponseWriter, errL []error, vo *analytics.VideoObject) {
+func handleError(labels *pbsmetrics.Labels, w http.ResponseWriter, errL []error, vo *analytics.VideoObject, enableDebug bool, errorCacheID *string) {
+	if enableDebug {
+		if rawUUID, _ := uuid.NewV4(); len(rawUUID) > 0 {
+			*errorCacheID = rawUUID.String()
+		}
+		errL = append(errL, fmt.Errorf("Debug cache ID: [%s]", *errorCacheID))
```
SyntaxNode: I'm a bit confused here. If we're sending back the cache id in the body when there's no video response, why can't we just send back the full debug info as well? If the caller can inspect the response for the id, can't they inspect the response for everything?

endpoints/openrtb2/video_auction.go
camrice: Updated to be case insensitive

endpoints/openrtb2/video_auction.go
camrice: Updated to have the cache directly available to the video endpoint. It is instead passed in from the router. Also added nil check for the cache.

exchange/exchange.go
camrice: Removed. Video endpoint now has a direct reference to the cache.

exchange/auction.go
camrice: Added a `debugLog` type that holds all info, including if the debug is enabled

endpoints/openrtb2/video_auction.go
camrice: Added checks around the caching to ensure that it is safe to make the call. Also updated key formatting.

For the TTL, should I set a default value and then check if the config has something, or if the config has no TTL value, ignore the caching?

endpoints/openrtb2/video_auction.go
camrice: Updated to use a deadline. I used the `defaultRequestTimeout` defined in the Video auction. As a temporary solution, I am not sure if I should add a config value but I am open to it if it better defined there.

exchange/exchange.go
camrice: Updated to build the debug ext only. This requires no modification, so there is just one marshal call to covert to a string.

exchange/exchange.go
camrice: I went with just including the debug info. It makes it simpler and doesn't duplicate the calls. However, I need to check if the output is sufficient for the resulting log.

exchange/auction.go
camrice: Sorry missed adding this change in. Will be added with unit test update.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -926,6 +927,10 @@ func (m *mockAmpExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.B
 	return response, nil
 }
 
+func (m *mockAmpExchange) GetCache() prebid_cache_client.Client {
```
camrice: These will be removed from unit tests as well

endpoints/openrtb2/video_auction.go
SyntaxNode: The `defaultRequestTimeout` is 5 seconds. I think that's too long to wait for a cache call. I think you can just hardcode it for 50ms to match the hardcoded elsewhere. Honestly, since those are non-configurable this one is probably fine with a hardcoded value as well.. at least for now.

endpoints/openrtb2/video_auction.go
SyntaxNode: Thank you. :)

endpoints/openrtb2/video_auction.go
camrice: It seems that at least running the server locally, the cache call here is exceeding 50ms. 100ms seems to work okay. Should it be definitely less than 50ms or is somewhere around 100ms for the timeout appropriate?

endpoints/openrtb2/video_auction.go
camrice: No problem!

exchange/auction.go
camrice: Change added

endpoints/openrtb2/video_auction.go
SyntaxNode: 50ms the minimal cache reserve time for AMP and is usually not an issue. I don't want to split hairs over 50ms vs 100ms, especially for a debugging only use case. 100ms would be fine.

exchange/exchange.go
SyntaxNode: Sweet. If that works for you, that's a really nice optimization.

endpoints/openrtb2/video_auction.go
```diff
@@ -91,38 +126,43 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
 	if err != nil {
-		handleError(&labels, w, []error{err}, &vo)
+		handleError(&labels, w, []error{err}, &vo, &debugLog)
 		return
 	}
 
 	resolvedRequest := requestJson
+	if debugLog.EnableDebug {
+		debugLog.Data = fmt.Sprintf("Request:\n%s", string(requestJson))
+		headerBytes, _ := json.Marshal(r.Header)
```
SyntaxNode: Please do not ignore the error.

endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,40 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+						Type:       debugLog.CacheType,
+						Data:       data,
+						TTLSeconds: debugLog.TTL,
+						Key:        "log_" + debugLog.CacheKey,
+					},
+				}
+				if deps.cache != nil {
+					ctx := context.Background()
+					var cancel context.CancelFunc
+					ctx, cancel = context.WithDeadline(ctx, start.Add(time.Duration(100)*time.Millisecond))
```
SyntaxNode: Can this be simplified as:
```
ctx, cancel : = context.Background().WithDeadline(...)
```

.. otherwise looking good.

exchange/exchange.go
```diff
@@ -162,7 +171,12 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 
 		if targData != nil {
 			auc.setRoundedPrices(targData.priceGranularity)
-			cacheErrs := auc.doCache(ctx, e.cache, targData, bidRequest, 60, &e.defaultTTLs, bidCategory)
+			if debugLog != nil && debugLog.EnableDebug {
+				bidResponseExt = e.makeExtBidResponse(adapterBids, adapterExtra, bidRequest, resolvedRequest, errs)
+				bidRespExtBytes, _ := json.Marshal(bidResponseExt)
```
SyntaxNode: Please do not ignore the error.

endpoints/openrtb2/video_auction.go
camrice: Ah missed these error checks. Added!

endpoints/openrtb2/video_auction.go
camrice: Simplified the initialization here

exchange/exchange.go
camrice: Added error check for this case

endpoints/openrtb2/amp_auction_test.go
camrice: These are removed

endpoints/openrtb2/video_auction.go
camrice: Understood! I will keep at 100ms for the time being.

exchange/exchange.go
camrice: Thanks! Just confirmed that this output is sufficient for our current needs.

endpoints/openrtb2/video_auction.go
camrice: Confirmed that we will be using this output for now, where just the cache ID is logged. We will keep an eye out for any requested changes regarding what should be returned in an error case.

endpoints/openrtb2/video_auction.go
```diff
@@ -242,7 +287,13 @@ func cleanupVideoBidRequest(videoReq *openrtb_ext.BidRequestVideo, podErrors []P
 	return videoReq
 }
 
-func handleError(labels *pbsmetrics.Labels, w http.ResponseWriter, errL []error, vo *analytics.VideoObject) {
+func handleError(labels *pbsmetrics.Labels, w http.ResponseWriter, errL []error, vo *analytics.VideoObject, debugLog *exchange.DebugLog) {
+	if debugLog != nil && debugLog.EnableDebug {
+		if rawUUID, _ := uuid.NewV4(); len(rawUUID) > 0 {
```
SyntaxNode: This will produce warnings that you're ignoring the error from NewV4. In practice, you're likely fine with a check on the uuid.. but checking for a non-nil error is preferred.

endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,38 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+	}
+	debugLog := exchange.DebugLog{
+		EnableDebug: strings.EqualFold(debugQuery, "true"),
+		CacheType:   prebid_cache_client.TypeXML,
+		TTL:         cacheTTL,
+	}
+
 	defer func() {
+		if len(debugLog.CacheKey) > 0 && vo.VideoResponse == nil {
+			debugLog.Data = fmt.Sprintf("<!--\n%s\n\nNo response created\n-->", debugLog.Data)
```
SyntaxNode: Go tooling reports there is no unit test coverage for inside of the if statement. Please add a test case.

endpoints/openrtb2/video_auction.go
camrice: Updated to check the error instead

endpoints/openrtb2/video_auction.go
camrice: Added a unit test for the failure case. Should be covered now.

endpoints/openrtb2/video_auction.go
```diff
@@ -79,7 +81,38 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+			data, err := json.Marshal(debugLog.Data)
```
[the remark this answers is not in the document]
hhhjort: Looks like down below `debugLog.Data` is a string. So marshalling that will create a string with all the JSON special characters escaped. Is this what is desired?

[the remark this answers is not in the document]
camrice: The output we are logging from testing is looking correct in terms of containing the right formatting and characters. Is there a more efficient or desirable way to perform this conversion to get this to be a `json.RawMessage`?

endpoints/openrtb2/video_auction.go
SyntaxNode: Looks good now. Thank you.

PR #1226, opened 2020-03-18
Added metrics support to endpoint aspect

VeronikaSolovei9:


pbsmetrics/metrics.go
```diff
@@ -168,6 +169,7 @@ func RequestStatuses() []RequestStatus {
 		RequestStatusErr,
 		RequestStatusNetworkErr,
 		RequestStatusBlacklisted,
+		RequestStatusQueueTimeout,
```
SyntaxNode: I'd like to start the discussion here if we want to consider a queued timeout a request. The feature is essentially a hack for nginx's queues not being able to discard messages which are meet some threshold for being expired. In other infrastructure topologies, the request might never make it to the app server. The total number of requests can be used (reasonably IMHO) for capacity planning and performance monitoring, but if we're quickly discarding these queue timeout.. should they be included? i'm personally leaning towards no, but connections and at least some server resources are used, so maybe?

If we consider this a request, this looks good. This addition adds 5 new metrics (for prometheus at least) since it's correlated to the request types (legacy, openrtb web, openrtb app, amp, and video) - which isn't a big deal when we're at around 3k metrics right now.

If we don't consider this a request, we would want to create a separate counter metric.

pbsmetrics/metrics.go
```diff
@@ -168,6 +169,7 @@ func RequestStatuses() []RequestStatus {
...
 	}
```
SyntaxNode: I think we should also add a metric to track the queue duration to give us better visibility into the performance. That'll give us an idea for how long wait times are looking in the queue and could power decisions for scaling or adjusting the queue window. Thoughts?

router/aspects/request_timeout_handler.go
```diff
@@ -3,11 +3,21 @@ package aspects
...
 	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/pbsmetrics"
 	"net/http"
 	"strconv"
 )
 
-func QueuedRequestTimeout(f httprouter.Handle, reqTimeoutHeaders config.RequestTimeoutHeaders) httprouter.Handle {
+func QueuedRequestTimeout(f httprouter.Handle, reqTimeoutHeaders config.RequestTimeoutHeaders, metricsEngine pbsmetrics.MetricsEngine, requestType pbsmetrics.RequestType) httprouter.Handle {
+
+	labels := pbsmetrics.Labels{
```
SyntaxNode: The fact that most of these values are set to unknown might be a good reason to move queue timeouts to their own metric.

pbsmetrics/metrics.go
VeronikaSolovei9: I got what you mean. 
This is still a request and together with other metrics, like NetworkError and Blacklisted this additional metric is on its place. Tracking queue duration is a good idea and if we want it, then it makes sense to move queued requests related data to separate metrics structure.

pbsmetrics/metrics.go
VeronikaSolovei9: Agree! Please see the above comment.

router/aspects/request_timeout_handler.go
VeronikaSolovei9: Agree! If we have a new metrics structure this will solve this

pbsmetrics/metrics.go
VeronikaSolovei9: Will move to separate metrics struct

pbsmetrics/metrics.go
hhhjort: I think counting the requests is probably the way to go. In an ideal world, we want to serve all the requests in the end.

VeronikaSolovei9: Hans, what do you mean by counting the requests? My understanding is that we send metrics object for each request (if timeout or bad headers happened) and Grafana will count it, isn't it? 
Scott, I'm implementing a separate structure to represent only queued requests and don't mix it with regular Labels. Is this what you meant? Just to make sure.

hhhjort: I was agreeing with you Veronica. :)

router/aspects/request_timeout_handler.go
SyntaxNode: Since the discussion is currently leaning towards moving ahead with a `RequestStatusQueueTimeout` status within the existing request count metric, it's best to use the same labels struct. We can improve that piece another time.

pbsmetrics/metrics.go
SyntaxNode: Ok. Sound reasoning. I'm good with this. The only negative aspect is the queue filtering is only for video auctions right now but due to cardinality would also have metrics against the other auction types. This will add 4 unused metrics, which isn't worth a concern IMHO.

pbsmetrics/metrics.go
SyntaxNode: To add a new metric to track the queue time (and possible outcome, such as timeouts) you'll need to add a method to the `MetricsEngine` interface and fill in the functionality for the Influx and Prometheus real engines, the "fake" ones for testing (MetricsEngineMock, DummyMetricsEngine), and the MultiMetricsEngine multiplexor.

VeronikaSolovei9: Yes, and fix unit tests.

pbsmetrics/metrics.go
```diff
@@ -8,12 +8,13 @@ import (
...
-	Browser       Browser
-	CookieFlag    CookieFlag
-	RequestStatus RequestStatus
+	Source               DemandSource
+	RType                RequestType
+	PubID                string // exchange specific ID, so we cannot compile in values
+	Browser              Browser
+	CookieFlag           CookieFlag
+	RequestStatus        RequestStatus
+	RequestQueueDuration float64
```
SyntaxNode: Unfortunately, there is no magic (aka reflection) involved with metrics. Adding a field to the labels struct makes the data available to the Influx and Prometheus metric engines, but they would need to specifically look for the new information and forward it in some way to their respective data stores.

Request time is currently tracked by the RecordRequestTime method, taking in both the Labels struct and a duration. That is not appropriate for recording queued request time because RecordRequestTime filters out all non-successful requests. I fear changing that behavior would introduce a breaking change and cause issues with existing monitoring queries folks may have.

I recommending adding a new metrics engine method along the lines of `RecordRequestQueueTime` possibly including the RType/RequestType. You'll need to add new metrics to both Influx (aka go-metrics) and Prometheus.

VeronikaSolovei9: Added initial attempt to support queued request timeout metrics. Unit tests will be added. Please review if this makes sense to you and I move in right direction.

VeronikaSolovei9:
Ready for review

router/aspects/request_timeout_handler.go
```diff
@@ -3,11 +3,15 @@ package aspects
...
 	"github.com/julienschmidt/httprouter"
 	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/pbsmetrics"
 	"net/http"
 	"strconv"
+	"time"
 )
 
-func QueuedRequestTimeout(f httprouter.Handle, reqTimeoutHeaders config.RequestTimeoutHeaders) httprouter.Handle {
+const NanosecondsInSecond = 1000000000
```
hhhjort, line 12: What about using `time.Second` instead? Then don't have to worry that you have the correct number of zeros. :)

router/aspects/request_timeout_handler.go
```diff
@@ -23,20 +27,25 @@ func QueuedRequestTimeout(f httprouter.Handle, reqTimeoutHeaders config.RequestT
 		reqTimeFloat, reqTimeFloatErr := strconv.ParseFloat(reqTimeInQueue, 64)
 		reqTimeoutFloat, reqTimeoutFloatErr := strconv.ParseFloat(reqTimeout, 64)
 
+		reqTimeDuration := time.Duration(reqTimeFloat * NanosecondsInSecond)
```
hhhjort, line 30: I assume this means that the header has request time in seconds.

pbsmetrics/metrics.go
```diff
@@ -272,4 +274,5 @@ type MetricsEngine interface {
 	RecordStoredReqCacheResult(cacheResult CacheResult, inc int)
 	RecordStoredImpCacheResult(cacheResult CacheResult, inc int)
 	RecordPrebidCacheRequestTime(success bool, length time.Duration)
+	RecordRequestQueueTime(requestStatus RequestStatus, requestType RequestType, length time.Duration)
```
SyntaxNode, line 277: I think request type is good. It might be interesting to observe differences between request types which might reveal interesting insights. I don't think RequestStatus is appropriate here though. It has high cardinality and we really just care about "accepted" or "rejected" from a timing perspective. I recommend changing that to a bool.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -187,6 +188,12 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
 		"Count of total requests to Prebid Server labeled by account.",
 		[]string{accountLabel})
 
+	metrics.requestsQueueTimer = newHistogram(cfg, metrics.Registry,
```
SyntaxNode, line 191: Please also add to preload.go to reduce runtime overhead and to allow the gatekeeping tests to function properly.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -187,6 +188,12 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
...
+		"request_queue_time",
+		"Seconds request was waiting in queue",
+		[]string{requestTypeLabel, requestStatusLabel},
```
SyntaxNode, line 194: There are 5 possible types and 6 possible status. This cardinality results in 30 series and the cost of each series is 14 metrics each. I think 420 new metrics for this timer is a bit much. If we bring the status cardinality down to a bool we get a better 140 new metrics value.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -187,6 +188,12 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
...
+		prometheus.DefBuckets)
```
SyntaxNode, line 195: You don't want to use DefBuckets. That's geared toward network connectivity measurements and maxes out at 5 seconds. Please use a new custom bucket series. I think you're expecting wait times up 5 minutes? Consider then buckets like:

0, 1, 5, 30, 60, 120, 180, 240, 300

The series (which I just guessed at) is geared towards measuring fast responses with less resolution spread equally up to 5 minutes.  

Note: Influx records time series differently and doesn't have this concern.

router/aspects/request_timeout_handler.go
VeronikaSolovei9: Sure, fixed!

router/aspects/request_timeout_handler.go
VeronikaSolovei9: Yes, fixed

pbsmetrics/metrics.go
VeronikaSolovei9: We now have 3 states: 
`RequestStatusErr` - in case request timeout headers are incorrect (wrong format)
`RequestStatusQueueTimeout` - in case requests stays too long in queue
and `RequestStatusOK`
Do you want me to refactor it and "merge" 2 error statuses into one?

pbsmetrics/prometheus/prometheus.go
VeronikaSolovei9: Good catch, added

pbsmetrics/prometheus/prometheus.go
VeronikaSolovei9: Makes sense, modified

pbsmetrics/prometheus/prometheus.go
VeronikaSolovei9: Will do if you think we need to merge two error statuses into one. Please see my comment above.

pbsmetrics/metrics.go
SyntaxNode: I don't think we need to record any metrics for the  `RequestStatusErr` state. If the headers are wrong, we won't be able to compute the duration, which leaves just timeout and ok.

The existing code will record the `RequestStatusErr` state in the request count metrics, so we know how often it happens.

I'm also fine to preload the video type only for now. Please add a comment to in the prometheus.go code since all other metrics are fully enumerated in the preload.

pbsmetrics/prometheus/prometheus.go
SyntaxNode: Answered above.

pbsmetrics/prometheus/prometheus.go
SyntaxNode: Great. Thank you.

pbsmetrics/metrics.go
VeronikaSolovei9: Done!

VeronikaSolovei9:
Ready for review

VeronikaSolovei9:
Previous commits squashed and rebased over latest upstream.

pbsmetrics/go_metrics.go
```diff
@@ -526,6 +537,10 @@ func (me *Metrics) RecordPrebidCacheRequestTime(success bool, length time.Durati
 	}
 }
 
+func (me *Metrics) RecordRequestQueueTime(success bool, requestType RequestType, length time.Duration) {
+	me.RequestsQueueTimer[requestType][success].Update(length)
```
SyntaxNode, line 541: This will produce a null reference exception if the requestType is not video. Not a problem, but please add a check here to ignore all non-video requests to be sure.

pbsmetrics/prometheus/preload.go
```diff
@@ -91,6 +91,13 @@ func preloadLabelValues(m *Metrics) {
 		adapterLabel: adapterValues,
 		actionLabel:  actionValues,
 	})
+
+	//to minimize memory usage, queuedTimeout metric is now supported for video endpoint only
+	//boolean value represents 2 general request statuses: accepted and rejected
+	preloadLabelValuesForHistogram(m.requestsQueueTimer, map[string][]string{
+		requestTypeLabel:   {isVideoLabel},
```
SyntaxNode, line 98: The `isVideoLabel` is specific to the `impressions_requests` metric. It works out because that value is coincidentally the same as `ReqTypeVideo`, but please use `ReqTypeVideo` here in case of future refactors.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -73,11 +74,17 @@ const (
 	markupDeliveryNurl = "nurl"
 )
 
+const (
+	requestSuccessLabel = "accepted"
```
SyntaxNode, line 78: Consider renaming to `requestAcceptedLabel` to better match its value.

router/aspects/request_timeout_handler_test.go
```diff
@@ -19,76 +21,62 @@ func TestAny(t *testing.T) {
 		reqTimeInQueue          string
 		reqTimeOut              string
 		setHeaders              bool
-		expectedRespCode        int
+		extectedRespCode        int
```
SyntaxNode, line 24: Looks like you introduced a typo here.

pbsmetrics/go_metrics.go
VeronikaSolovei9: Added `requestType` check

pbsmetrics/prometheus/preload.go
VeronikaSolovei9: Fixed!

pbsmetrics/prometheus/prometheus.go
VeronikaSolovei9: Refactored

router/aspects/request_timeout_handler_test.go
VeronikaSolovei9: Sure, done!

PR #1228, opened 2020-03-19
First pass at phase 1 TCF 2.0 support

hhhjort:
This will require updating the go-gdpr library first.

### DIFF
--- a/gdpr/impl.go
+++ b/gdpr/impl.go
@@ -4,7 +4,8 @@ import (
 	"context"
 
 	"github.com/prebid/go-gdpr/api"
-	"github.com/prebid/go-gdpr/consentconstants"
+	tcf1constants "github.com/prebid/go-gdpr/consentconstants"
+	consentconstants "github.com/prebid/go-gdpr/consentconstants/tcf2"
 	"github.com/prebid/go-gdpr/vendorconsent"
 	"github.com/prebid/go-gdpr/vendorlist"
 	"github.com/prebid/prebid-server/config"
@@ -100,7 +101,7 @@ func (p *permissionsImpl) allowPI(ctx context.Context, vendorID uint16, consent 
 			return true, nil
 		}
 	} else {
-		if (vendor.Purpose(consentconstants.InfoStorageAccess) || vendor.LegitimateInterest(consentconstants.InfoStorageAccess)) && parsedConsent.PurposeAllowed(consentconstants.InfoStorageAccess) && (vendor.Purpose(consentconstants.AdSelectionDeliveryReporting) || vendor.LegitimateInterest(consentconstants.AdSelectionDeliveryReporting)) && parsedConsent.PurposeAllowed(consentconstants.AdSelectionDeliveryReporting) && parsedConsent.VendorConsent(vendorID) {
+		if (vendor.Purpose(tcf1constants.InfoStorageAccess) || vendor.LegitimateInterest(tcf1constants.InfoStorageAccess)) && parsedConsent.PurposeAllowed(tcf1constants.InfoStorageAccess) && (vendor.Purpose(tcf1constants.AdSelectionDeliveryReporting) || vendor.LegitimateInterest(tcf1constants.AdSelectionDeliveryReporting)) && parsedConsent.PurposeAllowed(tcf1constants.AdSelectionDeliveryReporting) && parsedConsent.VendorConsent(vendorID) {
 			return true, nil
 		}
 	}
--- a/gdpr/impl.go
+++ b/gdpr/impl.go
@@ -20,7 +20,7 @@ import (
 type permissionsImpl struct {
 	cfg             config.GDPR
 	vendorIDs       map[openrtb_ext.BidderName]uint16
-	fetchVendorList []func(ctx context.Context, id uint16) (vendorlist.VendorList, error)
+	fetchVendorList map[uint8]func(ctx context.Context, id uint16) (vendorlist.VendorList, error)
 }
 
 func (p *permissionsImpl) HostCookiesAllowed(ctx context.Context, consent string) (bool, error) {
@@ -122,7 +122,7 @@ func (p *permissionsImpl) parseVendor(ctx context.Context, vendorID uint16, cons
 	if version < 1 || version > 2 {
 		return
 	}
-	vendorList, err := p.fetchVendorList[version-1](ctx, parsedConsent.VendorListVersion())
+	vendorList, err := p.fetchVendorList[version](ctx, parsedConsent.VendorListVersion())
 	if err != nil {
 		return
 	}
--- a/gdpr/impl.go
+++ b/gdpr/impl.go
@@ -3,6 +3,7 @@ package gdpr
 import (
 	"context"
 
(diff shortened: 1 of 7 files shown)


gdpr/impl.go
```diff
@@ -110,7 +117,7 @@ func (p *permissionsImpl) parseVendor(ctx context.Context, vendorID uint16, cons
 		return
 	}
 
-	vendorList, err := p.fetchVendorList(ctx, parsedConsent.VendorListVersion())
+	vendorList, err := p.fetchVendorList[parsedConsent.Version()-1](ctx, parsedConsent.VendorListVersion())
```
SyntaxNode: This should be safe. If version is anything other than 1 or 2, we would get an error from the library and not run this code.. right?

gdpr/vendorlist-fetching.go
```diff
@@ -11,35 +11,37 @@ import (
...
+type saveVendors func(uint16, api.VendorList)
 
 // This file provides the vendorlist-fetching function for Prebid Server.
 //
 // For more info, see https://github.com/prebid/prebid-server/issues/504
 //
 // Nothing in this file is exported. Public APIs can be found in gdpr.go
 
-func newVendorListFetcher(initCtx context.Context, cfg config.GDPR, client *http.Client, urlMaker func(uint16) string) func(ctx context.Context, id uint16) (vendorlist.VendorList, error) {
+func newVendorListFetcher(initCtx context.Context, cfg config.GDPR, client *http.Client, urlMaker func(uint16, int) string, tCFVer int) func(ctx context.Context, id uint16) (vendorlist.VendorList, error) {
```
SyntaxNode: Could we just name it TCFVer here? .. since capitalization doesn't affect method parameters.

gdpr/vendorlist-fetching.go
```diff
@@ -11,35 +11,37 @@ import (
...
 	// These save and load functions can be used to store & retrieve lists from our cache.
 	save, load := newVendorListCache()
 
 	withTimeout, cancel := context.WithTimeout(initCtx, cfg.Timeouts.InitTimeout())
 	defer cancel()
 	populateCache(withTimeout, client, urlMaker, save)
 
-	saveOneSometimes := newOccasionalSaver(cfg.Timeouts.ActiveTimeout())
+	saveOneSometimes := newOccasionalSaver(cfg.Timeouts.ActiveTimeout(), ctfVer)
```
SyntaxNode: Is ctfVer a typo of tcfVer? Those should be easy to sort through once the go-gdpr library is updated. I'll comment no further on it.

gdpr/impl.go
hhhjort: Actually once TCF 3.0 comes out, that definitely won't be true. We should do a check on this side.

gdpr/vendorlist-fetching.go
hhhjort: Done

gdpr/vendorlist-fetching.go
hhhjort: Yes, I am surprised the parser did not flag this.

gdpr/vendorlist-fetching.go
```diff
@@ -71,7 +79,7 @@ func vendorListURLMaker(version uint16) string {
 // The goal here is to update quickly when new versions of the VendorList are released, but not wreck
 // server performance if a bad CMP starts sending us malformed consent strings that advertize a version
 // that doesn't exist yet.
-func newOccasionalSaver(timeout time.Duration) func(ctx context.Context, client *http.Client, url string, saver saveVendors) {
+func newOccasionalSaver(timeout time.Duration, cTFVer int) func(ctx context.Context, client *http.Client, url string, saver saveVendors) {
 	lastSaved := &atomic.Value{}
```
guscarreon: Function `newOccasionalSaver` seems to not be passing `cTFVer int` to the `saveOne` call. In order to make sure
we are not missing anything the compiler wouldn't catch, I recommend merging the `prebid/go-gdpr` changes
first before commiting this one.
```
func newOccasionalSaver(timeout time.Duration, cTFVer int) func(ctx context.Context, client *http.Client, url string, saver saveVendors) {
	lastSaved := &atomic.Value{}
	lastSaved.Store(time.Time{})

	return func(ctx context.Context, client *http.Client, url string, saver saveVendors) {
		now := time.Now()
		if now.Sub(lastSaved.Load().(time.Time)).Minutes() > 10 {
			withTimeout, cancel := context.WithTimeout(ctx, timeout)
			defer cancel()
			saveOne(withTimeout, client, url, saver) //<-- missing `cTFVer` that depicts the GDPR version and determines the URL to send the http request to
			lastSaved.Store(now)
		}
	}
}
```

gdpr/impl.go
```diff
@@ -18,7 +19,7 @@ import (
 type permissionsImpl struct {
 	cfg             config.GDPR
 	vendorIDs       map[openrtb_ext.BidderName]uint16
-	fetchVendorList func(ctx context.Context, id uint16) (vendorlist.VendorList, error)
+	fetchVendorList []func(ctx context.Context, id uint16) (vendorlist.VendorList, error)
```
SyntaxNode: Probably best to make this a map. Otherwise it will start to break down when TCF 1 support is removed.

gdpr/vendorlist-fetching.go
hhhjort: New go-gdpr library updated and all the tests and issues fixed

gdpr/impl.go
```diff
@@ -93,14 +95,20 @@ func (p *permissionsImpl) allowPI(ctx context.Context, vendorID uint16, consent
 		return false, nil
 	}
 
-	if (vendor.Purpose(consentconstants.InfoStorageAccess) || vendor.LegitimateInterest(consentconstants.InfoStorageAccess)) && parsedConsent.PurposeAllowed(consentconstants.InfoStorageAccess) && (vendor.Purpose(consentconstants.AdSelectionDeliveryReporting) || vendor.LegitimateInterest(consentconstants.AdSelectionDeliveryReporting)) && parsedConsent.PurposeAllowed(consentconstants.AdSelectionDeliveryReporting) && parsedConsent.VendorConsent(vendorID) {
-		return true, nil
+	if parsedConsent.Version() == 2 {
+		// Need to add the location special purpose once the library supports it.
+		if (vendor.Purpose(consentconstants.InfoStorageAccess) || vendor.LegitimateInterest(consentconstants.InfoStorageAccess)) && parsedConsent.PurposeAllowed(consentconstants.InfoStorageAccess) && (vendor.Purpose(consentconstants.PersonalizationProfile) || vendor.LegitimateInterest(consentconstants.PersonalizationProfile)) && parsedConsent.PurposeAllowed(consentconstants.PersonalizationProfile) && parsedConsent.VendorConsent(vendorID) {
```
SyntaxNode: Does this code have test coverage?

SyntaxNode: Discussed in person. There are no tests for the TCF 1 version either. This needs to be addressed and can be done in a separate PR. We'll manually run end to end tests to verify functionality for now.

PR #1229, opened 2020-03-20
added VISX vendor ID for usersyncing

Aadeshp:


mk0x9:
@SyntaxNode change is approved.

Issue #1215, opened 2020-03-06
Optional pricegranularity.ranges parameter is enforced on openrtb2/video endpoint


joshuahaertel:
# Observed Behavior
Sending `"pricegranularity": {}` to the `openrtb2/video` endpoint results in a 500 error with the `empty granularity definition supplied` error message, while omitting `pricegranularity` appears to work.

# Expected Behavior
According to http://prebid.org/prebid-server/endpoints/openrtb2/video.html the `ranges` value isn't necessary. Regardless, `{}` should be just as valid as not sending anything.

# Suggested Fix
Instead of returning the error here:
https://github.com/prebid/prebid-server/blob/fd78c23caef9986556b8558443a7e0cb91831a3a/openrtb_ext/request.go#L152

Use the same default value used when nothing is passed in: https://github.com/prebid/prebid-server/blob/fd78c23caef9986556b8558443a7e0cb91831a3a/openrtb_ext/request.go#L111

It might not be a bad idea to make sure the same thing happens when `null` is passed in.

SyntaxNode:
@camrice Can you please take a look at this?

camrice:
This change sounds good to me. It will be more in line with the documentation. Let me bring this up with the team here.

SyntaxNode:
@joshuahaertel This has been implemented and released. Does this resolve your concern?

joshuahaertel:
Yes, it works great. Thank you!

PR #1230, opened 2020-03-20
Updated price granularity unmarshal to accept empty values and ranges

camrice:
Original issue: https://github.com/prebid/prebid-server/issues/1215

Updated the unmarshal for price granularity to use the default object if the request object is empty, or contains no ranges.

### DIFF
--- a/openrtb_ext/request.go
+++ b/openrtb_ext/request.go
@@ -148,10 +148,11 @@ func (pg *PriceGranularity) UnmarshalJSON(b []byte) error {
 			}
 			prevMax = gr.Max
 		}
-	} else {
-		return errors.New("Price granularity error: empty granularity definition supplied")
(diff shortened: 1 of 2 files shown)


hhhjort:
I am a little confused for the need for this? If there is no price granularity supplied, we will already use medium. This just covers the case where an empty or unparseable granularity is supplied? In the unparseable case, I think it may be better to return an error. What is the use case for someone supplying `{}` as the entry?

camrice:
This one should still catch an unmashaling error and return that instead of using the default price granularity. This change was made to match the documentation where the `precision` and `ranges` fields are optional. So as well as accepting a price granularity defined as `{}`, it would also allow ones where the `ranges` are not specified or `ranges` is an empty array. Let me know if this seems like too much of a corner case, or if there should be different behavior for these cases.

PR #1236, opened 2020-03-31
Typos Fix

SyntaxNode:
Follow-up from https://github.com/prebid/prebid-server/pull/1187

### DIFF
--- a/main.go
+++ b/main.go
@@ -42,9 +42,11 @@ func main() {
 	}
 }
 
+const configFileName = "pbs"
+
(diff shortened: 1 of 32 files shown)


main.go
```diff
@@ -42,9 +42,11 @@ func main() {
 	}
 }
 
+const configFileName = "pbs"
```
hhhjort, line 45: A little more than just a typo fix here. ;)

SyntaxNode: Busted!

PR #1238, opened 2020-03-31
#615 Beachfront URLs from config

kdesput:
Added getting Beachfront URLs from config


config/config.go
```diff
@@ -247,6 +247,9 @@ type Adapter struct {
 	// needed for Facebook
 	PlatformID string `mapstructure:"platform_id"`
 	AppSecret  string `mapstructure:"app_secret"`
+
+	// needed for Beachfront adapter with separate endpoint for video
+	VideoEndpoint string `mapstructure:"video_endpoint"`
```
hhhjort, line 252: We now have the `ExtraAdapterInfo`  config object which is meant to be a JSON string that can contain extra adapter config data. We are encouraging adapters to use this instead of adding on extra parameters to the main object, so we don't balloon out to a dozen new single-adapter fields in this config object.

muncha: Just a note that beachfront has noted this. Thank you.

adapters/beachfront/beachfront.go
```diff
@@ -663,19 +677,21 @@ func getIP(ip string) string {
...
-	return openrtb_ext.BidTypeBanner
-}
-
 func removeVideoElement(slice []beachfrontVideoRequest, s int) []beachfrontVideoRequest {
 	return append(slice[:s], slice[s+1:]...)
 }
 
-func NewBeachfrontBidder() *BeachfrontAdapter {
-	return &BeachfrontAdapter{}
+func NewBeachfrontBidder(bannerEndpoint string, extraAdapterInfo string) adapters.Bidder {
```
hhhjort, line 684: You may want to either provide a default video endpoint, or put some checks in to not attempt to bid on video traffic if one is not available. I can see some issues occurring if the adapter is trying to generate requests to a blank URL.

hhhjort:
Other than the point on handling the case of a missing video endpoint, it all looks good to me. Sorry for not catching that point sooner.

adapters/beachfront/beachfront.go
kdesput: I added the default video endpoint URL. Its value is used when the `extraAdapterInfo` is empty or it doesn't contain the value for video endpoint URL.

PR #1243, opened 2020-04-02
Updated condition for setting brand_category_uniqueness flag

camrice:
`brand_category_uniqueness` was always being set to enabled, as we created a default `includebrandcategory` in the video endpoint. This update omits the flag if no `includebrandcategory` is set.


SyntaxNode:
Was this a bug? I'm wondering if this fix will invalidate existing configs / introduce breaking changes.

camrice:
Yes, this was brought to our attention as a bug. I agree that this needs more verification. There isn't a major rush for this change, so let me plan a path forward for making sure that this doesn't break anything existing.

PR #1252, opened 2020-04-08
Bugfix: no bids from bidder handling

VeronikaSolovei9:


exchange/exchangetest/request-multi-bidders-one-no-resp.json
```diff
@@ -0,0 +1,169 @@
...
+            },
+            "audienceNetwork": {
+              "placementId": "some-other-placement"
+            }
+          }
+        }
+      ],
+      "ext": {
+        "prebid": {
+          "targeting": {}
```
camrice, line 40: Just realized, this requests needs the `includebrandcategory` in the `targeting` in order to represent the fix, correct? That way, the processing will trigger `applyCategoryMapping` where you pointed out the crash initially.

exchange/exchangetest/request-multi-bidders-one-no-resp.json
```diff
@@ -0,0 +1,169 @@
...
+              },
+              "bidType": "video"
+            }
+          ]
+        }
+      }
+    },
+    "audienceNetwork": {
+      "mockResponse": {
+        "pbsSeatBid": {
```
camrice, line 89: I believe that if `pbsSeatBid` is initialized like this, it will be an array of zero length and so it won't reach the crash case without your fix. Instead, should this be ` "pbsSeatBid": null`?

exchange/exchangetest/request-multi-bidders-one-no-resp.json
VeronikaSolovei9: You are right, fixed!

exchange/exchangetest/request-multi-bidders-one-no-resp.json
VeronikaSolovei9: Fixed!

exchange/exchangetest/request-multi-bidders-one-no-resp-test1.json
```diff
@@ -0,0 +1,227 @@
+{
```
camrice, line 1: Small change, but I think the filename should be changed to mark that this test is checking for debug info, just to make the purpose of this test file clearer.

VeronikaSolovei9: Done!

exchange/exchange.go
```diff
@@ -63,6 +63,9 @@ type exchange struct {
 type seatResponseExtra struct {
 	ResponseTimeMillis int
 	Errors             []openrtb_ext.ExtBidderError
+	// httpCalls is the list of debugging info. It should only be populated if the request.test == 1.
+	// This will become response.ext.debug.httpcalls.{bidder} on the final Response.
+	httpCalls []*openrtb_ext.ExtHttpCall
```
SyntaxNode, line 68: This member begins with a lower-case letter indicating that's private. The other members are public. I'm curious as to why this one is treated differently.

exchange/exchange.go
```diff
@@ -344,7 +351,12 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
 	// Wait for the bidders to do their thing
 	for i := 0; i < len(cleanRequests); i++ {
 		brw := <-chBids
-		adapterBids[brw.bidder] = brw.adapterBids
+
+		//if bidder returned no bids back - remove bidder from further processing
+		if brw.adapterBids != nil && len(brw.adapterBids.bids) != 0 {
```
SyntaxNode, line 356: What are the consequences of removing the adapter bids entirely (do the metrics still record properly?) versus setting the bids to an empty slice?

exchange/exchange.go
```diff
@@ -637,19 +649,20 @@ func (e *exchange) makeExtBidResponse(adapterBids map[openrtb_ext.BidderName]*pb
 		}
 	}
 
-	for a, b := range adapterBids {
-		if b != nil && req.Test == 1 {
+	for a, b := range adapterExtra {
```
SyntaxNode, line 652: Consider renaming variables for clarity. Something along the lines of:
```
for bidderName, reasponseExtra := range adapterExtra
```

I * think * before `a` was for adapter and `b` was for bids. Since we're changing the data types might be a good opportunity to make these variable names clearer.

exchange/exchange.go
```diff
@@ -637,19 +649,20 @@ func (e *exchange) makeExtBidResponse(adapterBids map[openrtb_ext.BidderName]*pb
...
+
+		if req.Test == 1 {
 			// Fill debug info
```
SyntaxNode, line 655: Nitpick: Consider removing comment.

exchange/exchange.go
```diff
@@ -322,6 +325,10 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
 			// Structure to record extra tracking data generated during bidding
 			ae := new(seatResponseExtra)
 			ae.ResponseTimeMillis = int(elapsed / time.Millisecond)
+			if request.Test == 1 && bids != nil {
```
SyntaxNode, line 328: Consider removing the `request.Test` check here to always capture the information (there's no overhead for doing so) and letting downstream code decide what to do with it (as it already does with repeating this check).

exchange/exchange.go
VeronikaSolovei9: I made it private same as we already have it in `pbsOrtbSeatBid struct`. But here, yes, it should be public. Good catch! Fixed.

exchange/exchange.go
VeronikaSolovei9: We don't want to carry bidder data if there are no bids back. For loops will skip empty array iterating and nothing bad will happen. This is just memory and processor time optimization. Hope this makes sense.

exchange/exchange.go
VeronikaSolovei9: Done!

exchange/exchange.go
VeronikaSolovei9: Done!

exchange/exchange.go
VeronikaSolovei9: Done!

PR #1256, opened 2020-04-10
replacing info@prebid.org maintainer email addrs

bretg:
info@prebid.org received email for one of these adapters. Fixing the addresses.

### DIFF
--- a/static/bidder-info/audienceNetwork.yaml
+++ b/static/bidder-info/audienceNetwork.yaml
@@ -1,5 +1,5 @@
 maintainer:
-  email: "info@prebid.org"
+  email: "none"
 capabilities:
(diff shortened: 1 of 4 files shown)


static/bidder-info/audienceNetwork.yaml
```diff
@@ -1,5 +1,5 @@
 maintainer:
-  email: "info@prebid.org"
+  email: "none"
```
SyntaxNode, line 2: This is fine. AppNexus/Xandr built and recently updated this adapter, but don't have an official relationship with Facebook. We'll continue to keep an eye on related issues. If Facebook would later like to own it, all the better. :)

PR #1257, opened 2020-04-13
Add kidoz bidder info

bretg:
got this info from email communication with kidoz

### DIFF
--- a/docs/bidders/kidoz.md
+++ b/docs/bidders/kidoz.md
@@ -1,6 +1,6 @@
 # Kidoz Bidder
 
-Kidoz is exclusively for Mobile app coppa compatible ads, 100% kid relevant and appropriate.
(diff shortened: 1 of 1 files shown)


docs/bidders/kidoz.md
```diff
@@ -0,0 +1,9 @@
+# Kidoz Bidder
+
+Kidoz is exclusively for Mobile app coppa compatible ads, 100% kid relevant and appropriate.
```
SyntaxNode, line 3: Could we slightly fix the capitalization? 
`Kidoz is exclusively for mobile app COPPA compatible`

bretg: done

SyntaxNode: Thanks.

PR #1260, opened 2020-04-15
Add nil check errors when setting native asset types

mansinahar:


exchange/bidder.go
```diff
@@ -255,13 +272,13 @@ func getNativeImpByImpID(impID string, request *openrtb.BidRequest) (*openrtb.Na
 	return nil, errors.New("Could not find native imp")
 }
 
-func getAssetByID(id int64, assets []nativeRequests.Asset) nativeRequests.Asset {
+func getAssetByID(id int64, assets []nativeRequests.Asset) (*nativeRequests.Asset, error) {
 	for _, asset := range assets {
 		if id == asset.ID {
-			return asset
+			return &asset, nil
```
SyntaxNode, line 278: Curious. Why the switch to a pointer?

exchange/bidder.go
```diff
@@ -255,13 +272,13 @@ func getNativeImpByImpID(impID string, request *openrtb.BidRequest) (*openrtb.Na
...
 		}
 	}
-	return nativeRequests.Asset{}
+	return nil, fmt.Errorf("Couldn't find asset with ID:%d in the request", id)
```
SyntaxNode, line 281: Consider rewording to a slightly more professional message. Something along the lines of "Unable to find asset with ID:%d in the request."

exchange/bidder_test.go
```diff
@@ -1083,6 +1086,141 @@ func TestErrorReporting(t *testing.T) {
 	}
 }
 
+func TestSetAssetTypes(t *testing.T) {
+	testCases := []struct {
+		respAsset nativeResponse.Asset
+		nativeReq nativeRequests.Request
+		errStr    string
```
SyntaxNode, line 1093: Please rename to avoid Hungarian notation. Consider "expectedError"

SyntaxNode:
> Given the length of `TestSetAssetTypes`, it'd be desirable to have a description of every test case. Nevertheless things look pretty solid and test cases seem to tackle most scenarios. Approved

It was a bit difficult to follow. I think a description of the test case would be a good idea.

exchange/bidder.go
mansinahar: This function was initially not returning an error and therefore I changed that to a pointer to be able to do a nil check where this function is called but then realized that it's better to return an error and ended up keeping it as a pointer as it doesn't matter much either way in this case.

Nevertheless have changed it back and updated the PR.

mansinahar:
That's a good point about test descriptions @SyntaxNode and @guscarreon. I agree and have updated the PR. Thanks for your feedback!

PR #1264, opened 2020-04-18
Add missing categories to AppNexus -> IAB mapping file.

jmaynardxandr:


static/adapter/appnexus/opts.json
```diff
@@ -19,9 +19,12 @@
     "18": "IAB13-6",
     "19": "IAB18-4",
     "20": "IAB1-5",
+    "22": "IAB19-28",
```
hhhjort, line 22: Odd that "22" is slightly out of order. Otherwise seems fine.

jmaynardxandr: Thanks for catching that, I fixed the order, my brain's not as good at sorting as I'd like. :)

PR #1265, opened 2020-04-20
populate the app ID in the FAN timeout notif url with the publisher ID and the auction with the request ID

Aadeshp:


adapters/audienceNetwork/facebook.go
```diff
@@ -455,18 +460,39 @@ func NewFacebookBidder(client *http.Client, platformID string, appSecret string)
...
 	if err != nil {
 		return &adapters.RequestData{}, []error{err}
 	}
 
-	uri := fmt.Sprintf("https://www.facebook.com/audiencenetwork/nurl/?partner=%s&app=%s&auction=%s&ortb_loss_code=2", fa.platformID, fa.platformID, auction_id)
+	// The publisher ID is either in the app object or the site object, depending on the supply of the request so we need
+	// to check both
+	pubID, err = jsonparser.GetString(req.Body, "app", "publisher", "id")
+	if err != nil {
+		pubID, err = jsonparser.GetString(req.Body, "site", "publisher", "id")
```
SyntaxNode, line 481: What if the app.publisher.id exists and is empty?

adapters/audienceNetwork/facebook_test.go
```diff
@@ -45,7 +45,7 @@ func TestJsonSamples(t *testing.T) {
 
 func TestMakeTimeoutNotice(t *testing.T) {
 	req := adapters.RequestData{
```
SyntaxNode, line 47: Please add a unit test for getting the pubID from the site.publisher.id.

adapters/audienceNetwork/facebook.go
Aadeshp: we would have rejected the impression — we validate that there is a publisher ID before sending a request to FAN (in extractPlacementAndPublisher)

hhhjort: We need to consider the case where the request comes from somewhere other than Appnexus.

SyntaxNode: The validation @Aadeshp mentions isn't specific to AppNexus, is it? The `extractPlacementAndPublisher` method is in the PBS adapter.

hhhjort: Oh, yes, the validation is in the facebook adapter. So it should be safe.

PR #1268, opened 2020-04-21
Added header User Agent decoding

VeronikaSolovei9:


endpoints/openrtb2/video_auction.go
```diff
@@ -617,7 +618,9 @@ func (deps *endpointDeps) parseVideoRequest(request []byte, headers http.Header)
 
 	//if Device.UA is not present in request body, init it with user-agent from request header if it's present
 	if req.Device.UA == "" {
-		req.Device.UA = headers.Get("User-Agent")
+		ua := headers.Get("User-Agent")
+		ua, _ = url.QueryUnescape(ua)
```
SyntaxNode, line 622: This is a change which affects everyone. How will this behave when the user agent is not url encoded? Also, please never ignore errors.

jmaynardxandr: fyi we tested with a value that's not encoded and QueryUnescape() returns the same value in that case. In the unlikely event that an unencoded value contains a decodable character sequence the return wouldn't match, but that's probably a pretty extreme edge case.

VeronikaSolovei9: Yes, fixed

SyntaxNode: Looking at samples of UA strings, the '+' character may sometimes appear and the '%' character really doesn't. The `+` character would get changed to a space if not actually encoded. Perhaps you can check for the presence of a '%' as a best guess if the UA is encoded? Whatever you think is the best choice for this usecase. Please document your decision / explanation in a code block and also in the docs. This is one of those small hidden things which may infuriate someone down the road trying to debug.

GoLang already checks for invalid escape sequences and errors out, which now causes you to fallback to the original string.. which is good.

VeronikaSolovei9: Good point! Check added. 
I also added a comment in code. Do you want me to mention it in prebid.org docs?

SyntaxNode: Comment is good for this PR. Thank you. I would like it to be mention in prebid.org docs as well, but I see that as a follow-up and not a blocker.

endpoints/openrtb2/video_auction.go
```diff
@@ -617,7 +618,18 @@ func (deps *endpointDeps) parseVideoRequest(request []byte, headers http.Header)
 
 	//if Device.UA is not present in request body, init it with user-agent from request header if it's present
 	if req.Device.UA == "" {
-		req.Device.UA = headers.Get("User-Agent")
+		ua := headers.Get("User-Agent")
+
+		//Check UA is encoded. Without it the `+` character would get changed to a space if not actually encoded
+		if strings.ContainsAny(ua, "%") {
+			var err error
+			req.Device.UA, err = url.QueryUnescape(ua)
```
hhhjort, line 626: This feels a little clumsy. Should we assign the `url.QueryUnescape()` result to a variable, and then assign it to `req.Device.UA` after we check `err == nil`? I am open on this question, can't quite envision a sequence that is definitely "not clumsy".

jmaynardxandr: We just did this way to avoid having to create a 3rd variable for it if we need to set it back to the original value for the error case, unless QueryUnescape() returns the original string in an error case then we can go back to using ua there and always set it (without needing to check the error in that case since either way ua would be what we want to use).

SyntaxNode: `req.Device.UA` needs to be assigned in any case. We're copying the UA from the http headers to the OpenRTB request. I thought it looked fine. I'm also fine if you want to use an intermediate variable.

It's not that we're changing `req.Device.UA` to possibly nil and then back to ua. It's nil from the start.

hhhjort: leave it it as is then. It looks clumsy to my eyes, but I couldn't find a better suggestion, so I opened it up to see if anyone else could see one.

SyntaxNode: Can always think on it and iterate in the future.

PR #1270, opened 2020-04-22
Fix debug log error messages

camrice:
This PR fixes the formatting of the debug log, specifically for browsers. It also fixes an issue where deal support and cache errors were not being appended to the response.

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -34,14 +34,14 @@ import (
 
 var defaultRequestTimeout int64 = 5000
 
-func NewVideoEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, videoFetcher stored_requests.Fetcher, categories stored_requests.CategoryFetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte, bidderMap map[string]openrtb_ext.BidderName, cache prebid_cache_client.Client) (httprouter.Handle, error) {
+func NewVideoEndpoint(ex exchange.Exchange, validator openrtb_ext.BidderParamValidator, requestsById stored_requests.Fetcher, videoFetcher stored_requests.Fetcher, categories stored_requests.CategoryFetcher, cfg *config.Configuration, met pbsmetrics.MetricsEngine, pbsAnalytics analytics.PBSAnalyticsModule, disabledBidders map[string]string, defReqJSON []byte, bidderMap map[string]openrtb_ext.BidderName, cache prebid_cache_client.Client, debugLogRegexp *regexp.Regexp) (httprouter.Handle, error) {
 
 	if ex == nil || validator == nil || requestsById == nil || cfg == nil || met == nil {
 		return nil, errors.New("NewVideoEndpoint requires non-nil arguments.")
 	}
 	defRequest := defReqJSON != nil && len(defReqJSON) > 0
 
-	return httprouter.Handle((&endpointDeps{ex, validator, requestsById, videoFetcher, categories, cfg, met, pbsAnalytics, disabledBidders, defRequest, defReqJSON, bidderMap, cache}).VideoAuctionEndpoint), nil
+	return httprouter.Handle((&endpointDeps{ex, validator, requestsById, videoFetcher, categories, cfg, met, pbsAnalytics, disabledBidders, defRequest, defReqJSON, bidderMap, cache, debugLogRegexp}).VideoAuctionEndpoint), nil
 }
 
 /*
@@ -93,7 +93,7 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		Enabled:   strings.EqualFold(debugQuery, "true"),
 		CacheType: prebid_cache_client.TypeXML,
 		TTL:       cacheTTL,
-		Matcher:   regexp.MustCompile(`[<>]`),
+		Regexp:    deps.debugLogRegexp,
 	}
 
 	defer func() {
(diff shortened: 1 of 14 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -130,11 +131,11 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 
 	resolvedRequest := requestJson
 	if debugLog.EnableDebug {
-		debugLog.Data = fmt.Sprintf("Request:\n%s", string(requestJson))
```
SyntaxNode, line 133: This is starting to get bit complicated and isn't covered by tests. Please pull out into helper/util funcs to test outside of the auction flow.

camrice: Updated implementation to be cleaner and use a few helper functions. Let me know if there is further cleanup I should do.

endpoints/openrtb2/video_auction.go
```diff
@@ -131,11 +118,11 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+		debugLog.Data.Request = string(requestJson)
 		if headerBytes, err := json.Marshal(r.Header); err == nil {
-			debugLog.Data = fmt.Sprintf("%s\n\nHeaders:\n%s", debugLog.Data, string(headerBytes))
+			debugLog.Data.Headers = string(headerBytes)
 		} else {
-			debugLog.Data = fmt.Sprintf("%s\n\nUnable to marshal headers data\n", debugLog.Data)
+			debugLog.Data.Headers = "Unable to marshal headers data"
```
SyntaxNode, line 125: Consider including the `err.Error()` message to figure out what went wrong.

endpoints/openrtb2/video_auction.go
```diff
@@ -281,6 +268,33 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+
+	toCache := []prebid_cache_client.Cacheable{
+		{
+			Type:       debugLog.CacheType,
+			Data:       data,
+			TTLSeconds: debugLog.TTL,
+			Key:        "log_" + debugLog.CacheKey,
+		},
+	}
+	if cache != nil {
```
SyntaxNode, line 289: Super Nitpick: Consider adding a line break prior to the `cache != nil` check.

exchange/auction.go
```diff
@@ -15,6 +16,29 @@ import (
 	"github.com/prebid/prebid-server/prebid_cache_client"
 )
 
+type DebugLog struct {
+	EnableDebug bool
```
SyntaxNode, line 20: There is a duplicate of the Debug name. Consider renaming to `Enabled`.

exchange/auction.go
```diff
@@ -15,6 +16,29 @@ import (
...
+}
+
+type DebugData struct {
+	Request  string
+	Headers  string
+	Response string
+}
+
+func (d *DebugLog) BuildDebugLog() {
+	d.Data.Request = fmt.Sprintf("<Request>%s</Request>", d.Data.Request)
```
SyntaxNode, line 35: Is it possible for the request, headers, or response string contains invalid xml?

exchange/auction.go
```diff
@@ -15,6 +16,29 @@ import (
...
+	CacheString string
+}
+
+type DebugData struct {
+	Request  string
+	Headers  string
+	Response string
+}
+
+func (d *DebugLog) BuildDebugLog() {
```
SyntaxNode, line 34: Consider renaming to something like `BuildCacheString` to make it clear what field is being modified.

endpoints/openrtb2/video_auction.go
camrice: Updated to include the error

endpoints/openrtb2/video_auction.go
camrice: Updated formatting

exchange/auction.go
camrice: Renamed to `Enabled`

exchange/auction.go
camrice: Added a regex replacer for `<>` characters in case. This is used before the tags are added, so it should fix formatting issues if those characters are included. Let me know if this is too aggressive of a replacer.

exchange/auction.go
camrice: Renamed to `BuildCacheString`

camrice:
Thank you! Really sorry for the slow progress on this one. Adding changes from comments above.

exchange/auction.go
hhhjort: Probably want to compile the `matcher` when you build the `DebugLog` object, and store it there, so we don't have to recompile every time this function is called.

camrice: Added `Matcher` as a new field. Updated the construction to compile the regex.

SyntaxNode: Awesome. The name 'Matcher' seems a bit ambiguous at that level, but other names I would suggest are awful.. like CacheStringNormlizationMatcher. Thoughts @hhhjort?

hhhjort: A little bit of a descriptive name would be good, especially in light of my next suggestion.

hhhjort: We probably should look into moving the matcher all the way into the `endpointDeps`, as it stands now we will recompile this for every request still, just locked down to only once per request. The other endpoints should be able to ignore the additional `endpointDeps` field, at least until they have a reason to use it.

camrice: Ah good point! Moved the regexp to be in `endpointDeps` and renamed the field from `Matcher` to just `Regexp`.

hhhjort: You could move the compiling into `NewVideoEndpoint()` rather than compiling it router and passing it in. Saves one argument in the function, but that is a minor nit pick at this point.

camrice:
Ah yeah taking another look, I could move the initialization of the regex to `NewVideoEndpoint` if we needed.

SyntaxNode:
Yeah, you might want to implement the suggestion by @hhhjort to move it down to the right level inside `NewVideoEndpoint`. This is really just a performance optimization, not a true injectable dependency.

camrice:
Updated regex initialization to `NewVideoEndpoint` as commented

PR #1273, opened 2020-04-23
Add AdOcean adapter

mmuras:
Add a new adapter for AdOcean


adapters/adocean/adocean.go
```diff
@@ -0,0 +1,317 @@
...
+					(new Image(1,1)).src = wu
+				}
+
+				if (su && !(navigator.sendBeacon && navigator.sendBeacon(su))) {
+					(new Image(1,1)).src = su
+				}
+			}();
+		</script>
+	`
+	whiteSpace := regexp.MustCompile(`\s+`)
```
hhhjort, line 312: I am thinking you may want to add this value as a parameter of your adapter object, and make this function a method of that object. Then you could compile `whitespace` just once on startup, rather than recompiling it every request.

mmuras: Thanks, it's changed. I've also added additional if to check if this object is not nil (on line 146), and corrected a little oversight at line 177 (instead of using newUri variable new string was generated).

mmuras:
@guscarreon do you have some estimate when you will have time for review?

adapters/adocean/adocean.go
```diff
@@ -0,0 +1,322 @@
...
+		if err := json.Unmarshal(request.User.Ext, &extUser); err == nil {
+			consentString = extUser.Consent
+		}
+	}
+
+	var httpRequests []*adapters.RequestData
+	for _, auction := range request.Imp {
+		newBidRequest, err := a.makeRequest(httpRequests, &auction, request, consentString)
+		if err != nil {
+			return nil, []error{err}
```
guscarreon, line 96: If there's even one element of the `[]Imp` array with an error, we are discarding the rest of them. Do we want this? Or did we mean to just `continue` and try to check the rest of the array and only append the "good ones" to `httpRequests`

config/config.go
```diff
@@ -499,6 +499,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderAdpone, "https://usersync.adpone.com/csync?redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dadpone%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%7Buid%7D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderAdtelligent, "https://sync.adtelligent.com/csync?t=p&ep=0&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dadtelligent%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%7Buid%7D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderAdmixer, "https://inv-nets.admixer.net/adxcm.aspx?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redir=1&rurl="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dadmixer%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24%24visitor_cookie%24%24")
+	// openrtb_ext.BidderAdOcean doesn't have a good default.
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderAdvangelists, "https://nep.advangelists.com/xp/user-sync?acctid={aid}&&redirect="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dadvangelists%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
```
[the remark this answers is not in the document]
mmuras: For now no, we expect that clients will set specific for them url in the configuration file (e.g. pbs.yaml):
`adapters:
    adocean:
        usersync_url: https://our-matchning-host/?url=https://client-prebid-server-instance/setuid?bidder=adocean&uid=$userID
`
It may change in the future, but now it's the preferred way.

adapters/adocean/adocean.go
mmuras: I've made changes so it will be more fault tolerant.

mmuras:
@hhhjort I've made some small changes (if there is error in one bid, the rest won't be discarded) after @guscarreon review, I think I need your approval again.

adapters/adocean/adocean.go
```diff
@@ -0,0 +1,324 @@
...
+	var parsedResponses = adapters.NewBidderResponseWithBidsCapacity(len(auctionIDs))
+	var errors []error
+
+	for _, auctionFullID := range auctionIDs {
+		auctionIDsSlice := strings.SplitN(auctionFullID, ":", 2)
+		auctionID := auctionIDsSlice[1]
+		slaveID := auctionIDsSlice[0]
+
+		for _, bid := range bidResponses {
+			if bid.ID == slaveID {
```
guscarreon, line 279: Small optimization. If it makes sense in the context of this adapter, we could go from the nested for loops of O(nm) time to O(n+m) consecutive loops at the cost of a little memory (currently, `prebid-server` can spare a bit more memory in return of less processing time). Again, this is only a suggestion and I'm ok if you think it doesn't make sense in this context.
```
252   func (a *AdOceanAdapter) MakeBids(
253       internalRequest *openrtb.BidRequest,
254       externalRequest *adapters.RequestData,
255       response *adapters.ResponseData,
256   ) (*adapters.BidderResponse, []error) {
257   *-- 15 lines: if response.StatusCode != http.StatusOK {---------------------
272  
    +     idMap := new(map[string]string,len(auctionIDs))
273       for _, auctionFullID := range auctionIDs {
274           auctionIDsSlice := strings.SplitN(auctionFullID, ":", 2)
275 -         auctionID := auctionIDsSlice[1]
276 -         slaveID := auctionIDsSlice[0]
    +         idMap[auctionIDsSlice[0]] = auctionIDsSlice[1]
    +     }
277  
278           for _, bid := range bidResponses {
279 -             if bid.ID == slaveID {
    +             if auctionID, found := idMap[bid.ID]; found {
280                   if bid.Error == "true" {
281                       break
282                   }
283  
284   *--  21 lines: price, _ := strconv.ParseFloat(bid.Price, 64)----------------
305                   parsedResponses.Currency = bid.Currency
306  
307                   break
308               }
309           }
310       }
311  
312       return parsedResponses, errors
313   }
adapters/adocean/adocean.go
```

adapters/adocean/adocean.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	if params.SlaveID == "" {
+		return &errortypes.BadInput{
+			Message: "SlaveId undefined",
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
mmuras: You're right, It's removed.

adapters/adocean/adocean.go
```diff
@@ -0,0 +1,324 @@
...
+
+func (a *AdOceanAdapter) MakeBids(
+	internalRequest *openrtb.BidRequest,
+	externalRequest *adapters.RequestData,
+	response *adapters.ResponseData,
+) (*adapters.BidderResponse, []error) {
+	if response.StatusCode != http.StatusOK {
+		return nil, []error{fmt.Errorf("Unexpected status code: %d. Network error?", response.StatusCode)}
+	}
```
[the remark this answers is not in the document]
mmuras: No, unless there is something wrong expected response is 200. If there is no bid we return json with error field set to true.

adapters/adocean/adocean.go
mmuras: I don't expect number of slaves to be high enough to make a visible difference, but your version is also easier to read so I've changed it to your version, thanks :).

mmuras:
@hhhjort one more time :)

PR #1275, opened 2020-04-27
Add Adapter Orbidder

arneschulz1984:


adapters/orbidder/orbidder.go
```diff
@@ -0,0 +1,139 @@
...
+type OrbidderAdapter struct {
+	endpoint string
+}
+
+// MakeRequests makes the HTTP requests which should be made to fetch bids from orbidder.
+func (rcv *OrbidderAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errs []error
+	var validImps []openrtb.Imp
+
+	// check if imps exists, if not do not return error and do send request to orbidder
```
SyntaxNode, line 22: Comment says "do not return error" when it does return an error.

adapters/orbidder/orbidder.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
+		return &errortypes.BadInput{
+			Message: err.Error(),
+		}
+	}
+
+	var orbidderExt openrtb_ext.ExtImpOrbidder
+	if err := json.Unmarshal(bidderExt.Bidder, &orbidderExt); err != nil {
+		return &errortypes.BadInput{
+			Message: "Wrong orbidder bidder ext",
```
SyntaxNode, line 74: Consider including the `err.Error()` to help the publisher understand what they got wrong.

config/config.go
```diff
@@ -525,6 +525,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderNanoInteractive, "https://ad.audiencemanager.de/hbs/cookie_sync?gdpr={{.GDPR}}&consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redirectUri="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dnanointeractive%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderNinthDecimal, "https://rtb.ninthdecimal.com/xp/user-sync?acctid={aid}&&redirect="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dninthdecimal%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderOpenx, "https://rtb.openx.net/sync/prebid?r="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dopenx%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24%7BUID%7D")
+	// openrtb_ext.BidderOrbidder doesn't have a good default.
```
SyntaxNode, line 528: Please remove this comment. This should only be here if Orbidder supports user syncing but doesn't have a common default url. The Ordbidder adapter does not support user syncing.

arneschulz1984: Comment is removed

adapters/orbidder/orbidder.go
arneschulz1984: Thanks for the hint, i added the original error message

adapters/orbidder/orbidder.go
arneschulz1984: fixed to: check if imps exists, if not return error and do send request to orbidder

adapters/orbidder/orbidder.go
```diff
@@ -0,0 +1,139 @@
...
+		return &errortypes.BadInput{
+			Message: "Wrong orbidder bidder ext, no account id",
+		}
+	}
+
+	if orbidderExt.PlacementId == "" {
+		return &errortypes.BadInput{
+			Message: "Wrong orbidder bidder ext, no placement id",
+		}
+	}
```
guscarreon, line 88: Given that these 2 parameters were marked as `required` in file static/bidder-params/orbidder.json and that prebid-server runs a parameter validation before getting to this part of the code, it is safe to assume the true branches of the `if` statements in lines 78 and 84 will be unreachable in production. We could discard those statements entirely.

arneschulz1984: Thanks, i will drop the if statements and the tests for this cases.

arneschulz1984: if conditions and not more needed tests are removed

PR #1279, opened 2020-04-29
Yeahmobi adapter

zhaojp:
Add a adapter for Yeahmobi. Please review the code and let me know if there is something wrong.


adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+type YeahmobiAdapter struct {
+	EndpointTemplate template.Template
+}
+
+func NewYeahmobiBidder(endpointTemplate string) adapters.Bidder {
+	tpl, err := template.New("endpointTemplate").Parse(endpointTemplate)
+	if err != nil {
+		glog.Fatal("Unknow url template")
```
SyntaxNode, line 24: Nitpick: Typo "Unknow"
.. also consider including the error message from `err` to provide useful debugging info on the nature of the error.

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil
+	}
+	return &YeahmobiAdapter{EndpointTemplate: *tpl}
+}
+
+func (adapter *YeahmobiAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var adapterRequests []*adapters.RequestData
+
+	adapterRequest, errors := adapter.makeRequest(request)
+	if adapterRequest != nil {
```
SyntaxNode, line 34: It's more common in Go to check for `errors != nil` to determine if the return variables are safe to use.

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,179 @@
...
+	}, errs
+}
+
+func transform(request *openrtb.BidRequest) {
+	for i, imp := range request.Imp {
+		if imp.Native != nil {
+			var nativeRequest map[string]interface{}
+			err := json.Unmarshal([]byte(request.Imp[i].Native.Request), &nativeRequest)
+			if err == nil {
+				if nativeRequest["native"] != nil {
```
SyntaxNode, line 81: You would typically use the second return parameter. It's a boolean which indicates if the key was found in the map.

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,179 @@
...
+					continue
+				}
+				request.Imp[i].Native.Request = "{\"native\":" + request.Imp[i].Native.Request + "}"
```
SyntaxNode, line 84: The `Native` object is shared memory across the other adapters. Please do not modify it directly. Make a shallow copy first, something like:

```
nativeCopy := *imp.Native
nativeCopy.Request = "foo"
request.Imp[i].Native = &nativeCopy
```

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,179 @@
...
+	return &extImpYeahmobi, errs
+
+}
+
+func (adapter *YeahmobiAdapter) getEndpoint(ext *openrtb_ext.ExtImpYeahmobi) (string, error) {
+	if ext.ZoneId == "" {
+		return "", errors.New("param of zoneId not config")
+	}
+
+	return macros.ResolveMacros(adapter.EndpointTemplate, macros.EndpointTemplateParams{Host: "gw-" + ext.ZoneId + "-bid.yeahtargeter.com"})
```
SyntaxNode, line 118: Since ZoneId is a user configurable string, please consider running through a url encode to protect this adapter from unexpected input.

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,179 @@
...
+			err := json.Unmarshal([]byte(request.Imp[i].Native.Request), &nativeRequest)
+			if err == nil {
+				if nativeRequest["native"] != nil {
+					continue
+				}
+				request.Imp[i].Native.Request = "{\"native\":" + request.Imp[i].Native.Request + "}"
+			}
+		}
+	}
+}
```
SyntaxNode, line 88: Looks like this logic is untested. Please add a native json test.

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,182 @@
...
+		err := json.Unmarshal(imp.Ext, &extBidder)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		err = json.Unmarshal(extBidder.Bidder, &extImpYeahmobi)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
```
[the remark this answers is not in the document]
zhaojp: We only need the first valid ext element and I fixed it in new commit.Thanks reviewed the code.

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,175 @@
...
+			err := json.Unmarshal([]byte(request.Imp[i].Native.Request), &nativeRequest)
+			if err == nil {
+				_, exists := nativeRequest["native"]
+				if exists {
+					continue
+				}
+				nativeCopy := *request.Imp[i].Native
+				nativeCopy.Request = "{\"native\":" + request.Imp[i].Native.Request + "}"
+				request.Imp[i].Native = &nativeCopy
+			}
```
guscarreon, line 85: One last double-check: do we want to not do anything with an `err!=nil`? Would any `else` actions make sense in the context of this adapter? Or do we just want to skip the "bad" `imp.Native != nil` entries?

zhaojp: Yes. we just want to skip the "bad" imp.Native != nil entries.And thanks again for the review.

adapters/yeahmobi/yeahmobi.go
```diff
@@ -0,0 +1,175 @@
...
+		if imp.Native != nil {
+			var nativeRequest map[string]interface{}
+			err := json.Unmarshal([]byte(request.Imp[i].Native.Request), &nativeRequest)
+			if err == nil {
+				_, exists := nativeRequest["native"]
+				if exists {
+					continue
+				}
+				nativeCopy := *request.Imp[i].Native
+				nativeCopy.Request = "{\"native\":" + request.Imp[i].Native.Request + "}"
```
SyntaxNode, line 83: Would it be a good idea to run this through the json.Marshal? .. or can you guarantee that wrapping `request.Imp[i].Native.Request` in quotes always produces valid json?

zhaojp: Yes.It's a good idea and I optimized it in the new commit.

PR #1281, opened 2020-04-30
OpenX adapter: Enable video for app

laurb9:
* Enable app video for OpenX adapter.


SyntaxNode:
Looks like the adapter already supports video requests and even has json test coverage. Was the exclusion from the yaml file an oversight?

jimee02:
@SyntaxNode, yes I believe this is an oversight.  We've supported video for app prior to this.

PR #1290, opened 2020-05-07
OpenX adapter: Pass rewarded video flag

laurb9:


adapters/openx/openx.go
```diff
@@ -142,6 +142,14 @@ func preprocess(imp *openrtb.Imp, reqExt *openxReqExt) error {
 		}
 	}
 
+	if imp.Video != nil {
+		if bidderExt.Prebid != nil && bidderExt.Prebid.IsRewardedInventory == 1 {
+			imp.Video.Ext = json.RawMessage(`{"rewarded":1}`)
+		} else {
+			imp.Video.Ext = nil
```
SyntaxNode, line 149: I'm curious why the setting to nil. Do you expect the caller to potentially fill this is out in an incompatible value?

SyntaxNode:
Before we merge this in, could you please respond to the comment?

adapters/openx/openx.go
laurb9: Thank you for reviewing. Wanted to be consistent and wipe video.ext in both rewarded/non-rewarded cases, as the same happens to imp.ext when customParams is missing.

The java version passes values only explicitly anyway, this change there is:
https://github.com/rubicon-project/prebid-server-java/pull/712

PR #1292, opened 2020-05-08
Introduce Adhese adapter

sander-adhese:


adapters/adhese/adhese.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	complete_url := fmt.Sprintf("%s%s%s%s%s",
+		host,
+		extractSlotParameter(params),
+		extractTargetParameters(params),
+		extractGdprParameter(request),
+		extractRefererParameter(request))
+
+	// If all the requests are invalid, Call to adaptor is skipped
+	if len(request.Imp) == 0 {
```
guscarreon, line 122: It is better place this check before line 86 where we dereference the first element of `request.Imp`
```
86         	var imp = &request.Imp[0]
```

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func getAdMarkup(adheseBid openrtb_ext.AdheseBid, adheseExt openrtb_ext.AdheseExt) string {
+	if adheseExt.Ext == "js" && ContainsAny(adheseBid.Body, []string{"<script", "<div", "<html"}) {
+		return adheseBid.Body + "<img src='" + adheseExt.ImpressionCounter + "' style='height:1px; width:1px; margin: -1px -1px; display:none;'/>"
+	} else if adheseExt.Ext == "js" && ContainsAny(adheseBid.Body, []string{"<?xml", "<vast"}) {
+		return adheseBid.Body
+	} else {
+		return adheseExt.Tag
+	}
```
guscarreon, line 223: To avoid repeating the conditional `adheseExt.Ext == "js"` maybe we could refactor for a cleaner read:
```
func getAdMarkup(adheseBid openrtb_ext.AdheseBid, adheseExt openrtb_ext.AdheseExt) string {
    if adheseExt.Ext == "js" {
        if ContainsAny(adheseBid.Body, []string{"<script", "<div", "<html"}) {
            adheseBid.Body + "<img src='" + adheseExt.ImpressionCounter + "' style='height:1px; width:1px; margin: -1px -1px; display:none;'/>"
        } 
        if ContainsAny(adheseBid.Body, []string{"<?xml", "<vast"}) {
            adheseBid.Body
        }
    } 
    return adheseExt.Tag
}
```

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,259 @@
...
+func NewAdheseAdapter(config *adapters.HTTPAdapterConfig, uri string) *AdheseAdapter {
+	return NewAdheseBidder(adapters.NewHTTPAdapter(config).Client, uri)
+}
+
+func NewAdheseBidder(client *http.Client, uri string) *AdheseAdapter {
+	template, _ := template.New("endpointTemplate").Parse(uri)
+	return &AdheseAdapter{http: &adapters.HTTPAdapter{Client: client}, endpointTemplate: *template}
+}
+
+func printAsJson(obj interface{}) {
```
guscarreon, line 256: Function `printAsJson` seems to not be used anywhere and can probably be safely removed.

openrtb_ext/imp_adhese.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+type ExtImpAdhese struct {
+	Account  string          `json:"account"`
+	Location string          `json:"location"`
+	Format   string          `json:"format"`
+	Keywords json.RawMessage `json:"targets,omitempty"`
+}
+
+type AdheseOriginData struct {
```
guscarreon, line 16: Lines 16 on better fit inside `adapters/adhese/adhese.go` itself or even on a `adapters/adhese/utils.go` file in order to restrict the use of these types to the Adhese adapter only.

static/bidder-params/adhese.json
```diff
@@ -0,0 +1,42 @@
...
+				  "items": {
+					"type": "string"
+				  }
+				}
+			  },
+			  "required": ["key", "value"]
+			}
+		  }
+	},
+	"required": ["account", "location", "format"]
```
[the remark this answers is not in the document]
sander-adhese: Yes, we always require all three parameters.

static/bidder-params/adhese.json
```diff
@@ -0,0 +1,42 @@
...
+		},
+		"location": {
+			"type": "string",
+			"description": "The location you want to refer to for a specific section or page, as defined in your Adhese inventory"
+		},
+		"format": {
+			"type": "string",
+			"description": "The format you accept for this unit, as defined in your Adhese inventory"
+		},
+		"data": {
```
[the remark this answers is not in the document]
sander-adhese: Indeed, this is an oversight. Thank you for spotting it!

adapters/adapterstest/test_json.go
```diff
@@ -208,7 +208,7 @@ func diffErrorLists(t *testing.T, description string, actual []error, expected [
 	t.Helper()
 
 	if len(expected) != len(actual) {
-		t.Fatalf("%s had wrong error count. Expected %d, got %d", description, len(expected), len(actual))
+		t.Fatalf("%s had wrong error count. Expected %d, got %v", description, len(expected), actual)
```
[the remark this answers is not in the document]
sander-adhese: Sure - I did not know how to obtain the actual error message otherwise.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	m := make(map[string][]string)
+	var targetParsed map[string]interface{}
+	json.Unmarshal(parameters.Keywords, &targetParsed)
+	for targetKey, targetRawValue := range targetParsed {
+		var targetingValues = targetRawValue.([]interface{})
+		for _, targetRawValKey := range targetingValues {
+			var targetValueParsed = targetRawValKey.(string)
+			cur, _ := m[targetKey]
+			new := cur[:]
```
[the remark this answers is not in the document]
sander-adhese: A valid request and great suggestion, thanks!

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	var params openrtb_ext.ExtImpAdhese
+	if err := json.Unmarshal(bidderExt.Bidder, &params); err != nil {
+		errs = append(errs, WrapError("Request could not be parsed as ExtImpAdhese due to: "+err.Error()))
+		return nil, errs
+	}
+
+	// Validate request
+	if params.Account == "" || params.Location == "" || params.Format == "" {
```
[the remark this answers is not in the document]
sander-adhese: Yes, makes sense

adapters/adhese/adhese.go
sander-adhese: Correct

sander-adhese:
Thanks @guscarreon ! What is the procedure now for merging it?

guscarreon:
> Thanks @guscarreon ! What is the procedure now for merging it?

Once a second reviewer, one of my teammates, approve, the PR will get merged.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/macros"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type AdheseAdapter struct {
+	http             *adapters.HTTPAdapter
```
SyntaxNode, line 21: This is an outdated interface only used for the legacy endpoint. You probably saw this in an older adapter. Since it doesn't look like you're using it anyway, please remove it.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+	endpointTemplate template.Template
+}
+
+func (a *AdheseAdapter) Name() string {
```
SyntaxNode, line 25: This is legacy as well and is not needed. You only need to implement the methods of `adapters.Bidder`. I'll try to find time in the next couple of weeks to update our new Adapter docs to make this clear. Sorry for misleading you on this.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+	return "adhese"
+}
+
+func (a *AdheseAdapter) SkipNoCookies() bool {
```
SyntaxNode, line 29: This is legacy as well and is not needed. You only need to implement the methods of `adapters.Bidder`.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+	return false
+}
+
+func extractSlotParameter(parameters openrtb_ext.ExtImpAdhese) string {
+	return fmt.Sprintf("/sl%s-%s", parameters.Location, parameters.Format)
```
SyntaxNode, line 34: `parameters.Location` and `parameters.Format` are both strings. Please escape them with `url.PathEscape(x)`.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func extractTargetParameters(parameters openrtb_ext.ExtImpAdhese) string {
+	if parameters.Keywords == nil || len(parameters.Keywords) == 0 {
```
SyntaxNode, line 38: Syntax Nitpick: The nil check isn't necessary. The `len` method works nicely with nil slices and will return 0 in that case.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+		return ""
+	}
+	var parametersAsString = ""
+	var targetParsed map[string]interface{}
+	json.Unmarshal(parameters.Keywords, &targetParsed)
```
SyntaxNode, line 43: The call to `json.Unmarshal` can return an error. Please do not ignore it. Something like this is likely appropriate, ignoring the keywords if there's a problem:

```
err := json.Unmarshal(parameters.Keywords, &targetParsed)
if err != nil {
  return ""
}
```

```

File: adapters/adhese/adhese.go [hunk 16]
```diff
@@ -0,0 +1,247 @@
...
+	for targetKey, targetRawValue := range targetParsed {
+		var targetingValues = targetRawValue.([]interface{})
+		parametersAsString += "/" + targetKey
```
SyntaxNode, line 46: Since this is going to be inserted in an url, please `url.PathEscape(x)` the targetKey values.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+		for _, targetRawValKey := range targetingValues {
+			var targetValueParsed = targetRawValKey.(string)
+			parametersAsString += targetValueParsed + ";"
+		}
+		parametersAsString = strings.TrimRight(parametersAsString, ";")
+	}
+
+	params := strings.Split(parametersAsString, "/")
```
SyntaxNode, line 54: I don't think it makes sense to build a string only to split it immediately after. Consider using a slice in the loop and then sort and join at the end.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	var params openrtb_ext.ExtImpAdhese
+	if err := json.Unmarshal(bidderExt.Bidder, &params); err != nil {
+		errs = append(errs, WrapError("Request could not be parsed as ExtImpAdhese due to: "+err.Error()))
+		return nil, errs
+	}
+
+	// Compose url
+	endpointParams := macros.EndpointTemplateParams{Host: "ads-" + params.Account + ".adhese.com"}
```
SyntaxNode, line 103: I know this is an existing pattern you're modeling from, but it's one I'd wish we change.

Can you please include the "ads-" and ".adhese.com" parts in the template so it's more obvious we're only substituting the account? I don't know if it makes sense for you to use PublisherID, ZoneID, or SourceId... but feel free to add AccountID to that struct if you need to. The new default url would then be: `https://ads-{{.AccountID}}.adhese.com/json`

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+		extractRefererParameter(request))
+
+	return []*adapters.RequestData{{
+		Method: "GET",
+		Uri:    complete_url,
+	}}, errs
+}
+
+func (a *AdheseAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if response.StatusCode != http.StatusOK {
```
SyntaxNode, line 124: Some adapters return a 204 if no bids were made. Does your server behave that way? If so, you'll want to account for that status code here.

You may also want to consider if all of these errors are truly of type `BadInput` rather than some kind of issue on the server side, which is better represented as `BadServerResponse`.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+	var bidResponse openrtb.BidResponse
+
+	var adheseBidResponseArray []AdheseBid
+	if err := json.Unmarshal(response.Body, &adheseBidResponseArray); err != nil {
+		return nil, []error{err, WrapError(fmt.Sprintf("Response %v could not be parsed as generic Adhese bid.", string(response.Body)))}
+	}
+
+	var adheseBid = adheseBidResponseArray[0]
+
+	if adheseBid.Origin == "JERLICIA" {
```
SyntaxNode, line 137: Can you please share some context on why this one publisher is treated specially?

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+
+		if err := json.Unmarshal(response.Body, &originDataArray); err != nil {
+			return nil, []error{err, WrapError(fmt.Sprintf("Response %v could not be parsed to JERLICIA origin data.", string(response.Body)))}
+		}
+		bidResponse = convertAdheseBid(adheseBid, extArray[0], originDataArray[0])
+	} else {
+		bidResponse = convertAdheseOpenRtbBid(adheseBid)
+	}
+
+	price, _ := strconv.ParseFloat(adheseBid.Extension.Prebid.Cpm.Amount, 64)
```
SyntaxNode, line 152: Please do not ignore the errors from `ParseFloat` and `ParseUint`.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+				BidType: getBidType(bid.AdM),
+			})
+
+		}
+	}
+	return bidderResponse, errs
+}
+
+func convertAdheseBid(adheseBid AdheseBid, adheseExt AdheseExt, adheseOriginData AdheseOriginData) openrtb.BidResponse {
+	adheseExtJson, _ := json.Marshal(adheseOriginData)
```
SyntaxNode, line 183: Please do not ignore the error from `json.Marshal`.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+
+func getBidType(bidAdm string) openrtb_ext.BidType {
+	if bidAdm != "" && ContainsAny(bidAdm, []string{"<?xml", "<vast"}) {
+		return openrtb_ext.BidTypeVideo
+	}
+	return openrtb_ext.BidTypeBanner
+}
+
+func WrapError(errorStr string) *errortypes.BadInput {
+	return &errortypes.BadInput{Message: errorStr}
```
SyntaxNode, line 227: See earlier comment. I don't think it's appropriate to wrap everything as `BadInput`, which signals that the publisher is making the call wrong. Some instances look better served by `BadServerResponse` in which you get a 500 back or malformed json.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func ContainsAny(raw string, keys []string) bool {
+	for i := 0; i < len(keys); i++ {
+		if strings.Contains(strings.ToLower(raw), keys[i]) {
```
SyntaxNode, line 232: Consider pulling up the conversion of raw to lower case before the for loop to save a bit on performance.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,247 @@
...
+	return false
+
+}
+
+func NewAdheseAdapter(config *adapters.HTTPAdapterConfig, uri string) *AdheseAdapter {
+	return NewAdheseBidder(adapters.NewHTTPAdapter(config).Client, uri)
+}
+
+func NewAdheseBidder(client *http.Client, uri string) *AdheseAdapter {
+	template, _ := template.New("endpointTemplate").Parse(uri)
```
SyntaxNode, line 245: Please do not ignore errors.

adapters/adhese/params_test.go
```diff
@@ -0,0 +1,53 @@
...
+	`null`,
+	`nil`,
+	``,
+	`[]`,
+	`true`,
+	`{"location": "loc1", "format": "for1"}`,
+	`{"account": "demo", "format": "for1"}`,
+	`{"account": "demo", "location": "loc1"}`,
+	`{"account": "demo", "location": "loc1", "format": "for1", "targets": null`,
+}
```
SyntaxNode, line 53: Please add cases for invalid types. The Go json parser is sensitive to types and we want to be sure via tests that the code handles those errors properly. For example, have account be a number instead of a string:

`{"account": 42, "location": "loc1", "format": "for1"}`,

static/bidder-params/adhese.json
```diff
@@ -0,0 +1,25 @@
+{
+	"$schema": "http://json-schema.org/draft-04/schema#",
+	"title": "Adhese Adapter Parameters",
+	"description": "Validation for parameters handled by the Adhese adapter",
+	"type": "object",
+	"properties": {
+		"account": {
+			"type": "string",
+			"description": "Your Adhese account name. If unknown, please contact your sales rep"
```
SyntaxNode, line 9: Please consider adding a `/doc/bidders/adhese.md` with instructions on how to contact an Adhese salea rep to register for these values.

adapters/adhese/adhese.go
sander-adhese: Thanks for the tip!

adapters/adhese/adhese.go
sander-adhese: This is our own internal legacy system.

adapters/adhese/adhese.go
sander-adhese: They don't,- so I'll add AccountID

adapters/adhese/adhese.go
SyntaxNode: Gotcha. Thank you for clarifying. Just wanted to make sure we're not hardcoding specific publishers.

adapters/adhese/adhese.go
```diff
@@ -0,0 +1,265 @@
...
+	if len(response.SeatBid) > 0 && len(response.SeatBid[0].Bid) > 0 {
+		response.SeatBid[0].Bid[0].AdM = adheseBid.Body
+	}
+	return response
+}
+
+func getAdMarkup(adheseBid AdheseBid, adheseExt AdheseExt) string {
+	if adheseExt.Ext == "js" {
+		if ContainsAny(adheseBid.Body, []string{"<script", "<div", "<html"}) {
+			return adheseBid.Body + "<img src='" + adheseExt.ImpressionCounter + "' style='height:1px; width:1px; margin: -1px -1px; display:none;'/>"
```
SyntaxNode, line 223: How does the ImpressionCounter handle privacy regulation? Is it possible for the url to be empty?

macros/macros.go
```diff
@@ -11,6 +11,7 @@ type EndpointTemplateParams struct {
 	PublisherID string
 	ZoneID      string
 	SourceId    string
+	AccountID   string
```
SyntaxNode, line 14: Thank you for adding this. It makes sense for your use case.

However, it looks like this accidentally broke the `validateAdapterEndpoint` method in config.go. On line 266 there's a call to `macros.ResolveMacros` which includes hardcoded example data. Please add a dummy account id to the hardcoded `EndpointTemplateParams` data alongside `dummyHost` and `dummyPublisherID`.

adapters/adhese/adhese.go
sander-adhese: I'm not the expert at Adhese. We fully implement the GDPR protocols which handles sensitive user data/tracking. As far as I'm aware we can track a bid in order to properly bill it - that is the only use of this counter. If you require so I can move it up the chain here and get a better answer formulated. In any case, it could be empty so I'll add it to the condition. I'm guessing that's the main reason to ask?

SyntaxNode: Yes, exactly.

sander-adhese:
Thanks guys, also for the solid feedback!

Issue #1272, opened 2020-04-23
Privacy update: removing user.ext.eids


bretg:
I posted a [PBS Privacy FAQ](https://prebid.org/faq/prebid-server-faq.html#how-does-prebid-server-support-privay-signals) entry on prebid.org and realized that we had missed some other places where there are user IDs: user.ext.eids and user.ext.digitrust

In short, wherever we remove user.buyeruid, we should also remove user.id, user.ext.eids, and user.ext.digitrust

Request changes:

1) Mobile ‘Limit Ad Tracking’ flag (and DNT HTTP header)
If PBS receives ‘device.lmt’ flag in the OpenRTB request, it should also remove user.id, user.ext.eids, and user.ext.digitrust

2) GDPR TCF1 and TCF2
Wherever user.buyeruid is removed, also remove user.id, user.ext.eids, and user.ext.digitrust

3) COPPA
If regs.coppa is set to ‘1’ on the OpenRTB request, also remove user.id, user.ext.eids, and user.ext.digitrust

4) CCPA
If regs.ext.us_privacy is parsed to find that the user has opted-out of a “sale”, also remove user.id, user.ext.eids, and user.ext.digitrust

SyntaxNode:
Sounds good to me. We'll work on this next week for PBS-Go.

SyntaxNode:
This is complete in PBS-Go and PBS-Java.

PR #1294, opened 2020-05-09
Privacy: Remove user.ext.eids

SyntaxNode:
Implements external ids portion of https://github.com/prebid/prebid-server/issues/1272


privacy/scrubber.go
```diff
@@ -99,6 +101,24 @@ func (scrubber) ScrubUser(user *openrtb.User, demographic ScrubStrategyDemograph
 	userCopy.BuyerUID = ""
 	userCopy.ID = ""
 
+	if len(userCopy.Ext) > 0 {
+		var extUser openrtb_ext.ExtUser
+		err := json.Unmarshal(userCopy.Ext, &extUser)
+		if err == nil {
+			needsToScrub := len(extUser.Eids) > 0 || extUser.DigiTrust != nil
```
SyntaxNode, line 108: Marshaling can be expensive. This is an optimization to avoid re-marshaling if there is nothing to remove.

hhhjort: I would suggest looking at jsonparser, as you can manipulate the json as raw json, eliminating the need for a full unmarshal/marshal loop. Would probably simplify your code as jsonparser can check if the element exists when you ask it to remove the element.

SyntaxNode: I ran the suggestion through benchmarking.

PR Code Using `encoding/json`:
```
BenchmarkNone-8            	 3737512	       281 ns/op	     320 B/op	       3 allocs/op
BenchmarkDigitrustOnly-8   	  687613	      1778 ns/op	     424 B/op	      10 allocs/op
BenchmarkEidOnly-8         	  336398	      3901 ns/op	    1152 B/op	      20 allocs/op
BenchmarkBoth-8            	  240844	      4661 ns/op	    1184 B/op	      22 allocs/op
BenchmarkBothPlusExtra-8   	  244676	      4988 ns/op	    1232 B/op	      23 allocs/op
```

Recommended Change using `jsonparser`:
```
BenchmarkNone-8            	17608536	        69.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkDigitrustOnly-8   	 6746899	       194 ns/op	       0 B/op	       0 allocs/op
BenchmarkEidOnly-8         	 3422544	       354 ns/op	       0 B/op	       0 allocs/op
BenchmarkBoth-8            	 2536236	       470 ns/op	       0 B/op	       0 allocs/op
BenchmarkBothPlusExtra-8   	 2240919	       548 ns/op	       0 B/op	       0 allocs/op
```

Clearly, the `jsonparser` implementation is much faster. However, it's not stable. The `Delete` method is listed as an experimental feature only and fails on this test string:
`{"consent":"anyConsent","eids":[{"source":"anySource","id":"anyId","uids":[{"id":"anyId","ext":{"id":42}}],"ext":{"id":42}}],"digitrust":{"id":"anyId","keyv":4,"pref":8}}`

SyntaxNode: I looked at other json libraries already in our go.mod file. `fastjson` does better than `encoding/json`:
```
BenchmarkNone-8            	 6708949	       174 ns/op	     104 B/op	       4 allocs/op
BenchmarkDigitrustOnly-8   	 1000000	      1211 ns/op	    1472 B/op	      11 allocs/op
BenchmarkEidOnly-8         	  565209	      2126 ns/op	    2976 B/op	      18 allocs/op
BenchmarkBoth-8            	  460468	      2629 ns/op	    3248 B/op	      22 allocs/op
BenchmarkBothPlusExtra-8   	  344578	      3518 ns/op	    6056 B/op	      25 allocs/op
```

.. but not with a very wide margin. I think searching for alternate json parsers is a bit beyond the scope of this PR. Since our usual fallback doesn't work well in this scenario, I think we should leave it alone for now. A future project can reassess using a new json parser.

hhhjort:
We no have some merge conflicts to resolve.

SyntaxNode:
> We now have some merge conflicts to resolve.

Fixed.

PR #1296, opened 2020-05-11
VIS.X: fix bid.ID, bid.CrID and set default currency value

mk0x9:


adapters/visx/visx.go
```diff
@@ -14,11 +14,38 @@ type VisxAdapter struct {
...
+
 // MakeRequests makes the HTTP requests which should be made to fetch bids.
 func (a *VisxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
 	var errors = make([]error, 0)
 
-	reqJSON, err := json.Marshal(request)
+	// copy the request, because we are going to mutate it
+	requestCopy := *request
+	if len(requestCopy.Cur) == 0 {
+		requestCopy.Cur = make([]string, 1)
```
SyntaxNode, line 44: This is fine. An alternative way of writing this is the one liner: `requestCopy.Cur = []string{"USD"}`

adapters/visx/visx.go
```diff
@@ -14,11 +14,38 @@ type VisxAdapter struct {
 	endpoint string
 }
 
+type visxBid struct {
```
SyntaxNode, line 17: I'm curious. What was broken with the `bid.ID` and `bid.CrID` previously?

adapters/visx/visx.go
```diff
@@ -62,8 +89,19 @@ func (a *VisxAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequ
...
+			bid := openrtb.Bid{}
+			bid.ID = internalRequest.ID
+			bid.CrID = fmt.Sprint(sb.Bid[i].UID)
+			bid.ImpID = sb.Bid[i].ImpID
+			bid.Price = sb.Bid[i].Price
+			bid.AdM = sb.Bid[i].AdM
+			bid.W = sb.Bid[i].W
+			bid.H = sb.Bid[i].H
+			bid.ADomain = sb.Bid[i].ADomain
+			bid.DealID = sb.Bid[i].DealID
```
SyntaxNode, line 101: Can you please add a test case to validate the DealID is mapped correctly? I think that's missing.

adapters/visx/visx.go
mk0x9: Problems with `bid.CrID` was resolved on the server side.

About `bid.ID` – our ad server doesn't return this field in the ad response. Since this parameter is required we set it in s2s adapter.

adapters/visx/visx.go
mk0x9: Done.

adapters/visx/visx.go
mk0x9: Done.

PR #1297, opened 2020-05-11
Bugfix for missing fields in imp.video

VeronikaSolovei9:
Fixes https://github.com/prebid/prebid-server/pull/1297

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -350,7 +350,7 @@ func (deps *endpointDeps) createImpressions(videoReq *openrtb_ext.BidRequestVide
 
 		impsArray := make([]openrtb.Imp, numImps)
 		for impInd := range impsArray {
-			newImp := createImpressionTemplate(storedImp, *videoData)
+			newImp := createImpressionTemplate(storedImp, videoData)
 			impsArray[impInd] = newImp
 			if reqExactDur {
 				//floor := int(math.Floor(ind/impDivNumber))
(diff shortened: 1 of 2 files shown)


endpoints/openrtb2/video_auction.go
```diff
@@ -380,12 +380,8 @@ func max(a, b int) int {
 	return b
 }
 
-func createImpressionTemplate(imp openrtb.Imp, video *openrtb.Video) openrtb.Imp {
-	imp.Video = &openrtb.Video{}
-	imp.Video.W = video.W
-	imp.Video.H = video.H
-	imp.Video.Protocols = video.Protocols
-	imp.Video.MIMEs = video.MIMEs
+func createImpressionTemplate(imp openrtb.Imp, video openrtb.Video) openrtb.Imp {
```
neririchter, line 383: Question: this implementation will ignore all video params inside the stored imp, is this the desired behavior?

endpoints/openrtb2/video_auction.go
```diff
@@ -380,12 +380,8 @@ func max(a, b int) int {
...
+	imp.Video = &video
```
SyntaxNode, line 384: Do you think it would be cleaner to leave the method signature the same as it was before and make the video copy within the method? That might be easier to read / understand.

VeronikaSolovei9: Refactored.

endpoints/openrtb2/video_auction.go
```diff
@@ -381,11 +381,10 @@ func max(a, b int) int {
 }
 
 func createImpressionTemplate(imp openrtb.Imp, video *openrtb.Video) openrtb.Imp {
-	imp.Video = &openrtb.Video{}
-	imp.Video.W = video.W
-	imp.Video.H = video.H
-	imp.Video.Protocols = video.Protocols
-	imp.Video.MIMEs = video.MIMEs
+	//for every new impression we need to have it's own copy of video object, because we customize it in further processing
+	newVideo := &openrtb.Video{}
```
SyntaxNode, line 385: The `&openrtb.Video{}` is thrown away immediately on the following line. Would this work for you?

```
newVideo := *video
imp.Video = &newVideo
```

VeronikaSolovei9: Yes, thank you! Fixed!

PR #1302, opened 2020-05-14
Enable full TCF2 support

hhhjort:


gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
+	}
+	if p.cfg.TCF2.Purpose2 {
+		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.BasicAdserving)
+	}
+	if p.cfg.TCF2.Purpose4 {
+		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.PersonalizationSelection)
+	}
+	if p.cfg.TCF2.Purpose7 {
+		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.AdPerformance)
+	}
```
guscarreon, line 151: If `allowPI` is `true`, after the true branches of any of the first 3 `if` statements, it still can end up being `false` in line 150. Did we have this in mind? If all `p.cfg.TCF2.Purpose1`, `p.cfg.TCF2.Purpose2`, `p.cfg.TCF2.Purpose4`, and `p.cfg.TCF2.Purpose7` are set to true, and `allowPI` remains `true` through after the first 3 `if` statement, it can still end up being `false` in line 150. Is that ok? Just double-checking.

gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
+	allowGeo = false
+	if !ok {
+		err = fmt.Errorf("Unable to access TCF2 parsed consent")
+		return
+	}
+	if p.cfg.TCF2.SpecialPurpose1 {
+		allowGeo = consent.SpecialFeatureOptIn(1) && vendor.SpecialPurpose(1)
+	} else {
+		allowGeo = true
+	}
```
guscarreon, line 137: Can we add tests, at least to cover the true branch of the `if` statement in line 129

gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
+	if consent.CheckPubRestriction(uint8(purpose), 0, vendorID) {
+		return false
+	}
+	if consent.CheckPubRestriction(uint8(purpose), 1, vendorID) {
+		return vendor.PurposeStrict(purpose) && consent.PurposeAllowed(purpose) && consent.VendorConsent(vendorID)
+	}
+	if consent.CheckPubRestriction(uint8(purpose), 2, vendorID) {
+		// Need LITransparency here
+		return vendor.LegitimateInterestStrict(purpose) && consent.PurposeLITransparency(purpose) && consent.VendorLegitInterest(vendorID)
```
guscarreon, line 167: Can we add coverage to some of these scenarios?

guscarreon: image

gdpr/impl.go
guscarreon: image

gdpr/impl.go
```diff
@@ -40,22 +42,23 @@ func (p *permissionsImpl) BidderSyncAllowed(ctx context.Context, bidder openrtb_
...
 
 	id, ok := p.vendorIDs[bidder]
 	if ok {
+
 		return p.allowPI(ctx, id, consent)
 	}
 
 	if consent == "" {
-		return p.cfg.UsersyncIfAmbiguous, nil
+		return p.cfg.UsersyncIfAmbiguous, p.cfg.UsersyncIfAmbiguous, nil
```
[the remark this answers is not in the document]
hhhjort: Yes, we want it twice. The config applies to both "do we want to scrub non-geo PI" and "do we want to scrub geo" questions.

gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
 	}
 
 	// InfoStorageAccess is the same across TCF 1 and TCF 2
+	if parsedConsent.Version() == 2 {
+		consent, ok := parsedConsent.(tcf2.ConsentMetadata)
+		if !ok {
+			err := fmt.Errorf("Unable to access TCF2 parsed consent")
+			return false, err
+		}
+		return p.checkPurpose(consent, vendor, vendorID, consentconstants.InfoStorageAccess), nil
```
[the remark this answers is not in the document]
hhhjort: It does work, that is the same recipe used in `allowPI()`. This causes `consent` to be a struct of type `tcf2.ConsentMetadata`, as long as the interface was pointing to is a struct of that type. If the interface contained anything else but a `tcf2.ConsentMetadata` type, then `ok` would be `false`.

Using this, we can access methods of `tcf2.ConsentMetadata` that are not included in the interface definition. In the go-gdpr library, the choice was either add the methods for the new fields to the interface, and figure out what the tcf1 code should do when they were requested, or just use this to access the new fields only when it is determined that we have a tcf2 consent string.

gdpr/impl.go
hhhjort: I am not sure how we would write the test. We would have to come up with a consent string that claims to be TCF2, but that the go-gdpr library resolves into a tcf1 consent object.

hhhjort: Actually it may be possible, calling this function directly. We would just need to get a TCF1 parsed consent string, and then call this function using that. This if statement is one of those "It should not be possible to get here, but we will catch this case just in case" safety statements.

gdpr/impl.go
hhhjort: Yes, that is correct. If `checkPurpose()` fails for any of the configured purposes, we want `allowPI` to be false. I could have put an if statement inside of each check to return immediately if `allowPI` got set to `false`, but I am not sure if that would have saved anything.

gdpr/impl.go
guscarreon: I see. If its too time consuming, maybe we don' need the test. Coverage-wise this function looks fine.

gdpr/impl.go
guscarreon: Can `checkPurpose` be false on say, line 141 but true on one of its latter calls? Do we want to return `true` for the `allowPI` value even if it was previously false?

gdpr/impl_test.go
```diff
@@ -204,21 +205,90 @@ func TestAllowPersonalInfo(t *testing.T) {
...
 
 	// Assert that an item that otherwise would not be allowed PI access, gets approved because it is found in the GDPR.NonStandardPublishers array
 	perms.cfg.NonStandardPublisherMap = map[string]int{"appNexusAppID": 1}
-	allowPI, err = perms.PersonalInfoAllowed(context.Background(), openrtb_ext.BidderAppnexus, "appNexusAppID", "BOS2bx5OS2bx5ABABBAAABoAAAABBwAA")
+	allowPI, _, err = perms.PersonalInfoAllowed(context.Background(), openrtb_ext.BidderAppnexus, "appNexusAppID", "BOS2bx5OS2bx5ABABBAAABoAAAABBwAA")
 	assertNilErr(t, err)
 	assertBoolsEqual(t, true, allowPI)
 }
 
+func TestAllowPersonalInfoTCF2(t *testing.T) {
```
SyntaxNode, line 223: Can you please refactor this test to use the table driven tests format? .. with descriptions added to the assert statements to narrow down the test failure please.

privacy/enforcement_test.go
```diff
@@ -97,9 +100,10 @@ func TestApply(t *testing.T) {
...
-				CCPA:  false,
-				COPPA: false,
-				GDPR:  true,
+				CCPA:    false,
+				COPPA:   false,
+				GDPR:    true,
+				GDPRGeo: true,
 			},
 			expectedDeviceIPv6:      ScrubStrategyIPV6Lowest16,
 			expectedDeviceGeo:       ScrubStrategyGeoReducedPrecision,
```
SyntaxNode, line 109: Please add a GDPR Geo only test.

privacy/enforcement.go
```diff
@@ -6,14 +6,15 @@ import (
...
+	CCPA    bool
+	COPPA   bool
+	GDPR    bool
+	GDPRGeo bool
 }
 
 // Any returns true if at least one privacy policy requires enforcement.
 func (e Enforcement) Any() bool {
-	return e.CCPA || e.COPPA || e.GDPR
+	return e.CCPA || e.COPPA || e.GDPR || e.GDPRGeo
```
SyntaxNode, line 17: Please add a test case for this.

gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
+		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.PersonalizationSelection)
+	}
+	if p.cfg.TCF2.Purpose7 {
+		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.AdPerformance)
+	}
+	return
+}
+
+func (p *permissionsImpl) checkPurpose(consent tcf2.ConsentMetadata, vendor api.Vendor, vendorID uint16, purpose tcf1constants.Purpose) bool {
+	if purpose == 1 && p.cfg.TCF2.PurposeOneTreatment.Enabled && consent.PurposeOneTreatment() {
```
SyntaxNode, line 156: Can you please explain what the trailing `consent.PurposeOneTreatment()` adds to the logic?

gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
+		return p.cfg.TCF2.PurposeOneTreatment.AccessAllowed
+	}
+	if consent.CheckPubRestriction(uint8(purpose), 0, vendorID) {
```
SyntaxNode, line 159: It's really hard to remember what the restriction type is. Consider adding constants.

0 = Not Allowed
1 = Require Consent
2 = Require Legitimate Interest

exchange/utils.go
```diff
@@ -59,10 +59,12 @@ func cleanOpenRTBRequests(ctx context.Context,
 			coreBidder := resolveBidder(bidder.String(), aliases)
 
 			var publisherID = labels.PubID
-			ok, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, publisherID, consent)
+			ok, geo, err := gDPR.PersonalInfoAllowed(ctx, coreBidder, publisherID, consent)
 			privacyEnforcement.GDPR = !ok && err == nil
+			privacyEnforcement.GDPRGeo = !geo && err == nil
```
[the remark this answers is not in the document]
hhhjort: This is the case when we don't know if GDPR applies, and have no consent string, So the only way they could be different is if we want to default to scrubbing one or the other when we don't know, but not both.

gdpr/impl.go
hhhjort: Possibly, it is an extra check as I have to check if the value is false before I move on. It would be one extra check inside each of those 4 purpose checks, so 4 extra ifs that are always executed vs needing to execute 2, 4, or 6 extra if statements if one of the early purposes failed their check.

gdpr/impl_test.go
hhhjort: Those would be some huge tables I fear. I compacted down the TCF2 test environment setup by defining a "standard base" environment, and then tweaking the needed values for each test. That standard+tweak doesn't work well in a table setup, and I would instead be forced to replicate most if not all of the standard block for each entry in the table. Or do you have an idea on how to set up the tables so that they don't become a huge unreadable mess?

gdpr/impl.go
hhhjort: First comparison is to ensure we are considering purpose 1. The second comparison is to see if the PBS Host as configured to care about the PurposeOneTreatment flag. The last, `consent.PurposeOneTreatment()` checks to see if the flag is set in the consent string. It is last because it is the most expensive check.

exchange/utils.go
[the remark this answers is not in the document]
hhhjort: It should be able to return different values for TCF2 consent strings. All the TCF1 logic will lock these two to be the same value.

config/config.go
```diff
@@ -164,6 +165,22 @@ func (t *GDPRTimeouts) ActiveTimeout() time.Duration {
 	return time.Duration(t.ActiveVendorlistFetch) * time.Millisecond
 }
 
+// TCF2 defines the TCF2 specific configurations for GDPR
+type TCF2 struct {
+	Enabled             bool                 `mapstructure:"enabled"`
+	Purpose1            bool                 `mapstructure:"purpose1"`
```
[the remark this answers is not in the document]
hhhjort: Ok

gdpr/impl.go
```diff
@@ -40,22 +42,23 @@ func (p *permissionsImpl) BidderSyncAllowed(ctx context.Context, bidder openrtb_
...
+func (p *permissionsImpl) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, consent string) (bool, bool, error) {
 	_, ok := p.cfg.NonStandardPublisherMap[PublisherID]
 	if ok {
-		return true, nil
+		return true, true, nil
 	}
 
 	id, ok := p.vendorIDs[bidder]
 	if ok {
+
```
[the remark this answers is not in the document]
hhhjort: Nope, thanks

gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
+			return true, true, nil
 		}
 	}
-	return false, nil
+	return false, false, nil
+}
+
+func (p *permissionsImpl) allowPITCF2(parsedConsent api.VendorConsents, vendor api.Vendor, vendorID uint16) (allowPI bool, allowGeo bool, err error) {
+	consent, ok := parsedConsent.(tcf2.ConsentMetadata)
+	err = nil
```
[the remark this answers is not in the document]
hhhjort: Not strictly necessary, but it results in the minimum number of assignment statements. Otherwise I would be putting more assignment statements inside of if blocks before the `return` fires.

Or if you mean that I am assigning these values to the null defaults go would give them anyway, I guess not. I think it improves readability to explicitly say what their defaults are, and depending on how good the compiler optimizer is it should not cost anything.

gdpr/impl.go
```diff
@@ -74,38 +77,99 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
+	allowPI = false
+	allowGeo = false
+	if !ok {
+		err = fmt.Errorf("Unable to access TCF2 parsed consent")
+		return
+	}
+	if p.cfg.TCF2.SpecialPurpose1 {
```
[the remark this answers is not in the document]
hhhjort: Does that save anything, since I would first have to assign `allowGeo` to true, and then evaluate all those things to reassign `allowGeo` to the resulting value, which may be true or false?

gdpr/impl.go
hhhjort: I don't have a consent string generator that will do publisher restrictions. The two consent strings generated by hand for go-gdpr were poorly crafted as far as supporting anything but the check publisher restriction function, so very limited coverage here now.

config/config.go
```diff
@@ -165,6 +166,26 @@ func (t *GDPRTimeouts) ActiveTimeout() time.Duration {
 	return time.Duration(t.ActiveVendorlistFetch) * time.Millisecond
 }
 
+// TCF2 defines the TCF2 specific configurations for GDPR
+type TCF2 struct {
+	Enabled             bool                 `mapstructure:"enabled"`
+	Purpose1            PurposeDetail        `mapstructure:"purpose1"`
+	Purpose2            PurposeDetail        `mapstructure:"purpose2"`
+	Purpose7            PurposeDetail        `mapstructure:"purpose7"`
+	SpecialPurpose1     PurposeDetail        `mapstructure:"special_purpose1"`
```
SyntaxNode, line 175: When we add config options for other purposes in the future for the 5/15 spec update, will we add them as Purpose3, etc... or switch to a slice?

config/config.go
```diff
@@ -769,6 +790,14 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.timeouts_ms.init_vendorlist_fetches", 0)
 	v.SetDefault("gdpr.timeouts_ms.active_vendorlist_fetch", 0)
 	v.SetDefault("gdpr.non_standard_publishers", []string{""})
+	v.SetDefault("gdpr.tcf2.enabled", true)
```
SyntaxNode, line 793: Does it make sense to enable everything by default? Honest question. Not sure how we did it for TCF1.

gdpr/impl.go
```diff
@@ -2,11 +2,13 @@ package gdpr
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/prebid/go-gdpr/api"
 	tcf1constants "github.com/prebid/go-gdpr/consentconstants"
 	consentconstants "github.com/prebid/go-gdpr/consentconstants/tcf2"
 	"github.com/prebid/go-gdpr/vendorconsent"
+	tcf2 "github.com/prebid/go-gdpr/vendorconsent/tcf2"
```
SyntaxNode, line 11: Does this need an alias defined? Wouldn't it just be tcf2 without it?

gdpr/impl.go
```diff
@@ -78,38 +80,104 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
-			return true, nil
+			return true, true, nil
 		}
 	}
-	return false, nil
+	return false, false, nil
+}
+
+func (p *permissionsImpl) allowPITCF2(parsedConsent api.VendorConsents, vendor api.Vendor, vendorID uint16) (allowPI bool, allowGeo bool, err error) {
+	consent, ok := parsedConsent.(tcf2.ConsentMetadata)
```
SyntaxNode, line 132: This check is also performed at the call site. Can we pass the casted tcf2.ConsentMetadata object?

privacy/enforcement.go
hhhjort: added

config/config.go
hhhjort: Is that the cookie sync update where we check for purposes 2-10? I am not sure what the use case would be there for per purpose checks. Currently, if we don't enforce purpose 1, then `allowSync` always passes. When we add in the 2-10 check, what would it mean to not enforce purpose 3 for example? In the `allowSync`, we are not enforcing any purpose other than 1, we just are checking if any one of the other purposes provide a valid reason to want to sync.

gdpr/impl.go
hhhjort: No, it would be `vendorconsent` if undefined, and conflict with the line above.

gdpr/impl.go
hhhjort: Where is that? I am only finding it here and in `allowSync`

config/config.go
hhhjort: I don't think the defaults have changed since TCF1 went in. We don't have an option to turn off TCF1 completely, just what to assume if there is no GDPR signal at all. It makes sense to have everything on, as TCF1 was hardcoded with everything on.

config/config.go
SyntaxNode: I had to go back and check. It's new changes to the 'Passing User IDs into an auction' which now states:

Any Purpose 2-10. User IDs are important for more than personalizing ads - they can be used in frequency capping, building profiles, counting unique users, etc. So Prebid Server should pass User IDs through the auction if any of Purposes 2-10 pass the legal basis test.

Would that require expanding the purpose configs? Perhaps not?

gdpr/impl.go
SyntaxNode: Oh, ok.

gdpr/impl.go
SyntaxNode: Yeah. I'm not sure what I saw earlier. I can't find it now.

config/config.go
hhhjort: I think for purpose 2-10, we would just have a list of purposes that the host does not consider a valid reason for PI. That is if the host declares purpose 3 as not valid (on the list), then a vendor who only claims purpose 3 would not get the PI. `Purpose1` and `SpecialPurpose1` will likely stay, 2 and 7 would go away and be replaced by the list. The key change in behavior is we are now looking for "any of" rather than "all" purposes.

Another open question I discussed with Brett after reading the most recent changes in the document is the different types of PI. Geo is already captured with `SpecialPurpose1`, though not explicitly called out that much. User IDs are discussed in the document, but we also have IP addresses, age, and gender data to potentially be scrubbed. Brett said he would bring that up with the group for clarification.

SyntaxNode: Cool. I saw Bret started that discussion. Should the pending results of that discussion hold up this PR?

hhhjort: I don't think so ... the only conflict I anticipate is the removal of the `Purpose2` and `Purpose7` sections ... the new stuff should end up in a new section I think

PR #1304, opened 2020-05-17
privacy: Potential JSON injection

djcsdy:
`privacy/ccpa/policy.go` and `privacy/gdpr/policy.go` include code that attempts to generate JSON by concatenating strings. If the `us_privacy` or `consent` fields contain the character `"`, this can produce malformed JSON, or JSON with an unintended structure.

This PR fixes the problem by encoding all JSON using the `json` module.

The included tests demonstrate that an attacker could exploit this bug to trick Prebid Server into mutating an OpenRTB bid request to contain arbitrary data, bypassing validation. These tests fail against the existing code.

### DIFF
--- a/privacy/ccpa/policy.go
+++ b/privacy/ccpa/policy.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/buger/jsonparser"
 	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
@@ -48,9 +47,14 @@ func (p Policy) Write(req *openrtb.BidRequest) error {
 		return err
 	}
 
-	jsonString, err := json.Marshal(p.Value)
+	var extMap map[string]interface{}
+	err := json.Unmarshal(req.Regs.Ext, &extMap)
 	if err == nil {
-		req.Regs.Ext, err = jsonparser.Set(req.Regs.Ext, jsonString, "us_privacy")
+		extMap["us_privacy"] = p.Value
+		ext, err := json.Marshal(extMap)
+		if err == nil {
+			req.Regs.Ext = ext
+		}
 	}
 	return err
 }
--- a/privacy/ccpa/policy.go
+++ b/privacy/ccpa/policy.go
@@ -40,10 +40,11 @@ func (p Policy) Write(req *openrtb.BidRequest) error {
 		req.Regs = &openrtb.Regs{}
 	}
 
-	var err error
-
 	if req.Regs.Ext == nil {
-		req.Regs.Ext, err = json.Marshal(openrtb_ext.ExtRegs{USPrivacy: p.Value})
+		ext, err := json.Marshal(openrtb_ext.ExtRegs{USPrivacy: p.Value})
+		if err == nil {
+			req.Regs.Ext = ext
+		}
 		return err
 	}
 
--- a/privacy/ccpa/policy.go
+++ b/privacy/ccpa/policy.go
@@ -40,13 +40,17 @@ func (p Policy) Write(req *openrtb.BidRequest) error {
 		req.Regs = &openrtb.Regs{}
 	}
 
+	var err error
+
 	if req.Regs.Ext == nil {
-		req.Regs.Ext = json.RawMessage(`{"us_privacy":"` + p.Value + `"}`)
-		return nil
+		req.Regs.Ext, err = json.Marshal(openrtb_ext.ExtRegs{USPrivacy: p.Value})
+		return err
 	}
 
-	var err error
-	req.Regs.Ext, err = jsonparser.Set(req.Regs.Ext, []byte(`"`+p.Value+`"`), "us_privacy")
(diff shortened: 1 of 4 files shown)


SyntaxNode:
Thank you reviewing the project for potential security risks. I greatly appreciate the scrutiny to ensure Prebid Server doesn't provide attack vectors which put our customers at risk.

In isolation, I agree this method poses a security risk. I don't see it as a possible attack vector though due to two reasons:
1) This method is only called for AMP requests, in which the consent string is validated as either a CCPA or GDPR TCF1/TCF2 consent string. 
2) Go's `json.Marshal` method validates embedded json to avoid injection attacks such as this. See: https://play.golang.org/p/gumbqihqXJ6

The benefit from this change is that we are protected from potentially harmful future uses of this method or if the Go json marshaller changes in the future and no longer properly validates embedded json.

The cost is performance. Use of the `json.Marshal` method takes 3x longer.
```
BenchmarkWrite-8      	 3492728	       359 ns/op	      32 B/op	       1 allocs/op
BenchmarkWriteNew-8   	 1000000	      1021 ns/op	      64 B/op	       2 allocs/op
```

Thoughts?

djcsdy:
@SyntaxNode Agreed that it is not possible to inject JSON outside the `ext` subtrees, which is where the most potential for mischief lies.

It *is* possible to inject arbitrary JSON inside the `user.ext` or `regs.ext` subtrees. I can’t think of a way to exploit this with OpenRTB extensions that I’m aware of, but it could easily become exploitable without notice.

IMHO it is almost never a good idea to construct structured data by concatenating strings (of characters or bytes) outside of a specialised serializer or marshaller. IME even many programmers with decades of experience do not understand how badly this approach can go wrong.

Programmers use the code that already exists in a project as an exemplar of how to write new code, so if you use a particular approach to solve a problem, the project as a whole is likely to use a similar approach to solve similar problems in the future. So although this particular code might be safe (which is already a wobbly assumption), it will encourage similar unsafe code to creep in in future. I did a quick search of the project when I proposed this change and found no other instances of code that constructs JSON by string concatenation, so this is a chance to nip a potential future problem in the bud.

FWIW I am still not particularly happy with this code as an exemplar of JSON-handling code even after my changes. It still uses string concatenation, it’s just more careful to ensure every component is properly encoded before concatenation. There is probably a better way to write this in terms of JSON-handling correctness and clarity. I went with the minimal change that solved the problem.

Everything stated here is just my opinion. It’s up to you to decide if the performance difference is worth the risk.

SyntaxNode:
@djcsdy I agree it's a good idea to set positive examples in the code base. I was benchmarking json encoders last week, but that's going to be a larger effort. We can go with the marshalling approach you have here.

privacy/ccpa/policy.go
```diff
@@ -40,13 +40,17 @@ func (p Policy) Write(req *openrtb.BidRequest) error {
...
 	if req.Regs.Ext == nil {
-		req.Regs.Ext = json.RawMessage(`{"us_privacy":"` + p.Value + `"}`)
-		return nil
+		req.Regs.Ext, err = json.Marshal(openrtb_ext.ExtRegs{USPrivacy: p.Value})
+		return err
 	}
 
-	var err error
-	req.Regs.Ext, err = jsonparser.Set(req.Regs.Ext, []byte(`"`+p.Value+`"`), "us_privacy")
+	jsonString, err := json.Marshal(p.Value)
```
SyntaxNode, line 50: I'm wondering if it's just better / safer to do a full json.Unmarshal / set the privacy string / json.Marshal. If you feel that's a bit much, I can address it in the future. It would require changing the unit tests from the arbitrary existing json to valid fields of the Reg.Ext.

privacy/ccpa/policy.go
```diff
@@ -40,13 +40,17 @@ func (p Policy) Write(req *openrtb.BidRequest) error {
 		req.Regs = &openrtb.Regs{}
 	}
 
+	var err error
+
 	if req.Regs.Ext == nil {
-		req.Regs.Ext = json.RawMessage(`{"us_privacy":"` + p.Value + `"}`)
-		return nil
+		req.Regs.Ext, err = json.Marshal(openrtb_ext.ExtRegs{USPrivacy: p.Value})
```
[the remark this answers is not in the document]
djcsdy: Good call, I will make this change.

privacy/ccpa/policy.go
djcsdy: I considered this but came to the conclusion that that would result in any fields that are not defined in `Regs.Ext` being stripped from the JSON, which is probably not what is expected. But I am happy to make this change if you are really sure.

djcsdy: Alternatively I could make the change you suggest but unmarshal to a map, so that all fields are preserved. Let me know what you think.

SyntaxNode: Gotcha. I vote for the map approach. I didn't think of that. Good idea.

djcsdy:
Should be ready for merge now.

PR #1306, opened 2020-05-18
eplanning: Support for apps

chino117:


adapters/eplanning/eplanning.go
```diff
@@ -117,7 +118,19 @@ func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqIn
...
+	} else {
+		requestTarget = pageDomain
+	}
+
+	var uri string
+	if request.App != nil {
+		uri = adapter.URI + fmt.Sprintf("/%s/%s/%s/%s?r=pbs&ncb=1&e=%s", clientID, dfpClientID, requestTarget, sec, strings.Join(spacesStrings, "+"))
+	} else {
+		uri = adapter.URI + fmt.Sprintf("/%s/%s/%s/%s?r=pbs&ncb=1&ur=%s&e=%s", clientID, dfpClientID, requestTarget, sec, url.QueryEscape(pageURL), strings.Join(spacesStrings, "+"))
+	}
```
guscarreon, line 133: For a more secure URL build, can we use golang's `"net/url"` library to add the path and the query to the URL?
```
uriObj, err := url.Parse(adapter.URI)
uriObj.Path = fmt.Sprintf("/%s/%s/%s/%s", clientID, dfpClientID, requestTarget, sec)
query := url.Values{}
query.Set("r", "pbs")
query.Set("ncb", 1)
query.Set("e", strings.Join(spacesStrings, "+"))
uriObj.RawQuery = query.Encode()

uri = uriObj.String()
```

adapters/eplanning/eplanning.go
```diff
@@ -135,6 +148,19 @@ func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqIn
...
+		}
+		if request.App.ID != "" {
+			uri = uri + fmt.Sprintf("&appid=%s", request.App.ID)
+		}
+		if request.Device != nil && request.Device.IFA != "" {
+			uri = uri + fmt.Sprintf("&ifa=%s", request.Device.IFA)
+		}
+		uri = uri + fmt.Sprintf("&app=%s", requestTargetInventory)
+	}
+
```
guscarreon, line 163: With the `"net/url"` library queries are easier to manage:
```
151     if request.App != nil {
152         if request.App.Name != "" {
153 -           uri = uri + fmt.Sprintf("&appn=%s", request.App.Name)
    +           query.Set("appn", request.App.Name)
154         }
155         if request.App.ID != "" {
156 -           uri = uri + fmt.Sprintf("&appid=%s", request.App.ID)
    +           query.Set("appid", request.App.ID)
157         }
158         if request.Device != nil && request.Device.IFA != "" {
159 -           uri = uri + fmt.Sprintf("&ifa=%s", request.Device.IFA)
    +           query.Set("ifa", request.Device.IFA)
160         }
161 -       uri = uri + fmt.Sprintf("&app=%s", requestTargetInventory)
    +           query.Set("app", requestTargetInventory)
162     }
163
```

adapters/eplanning/eplanning.go
```diff
@@ -117,24 +118,61 @@ func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqIn
...
+		errors = append(errors, err)
+		return nil, errors
+	}
+
+	uriObj.Path = uriObj.Path + fmt.Sprintf("/%s/%s/%s/%s", clientID, dfpClientID, requestTarget, sec)
+	query := url.Values{}
+	query.Set("r", "pbs")
+	query.Set("ncb", "1")
+	if request.App == nil {
+		query.Set("ur", url.QueryEscape(pageURL))
```
SyntaxNode, line 139: You don't need the `url.QueryEscape` call here since you're now using the `query` object. `query.Encode()` will escape the values. Consider adding a test case for when the pageURL contains a character which needs to be escaped.

PR #1311, opened 2020-05-20
Restore the AMP privacy exception as an option.

hhhjort:
Also note I have restored the ScrubStrategyUserNone even though it is not used in this code. It can be a source of confusion that some things are "silently" scrubbed and will be useful for the TCF 2 PR. The old (recently merged) code depended on the fact that if the scrubber was invoked at all, there were some things we would always scrub. This creates confusion when trying to extend the privacy scrubber, as you have to check in an unexpected place if there are more things that you may or may not want scrubbed in your new use case.

### DIFF
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -43,6 +43,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 
 	gdpr := extractGDPR(orig, usersyncIfAmbiguous)
 	consent := extractConsent(orig)
+	isAMP := (labels.RType == pbsmetrics.ReqTypeAMP) && gDPR.AMPException()
 
 	privacyEnforcement := privacy.Enforcement{
 		COPPA: orig.Regs != nil && orig.Regs.COPPA == 1,
@@ -65,7 +66,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 			privacyEnforcement.GDPR = false
 		}
 
-		privacyEnforcement.Apply(bidReq)
+		privacyEnforcement.Apply(bidReq, isAMP)
 	}
 
 	return
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -43,7 +43,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 
 	gdpr := extractGDPR(orig, usersyncIfAmbiguous)
 	consent := extractConsent(orig)
-	isAMP := (labels.RType == pbsmetrics.ReqTypeAMP) && gDPR.AMPException()
+	ampGDPRException := (labels.RType == pbsmetrics.ReqTypeAMP) && gDPR.AMPException()
 
 	privacyEnforcement := privacy.Enforcement{
 		COPPA: orig.Regs != nil && orig.Regs.COPPA == 1,
@@ -66,7 +66,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 			privacyEnforcement.GDPR = false
 		}
 
-		privacyEnforcement.Apply(bidReq, isAMP)
+		privacyEnforcement.Apply(bidReq, ampGDPRException)
 	}
 
 	return
--- a/privacy/enforcement.go
+++ b/privacy/enforcement.go
@@ -17,14 +17,14 @@ func (e Enforcement) Any() bool {
 }
 
 // Apply cleans personally identifiable information from an OpenRTB bid request.
-func (e Enforcement) Apply(bidRequest *openrtb.BidRequest, isAMP bool) {
-	e.apply(bidRequest, isAMP, NewScrubber())
+func (e Enforcement) Apply(bidRequest *openrtb.BidRequest, ampGDPRException bool) {
+	e.apply(bidRequest, ampGDPRException, NewScrubber())
 }
 
-func (e Enforcement) apply(bidRequest *openrtb.BidRequest, isAMP bool, scrubber Scrubber) {
+func (e Enforcement) apply(bidRequest *openrtb.BidRequest, ampGDPRException bool, scrubber Scrubber) {
 	if bidRequest != nil && e.Any() {
-		bidRequest.Device = scrubber.ScrubDevice(bidRequest.Device, !isAMP, e.getIPv6ScrubStrategy(), e.getGeoScrubStrategy())
-		bidRequest.User = scrubber.ScrubUser(bidRequest.User, e.getUserScrubStrategy(isAMP), e.getGeoScrubStrategy())
+		bidRequest.Device = scrubber.ScrubDevice(bidRequest.Device, e.getIPv6ScrubStrategy(), e.getGeoScrubStrategy())
+		bidRequest.User = scrubber.ScrubUser(bidRequest.User, e.getUserScrubStrategy(ampGDPRException), e.getGeoScrubStrategy())
 	}
 }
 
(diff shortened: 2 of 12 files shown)


privacy/scrubber.go
```diff
@@ -68,9 +74,11 @@ func (scrubber) ScrubDevice(device *openrtb.Device, ipv6 ScrubStrategyIPV6, geo
 	deviceCopy.DIDSHA1 = ""
 	deviceCopy.DPIDMD5 = ""
 	deviceCopy.DPIDSHA1 = ""
-	deviceCopy.IFA = ""
-	deviceCopy.MACMD5 = ""
-	deviceCopy.MACSHA1 = ""
+	if macAndIFA {
```
SyntaxNode, line 77: This is actually not related to AMP. We were only removing these fields for COPPA, but after consolation with our privacy advisors we are extending it to all privacy policies (COPPA, GDPR, CCPA). 

 Can we leave this logic as it was before please?

privacy/enforcement.go
```diff
@@ -17,14 +17,14 @@ func (e Enforcement) Any() bool {
...
+func (e Enforcement) Apply(bidRequest *openrtb.BidRequest, isAMP bool) {
+	e.apply(bidRequest, isAMP, NewScrubber())
 }
 
-func (e Enforcement) apply(bidRequest *openrtb.BidRequest, scrubber Scrubber) {
+func (e Enforcement) apply(bidRequest *openrtb.BidRequest, isAMP bool, scrubber Scrubber) {
 	if bidRequest != nil && e.Any() {
-		bidRequest.Device = scrubber.ScrubDevice(bidRequest.Device, e.getIPv6ScrubStrategy(), e.getGeoScrubStrategy())
-		bidRequest.User = scrubber.ScrubUser(bidRequest.User, e.getDemographicScrubStrategy(), e.getGeoScrubStrategy())
+		bidRequest.Device = scrubber.ScrubDevice(bidRequest.Device, !isAMP, e.getIPv6ScrubStrategy(), e.getGeoScrubStrategy())
```
mansinahar, line 26: Just a nit pick around readability:

The second argument to the `ScrubDevice` function is a flag to indicate if we wanna scrub `mac` and `IFA` values from the device struct and from this code I understand that we don't wanna scrub this for AMP requests. Can we please hold if mac and IFA values be scrubbed or not based on `isAMP` in a different variable and pass that variable? It just helps with readability. Right now for anyone to understand what passing `!isAMP` means in that method call, they need to go and check the function definition. If we create a new variable like `scrubMacAndIFA`, it becomes clear from this line itself. 

Also, other option could be to create something like `ScrubStrategyUser` for device as well.

exchange/utils.go
```diff
@@ -43,6 +43,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 
 	gdpr := extractGDPR(orig, usersyncIfAmbiguous)
 	consent := extractConsent(orig)
+	isAMP := (labels.RType == pbsmetrics.ReqTypeAMP) && gDPR.AMPException()
```
mansinahar, line 46: Correct me if I am wrong here but `AMPException` flag being set means that we don't want AMP requests to be considered for GDPR and so keeping it `false` by default means that AMP requests will be subject to GDPR as well, right?

So based on that the above line will actually set `isAMP` to true only if the request is of type AMP and if the flag is set (meaning that we wanna exempt AMP requests from GDPR). Isn't that the opposite of what we want? Meaning that we want `isAMP` to be true only if the request is of type AMP and if the flag is not set i.e. we don't want AMP to be exempted from enforcing GDPR. In other words:
```
isAMP := (labels.RType == pbsmetrics.ReqTypeAMP) && !gDPR.AMPException()
```

exchange/utils.go
```diff
@@ -65,7 +66,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 			privacyEnforcement.GDPR = false
 		}
 
-		privacyEnforcement.Apply(bidReq)
+		privacyEnforcement.Apply(bidReq, isAMP)
```
mansinahar, line 69: Thoughts around setting `isAMP` in the request context and passing the context to the `privacyEnforcement` function? 

Reason being, if we have other specific conditions based on which we want to tweak the privacy enforcement in future, this function signature would keep getting long, making it difficult to understand the purpose of the function from the signature itself and it might be helpful if we send all the information as part of the request context or a privacy specific struct.

privacy/enforcement.go
SyntaxNode: See my previous comment. Device IFA and MAC shouldn't be tied to AMP. We always want to scrub those. There are no issues reported with that logic.

exchange/utils.go
SyntaxNode: I read it as:
- `AMPException = true` Never scrub the buyer id or user id for AMP requests.
- `AMPException = false` Scrub the buyer id or user id for AMP requests per normal GDPR logic.

.. with the default of `false` not reverting the change back. So, the code as @hhhjort has it looks fine to me.

exchange/utils.go
SyntaxNode: I think this just a revert of the AMP commit. This looks like how it was written before. I didn't feel that isAMP was a good fit for the `Enforcement` struct and since it was going away soon, I didn't worry. That's obviously not longer the case (at least not for a little bit) - but I'm fine leaving it as is for now. There are no other conditions which I foresee being added. This one is a bit tricky because it's information from the endpoint, and not from the literal OpenRTB request itself.

mansinahar: Yeah, I agree that this information coming from the endpoint rather than the OpenRTB request itself makes it a little tricky but IMO, that is the kind of request level information that context can be used to hold and passed along in the code.

Nevertheless, I didn't know that's how this code was before (I have been away too long from Prebid Server 😅) and is part of the revert that this PR is doing so makes sense to keep it as is for this PR 👍

privacy/scrubber.go
hhhjort: The logic currently is we remove these unless the AMP flag is set in the config and the request is AMP. The idea is that the old AMP behavior is preserved for AMP requests, but the new behavior is preserved otherwise.

privacy/scrubber.go
```diff
@@ -90,17 +98,23 @@ func (scrubber) ScrubDevice(device *openrtb.Device, ipv6 ScrubStrategyIPV6, geo
...
+		userCopy.ID = ""
+		userCopy.Yob = 0
+		userCopy.Gender = ""
+	case ScrubStrategyUserBuyerIDOnly:
+		userCopy.BuyerUID = ""
+		userCopy.ID = ""
+	case ScrubStrategyUserAgeAndGender:
 		userCopy.Yob = 0
 		userCopy.Gender = ""
 	}
```
SyntaxNode, line 120: ScrubStrategyUserFull -> ScrubStrategyUserIDAndDemographic
ScrubStrategyUserBuyerIDOnly -> ScrubStrategyUserID
ScrubStrategyUserNone

privacy/enforcement.go
```diff
@@ -52,10 +52,13 @@ func (e Enforcement) getGeoScrubStrategy() ScrubStrategyGeo {
...
 
-func (e Enforcement) getDemographicScrubStrategy() ScrubStrategyDemographic {
+func (e Enforcement) getUserScrubStrategy(isAMP bool) ScrubStrategyUser {
 	if e.COPPA {
-		return ScrubStrategyDemographicAgeAndGender
+		return ScrubStrategyUserFull
 	}
 
-	return ScrubStrategyDemographicNone
+	if isAMP {
```
SyntaxNode, line 60:
```
if e.COPPA {
  return ScrubStrategyUserIDAndDemographic
}

if e.GDPR && isAMP {
 return ScrubStrategyUserNone
}

return ScrubStrategyUserID
```

privacy/enforcement.go
```diff
@@ -17,14 +17,14 @@ func (e Enforcement) Any() bool {
...
 
 // Apply cleans personally identifiable information from an OpenRTB bid request.
-func (e Enforcement) Apply(bidRequest *openrtb.BidRequest) {
-	e.apply(bidRequest, NewScrubber())
+func (e Enforcement) Apply(bidRequest *openrtb.BidRequest, isAMP bool) {
+	e.apply(bidRequest, isAMP, NewScrubber())
 }
 
-func (e Enforcement) apply(bidRequest *openrtb.BidRequest, scrubber Scrubber) {
+func (e Enforcement) apply(bidRequest *openrtb.BidRequest, isAMP bool, scrubber Scrubber) {
```
SyntaxNode, line 24: Lets rename `isAMP` to `ampGDPRException` for clarity.

privacy/scrubber.go
hhhjort: reverted

PR #1319, opened 2020-05-28
Add Adtarget server adapter

GeneGenie:


adapters/adtarget/adtarget.go
```diff
@@ -0,0 +1,190 @@
...
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+
+	reqs := make([]*adapters.RequestData, 0, totalReqs)
+
+	imps := request.Imp
+	request.Imp = make([]openrtb.Imp, 0, len(imps))
+	for sourceId, impIds := range imp2source {
```
SyntaxNode, line 58: Consider renaming `impIds` to something like `impIndexes` to avoid confusing an Id correlation with a slice index correlation.

adapters/adtarget/adtarget.go
```diff
@@ -0,0 +1,190 @@
...
+
+		reqs = append(reqs, &adapters.RequestData{
+			Method:  "POST",
+			Uri:     a.endpoint + fmt.Sprintf("?aid=%d", sourceId),
+			Body:    body,
+			Headers: headers,
+		})
+	}
+
+	if 0 == len(reqs) {
```
SyntaxNode, line 79: Is this check necessary? I'm not sure it matters if the response is nil vs an empty slice.

adapters/adtarget/adtarget.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil, errors
+	}
+
+	return reqs, errors
+
+}
+
+func (a *AdtargetAdapter) MakeBids(bidReq *openrtb.BidRequest, unused *adapters.RequestData, httpRes *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+
+	if httpRes.StatusCode == http.StatusNoContent {
```
SyntaxNode, line 89: It's also common for adapters to check for `StatusBadRequest` for a `BadInput` error and then anything else not `StatusOK` as an unexpected server error.

adapters/adtarget/adtarget.go
```diff
@@ -0,0 +1,190 @@
...
+
+	impExtBuffer, err = json.Marshal(&adtargetImpExt{
+		Adtarget: impExt,
+	})
+
+	if impExt.BidFloor > 0 {
+		imp.BidFloor = impExt.BidFloor
+	}
+
+	imp.Ext = impExtBuffer
```
SyntaxNode, line 181: Since this method is altering the value of `imp.Ext` and potentially `imp.BidFloor`, consider renaming the method from `validateImpression` to something along the lines of `validateImpressionAndSetExt` to make it clear this method also mutate the Imp.

static/bidder-info/adtarget.yaml
```diff
@@ -0,0 +1,10 @@
+maintainer:
+  email: "kamil@adtarget.com.tr"
+capabilities:
+  app:
+    mediaTypes:
+      - banner
```
SyntaxNode, line 6: Should the adapter code enforce that video impressions aren't valid for app requests?

SyntaxNode:
@GeneGenie Friendly ping. :)

GeneGenie:
@SyntaxNode hi there :) 
will submit changes asap

static/bidder-info/adtarget.yaml
GeneGenie: hm, actually there is no reason to ignore video for app

adapters/adtarget/adtarget.go
GeneGenie: not necessary but micro-optimized for callee

SyntaxNode:
> @SyntaxNode hi there :)
> will submit changes asap

Thank you. There was no need to address it immediately. Just wanted to give you a friendly reminder about the pending feedback. Looks good to me now. You just have a merge conflict to resolve.

PR #1320, opened 2020-05-28
Removed some redundant methods

telariaEng:
Removed the decompression method since that can be an option provided in the server config.


adapters/telaria/telaria.go
```diff
@@ -79,7 +77,7 @@ func GetHeaders(request *openrtb.BidRequest) *http.Header {
 	headers.Add("Content-Type", "application/json;charset=utf-8")
 	headers.Add("Accept", "application/json")
 	headers.Add("X-Openrtb-Version", "2.5")
-	headers.Add("Accept-Encoding", "gzip")
+	/*headers.Add("Accept-Encoding", "gzip")*/
```
SyntaxNode, line 80: You can remove the comment entirely.

telariaEng:
> > Removed the decompression method since that can be an option provided in the server config.
> 
> It's not even an option, actually. It's always enabled.

I see. What does the `enable_gzip` option in pbs.yaml do then?

SyntaxNode:
> I see. What does the `enable_gzip` option in pbs.yaml do then?

That enables gzip compression on incoming bid requests. Outgoing bid requests to adapters will automatically request gzip compression and silently handle decompression.

PR #1322, opened 2020-05-29
Timeout notification monitoring and debugging

hhhjort:


exchange/bidder.go
```diff
@@ -365,8 +374,24 @@ func (bidder *bidderAdapter) doTimeoutNotification(timeoutBidder adapters.Timeou
 		httpReq, err := http.NewRequest(toReq.Method, toReq.Uri, bytes.NewBuffer(toReq.Body))
 		if err == nil {
 			httpReq.Header = req.Headers
-			ctxhttp.Do(ctx, bidder.Client, httpReq)
-			// No validation yet on sending notifications
+			httpResp, err := ctxhttp.Do(ctx, bidder.Client, httpReq)
+			success := true
```
SyntaxNode, line 378: `success := (err == nil && httpResp.StatusCode >= 200 && httpResp.StatusCode < 300)`?

pbsmetrics/go_metrics.go
```diff
@@ -544,6 +544,11 @@ func (me *Metrics) RecordRequestQueueTime(success bool, requestType RequestType,
 
 }
 
+func (me *Metrics) RecordTimeoutNotice(sucess bool) {
+	// Not yet implemented for influx
```
SyntaxNode, line 548: Why not?

pbsmetrics/metrics.go
```diff
@@ -275,4 +275,5 @@ type MetricsEngine interface {
 	RecordStoredImpCacheResult(cacheResult CacheResult, inc int)
 	RecordPrebidCacheRequestTime(success bool, length time.Duration)
 	RecordRequestQueueTime(success bool, requestType RequestType, length time.Duration)
+	RecordTimeoutNotice(sucess bool)
```
SyntaxNode, line 278: Should this be a per-adapter metric?

config/util/loggers_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+var expected string = `This is test line 2
+This is test line 3
+`
+
+func TestLogRandomSample(t *testing.T) {
+	rand.Seed(1337)
```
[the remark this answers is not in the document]
hhhjort: Oh, that is an idea

pbsmetrics/go_metrics.go
hhhjort: We don't use it, I don't know who does, and this is just for debugging facebook timeout notifications. Given Facebook's willingness to work with prebid, I doubt anyone else needs this, so wasn't worth the effort. Influx and Prometheus have never been 1:1 in metrics, due to their different feature sets.

pbsmetrics/metrics.go
hhhjort: Eventually it may need to be. At this point I did not want to pollute the prometheus metric space with 59 unused adapter metrics.

hhhjort: Shouldn't be a breaking change. If we add an adapter label later, any Prometheus queries written against the current labels should still work, just sum over all adapters as expected. Could be missleading if we hardcode support only for facebook, and someone adds timeout support for another adapter without catching the hardcoded part in the code.

SyntaxNode: The sum over all adapters isn't automatic though. The PromQL query would return a series of metrics instead of a single value. Perhaps we can add a new per-adapter metric in the future when needed.

pbsmetrics/go_metrics.go
SyntaxNode: I understand that Xandr doesn't use Influx, but I'm quite sure there are others hosts who do use Influx and perhaps one of them will be successful in working with Facebook. I like to think we try to maintain parity between Influx and Prometheus.. although its admittedly not perfect.

config/util/loggers_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"bytes"
+	"fmt"
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var expected string = `This is test line 2
```
SyntaxNode, line 12: Please move into the `TestLogRandomSample` method as a const.

exchange/bidder.go
```diff
@@ -82,16 +87,20 @@ type pbsOrtbSeatBid struct {
...
+		me:       me,
 	}
 }
 
 type bidderAdapter struct {
-	Bidder adapters.Bidder
-	Client *http.Client
+	Bidder   adapters.Bidder
+	Client   *http.Client
+	DebugCfg config.Debug
```
SyntaxNode, line 102: Consider renaming to DebugConfig for ease of reading.

config/config.go
```diff
@@ -428,6 +430,30 @@ type DefReqFiles struct {
...
+
+func (cfg *Debug) validate(errs configErrors) configErrors {
+	return cfg.TimeoutNotification.validate(errs)
+}
+
+type TimeoutNotification struct {
+	// Log timeout notifications in the application log
+	Log bool `mapstructure:"log"`
+	// Fraction of notifications to log
+	Sampling float32 `mapstructure:"sampling"`
```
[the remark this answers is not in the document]
SyntaxNode: Do you not agree with this one?

config/util/loggers_test.go
```diff
@@ -0,0 +1,31 @@
...
+This is test line 3
+`
+
+func TestLogRandomSample(t *testing.T) {
```
SyntaxNode, line 16: Now that we have extracted LogRandomSample, you can test the `logRandomSampleImpl` method with a mocked constant value and only need to test the two conditions of 'log' or 'ignore'.

hhhjort: True, but is it worth redoing the test to break it up into those two separate cases?

config/config.go
```diff
@@ -428,6 +430,30 @@ type DefReqFiles struct {
 	FileName string `mapstructure:"name"`
 }
 
+type Debug struct {
+	TimeoutNotification TimeoutNotification `mapstructure:"timeout_notification"`
+}
+
+func (cfg *Debug) validate(errs configErrors) configErrors {
+	return cfg.TimeoutNotification.validate(errs)
+}
```
guscarreon, line 439: Should this function run inside `func (cfg *Configuration) validate() configErrors`? I can't find the place where it gets called.

config/util/loggers.go
```diff
@@ -0,0 +1,24 @@
...
+
+type logMsg func(string, ...interface{})
+
+type randomGenerator func() float32
+
+// LogRandomSample will log a randam sample of the messages it is sent, based on the chance to log
+// chance = 1.0 => always log,
+// chance = 0.0 => never log
+func LogRandomSample(msg string, logger logMsg, chance float32) {
+	logRandomSampleImpl(msg, logger, chance, rand.Float32)
```
guscarreon, line 15: Locally I added the test that follows to `config/util/loggers_test.go` and it failed with the error message shown below. Do we need to call `rand.New` from inside `LogRandomSample(msg string, logger logMsg, chance float32)` too? Can we add a test to cover this function?
```
func TestLogRandomSampleWrapper(t *testing.T) {
	var buf bytes.Buffer

	mylogger := func(msg string, args ...interface{}) {
		buf.WriteString(fmt.Sprintf(fmt.Sprintln(msg), args...))
	}

	LogRandomSample("This is test line 1", mylogger, 0.5)
	LogRandomSample("This is test line 2", mylogger, 0.5)
	LogRandomSample("This is test line 3", mylogger, 0.5)
	LogRandomSample("This is test line 4", mylogger, 0.5)
	LogRandomSample("This is test line 5", mylogger, 0.5)

	assert.EqualValues(t, expected, buf.String())
}
```
**Result**
```
config/util/loggers_test.go|46| 
|| 	Error Trace:	loggers_test.go:46
|| 	Error:      	Not equal: 
|| 	            	expected: "This is test line 2\nThis is test line 3\n"
|| 	            	actual  : "This is test line 4\nThis is test line 5\n"
|| 	            	
|| 	            	Diff:
|| 	            	--- Expected
|| 	            	+++ Actual
|| 	            	@@ -1,3 +1,3 @@
|| 	            	-This is test line 2
|| 	            	-This is test line 3
|| 	            	+This is test line 4
|| 	            	+This is test line 5
|| 	            	 
|| 	Test:       	TestLogRandomSampleWrapper
```

PR #1328, opened 2020-06-02
Change on eplanning endpoint (hostname)

sebaperez:
I'm sorry guys, after a review we noticed that some changes on protocol and hostname are required.

This PR is part of the already merged #1327 

Changes are on endpoint protocol and hostname. This changes won't affect older versions


config/config.go
```diff
@@ -718,7 +718,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.datablocks.endpoint", "http://{{.Host}}/openrtb2?sid={{.SourceId}}")
 	v.SetDefault("adapters.emx_digital.endpoint", "https://hb.emxdgt.com")
 	v.SetDefault("adapters.engagebdr.endpoint", "http://dsp.bnmla.com/hb")
-	v.SetDefault("adapters.eplanning.endpoint", "https://ads.us.e-planning.net/pbs/1")
+	v.SetDefault("adapters.eplanning.endpoint", "http://rtb.e-planning.net/pbs/1")
```
SyntaxNode, line 721: Curious, why the switch away from https? Wouldn't a secure channel be preferred?

sebaperez: https requires more CPU usage than http. It's OK for us to use http as there isn't a browser request involved here, just a connection between a PBS instance and our server

PR #1331, opened 2020-06-03
feat: Add new logger module - Pubstack Analytics Module

gpolaert:
Add a new logger module: Pubstack
Ability for users to send bid events directly to Pubstack.

### DIFF
--- a/analytics/config/config.go
+++ b/analytics/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"github.com/golang/glog"
 	"github.com/prebid/prebid-server/analytics"
+	"github.com/prebid/prebid-server/analytics/clients"
 	"github.com/prebid/prebid-server/analytics/filesystem"
 	"github.com/prebid/prebid-server/analytics/pubstack"
 	"github.com/prebid/prebid-server/config"
@@ -20,6 +21,7 @@ func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
 	}
 	if analytics.Pubstack.Enabled {
 		pubstackModule, err := pubstack.NewPubstackModule(
+			clients.GetDefaultHttpInstance(),
 			analytics.Pubstack.ScopeId,
 			analytics.Pubstack.IntakeUrl,
 			analytics.Pubstack.ConfRefresh,
--- a/analytics/config/config.go
+++ b/analytics/config/config.go
@@ -9,12 +9,20 @@ import (
 	"github.com/prebid/prebid-server/config"
 )
 
+type analyticsModule analytics.PBSAnalyticsModule
+
+type pbsAnalyticsModule struct {
+	enabledModules []analyticsModule
+}
+
 //Modules that need to be logged to need to be initialized here
 func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
-	modules := make(enabledAnalytics, 0)
+
+	instance := &pbsAnalyticsModule{enabledModules: make([]analyticsModule, 0)}
+
 	if len(analytics.File.Filename) > 0 {
 		if mod, err := filesystem.NewFileLogger(analytics.File.Filename); err == nil {
-			modules = append(modules, mod)
+			instance.enabledModules = append(instance.enabledModules, mod)
 		} else {
 			glog.Fatalf("Could not initialize FileLogger for file %v :%v", analytics.File.Filename, err)
 		}
@@ -29,43 +37,40 @@ func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
 			analytics.Pubstack.Buffers.BufferSize,
 			analytics.Pubstack.Buffers.Timeout)
 		if err == nil {
-			modules = append(modules, pubstackModule)
+			instance.enabledModules = append(instance.enabledModules, pubstackModule)
 		} else {
 			glog.Fatalf("Could not initialize PubstackModule: %v", err)
 		}
 	}
-	return modules
+	return instance
 }
 
-//Collection of all the correctly configured analytics modules - implements the PBSAnalyticsModule interface
-type enabledAnalytics []analytics.PBSAnalyticsModule
-
-func (ea enabledAnalytics) LogAuctionObject(ao *analytics.AuctionObject) {
-	for _, module := range ea {
+func (pam pbsAnalyticsModule) LogAuctionObject(ao *analytics.AuctionObject) {
+	for _, module := range pam.enabledModules {
 		module.LogAuctionObject(ao)
 	}
 }
 
-func (ea enabledAnalytics) LogVideoObject(vo *analytics.VideoObject) {
-	for _, module := range ea {
+func (pam pbsAnalyticsModule) LogVideoObject(vo *analytics.VideoObject) {
+	for _, module := range pam.enabledModules {
 		module.LogVideoObject(vo)
 	}
 }
 
-func (ea enabledAnalytics) LogCookieSyncObject(cso *analytics.CookieSyncObject) {
-	for _, module := range ea {
+func (pam pbsAnalyticsModule) LogCookieSyncObject(cso *analytics.CookieSyncObject) {
+	for _, module := range pam.enabledModules {
 		module.LogCookieSyncObject(cso)
 	}
 }
 
-func (ea enabledAnalytics) LogSetUIDObject(so *analytics.SetUIDObject) {
-	for _, module := range ea {
+func (pam pbsAnalyticsModule) LogSetUIDObject(so *analytics.SetUIDObject) {
+	for _, module := range pam.enabledModules {
 		module.LogSetUIDObject(so)
 	}
 }
 
-func (ea enabledAnalytics) LogAmpObject(ao *analytics.AmpObject) {
-	for _, module := range ea {
+func (pam pbsAnalyticsModule) LogAmpObject(ao *analytics.AmpObject) {
+	for _, module := range pam.enabledModules {
 		module.LogAmpObject(ao)
 	}
 }
--- a/analytics/config/config.go
+++ b/analytics/config/config.go
@@ -39,7 +39,7 @@ func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
 		if err == nil {
 			instance.enabledModules = append(instance.enabledModules, pubstackModule)
 		} else {
-			glog.Fatalf("Could not initialize PubstackModule: %v", err)
+			glog.Errorf("Could not initialize PubstackModule: %v", err)
 		}
 	}
 	return instance
--- a/analytics/config/config.go
+++ b/analytics/config/config.go
@@ -9,20 +9,12 @@ import (
 	"github.com/prebid/prebid-server/config"
 )
 
-type analyticsModule analytics.PBSAnalyticsModule
-
-type pbsAnalyticsModule struct {
-	enabledModules []analyticsModule
-}
-
 //Modules that need to be logged to need to be initialized here
 func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
-
-	instance := &pbsAnalyticsModule{enabledModules: make([]analyticsModule, 0)}
-
+	modules := make(enabledAnalytics, 0)
 	if len(analytics.File.Filename) > 0 {
 		if mod, err := filesystem.NewFileLogger(analytics.File.Filename); err == nil {
-			instance.enabledModules = append(instance.enabledModules, mod)
+			modules = append(modules, mod)
 		} else {
 			glog.Fatalf("Could not initialize FileLogger for file %v :%v", analytics.File.Filename, err)
 		}
@@ -37,40 +29,43 @@ func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
 			analytics.Pubstack.Buffers.BufferSize,
 			analytics.Pubstack.Buffers.Timeout)
 		if err == nil {
-			instance.enabledModules = append(instance.enabledModules, pubstackModule)
+			modules = append(modules, pubstackModule)
 		} else {
 			glog.Errorf("Could not initialize PubstackModule: %v", err)
 		}
 	}
-	return instance
+	return modules
 }
 
-func (pam pbsAnalyticsModule) LogAuctionObject(ao *analytics.AuctionObject) {
-	for _, module := range pam.enabledModules {
+//Collection of all the correctly configured analytics modules - implements the PBSAnalyticsModule interface
+type enabledAnalytics []analytics.PBSAnalyticsModule
+
+func (ea enabledAnalytics) LogAuctionObject(ao *analytics.AuctionObject) {
+	for _, module := range ea {
 		module.LogAuctionObject(ao)
 	}
 }
 
-func (pam pbsAnalyticsModule) LogVideoObject(vo *analytics.VideoObject) {
-	for _, module := range pam.enabledModules {
+func (ea enabledAnalytics) LogVideoObject(vo *analytics.VideoObject) {
+	for _, module := range ea {
 		module.LogVideoObject(vo)
 	}
 }
 
-func (pam pbsAnalyticsModule) LogCookieSyncObject(cso *analytics.CookieSyncObject) {
-	for _, module := range pam.enabledModules {
+func (ea enabledAnalytics) LogCookieSyncObject(cso *analytics.CookieSyncObject) {
+	for _, module := range ea {
 		module.LogCookieSyncObject(cso)
 	}
 }
 
-func (pam pbsAnalyticsModule) LogSetUIDObject(so *analytics.SetUIDObject) {
-	for _, module := range pam.enabledModules {
+func (ea enabledAnalytics) LogSetUIDObject(so *analytics.SetUIDObject) {
+	for _, module := range ea {
 		module.LogSetUIDObject(so)
 	}
 }
 
-func (pam pbsAnalyticsModule) LogAmpObject(ao *analytics.AmpObject) {
-	for _, module := range pam.enabledModules {
+func (ea enabledAnalytics) LogAmpObject(ao *analytics.AmpObject) {
+	for _, module := range ea {
 		module.LogAmpObject(ao)
 	}
 }
--- a/analytics/config/config.go
+++ b/analytics/config/config.go
@@ -4,6 +4,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/prebid/prebid-server/analytics"
 	"github.com/prebid/prebid-server/analytics/filesystem"
+	"github.com/prebid/prebid-server/analytics/pubstack"
 	"github.com/prebid/prebid-server/config"
 )
 
@@ -15,6 +16,20 @@ func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
 			modules = append(modules, mod)
 		} else {
 			glog.Fatalf("Could not initialize FileLogger for file %v :%v", analytics.File.Filename, err)
+		}
+	}
+	if analytics.Pubstack.Enabled {
+		pubstackModule, err := pubstack.NewPubstackModule(
+			analytics.Pubstack.ScopeId,
+			analytics.Pubstack.IntakeUrl,
+			analytics.Pubstack.ConfRefresh,
+			analytics.Pubstack.Buffers.EventCount,
+			analytics.Pubstack.Buffers.BufferSize,
+			analytics.Pubstack.Buffers.Timeout)
+		if err == nil {
+			modules = append(modules, pubstackModule)
+		} else {
+			glog.Fatalf("Could not initialize PubstackModule: %v", err)
 		}
 	}
 	return modules
--- a/analytics/pubstack/README.md
+++ b/analytics/pubstack/README.md
@@ -1,0 +1,28 @@
+# Pubstack Analytics
+
+In order to use the pubstack analytics module, it needs to be configured by the host.
+
+You can configure the server using the following environment variables:
+
+```bash
+export PBS_ANALYTICS_PUBSTACK_ENABLED="true"
+export PBS_ANALYTICS_PUBSTACK_ENDPOINT="https://openrtb.preview.pubstack.io/v1/openrtb2"
+export PBS_ANALYTICS_PUBSTACK_SCOPEID=<your scopeId here> # should be an UUIDv4
+```
+
+Or using the pbs configuration file and by appending the following block:
+
+```yaml
+analytics:
+    pubstack:
+      # Required properties
+      enabled: true
+      endpoint: "https://openrtb.preview.pubstack.io/v1/openrtb2"
+      scopeid: "<scopeId>" # The scopeId provided by the Pubstack Support Team
+      # Optional properties (advanced configuration)
+      configuration_refresh_delay: "2h" # Dynamic configuration delay
+      buffers: # Flush events to Pubstack when (first condition reached)
+        size: "2MB" # greater than 2MB
+        count : 100 # greater than 100 events
+        timeout: "15m" # greater than 15 minutes
+```
--- a/analytics/pubstack/eventchannel/eventchannel.go
+++ b/analytics/pubstack/eventchannel/eventchannel.go
@@ -1,0 +1,146 @@
+package eventchannel
+
+import (
+	"bytes"
+	"compress/gzip"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/golang/glog"
+)
+
+type Metrics struct {
+	bufferSize int64
+	eventCount int64
+}
+type Limit struct {
+	maxByteSize   int64
+	maxEventCount int64
+	maxTime       time.Duration
+}
+type EventChannel struct {
+	gz   *gzip.Writer
+	buff *bytes.Buffer
+
+	ch      chan []byte
+	endCh   chan int
+	metrics Metrics
+	mux     sync.Mutex
+	send    Sender
+	limit   Limit
+}
+
+func NewEventChannel(sender Sender, maxByteSize, maxEventCount int64, maxTime time.Duration) *EventChannel {
+	b := &bytes.Buffer{}
+	gzw := gzip.NewWriter(b)
+
+	c := EventChannel{
+		gz:      gzw,
+		buff:    b,
+		ch:      make(chan []byte),
+		endCh:   make(chan int),
+		metrics: Metrics{},
+		send:    sender,
+		limit:   Limit{maxByteSize, maxEventCount, maxTime},
+	}
+
+	termCh := make(chan os.Signal)
+	signal.Notify(termCh, os.Interrupt, syscall.SIGTERM)
+
+	go c.start(termCh)
+	return &c
+}
+
+func (c *EventChannel) Push(event []byte) {
+	c.ch <- event
+}
+
+func (c *EventChannel) Close() {
+	c.endCh <- 1
+}
+
+func (c *EventChannel) buffer(event []byte) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
+	_, err := c.gz.Write(event)
+	if err != nil {
+		glog.Warning("[pubstack] fail to compress, skip the event")
+		return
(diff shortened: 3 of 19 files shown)


gpolaert:
Let me know what I have to do this that PR ;)

SyntaxNode:
Sweet! The first analytics module! :)

We'll review the PR tomorrow.

gpolaert:
:pray: Let me know how we can improve this first version/draft ;)

analytics/pubstack/eventchannel/channel.go
```diff
@@ -0,0 +1,127 @@
...
+	req, err := http.NewRequest(http.MethodPost, c.intake.String(), bytes.NewReader(payload))
+	if err != nil {
+		glog.Error(err)
+		return
+	}
+
+	req.Header.Set("Content-Type", "application/octet-stream")
+	req.Header.Set("Content-Encoding", "gzip")
+
+	resp, err := http.DefaultClient.Do(req)
```
SyntaxNode, line 98: You shouldn't use the default client. Instead, I think we should create an `analyticsHttpClient` or use the existing `generalHttpClient` and pass it down from the router composition root. We need to be able to configure values such as timeouts. Also, we have an upcoming project to optimize outgoing connections and we may request related changes later this summer.

analytics/pubstack/eventchannel/channel.go
```diff
@@ -0,0 +1,127 @@
...
+		return
+	}
+	if resp.StatusCode != http.StatusOK {
+		glog.Errorf("wrong code received %d instead of %d", resp.StatusCode, http.StatusOK)
+		return
+	}
+}
+
+func NewChannel(intake, route string, maxSize, maxCount int64, maxTime time.Duration) *Channel {
+	u, _ := url.Parse(intake)
```
SyntaxNode, line 109: Please don't ignore parse errors.

analytics/pubstack/helpers.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"github.com/prebid/prebid-server/analytics"
+	"net/http"
+	"net/url"
+)
+
+type RequestType string
+
+const (
+	HEALTH = "/v1/health"
```
SyntaxNode, line 15: Capitalization of constants is not advised in Go due to the capitalization of the first character denoting package scope. This should be `health` or `healthPath`. 

I realize older code in this repo is also using this anti-pattern, so I understand where you got it from.

analytics/pubstack/helpers.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func testEndpoint(endpoint *url.URL) error {
```
SyntaxNode, line 18: This looks unused.

analytics/pubstack/helpers.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	if r.StatusCode != http.StatusOK {
+		return fmt.Errorf("receive code %d instead of %d", r.StatusCode, http.StatusOK)
+	}
+	return nil
+}
+
+func sendPayloadToTarget(payload []byte, endpoint string) error {
+	resp, err := http.Post(endpoint, "application/json", bytes.NewBuffer(payload))
```
SyntaxNode, line 33: See earlier comment. Please don't use the default http client.

analytics/pubstack/helpers/json.go
```diff
@@ -0,0 +1,95 @@
...
+	}{
+		Scope: scope,
+		alias: (*alias)(ao),
+	})
+
+	if err == nil {
+		b = append(b, byte('\n'))
+		return b, nil
+	}
+	return []byte(""), fmt.Errorf("transactional logs error: auction object badly formed %v", err)
```
SyntaxNode, line 25: You should be fine to return `nil` instead of `[]byte("")` and avoid an allocation. This comment applies to all similar methods.

analytics/pubstack/helpers.go
```diff
@@ -0,0 +1,59 @@
...
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("wrong code received %d instead of %d", resp.StatusCode, http.StatusOK)
+	}
+	return nil
+}
+
+func jsonifyAuctionObject(ao *analytics.AuctionObject, scope string) ([]byte, error) {
```
SyntaxNode, line 44: Is this a duplicate of the one in json helpers below?

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,276 @@
...
+
+type Configuration struct {
+	ScopeId  string          `json:"scopeId"`
+	Endpoint string          `json:"endpoint"`
+	Features map[string]bool `json:"features"`
+}
+
+// routes for events
+const (
+	AUCTION    = "auction"
```
SyntaxNode, line 37: See earlier comment about avoiding all caps constants and rename appropriately.

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	return &bufferConfig{
+		pDuration,
+		int64(count),
+		pSize,
+	}, nil
+}
+
+type PubstackModule struct {
+	chans    map[string]*eventchannel.Channel
```
SyntaxNode, line 67: Consider not calling it `chans`. Too confusing with Go's special `chan`.

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func getConfiguration(scope string, intake string) (*Configuration, error) {
+	u, err := url.Parse(intake)
+	if err != nil {
+		return nil, err
+	}
+
+	u.Path = path.Join(u.Path, "bootstrap")
+	q, _ := url.ParseQuery(u.RawQuery)
```
SyntaxNode, line 206: Why are you reparsing `u`?

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,276 @@
...
+
+	q.Add("scopeId", scope)
+	u.RawQuery = q.Encode()
+
+	res, err := http.DefaultClient.Get(u.String())
```
SyntaxNode, line 211: See earlier comment, please don't use the default client.

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,276 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
+	data, err := ioutil.ReadAll(res.Body)
```
SyntaxNode, line 217: Please use `json.Decoder` for unmarshaling from a stream. See: https://stackoverflow.com/questions/21197239/decoding-json-using-json-unmarshal-vs-json-newdecoder-decode?rq=1

config/config.go
```diff
@@ -190,6 +191,20 @@ type FileLogs struct {
 	Filename string `mapstructure:"filename"`
 }
 
+type Pbstck struct {
```
SyntaxNode, line 194: Why not "Pubstack"?

config/config.go
```diff
@@ -190,6 +191,20 @@ type FileLogs struct {
...
+	Enabled     bool         `mapstructure:"enabled"`
+	ScopeId     string       `mapstructure:"scopeid"`
+	IntakeUrl   string       `mapstructure:"endpoint"`
+	Buffers     PbstckBuffer `mapstructure:"buffers"`
+	ConfRefresh string       `mapstructure:"configuration_refresh_delay"`
+}
+
+type PbstckBuffer struct {
```
SyntaxNode, line 202: Why not "PubstackBuffer"?

analytics/pubstack/eventchannel/channel.go
```diff
@@ -0,0 +1,127 @@
...
+			}
+		// time between flushes has passed
+		case <-ticker.C:
+			c.flush()
+		}
+	}
+}
+
+func (c *Channel) flush() {
+	c.resetMetrics()
```
SyntaxNode, line 73: I'm concerned of a race condition between flushing and writing new data. Did you run tests with race detection turned on?

analytics/pubstack/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Or using the pbs configuration file and by appending the following block:
+
+```yaml
+analytics:
+    pubstack:
+      enabled: true
+      endpoint: "https://openrtb.preview.pubstack.io/v1/openrtb2"
+      scopeid: <your scopeId here> # should be an UUIDv4
+```
```
SyntaxNode, line 22: Please include details of configuring the buffer and providing memory usage guidance to hosts.

analytics/pubstack/eventchannel/channel.go
```diff
@@ -0,0 +1,127 @@
...
+	intake  *url.URL
+	gz      *gzip.Writer
+	buff    *bytes.Buffer
+	ch      chan []byte
+	metrics ChannelMetrics
+}
+
+// Add : add a new event to be processed
+func (c *Channel) Add(event []byte) {
+	c.ch <- event
```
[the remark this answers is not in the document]
SyntaxNode: Thank you. :)

analytics/pubstack/eventchannel/channel.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// read gzipped content
+	payload := make([]byte, c.buff.Len())
+	_, err := c.buff.Read(payload)
+	if err != nil {
+		glog.Warning("Fail to read gzipped buffer")
+	}
+
+	// clean buffers and writers
+	c.buff = bytes.NewBufferString("")
```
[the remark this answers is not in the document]
SyntaxNode: This one a feel a bit strongly about. Why not `c.buff.Reset()`?

analytics/pubstack/eventchannel/eventchannel.go
```diff
@@ -0,0 +1,127 @@
...
+	req, err := http.NewRequest(http.MethodPost, c.intake.String(), bytes.NewReader(payload))
+	if err != nil {
+		glog.Error(err)
+		return
+	}
+
+	req.Header.Set("Content-Type", "application/octet-stream")
+	req.Header.Set("Content-Encoding", "gzip")
+
+	resp, err := http.DefaultClient.Do(req)
```
SyntaxNode, line 98: It's really not preferable to use the default client. I'd like to see the http client included in the EventChannel struct and also in the PubstackModule struct to support the call to refresh the configuration. If you're not comfortable making the change in the core to pass down a different http client.. you can reference the default client at the top most level and we can wire it up differently in a follow-up PR.

SyntaxNode:
Thank you for the first pass update. Heading in a good direction. The build passed on Go 1.12 and 1.13. It failed on 1.14 with the error:
```
[pubstack] Fail to initialize pubstack module, fail to acquire configuration
```
.. which looks like a failed http call. You'll want to stub out the http client to use canned responses in place of actual calls during the tests. My recommendation to allow configuration of the http client to avoid using the default client would help here.

analytics/config/config.go
```diff
@@ -17,6 +18,18 @@ func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
...
+	if analytics.Pubstack.Enabled {
+		pubstack, err := pubstack.NewPubstackModule(
+			analytics.Pubstack.ScopeId,
+			analytics.Pubstack.IntakeUrl,
+			analytics.Pubstack.ConfRefresh,
+			analytics.Pubstack.Buffers.EventCount,
+			analytics.Pubstack.Buffers.BufferSize,
+			analytics.Pubstack.Buffers.Timeout)
+		if err == nil {
+			modules = append(modules, pubstack)
```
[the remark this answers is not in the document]
gpolaert: No sure to understand here, what do you mean?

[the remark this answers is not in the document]
SyntaxNode: In `NewPubstackModule` I see a pattern of writing to the log and returning an error. I was curious what you thought about only returning the error and then writing it to the log at this location.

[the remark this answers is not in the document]
gpolaert: Make sense :)

analytics/pubstack/eventchannel/channel.go
gpolaert: I have reworked that part. I'll appreciate if you can look again here.
The race detection only shows only an issue with the thread initializer, like all adapters, I think we are good there.

analytics/pubstack/eventchannel/eventchannel.go
gpolaert: As we can see, I've create a dedicated package/client to centralize default client calls
This part must be refactored after your client redesign

analytics/pubstack/eventchannel/channel.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+func (c *Channel) flush() {
+	c.resetMetrics()
+	// finish writing gzip header
+	c.gz.Close()
+
+	// read gzipped content
+	payload := make([]byte, c.buff.Len())
```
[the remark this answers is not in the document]
gpolaert: I've kept the buffer copy to process asynchronously the http call. 
I've also refactored that part, let me know if you are okay with this design

analytics/pubstack/helpers/json.go
```diff
@@ -0,0 +1,95 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/prebid/prebid-server/analytics"
+)
+
+// JsonifyAuctionObject helpers to serialize auction into json line
+func JsonifyAuctionObject(ao *analytics.AuctionObject, scope string) ([]byte, error) {
+	type alias analytics.AuctionObject
+	b, err := json.Marshal(&struct {
```
hhhjort, line 13: Is there a reason you define the alias type, rather than simply declaring that the value of alias in the struct is of the appropriate type?

Have you tested that this will marshal correctly with `alias` unexported?

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,301 @@
...
+	endpoint, err := url.Parse(baseUrl)
+	if err != nil {
+		glog.Fatal(err)
+	}
+	endpoint.Path = path.Join(endpoint.Path, "intake", module)
+	return eventchannel.NewHttpSender(endpoint.String())
+}
+
+func (p *PubstackModule) initEventChannels() {
+	if p.isFeatureEnable(amp) {
```
hhhjort, line 272: Since `initEventChannels()` is called each time the config is refreshed, don't we have a resource/memory leak. Say for example the config is set to refresh hourly and an instance of PBS is left running for a week. That would result in 168 amp event channels, each with its own go routine listening for messages. Of course only one instance would receive any messages other than the periodic flush buffers timer.

gpolaert: Good catch! I've refactored that part, and closed the channels

hhhjort: I believe you have created a race condition here. By deleting the old channels first, you create a temporary condition where the map `p.eventChannels` will not have a value for a feature, like amp, while `p.isFeatureEnabled` is still true. You use only `p.isFeatureEnabled` to guard against calling `p.eventChannels[feature].Push(payload)` on a feature that is not present in `p.eventChannels`. You could put the existence check in `p.logObject`, but that would still leave a tiny window for race conditions. Better to save the pointers to the old channels, and then close the old channel after the new channel has been slotted into the map.

This may still not close all the race condition possibility upon reflection. Updating a map with a new value may not be an atomic action, certainly not for a struct where values need to be copied in. In the case of a pointer, it is less clear, as in go pointers are not a simple address, but also contain metadata about the object they are pointing to. You may need something like this in `logObject` to close out the possibility completely:
```
ch, ok := p.eventChannels[feature]
if ok && ch != nil {
    ch.Push(payload)
}
```
I would hope go is smart enough to not signal `ok` on pulling a map value if the map is mid update, but I have not researched this point before.

hhhjort: Perhaps the safest approach is to add an `isActive bool` property to `PubstackModule`, set it to false at the start of `setEventChannels()`, and not set it back to true until the new channels have been created. Then have a line to have `logObject()` return before doing anything if this flag is false. Alternatively you could have `setEventChannels()` set the same `p.mux.Lock()` to lock out `logObject()` until the channels are stable again. You should also consider locking your module until the initial config load is done, you don't want the analytics to start flowing before the module is up and running. ... Reviewing again, it seems you don't return the pubstack object until the initial config load is done, so that is probably not an issue.

gpolaert: Thanks for your reply, I will check it today

gpolaert: *Disclaimer, I've rebased the branch on the lastest master and squash all previous commits to clean the PR.*

What do you think moving the Mutex in each `Log*Object` methods? Since, we can't use generics, it doesn't make a lot of sense to have the `log object` method jsut for pushing the events.

I've also refactored the `setEventChannels` to `configure`. The goal is to create an "atomic" operation (with the Mutex) to update the module. So, during the update, the `Log*Object` methods are blocked/synced. 

I think it removes the race condition.

It's been a long time I didn't code in go, so I have to switch my mind from an eventloop model (js) to a multi thread app :sunglasses:

gpolaert:
@SyntaxNode How can I re-enable travis?

SyntaxNode:
> @SyntaxNode How can I re-enable travis?

It's enabled, but it's been a bit flaky with reporting back to GitHub. I'll kick it off again.

analytics/pubstack/pubstack_module.go
hhhjort: Yes, that should do it. The race condition is on losing the channel, so at the time of the `p.eventChannels[].Push(payload)` the channel must exist. You do run into the possibility of the shutdown being processed while there are still payloads in the queue, so you could potentially lose a few events. But that should be an insignificant fraction of the total, especially if you have a nice long delay between config refresh events.

gpolaert:
Hello,

What's the next step here :)?

SyntaxNode:
> Hello,
> 
> What's the next step here :)?

Every code PR needs two approvals before being merged. I'm assigned as the second reviewer, but I've had a meeting heavy week. Sorry. I'm unintentionally holding you up here. I'll give it another pass today or worst case over the weekend. I'd really like to get this included in the next release.

I greatly appreciate your responsiveness and support to the feedback. Building a module has a lot more complexity than an adapter.

analytics/pubstack/eventchannel/eventchannel.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	termCh := make(chan os.Signal)
+	signal.Notify(termCh, os.Interrupt, syscall.SIGTERM)
+
+	go c.start(termCh)
+	return &c
+}
+
+func (c *EventChannel) Push(event []byte) {
```
SyntaxNode, line 57: Please add tests for the Push and Close methods.

analytics/pubstack/eventchannel/eventchannel.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// send events (async)
+	go c.send(payload)
+}
+
+func (c *EventChannel) start(termCh chan os.Signal) {
+	ticker := time.NewTicker(c.limit.maxTime)
+
+	for {
+		select {
```
SyntaxNode, line 126: Please add test coverage for the select statement cases. There is currently no coverage.

analytics/pubstack/eventchannel/sender.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/golang/glog"
+	"github.com/prebid/prebid-server/analytics/clients"
+	"net/http"
+	"net/url"
+	"path"
+)
+
+type Sender = func(payload []byte)
+
+func NewHttpSender(endpoint string) Sender {
```
SyntaxNode, line 14: Please add a httpClient parameter and use it instead of `clients.GetDefaultHttpInstance()`. The client should be passed down from `PubstackModule.configure`.

analytics/pubstack/eventchannel/sender.go
```diff
@@ -0,0 +1,43 @@
+package eventchannel
```
SyntaxNode, line 1: Please add tests for this file. There is currently no coverage.

analytics/pubstack/helpers/json.go
```diff
@@ -0,0 +1,89 @@
+package helpers
```
SyntaxNode, line 1: Please add tests for this file.

analytics/pubstack/eventchannel/eventchannel.go
```diff
@@ -0,0 +1,146 @@
...
+	maxTime       time.Duration
+}
+type EventChannel struct {
+	gz   *gzip.Writer
+	buff *bytes.Buffer
+
+	ch      chan []byte
+	endCh   chan int
+	metrics Metrics
+	mux     sync.Mutex
```
SyntaxNode, line 31: Exactly which memory is being protected by the mux? It looks like the gz writier. Consider adding a comment to make this clear.

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("fail to parse the module args, arg=analytics.pubstack.configuration_refresh_delay, :%v", err)
+	}
+
+	bufferCfg, err := newBufferConfig(maxEventCount, maxByteSize, maxTime)
+	if err != nil {
+		return nil, fmt.Errorf("fail to parse the module args, arg=analytics.pubstack.buffers, :%v", err)
+	}
+
+	defaultFeatures := make(map[string]bool)
```
SyntaxNode, line 63: Style Option: You may also write this as a literal map initializer, if you prefer:
```
defaultFeatures := map[string]bool{
	auction: false,
	video: false,
	amp: false,
	cookieSync: false,
	serUID: false,
}
```

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,246 @@
...
+	size    int64
+}
+
+type PubstackModule struct {
+	eventChannels map[string]*eventchannel.EventChannel
+	configCh      chan *Configuration
+	scope         string
+	cfg           *Configuration
+	buffsCfg      *bufferConfig
+	mux           sync.Mutex
```
SyntaxNode, line 45: Is this mux protecting the cfg object? If so, consider changing this to to a RWMutex since the reads will greatly outweigh the writes.

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,246 @@
...
+
+	go pb.setup()
+	go pb.start(refreshDelay, endCh)
+
+	glog.Info("[pubstack] Pubstack analytics configured and ready")
+	return &pb, nil
+}
+
+func (p *PubstackModule) LogAuctionObject(ao *analytics.AuctionObject) {
+	p.mux.Lock()
```
SyntaxNode, line 94: Is this lock protecting the config via `isFeatureEnable` or the Push via `Push(payload)`.. or both? I'm asking because if it's just one or the either, we can restrict it's scope to minimize the lock time.

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,246 @@
...
+		p.eventChannels[video] = eventchannel.NewEventChannel(eventchannel.BuildEndpointSender(p.cfg.Endpoint, video), p.buffsCfg.size, p.buffsCfg.count, p.buffsCfg.timeout)
+	}
+	if p.isFeatureEnable(setUID) {
+		p.eventChannels[setUID] = eventchannel.NewEventChannel(eventchannel.BuildEndpointSender(p.cfg.Endpoint, setUID), p.buffsCfg.size, p.buffsCfg.count, p.buffsCfg.timeout)
+	}
+}
+
+func (p *PubstackModule) isFeatureEnable(feature string) bool {
+	val, ok := p.cfg.Features[feature]
+	if !(ok && val) {
```
SyntaxNode, line 242: Syntax Nitpick: You can avoid the branch with `return ok && val`

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+type bufferConfig struct {
+	timeout time.Duration
+	count   int64
+	size    int64
+}
+
+type PubstackModule struct {
+	eventChannels map[string]*eventchannel.EventChannel
```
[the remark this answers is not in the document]
gpolaert: Just to be sure, like that way?

```go
pubstackModule, err := pubstack.NewPubstackModule(
  clients.GetDefaultHttpInstance(),
  analytics.Pubstack.ScopeId,
  ...
```

analytics/config/config.go
```diff
@@ -17,6 +18,20 @@ func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
 			glog.Fatalf("Could not initialize FileLogger for file %v :%v", analytics.File.Filename, err)
 		}
 	}
+	if analytics.Pubstack.Enabled {
```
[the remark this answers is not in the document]
gpolaert: I've switched to an error log instead of a fatal. 
But the question is what do we want to do if we can't load a required module. Should we fail? If yes, I think we have to redesign this part to make it more testable.

analytics/pubstack/eventchannel/eventchannel.go
gpolaert: I've renamed the mutex, which will be more readable everywhere

analytics/pubstack/pubstack_module.go
gpolaert: I didn't know it, it's perfect. I also used it for the second mutex which it's the same pattern (N Readers- 1 Writer)

analytics/pubstack/pubstack_module.go
gpolaert: Both, because if the configuration is updated, the flag and the map(push func) could have a new value.
Now, I'm using the RWMutex to improve the lock

analytics/config/config.go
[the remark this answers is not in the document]
hhhjort: I think we should abort the startup if the analytics config cannot be processed.

gpolaert:
I did my best to improve the code coverage.
The last part non tested are mainly about checking return status of other libs (like url, json ...).

Let me know if it's okay

gpolaert:
Hello guys, 

any news? :)

SyntaxNode:
> Hello guys,
> 
> any news? :)

Sorry. Been a little busy. I commented on your other related PR. Looking at this tonight for feedback tomorrow morning.

analytics/pubstack/eventchannel/eventchannel.go
```diff
@@ -0,0 +1,138 @@
...
+func (c *EventChannel) Push(event []byte) {
+	c.ch <- event
+}
+
+func (c *EventChannel) Close() {
+	c.endCh <- 1
+}
+
+func (c *EventChannel) buffer(event []byte) {
+	c.muxGzBuffer.RLock()
```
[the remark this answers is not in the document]
gpolaert: You've right.
The RWMutex doesn't have any more sense here, so I revert to a simple Mutex.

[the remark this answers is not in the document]
gpolaert: Euh. Just see your next comment, so I will still use the RWMutex :)

analytics/pubstack/pubstack_module.go
```diff
@@ -0,0 +1,264 @@
...
+				if err != nil {
+					glog.Errorf("[pubstack] Fail to fetch remote configuration: %v", err)
+				}
+			}()
+		}
+	}
+
+}
+
+func (p *PubstackModule) updateConfig(config *Configuration) {
```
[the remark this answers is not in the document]
gpolaert: This could be great

But I try to respect the mantra "the best code is no code at all" :).

Because the update process is very infrequent (every 2h), I think the optimization will be very limited here. 
But because it's even more infrequent to have a configuration update, I can add a simple test to check if the config is new and if yes, update all channels

What do you think?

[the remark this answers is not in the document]
SyntaxNode: > I can add a simple test to check if the config is new and if yes, update all channels

Sounds good to me.

gpolaert:
Thanks gyus! (@hhhjort and @SyntaxNode)
Your reviews help me a lot to better understand the specific of this language :sweat_smile:

PR #1339, opened 2020-06-05
AdOcean adapter - Support for sizes defined in prebid configuration.

mmuras:
Support for sizes defined in Prebid.js configuration. Previously only sizes defined on SSP side were taken into account, the ones from Prebid was ignored.

### DIFF
--- a/adapters/adocean/adocean.go
+++ b/adapters/adocean/adocean.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"text/template"
@@ -19,7 +20,7 @@ import (
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
-const adapterVersion = "1.0.0"
+const adapterVersion = "1.1.0"
 const maxUriLength = 8000
 const measurementCode = `
 	<script>
@@ -49,6 +50,12 @@ type ResponseAdUnit struct {
 	WinURL   string `json:"winUrl"`
 	StatsURL string `json:"statsUrl"`
 	Error    string `json:"error"`
+}
+
+type requestData struct {
+	Url        string
+	Headers    *http.Header
+	SlaveSizes map[string]string
 }
 
 func NewAdOceanBidder(client *http.Client, endpointTemplateString string) *AdOceanAdapter {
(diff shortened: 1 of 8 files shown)


adapters/adocean/adocean.go
```diff
@@ -89,44 +96,58 @@ func (a *AdOceanAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adap
...
+			Headers: *requestData.Headers,
+		})
+	}
+
 	return httpRequests, errors
 }
 
-func (a *AdOceanAdapter) makeRequest(existingRequests []*adapters.RequestData, imp *openrtb.Imp, request *openrtb.BidRequest, consentString string) (*adapters.RequestData, error) {
+func (a *AdOceanAdapter) addNewBid(
+	requestsData *[]*requestData,
```
hhhjort, line 121: The usual go way of handling this is to return the `requestData` slice rather than passing a pointer in. I am not sure how go would handle the case where you would have to resize the slice larger, it would probably work, and should not be an issue with the current code. But I am not sure it will work as expected. Is there a reason you are not returning both `requestsData` and `error` instead of using this pointer magic?

adapters/adocean/adocean.go
```diff
@@ -89,44 +96,58 @@ func (a *AdOceanAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adap
...
 	if addedToExistingRequest {
-		return nil, nil
+		return nil
 	}
 
-	url, err := a.makeURL(&adOceanExt, imp.ID, request, consentString)
+	slaveSizes := map[string]string{}
+	slaveSizes[adOceanExt.SlaveID] = getImpSizes(imp)
+
+	url, err := a.makeURL(&adOceanExt, imp, request, slaveSizes, consentString)
```
hhhjort, line 148: Here you create a new `slaveSizes` map, add only one entry, and then pass that along. Why are you creating a map with one entry, rather than just passing the sizes, or potentially the key and the sizes?

mmuras: In this place request is created initially, but it can be changed to contain more slaves as loop will continue.
It is used later to check if master has a slave with given ID (on line 189) because this IDs have to be unique within one master. Also this map can have more elements if user configured different slave (if so it will be added at line 194). And at the end, after adding all placement, setSlaveSizesParam function will convert entire map into single GET parameter in this form: aosspsizes=myaowafpdwlrks~300x250-myaozpniqismex~300x250_320x600

adapters/adocean/adocean.go
mmuras: Well, It just seemed more logical to just pass a pointer to a slice if a function is going to change the slice size, instead of returning it, but I have no strong opinion on this matter and can change it so the slice will be return if you think that, it's better.

hhhjort: It seems like your method should work, but every bit of go code I have ever seen that does something like that uses the return slice method. The go append builtin uses the return slice, which is as close to an official recommendation as you can get. So it should be safer to move to that method.

adapters/adocean/adocean.go
hhhjort: In line 145 you declare a new `slaveSizes` map, line 146 is where you add an entry. I didn't spot any place else where you add an entry, so next time through the loop line 145 will declare a new map. I don't see how you will get a second entry unless I am missing something.

mmuras: At a line 140 there is a function "addToExistingRequest" so if it is possible (the same master, different slave) a new slave will be added to map created in previous iteration.
It working like that:
- can this slave be added to existing request? If yes add it to existing one and end at the line 142 (new entry will be added at line 194, inside addToExistingRequest function)
- if previous condition is not true, a new request with new map will be created

adapters/adocean/adocean.go
mmuras: It's changed.

adapters/adocean/adocean.go
hhhjort: That still doesn't change that once you hit line 149, it is only possible for one entry to exist in `slaveSizes`, you will have returned out of the function already if you added to an existing map. I don't see anywhere else where `a.makeURL` is called to require that a map be sent. This doesn't make the code wrong, but does worry me that you might have been trying to do something that will not actually work as intended.

mmuras: Maybe I am missing something but it works like that:
- on lines 102-107 you have a loop which is invoking addNewBid function. This function is creating a new entry in requestsData slice **or** (that's important part) add additional data (for example slave sizes) to existing entry.

for example we have auction for:
-- master: masterA slave: slaveAA
-- master: masterB slave: slaveBB
-- master: masterA slave: slaveAA
-- master: masterA slave: slaveAB

- and now we in fist iteration and first execution of addNewBid function, because there is no previous entries it obviously will create a new requestData struct which will contain url, headers and slaveSizez with one entry.
- now we in second iteration, because it's different master new request will be created as well, so we have a second struct with url, headers and map with one entry
- third iteration situation is repeating because we the same slave which was used in first iteration, so new struct will be created, so now we have 3 structs where each will have map slaveSizez with one entry
- but now we in fourth iteration: addNewBid function is invoked once more, we go instruction by instruction and we at the line 141, addToExistingRequest is invoked, now this function see that there is place in first request for current slave. Existing url is decoded into components and at the lines 194-196 additional ones are added. Now inside our SlaveSizes map we have 2 entries, after option amending this components are converted again to url. 
After addToExistingRequest returns with true, addNewBid also is returning, it does not create a new Map. So in the end we have 3 maps for 4 requests (in this case). 
- there may be question why url maybe converted to string more than once, but that is necessary to know length it.

hhhjort: Yes, I understand that there is a case where you can have multiple entries in the map. I was just focusing on where you invoke `a.makeURL()`, which seems to only happen on line 149. The only way to get to line 149 insures the `slaveSizes` used will only have one entry. My worry is that you intended to have additional paths to line 149, or other places in your code where you call `a.makeURL`. If this is not the case, then the code should work, though the signature for calling `a.makeURL` is a little more complex than it needs to be.

mmuras: OK, now I understand. So in this case the only reason makeUrl is receiving a map instead of just sizes of this slave (SlaveID is in "params" parameter so it doesn't have to be passed) is because it is invoking setSlaveSizesParam function which is expecting that map (and setSlaveSizesParam is invoked in place where it can receive a map with more than one entry). Of course I would create a second map but I'm not sure if it is worth that.

hhhjort: That is fine, I already approved this PR. I just wanted to be sure all the parts are working as intended, since there seemed to be a mismatch between what the code could do, and what it was actually doing.

adapters/adocean/adocean.go
```diff
@@ -147,43 +169,52 @@ func (a *AdOceanAdapter) makeRequest(existingRequests []*adapters.RequestData, i
...
-func addToExistingRequest(existingRequests []*adapters.RequestData, newParams *openrtb_ext.ExtImpAdOcean, auctionID string) bool {
-requestsLoop:
-	for _, request := range existingRequests {
-		endpointURL, _ := url.Parse(request.Uri)
-		queryParams := endpointURL.Query()
+func addToExistingRequest(requestsData []*requestData, newParams *openrtb_ext.ExtImpAdOcean, imp *openrtb.Imp, testImp bool) bool {
+	auctionID := imp.ID
+
+	for _, requestData := range requestsData {
+		endpointUrl, _ := url.Parse(requestData.Url)
```
SyntaxNode, line 185: The `makeURL` method is building a native Url object and converting it into a string. The first thing we're doing here is parsing it from a string to a Url object. Can you change `makeURL` to return a Url object and avoid this rework here?

mmuras: It's changed.

mmuras:
@bsardo ?

PR #1343, opened 2020-06-08
Add Outgoing Connection Metrics

guscarreon:
Add metrics to track the number of connections per host and differentiate between an actively used connection vs an idle connection.

Technical Reference:
https://medium.com/@deeeet/trancing-http-request-latency-in-golang-65b2463f548c
https://blog.golang.org/http-tracing


pbsmetrics/go_metrics.go
```diff
@@ -70,6 +70,11 @@ type AdapterMetrics struct {
 	BidsReceivedMeter metrics.Meter
 	PanicMeter        metrics.Meter
 	MarkupMetrics     map[openrtb_ext.BidType]*MarkupDeliveryMetrics
+	ConnSuccess       metrics.Counter
```
SyntaxNode, line 73: As you suspect, we may not need all of this information. Here's my take: 
- `ConnSuccess` is the default assumption. I don't think we need to log that counter. This also looks like the sum of `ConnCreated` and `ConnReused` which can be computed easily in queries if desired.
- `ConnError ` is likely already surfaced with other metrics. I don't remember exactly. I think it either turns into a panic or an adapter error. Thoughts on if this is worth a counter?
- `ConnCreated` seems very useful in tracking connection management.
- `ConnReused ` seems very useful in tracking connection management.

There's also two metrics I'd like to throw out for consideration:
 - Connection wait time. This one is a bit tricky and might not easily be done with the built-in tracers. The idea is to measure the time when the connection request is first made until a connection is either created or reused. This will tell us how bad the congestion in the connection pool is and how much delay is being introduced. This would be expensive, for sure... likely 12 additional metrics (at least for Prometheus) per adapter... which is now roughly 840 in total. Still, I think it would be super useful information.
- Dns resolution wait time. I don't think this is worthwhile to record per adapter, but might be interesting in aggregate for hosts to understand how their DNS resolution layer is performing.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -178,6 +185,11 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
 		"Count of requests labeled by adapter, if has a cookie, and if it resulted in bids.",
 		[]string{adapterLabel, cookieLabel, hasBidsLabel})
 
+	metrics.adapterConnections = newCounter(cfg, metrics.Registry,
+		"adapter_connections",
+		"Count of connections to bidder hosts labeled by their adapters. Tracks if connection was successful, connection was reused, or if connection was left idle",
```
SyntaxNode, line 190: These should be separated out into their own metrics. I don't think the correlation serves meaningful purposes here. Some of the labels are mutually exclusive (success vs error, reused vs created) and knowing the relation between reused vs created and idle doesn't seem very interesting.

pbsmetrics/prometheus/prometheus_test.go
```diff
@@ -61,7 +61,7 @@ func TestMetricCountGatekeeping(t *testing.T) {
 	// Verify Per-Adapter Cardinality
 	// - This assertion provides a warning for newly added adapter metrics. Threre are 40+ adapters which makes the
 	//   cost of new per-adapter metrics rather expensive. Thought should be given when adding new per-adapter metrics.
-	assert.True(t, perAdapterCardinalityCount <= 22, "Per-Adapter Cardinality")
+	assert.True(t, perAdapterCardinalityCount <= 54, "Per-Adapter Cardinality")
```
SyntaxNode, line 64: Yeah, see? With having these as labels on the same adapter the cardinality is very high for the information we're recording and we're adding 32 new metrics instead of just 5 per adapter.

pbsmetrics/go_metrics.go
mansinahar: - I agree with @SyntaxNode. We only need to know if a new connection was created for a bidder vs if an existing connection was reused. Either of those would mean that we were able to connect with the bidder successfully. Note that we already have a metric called, `adapter_errors` which should record any type of adapter related error including not being able to connect to the adapter.

- I think adding connection wait time metric is worth it because we wanna make sure that our suspicion that creating new connections is the cause of increased latencies during spikey traffic, is indeed true. I believe you should be able to do that by recording the timestamp in `GetConn()` and `GotConn()` and then subtracting the two. This will represent the entire time taken to create a new connection (including DNS lookup and TLS handshake time). Although, when we record this, we wanna make sure that we're recording it only when we're creating a new connection and not when we are reusing an existing one.

- One other thing to consider is redirects. In case of redirects, these trace functions may be called multiple times which will then in turn record these metrics multiple time for each redirect connection as well. This could lead to misleading numbers in our metrics. This one is a little tricky and will require a little more thought as to how we can handle this.

guscarreon:
Below is the `diff` command output in prometheus metrics after a post request to bidder `appnexus`

**A new connection and connection wait time**
```
168c168
< prebid_server_adapter_connection_created{adapter="appnexus"} 0
---
> prebid_server_adapter_connection_created{adapter="appnexus"} 1
375a376,390
> # HELP prebid_server_adapter_connection_wait Seconds from when the connection was requested until it is either created or reused
> # TYPE prebid_server_adapter_connection_wait histogram
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.05"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.1"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.15"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.2"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.25"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.3"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.4"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.5"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.75"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="1"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="+Inf"} 1
> prebid_server_adapter_connection_wait_sum{adapter="appnexus"} 0.03102841
> prebid_server_adapter_connection_wait_count{adapter="appnexus"} 1
602c617
```

**DNS lookup metrics**
```
< prebid_server_connections_closed 0
---
> prebid_server_connections_closed 1
4562c4577
< prebid_server_connections_opened 0
---
> prebid_server_connections_opened 1
4568,4580c4583,4595
< prebid_server_dns_lookup_time_bucket{le="0.05"} 0
< prebid_server_dns_lookup_time_bucket{le="0.1"} 0
< prebid_server_dns_lookup_time_bucket{le="0.15"} 0
< prebid_server_dns_lookup_time_bucket{le="0.2"} 0
< prebid_server_dns_lookup_time_bucket{le="0.25"} 0
< prebid_server_dns_lookup_time_bucket{le="0.3"} 0
< prebid_server_dns_lookup_time_bucket{le="0.4"} 0
< prebid_server_dns_lookup_time_bucket{le="0.5"} 0
< prebid_server_dns_lookup_time_bucket{le="0.75"} 0
< prebid_server_dns_lookup_time_bucket{le="1"} 0
< prebid_server_dns_lookup_time_bucket{le="+Inf"} 0
< prebid_server_dns_lookup_time_sum 0
< prebid_server_dns_lookup_time_count 0
---
> prebid_server_dns_lookup_time_bucket{le="0.05"} 1
> prebid_server_dns_lookup_time_bucket{le="0.1"} 1
> prebid_server_dns_lookup_time_bucket{le="0.15"} 1
> prebid_server_dns_lookup_time_bucket{le="0.2"} 1
> prebid_server_dns_lookup_time_bucket{le="0.25"} 1
> prebid_server_dns_lookup_time_bucket{le="0.3"} 1
> prebid_server_dns_lookup_time_bucket{le="0.4"} 1
> prebid_server_dns_lookup_time_bucket{le="0.5"} 1
> prebid_server_dns_lookup_time_bucket{le="0.75"} 1
> prebid_server_dns_lookup_time_bucket{le="1"} 1
> prebid_server_dns_lookup_time_bucket{le="+Inf"} 1
> prebid_server_dns_lookup_time_sum 0.022126272
> prebid_server_dns_lookup_time_count 1
4587c4602
```

pbsmetrics/go_metrics_test.go
```diff
@@ -1,14 +1,18 @@
...
 	"testing"
+	"time"
 
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	metrics "github.com/rcrowley/go-metrics"
 	"github.com/stretchr/testify/assert"
 )
 
+const OneSecond time.Duration = time.Second
```
mansinahar, line 14: `time.Second` is already a constant, any reason why we wanna define it as a constant again over here with a diff name?

pbsmetrics/metrics_mock.go
```diff
@@ -52,6 +53,15 @@ func (me *MetricsEngineMock) RecordAdapterRequest(labels AdapterLabels) {
...
 }
 
+// RecordAdapterConnections mock
+func (me *MetricsEngineMock) RecordAdapterConnections(bidderName openrtb_ext.BidderName, connSuccess bool, info httptrace.GotConnInfo, obtainConnectionTime time.Duration) {
+	me.Called(bidderName, connSuccess, info, obtainConnectionTime)
+}
+
+// RecordDNSTime mock
+func (me *MetricsEngineMock) RecordDNSTime(dnsLookupTime time.Duration) {
+}
```
mansinahar, line 63: Why don't we call the `Called` function for this method? Looks like something you might have missed by mistake?

pbsmetrics/prometheus/prometheus.go
```diff
@@ -63,6 +70,10 @@ const (
 	requestStatusLabel   = "request_status"
 	requestTypeLabel     = "request_type"
 	successLabel         = "success"
+
+	adapterConnErrorLabel   = "adapter_connection_error"
```
mansinahar, line 74: Since you've defined separate metrics for `ConnError`, `ConnReused` and `ConnCreated`, any reason why we also need labels? Also, they don't seem to be used anywhere.

exchange/bidder.go
```diff
@@ -321,6 +324,7 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 	}
 	httpReq.Header = req.Headers
 
+	ctx = bidder.AddClientTrace(ctx)
```
SyntaxNode, line 327: Should we add a feature flag around this option? Just in case it doesn't work as expected for some reason in prod or if a host would like to disable the feature to save a little bit in memory usage. If so, the prometheus preload should also only preload the metrics if the feature flag is enabled.

pbsmetrics/go_metrics.go
```diff
@@ -234,6 +244,11 @@ func makeBlankAdapterMetrics() *AdapterMetrics {
 		BidsReceivedMeter: blankMeter,
 		PanicMeter:        blankMeter,
 		MarkupMetrics:     makeBlankBidMarkupMetrics(),
+		ConnError:         metrics.NilCounter{},
+		ConnCreated:       metrics.NilCounter{},
+		ConnReused:        metrics.NilCounter{},
+		ConnIdleTime:      &metrics.NilTimer{},
```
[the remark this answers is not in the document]
guscarreon: I thought it'd be useful to know which connections are more efficient. I'm not sure if I'm going for a "nice to have" here instead of a "must have" or we "definitely need" this. Do think we should discard them?

pbsmetrics/go_metrics.go
```diff
@@ -234,6 +244,11 @@ func makeBlankAdapterMetrics() *AdapterMetrics {
 		BidsReceivedMeter: blankMeter,
 		PanicMeter:        blankMeter,
 		MarkupMetrics:     makeBlankBidMarkupMetrics(),
+		ConnError:         metrics.NilCounter{},
```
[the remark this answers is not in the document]
guscarreon: Pretty good catch! The error metrics were not getting updated at all. I can remove them if you want, for the moment I left them there because I think it'd be interesting to know whether or not an adapter endpoint is having connection issues. But I do realize this is probably just a "nice to have" and not a "we definitely need this". We can discuss this in standup if you want and I can removed them if you want me to.

exchange/bidder.go
```diff
@@ -412,3 +416,34 @@ type httpCallInfo struct {
 	response *adapters.ResponseData
 	err      error
 }
+
+// This function adds an httptrace.ClientTrace object to the context so, if connection with the bidder
+// endpoint is established, we can keep track of whether the connection was newly created, reused, and
+// even know if it was left idle and for how much time.
+func (bidder *bidderAdapter) AddClientTrace(ctx context.Context) context.Context {
```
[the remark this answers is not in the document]
guscarreon: Good point. Let me go back to the whiteboard. Maybe we can modify the context in `exchange/exchange.go`.

pbsmetrics/metrics_mock.go
guscarreon: Not necessary for `TestRecordDNSTime`

pbsmetrics/go_metrics_test.go
guscarreon: Good point. Corrected

pbsmetrics/prometheus/prometheus.go
guscarreon: Pretty good catch. This is a leftover from the previous commit. Removed

exchange/bidder.go
guscarreon: Added. Let me know your thoughts with the addition

guscarreon:
Modified according to feedback. Kept both the error connection metrics and the connection idle time. Added a configuration boolean flag to disable the adapter connection metrics. 

Ran the server locally and posted a request to check out if the metrics work as expected. These are the metrics before and after said request to the appnexus bidder.

```
170c170
< prebid_server_adapter_connection_created{adapter="appnexus"} 0
---
> prebid_server_adapter_connection_created{adapter="appnexus"} 1
578,590c578,590
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.05"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.1"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.15"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.2"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.25"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.3"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.4"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.5"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.75"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="1"} 0
< prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="+Inf"} 0
< prebid_server_adapter_connection_wait_sum{adapter="appnexus"} 0
< prebid_server_adapter_connection_wait_count{adapter="appnexus"} 0
---
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.05"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.1"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.15"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.2"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.25"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.3"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.4"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.5"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="0.75"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="1"} 1
> prebid_server_adapter_connection_wait_bucket{adapter="appnexus",le="+Inf"} 1
> prebid_server_adapter_connection_wait_sum{adapter="appnexus"} 0.036550788
> prebid_server_adapter_connection_wait_count{adapter="appnexus"} 1
    .
    .
    .
5581c5581
< prebid_server_connections_closed 0
---
> prebid_server_connections_closed 1
5588c5588
< prebid_server_connections_opened 0
---
> prebid_server_connections_opened 1
5594,5606c5594,5606
< prebid_server_dns_lookup_time_bucket{le="0.05"} 0
< prebid_server_dns_lookup_time_bucket{le="0.1"} 0
< prebid_server_dns_lookup_time_bucket{le="0.15"} 0
< prebid_server_dns_lookup_time_bucket{le="0.2"} 0
< prebid_server_dns_lookup_time_bucket{le="0.25"} 0
< prebid_server_dns_lookup_time_bucket{le="0.3"} 0
< prebid_server_dns_lookup_time_bucket{le="0.4"} 0
< prebid_server_dns_lookup_time_bucket{le="0.5"} 0
< prebid_server_dns_lookup_time_bucket{le="0.75"} 0
< prebid_server_dns_lookup_time_bucket{le="1"} 0
< prebid_server_dns_lookup_time_bucket{le="+Inf"} 0
< prebid_server_dns_lookup_time_sum 0
< prebid_server_dns_lookup_time_count 0
---
> prebid_server_dns_lookup_time_bucket{le="0.05"} 1
> prebid_server_dns_lookup_time_bucket{le="0.1"} 1
> prebid_server_dns_lookup_time_bucket{le="0.15"} 1
> prebid_server_dns_lookup_time_bucket{le="0.2"} 1
> prebid_server_dns_lookup_time_bucket{le="0.25"} 1
> prebid_server_dns_lookup_time_bucket{le="0.3"} 1
> prebid_server_dns_lookup_time_bucket{le="0.4"} 1
> prebid_server_dns_lookup_time_bucket{le="0.5"} 1
> prebid_server_dns_lookup_time_bucket{le="0.75"} 1
> prebid_server_dns_lookup_time_bucket{le="1"} 1
> prebid_server_dns_lookup_time_bucket{le="+Inf"} 1
> prebid_server_dns_lookup_time_sum 0.022831942
> prebid_server_dns_lookup_time_count 1
```

exchange/bidder.go
```diff
@@ -87,20 +88,27 @@ type pbsOrtbSeatBid struct {
...
+	BidderName  openrtb_ext.BidderName
 	Client      *http.Client
 	DebugConfig config.Debug
-	me          pbsmetrics.MetricsEngine
+	me          bidderMetrics
+}
+
+type bidderMetrics struct {
+	engine   pbsmetrics.MetricsEngine
+	disabled config.DisabledMetrics
```
SyntaxNode, line 111: I would normally expect a property called `disabled` to be a bool. `disabledMetrics` perhaps?

exchange/bidder.go
```diff
@@ -334,6 +347,7 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 			}
 
 		}
+		bidder.me.engine.RecordAdapterFailedConnections(bidder.BidderName)
```
SyntaxNode, line 350: I'm curious about this. A connection may fail due to our context deadline, which I don't think should be reported as a connection error. Are there other ways for `ctxhttp.Do` to error which has nothing to do with retrieving a connection?

pbsmetrics/go_metrics.go
```diff
@@ -269,6 +286,11 @@ func registerAdapterMetrics(registry metrics.Registry, adapterOrAccount string,
 		openrtb_ext.BidTypeAudio:  makeDeliveryMetrics(registry, adapterOrAccount+"."+exchange, openrtb_ext.BidTypeAudio),
 		openrtb_ext.BidTypeNative: makeDeliveryMetrics(registry, adapterOrAccount+"."+exchange, openrtb_ext.BidTypeNative),
 	}
+	am.ConnError = metrics.GetOrRegisterCounter(fmt.Sprintf("%[1]s.%[2]s.connections_error", adapterOrAccount, exchange), registry)
+	am.ConnCreated = metrics.GetOrRegisterCounter(fmt.Sprintf("%[1]s.%[2]s.connections_created", adapterOrAccount, exchange), registry)
+	am.ConnReused = metrics.GetOrRegisterCounter(fmt.Sprintf("%[1]s.%[2]s.connections_reused", adapterOrAccount, exchange), registry)
+	am.ConnIdleTime = metrics.GetOrRegisterTimer(fmt.Sprintf("%[1]s.%[2]s.connection_idle_time", adapterOrAccount, exchange), registry)
```
SyntaxNode, line 292: I think this was asked before, but I forgot the answer. Per-adapter metrics are very expensive, especially for timers. Adding a timer for Prometheus results in nearly 1,000 new unique metrics. We'll use `ConnWaitTime` to tweak connection pools and so I think it earns it worth. What will we use `ConnIdleTime` for?

pbsmetrics/metrics_mock.go
SyntaxNode: I still recommend to add the calls, so future tests can rely on that behavior if needed in the future.

config/config.go
```diff
@@ -675,6 +680,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("http_client_cache.idle_connection_timeout_seconds", 60)
 	// no metrics configured by default (metrics{host|database|username|password})
 	v.SetDefault("metrics.disabled_metrics.account_adapter_details", false)
+	v.SetDefault("metrics.disabled_metrics.adapter_connections_metrics", false)
```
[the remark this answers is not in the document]
guscarreon: Makes sense, set to `true` so they are disabled by default.

exchange/bidder.go
guscarreon: Makes total sense. Corrected.

exchange/bidder.go
guscarreon: True. This metric will be removed per our conversation

pbsmetrics/go_metrics.go
guscarreon: Per our conversation, `ConnIdleTime` was removed

exchange/bidder.go
```diff
@@ -412,3 +425,34 @@ type httpCallInfo struct {
 	response *adapters.ResponseData
 	err      error
 }
+
+// This function adds an httptrace.ClientTrace object to the context so, if connection with the bidder
+// endpoint is established, we can keep track of whether the connection was newly created, reused, and
+// even know if it was left idle and for how much time.
```
SyntaxNode, line 431: Might want to update the comment to remove the idle references.

exchange/bidder_test.go
```diff
@@ -1231,6 +1240,100 @@ func TestSetAssetTypes(t *testing.T) {
...
+			w.Write([]byte("getBody"))
+		} else {
+			w.Write([]byte("postBody"))
+		}
+	})
+	server := httptest.NewServer(handler)
+	defer server.Close()
+
+	//setup influx metrics that give us access to the DNSLookupTimer
+	cfg := config.Configuration{}
```
SyntaxNode, line 1310: Would it be simpler to use a metrics mock here instead of relying on the Influx implementation?

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20180720115003-f9ffefc3facf
 	github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973 // indirect
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869
```
SyntaxNode, line 13: What's this from?

exchange/bidder.go
guscarreon: Done

exchange/bidder_test.go
guscarreon: surprisingly, it was a little bit hard to make them work. Will keep looking into it tomorrow.

SyntaxNode: What was the trouble? You should need 1 expectation on the mock RecordDNSTime with a 0 duration, yeah?

go.mod
guscarreon: I've no idea. Removed

exchange/bidder_test.go
guscarreon: Yep. that 0 duration value was not testing anything. Except that `ClientTrace.DNSDone()` wasn't been called.

exchange/bidder_test.go
```diff
@@ -1231,6 +1241,93 @@ func TestSetAssetTypes(t *testing.T) {
...
+		Bidder: &mixedMultiBidder{},
+		Client: &http.Client{Transport: DNSDoneTripper{}},
+		me:     bidderMetrics{engine: metricsConfig.NewMockMetricsEngine(mockMetricsData)},
+	}
+
+	// Run test
+	bidder.doRequest(context.Background(), &adapters.RequestData{Method: "POST", Uri: "http://www.example.com/"})
+
+	// Assert DNS results, mock metrics should have recorded a 1.00 duration
+	assert.Equal(t, mockMetricsData.DNSLookupTimer, 1.00, "Function ClientTracer.DNSDone(info httptrace.DNSStartInfo) should have been called and seems it wasn't")
```
SyntaxNode, line 1328: Where does the `1.00` come from?

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/influxdata/influxdb v1.6.1 // indirect
 	github.com/julienschmidt/httprouter v1.1.0
 	github.com/k0kubun/colorstring v0.0.0-20150214042306-9440f1994b88 // indirect
+	github.com/kr/pretty v0.2.0 // indirect
```
SyntaxNode, line 25: I don't think you're using the changes in go.mod and go.sum.

pbsmetrics/config/metrics.go
```diff
@@ -284,3 +308,108 @@ func (me *DummyMetricsEngine) RecordTimeoutNotice(success bool) {
 // RecordReq as a noop
 func (me *DummyMetricsEngine) RecordTCFReq(version pbsmetrics.TCFVersionValue) {
 }
+
+// MockMetricsEngine will have fields to update via the MetricsEngine interface methods
+type MockMetricsData struct {
```
SyntaxNode, line 313: Please use the existing `MetricsEngineMock` instead of creating a new one.

exchange/bidder.go
```diff
@@ -87,20 +88,27 @@ type pbsOrtbSeatBid struct {
...
+	BidderName  openrtb_ext.BidderName
 	Client      *http.Client
 	DebugConfig config.Debug
-	me          pbsmetrics.MetricsEngine
+	me          bidderMetrics
+}
+
+type bidderMetrics struct {
+	engine          pbsmetrics.MetricsEngine
+	disabledMetrics config.DisabledMetrics
```
mansinahar, line 111: Do we need to have the `disabledMetrics` info at the bidder level? Right now the way it is configured, we can either enable connection metrics for all adapter or no adapters and so I am wondering if this is needed at the bidder level?

exchange/bidder.go
```diff
@@ -321,6 +329,11 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 	}
 	httpReq.Header = req.Headers
 
+	// If adapter connection metrics are not disabled, add the client trace
+	// to get complete connection info into our metrics
+	if !bidder.me.disabledMetrics.AccountAdapterDetails {
+		ctx = bidder.addClientTrace(ctx)
```
mansinahar, line 335: As I mentioned above, since the connection metrics can be enabled/disabled for all adapters, it might be easier to just add the http tracer earlier in the code path a i.e. somewhere before `requestBid` is called. That way you'd have to add the http tracer only once because the same context gets propagated downstream.

go.mod
mansinahar: I believe the `//indirect` comment indicates that dependency is used by one of the other modules that we depend on.
```
The indirect comment indicates a dependency is not used directly by this module, only indirectly by other module dependencies. See go help modules for details.
```

Ref: https://blog.golang.org/using-go-modules

exchange/bidder_test.go
guscarreon: The mock metrics were written to write a value. The value `1.00` seemed to be a good test value to write

go.mod
guscarreon: The vim-go plug-in added them automatically, so I assumed they were needed.

exchange/bidder.go
guscarreon: The idea, as understood in the discussions we had on this PR, is to be able to enable this metrics by adapter.

exchange/bidder.go
guscarreon: It was suggested in one the meetings we held. In the future we'll probably want to collect metrics only for certain adapters.

exchange/bidder.go
```diff
@@ -321,6 +329,11 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 	}
 	httpReq.Header = req.Headers
 
+	// If adapter connection metrics are not disabled, add the client trace
+	// to get complete connection info into our metrics
+	if !bidder.me.disabledMetrics.AccountAdapterDetails {
```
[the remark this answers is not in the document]
guscarreon: Wow :open_mouth: !! wrong boolean value. Corrected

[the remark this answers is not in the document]
guscarreon: Pretty good catch!

go.mod
SyntaxNode: Ah, ok. I missed the indirect comment.

pbsmetrics/config/metrics.go
guscarreon: :laughing: I just used `MetricsEngineMock` in `TestCallRecordAdapterConnections` above and completely forgot about it. Jesus, I'm sorry. Fixed and removed.

exchange/bidder_test.go
guscarreon: Test has been modified to use the `mock.Mock` metrics and is no longer asserting this way

exchange/bidder.go
```diff
@@ -87,20 +88,27 @@ type pbsOrtbSeatBid struct {
...
-	me          pbsmetrics.MetricsEngine
+	me          bidderMetrics
+}
+
+type bidderMetrics struct {
+	engine          pbsmetrics.MetricsEngine
+	disabledMetrics config.DisabledMetrics
 }
 
 func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.BidRequest, name openrtb_ext.BidderName, bidAdjustment float64, conversions currencies.Conversions, reqInfo *adapters.ExtraRequestInfo) (*pbsOrtbSeatBid, []error) {
```
mansinahar, line 114: Since you've added `BidderName` to the `bidderAdapter`, can we please remove the `name` parameter from this function args list as it is redundant?

guscarreon: `exchange/legacy.go`'s implementation of `requestBid` does use this parameter. Do you think it is better to keep it? I know it's unnecessary in every implementation of `requestBid` out there except for `exchange/legacy.go`, but for the moment, it is better to keep it to not break anything.

Another option is to eliminate `bidderAdapter`'s `BidderName` field and move `ctx = bidder.addClientTrace(ctx)` upwards from `doRequest` to `requestBid(ctx context.Context, request *openrtb.BidRequest, bidAdjustment float64, conversions currencies.Conversions, reqInfo *adapters.ExtraRequestInfo)` in order to make use of its `name` argument. Though I bet it is better not to rely on `name` given that in the future `legacy.go` will eventually be dropped and we want less arguments in function signatures. What do you think?

exchange/bidder.go
guscarreon: You mean consolidating the `DebugConfig` and the `disabledMetrics` fields? 

So go from this: (current version)
```
type bidderAdapter struct {
    Bidder      adapters.Bidder
    Client      *http.Client
    DebugConfig config.Debug
    me          pbsmetrics.MetricsEngine
}
```

To this: 
```
type bidderAdapter struct {
    Bidder      adapters.Bidder
    Client      *http.Client
    me          pbsmetrics.MetricsEngine
    config      bidderAdapterConfig
}
type bidderAdapterConfig struct {
    DebugConfig   config.Debug
    MetricsConfig config.DisabledMetrics
}
```

And not this: (proposed in PR)
```
type bidderAdapter struct {
    Bidder      adapters.Bidder
    BidderName  openrtb_ext.BidderName
    Client      *http.Client
    DebugConfig config.Debug
    me          bidderMetrics
}

type bidderMetrics struct {
    engine          pbsmetrics.MetricsEngine
    disabledMetrics config.DisabledMetrics
}
```

Did I get it right?

mansinahar: Yup! That's right. Although, I think rather than having the entire `DisabledMetrics` struct in `bidderAdapterConfig`, it might be better to have just a boolean for connection metrics that indicates if connection metrics for this bidder is enabled or not (Right now, it'd be true for all bidders). Because we aren't using the `AccountAdapterDetails` value anywhere here and also if we add more metrics to the `DisabledMetrics` struct in the future, not each of them would make sense at the bidder level and so we might wanna keep only the ones we're interested in over here rather than copying the entire `DisabledMetrics` struct.

guscarreon: Done

exchange/bidder_test.go
```diff
@@ -1233,6 +1243,94 @@ func TestSetAssetTypes(t *testing.T) {
...
+	}
+
+	bidderImpl := &goodSingleBidder{
+		httpRequest: &adapters.RequestData{
+			Method:  "POST",
+			Uri:     server.URL,
+			Body:    []byte("{\"key\":\"val\"}"),
+			Headers: http.Header{},
+		},
+		bidResponse: mockBidderResponse,
```
mansinahar, line 1274: We can just have the `bidResponse` here to be an empty `adapters.BidResponse` struct as we don't really care about the bid response in this case. Or the other thing you can do is extract this single bid response to a global variable in this test file because I am sure this same response is used elsewhere in this test file too. This will reduce the redundant code.

exchange/bidder_test.go
```diff
@@ -1233,6 +1243,94 @@ func TestSetAssetTypes(t *testing.T) {
...
+	// http.Client that runs our mock RoundTripper so DNSDone(httptrace.DNSDoneInfo{})
+	// gets called
+	bidder := &bidderAdapter{
+		Bidder: &mixedMultiBidder{},
+		Client: &http.Client{Transport: DNSDoneTripper{}},
+		me:     metricsMock,
+	}
+
+	// Run test
+	bidder.doRequest(context.Background(), &adapters.RequestData{Method: "POST", Uri: "http://www.example.com/"})
```
mansinahar, line 1328: Can't we have the above test testing `RecordAdapterConnections` also just go through `doRequest` rather than `requestBid` as you're introducing the change in the `doRequestImpl` function and so testing on that level for unit tests should be enough, no?

pbsmetrics/go_metrics_test.go
```diff
@@ -179,6 +184,160 @@ func TestRecordBidTypeDisabledConfig(t *testing.T) {
...
+		expDuration time.Duration
+	}
+	testCases := []struct {
+		description string
+		in          testIn
+		out         testOut
+	}{
+		{
+			description: "Five second DNS lookup time",
+			in: testIn{
```
mansinahar, line 201: What's the reason behind wrapping the `dnsLookupDuration` and `expDuration` in `testIn` and `testOut` structs?

pbsmetrics/go_metrics_test.go
```diff
@@ -179,6 +184,160 @@ func TestRecordBidTypeDisabledConfig(t *testing.T) {
...
+			out:         testOut{},
+		},
+	}
+	for i, test := range testCases {
+		registry := metrics.NewRegistry()
+		m := NewMetrics(registry, []openrtb_ext.BidderName{openrtb_ext.BidderAppnexus}, config.DisabledMetrics{AccountAdapterDetails: true})
+
+		m.RecordDNSTime(test.in.dnsLookupDuration)
+
+		assert.Equal(t, test.out.expDuration.Nanoseconds(), m.DNSLookupTimer.Sum(), "Test [%d] incorrect DNS lookup time", i)
```
mansinahar, line 220: Might be good to add the test `description` as part of the error message here rather than the test case index so that the reader doesn't necessarily have to come back here to map the index to the test case

config/config.go
```diff
@@ -379,6 +379,11 @@ type Metrics struct {
 type DisabledMetrics struct {
 	// True if we want to stop collecting account-to-adapter metrics
 	AccountAdapterDetails bool `mapstructure:"account_adapter_details"`
+
+	// True if we don't want to collect metrics about the connections prebid
+	// server establishes with bidder servers such as the number of connections
+	// that were created or reused and the connection idle and waiting times.
```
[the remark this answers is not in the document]
guscarreon: Done

exchange/bidder.go
```diff
@@ -87,20 +88,30 @@ type pbsOrtbSeatBid struct {
...
-		Bidder:      bidder,
-		Client:      client,
-		DebugConfig: cfg.Debug,
-		me:          me,
+		Bidder:     bidder,
+		BidderName: name,
+		Client:     client,
+		me:         me,
+		config: bidderAdapterConfig{
+			DebugConfig:          cfg.Debug,
```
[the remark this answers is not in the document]
guscarreon: Done

exchange/bidder.go
```diff
@@ -87,20 +88,30 @@ type pbsOrtbSeatBid struct {
...
+			LogConnectionMetrics: cfg.Metrics.Disabled.AdapterConnectionMetrics,
```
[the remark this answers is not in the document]
guscarreon: Done

pbsmetrics/config/metrics.go
```diff
@@ -237,6 +253,14 @@ func (me *DummyMetricsEngine) RecordAdapterPanic(labels pbsmetrics.AdapterLabels
 func (me *DummyMetricsEngine) RecordAdapterRequest(labels pbsmetrics.AdapterLabels) {
 }
 
+// RecordAdapterConnections as a noop
+func (me *DummyMetricsEngine) RecordAdapterConnections(bidderName openrtb_ext.BidderName, info httptrace.GotConnInfo, obtainConnectionTime time.Duration) {
+}
+
+// Times the DNS resolution process
```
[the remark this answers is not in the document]
guscarreon: Done

pbsmetrics/go_metrics.go
```diff
@@ -268,6 +275,11 @@ func makeBlankAdapterMetrics() *AdapterMetrics {
 		PanicMeter:        blankMeter,
 		MarkupMetrics:     makeBlankBidMarkupMetrics(),
 	}
+	if disabledMetrics.AdapterConnectionMetrics {
```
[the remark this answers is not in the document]
guscarreon: Totally right. That's a bug

pbsmetrics/go_metrics.go
```diff
@@ -472,6 +487,34 @@ func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
...
+	info httptrace.GotConnInfo,
+	obtainConnectionTime time.Duration) {
+
+	if me.MetricsDisabled.AdapterConnectionMetrics {
+		return
+	}
+
+	am, ok := me.AdapterMetrics[adapterName]
+	if !ok {
+		glog.Errorf("Trying to log adapter connection metrics for %s: adapter metrics not found", string(adapterName))
```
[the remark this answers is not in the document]
guscarreon: Ok

pbsmetrics/go_metrics.go
```diff
@@ -472,6 +487,34 @@ func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
 	}
 }
 
+// Keeps track of created and reused connections to adapter bidders and the time from the
+// connection request, to the connection creation, or reuse from the pool across all engines
+func (me *Metrics) RecordAdapterConnections(adapterName openrtb_ext.BidderName,
+	info httptrace.GotConnInfo,
+	obtainConnectionTime time.Duration) {
```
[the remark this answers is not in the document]
guscarreon: Done

pbsmetrics/go_metrics.go
```diff
@@ -472,6 +487,34 @@ func (me *Metrics) RecordAdapterRequest(labels AdapterLabels) {
 	}
 }
 
+// Keeps track of created and reused connections to adapter bidders and the time from the
+// connection request, to the connection creation, or reuse from the pool across all engines
+func (me *Metrics) RecordAdapterConnections(adapterName openrtb_ext.BidderName,
+	info httptrace.GotConnInfo,
```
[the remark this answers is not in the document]
guscarreon: Pretty good idea. Modified

pbsmetrics/prometheus/preload.go
```diff
@@ -85,6 +85,20 @@ func preloadLabelValues(m *Metrics) {
...
+			adapterLabel: adapterValues,
+		})
+
+		preloadLabelValuesForHistogram(m.adapterConnectionWaitTime, map[string][]string{
+			adapterLabel: adapterValues,
+		})
+	}
+
 	preloadLabelValuesForHistogram(m.adapterRequestsTimer, map[string][]string{
 		adapterLabel: adapterValues,
```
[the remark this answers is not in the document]
guscarreon: Only vectors need to be preloaded. In other words a `prometheus.HistogramVec` needs preloading whereas a `prometheus.Histogram` does not.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -29,23 +30,29 @@ type Metrics struct {
 	storedImpressionsCacheResult *prometheus.CounterVec
 	storedRequestCacheResult     *prometheus.CounterVec
 	timeoutNotifications         *prometheus.CounterVec
+	dnsLookupTime                prometheus.Histogram
```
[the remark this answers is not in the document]
guscarreon: Other metrics need labeling. Take `requestsQueueTimer`: The reason it was defined as a `*prometheus.HistogramVec` over a simple `Histogram` is that we wan't to differentiate the data by whatever values the `labels pbsmetrics.Labels` in its `RType` and `RequestStatus` fields:
```
func (m *Metrics) RecordRequest(labels pbsmetrics.Labels) {
	m.requests.With(prometheus.Labels{
		requestTypeLabel:   string(labels.RType),
		requestStatusLabel: string(labels.RequestStatus),
	}).Inc()
  .
  .
  .
```

pbsmetrics/prometheus/prometheus.go
```diff
@@ -220,7 +233,25 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
...
+		metrics.adapterReusedConnections = newCounter(cfg, metrics.Registry,
+			"adapter_connection_reused",
+			"Count that keeps track of reused connections when contacting adapter bidder endpoints.",
+			[]string{adapterLabel})
+
+		metrics.adapterConnectionWaitTime = newHistogramVec(cfg, metrics.Registry,
+			"adapter_connection_wait",
+			"Seconds from when the connection was requested until it is either created or reused",
+			[]string{adapterLabel},
+			requestTimeBuckets)
```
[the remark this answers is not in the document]
guscarreon: Renamed to `standardTimeBuckets`

exchange/bidder_test.go
```diff
@@ -1233,6 +1243,94 @@ func TestSetAssetTypes(t *testing.T) {
...
+			Body:    []byte("{\"key\":\"val\"}"),
+			Headers: http.Header{},
+		},
+		bidResponse: mockBidderResponse,
+	}
+
+	// Setup mock metrics
+	metrics := &pbsmetrics.MetricsEngineMock{}
+	expectedAdapterName := openrtb_ext.BidderAppnexus
+	compareConnInfo := func(info httptrace.GotConnInfo) bool { return !info.Reused }
```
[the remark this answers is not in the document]
guscarreon: After your suggestion above about modifying the second argument of this function, `compareConnInfo` is no longer needed and was substituted by a boolean value.

exchange/bidder_test.go
```diff
@@ -1233,6 +1243,94 @@ func TestSetAssetTypes(t *testing.T) {
 	}
 }
 
+func TestCallRecordAdapterConnections(t *testing.T) {
+	// Setup mock server
+	respStatus := 200
+	respBody := "{\"bid\":false}"
+	server := httptest.NewServer(mockHandler(respStatus, "getBody", respBody))
```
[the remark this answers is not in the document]
guscarreon: Many other tests do the exact same. I'm leaving it for consistency. A future PR can probably clean, refactor and make the code a bit easier to read in the entire file.

exchange/bidder_test.go
guscarreon: You are right. Looks better. Done.

exchange/bidder_test.go
guscarreon: Does it hurt to go through that extra layer?

pbsmetrics/go_metrics_test.go
guscarreon: Makes sense. Modified

pbsmetrics/go_metrics_test.go
guscarreon: The index was supposed to be for testing purposes. Modified.

pbsmetrics/prometheus/prometheus.go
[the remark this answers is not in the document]
guscarreon: Renamed to `requestsQueueTimer`

exchange/bidder.go
mansinahar: Ah ok! Yeah, requestBid is part of the interface so I guess we'll have to let it be for now.

exchange/bidder_test.go
mansinahar: Well, for unit tests it is better if we keep the tests close to the functions that are being modified because say tomorrow if somebody makes some changes to the `requestBid` function that has nothing to do directly with the connection metrics but can potentially lead to the failure of this test.

PR #1346, opened 2020-06-10
Adding Smartadserver adapter

tadam75:
Hello,

I have created this PR to add the Smart Adserver adapter. 

Thanks for reviewing it :)


adapters/smartadserver/smartadserver.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if response.StatusCode == http.StatusBadRequest {
+		return nil, []error{&errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
+		}}
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return nil, []error{fmt.Errorf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode)}
```
SyntaxNode, line 131: I recommend to use `&errortypes.BadServerResponse` here instead.

adapters/smartadserver/smartadserver.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+	return bidResponse, []error{}
+}
+
+// BuildEndpointURL : Builds endpoint url
+func (a *SmartadserverAdapter) BuildEndpointURL(params *openrtb_ext.ExtImpSmartadserver) (string, error) {
+
+	host := a.host
+	if params.Domain != "" {
```
SyntaxNode, line 158: Why would a bidder need to define their own domain? Is it entirely free form?

static/bidder-params/smartadserver.json
```diff
@@ -0,0 +1,39 @@
...
+
+  "type": "object",
+  "properties": {
+    "domain": {
+      "type": "string",
+      "description": "Server domain."
+    },
+    "siteId": {
+      "type": "integer",
+      "description": "The page id.",
```
SyntaxNode, line 14: Should this be "The site id." ?

static/bidder-info/smartadserver.yaml
```diff
@@ -0,0 +1,11 @@
+maintainer:
+  email: "support@smartadserver.com"
```
SyntaxNode, line 2: This looks like it might more of a general support email. I just want to confirm that if we send an email about the code in this repo we will routed to the correct team at Smart.

SyntaxNode:
Test coverage looks great. As you likely know, it's technically possible -  but extremely unlikely - for `json.Marshal` calls to fail. I appreciate you handling the errors just in case, but of course testing those flows isn't expected. The only suggestion I have is to add a test with no imps to cover the `len(request.Imp) == 0` condition.

adapters/smartadserver/smartadserver.go
tadam75: We used this domain parameter as a convenient way to call our local&development testing environments. 
Indeed, it's not really for bidders.

static/bidder-info/smartadserver.yaml
tadam75: Yes I have double checked it, this will redirect to the relevant team :)

tadam75:
Hello @SyntaxNode ,
Thank you for your comments and suggestions, it helps a lot. I have taken them all into account in my last commit. 
Just tell me if I have to remove the domain parameter.

adapters/smartadserver/smartadserver.go
SyntaxNode: Ah, ok. Thank you for the context. Could you please remove this ability? It would enable the adapter to act as a publisher defined proxy, which we'd like to avoid. For testing, please use a config or environment variable override to specify local host (that's what we do).

tadam75:
Hello @SyntaxNode ,
No problem, I have removed the domain parameter.

openrtb_ext/imp_smartadserver.go
```diff
@@ -0,0 +1,10 @@
+package openrtb_ext
+
+// ExtImpSmartadserver defines the contract for bidrequest.imp[i].ext.smartadserver
+type ExtImpSmartadserver struct {
+	Domain    string `json:"domain"`
```
guscarreon, line 5: `Domain` doesn't seem to be used by your adapter. I'm not saying we don't want to keep it, I'm just double checking it is not a leftover or something.

SyntaxNode: I missed that. Good eye, @guscarreon.

tadam75: Sorry I forgot it 😕

static/bidder-params/smartadserver.json
```diff
@@ -0,0 +1,35 @@
...
+      "type": "integer",
+      "description": "The network id.",
+      "minimum": 1
+    }
+  },
+  "dependencies": {
+    "siteId": { "required": ["pageId", "formatId"] },
+    "pageId": { "required": ["siteId", "formatId"] },
+    "formatId": { "required": ["siteId", "pageId"] }
+  },
```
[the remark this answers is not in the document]
tadam75: It was the cleanest way I've found to describe the behaviour : "siteId, pageId and formatId have to all be provided or all empty." (I wrote it in the documentation file).

So it's kind of a circular dependency but it seems to be handled correctly by the schema.

tadam75:
Hello @guscarreon ,
Thank you for reviewing this PR. I've added the params_test file 😊

PR #1349, opened 2020-06-11
Extra logging for timeout notifications

hhhjort:
More complete coverage of timeout notification issues.

### DIFF
--- a/exchange/bidder.go
+++ b/exchange/bidder.go
@@ -389,7 +389,15 @@ func (bidder *bidderAdapter) doTimeoutNotification(timeoutBidder adapters.Timeou
 			}
 		} else {
 			bidder.me.RecordTimeoutNotice(false)
+			if bidder.DebugConfig.TimeoutNotification.Log {
+				msg := fmt.Sprintf("TimeoutNotification: Failed to make timeout request: method(%s), uri(%s), error(%s)", toReq.Method, toReq.Uri, err.Error())
+				util.LogRandomSample(msg, glog.Warningf, bidder.DebugConfig.TimeoutNotification.SamplingRate)
+			}
 		}
+	} else if bidder.DebugConfig.TimeoutNotification.Log {
+		reqJSON, _ := json.Marshal(req)
+		msg := fmt.Sprintf("TimeoutNotification: Failed to generate timeout request: error(%s), bidder request(%s)", errL[0].Error(), string(reqJSON))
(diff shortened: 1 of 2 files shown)


exchange/bidder.go
```diff
@@ -389,7 +389,15 @@ func (bidder *bidderAdapter) doTimeoutNotification(timeoutBidder adapters.Timeou
 			}
 		} else {
 			bidder.me.RecordTimeoutNotice(false)
+			if bidder.DebugConfig.TimeoutNotification.Log {
+				msg := fmt.Sprintf("TimeoutNotification: Failed to make timeout request: method(%s), uri(%s), error(%s)", toReq.Method, toReq.Uri, err.Error())
+				util.LogRandomSample(msg, glog.Warningf, bidder.DebugConfig.TimeoutNotification.SamplingRate)
+			}
 		}
+	} else if bidder.DebugConfig.TimeoutNotification.Log {
+		reqJSON, _ := json.Marshal(req)
```
SyntaxNode, line 398: However unlikely it is for the marshaling to fail, please don't ignore the error response.

SyntaxNode, line 398: Should a metric be added here as well for an error in generating the timeout notification event request?

hhhjort: I am a little wary of creating too many metrics here, especially if we may move to per adapter metrics in the future. Do you think the same failure metric would make sense here?

hhhjort:
Gus, do you have a suggestion on how we can test this function?

hhhjort:
Managed to throw together a very basic test ... doesn't quite verify that it does anything, just that the code does not fail.

exchange/bidder_test.go
```diff
@@ -1229,6 +1229,64 @@ func TestSetAssetTypes(t *testing.T) {
...
+			TimeoutNotification: config.TimeoutNotification{
+				Log: true,
+			},
+		},
+		me: &metricsConfig.DummyMetricsEngine{},
+	}
+	if tb, ok := bidder.Bidder.(adapters.TimeoutBidder); !ok {
+		t.Error("Failed to cast bidder to a TimeoutBidder")
+	} else {
+		bidder.doTimeoutNotification(tb, &adapters.RequestData{})
```
SyntaxNode, line 1286: The setup is good. There's no assertion after the call to `doTimeoutNotification`. @bsardo encountered the same trouble in the story he's working on. I don't see a way to make it better now, but we should look to wrap the logger instead of calling the static glog directly so we can inject and inspect log entries.. as well as provide a mechanism to swap out to logrus in the future (via a config option).

PR #1350, opened 2020-06-11
Build With Go 1.14

SyntaxNode:
All PRs are now building with Go 1.14, so it's time for the final phase and build our release binaries with 1.14.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: go
 
 go:
-  - '1.12'
   - '1.13'
-  - '1.14'
+  - '1.14.2'
 
 go_import_path: github.com/prebid/prebid-server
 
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,9 +1,8 @@
 language: go
 
 go:
-  - '1.12'
   - '1.13'
-  - '1.14'
+  - '1.14.2'
```
SyntaxNode, line 5: Current version is 1.14.4. We recently updated Prebid Cache to Go 1.14.2 without any production issues, so using the same version here for consistency. If you feel differently regarding the minor version, please let me know.

SyntaxNode:
> LGTM in general! Just one thing to note is that, there was a new `go` directive added to the `go.mod` file starting v1.12. We likely wanna up that to `1.13` as well as it is suppose to be the minimum version that we support.
> 
> https://golang.org/ref/mod#tmp_12

Sounds good to me. Updated.

PR #1354, opened 2020-06-12
Adds Avocet adapter

scritchley:
This adds an adapter for Avocet. The adapter is disabled by default as each integration requires a specific endpoint.

Thanks in advance for your time.


adapters/avocet/avocet_test.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		want   *adapters.BidderResponse
+		errs   []error
+	}{
+		{
+			"204 No Content indicates no bids",
```
SyntaxNode, line 97: Nice use of table driven tests. Using this format, please consider specifying the struct field name to improve readability. Ie:

```
{
	name: ...,
	fields: ...,
	args: ...,
	want: ...,
	errs: ...,
}
```

adapters/avocet/avocet.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// AvocetAdapter implements a adapters.Bidder compatible with the Avocet advertising platform.
+type AvocetAdapter struct {
+	// Endpoint is a http endpoint to use when making requests to the Avocet advertising platform.
+	Endpoint string
+}
+
+func (a *AvocetAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	headers := http.Header{}
```
SyntaxNode, line 21: There's a lovely elegancy to just passing through the OpenRTB request. It's common for adapters to do a bit of checking before making the network call to place a bid, aka.. to fail fast. This typically includes checking for no impressions and checking for a compatible ad type. It's also common to pull out the W and H of banner ads from the formats slice.

Do you think any of this applies to the Avocet adapter?

scritchley:
Thanks for the comments, I've addressed most of them. On the point of validation within the `MakeRequests` method, I've added a check for a zero-length slice of imps, which is hopefully enough. I don't think we need much more validation than that and I'd almost prefer us to receive the http requests in cases of misconfiguration or missing objects, as we can debug and assist publishers more easily that way (without requiring their prebid-server logs). Hopefully that makes sense. Thanks!

adapters/avocet/avocet.go
```diff
@@ -0,0 +1,124 @@
...
+	if len(request.Imp) == 0 {
+		return nil, nil
+	}
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	body, err := json.Marshal(request)
+	if err != nil {
+		return nil, []error{&errortypes.FailedToRequestBids{
```
SyntaxNode, line 30: I think this is appropriate. The description for this error type is:
```
// FailedToRequestBids is an error to cover the case where an adapter failed to generate any http requests to get bids,
// but did not generate any error messages. This should not happen in practice and will signal that an adapter is poorly
// coded. If there was something wrong with a request such that an adapter could not generate a bid, then it should
// generate an error explaining the deficiency. Otherwise it will be extremely difficult to debug the reason why an
// adapter is not bidding.
```

Although I'd like to change that comment to exclude the "poorly coded" wording, I believe this error type is appropriate here. Failure to marshal is more of a sign of an internal server issue and not caused by user input. The chance of this happening is very very small.

PR #1357, opened 2020-06-15
Added additional Ext Param

telariaEng:
Added an `Extra` param to our ext object.


adapters/telaria/telaria.go
```diff
@@ -204,7 +210,13 @@ func (a *TelariaAdapter) MakeRequests(requestIn *openrtb.BidRequest, reqInfo *ad
...
 
 		// Swap the tagID with adCode
-		request.Imp[i].TagID = telariaExt.AdCode
+		request.Imp[i].TagID = telariaImpExt.AdCode
+	}
+
+	// Add the Extra from Imp to the top level Ext
+	if telariaImpExt != nil && telariaImpExt.Extra != nil {
+		// ignore the error because these params are optional
+		request.Ext, _ = json.Marshal(&telariaBidExt{Extra: telariaImpExt.Extra})
```
guscarreon, line 219: Can we please add a test case to cover the newly added functionality inside the if statement in line 217. Please also try to add the case where `telariaImpExt != nil` evaluates to `true` and `telariaImpExt.Extra != nil` evaluates to `false`.
image

telariaEng: Done.

guscarreon: @telariaEng thank you so much for addressing the feedback and your quick response. One more thing; I realize the chances of `json.Marshal` throwing an error are slim, but can we add an error handling mechanism right after line 219 for it please? When `json.Marshal` throws an error it has to do with malformed information that I believe is relevant to handle.

telariaEng: Thanks for the response. The reason I'm ignoring the error over there is because we don't want to throw away the request if there is a malformed JSON in there (we want to proceed with the request by just ignoring the params passed via the `Extra` object). How would you suggest I go about doing that?

SyntaxNode: Totally understandable if you'd like to ignore the `Extra` object if it's malformed. That error would actually happen when you call `Unmarshal` in `FetchTelariaExtImpParams`.. and even then, I believe the server would reject the malformed JSON earlier in the pipeline. There's quite a bit of validation and processing which occurs before your adapter runs.

Really.. there really isn't a good reason for `Marshal` to ever fail here. If it did, it would indiciate something is extremely wrong such as hardware failure or memory corruption. So, while ignoring errors from `Unmarshal` makes sense in some situations, you really don't want to do that for `Marshal`.

telariaEng: @SyntaxNode thank you for your feedback. So should I leave the code as is then?

SyntaxNode: @telariaEng Could you please add the error check? It likely will never be hit, but if it ever does it will provide the server host with useful information.

telariaEng: @SyntaxNode  Done.

PR #1358, opened 2020-06-16
Log account id and all bidder names when recovering from OpenRTB auction bidder…

bsardo:
… panic (#446)

### DIFF
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -299,7 +299,7 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
 	for bidderName, req := range cleanRequests {
 		// Here we actually call the adapters and collect the bids.
 		coreBidder := resolveBidder(string(bidderName), aliases)
-		bidderRunner := e.recoverSafely(func(aName openrtb_ext.BidderName, coreBidder openrtb_ext.BidderName, request *openrtb.BidRequest, bidlabels *pbsmetrics.AdapterLabels, conversions currencies.Conversions) {
+		bidderRunner := e.recoverSafely(cleanRequests, func(aName openrtb_ext.BidderName, coreBidder openrtb_ext.BidderName, request *openrtb.BidRequest, bidlabels *pbsmetrics.AdapterLabels, conversions currencies.Conversions) {
 			// Passing in aName so a doesn't change out from under the go routine
 			if bidlabels.Adapter == "" {
 				glog.Errorf("Exchange: bidlables for %s (%s) missing adapter string", aName, coreBidder)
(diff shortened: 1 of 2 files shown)


exchange/exchange.go
```diff
@@ -369,11 +369,25 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
...
 }
 
-func (e *exchange) recoverSafely(inner func(openrtb_ext.BidderName, openrtb_ext.BidderName, *openrtb.BidRequest, *pbsmetrics.AdapterLabels, currencies.Conversions), chBids chan *bidResponseWrapper) func(openrtb_ext.BidderName, openrtb_ext.BidderName, *openrtb.BidRequest, *pbsmetrics.AdapterLabels, currencies.Conversions) {
+func (e *exchange) recoverSafely(cleanRequests map[openrtb_ext.BidderName]*openrtb.BidRequest, inner func(openrtb_ext.BidderName, openrtb_ext.BidderName, *openrtb.BidRequest, *pbsmetrics.AdapterLabels, currencies.Conversions), chBids chan *bidResponseWrapper) func(openrtb_ext.BidderName, openrtb_ext.BidderName, *openrtb.BidRequest, *pbsmetrics.AdapterLabels, currencies.Conversions) {
 	return func(aName openrtb_ext.BidderName, coreBidder openrtb_ext.BidderName, request *openrtb.BidRequest, bidlabels *pbsmetrics.AdapterLabels, conversions currencies.Conversions) {
 		defer func() {
 			if r := recover(); r != nil {
-				glog.Errorf("OpenRTB auction recovered panic from Bidder %s: %v. Stack trace is: %v", coreBidder, r, string(debug.Stack()))
+
+				var allBidders string = ""
```
SyntaxNode, line 377: Syntax Nitpick: This is more commonly written as `allBidders := ""`

exchange/exchange.go
```diff
@@ -369,11 +369,25 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
...
+				var sb strings.Builder
```
SyntaxNode, line 378: Syntax Nitpick: I've seen a lot of code written this way, so it's cool with me. If you're interested, the non-var alternative is: `sb  := strings.Builder{}`

exchange/exchange.go
```diff
@@ -369,11 +369,25 @@ func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext
...
+				for k := range cleanRequests {
+					sb.WriteString(string(k))
+					sb.WriteString(",")
+				}
+				if sb.Len() > 0 {
+					allBidders = sb.String()[:sb.Len()-1]
+				}
+
+				glog.Errorf(
```
SyntaxNode, line 387: Syntax Nitpick: I would recommend the first line of the string to be moved up a line.. something like:
```
glog.Errorf("OpenRTB auction recovered panic from Bidder %s: %v. "+
    "Account id: %s, All Bidders: %s, Stack trace is: %v",
coreBidder, r, bidlabels.PubID, allBidders, string(debug.Stack()))
```

exchange/exchange.go
bsardo: 👍

PR #1359, opened 2020-06-16
Consumable: Correct bid type, should always be "banner".

djcsdy:
I am raising this PR on behalf of Consumable, CC @naffis.

This PR (hopefully) fixes issue #862 for Consumable, where the Consumable adapter would respond incorrectly when bidding on multi-format slots.

My understanding is that `BidTypeVideo` is for VAST units.

Consumable units are always HTML, never VAST. Sometimes the HTML unit itself embeds video, but my understanding is that from Prebid's point of view the `BidType` for that unit is still `BidTypeBanner`.

Therefore my understanding is that Consumable units are always `BidTypeBanner`.

Please let me know if my understanding is incorrect. I found the documentation for this quite difficult to understand.

### DIFF
--- a/adapters/consumable/consumable.go
+++ b/adapters/consumable/consumable.go
@@ -269,7 +269,7 @@ func (a *ConsumableAdapter) MakeBids(
 
 			bidderResponse.Bids = append(bidderResponse.Bids, &adapters.TypedBid{
 				Bid:     &bid,
-				BidType: getMediaTypeForImp(getImp(bid.ImpID, internalRequest.Imp)),
+				BidType: openrtb_ext.BidTypeBanner,
 			})
 		}
 	}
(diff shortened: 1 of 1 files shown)


adapters/consumable/consumable.go
```diff
@@ -269,7 +269,7 @@ func (a *ConsumableAdapter) MakeBids(
 
 			bidderResponse.Bids = append(bidderResponse.Bids, &adapters.TypedBid{
 				Bid:     &bid,
-				BidType: getMediaTypeForImp(getImp(bid.ImpID, internalRequest.Imp)),
+				BidType: openrtb_ext.BidTypeBanner,
```
mansinahar, line 272: @djcsdy Just a suggestion: Would be good to add a comment here explaining the reason why it's always set to banner.

djcsdy: Done, thanks.

PR #1360, opened 2020-06-17
Metrics for TCF 2 adoption

hhhjort:


exchange/utils.go
```diff
@@ -64,10 +75,18 @@ func cleanOpenRTBRequests(ctx context.Context,
...
+		parsedConsent, err := vendorconsent.ParseString(consent)
+		if err == nil {
+			cleanMetrics.tcfVersion = int(parsedConsent.Version())
+		}
+	}
 	// bidder level privacy policies
 	for bidder, bidReq := range requestsByBidder {
 
 		if gdpr == 1 {
+			cleanMetrics.gdprEnforced = true
```
SyntaxNode, line 89: This doesn't seem necessary since we have it done on line 79.

exchange/utils.go
```diff
@@ -17,6 +19,15 @@ import (
 	"github.com/prebid/prebid-server/privacy/lmt"
 )
 
+// cleanMetrics is a struct to export any metrics data resulting from cleanOpenRTBRequests(). It starts with just
+// the TCF version, but made a struct to facilitate future expansion
+type cleanMetrics struct {
+	// A simple flag if GDPR is being enforced on this request.
+	gdprEnforced bool
+	// a zero value means a missing or invalid GDPR string
+	tcfVersion int
```
SyntaxNode, line 28: Consider renaming `tcfVersion` to `gdprTcfVersion` to make the groping with `gdprEnforced` obvious as we'll add ccpa, lmt, and coppa in the future.

pbsmetrics/go_metrics.go
```diff
@@ -218,6 +228,10 @@ func NewMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName, d
 
 	newMetrics.TimeoutNotificationSuccess = metrics.GetOrRegisterMeter("timeout_notification.ok", registry)
 	newMetrics.TimeoutNotificationFailure = metrics.GetOrRegisterMeter("timeout_notification.failed", registry)
+
+	newMetrics.TCF1Meter = metrics.GetOrRegisterMeter("privacy.tcf.v1", registry)
```
SyntaxNode, line 232: I really like the new 'dotted' name format. Other metrics use a combination of underscore and dots. Thoughts on consistency?

SyntaxNode, line 232: I don't think it's clear from the name that this is bid request privacy metrics. We also have user sync related metrics, which I wouldn't mind refactoring in the future. Perhaps "privacy.requests.tcf.v1" or "request_privacy.tcf.v1"?

pbsmetrics/prometheus/prometheus.go
```diff
@@ -153,11 +156,16 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
...
 		[]string{cacheResultLabel})
 
-	metrics.timeout_notifications = newCounter(cfg, metrics.Registry,
+	metrics.timeoutNotifications = newCounter(cfg, metrics.Registry,
 		"timeout_notification",
 		"Count of timeout notifications triggered, and if they were successfully sent.",
 		[]string{successLabel})
 
+	metrics.tcfMetrics = newCounter(cfg, metrics.Registry,
+		"privacy_tcf",
```
SyntaxNode, line 165: Similar naming concern as with Influx. How about 'request_privacy_tcf'?

pbsmetrics/prometheus/prometheus.go
```diff
@@ -153,11 +156,16 @@ func NewMetrics(cfg config.PrometheusMetrics) *Metrics {
...
+		"Count of TCF versions for requests under GDPR.",
```
SyntaxNode, line 166: .. for requests with GDPR enforced.

exchange/utils.go
hhhjort: Done

exchange/utils.go
hhhjort: Yes, I moved this line to above, or rather copied it instead of moving. Thanks

pbsmetrics/go_metrics.go
hhhjort: timeout_notification should remain underscore because it is a compound name, rather than notification being a subcategory of timeout.

hhhjort: That is a good point, I have not included sync stats.

pbsmetrics/prometheus/prometheus.go
SyntaxNode: The change to the Influx name looks good. Can you please change the name here to match?

hhhjort: I don't need to change the name since I made the source a label instead. If we add more sources (usersync), we will want the same version options, so will be the same cardinality as having per source metrics.

pbsmetrics/prometheus/prometheus.go
hhhjort: fixed

exchange/utils.go
```diff
@@ -64,6 +75,13 @@ func cleanOpenRTBRequests(ctx context.Context,
 		LMT:   lmtPolicy.ShouldEnforce(),
 	}
 
+	if gdpr == 1 {
+		cleanMetrics.gdprEnforced = true
```
SyntaxNode, line 79: Cool. So we're logging when enforcement is in play, but not if the enforcement resulted in action. Privacy enforcement still could be avoided for some bidders depending on permission setup. The same will soon be true for CCPA. I don't have a good suggestion here, just wanted to call it out.

pbsmetrics/prometheus/prometheus.go
SyntaxNode: Cool. Could you please add this to the preload code?

hhhjort:
preload added

pbsmetrics/go_metrics.go
```diff
@@ -562,6 +582,14 @@ func (me *Metrics) RecordTimeoutNotice(success bool) {
 	return
 }
 
+func (me *Metrics) RecordTCFReq(version int) {
+	if version >= 0 && version < len(me.TCFReqVersion) {
+		me.TCFReqVersion[version].Mark(1)
+	}
+	me.TCFReqVersion[0].Mark(1)
```
SyntaxNode, line 589: Why are we always marking index 0?

pbsmetrics/go_metrics.go
```diff
@@ -562,6 +582,14 @@ func (me *Metrics) RecordTimeoutNotice(success bool) {
 	return
 }
 
+func (me *Metrics) RecordTCFReq(version int) {
+	if version >= 0 && version < len(me.TCFReqVersion) {
+		me.TCFReqVersion[version].Mark(1)
```
SyntaxNode, line 587: The use of index to correlate to a version string relies on the specific order of `TCFVersions()`. Do other metrics really do this? Would a map by the version string be safer?

pbsmetrics/go_metrics_test.go
```diff
@@ -56,6 +56,10 @@ func TestNewMetrics(t *testing.T) {
 
 	ensureContains(t, registry, "timeout_notification.ok", m.TimeoutNotificationSuccess)
 	ensureContains(t, registry, "timeout_notification.failed", m.TimeoutNotificationFailure)
+	ensureContains(t, registry, "privacy.request.tcf.v1", m.TCFReqVersion[1])
```
SyntaxNode, line 59: The special known indexes of `TCFReqVersion` seem a bit fragile IMHO.

pbsmetrics/prometheus/prometheus.go
```diff
@@ -28,7 +28,8 @@ type Metrics struct {
 	requestsWithoutCookie        *prometheus.CounterVec
 	storedImpressionsCacheResult *prometheus.CounterVec
 	storedRequestCacheResult     *prometheus.CounterVec
-	timeout_notifications        *prometheus.CounterVec
+	timeoutNotifications         *prometheus.CounterVec
+	tcfMetrics                   *prometheus.CounterVec
```
mansinahar, line 32: Can we please call this `tcfVersion` or something along those lines so that we're consistent with the rest of the metric names?

pbsmetrics/go_metrics.go
```diff
@@ -48,9 +48,13 @@ type Metrics struct {
 	ImpsTypeAudio  metrics.Meter
 	ImpsTypeNative metrics.Meter
 
+	// Notification timeout metrics
 	TimeoutNotificationSuccess metrics.Meter
 	TimeoutNotificationFailure metrics.Meter
 
+	// TCF adaption metrics
+	TCFReqVersion []metrics.Meter
```
mansinahar, line 56: I'd agree with Scott here about having indexes of a slice correlate with the TCF versions can be a little fragile. We should probably create a map here with the keys being of type `TCFVersionValue`

pbsmetrics/go_metrics.go
hhhjort: Oops, should be a return above. Thanks,. `[0]` is the "err" version.

pbsmetrics/go_metrics.go
hhhjort: Other metrics do not do this. They resolve the value upstream, and pass the code to the Record method. In this case, we already have version resolved to an integer, and felt it would make more sense to stringify it on the metrics side, which is probably the first place we will look if at some point down the line we want to change the strings that get recorded.

hhhjort: If I were to do this the way other metrics do it, I would have some code that looks like:
`if version == 1 {
    val = TCFVersionV1
} else if version == 2{
    val = TCFVersionV2
} else {
    val = TCFVersionErr
}`
However all the other metrics already produce lists of values in a predefined order. They just don't have an integer predefined for the value they represent that makes a good index. It is usually decoding HTTP status codes or checking error types, or just hardcoded based on the endpoint code it is in. Version number is unique in this way that it is easier to use the version number as the index to the version number than to have a set of ifs or a case statement to resolve the the version number to the stringify the version number. Or go back to my original version where I do not predefine the version strings, but generate them on the fly with `Sprinft()`

pbsmetrics/go_metrics.go
```diff
@@ -562,6 +584,15 @@ func (me *Metrics) RecordTimeoutNotice(success bool) {
 	return
 }
 
+func (me *Metrics) RecordTCFReq(version int) {
+	if version >= 0 && version < len(me.TCFVersions) {
```
SyntaxNode, line 588: Could this be simplified / future proofed with a map lookup?:
```
if m, ok := me.TCFVersions[version]; ok {
  m.Mark(1)
  return
}
me.TCFReqVersion[TCFVersionErr].Mark(1)
return
```

pbsmetrics/metrics_mock.go
```diff
@@ -106,3 +106,8 @@ func (me *MetricsEngineMock) RecordRequestQueueTime(success bool, requestType Re
 func (me *MetricsEngineMock) RecordTimeoutNotice(success bool) {
 	me.Called(success)
 }
+
+// RecordTCF mock
```
SyntaxNode, line 110: Typo: RecordTCFReq mock

pbsmetrics/go_metrics.go
hhhjort: Oh, I see ... it does push some logic to the map creation. This would require we set up a mapping of version numbers to version keys, something we don't do anywhere else in metrics. All that sort of mapping is done in the PBS core code itself, with the keys then passed to the metrics code. I think we would need to run through several TCF version iterations before we would really have to worry about saving space for omitting the older versions, at which point metrics itself is probably ripe for a refactor anyway after all those years more of development.

SyntaxNode: Discussed in person. I misread the code. The key was pulled out to a separate variable it make it easier to follow.

hhhjort:
Where would `convertTCFVerNumToStr` live, since it would not be specific to one of the metrics packages?

pbsmetrics/go_metrics.go
```diff
@@ -562,6 +582,11 @@ func (me *Metrics) RecordTimeoutNotice(success bool) {
 	return
 }
 
+func (me *Metrics) RecordTCFReq(version TCFVersionValue) {
+	me.TCFReqVersion[version].Mark(1)
```
SyntaxNode, line 586: If version isn't found in the map, would it be better to hardcode the unknown value or ignore the request instead of potentially getting a nil ref panic?

PR #1361, opened 2020-06-17
IQZone bidder adapter

trchandraprakash:
Hi @SyntaxNode This is a new adapter for IQZone.


Gopkg.lock
```diff
@@ -0,0 +1,628 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
```
SyntaxNode, line 1: The project has migrated to Go Modules and no longer uses dep. Please remove the gopkg files from the PR.

adapters/iqzone/iqzone.go
```diff
@@ -0,0 +1,236 @@
...
+	if len(pub2impressions) == 0 {
+		return nil, errs
+	}
+
+	result := make([]*adapters.RequestData, 0, len(pub2impressions))
+	for k, imps := range pub2impressions {
+		bidRequest, err := adapter.buildAdapterRequest(request, &k, imps)
+		if err != nil {
+			errs = append(errs, err)
+			return nil, errs
```
SyntaxNode, line 42: Should this be a continue instead to allow for a partial request of the valid imps, or do you want to specifically error out if one of the imps is invalid? I wouldn't normally expect errors from the `buildAdapterRequest`. If it does error, something is likely VERY wrong.. along the lines of memory corruption since the errors are only generated by marshal attemps and template building.

adapters/iqzone/iqzone.go
```diff
@@ -0,0 +1,236 @@
...
+
+// Builds enpoint url based on adapter-specific pub settings from imp.ext
+func (adapter *IQZoneAdapter) buildEndpointURL(params *openrtb_ext.ExtImpIQZone) (string, error) {
+	endpointParams := macros.EndpointTemplateParams{PublisherID: params.PublisherID}
+	return macros.ResolveMacros(adapter.EndpointTemplate, endpointParams)
+}
+
+//MakeBids translates IQZone bid response to prebid-server specific format
+func (adapter *IQZoneAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	var msg = ""
```
SyntaxNode, line 187: Nitpick: I'd probably remove this declaration and convert the other instances to declare in place, like:
```
msg := fmt.Sprintf("Unexpected http status code: %d", response.StatusCode)
```

adapters/iqzone/iqzone_test.go
```diff
@@ -0,0 +1,10 @@
+package iqzone
+
+import (
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"testing"
+)
+
+func TestJsonSamples(t *testing.T) {
+	adapterstest.RunJSONBidderTest(t, "iqzonetest", NewIQZoneBidder("http://nep.advangelists.com/xp/get?pubid={{.PublisherID}}"))
```
SyntaxNode, line 9: Does IQZone use the Advangelists servers? .. or is this a copy/paste oversight?

adapters/iqzone/params_test.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}
+}
+
+var validParams = []string{
+	`{"pubid": "19f1b372c7548ec1fe734d2c9f8dc688"}`,
+}
+
+var invalidParams = []string{
+	`{"publisher": "19f1b372c7548ec1fe734d2c9f8dc688"}`,
```
SyntaxNode, line 40: Please add a case of pubid being the wrong type, like:
```
`{"pubid": 42}`,
```

Gopkg.lock
trchandraprakash: Removed files

adapters/iqzone/iqzone.go
trchandraprakash: Thanks, i added to continue with valid impression.

adapters/iqzone/iqzone.go
trchandraprakash: done.

adapters/iqzone/iqzone_test.go
trchandraprakash: yes they do.

adapters/iqzone/params_test.go
trchandraprakash: done.

adapters/iqzone/iqzone_test.go
hhhjort: Is there a reason to have a dedicated adapter then rather than simply declaring an alias of the Advangelists adapter? It would mean we would only have one adapter to maintain rather than two.

trchandraprakash: Thanks @hhhjort. It's part of our client and tech need to get a new adapter for them.

trchandraprakash:
Hi @SyntaxNode, Let me know when we can make it available for our client.

trchandraprakash:
Hi, Can you please confirm all good, and when adapter can be made available for client to use.

adapters/iqzone/iqzone_test.go
hhhjort: How is the processing different between this adapter and the Advangelists one? The alias system was developed just for this use case, to avoid duplicate work. So I would like to understand what is missing to make it work for you.

trchandraprakash: Hi, Processing in both adapters is same at this moment. We wanted to have a different adapter name for our customer.  What changes are needed for implementing alias, do i have to update advangelists adapter.

hhhjort: An alias is done on the adapter side, or potentially the PBS host side. See [Auction Docs](https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/auction.md) , the section on Aliases. A PBS host can define a default request that gets merged with all incoming requests. Any aliases defined there will be available for any publishers to use without them having to declare them in their request.

trchandraprakash: @hhhjort 
PBS host - It seems all of the config is needed on server. I do not see any change in adapter. 
on Adapter - Is there any default request concept too, or a doc to look at what changes are needed.

hhhjort: All the config is done on the PBS host, no code change is needed. This is the default request documentation: https://github.com/prebid/prebid-server/blob/master/docs/developers/default-request.md

SyntaxNode:
This PR presents a new scenario which we don't currently have a solution for in Prebid Server. Aliases solve the problem, but they're a config for a single host and not a server-wide solution. I'm going to propose a solution this week, but that doesn't solve your immediate case. If you use the Xandr/AppNexus host, I can add the alias there for now until we solve the bigger picture.

trchandraprakash:
Hi @SyntaxNode, Sorry i was not able to get back. i had asked client to use alias, it was bit disappointing for them, but haven't received any response yet. Please wait for a week i will let you know. 
Btw what is the solution you are planning here?

Good to know you can help on AppNexus prebid hosting environment.
I am not sure if this is right place to ask, but recently one of our supply partner mention that they are not receiving video bids from us, after checking some logs they provided, i believe we need to implement prebid cache, can you confirm if we have to send cache id and url to Appnexus in case of video responses.

trchandraprakash:
Hi @SyntaxNode it seem for client alias is too technical, can you please help in creating alias for IQZone in AppNexus hosted environment.

trchandraprakash:
> Hi @SyntaxNode it seem for client alias is too technical, can you please help in creating alias for IQZone in AppNexus hosted environment.

Hi @SyntaxNode  Can you please help me on above, i just need iqzone to be alias of Advangelists in AppNexus.

SyntaxNode:
> Hi @SyntaxNode Can you please help me on above, i just need iqzone to be alias of Advangelists in App

We can do that. No problem. Expect that to go out today or tomorrow.

trchandraprakash:
Hi @SyntaxNode, Just want to make sure if changes are in place, before i confirm to client.

Also, just adding my previous question, let me know if you can help here?
"I am not sure if this is right place to ask, but recently one of our supply partner mention that they are not receiving video bids from us, after checking some logs they provided, i believe we need to implement prebid cache, can you confirm if we have to send cache id and url to Appnexus in case of video responses."

hhhjort:
Video tends to make use of cache, but it doesn't always. Prebid Server is a middle man, it distributes bid requests to the bidders, and collects the bid responses back again into one large response. Depending on what is on the user end to display the video, cache may or may not be needed. The openrtb response returns everything needed to display the ad, but quite a few players require that a video ad be given in the form of a URL to something rather being provided directly from the code on the page. This URL for a specific ad is what cache is for.

So they should be receiving your video bids. If they are looking at targeting keys rather than the openrtb bids, the keys may be missing if caching was requested, but PBS was unable to cache the results. But I would expect that all the PBS bids would be missing rather than just yours.

trchandraprakash:
@hhhjort  Thanks. 

So if you see below targeting from sample response which our publisher got from Rythmone (bidder), this works for them. 

I can see couple of cache_ids parameters, so most likely rhythmone is caching VAST/bids on their side, via prebid cache right?
Or VAST / bids are being cached on host which is Appnexus and targeting is getting appended?

              "targeting": {
                "hb_bidder": "rhythmone",
                "hb_bidder_rhythmone": "rhythmone",
                "hb_cache_id": "d0c43530-1ce6-4f87-8f9f-f9ae62055096",
                "hb_cache_id_rhythmon": "d0c43530-1ce6-4f87-8f9f-f9ae62055096",
                "hb_env": "mobile-app",
                "hb_env_rhythmone": "mobile-app",
                "hb_pb": "7.50",
                "hb_pb_rhythmone": "7.50",
                "hb_size": "320x480",
                "hb_size_rhythmone": "320x480",
                "hb_uuid": "c9ebefd3-1c48-45e9-9269-9f52c4917df9",
                "hb_uuid_rhythmone": "c9ebefd3-1c48-45e9-9269-9f52c4917df9"
              },
              "type": "video"
            }
          }
        }
      ],
      "seat": "rhythmone"

hhhjort:
Assuming that is a response from PBS, then the caching is being done by PBS. PBS is configured to work with prebid cache, it is not aware of any caching being done further up the chain, so that will not appear in the generated targeting parameters. I can imagine a situation where the "ad" that got stored in prebid cache is simply a tag to retrieve the add from the bidder's system. And the ad in the bidder's system is simply a tag to retrieve the ad from the advertiser's agency's system. The point of the prebid level caching is to be able to guarantee that we have a URL that can be used to retrieve the ad, rather than relying on the ability of the viewer to extract and display the ad from the openrtb bid. If you are not hosting a prebid server instance, you do not need to worry about prebid cache.

trchandraprakash:
Hi @hhhjort Sorry for delay and thank you  for details. 

I may have more questions later, but for now, i will take your comments as all good. And based on your comments it seems a general VAST RTB response on prebid will work.

On an another note, regarding this PR, IQZone publisher "Weatherbug" still not finding IQZone alias'ed bidder on Appnexus. Any idea what could be the issue.

hhhjort:
Are they using our free prebid server or our integrated premium offering? If the latter they should work with their account manager to get that supported.

SyntaxNode:
> On an another note, regarding this PR, IQZone publisher "Weatherbug" still not finding IQZone alias'ed bidder on Appnexus. Any idea what could be the issue.

I think this is veering a bit too much into a specific issue with the Xandr host.  Please continue this discussion over email at prebid-server@xandr.com.

SyntaxNode:
I created a proposal for a long term solution:
https://github.com/prebid/prebid-server/issues/1429

trchandraprakash:
@hhhjort Thanks, I do not know if they have paid or free account, will check and drop an email over prebid-server@xandr.com for further discussion.
@SyntaxNode Thanks. I like your proposed changes.

Issue #1332, opened 2020-06-03
IPv6 Support


bretg:
IPv6 is becoming more common, and Prebid Server needs some updates to handle it in a privacy-appropriate way. Note that IPv6 addressing is quite different than IPv4 - there are a number of good resources on the net to understand them.

## Config

We should support the following host configuration options:

- A setting to anonymize the network piece when we need to protect user location. Suggested config: ipv6.anon-left-mask-bits. Default to 56 bits. Throw an error if this isn't >=56 and <=32.
- A setting to remove a number of bits on the right regardless of whether it's a situation that calls for explicit privacy. The idea is that no one downstream needs the whole device MAC address. Suggested config: ipv6.always-mask-right. Default to default 64 bits
- A setting to define the network ranges considered "private" and therefore ignored. Suggested config: ipv6.private-networks. If not specified, default to [::1/128, fc00::/7, fe80::/10]

## Proposed Algorithm
1. If the device.ip field is present in the incoming request, validate it.
   1. If not valid or is determined to be a private network, remove.
   2. If the current privacy environment requires, anonymize the address. (see algorithm below)
2. If the device.ipv6 field is present in the incoming request, validate it.
   1. If not valid or is determined to be a private network, remove.
   2. If the current privacy environment requires, anonymize the address. (see algorithm below)
   3. Even if we don't need to anonymize the address, check the always-mask-right configuration to see if we need to remove any bits off the right side.
3. If device.ip and/or device.ipv6 are validated, we're done. We're going to trust the upstream clients to provide these correctly.
4. If we don't have either, then we need to find one. Start with the True-Client-IP header. If it exists:
    1.  Validate, anonymize, and mask as needed
    2. Write to the appropriate device.{ip,ipv6} field
5. Else, continue by parsing the comma-separated IP values in XFF and looping through the addresses from left to right
   1. If the address is IPv6, check to see if matches one of these the configured IPv6 private networks
      1. If so, skip it and go to the next address
      2. If the current privacy environment requires, anonymize the address. (see algorithm below)
      3. Even if we don't need to anonymize the address, check the always-mask-right configuration to see if we need to remove any bits off the right side.
      4. If the OpenRTB device.ipv6 field isn't already present, write the new address. If there was an address already, and there's a post-mask difference between the addresses, log a warning.
      5. Break out of loop
[...]

Issue #1106, opened 2019-11-06
Prebid Server: change how IP addresses are determined using XFF header


bretg:
It's come to our attention that Prebid Server isn't processing the X-Forwarded-For (XFF) HTTP header quite right.

Currently, it's taking the leftmost entry from the comma-separated list, but sometimes that leftmost entry is a 'private' IP (e.g. NAT) that's not useful for geo-targeting purposes. A better algorithm would be:

1) First, use the True-Client-IP header if it's present
2) Else, parse the comma-separated IP values in XFF and loop through them
    1a) compare the value against the [private network table](https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses)
    1b) if the address matches a private network pattern, skip it and go on to the next one
3) in this way, set device.ip to the leftmost IP address that's not a private address.

dimention:
We have this problem, the determined IP is private, and it is sent to SSPs, that don't bid on it.

rpanchyk:
Implemented for PBS-Java in PR https://github.com/rubicon-project/prebid-server-java/pull/515

bretg:
Reviewed in PBS committee today and added reference to `True-Client-IP`

SyntaxNode:
This was recently implemented in PBS-Go along with IPv6 support.

PR #1362, opened 2020-06-19
Improved IPv6 Support + Private Network Filtering

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/1106 and the non-privacy features of https://github.com/prebid/prebid-server/issues/1332.


util/iputil/parse.go
```diff
@@ -0,0 +1,28 @@
...
+const (
+	IPvUnknown IPVersion = 0
+	IPv4       IPVersion = 4
+	IPv6       IPVersion = 6
+)
+
+// ParseIP parses v as an ip address returning the result and version, or nil and unknown if invalid.
+func ParseIP(v string) (net.IP, IPVersion) {
+	if ip := net.ParseIP(v); ip != nil {
+		if strings.ContainsRune(v, ':') {
```
SyntaxNode, line 21: net.ParseIP automatically converts IPv4 addresses to IPv6. There's no way to tell if the original value was IPv4 or IPv6 without the above check. This approach is common in other Go libraries.

config/config.go
```diff
@@ -497,6 +499,21 @@ func New(v *viper.Viper) (*Configuration, error) {
...
+	if err := c.RequestValidation.Parse(); err != nil {
+		return nil, err
+	}
+
+	if err := isValidCookieSize(c.HostCookie.MaxCookieSizeBytes); err != nil {
+		glog.Fatal(fmt.Printf("Max cookie size %d cannot be less than %d \n", c.HostCookie.MaxCookieSizeBytes, MIN_COOKIE_SIZE_BYTES))
+		return nil, err
+	}
+
+	glog.Info("Logging the resolved configuration:")
```
hhhjort, line 511: Why did you move this logging above the map generation? This way the generated maps will not be included in the logs.

SyntaxNode: The logging call looks to correlate to the validate call. Such that, if validation fails you can refer to the logged configuration to figure out what went wrong. The hash sets don't affect validation, they're just an optimization. so I moved them below the validation call.

util/iputil/matcher.go
```diff
@@ -0,0 +1,48 @@
+package iputil
+
+import (
+	"net"
+)
+
+// IPMatcher is the interface for matching on an ip address and version.
+type IPMatcher interface {
```
hhhjort, line 8: I wonder if we should change "Match" to "Validate". We have the counterintuitive result that if `Match()` returns true, the IP was not on the list of IPs in the IPMatcher

SyntaxNode: Yeah. I cringed at that too. We're looking for public networks from a list of private networks, so there's got to be a negation somewhere. Are you suggesting only a name change or to flip the logic too?

I choose the matcher name by following the naming pattern used by the standard sort package. I'm sure having FindIP take an IPValidator works too. :)

hhhjort: I was just thinking of changing the name. If we flip the logic, then there is no need to change the name in my view.

SyntaxNode: Sure.

endpoints/openrtb2/auction.go
```diff
@@ -914,13 +920,27 @@ func validateRegs(regs *openrtb.Regs) error {
...
 // setFieldsImplicitly uses _implicit_ information from the httpReq to set values on bidReq.
 // This function does not consume the request body, which was set explicitly, but infers certain
 // OpenRTB properties from the headers and other implicit info.
 //
 // This function _should not_ override any fields which were defined explicitly by the caller in the request.
 func (deps *endpointDeps) setFieldsImplicitly(httpReq *http.Request, bidReq *openrtb.BidRequest) {
-	setDeviceImplicitly(httpReq, bidReq)
+	sanitizeRequest(bidReq, deps.privateNetworkIPMatcher)
+
+	setDeviceImplicitly(httpReq, bidReq, deps.privateNetworkIPMatcher)
```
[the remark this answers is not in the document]
SyntaxNode: You're right, the information is available in the cfg object already. My thought in passing it down in the `endpointDep` is to avoid recreating the matcher object for each request, to save a bit on memory usage.

config/config.go
SyntaxNode: I switched it back.

util/iputil/matcher.go
SyntaxNode: Renamed to IPValidator and IPValidator.IsValid.

PR #1364, opened 2020-06-22
Fall back to constant rates when the currency rates endpoint i…

bsardo:
…s not accessible and rates are stale

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -212,6 +212,7 @@ type Analytics struct {
 type CurrencyConverter struct {
 	FetchURL             string `mapstructure:"fetch_url"`
 	FetchIntervalSeconds int    `mapstructure:"fetch_interval_seconds"`
+	StaleRatesSeconds    int    `mapstructure:"stale_rates_seconds"`
 }
 
 func (cfg *CurrencyConverter) validate(errs configErrors) configErrors {
@@ -853,6 +854,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("lmt.enforce", true)
 	v.SetDefault("currency_converter.fetch_url", "https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json")
 	v.SetDefault("currency_converter.fetch_interval_seconds", 1800) // fetch currency rates every 30 minutes
+	v.SetDefault("currency_converter.stale_rates_seconds", 7200)    // fetched currency rates are stale at 120 minutes
 	v.SetDefault("default_request.type", "")
 	v.SetDefault("default_request.file.name", "")
 	v.SetDefault("default_request.alias_info", false)
--- a/config/config.go
+++ b/config/config.go
@@ -854,7 +854,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("lmt.enforce", true)
 	v.SetDefault("currency_converter.fetch_url", "https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json")
 	v.SetDefault("currency_converter.fetch_interval_seconds", 1800) // fetch currency rates every 30 minutes
-	v.SetDefault("currency_converter.stale_rates_seconds", 7200)    // fetched currency rates are stale at 120 minutes
+	v.SetDefault("currency_converter.stale_rates_seconds", 0)
 	v.SetDefault("default_request.type", "")
 	v.SetDefault("default_request.file.name", "")
 	v.SetDefault("default_request.alias_info", false)
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -168,7 +168,7 @@ func (rc *RateConverter) LastUpdated() time.Time {
 
 // Rates returns current conversions rates
 func (rc *RateConverter) Rates() Conversions {
-	if rates := rc.rates.Load(); rates != nil && rates != (*Rates)(nil) {
+	if rates := rc.rates.Load(); rates != (*Rates)(nil) && rates != nil {
 		return rates.(*Rates)
 	}
 	return rc.constantRates
--- a/currencies/rate_converter.go
+++ b/currencies/rate_converter.go
@@ -170,7 +170,9 @@ func (rc *RateConverter) LastUpdated() time.Time {
 
 // Rates returns current conversions rates
 func (rc *RateConverter) Rates() Conversions {
-	if rates := rc.rates.Load(); rc.CheckRatesCleared(rates) && rates != nil {
+	// atomic.Value field rates is an empty interface and will be of type *Rates the first time rates are stored
+	// or nil if the rates have never been stored
+	if rates := rc.rates.Load(); rates != (*Rates)(nil) && rates != nil {
 		return rates.(*Rates)
 	}
 	return rc.constantRates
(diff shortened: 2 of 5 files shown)


currencies/rate_converter.go
```diff
@@ -160,14 +165,26 @@ func (rc *RateConverter) LastUpdated() time.Time {
...
-	if rc.constantRates != nil {
-		// Converter is not active, returning the constant rates
-		return rc.constantRates
-	}
-	if rates := rc.rates.Load(); rates != nil {
+
+	rates := rc.rates.Load()
+	lastUpdated := rc.lastUpdated.Load()
+
+	if rates != nil && lastUpdated != nil && !rc.CheckStaleRates() {
```
mansinahar, line 172: How about checking for stale rates during periodic fetching if we were unable to fetch the rates? On the first thought, the approach in this PR seemed okay too but now I am thinking that we'd call `Rates()` for every auction and will therefore go through this piece of code a lot more times as compared to the other approach which I don't think is needed.

The only drawback I can think of the approach that I am suggesting i.e. checking for stale rates during periodic fetching is that if we have say for example, a `fetchInterval` of 45 mins and `staleThreshold` of 120 mins then we will be using the stale rates for 135 mins instead of 120 mins. But, I am not super concerned about this scenario because these are both configurable and we (or anyone else who hosts prebid server) can keep this in mind when they configure these values. Perhaps we can add a warning of some sort where the `StaleRatesSeconds` config option is defined.

bsardo: Good point on the potential performance impact. Under normal circumstances, we shouldn't need to do the time math.

bsardo: @mansinahar and I talked about this offline. I'm going to look at moving the stale rate detection logic to the periodic task and clear the rates when the rates are stale. This logic would then just check for rate presence and use the constant rates if they are nil.

currencies/rate_converter.go
```diff
@@ -160,14 +168,22 @@ func (rc *RateConverter) LastUpdated() time.Time {
...
+	if rates := rc.rates.Load(); rates != nil && rates != (*Rates)(nil) {
 		return rates.(*Rates)
 	}
-	return nil
+	return rc.constantRates
+}
+
+// CheckStaleRates checks if loaded third party conversion rates are stale
+func (rc *RateConverter) CheckStaleRates() bool {
+	currentTime := time.Now().UTC()
```
mansinahar, line 179: We likely wanna add a check here to make sure the `staleRatesThreshold` value is greater than `0`. If it is `0` then we should never consider the rates stale i.e. the same behavior that we had before.

config/config.go
```diff
@@ -853,6 +854,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("lmt.enforce", true)
 	v.SetDefault("currency_converter.fetch_url", "https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json")
 	v.SetDefault("currency_converter.fetch_interval_seconds", 1800) // fetch currency rates every 30 minutes
+	v.SetDefault("currency_converter.stale_rates_seconds", 7200)    // fetched currency rates are stale at 120 minutes
```
mansinahar, line 857: I think we should default this to `0` and consider `rates` to be never stale if it is set to `0` i.e. the behavior that we have today. So, if somebody who is hosting prebid server today don't use this new config they will still see the same behavior as before.

currencies/rate_converter.go
```diff
@@ -160,14 +168,25 @@ func (rc *RateConverter) LastUpdated() time.Time {
 
 // Rates returns current conversions rates
 func (rc *RateConverter) Rates() Conversions {
-	if rc.constantRates != nil {
-		// Converter is not active, returning the constant rates
-		return rc.constantRates
-	}
-	if rates := rc.rates.Load(); rates != nil {
+	if rates := rc.rates.Load(); rates != (*Rates)(nil) && rates != nil {
```
SyntaxNode, line 171: How is `rates != (*Rates)(nil)` different than `rates != nil`?

currencies/rate_converter.go
```diff
@@ -111,6 +116,9 @@ func (rc *RateConverter) Update() error {
 		rc.rates.Store(rates)
 		rc.lastUpdated.Store(time.Now())
 	} else {
+		if rc.CheckStaleRates() {
+			rc.rates.Store((*Rates)(nil))
+		}
 		glog.Errorf("Error updating conversion rates: %v", err)
```
SyntaxNode, line 122: Ok, let me check if I've got this right. If there's an error fetching the rates json and we should check for stale rates (CheckStaleRates returns true), set the rates to nil which is a special marker to tell it to update next time?

currencies/rate_converter.go
```diff
@@ -160,14 +168,25 @@ func (rc *RateConverter) LastUpdated() time.Time {
 
 // Rates returns current conversions rates
 func (rc *RateConverter) Rates() Conversions {
```
SyntaxNode, line 170: Ok, cool. To verify, it's not impossible to get a nil response from Rates(), right?

currencies/rate_converter.go
```diff
@@ -160,14 +168,25 @@ func (rc *RateConverter) LastUpdated() time.Time {
...
-	return nil
+	return rc.constantRates
+}
+
+// CheckStaleRates checks if loaded third party conversion rates are stale
+func (rc *RateConverter) CheckStaleRates() bool {
+	if rc.staleRatesThreshold <= 0 {
+		return false
+	}
+	currentTime := time.Now().UTC()
```
SyntaxNode, line 182: If you factor out the `time.Now()`call to a dependency, you can mock the time and avoid the sleeps in the tests.

currencies/rate_converter.go
SyntaxNode, line 122: Consider adding to the error log that we now fall back to a constant rate conversion.

currencies/rate_converter.go
bsardo: 👍 will handle this in the currency rate refactor PR.

currencies/rate_converter.go
```diff
@@ -168,12 +170,24 @@ func (rc *RateConverter) LastUpdated() time.Time {
 
 // Rates returns current conversions rates
 func (rc *RateConverter) Rates() Conversions {
-	if rates := rc.rates.Load(); rates != (*Rates)(nil) && rates != nil {
+	if rates := rc.rates.Load(); rc.CheckRatesCleared(rates) && rates != nil {
```
bsardo, line 173: I didn't want to pass `rates` in here but I also didn't want to call `rc.rates.Load()` a second time in this method because it seems like unnecessary work. Given that fact, I'm not sure how much added value we get here by the self documenting method name (which I was initially on board with when we discussed) since we have introduced a little awkwardness into this method call.

SyntaxNode: Two options here. You can write the helper method a bit differently, as you don't need to add a RateConverter receiver. ie:
```
func checkRatesCleared(rates interface{}) bool {
	// atomic.Value field rates is an empty interface and will be of type *Rates the first time rates are stored
	return rates != (*Rates)(nil)
}
```

.. or remove the helper method if you feel it hurts the flow. Whichever you think is best.

PR #1366, opened 2020-06-22
PBS-632 add max connections per host config setting to general http a…

bsardo:
…nd cache clients

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -74,6 +74,7 @@ type Configuration struct {
 const MIN_COOKIE_SIZE_BYTES = 500
 
 type HTTPClient struct {
+	MaxConnsPerHost     int `mapstructure:"max_connections_per_host"`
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -668,9 +669,11 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("host_cookie.value", "")
 	v.SetDefault("host_cookie.ttl_days", 90)
 	v.SetDefault("host_cookie.max_cookie_size_bytes", 0)
+	v.SetDefault("http_client.max_connections_per_host", 0) // unlimited
```
SyntaxNode, line 672: Sounds good. This is the default we have today.

bsardo:
Oops! I had a test file exclude filter turned on which is why I wasn't seeing tests for the config. Updated.

PR #1367, opened 2020-06-23
Add ext.bidder.zoneid for Kubient adapater

mapcuk:


adapters/kubient/kubient.go
```diff
@@ -28,6 +28,24 @@ func (adapter *KubientAdapter) MakeRequests(
 	requestsToBidder []*adapters.RequestData,
 	errs []error,
 ) {
+	var bidderExt adapters.ExtImpBidder
+	if err := json.Unmarshal(openRTBRequest.Imp[0].Ext, &bidderExt); err != nil {
```
SyntaxNode, line 32: Please do not assume that there is at least 1 imp in the request. It's a common approach to check the number of Imps at the start of this method and exit early:

```
	if len(request.Imp) == 0 {
			Message: "No impression in the bid request",
		}}
	}
```

Please also add a test case with no impressions.

static/bidder-params/kubient.json
```diff
@@ -3,5 +3,12 @@
...
   "type": "object",
-  "properties": {  }
+  "properties": {
+    "zoneid": {
+      "type": "string",
+      "description": "Zone ID identifies Kubient placement ID.",
+      "minLength": 1
+    }
+  },
+  "required": ["zoneid"]
```
SyntaxNode, line 13: Be careful here. I'm stealing this warning from @hhhjort:

> Do you have any current traffic coming from PBS? When you set zoneid as required, if a request comes in missing zoneid once it is set to required, it will mark the whole request as failed and not send a bid request to any bidder. So adding a required field can break the entire prebid setup of a publisher.
>
> This behavior is to help ensure publishers set up their configs correctly. Otherwise a bidder might set up a request with 5 bidders, but bidder ABC has a bad config and always errors out. If the other 4 are working normally, the publisher might not notice they have a bad config for quite a while. If some pages have a valid ABC config, they might just conclude that ABC just doesn't bid much if they don't troubleshoot carefully.
>
> If there are publishers who currently have configs that omit zoneid. we would recommend that initially the adapter not advertise zoneid as required in bidder-params, but return an error from within the adapter if it is missing. An error from within the adapter will not invalid the whole request, allowing the other bidders to function. Then after all publishers have had a reasonable amount of time to update their configs, we can do the additional step of making it required in bidder-params.

adapters/kubient/kubient.go
mapcuk: @SyntaxNode added test for num of imps. Every imp object is checked.

static/bidder-params/kubient.json
mapcuk: Currently we don't have any traffic from prebid.
I'll change JSON schema to make **zoneid** as optional.

hhhjort: Feel free at some later point (a few months) to submit a PR to make it required. I have less sympathy for a publisher who has been ignoring months of warning messages and the adapter not working. Just want to be sure publishers have sufficient warning before making a breaking change.

PR #1368, opened 2020-06-23
OpenX + Facebook Hardening

SyntaxNode:
Mixed a few instances of editing shared memory. Added unit test coverage of the Facebook adapter and removed unused legacy components.


adapters/audienceNetwork/facebook.go
```diff
@@ -239,102 +232,106 @@ func (this *FacebookAdapter) extractPlacementAndPublisher(out *openrtb.Imp) (str
...
 
-	return placementId, publisherId, nil
+	return placementID, publisherID, nil
 }
 
 // XXX: This entire function is just a hack to get around mxmCherry 11.0.0 limitations, without
 // having to fork the library and maintain our own branch
-func modifyImpCustom(json []byte, imp *openrtb.Imp) ([]byte, error) {
+func modifyImpCustom(jsonData []byte, imp *openrtb.Imp) ([]byte, error) {
 	impType, ok := resolveImpType(imp)
```
SyntaxNode, line 267: Removed use of jsonparser.Set and replaced with standard library. The Set and Delete operations in jsonparser are experimental and shouldn't be used.

adapters/openx/openx.go
```diff
@@ -143,11 +143,13 @@ func preprocess(imp *openrtb.Imp, reqExt *openxReqExt) error {
 	}
 
 	if imp.Video != nil {
+		videoCopy := *imp.Video
```
SyntaxNode, line 146: @laurb9 I missed this in the review of https://github.com/prebid/prebid-server/pull/1290. Video is shared memory and should be copied before mutation.

util/maputil/maputil_test.go
```diff
@@ -0,0 +1,30 @@
...
+		},
+	}
+
+	for _, test := range testCases {
+		resultMap, resultOK := ReadEmbeddedMap(test.value, test.key)
+
+		assert.Equal(t, test.expectedMap, resultMap)
+		assert.Equal(t, test.expectedOK, resultOK)
+	}
+}
```
SyntaxNode, line 30: Eek! I need to finish this. Coming soon.

adapters/audienceNetwork/facebook.go
```diff
@@ -178,8 +170,10 @@ func (this *FacebookAdapter) modifyImp(out *openrtb.Imp) error {
 		}
 	}
 
-	switch impType {
-	case openrtb_ext.BidTypeBanner:
+	if impType == openrtb_ext.BidTypeBanner {
+		bannerCopy := *out.Banner
```
SyntaxNode, line 174: We missed this one too. Accidental editing of shared memory.

adapters/openx/openx.go
laurb9: Thanks. I thought this was safe because the adapter makes a copy of the entire request (unlike others), but I should have noticed the pointer.

By the way, do you use a tool to find these problems ? I'm just starting in Go and this could be useful to add to my toolbox.

adapters/audienceNetwork/audienceNetworktest/supplemental/server-error-500.json
```diff
@@ -0,0 +1,93 @@
+{
```
bsardo, line 1: It looks like we're missing test coverage for the the error handling block on line 368 of facebook.go where we check that the stringified json `adm` is well-formed: 
```
var obj facebookAdMarkup
if err := json.Unmarshal([]byte(bid.AdM), &obj); err != nil {
	errs = append(errs, &errortypes.BadServerResponse{
		Message: err.Error(),
	})
	continue
}
```
Perhaps you want to punt on this but I see you added coverage for the other error paths so I figured I'd mention it.

adapters/openx/openx.go
SyntaxNode: IMHO, Go doesn't provide the necessary syntax to express memory safety in code as other languages do. The best tool we have to find these kinds of issues is the [race detector](https://golang.org/doc/articles/race_detector.html). However, that tool isn't a static analyzer. It's limited to checking code that's running during a test (which is why we push for race detector test cases so much during code reviews). If you find a good static memory safety analyzer, please let us know. :)

util/maputil/maputil_test.go
SyntaxNode: Finished.

adapters/audienceNetwork/facebook_test.go
```diff
@@ -93,3 +95,25 @@ func TestMakeTimeoutNoticeBadRequest(t *testing.T) {
 	assert.NotNil(t, err, "Facebook MakeTimeoutNotification() did not return an error")
 
 }
+
+func TestNewFacebookBidderMissingPlatformID(t *testing.T) {
+	result := NewFacebookBidder("", "anyAppSecret")
+
+	expected := &adapters.MisconfiguredBidder{
+		Name:  "audienceNetwork",
+		Error: fmt.Errorf("Audience Network is not configured properly on this Prebid Server deploy. If you believe this should work, contact the company hosting the service and tell them to check their configuration."),
```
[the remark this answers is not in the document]
SyntaxNode: `fmt.Errorf` is basically a shorter form of `errors.New(fmt.Sprintf("", a, b, c...))`. If there's no formatting, it's a waste to use `fmt.Errorf`. I'll fix that.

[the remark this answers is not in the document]
SyntaxNode: This is a test file. I don't really care much to performance optimize unit tests. Might be a good idea for non-test code if the compiler isn't doing that for us already. Worth a check if you have some free time.

adapters/audienceNetwork/audienceNetworktest/supplemental/server-error-500.json
SyntaxNode: Thanks for calling it out. I've added a test for this case, but I'm fine cutting off coverage expansion for this PR. It's also not possible to test `json.Marshal` failures in most cases without extracting and mocking the json package.

Issue #1091, opened 2019-10-28
Prebid Server: Multiple Root SChain nodes


bretg:
(This is the Prebid Server side of https://github.com/prebid/Prebid.js/issues/3870)

Copying the use case from that issue for context:

```
I'm a publisher. I work with SuperAds, a fictitious monetization platform who provides me prebid.js implementation and access to multiple exchanges. SuperAds does this by configuring prebid.js for me, with multiple bidders (Exchanges A and B), making use of the seller account (publisher ID) for SuperAds on those exchanges. However, I also have a direct relationship with ShinyClicks, an ad exchange with a prebid.js adapter, and SuperAds configures that in prebid.js for me too.

SuperAds has assigned me publisher ID 123 internally.

In this case, the supply chains needing to be sent out in bid requests by prebid.js would be:

Bidder for Exchange A: superads.com, 123
Bidder for Exchange B: superads.com, 123
Bidder for ShinyClicks: (I own the account on ShinyClicks)

In the first two cases, the exchange pays SuperAds, who pays me. I don't control the accounts on the exchanges.
In the third case, ShinyClicks directly pays me, as I control that account.
```

In this scenario, Prebid Server is not receiving a single simple schain on source.ext.schain. Rather, it will receive a list of schains that are appropriate for different bidders. Here's the proposed location and format:
```
ext.prebid.schains: [
   { bidders: ["shinyClicks"], schain: { SCHAIN OBJECT }},
   { bidders: ["*"], schain: { SCHAIN OBJECT }}
]
```

Prebid Server's job would be to parse this list and pass the correct source.ext.schain to each bidder. It should also remove ext.prebid.schains before sending to the adapter.

If there's already an source.ext.schain and a bidder is named in ext.prebid.schains, the latter would override the former.

bretg:
This is done in PBS-Java.

bjorn-lw:
I'd like to bump this one. We have cases where this support is needed.

SyntaxNode:
Noted. We've been focused on privacy policies a lot recently. I understand it's a common ask from the community and have bumped up its priority within Xandr. I added the 'help wanted' tag if someone else is able to get to it first.

bsardo:
I'm picking this up for PBS-Go.

bsardo:
[...]

PR #1374, opened 2020-06-30
Add support for multiple root schain nodes

bsardo:


openrtb_ext/request.go
```diff
@@ -12,12 +12,43 @@ type ExtRequest struct {
...
+	Nodes    []*ExtRequestPrebidSChainSChainNode `json:"nodes"`
+	Ver      string                              `json:"ver"`
+	Ext      json.RawMessage                     `json:"ext,omitempty"`
+}
+
+// ExtRequestPrebidSChainSChainNode defines the contract for bidrequest.ext.prebid.schains[i].schain[i].nodes
+type ExtRequestPrebidSChainSChainNode struct {
+	Asi    string          `json:"asi"`
+	Sid    string          `json:"sid"`
+	Rid    string          `json:"rid,omitempty"`
```
bsardo, line 42: What casing should I use for these and the `Hp` field below? I'd rather they be all lower case but i need the first letter to be upper so I can populate via json unmarshal. Given that, I'm guessing this is better than all upper case.

openrtb_ext/request.go
```diff
@@ -12,12 +12,43 @@ type ExtRequest struct {
...
+}
+
+// ExtRequestPrebid defines the contract for bidrequest.ext.prebid.schains
+type ExtRequestPrebidSChain struct {
+	Bidders []string                     `json:"bidders,omitempty"`
+	SChain  ExtRequestPrebidSChainSChain `json:"schain"`
+}
+
+// ExtRequestPrebidSChainSChain defines the contract for bidrequest.ext.prebid.schains[i].schain
+type ExtRequestPrebidSChainSChain struct {
```
bsardo, line 31: The name of these types mimid what is used in the Java version.

bsardo:
Related issue: #1091 

Specification: https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/supplychainobject.md

exchange/utils_test.go
```diff
@@ -135,6 +135,80 @@ func TestCleanOpenRTBRequestsCCPA(t *testing.T) {
...
+		{
+			description:  "No schains in root ext and empty source ext",
+			inSourceExt:  json.RawMessage(``),
+			inExt:        json.RawMessage(`{"prebid":{"schains":[]}}`),
+			outSourceExt: json.RawMessage(``),
+			outExt:       json.RawMessage(`{"prebid":{}}`),
+		},
+		{
+			description:  "Use source schain -- no bidder schain or wildcard schain in ext.prebid.schains",
+			inSourceExt:  json.RawMessage(`{"schain":{"complete":1,"nodes":[{"asi":"example.com","sid":"example1","rid":"ExampleReq1","hp":1}],"ver":"1.0"}}`),
```
bsardo, line 162: My hope is that we can eventually come up with a better way to set raw json inputs and validate raw json outputs since this is pretty hard to read.

exchange/utils.go
```diff
@@ -110,11 +151,59 @@ func splitBidRequest(req *openrtb.BidRequest, impsByBidder map[string][]openrtb.
...
+	} else {
+		selectedSChain = wildCardSChain
+	}
+
+	// set source
+	var source openrtb.Source
+	schain := openrtb_ext.ExtRequestPrebidSChain{
+		SChain: *selectedSChain,
+	}
+	sourceExt, err := json.Marshal(schain)
```
bsardo, line 199: There's a performance optimization opportunity here that I punted on. We could cache each marshalled schain and copy the byte array to `sourceExt` instead of marshalling again. I didn't think this was worth it for a first pass but if others think it is worth exploring now, LMK.

exchange/utils.go
```diff
@@ -85,12 +107,31 @@ func cleanOpenRTBRequests(ctx context.Context,
...
 	requestsByBidder := make(map[openrtb_ext.BidderName]*openrtb.BidRequest, len(impsByBidder))
 	explicitBuyerUIDs, err := extractBuyerUIDs(req.User)
 	if err != nil {
 		return nil, []error{err}
 	}
+
+	var requestExt openrtb_ext.ExtRequest
+	var sChainsByBidder map[string]*openrtb_ext.ExtRequestPrebidSChainSChain
+	if len(req.Ext) > 0 {
+		err := json.Unmarshal(req.Ext, &requestExt)
```
bsardo, line 126: This data is also unmarshaled here: https://github.com/prebid/prebid-server/blob/master/exchange/exchange.go#L118-L122

I'd rather only unmarshal once but things might get a little convoluted without a refactor so I figured I would punt on that for now.

exchange/utils_test.go
```diff
@@ -135,6 +135,80 @@ func TestCleanOpenRTBRequestsCCPA(t *testing.T) {
...
+			inExt:        json.RawMessage(``),
+			outSourceExt: json.RawMessage(``),
+			outExt: json.RawMessage(``),
+		},
+		{
+			description:  "No schains in root ext and empty source ext",
+			inSourceExt:  json.RawMessage(``),
+			inExt:        json.RawMessage(`{"prebid":{"schains":[]}}`),
+			outSourceExt: json.RawMessage(``),
+			outExt:       json.RawMessage(`{"prebid":{}}`),
```
bsardo, line 158: I'm not sure if the expected behavior should be to return an empty `prebid` or if the key should be deleted. I think deleting the key makes more sense but I didn't go down that path because of performance concerns when performing a deep equal comparison using the reflect package to check for an empty struct prior to marshaling. I also don't think you can compare the struct that maps to `prebid` to a zero value struct since it has nested structs. I'm looking for some feedback here.

exchange/utils.go
```diff
@@ -17,6 +17,28 @@ import (
...
+	bidderToSChains := make(map[string]*openrtb_ext.ExtRequestPrebidSChainSChain)
+
+	if req.Prebid.SChains == nil || len(req.Prebid.SChains) == 0 {
+		return bidderToSChains
+	}
+
+	for _, schainWrapper := range req.Prebid.SChains {
+		if schainWrapper != nil && len(schainWrapper.Bidders) > 0 {
+			for _, bidder := range schainWrapper.Bidders {
+				if _, present := bidderToSChains[bidder]; present {
```
hhhjort, line 30: This seems strange. If a bidder is in the list, and that bidder also has an entry in `bidderToSChains`, we delete that entry, otherwise we add an entry for the bidder into the map? This would mean if a bidder appears an even number of times, it will be absent from the map. If it appears an odd number of times, it will be in the map with the last `SChain` it was seen with. What is the logic supposed to be here?

exchange/utils_test.go
hhhjort: You could define the strings elsewhere, where you can include newlines and spacing for readability, and then `json.RawMessage` those variables at this line. It does move the JSON further away from the test, so there is a tradeoff.

exchange/utils.go
bsardo: Good catch. I don't see any info in the spec or in the applicable github issues/PRs about how to handle the case where multiple non-wildcard schains are defined for a bidder. The java version appears to take the same approach of ignoring all schains for a bidder if multiple are defined for a bidder, which is where I got the idea from. I'm not a java expert but it looks like the java version has the same bug. I'll confirm that this is the expected behavior and update the logic accordingly.

exchange/utils.go
```diff
@@ -17,6 +17,28 @@ import (
...
+					delete(bidderToSChains, bidder)
+					continue
```
SyntaxNode, line 32: I think it would make more sense here (easier to read IMHO) to use an if / else here than the continue keyword.

exchange/utils.go
```diff
@@ -85,12 +107,31 @@ func cleanOpenRTBRequests(ctx context.Context,
...
+	var requestExt openrtb_ext.ExtRequest
+	var sChainsByBidder map[string]*openrtb_ext.ExtRequestPrebidSChainSChain
+	if len(req.Ext) > 0 {
+		err := json.Unmarshal(req.Ext, &requestExt)
+		if err != nil {
+			return nil, []error{err}
+		}
+		sChainsByBidder = bidderToPrebidSChains(&requestExt)
+	} else {
+		sChainsByBidder = make(map[string]*openrtb_ext.ExtRequestPrebidSChainSChain, 0)
```
SyntaxNode, line 132: I believe the 0 capacity is implicit.

exchange/utils.go
```diff
@@ -110,11 +151,59 @@ func splitBidRequest(req *openrtb.BidRequest, impsByBidder map[string][]openrtb.
...
 }
 
+func prepareExt(req *openrtb.BidRequest, unpackedExt *openrtb_ext.ExtRequest) {
+	if len(req.Ext) == 0 {
+		return
+	}
+	extCopy := *unpackedExt
+	extCopy.Prebid.SChains = nil
+	reqExt, err := json.Marshal(extCopy)
+	if err != nil {
```
SyntaxNode, line 170: This would be more testable if you switch to a check for no error and set the Ext:
```
if err == nil {
  req.Ext = reqExt
}
```

exchange/utils.go
SyntaxNode: Application wide we are guilty of re-marshaling a ton. I agree this is fine for now and part of a much larger performance fix in the (near?) future.

openrtb_ext/request.go
SyntaxNode: We largely follow Go's naming convention. If it's an acronym and needs to be public, the entire name should be capitalized.

exchange/utils.go
bsardo: The team decided that this is a data error that should be detected during request validation preventing further processing.

endpoints/openrtb2/auction.go
```diff
@@ -362,6 +366,11 @@ func validateBidAdjustmentFactors(adjustmentFactors map[string]float64, aliases
 	return nil
 }
 
+func validateSChains(req *openrtb_ext.ExtRequest) error {
+	_, err := exchange.BidderToPrebidSChains(req)
```
hhhjort, line 370: Do we want a simplified version that just checks for duplicate bidders, or perhaps save the result from here for use later?

endpoints/openrtb2/auction.go
```diff
@@ -290,6 +290,10 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) []error {
 		if err := validateBidAdjustmentFactors(bidExt.Prebid.BidAdjustmentFactors, aliases); err != nil {
 			return []error{err}
 		}
+
+		if err := validateSChains(bidExt); err != nil {
+			return []error{err}
+		}
```
guscarreon, line 296: `*openrtb_ext.ExtRequest` also gets unmarshaled and validated in files `endpoints/openrtb2/amp_auction.go` and `endpoints/openrtb2/video_auction.go` we could probably also `validateSChains(bidExt)` from there in order to prevent a defective "ext"JSON field from getting all the way to `splitBidRequests`

bsardo: The `openrtb2/amp_auction.go` and `openrtb2/video_auction.go` endpoints both call the `validateRequest` method in `openrtb2/auction.go` which calls `validateSChains` so we're covered. Those endpoints forgo processing if the method returns an endpoint as well.

guscarreon: Ohh I see

PR #1375, opened 2020-06-30
Fix no bid debug log

camrice:
* Added `omitempty` tags for `AdPods.Targeting`. In cases where the `AdPod` is empty, a new pod is created with the log cache ID.
* Added cache logging in case where final length of `AdPods` is 0
* Added cache logging for case where all bids are rejected and for empty responses
* Critical error cases now log the errors in the cache
* Fixed case where logs were attempted to be cached when no other items would be cached in `doCache`
* Updated error cache logging to be a method for `DebugLog`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -468,7 +468,8 @@ type Cache struct {
 	// adjusts the time dynamically.
 	ExpectedTimeMillis int `mapstructure:"expected_millis"`
 
-	DefaultTTLs DefaultTTLs `mapstructure:"default_ttl_seconds"`
+	DefaultTTLs         DefaultTTLs `mapstructure:"default_ttl_seconds"`
+	DebugDeadlineMillis int         `mapstructure:"debug_deadline_ms"`
 }
 
 // Default TTLs to use to cache bids for different types of imps.
@@ -693,6 +694,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("cache.default_ttl_seconds.video", 0)
 	v.SetDefault("cache.default_ttl_seconds.native", 0)
 	v.SetDefault("cache.default_ttl_seconds.audio", 0)
+	v.SetDefault("cache.debug_deadline_ms", 100)
 	v.SetDefault("external_cache.host", "")
 	v.SetDefault("external_cache.path", "")
 	v.SetDefault("recaptcha_secret", "")
--- a/config/config.go
+++ b/config/config.go
@@ -468,8 +468,7 @@ type Cache struct {
 	// adjusts the time dynamically.
 	ExpectedTimeMillis int `mapstructure:"expected_millis"`
 
-	DefaultTTLs         DefaultTTLs `mapstructure:"default_ttl_seconds"`
-	DebugDeadlineMillis int         `mapstructure:"debug_deadline_ms"`
+	DefaultTTLs DefaultTTLs `mapstructure:"default_ttl_seconds"`
 }
 
 // Default TTLs to use to cache bids for different types of imps.
@@ -694,7 +693,6 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("cache.default_ttl_seconds.video", 0)
 	v.SetDefault("cache.default_ttl_seconds.native", 0)
 	v.SetDefault("cache.default_ttl_seconds.audio", 0)
-	v.SetDefault("cache.debug_deadline_ms", 100)
 	v.SetDefault("external_cache.host", "")
 	v.SetDefault("external_cache.path", "")
 	v.SetDefault("recaptcha_secret", "")
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -122,7 +122,7 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 
 	defer func() {
 		if len(debugLog.CacheKey) > 0 && vo.VideoResponse == nil {
-			err := putDebugLogError(deps.cache, &debugLog, start)
+			err := putDebugLogError(deps.cache, &debugLog)
 			if err != nil {
 				vo.Errors = append(vo.Errors, err)
 			}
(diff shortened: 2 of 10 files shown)


SyntaxNode:
Can you please fix the build failure?
```
endpoints/openrtb2/video_auction_test.go:1228:21: too few values in endpointDeps literal
```

camrice:
Updated to fix unit test failure

endpoints/openrtb2/video_auction.go
```diff
@@ -279,6 +279,21 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		bidResp.Ext = response.Ext
 	}
 
+	if len(bidResp.AdPods) == 0 && debugLog.Enabled {
+		err := debugLog.PutDebugLogError(deps.cache, vo.Errors)
+		if err != nil {
+			vo.Errors = append(vo.Errors, err)
+		} else {
+			bidResp.AdPods = append(bidResp.AdPods, &openrtb_ext.AdPod{
```
mansinahar, line 287: Question: Would we want to return this `AdPod` with debug key only when we have 0 `AdPods` in the `bidResp`? What about the case when we get an error during any of the previous steps like while holding the auction or while building the response? In those cases we don't want to send the `AdPod` with debug key back?

Also, how will the user be able to differentiate between a normal bid response with an actual ad pod vs a bid response with a debug ad pod?

exchange/auction.go
```diff
@@ -47,6 +48,52 @@ func (d *DebugLog) BuildCacheString() {
...
+		{
+			Type:       d.CacheType,
+			Data:       data,
+			TTLSeconds: d.TTL,
+			Key:        "log_" + d.CacheKey,
+		},
+	}
+
+	if cache != nil {
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Duration(100)*time.Millisecond))
```
mansinahar, line 89: Again, I understand that this is not a new change in your PR but curious as to why we're using 100ms as the timeout for the cache request? I mean how did we determine that 100ms is a good value here?

exchange/auction.go
```diff
@@ -179,9 +226,9 @@ func (a *auction) doCache(ctx context.Context, cache prebid_cache_client.Client,
 		}
 	}
 
-	if debugLog != nil && debugLog.Enabled {
-		debugLog.BuildCacheString()
+	if len(toCache) > 0 && debugLog != nil && debugLog.Enabled {
```
mansinahar, line 229: So here we wanna add debug log if there is something to cache? Wouldn't `toCache` greater than 0 mean we had valid bids?

camrice: Yes, so this check is here due to a case where the debug log caching is attempted, but no bids make it to the cache. For example, if there are bids returned but they are all rejected due to not having a category, this caching function would still be called. What was happening was that there was still a call to the cache for the debug log, but it didn't have a proper cache key. So if this check fails, we do the debug log later when we check if there were any adpods generated.

exchange/auction.go
camrice: 100ms was kind of chosen as just a safe starting point until this could be a config value. I will check into adding this.

endpoints/openrtb2/video_auction.go
camrice: For the first question, the error cases should be critical errors, which only return a message rather than a response with pods. So I kept the same error behavior for those failures, but the cache key is appended to the error message for log checking.

For the second question, the differentiator for the debug log `AdPod` is that it will not contain an `hb_pb` and `hb_pb_cat_dur`, and the `podid` will be 0.

exchange/auction.go
```diff
@@ -47,6 +48,52 @@ func (d *DebugLog) BuildCacheString() {
...
+
+	if len(d.CacheKey) == 0 {
+		rawUUID, err := uuid.NewV4()
+		if err != nil {
+			return err
+		}
+		d.CacheKey = rawUUID.String()
+	}
+
+	data, err := json.Marshal(d.CacheString)
```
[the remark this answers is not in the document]
camrice: This marshal here is mainly just for converting the `d.CacheString` to `json.RawMessage` for the cache structure. The only difference I believe is that the resulting string will have quotations, so the cache recognizes it as valid JSON.

exchange/auction.go
camrice: Added a config value for the cache deadline. Let me know if this improves this portion.

exchange/auction.go
mansinahar: So if all the bids were say rejected then basically the above for loop won't execute and which would mean that `toCache` would be of length `0`, no? The current check to me looks like there are bids to be cached but we're still attempting debug log caching which is ok if we wanna do debug log caching as long as debugging is enabled and no matter if there's an error or not.

endpoints/openrtb2/video_auction.go
camrice: The error message usage is meant to cover error cases like a missing field in the request or unmarshalling errors, where we can only return a critical error message. What was missed initially was the case where there weren't any critical errors, but there weren't any bids either. The creating of the debug pod is meant to cover this case, while the error message usage will cover when the request fails for some reason and we need to see the cache ID.

The cache key is appended to the error message in `video_auction.go::handleError()`.

exchange/auction.go
camrice: Yes, so this won't change debug log behavior based off of errors. It is only based off of if we are attempting to cache bids, mostly due to needing the use the same cache key as the bids.

config/config.go
```diff
@@ -441,7 +441,8 @@ type Cache struct {
 	// adjusts the time dynamically.
 	ExpectedTimeMillis int `mapstructure:"expected_millis"`
 
-	DefaultTTLs DefaultTTLs `mapstructure:"default_ttl_seconds"`
+	DefaultTTLs         DefaultTTLs `mapstructure:"default_ttl_seconds"`
+	DebugDeadlineMillis int         `mapstructure:"debug_deadline_ms"`
```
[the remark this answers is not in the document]
camrice: I believe I attempted to use this value before and kept getting timeout errors from the cache call. I think `ExpectedTimeMillis` is only used for the `getAllBids` function if I'm not mistaken.

endpoints/openrtb2/video_auction.go
camrice: The error message usage is meant to cover error cases like a missing field in the request or unmarshalling errors, where we can only return a critical error message. What was missed initially was the case where there weren't any critical errors, but there weren't any bids either. The creating of the debug pod is meant to cover this case, while the error message usage will cover when the request fails for some reason and we need to see the cache ID.

The cache key is appended to the error message in video_auction.go::handleError().

camrice:
Just a general note, I was trying to fix my latest comments but I can't seem to pick the ones I want to reply to. My latest comment is meant to address the two comments on `video_auction.go` and `auction.go`.

camrice:
Just rebased to make sure this branch was up to date. Otherwise no new changes.

PR #1378, opened 2020-07-02
Privacy Request Metrics

SyntaxNode:
Following up on https://github.com/prebid/prebid-server/pull/1360 to extend to CCPA, COPPA, and LMT for requests.


exchange/utils.go
```diff
@@ -75,13 +66,20 @@ func cleanOpenRTBRequests(ctx context.Context,
...
+	privacyLabels.LMTEnforced = privacyEnforcement.LMT
+
 	if gdpr == 1 {
-		cleanMetrics.gdprEnforced = true
+		privacyLabels.GDPREnforced = true
 		parsedConsent, err := vendorconsent.ParseString(consent)
 		if err == nil {
-			cleanMetrics.gdprTcfVersion = int(parsedConsent.Version())
+			version := int(parsedConsent.Version())
+			privacyLabels.GDPRTCFVersion = pbsmetrics.TCFVersionValue(version)
```
guscarreon, line 79: Shouldn't this call refer to the `TCFVersionToValue` function found in `pbsmetrics/metrics.go` instead?
```
280 func TCFVersionToValue(version int) TCFVersionValue {
281     switch {
282     case version == 1:
283         return TCFVersionV1
284     case version == 2:
285         return TCFVersionV2
286     }
287     return TCFVersionErr
288 }
pbsmetrics/metrics.go
```

guscarreon: Despite the fact that this part of the code appears to be covered by test cases, I could not find a test inside the `exchange` directory asserting the contents of the third return value of `cleanOpenRTBRequests`, `privacyLabels pbsmetrics.PrivacyLabels`. Should we include a test to assert this logic?

image

SyntaxNode: You're exactly right. Bug fixed. Tests added.

SyntaxNode:
This PR isn't behaving correctly. Something is out of sync. I'm going to close and recreate.

Issue #1344, opened 2020-06-08
Empty httpcalls array for bidder.


moteus:
I just try to make a request for the Pubnative bidder with test flag on, but got a response like
```JSON
{
  "id": "***",
  "ext": {
    "debug": {
      "httpcalls": {
        "pubnative": [] <------ here empty array.
      },
      "resolvedrequest": {<HERE MY REQUEST>}
      }
    },
    "responsetimemillis": {
      "pubnative": 315
    },
    "tmaxrequest": 800
  }
}
```
I suspect this is because of the pubnative adapter [set the test flag to 0](https://github.com/prebid/prebid-server/blob/master/adapters/pubnative/pubnative.go#L86)

SyntaxNode:
That seems very likely, yes. I'll discuss this with the team early next week. Perhaps we should record the value of the test flag before the adapter does its thing and modifies the request.

bretg:
@fhartwig - would like to understand why your adapter is ignoring/overriding the test flag. In general, the expected behavior for bidders would be to discount impressions when 'test' is specified. If the aim is to avoid test requests to your bidder, I would think your adapter would just invalidate the auction rather than ignoring the explicit signal that this isn't a normal user.

Note that this relates tangentially to https://github.com/prebid/prebid-server/issues/745 -- where we define a separate flag for "debug" vs "test". It's an important distinction:
- test means this isn't a real user. the bidder probably shouldn't charge for the bid.
- debug means it could be a real user, but the publisher wants the extra tracing for confirmation or troubleshooting

@SyntaxNode - is support for the debug flag still on your radar?

fhartwig:
> @fhartwig - would like to understand why your adapter is ignoring/overriding the test flag. In general, the expected behavior for bidders would be to discount impressions when 'test' is specified. If the aim is to avoid test requests to your bidder, I would think your adapter would just invalidate the auction rather than ignoring the explicit signal that this isn't a normal user.

@bretg I don't think there is a good reason for this behaviour, it's probably just something that was left over from the initial development and testing of the adapter. I will prepare a PR to remove this.

PR #1379, opened 2020-07-02
Empty seatBid.httpcalls when adapter modifies request test flag

guscarreon:
Issue #1334 exposed the inability of **Prebid Server** to record `*openrtb_ext.ExtHttpCall` information into the `pbsOrtbSeatBid` objects when a particular adapter's `MakeRequests()` implementation overrides the `openrtb.BidRequest` `Test` flag value. This PR attempts to correct this bug by recording the `Test` flag before  the `MakeRequests()` call in order to avoid adding unnecessary parameters.

### DIFF
--- a/exchange/bidder_test.go
+++ b/exchange/bidder_test.go
@@ -115,6 +115,67 @@ func TestSingleBidder(t *testing.T) {
 }
 
 // TestMultiBidder makes sure all the requests get sent, and the responses processed.
+func TestSingleBidderModifiesTestFlagValue(t *testing.T) {
+	respStatus := 200
+	respBody := "{\"bid\":false}"
+	server := httptest.NewServer(mockHandler(respStatus, "getBody", respBody))
+	defer server.Close()
+
+	requestHeaders := http.Header{}
+	requestHeaders.Add("Content-Type", "application/json")
+
+	bidAdjustment := 2.0
+	firstInitialPrice := 3.0
+	secondInitialPrice := 4.0
+	mockBidderResponse := &adapters.BidderResponse{
+		Bids: []*adapters.TypedBid{
+			{
+				Bid: &openrtb.Bid{
+					Price: firstInitialPrice,
+				},
+				BidType:      openrtb_ext.BidTypeBanner,
+				DealPriority: 4,
+			},
+			{
+				Bid: &openrtb.Bid{
+					Price: secondInitialPrice,
+				},
+				BidType:      openrtb_ext.BidTypeVideo,
+				DealPriority: 5,
+			},
+		},
+	}
+
+	bidderImpl := &requestModifyingBidder{
+		httpRequest: &adapters.RequestData{
+			Method:  "POST",
+			Uri:     server.URL,
+			Body:    []byte("{\"key\":\"val\"}"),
+			Headers: http.Header{},
+		},
+		bidResponse: mockBidderResponse,
+	}
+	bidder := adaptBidder(bidderImpl, server.Client(), &config.Configuration{}, &metricsConfig.DummyMetricsEngine{})
+	currencyConverter := currencies.NewRateConverterDefault()
+
+	// An openrtb.BidRequest with the test flag on that should make requestBid(...) log httpCalls to our return seatBid object
+	bidRequestTestFlagOn := &openrtb.BidRequest{Test: 1}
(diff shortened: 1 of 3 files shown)


mansinahar:
@guscarreon Since the original issue #1344 reported a problem of empty `httpcalls` in the `response.ext.debug` object (rather than the `pbsOrtbSeatBid` object), can we please also add test on a request level to make sure that it is working as expected? This is just to confirm that the `httpcalls` information is getting properly propagated from the `pbsOrtbSeatBid` object to the final response object.

guscarreon:
> @guscarreon Since the original issue #1344 reported a problem of empty `httpcalls` in the `response.ext.debug` object (rather than the `pbsOrtbSeatBid` object), can we please also add test on a request level to make sure that it is working as expected? This is just to confirm that the `httpcalls` information is getting properly propagated from the `pbsOrtbSeatBid` object to the final response object.

Makes sense. Added test case to assert if httpcalls is found in response.ext.debug itself

exchange/bidder_test.go
```diff
@@ -114,6 +114,67 @@ func TestSingleBidder(t *testing.T) {
...
+	server := httptest.NewServer(mockHandler(respStatus, "getBody", respBody))
+	defer server.Close()
+
+	requestHeaders := http.Header{}
+	requestHeaders.Add("Content-Type", "application/json")
+
+	bidAdjustment := 2.0
+	firstInitialPrice := 3.0
+	secondInitialPrice := 4.0
+	mockBidderResponse := &adapters.BidderResponse{
```
mansinahar, line 130: I think we have an opportunity here to extract a simple bid response and use that in this test and other tests rather than creating a new bid response for each test function. This will reduce redundant code for defining similar bid responses

exchange/exchange_test.go
```diff
@@ -140,6 +140,107 @@ func TestCharacterEscape(t *testing.T) {
...
+				ResponseBody: expectedExtHttpCallResponseBody,
+				Status:       expectedExtHttpCallStatus,
+			},
+		},
+	}
+
+	var errList []error
+
+	// Run test
+	outBidResponse, err := e.buildBidResponse(context.Background(),
```
mansinahar, line 209: I am not sure what's the purpose of this test because `buildBidResponse` doesn't really exercise `requestBid` which is where you're making the change. My suggestion about a request level test meant to exercise an entire auction request that would go through `HoldAuction` and for which we can likely utilize the json tests that we have

exchange/bidder_test.go
```diff
@@ -114,6 +114,67 @@ func TestSingleBidder(t *testing.T) {
 	}
 }
 
+// TestMultiBidder makes sure all the requests get sent, and the responses processed.
```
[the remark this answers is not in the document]
guscarreon: This test case is no longer needed in PR #1387

exchange/bidder_test.go
```diff
@@ -114,6 +114,67 @@ func TestSingleBidder(t *testing.T) {
...
+func TestSingleBidderModifiesTestFlagValue(t *testing.T) {
+	respStatus := 200
+	respBody := "{\"bid\":false}"
+	server := httptest.NewServer(mockHandler(respStatus, "getBody", respBody))
```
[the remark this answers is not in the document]
guscarreon: This test case is no longer needed in PR #1387

exchange/bidder_test.go
guscarreon: This test case is no longer needed in PR #1387

exchange/bidder_test.go
```diff
@@ -114,6 +114,67 @@ func TestSingleBidder(t *testing.T) {
...
+	}
+	bidder := adaptBidder(bidderImpl, server.Client(), &config.Configuration{}, &metricsConfig.DummyMetricsEngine{})
+	currencyConverter := currencies.NewRateConverterDefault()
+
+	// An openrtb.BidRequest with the test flag on that should make requestBid(...) log httpCalls to our return seatBid object
+	bidRequestTestFlagOn := &openrtb.BidRequest{Test: 1}
+	seatBid, errs := bidder.requestBid(context.Background(), bidRequestTestFlagOn, "test", bidAdjustment, currencyConverter.Rates(), &adapters.ExtraRequestInfo{})
+
+	// Assertions
+	if len(errs) != 0 {
```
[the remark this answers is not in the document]
guscarreon: This test case is no longer needed in PR #1387

exchange/exchange_test.go
guscarreon: You are right, PR 1387 corrects this test case and runs `exchange.HoldAuction(...)` instead

PR #1381, opened 2020-07-03
user.ext.eids support in adform adapter

braizhas:


adapters/adform/adform.go
```diff
@@ -513,9 +519,30 @@ func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []erro
...
+		_, ok := eidsMap[eid.Source]
+		if !ok {
+			eidsMap[eid.Source] = make(map[string][]int)
+		}
+		for _, uid := range eid.Uids {
+			eidsMap[eid.Source][uid.ID] = append(eidsMap[eid.Source][uid.ID], uid.Atype)
+		}
+	}
+
+	eidsString, _ := json.Marshal(eidsMap)
```
SyntaxNode, line 542: Please do not ignore `json.Marshal` errors.

adapters/adform/adform.go
```diff
@@ -513,9 +519,30 @@ func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []erro
 		consent:       consent,
 		digitrust:     digitrust,
 		currency:      requestCurrency,
+		eids:          eids,
 	}, errors
 }
 
+func encodeEids(eids *[]openrtb_ext.ExtUserEid) string {
```
SyntaxNode, line 526: I recommend to keep this simple and pass in the slice directly. Then you don't need to dereference for then nil check or for loop.

PR #1384, opened 2020-07-07
Pubmatic: Support for video duration and primary category

PubMatic-OpenWrap:


adapters/pubmatic/pubmatic.go
```diff
@@ -549,9 +561,21 @@ func (a *PubmaticAdapter) MakeBids(internalRequest *openrtb.BidRequest, external
 	for _, sb := range bidResp.SeatBid {
 		for i := 0; i < len(sb.Bid); i++ {
 			bid := sb.Bid[i]
+			impVideo := &openrtb_ext.ExtBidPrebidVideo{
+				PrimaryCategory: head(bid.Cat),
```
jmaynardxandr, line 565: Hi, this looks good over all, but I think there is some confusion about the PrimaryCategory field. The intended use for this field was if the bidder knows already what ad server the seller uses and what categories are used on that ad server they can set the primary ad server's category value in it directly and bypass the IAB -> AdServer category lookup that is done later on in exchange.go. This and the adoppler.go adapter both look to be interpreting this field as holding the main category among the possible list of categories found in the bid.Cat array. We are considering to either change the behavior for this field to be like these adapters are expecting, or maybe rename this field to something better indicative of the currently intended purpose. For now though putting an IAB category in this field will prevent the mapping from happening and cause the IAB category to be sent to the ad server in the hb_pb_cat_dur key-value. To return a single IAB category that gets remapped properly at present you would need to trim the bid.Cats array to only contain a single category (and leave the PrimaryCategory field blank).

hhhjort: @PubMatic-OpenWrap : Have you seen Jason's feedback?

PubMatic-OpenWrap: Hi @jmaynardxandr , @hhhjort
I have added code changes for the same, please review.
Thanks

adapters/pubmatic/pubmatic.go
```diff
@@ -48,6 +47,15 @@ type pubmaticParams struct {
 	Keywords    map[string]string `json:"keywords,omitempty"`
 }
 
+type pubmaticBidExtVideo struct {
+	Duration *int `json:"duration,omitempty"`
+}
+
+type pubmaticBidExt struct {
+	BidType           *int                 `json:"BidType,omitempty"`
```
SyntaxNode, line 55: Is `BidType` the correct capitalization of this json field, or is this a copy/paste error?

adapters/pubmatic/pubmatic.go
```diff
@@ -549,9 +561,24 @@ func (a *PubmaticAdapter) MakeBids(internalRequest *openrtb.BidRequest, external
 	for _, sb := range bidResp.SeatBid {
 		for i := 0; i < len(sb.Bid); i++ {
 			bid := sb.Bid[i]
+			impVideo := &openrtb_ext.ExtBidPrebidVideo{}
+
+			if len(bid.Cat) > 1 {
+				bid.Cat = []string{bid.Cat[0]}
```
SyntaxNode, line 567: Consider using `bid.Cat = bid.Cat[0:1]` to avoid instantiating a new array.

adapters/pubmatic/pubmatic.go
PubMatic-OpenWrap: Yes, this field is coming like this only in the pubmatic response, we are going to change it in future.

adapters/pubmatic/pubmatic.go
PubMatic-OpenWrap: Added changes for the same

PR #1390, opened 2020-07-10
Timeout fix

hhhjort:
This is a slight refactor to allow for some deeper tests of the notification timeout system. Putting up to get comments on if we want to merge this in to the code base.


exchange/bidder_test.go
```diff
@@ -1254,15 +1256,28 @@ func TestTimeoutNotificationOff(t *testing.T) {
 	if tb, ok := bidder.Bidder.(adapters.TimeoutBidder); !ok {
 		t.Error("Failed to cast bidder to a TimeoutBidder")
 	} else {
-		bidder.doTimeoutNotification(tb, &adapters.RequestData{})
+		bidder.doTimeoutNotification(tb, &adapters.RequestData{}, glog.Warningf)
 	}
 }
 
 func TestTimeoutNotificationOn(t *testing.T) {
+	ctx, cancelFunc := context.WithDeadline(context.Background(), time.Now().Add(-7*time.Hour))
```
SyntaxNode, line 1264: Does the time matter since we're cancelling the context immediately? If so, consider using just `time.Now()` or perhaps `time.Time{}`.

exchange/bidder_test.go
```diff
@@ -1254,15 +1256,28 @@ func TestTimeoutNotificationOff(t *testing.T) {
...
+	cancelFunc()
 	respBody := "{\"bid\":false}"
 	respStatus := 200
-	server := httptest.NewServer(mockHandler(respStatus, "getBody", respBody))
+	server := httptest.NewServer(mockSlowHandler(respStatus, "getBody", respBody))
```
SyntaxNode, line 1268: The slow handler waits 500ms. I think that's too long for a unit test. Ideally we'd inject a mock context to the timeout caller, but since we can't I think using smaller timeouts is a good compromise. Let's use 201ms or 205ms since the method is hardcoded for 200ms.

exchange/bidder_test.go
```diff
@@ -1272,21 +1287,40 @@ func TestTimeoutNotificationOn(t *testing.T) {
 			Headers: http.Header{},
 		},
 	}
+	// bidderImpl := audienceNetwork.NewFacebookBidder("test-platform-id", "test-app-secret")
```
SyntaxNode, line 1290: Consider removing the comment.

exchange/bidder_test.go
```diff
@@ -1272,21 +1287,40 @@ func TestTimeoutNotificationOn(t *testing.T) {
...
+	bidderInfo := adapters.BidderInfo{
+		Status: adapters.StatusActive,
+		Capabilities: &adapters.CapabilitiesInfo{
+			App: &adapters.PlatformInfo{
+				MediaTypes: []openrtb_ext.BidType{openrtb_ext.BidTypeBanner},
+			},
+		},
+	}
+	ibidder := adapters.EnforceBidderInfo(bidderImpl, bidderInfo)
```
SyntaxNode, line 1299: `iBidder` isn't very descriptive. Consider `wrappedBidder`.

exchange/bidder_test.go
```diff
@@ -1272,21 +1287,40 @@ func TestTimeoutNotificationOn(t *testing.T) {
...
+	if b, ok := corebidder.(*adapters.InfoAwareBidder); ok {
+		corebidder = b.Bidder
+	}
+	if _, ok := corebidder.(adapters.TimeoutBidder); !ok {
+		t.Fatal("Failed to cast bidder to a TimeoutBidder")
 	} else {
-		bidder.doTimeoutNotification(tb, &adapters.RequestData{})
+		bidder.doRequestImpl(ctx, &bidRequest, mylogger)
 	}
+	time.Sleep(300 * time.Millisecond)
```
SyntaxNode, line 1320: We can probably lower this to around 210ms.

exchange/exchange_test.go
```diff
@@ -1830,6 +1830,18 @@ func mockHandler(statusCode int, getBody string, postBody string) http.Handler {
 	})
 }
 
+func mockSlowHandler(statusCode int, getBody string, postBody string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(500 * time.Millisecond)
+		w.WriteHeader(statusCode)
+		if r.Method == "GET" {
```
SyntaxNode, line 1837: I don't think the method type is important for the test. Consider accepting a single `body string` and always return it.

exchange/bidder_test.go
```diff
@@ -1272,21 +1287,40 @@ func TestTimeoutNotificationOn(t *testing.T) {
...
 		me: &metricsConfig.DummyMetricsEngine{},
 	}
-	if tb, ok := bidder.Bidder.(adapters.TimeoutBidder); !ok {
-		t.Error("Failed to cast bidder to a TimeoutBidder")
+	var corebidder adapters.Bidder = bidder.Bidder
+	if b, ok := corebidder.(*adapters.InfoAwareBidder); ok {
+		corebidder = b.Bidder
+	}
+	if _, ok := corebidder.(adapters.TimeoutBidder); !ok {
+		t.Fatal("Failed to cast bidder to a TimeoutBidder")
```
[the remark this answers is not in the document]
hhhjort: Yes, no point in continuing if this test fails.

exchange/bidder_test.go
```diff
@@ -1254,39 +1256,83 @@ func TestTimeoutNotificationOff(t *testing.T) {
...
-		bidder.doTimeoutNotification(tb, &adapters.RequestData{})
+
+	// Unwrap To Mimic exchange.go Casting Code
+	var coreBidder adapters.Bidder = bidderAdapter.Bidder
+	if b, ok := coreBidder.(*adapters.InfoAwareBidder); ok {
+		coreBidder = b.Bidder
+	}
+	if _, ok := coreBidder.(adapters.TimeoutBidder); !ok {
+		t.Fatal("Failed to cast bidder to a TimeoutBidder")
+	}
```
guscarreon, line 1303: What if the logic inside `doRequest` changes? this logic would no longer apply, would it?

exchange/bidder.go
```diff
@@ -312,6 +312,10 @@ func makeExt(httpInfo *httpCallInfo) *openrtb_ext.ExtHttpCall {
 // doRequest makes a request, handles the response, and returns the data needed by the
 // Bidder interface.
 func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.RequestData) *httpCallInfo {
+	return bidder.doRequestImpl(ctx, req, glog.Warningf)
+}
+
+func (bidder *bidderAdapter) doRequestImpl(ctx context.Context, req *adapters.RequestData, logger util.LogMsg) *httpCallInfo {
```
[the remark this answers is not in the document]
SyntaxNode: Yes. The purpose of `doRequest` is to inject the logger. There's cleaner ways of doing this, but it requires larger refactoring which we're punting on for this PR.

exchange/bidder_test.go
SyntaxNode: It would not. The entire auction core is in our sights for a refactor soon.

PR #1394, opened 2020-07-13
Refactor rate converter separating scheduler from converter logic to improve testability

bsardo:


currencies/ticker_task_test.go
```diff
@@ -0,0 +1,308 @@
+package currencies_test
```
bsardo, line 1: The tests in this file were copied from `rate_converter_test.go` and slightly modified. I'm still using sleep here. I think it'll be a fair amount of extra ,not straightforward code to mock the ticker and I wasn't sure if it would be a good use of my time right now, especially since I think we should still have some integration tests anyway and these tests serve that purpose. Open to feedback on this of course.

currencies/ticker_task.go
```diff
@@ -0,0 +1,70 @@
...
+	tt.done <- true
+	close(tt.done)
+}
+
+// Start begins periodic fetching at the given interval
+// It triggers a run before beginning the timer if specified
+// It returns a chan in which the number of data updates everytime a new update was done
+func (tt *TickerTask) run() {
+	ticker := time.NewTicker(tt.fetchingInterval)
+	ticksCount := 0
```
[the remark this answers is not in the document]
bsardo: I don't think so. I couldn't figure out why `ticksCount` or the type `NewRateConverterWithNotifier` existed until I saw this test. I figured I might as well continue with this testing approach instead of mocking the ticker.

currencies/rate_converter.go
```diff
@@ -93,6 +89,10 @@ func (rc *RateConverter) fetch() (*Rates, error) {
 		return nil, err
 	}
 
+	if response.StatusCode >= 400 {
```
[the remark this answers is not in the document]
bsardo: Yeah this addresses a bug I found. If the rate converter returned a non-2xx with a well-formed response body, no error would be caught so the request would be considered successful. As a result, if we had successfully retrieved the currency rates prior to the non-2xx's, we would never fall back to constant rates because we would never see the rates as stale.
The old tests checked for 4xx codes but all of the tests had nil response bodies which resulted in an error when either `ioutil.ReadAll` or `json.Unmarshal` was executed, I can't remember which.

exchange/bidder_test.go
```diff
@@ -516,9 +517,19 @@ func TestMultiCurrencies(t *testing.T) {
...
 			&http.Client{},
 			mockedHTTPServer.URL,
-			time.Duration(10)*time.Second,
+			// time.Duration(10)*time.Second,
+			time.Duration(10)*time.Millisecond,
 			time.Duration(24)*time.Hour,
+			clock.NewRealClock(),
 		)
+		time.Sleep(time.Duration(500) * time.Millisecond)
+		currencyConverterTickerTask := currencies.NewTickerTask(time.Duration(10)*time.Second, currencyConverter)
```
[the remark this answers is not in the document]
bsardo: This test is verifying the `requestBid` function when provided with valid rate conversions. We don't need to run the timer here. We can set up our conversions by creating a new rate converter, call the converter method `Run()` once so that the mocked rates are stored in the converter state and then calling `Rates()` on the converter to retrieve the conversions to be passed into `requestBid`.

currencies/rate_converter_test.go
```diff
@@ -27,417 +25,167 @@ func getMockRates() []byte {
...
-	mockedHttpServer := httptest.NewServer(http.HandlerFunc(
-		func(rw http.ResponseWriter, req *http.Request) {
-			calledURLs = append(calledURLs, req.RequestURI)
-			rw.WriteHeader(http.StatusOK)
-			rw.Write([]byte(getMockRates()))
-		}),
-	)
-
-	defer mockedHttpServer.Close()
+type MockClock struct {
```
SyntaxNode, line 28: Consider naming this `FakeClock` since it doesn't have any mocking behavior and change `fakeTime` to just `time`.

currencies/ticker_task_test.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+type MockRunner struct {
+	updateNotifier chan<- int
+	runCount       int
+}
+
+func (mcc *MockRunner) Run() error { return nil }
+func (mcc *MockRunner) Notify()    { mcc.runCount++; mcc.updateNotifier <- mcc.runCount }
```
SyntaxNode, line 23: Let's remove `Notify` and have the `Run` method increase the count for the test to inspect.

currencies/ticker_task_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+func TestStop(t *testing.T) {
+	// Setup:
+	ticks := make(chan int)
+	mockRunner := &MockRunner{updateNotifier: ticks}
+	interval := time.Duration(1) * time.Millisecond
```
SyntaxNode, line 29: Can this be simplified as `1 * time.Millisecond`.

currencies/ticker_task_test.go
```diff
@@ -0,0 +1,306 @@
...
+		func(rw http.ResponseWriter, req *http.Request) {
+			calledURLs = append(calledURLs, req.RequestURI)
+			rw.WriteHeader(http.StatusOK)
+			rw.Write([]byte(getMockRates()))
+		}),
+	)
+
+	// Execute:
+	interval := time.Duration(0)
+	currencyConverter := currencies.NewRateConverter(
```
SyntaxNode, line 73: No need to use the `NewRateConverter` in this test.

currencies/rate_converter.go
```diff
@@ -46,39 +48,25 @@ func NewRateConverterDefault() *RateConverter {
...
 }
 
-// NewRateConverterWithNotifier returns a new RateConverter
-// it allow to pass an update chan in which the number of ticks will be passed after each tick
-// allowing clients to listen on updates
-// Do not use this method
-func NewRateConverterWithNotifier(
+// NewConfiguredRateConverter returns a new RateConverter
+// Do not call this method directly; it is to improve testability only
+func NewConfiguredRateConverter(
```
SyntaxNode, line 53: Do we need the separate constructor? The purpose for the time dependency is to enable unit tests. The unit tests can construct a rate converter normally and then change the time to a mocked version since the tests are part of the same package.. or just omit the constructor entirely and directly set the struct for testing.

util/task/ticker_task.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Stop stops the periodic task but the task runner maintains state
+func (t *TickerTask) Stop() {
+	close(t.done)
+}
+
+// run creates a ticker that ticks at the specified interval. On each tick,
+// the task is executed
+func (t *TickerTask) run() {
```
SyntaxNode, line 42: Consider renaming to `runRecurring` to better differentiate from the `t.runner.Run()` call.

currencies/rate_converter_test.go
```diff
@@ -557,25 +271,39 @@ func TestRatesAreNeverStale(t *testing.T) {
...
+	// Second Update call results in error but rates from a day ago are still valid
+	err2 := currencyConverter.Run()
+	assert.NotNil(t, err2)
+
+	// Verify rates are valid and last update ts is correct
+	assert.Equal(t, expectedRates, currencyConverter.Rates(), "Conversions.Rates weren't the expected ones")
+	assert.Equal(t, initialFakeTime, currencyConverter.LastUpdated(), "LastUpdated should be set")
 }
 
 func TestRace(t *testing.T) {
```
SyntaxNode, line 306: I guess we can leave the logic alone. I don't follow it's design. 50 clients seems like a lot and the random intervals only decrease the change of a race condition and lower the effectiveness of the test. Since order of operations will be impossible to predict here, this looks like a stub for Go's memory race detector to report on unprotected access.

currencies/rate_converter.go
bsardo: That's a good point. I was trying to keep `time` private but you're right, we only added it to enable UTs so I should just make it public and then we can overwrite it in tests.

currencies/rate_converter_test.go
bsardo: Yeah I don't understand it either. I took a look at the issue that resulted in this test but I still hesitated to edit this because I was afraid I would be removing coverage of some edge case I'm not seeing. I'll take a stab at writing a simpler race test given it seems overcomplicated to both of us.

currencies/rate_converter.go
SyntaxNode: You don't need to make it public for the unit tests. Access modifiers in Go are only at the package level and unit tests run within the same package.

bsardo: Yeah I could make it private but I wanted to continue following the practice of testing the module from a `_test` package to ensure we are only touching the API and publicly accessible fields.

util/task/ticker_task_test.go
```diff
@@ -0,0 +1,62 @@
...
+func (mcc *MockRunner) Run() error {
+	mcc.RunCount++
+	return nil
+}
+
+func TestStartWithSingleRun(t *testing.T) {
+	// Setup:
+	runner := &MockRunner{RunCount: 0}
+	interval := 0 * time.Millisecond
+	ticker := task.NewTickerTask(interval, runner)
```
guscarreon, line 24: `TestStop` does make a `ticker.Stop()` call. Are we missing that call here?

util/task/ticker_task_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func TestStartWithPeriodicRun(t *testing.T) {
+	// Setup:
+	runner := &MockRunner{RunCount: 0}
+	interval := 10 * time.Millisecond
+	ticker := task.NewTickerTask(interval, runner)
+
+	// Execute:
+	ticker.Start()
+	time.Sleep(25 * time.Millisecond)
```
guscarreon, line 42: `TestStop` does make a `ticker.Stop()` call. Are we missing that call here?

main.go
```diff
@@ -53,8 +54,10 @@ func loadConfig() (*config.Configuration, error) {
 func serve(revision string, cfg *config.Configuration) error {
 	fetchingInterval := time.Duration(cfg.CurrencyConverter.FetchIntervalSeconds) * time.Second
 	staleRatesThreshold := time.Duration(cfg.CurrencyConverter.StaleRatesSeconds) * time.Second
-	currencyConverter := currencies.NewRateConverter(&http.Client{}, cfg.CurrencyConverter.FetchURL,
-		fetchingInterval, staleRatesThreshold)
+	currencyConverter := currencies.NewRateConverter(&http.Client{}, cfg.CurrencyConverter.FetchURL, fetchingInterval, staleRatesThreshold)
+	
+	currencyConverterTickerTask := task.NewTickerTask(fetchingInterval, currencyConverter)
+	currencyConverterTickerTask.Start()
```
guscarreon, line 60: I'm not sure if its necessary but, should we also add `defer currencyConverterTickerTask.Stop()`?

util/task/ticker_task_test.go
bsardo: We shouldn't need it here because when `Start` is called, the interval is zero so the go routine the instantiates the ticker is never started. Since it never starts, there is no need to set a signal on the `done` channel because we don't need to stop the ticker.
In `TestStop`, we are testing the `Stop` method which is why it is called. Also, in that test the go routine starts and the ticker is instantiated.

currencies/rate_converter.go
```diff
@@ -30,55 +32,23 @@ func NewRateConverter(
...
-	rc.Update()                   // Make sure to populate data before to return the converter
-	go rc.startPeriodicFetching() // Start periodic ticking
-
-	return rc
+// NewRateConverterDefault returns a RateConverter with default values.
+// By default there will be no currencies conversions done.
+// `currencies.ConstantRate` will be used.
+func NewRateConverterDefault() *RateConverter {
+	return NewRateConverter(&http.Client{}, "", time.Duration(0), time.Duration(0))
 }
```
guscarreon, line 52: This function seems to not be used anywhere else but in `_test.go` after your refactor. Do you think we can safely get rid of it?

util/task/ticker_task_test.go
SyntaxNode: This is perhaps a bit of overkill for the tests, but if we think it would prevent issues then you have my support. This brings up a good thought though, are we probably handling graceful shutdowns? When PBS receives a quit signal from the OS, are we stopping the fetchers? If not, that might be a good follow-up PR.

util/task/ticker_task_test.go
bsardo: Adding `Stop()` here makes sense as it would lead to more consistent test results. However, I think we can just combine this test with `TestStop` since testing the stop method is dependent on a task running periodically.

main.go
bsardo: I don't think it's necessary; `Stop()` is just closing a channel which halts the timer. I guess I don't see any risk in an extra tick or two occurring while in the process of a shutdown, especially since the ticker is just being used for rate conversions at this time.

From the golang docs: "Channels aren't like files; you don't usually need to close them. Closing is only necessary when the receiver must be told there are no more values coming, such as to terminate a range loop."

Also deferred calls don't run when the main go routine exits. I think we would need to use sync waitgroups for this sort of thing.

util/task/ticker_task_test.go
bsardo: I can open a story to ensure we are gracefully shutting down.

util/task/ticker_task_test.go
SyntaxNode: I vote to leave the tests separate. `TestStartWithPeriodicRun` is validating an expected run count and adding `Stop()` is just for cleanup. The `TestStop` test has additional wait time to ensure the stop call actually worked. Leaving them separate should point us more clearly to a potential issue.

currencies/rate_converter.go
bsardo: I don't think this function was ever used in production code. I deleted it and replaced the calls with calls to `NewRateConverter`.

util/task/ticker_task_test.go
bsardo: 👍 I'll leave the two tests separate and add `Stop()`.

currencies/rate_converter.go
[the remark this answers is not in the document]
laurb9: Should we not be checking just for == 200 then ?
The other 2xx and 3xx codes do return useful data either, and we would see a unmarshal error instead of the unexpected status code, which is less useful for troubleshooting.

main.go
```diff
@@ -53,8 +54,10 @@ func loadConfig() (*config.Configuration, error) {
 func serve(revision string, cfg *config.Configuration) error {
 	fetchingInterval := time.Duration(cfg.CurrencyConverter.FetchIntervalSeconds) * time.Second
 	staleRatesThreshold := time.Duration(cfg.CurrencyConverter.StaleRatesSeconds) * time.Second
-	currencyConverter := currencies.NewRateConverter(&http.Client{}, cfg.CurrencyConverter.FetchURL,
-		fetchingInterval, staleRatesThreshold)
+	currencyConverter := currencies.NewRateConverter(&http.Client{}, cfg.CurrencyConverter.FetchURL, fetchingInterval, staleRatesThreshold)
```
laurb9, line 57: `fetchingInterval` is no longer used in the currency module, which makes this call a little misleading :)

util/task/ticker_task.go
```diff
@@ -0,0 +1,53 @@
...
+	Run() error
+}
+
+type TickerTask struct {
+	interval time.Duration
+	runner   Runner
+	done     chan struct{}
+}
+
+func NewTickerTask(interval time.Duration, runner Runner) TickerTask {
```
laurb9, line 17: What do you think about taking a function instead of the interface ? 
So an ad-hoc closure could be passed as well, without having to define a type and method for it.

main.go
bsardo: Good point. I left this in because the `GetInfo` method needed it but I should refactor that.

util/task/ticker_task_test.go
laurb9: I have seen the shutdown pattern in fetchers and I was thinking that since those goroutines maintain no external state to gracefully sync, are there any drawbacks to just letting them be terminated ? Their connections are going to be closed when the process ends anyway, and that seems to be the only time this happens. We don't reload configs, where this could be useful.

endpoints/currency_rates.go
```diff
@@ -57,8 +58,8 @@ func newCurrencyRatesInfo(rateConverter rateConverter) currencyRatesInfo {
 }
 
 // NewCurrencyRatesEndpoint returns current currency rates applied by the PBS server.
-func NewCurrencyRatesEndpoint(rateConverter rateConverter) http.HandlerFunc {
-	currencyRateInfo := newCurrencyRatesInfo(rateConverter)
+func NewCurrencyRatesEndpoint(rateConverter rateConverter, rateConverterTickerTask *task.TickerTask) http.HandlerFunc {
```
bsardo, line 61: I was curious if anyone thought I should create a ticker interface and use that here instead of `*task.TickerTask` so `currency_rates_test.go` is not dependent on the task package.

endpoints/currency_rates.go
```diff
@@ -24,7 +25,7 @@ type rateConverter interface {
 }
 
 // newCurrencyRatesInfo creates a new CurrencyRatesInfo instance.
-func newCurrencyRatesInfo(rateConverter rateConverter) currencyRatesInfo {
+func newCurrencyRatesInfo(rateConverter rateConverter, rateConverterTickerTask *task.TickerTask) currencyRatesInfo {
```
SyntaxNode, line 28: This doesn't need the `rateConverterTickerTask` though - right?, just the fetching interval. How about passing the fetching interval directly?

endpoints/currency_rates.go
SyntaxNode: It doesn't even need a `*task.TickerTask` though, does it? Just pass down the fetching interval duration.

endpoints/currency_rates_test.go
```diff
@@ -8,26 +8,32 @@ import (
...
-		description  string
+		inputConverter  rateConverter
+		inputTickerTask *task.TickerTask
+		expectedBody    string
+		expectedCode    int
+		description     string
 	}{
 		{
 			nil,
+			task.NewTickerTask(time.Duration(0), &FakeRunner{}),
```
SyntaxNode, line 30: Yeah, that's a lot of work just to pass down the duration IMHO.

currencies/rate_converter.go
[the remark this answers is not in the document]
SyntaxNode: The 3xx status codes will be handled by the networking library. So it's really then just the 2xx codes which is a concern. I agree it would be strange to get a non-200 code from the rate converter service, but really any 2xx is considered success so I'm fine with accepting that as a valid server response and letting the marshal code below this handle any errors with the content of the response.

util/task/ticker_task_test.go
SyntaxNode: Perhaps it might not be needed. We'll created an internal story to check if there's a need for a graceful shutdown.

PR #1399, opened 2020-07-15
Adtlelligent: add video for InApp support

GeneGenie:
Added Video mediatype for InApp requests for Adtelligent bidder

### DIFF
--- a/static/bidder-info/adtelligent.yaml
+++ b/static/bidder-info/adtelligent.yaml
@@ -4,6 +4,7 @@ capabilities:
   app:
     mediaTypes:
       - banner
+      - video
   site:
     mediaTypes:
       - banner


static/bidder-info/adtelligent.yaml
```diff
@@ -4,6 +4,7 @@ capabilities:
   app:
     mediaTypes:
       - banner
+      - video
```
bsardo, line 7: Looking at the test coverage, it appears as though there is sufficient coverage for video with `exemplary/simple-video.json` handling the case where the response code is 204 with an empty body and `exemplary/media-type-mapping.json` handling the 200 status code case. However, I did notice that we're missing a status code 200 json test for the banner type. `exemplary/simple-banner.json` covers the status code 204 case but it would probably be good to add a 200 test that demonstrates the `MakeBid` normal flow for this bid type. Perhaps this can be added in a separate PR?

SyntaxNode: I vote for a separate PR. That seems beyond the scope of this change. Sound good, @GeneGenie ?

GeneGenie: Yeah, I guess separate PR is the best solution. Thank you for such deep analisys!

Issue #1040, opened 2019-09-20
Accept account argument on AMP requests


bretg:
The one scenario where Prebid Server can't get account ID is AMP. That could become a problem if [stored request IDs aren't globally unique](https://github.com/prebid/prebid-server/issues/1024).

The proposed solution is to have Prebid Server look for the optional `account` parameter on the /amp endpoint, treating it as site.publisher.id if present. 

If the `account` param is empty on the AMP request, it should be ignored.

SyntaxNode:
Implemented in PBS-Go by @danielguedesb.

PR #1403, opened 2020-07-16
Parse Site.Publisher.ID from Amp Auction HTTP Req Query Parameter

danielguedesb:
- Parse Site.Publisher.ID from Amp Auction HTTP Req Query Parameter "account"


bretg:
Nice - thanks - related to issue https://github.com/prebid/prebid-server/issues/1040

PR #1408, opened 2020-07-17
GumGum: adds support for video

susyt:


adapters/gumgum/gumgum.go
```diff
@@ -128,6 +147,26 @@ func preprocess(imp *openrtb.Imp) (string, error) {
...
+	for _, imp := range imps {
+		if imp.ID == impID && imp.Banner != nil {
+			return openrtb_ext.BidTypeBanner
+		}
+	}
+	return openrtb_ext.BidTypeVideo
+}
+
+func validateVideoParams(video *openrtb.Video) (err error) {
+	if video.W != 0 && video.H != 0 && video.MinDuration != 0 && video.MaxDuration != 0 && video.Placement != 0 && video.Linearity != 0 {
```
SyntaxNode, line 160: Do you think it would be easier to read with the logic reversed?
```
if video.W == 0 || video.H == 0 || video.MinDuration == 0 || video.MaxDuration == 0 || video.Placement == 0 || video.Linearity == 0 {
  return &errortypes.BadInput{
    Message: "Invalid or missing video fields",
  }
}
return nil
```

adapters/gumgum/gumgum.go
```diff
@@ -98,12 +109,20 @@ func (g *GumGumAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRe
 
 	for _, sb := range bidResp.SeatBid {
 		for i := range sb.Bid {
+			// TODO: have gumgum server send mediaType in ext
```
SyntaxNode, line 112: Is this something you want to do before merging this PR?

adapters/gumgum/gumgum.go
```diff
@@ -128,6 +147,26 @@ func preprocess(imp *openrtb.Imp) (string, error) {
...
+		}
+	}
+	return openrtb_ext.BidTypeVideo
+}
+
+func validateVideoParams(video *openrtb.Video) (err error) {
+	if video.W != 0 && video.H != 0 && video.MinDuration != 0 && video.MaxDuration != 0 && video.Placement != 0 && video.Linearity != 0 {
+		return
+	} else {
+		return &errortypes.BadInput{
```
SyntaxNode, line 163: Could you please add a json unit test for this error?

adapters/gumgum/gumgum.go
susyt: Yes, that does read better - I'll update it. Thanks!

adapters/gumgum/gumgum.go
susyt: I'd like to but unfortunately the change would take longer than this week to get out :( I plan on creating another PR once we are able to send the mediaType in the ext object

SyntaxNode: Not a problem. The TODO makes it look like something is missing from the adapter. Since the adapter as complete as possible right now (until you add server side support), could you please remove the TODO?

susyt: @SyntaxNode - removed the comment! Is there anything else that needs to be modified?

PR #1410, opened 2020-07-20
Refactor: bring more clarity on how analytics modules are handled

gpolaert:
I'm working on the Pubstack Analytics Module.
I'm not an experimented go developer.

I think have a dedicated struct to push all enabled modules is more explicit, than returning the current `modules` object.
Right now, `modules` is used to provide the implementation of `analytics.PBSAnalyticsModule` AND to store enabled modules, and it can be confusing 

With this struct, other developers will not have to search how and where are stored the enabled modules (if there are not familiar with go)

### DIFF
--- a/analytics/config/config.go
+++ b/analytics/config/config.go
@@ -7,48 +7,53 @@ import (
 	"github.com/prebid/prebid-server/config"
 )
 
+type analyticsModule analytics.PBSAnalyticsModule
+
+type pbsAnalyticsModule struct {
+	enabledModules []analyticsModule
+}
+
 //Modules that need to be logged to need to be initialized here
 func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
-	modules := make(enabledAnalytics, 0)
+
+	instance := &pbsAnalyticsModule{enabledModules: make([]analyticsModule, 0)}
+
 	if len(analytics.File.Filename) > 0 {
 		if mod, err := filesystem.NewFileLogger(analytics.File.Filename); err == nil {
-			modules = append(modules, mod)
+			instance.enabledModules = append(instance.enabledModules, mod)
 		} else {
 			glog.Fatalf("Could not initialize FileLogger for file %v :%v", analytics.File.Filename, err)
 		}
 	}
-	return modules
+	return instance
 }
 
-//Collection of all the correctly configured analytics modules - implements the PBSAnalyticsModule interface
-type enabledAnalytics []analytics.PBSAnalyticsModule
-
-func (ea enabledAnalytics) LogAuctionObject(ao *analytics.AuctionObject) {
-	for _, module := range ea {
+func (pam pbsAnalyticsModule) LogAuctionObject(ao *analytics.AuctionObject) {
+	for _, module := range pam.enabledModules {
 		module.LogAuctionObject(ao)
 	}
 }
 
-func (ea enabledAnalytics) LogVideoObject(vo *analytics.VideoObject) {
-	for _, module := range ea {
+func (pam pbsAnalyticsModule) LogVideoObject(vo *analytics.VideoObject) {
+	for _, module := range pam.enabledModules {
 		module.LogVideoObject(vo)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


analytics/config/config.go
```diff
@@ -7,48 +7,53 @@ import (
 	"github.com/prebid/prebid-server/config"
 )
 
+type analyticsModule analytics.PBSAnalyticsModule
+
+type pbsAnalyticsModule struct {
```
mansinahar, line 12: I am okay with this idea of wrapping the array of modules in a struct but what do you think about naming this struct to be something like `anlyticsModules` and the field to be called `enabled` so that when you access it, it'd be like `analyticsModules.enabled`? 
That would help communicate what it means more clearly IMHO. The name `pbsAnalyticsModule` right now is the same as the interface name `PbsAnalyticsModule` except for the upper case `P` in the front and I found that to be confusing while reading the code.

analytics/config/config.go
```diff
@@ -7,48 +7,53 @@ import (
 	"github.com/prebid/prebid-server/config"
 )
 
+type analyticsModule analytics.PBSAnalyticsModule
```
mansinahar, line 10: Any reason why we need this new type to represent the interface? `analyticsModule` is serving pretty much as an alias to `analytics.PBSAnalyticsModule`

analytics/config/config.go
SyntaxNode: I agree with adding clarity.

In both the current and proposed approach, we're building an array/slice of analytic modules. In the current approach, we're declaring a type alias of `[]analytics.PBSAnalyticsModule`. In the proposed approach, we're creating a new struct with a member of `[]analytics.PBSAnalyticsModule`. This is a minor difference, IMHO.

The current approach is idiomatic Go and common throughout the core libraries and in PBS. For example, BidderName is an alias of string. Perhaps renaming and reorganizing a bit can add of clarity?

Would it be clearer to rename `enabledAnalytics` to something like `Analytics` or `AnalyticsMultiplexor` and move it to the `analytics/core.go` file to sit side-by-side with the `PBSAnalyticsModule`? I also vote for renaming the `PBSAnalyticsModule` to simply `AnalyticsModule`.

gpolaert: Why I've submitted this PR?

First thing, this was hard for me because I didn't practice go for a while. So, I'm a noob here :)

Here are the main "blockers"
- Not able to test easily the number of modules loaded (a cast is required)
- Not able to test the type of loaded modules, adding a type alias for each new module (FileLogger, PubstackModule ...) could be an idea
- Misleading between the interface definition and the current struct. But here, it's more related to how go is working. 
- Hard to test (fatal) when a module can't be loaded
- Hard to test if a module stops working after a while.

For the current struct or the naming, I don't have a strong opinion here.

analytics/config/config.go
gpolaert: I'm using it for the test. Not sure, it's the best approach to check if the module is loaded.
A better solution will be a dedicated type alias for each module in order to check their presence in the enabledAnalytics slice.

analytics/config/config.go
SyntaxNode: Ah, ok. So the essence of what you're asking for is an easier / more obvious way to inspect the slice of enabled analytics. Since the method is returning an interface, we'll always need to resort to casting in the tests. Perhaps another option is to keep the struct and break apart the interface from the list of adapter for easier to understand flow, like:

```
func NewPBSAnalytics(analytics *config.Analytics) analytics.PBSAnalyticsModule {
  return initAnalytics(analytics)
}

func initAnalytics(analytics *config.Analytics) []analytics.PBSAnalyticsModule {
  .. do all the things / test against this method ...
}
```

.. in which case I don't see much of a difference between the struct or the slice alias.

SyntaxNode:
@gpolaert What did you think of my most recent feedback?

analytics/config/config.go
gpolaert: After thinking about it again, I'm not sure that it brings much more clarity
What do you think about closing this issue?

SyntaxNode: Sure. I'll see what I can do in the future to make it a bit cleaner. I'm fine if you close this.

Issue #1402, opened 2020-07-16
Refreshing file-based stored configs


laurb9:
## Refreshable FileFetcher

The current FileFetcher eagerly loads all the configs in memory and never looks back. It does so by recursively scanning the directory, possibly intended to support a hashed path or other file organization. This makes it fast and shareable, but less suited if stored requests need to be updated within the lifetime of the server process, which is our use case.

I am proposing a FileFetcher whose behavior is more consistent with the other Fetchers, as in can retrieve items on-demand, but does not store the data itself. This allows it to work with the Cache like the others, which in turn makes it refreshable in a way that is configurable externally.

### Implementation

If we need to maintain compatibility with existing path structures, the existing FileFetcher is not ideal for this. The recursive directory ingestion means that the 1:1 mapping between the requested key and the resource, needed for on-demand loading, does not exist. We could attempt direct mapping when it exists (flat dir), but this will result in inconsistent behavior.

Alternatively, a separate clean FileFetcher can be implemented with the defined behavior.

(edit) now that I wrote all this, I realize two filefetchers can be confusing so a config mode switch is probably better.

SyntaxNode:
Thank you for the proposal.

The most common use case of the FileFetcher is local testing, which is likely why it was built to only load configurations at startup. I certainly have no issue with it being expanded to also include a lazy loading approach. I don't recall how the Cache wrapper is wired up, but so long as it would apply to the changes proposed, I'm on board.

I agree having two top level FileFetchers can be a bit confusing.

It looks like the FileFetcher currently returns an eagerFetcher. A nice approach could be implement a lazyFetcher which always goes direct to the file system and like you suggested, using a config flag to decide which one to instantiate. Perhaps another option can be to create a less forgiving Cache layer which binds to the first read and never again refreshes - moving that concern out of the FileFetcher entirely.. Again though, I'm not sure where the Cache is wired up and how easy that may be to change.

By "recursively scanning the directory" are you referring to it loading all files in the directory as opposed to searching for a specific file on demand? That's an area of code I haven't looked at in a while. I know there's an option somewhere in there to load a specific stored request... but I don't know how that's expressed in code. Do you know where that is?

> (edit) now that I wrote all this, I realize two filefetchers can be confusing so a config mode switch is probably better.

+1

Are you planning to implement this proposal?

laurb9:
[...]

PR #1411, opened 2020-07-21
Lazy FileFetcher (#1402)

laurb9:
Replaces the eager file fetcher for stored requests and categories with a lazy one.
This allows the cache layer to work and control refresh intervals for stored requests, but not for categories yet.

This configuration now works to cache files with the given settings. Before the change, it would cache data that is already in memory.
```yaml
stored_requests:
  filesystem: true
  directorypath: "./stored_requests/data/by_id"
  in_memory_cache:
    type: lru
    ttl_seconds: 10
    request_cache_size_bytes: 1000
    imp_cache_size_bytes: 5000
```

### DIFF
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -17,6 +17,10 @@ type StoredRequests struct {
 	Files bool `mapstructure:"filesystem"`
 	//If data should be loaded from file system, path should be specified in configuration
 	Path string `mapstructure:"directorypath"`
+	// Lazy-load the data.
+	//   false (default): preload all the files in cache
+	//   true: load data only when incoming requests ask for the stored resource
+	FilesLazyLoad bool `mapstructure:"filesystem_lazy_load"`
 	// Postgres configures Fetchers and EventProducers which read from a Postgres DB.
 	// Fetchers are in stored_requests/backends/db_fetcher/postgres.go
 	// EventProducers are in stored_requests/events/postgres
@@ -105,6 +109,10 @@ type FileFetcherConfig struct {
 	Enabled bool `mapstructure:"enabled"`
 	// Path to the directory this file fetcher gets data from.
 	Path string `mapstructure:"directorypath"`
+	// Lazy-load the data.
+	//   false (default): preload all the files in cache
+	//   true: load data only when incoming requests ask for the stored resource
+	LazyLoad bool `mapstructure:"lazy_load"`
 }
 
 // HTTPFetcherConfigSlim configures a stored_requests/backends/http_fetcher/fetcher.go
@@ -133,6 +141,10 @@ func (cfg *StoredRequests) validate(errs configErrors) configErrors {
 		}
 		if cfg.Postgres.CacheInitialization.Query != "" {
 			errs = append(errs, errors.New("stored_requests.postgres.initialize_caches.query must be empty if stored_requests.in_memory_cache=none"))
+		}
+		if cfg.Files && !cfg.FilesLazyLoad {
+			// auto configure an in-memory static cache to replicate previous behavior (load and store all)
+			cfg.InMemoryCache.Type = "unbounded"
 		}
 	}
 	errs = cfg.InMemoryCache.validate(errs)
--- a/stored_requests/backends/file_fetcher/fetcher.go
+++ b/stored_requests/backends/file_fetcher/fetcher.go
@@ -5,126 +5,107 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"strings"
+	"os"
+	"path"
+	"path/filepath"
 
+	"github.com/golang/glog"
 	"github.com/prebid/prebid-server/stored_requests"
 )
 
-// NewFileFetcher _immediately_ loads stored request data from local files.
-// These are stored in memory for low-latency reads.
-//
-// This expects each file in the directory to be named "{config_id}.json".
-// For example, when asked to fetch the request with ID == "23", it will return the data from "directory/23.json".
+// NewFileFetcher lazy-loads various kinds of objects from the given directory
+// Expected directory structure:
+// - stored_requests/{id}.json for stored requests
+// - stored_imps/{id}.json for stored imps
+// - {adserver}.json non-publisher specific primary adserver categories
+// - {adserver}/{adserver}_{account_id}.json publisher specific categories for primary adserver
 func NewFileFetcher(directory string) (stored_requests.AllFetcher, error) {
-	storedData, err := collectStoredData(directory, FileSystem{make(map[string]FileSystem), make(map[string]json.RawMessage)}, nil)
-	return &eagerFetcher{storedData, nil}, err
-}
-
-type eagerFetcher struct {
-	FileSystem FileSystem
-	Categories map[string]map[string]stored_requests.Category
-}
-
-func (fetcher *eagerFetcher) FetchRequests(ctx context.Context, requestIDs []string, impIDs []string) (map[string]json.RawMessage, map[string]json.RawMessage, []error) {
-	storedRequests := fetcher.FileSystem.Directories["stored_requests"].Files
-	storedImpressions := fetcher.FileSystem.Directories["stored_imps"].Files
-	errs := appendErrors("Request", requestIDs, storedRequests, nil)
-	errs = appendErrors("Imp", impIDs, storedImpressions, errs)
-	return storedRequests, storedImpressions, errs
-}
-
-func (fetcher *eagerFetcher) FetchCategories(ctx context.Context, primaryAdServer, publisherId, iabCategory string) (string, error) {
-	fileName := primaryAdServer
-
-	if len(publisherId) != 0 {
-		fileName = primaryAdServer + "_" + publisherId
-	}
-
-	if fetcher.Categories == nil {
-		fetcher.Categories = make(map[string]map[string]stored_requests.Category)
-	}
-	if data, ok := fetcher.Categories[fileName]; ok {
-		return data[iabCategory].Id, nil
-	}
-
-	if primaryAdServerDir, found := fetcher.FileSystem.Directories[primaryAdServer]; found {
-
-		if file, ok := primaryAdServerDir.Files[fileName]; ok {
-
-			tmp := make(map[string]stored_requests.Category)
-
-			if err := json.Unmarshal(file, &tmp); err != nil {
-				return "", fmt.Errorf("Unable to unmarshal categories for adserver: '%s', publisherId: '%s'", primaryAdServer, publisherId)
-			}
-			fetcher.Categories[fileName] = tmp
-			resultCategory := tmp[iabCategory].Id
-			primaryAdServerDir.Files[fileName] = nil
-
-			if len(resultCategory) == 0 {
-				return "", fmt.Errorf("Unable to find category for adserver '%s', publisherId: '%s', iab category: '%s'", primaryAdServer, publisherId, iabCategory)
-			}
-			return resultCategory, nil
-		} else {
-			return "", fmt.Errorf("Unable to find mapping file for adserver: '%s', publisherId: '%s'", primaryAdServer, publisherId)
-
-		}
-
-	}
-
-	return "", fmt.Errorf("Category '%s' not found for server: '%s', publisherId: '%s'",
-		iabCategory, primaryAdServer, publisherId)
-
-}
-
-type FileSystem struct {
-	Directories map[string]FileSystem
-	Files       map[string]json.RawMessage
-}
-
-func collectStoredData(directory string, fileSystem FileSystem, err error) (FileSystem, error) {
+	_, err := ioutil.ReadDir(directory)
 	if err != nil {
-		return FileSystem{nil, nil}, err
+		return &fileFetcher{}, err
 	}
-	fileInfos, err := ioutil.ReadDir(directory)
-	if err != nil {
-		return FileSystem{nil, nil}, err
-	}
-	data := make(map[string]json.RawMessage)
-
-	for _, fileInfo := range fileInfos {
-		if fileInfo.IsDir() {
-
-			fs := FileSystem{make(map[string]FileSystem), make(map[string]json.RawMessage)}
-			fileSys, innerErr := collectStoredData(directory+"/"+fileInfo.Name(), fs, err)
-			if innerErr != nil {
-				return FileSystem{nil, nil}, innerErr
-			}
-			fileSystem.Directories[fileInfo.Name()] = fileSys
-
-		} else {
-			if strings.HasSuffix(fileInfo.Name(), ".json") { // Skip the .gitignore
-				fileData, err := ioutil.ReadFile(fmt.Sprintf("%s/%s", directory, fileInfo.Name()))
-				if err != nil {
-					return FileSystem{nil, nil}, err
-				}
-				data[strings.TrimSuffix(fileInfo.Name(), ".json")] = json.RawMessage(fileData)
-
+	// read - but don't store - all the files to warm os cache
+	go filepath.Walk(directory, func(path string, info os.FileInfo, err error) error {
+		if err == nil && filepath.Ext(path) == ".json" {
+			if _, err := ioutil.ReadFile(path); err != nil {
+				glog.Warningf("Error reading %s: %v", path, err)
 			}
 		}
-
-	}
-	fileSystem.Files = data
-	return fileSystem, err
+		return nil
+	})
+	return &fileFetcher{
+		StoredRequestsDir: path.Join(directory, "stored_requests"),
+		StoredImpsDir:     path.Join(directory, "stored_imps"),
+		AccountsDir:       path.Join(directory, "accounts"),
+		CategoriesDir:     path.Clean(directory),
+		Categories:        make(map[string]map[string]stored_requests.Category),
+	}, nil
 }
 
-func appendErrors(dataType string, ids []string, data map[string]json.RawMessage, errs []error) []error {
+type fileFetcher struct {
+	StoredRequestsDir string
+	StoredImpsDir     string
+	AccountsDir       string
+	CategoriesDir     string
+	Categories        map[string]map[string]stored_requests.Category
+}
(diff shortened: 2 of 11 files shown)


stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -5,126 +5,107 @@ import (
...
-
-type eagerFetcher struct {
-	FileSystem FileSystem
-	Categories map[string]map[string]stored_requests.Category
+	_, err := ioutil.ReadDir(directory)
+	if err != nil {
+		return &fileFetcher{}, err
+	}
+	// read - but don't store - all the files to warm os cache
+	go filepath.Walk(directory, func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
laurb9: Reading the files causes the OS to store the file data and dir metadata in its file system cache. A subsequent open/read will read from that cache so it will be faster. Each OS has its own version with similar behavior. Linux has the buffers and page cache.

That being said, it is just a nicety to avoid a slow first request with a given stored request, mainly for testing since in real traffic it matters very little, and I can remove it.

laurb9:
> It does change the default behavior and not all hosts may want that. Can you please include an option either at this level or via a caching layer to retain the default "load all at once and cache" behavior with adding an option for on-demand loading?

I have to give it some thought. The plan was to let the in_memory_cache do the storing, to avoid data duplication in memory. Getting the in_memory_cache layer primed by using something similar to the above filepath.Walk code can work but moves some of this directory handling to a module that should not care about it. Scanning and storing in this module makes this less of an improvement...

laurb9:
I should also mention prebid-server currently uses 2 to 4 times the memory for these files due to this prefetching behavior, which is indirectly solved by this as well.

In https://github.com/prebid/prebid-server/blob/master/stored_requests/config/config.go#L119-L123
The amp and auction fetchers are instantiated from the same configuration, and each preload a copy of the same data dir. If the categories happen to be stored in the same root, then all three fetchers will load all the stored reqs and categories, etc.

I've verified this manually by adding a 500M json stored imp and watching the RSS of the process change by 3x that :)

stored_requests/backends/file_fetcher/fetcher.go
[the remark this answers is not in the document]
SyntaxNode: Gotcha. I recommend to either remove it or give the host an option to warm up the cache.

SyntaxNode:
> I have to give it some thought. The plan was to let the in_memory_cache do the storing, to avoid data duplication in memory. Getting the in_memory_cache layer primed by using something similar to the above filepath.Walk code can work but moves some of this directory handling to a module that should not care about it. Scanning and storing in this module makes this less of an improvement...

Sure. The end goal is to keep the same behavior we have today and then let hosts opt-in to new / better behavior.

> I should also mention prebid-server currently uses 2 to 4 times the memory for these files due to this prefetching behavior, which is indirectly solved by this as well.

I didn't realize it did that. Nice to have that addressed.

laurb9:
I implemented the previous behavior by adding a files event producer to populate
the cache instead of just loading everything locally inside the fetcher.

This makes it consistent with the other modules, and able to use the cache. But it
resulted in more code and complexity than I'd liked, so I am less convinced of its
worth.

1. The flag lazy_load=false (default) replicates existing behavior, but it does define
an unbounded cache if no cache was defined, to not force hosts to change configs.

2. With lazy_load=false but another cache, the loaded data will expire as prescribed
and will be reloaded.

3. With lazy_load=true, the stored data is loaded only when requested. Any cache
will work if given (but is not required, to keep in line with how the other configs work).

CC @bsardo if you think this should wait until your refactor.

stored_requests/config/config.go
```diff
@@ -201,6 +194,9 @@ func newFetcher(cfg *config.StoredRequestsSlim, client *http.Client, db *sql.DB)
 
 	if cfg.Files.Enabled {
 		fFetcher := newFilesystem(cfg.Files.Path)
+		if !cfg.Files.LazyLoad {
+			fFetcher.(stored_requests.CategoryFetcher).FetchAllCategories(context.Background())
+		}
```
bsardo, line 199: I think this should live elsewhere so this function always just returns a fetcher. Perhaps this should live in the `newEventProducers` function and we use an event to perform the initial load like the other backends are. This logic can probably be combined with the logic run during polling which doesn't exist for file fetching right now but we should probably add.

stored_requests/events/files/files.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (f *fileEventProducer) Saves() <-chan events.Save {
+	return f.saves
+}
+func (f *fileEventProducer) Invalidations() <-chan events.Invalidation {
+	return f.invalidations
+}
+
+// NewFilesLoader returns an EventProducer preloaded with all the stored reqs+imps
+func NewFilesLoader(cfg config.FileFetcherConfig) events.EventProducer {
```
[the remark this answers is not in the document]
laurb9: Correct, I did not think filesystem polling adds a lot of value, the periodic reloader using tasks will be simpler and achieves a similar purpose. The simple fact that the data can actually be refreshed in *some* way is in my opinion a massive improvement over the static data set.

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -5,126 +5,130 @@ import (
...
+	_, err := ioutil.ReadDir(directory)
+	if err != nil {
+		return &fileFetcher{}, err
+	}
+	return &fileFetcher{
+		StoredRequestsDir: path.Join(directory, "stored_requests"),
+		StoredImpsDir:     path.Join(directory, "stored_imps"),
+		AccountsDir:       path.Join(directory, "accounts"),
+		CategoriesDir:     path.Clean(directory),
+		Categories:        make(map[string]map[string]stored_requests.Category),
```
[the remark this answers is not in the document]
laurb9: Agreed. Could not find a clean way to quickly refactor this, as it needs native cache as opposed to json.

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -5,126 +5,130 @@ import (
...
+	var ids, files []string
+	var err error
+	path := filepath.Join(dir, glob)
+	if files, err = filepath.Glob(path); err != nil {
+		return nil, append(errs, fmt.Errorf("Error scanning for %s in %s: %v", dataType, path, err))
 	}
-	if data, ok := fetcher.Categories[fileName]; ok {
-		return data[iabCategory].Id, nil
+	for _, f := range files {
+		ids = append(ids, filepath.Base(f))
```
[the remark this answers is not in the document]
laurb9: Clearly! 
I must have assumed Base and Ext were counterparts 🤔

stored_requests/fetcher.go
```diff
@@ -23,11 +23,15 @@ type Fetcher interface {
 	//
 	// The returned objects can only be read from. They may not be written to.
 	FetchRequests(ctx context.Context, requestIDs []string, impIDs []string) (requestData map[string]json.RawMessage, impData map[string]json.RawMessage, errs []error)
+	// FetchAllRequests returns comprehensive maps containing all the requests and imps available in this fetcher
+	FetchAllRequests(ctx context.Context) (requestData map[string]json.RawMessage, impData map[string]json.RawMessage, errs []error)
```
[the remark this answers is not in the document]
laurb9: I wanted minimal code impact and took advantage of AllFetcher being defined not as a collection of interfaces but rather as a collection of functions protos that were also used in other interfaces. I am sure once the other refactor happens, a cleaner solution will reveal itself.

I believe the fetcher module should be providing both "get one" and "get all" functionality. It does not need to be exposed in the same interface but similar code being spread out in different modules (fetcher and events) is not ideal.

I think AllFetcher's reason for existence is to simplify the instantiation code, at the cost of adding complexity and obfuscation to the design. I tried to alleviate some of it in #1431 by removing the requirement that cache follows the same interface, when it did not need to. But it is my belief that it has no functional reason to exist.

stored_requests/config/config.go
laurb9: The categories fetcher wants to store the map in itself so creating another fetcher like `fileEvents.NewFilesLoader` does is a no-go, it was hard to find another place that has both the fetcher and its external config. This should be revisited if categories fetcher is refactored.

config/stored_requests.go
```diff
@@ -17,6 +17,10 @@ type StoredRequests struct {
 	Files bool `mapstructure:"filesystem"`
 	//If data should be loaded from file system, path should be specified in configuration
 	Path string `mapstructure:"directorypath"`
+	// Lazy-load the data.
+	//   false (default): preload all the files in cache
+	//   true: load data only when incoming requests ask for the stored resource
+	LazyLoad bool `mapstructure:"lazy_load"`
```
SyntaxNode, line 23: Since this only applies to the files source, please consider renaming to `FilesLazyLoad` with a json name of `filesystem_ lazy_load`. Ideally, we would package the `File`, `Path`, and `LazyLoad` setting to their own group like `Postgres` and `HTTP`, but that would break existing configs. At least this approach was taken in `StoredRequestsSlim`.

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -5,126 +5,133 @@ import (
...
 )
 
-// NewFileFetcher _immediately_ loads stored request data from local files.
-// These are stored in memory for low-latency reads.
-//
-// This expects each file in the directory to be named "{config_id}.json".
-// For example, when asked to fetch the request with ID == "23", it will return the data from "directory/23.json".
+// NewFileFetcher lazy-loads various kinds of objects from the given directory
+// Expected directory structure:
+// - stored_requests/{id}.json for stored requests
```
SyntaxNode, line 17: Compatibility Check: are the directories unchanged in this PR?

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -5,126 +5,133 @@ import (
...
+// - stored_imps/{id}.json for stored imps
+// - {adserver}.json non-publisher specific primary adserver categories
+// - {adserver}/{adserver}_{account_id}.json publisher specific categories for primary adserver
 func NewFileFetcher(directory string) (stored_requests.AllFetcher, error) {
-	storedData, err := collectStoredData(directory, FileSystem{make(map[string]FileSystem), make(map[string]json.RawMessage)}, nil)
-	return &eagerFetcher{storedData, nil}, err
+	_, err := ioutil.ReadDir(directory)
```
SyntaxNode, line 22: Nitpick: Since `err` is short lived, consider writing as:
```
if _, err := ioutil.ReadDir(directory); err != nil {
  return &fileFetcher{}, err
}
```

stored_requests/config/config.go
SyntaxNode: It's non-obvious through the code that this is prefetching the files to keep in a memory store. Perhaps there's a better way of expressing this, or else consider adding a comment.

stored_requests/events/files/files.go
```diff
@@ -0,0 +1,47 @@
...
+	fp := &fileEventProducer{
+		saves:         make(chan events.Save, 1),
```
SyntaxNode, line 28: I've seen this done elsewhere in our code, but why are we using a  buffered channel for the save events?

stored_requests/config/config.go
laurb9: I agree that it's not obvious, I will add a comment.

config/stored_requests.go
laurb9: For the naming I had followed `directorypath` which is also unqualified and is converted to `Files.path`.

Will rename as you suggest, it is more clear.

On this topic, I've opened #1453 which attempts to consolidate the Slim and non-Slim versions of the configuration and provides a migration path so all the stored data configurations can be finally aligned.

stored_requests/backends/file_fetcher/fetcher.go
laurb9: Directories are unchanged.

The directory structure documented here reflects the fact the original fetcher loads both requests and categories from the same root, the differentiation happens by which methods are called.

The original comment left out this information about the structure understood by the fetcher, which is not easy to discover. I can add more verbiage to indicate that the data sets depend on the data type.

stored_requests/events/files/files.go
laurb9: The capacity is needed to preload the data set in the EventProducer before returning, so it is a little closer to the original behavior which also did a blocking preload of all files.

stored_requests/backends/file_fetcher/fetcher.go
SyntaxNode: I think the comment is good as is. Just wanted to verify this is indeed the structure currently in place. As you said, it's not easy to discover. This should help a lot. Thank you.

stored_requests/events/files/files.go
```diff
@@ -0,0 +1,47 @@
...
+		invalidations: make(chan events.Invalidation),
```
mansinahar, line 29: Do we need the `invalidations` channel in the `fileEventProducer`? Doesn't look like we're using it

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -5,126 +5,133 @@ import (
...
 
-			if err := json.Unmarshal(file, &tmp); err != nil {
-				return "", fmt.Errorf("Unable to unmarshal categories for adserver: '%s', publisherId: '%s'", primaryAdServer, publisherId)
-			}
-			fetcher.Categories[fileName] = tmp
-			resultCategory := tmp[iabCategory].Id
-			primaryAdServerDir.Files[fileName] = nil
+// FetchAllRequests returns comprehensive maps containing all the requests and imps in filesystem
+func (fetcher *fileFetcher) FetchAllRequests(ctx context.Context) (requests map[string]json.RawMessage, imps map[string]json.RawMessage, errs []error) {
+	requests, errs = fetchAllObjects(fetcher.StoredRequestsDir, "*.json", "Request")
```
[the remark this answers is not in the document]
laurb9: Good point. I was planning to take a look at this after #1446 which introduces the DataTypes. The bare strings here come from the previous code, but they are in a few other places not touched by this PR (cache, db fetcher etc) and wanted to avoid starting a change from two places.

laurb9:
Integrated feedback and rebased on latest master.

stored_requests/events/files/files.go
laurb9: It was there more for symmetry rather than anything functional. Will remove.

stored_requests/backends/file_fetcher/fetcher_test.go
```diff
@@ -97,63 +104,64 @@ func assertErrorCount(t *testing.T, num int, errs []error) {
...
+
+func TestCategoriesFetcherAllCategories(t *testing.T) {
+	fetcher := newCategoryFetcher(t, "./test/category-mapping")
+	categories, errs := fetcher.FetchAllCategories(nil)
+	assertErrorCount(t, 1, errs)
+	assert.EqualError(t, errs[0], `Unable to unmarshal categories from "test/category-mapping/test/test_broken.json"`)
+	assert.Equalf(t, len(categories), 2, "Expected 2 categories preloaded, got %d", len(categories))
+	assert.Containsf(t, categories, "test/test", "test/test missing from preloaded data set")
+	assert.Containsf(t, categories, "test/test_categories", "test/test_categories missing from preloaded data set")
+	cat, err := fetcher.FetchCategories(nil, "test", "categories", "IAB1-1")
```
mansinahar, line 127: Any reason for incorporating the `FetchCategories` test as well here? These should be covered in the `FetchCategories` specific code down below, no?

laurb9: It was intentional, it is the only way to check that the preloaded categories were actually decoded and work as expected.

SyntaxNode:
I tested the changes with the AMP endpoint.

The default behavior is still a bit different from what we have today with lazy loading disabled. The files do indeed seem to be cached at startup, but if a file isn't cached the file system will still be dynamically checked. The expected behavior is with the lazy loading disabled, the file system is never checked. Perhaps a new cache type can solve the problem in which a cache miss will immediately return and not hit the store.

With lazy loading enabled, I verified that each request went out to the file system as expected. We might want to advise hosts of this behavior and encourage some kind of in-memory cache, even if short lived, to avoid hammering their disks.

laurb9:
> if a file isn't cached the file system will still be dynamically checked. The expected behavior is with the lazy loading disabled, the file system is never checked.

Also, with a manually configured cache other than in-memory, the items loaded would expire and have to be refreshed, even if we bulk loaded them at start.

There are ways around it, but the bottom line is that with the prefetch, this change ballooned from using less code than before to using twice as much, and my confidence that it's worth including in the code base is somewhat shaken.

> With lazy loading enabled, I verified that each request went out to the file system as expected. We might want to advise hosts of this behavior and encourage some kind of in-memory cache, even if short lived, to avoid hammering their disks.

Indeed, one is expected to configure a cache as shown in https://github.com/prebid/prebid-server/pull/1411#issue-454090085 
This PR makes the FileFetcher behavior consistent with the HTTP and DB ones, in that they all can be fronted by that cache.

In fact, the current configuration allows setting cache with a TTL that never works for files, which I'd argue is a little worse.

SyntaxNode:
> Also, with a manually configured cache other than in-memory, the items loaded would expire and have to be refreshed, even if we bulk loaded them at start.

That's a good point. 

> There are ways around it, but the bottom line is that with the prefetch, this change ballooned from using less code than before to using twice as much, and my confidence that it's worth including in the code base is somewhat shaken.

Do you need this change made? I have no problem with proceeding. I rather we fix it right, and I think the approach this PR is heading in is very good.

laurb9:
Thank you. Appreciate all the review work that has gone into this. I would like to focus on #1426 (account configuration) instead, and put this PR on the back-burner or close it.

Planning to keep the simple lazy fetcher internally for the time being and to start looking into setting up an HTTP API.

PR #1414, opened 2020-07-23
Adds keyvalue hb_format support

### LINKED ISSUE #1412: Support `hb_format` for multiformat AMP requests
Prebid.js supports `hb_format` to enable multiformat imps on AMP, see [multi-format](http://prebid.org/adops/setting-up-prebid-multi-format-in-dfp.html). The spec itself already exists in the prebid.js documentation, we just need to implement this on PBS.

In line with other flags we have on key value generation, I propose we add an `includeformat` flag to the `targeting` extension. So the targeting API would look like:
```
{
  "ext": {
    "prebid": {
      "targeting": {
        "pricegranularity": {
          "precision": 2,
          "ranges": [{
            "max": 20.00,
            "increment": 0.10 // This is equivalent to the deprecated "pricegranularity": "medium"
          }]
        },
        "includewinners": false, // Optional param defaulting to true
        "includebidderkeys": false // Optional param defaulting to true
        "includeformat": false //Optional param defaulting to false
      }
    }
  }
}
```
Thus all current requests would not get `hb_format` added to the key values, but requests that need it can add the flag in order to get access to this value.

hhhjort:
Resolves #1412

### DIFF
--- a/docs/endpoints/openrtb2/auction.md
+++ b/docs/endpoints/openrtb2/auction.md
@@ -173,7 +173,7 @@ to set these params on the response at `response.seatbid[i].bid[j].ext.prebid.ta
         },
         "includewinners": false, // Optional param defaulting to true
         "includebidderkeys": false // Optional param defaulting to true
-        "includeformat": false //Optional param defaulting to false
+        "includeformat": false // Optional param defaulting to false
       }
     }
   }
--- a/docs/endpoints/openrtb2/auction.md
+++ b/docs/endpoints/openrtb2/auction.md
@@ -185,7 +185,7 @@ For backwards compatibility the following strings will also be allowed as price 
 
 One of "includewinners" or "includebidderkeys" must be true (both default to true if unset). If both were false, then no targeting keys would be set, which is better configured by omitting targeting altogether.
 
-The parameter "includeformat" is for multiformat support. It will add the key `hb_format` and/or `hb_format_{bidderName}` as per "includewinners" and "includebidderkeys" above.
+The parameter "includeformat" indicates the type of the bid (banner, video, etc) for multiformat requests. It will add the key `hb_format` and/or `hb_format_{bidderName}` as per "includewinners" and "includebidderkeys" above.
 
 MediaType PriceGranularity (PBS-Java only) - when a single OpenRTB request contains multiple impressions with different mediatypes, or a single impression supports multiple formats, the different mediatypes may need different price granularities. If `mediatypepricegranularity` is present, `pricegranularity` would only be used for any mediatypes not specified. 
 
--- a/docs/endpoints/openrtb2/auction.md
+++ b/docs/endpoints/openrtb2/auction.md
@@ -173,6 +173,7 @@ to set these params on the response at `response.seatbid[i].bid[j].ext.prebid.ta
         },
         "includewinners": false, // Optional param defaulting to true
         "includebidderkeys": false // Optional param defaulting to true
+        "includeformat": false //Optional param defaulting to false
       }
     }
   }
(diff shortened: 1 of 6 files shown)


docs/endpoints/openrtb2/auction.md
```diff
@@ -173,6 +173,7 @@ to set these params on the response at `response.seatbid[i].bid[j].ext.prebid.ta
         },
         "includewinners": false, // Optional param defaulting to true
         "includebidderkeys": false // Optional param defaulting to true
+        "includeformat": false //Optional param defaulting to false
```
SyntaxNode, line 176: Nitpick: Please add a space before the comment for consistency.

exchange/targeting_test.go
```diff
@@ -247,3 +247,166 @@ func (f *mockFetcher) GetId(bidder openrtb_ext.BidderName) (string, bool) {
...
+type TargetingTestData struct {
+	Description                string
+	TargetData                 targetData
+	Auction                    auction
+	IsApp                      bool
+	CategoryMapping            map[string]string
+	ExpectedBidTargetsByBidder map[string]map[openrtb_ext.BidderName]map[string]string
+}
+
+func TestSetTargeting(t *testing.T) {
```
SyntaxNode, line 260: Please put test fixture after the test cases.

exchange/targeting_test.go
```diff
@@ -247,3 +247,166 @@ func (f *mockFetcher) GetId(bidder openrtb_ext.BidderName) (string, bool) {
...
+	for _, test := range TargetingTests {
+		auc := &test.Auction
+		// Set rounded prices from the auction data
+		auc.setRoundedPrices(test.TargetData.priceGranularity)
+		winningBids := make(map[string]*pbsOrtbBid)
+		// Set winning bids from the auction data
+		for imp, bidsByBidder := range auc.winningBidsByBidder {
+			for _, bid := range bidsByBidder {
+				if winningBid, ok := winningBids[imp]; ok {
+					if winningBid.bid.Price < bid.bid.Price {
```
SyntaxNode, line 270: Why do we need to filter by price in a unit test?

exchange/targeting_test.go
```diff
@@ -247,3 +247,166 @@ func (f *mockFetcher) GetId(bidder openrtb_ext.BidderName) (string, bool) {
...
+					test.Description,
+					string(bidder),
+					imp)
+			}
+		}
+	}
+
+}
+
+var TargetingTests []TargetingTestData = []TargetingTestData{
```
SyntaxNode, line 296: Some of these tests aren't related to hb_format. Did we have testing gaps before that you're addressing?

exchange/targeting_test.go
hhhjort: What do you mean by "test fixture"?

exchange/targeting_test.go
hhhjort: I believe so. We had no testing on the `setTargeting()` level. I believe the higher level tests would end up covering some cases, but they are rather far removed, so would be a lot of work to confirm which cases are covered at that level. As you can see, `setTargeting()` requires a lot of data to be set up, which is probably why it was not done before. But now with this framework in place we can easily expand it next time we touch targeting or find a specific hole in our testing.

mansinahar: Thanks a lot @hhhjort for adding these tests. They are definitely very helpful. Could we maybe also add some test cases around the other targeting keys like the cache ones and the deal ones? I understand that this PR doesn't make any code changes for those but since we have this framework in place now, I am hoping that it wouldn't be too much of an effort to add that. I will leave this up to you though.

docs/endpoints/openrtb2/auction.md
```diff
@@ -184,6 +185,8 @@ For backwards compatibility the following strings will also be allowed as price
 
 One of "includewinners" or "includebidderkeys" must be true (both default to true if unset). If both were false, then no targeting keys would be set, which is better configured by omitting targeting altogether.
 
+The parameter "includeformat" is for multiformat support. It will add the key `hb_format` and/or `hb_format_{bidderName}` as per "includewinners" and "includebidderkeys" above.
```
mansinahar, line 188: What do you think about modifying this a little bit to say:
```
The parameter "includeformat" indicates the type of the bid (banner, video, etc) for multiformat requests. It will add the key ....
```

hhhjort:
I have worked with @jaiminpanchal27 to insure this works and is inline with what prebid.js is doing.

PR #1418, opened 2020-07-27
New Adprime adapter

Adprime:


Adprime:
@SyntaxNode is it possible to change assignees to proceed with the review, as there are no updates since commits were made?

adapters/adprime/adprime.go
```diff
@@ -0,0 +1,140 @@
...
+
+		var bidderExt adapters.ExtImpBidder
+		if err = json.Unmarshal(reqCopy.Imp[0].Ext, &bidderExt); err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
+		if err = json.Unmarshal(bidderExt.Bidder, &adprimeExt); err != nil {
+			errs = append(errs, err)
+			continue
```
guscarreon, line 50: Given that there's only one parameter in `openrtb_ext.ExtImpAdprime` that we care for (`TagID`), maybe we can optimize a bit using `jsonparser` instead of unmarshalling twice.
```
31   func (a *AdprimeAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
32       var errs []error
33   *--  4 lines: var adprimeExt openrtb_ext.ExtImpAdprime--------------------------------------------------
37
38       reqCopy := *request
39       for _, imp := range request.Imp {
40           reqCopy.Imp = []openrtb.Imp{imp}
41
42 -         var bidderExt adapters.ExtImpBidder
43 -         if err = json.Unmarshal(reqCopy.Imp[0].Ext, &bidderExt); err != nil {
44 -             errs = append(errs, err)
45 -             continue
46 -         }
47 -
48 -         if err = json.Unmarshal(bidderExt.Bidder, &adprimeExt); err != nil {
   +         tagID, err = jsonparser.GetString(reqCopy.Imp[0].Ext, ... , "TagID") // write keys in the ... part
   +         if err != nil {
49               errs = append(errs, err)
50               continue
51           }
52
53 -         reqCopy.Imp[0].TagID = adprimeExt.TagID
   +         reqCopy.Imp[0].TagID = tagID
54
55   *--  5 lines: adapterReq, errors := a.makeRequest(&reqCopy)---------------------------------------------
60       }
61       return adapterRequests, errs
62   }
adapters/adprime/adprime.go
```

Adprime: This one done

adapters/adprime/adprime.go
```diff
@@ -0,0 +1,136 @@
...
+
+	if response.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if response.StatusCode == http.StatusNotFound {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
+		}}
+	}
```
mansinahar, line 94: Would be good to add a check here for HTTP status codes other than `StatusNoContent` and `StatusNotFound`. You can do that by doing something like:
```
		Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
	}}
}
```

adapters/adprime/adprime.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// NewAdprimeBidder Initializes the Bidder
+func NewAdprimeBidder(endpoint string) *AdprimeAdapter {
+	return &AdprimeAdapter{
+		URI: endpoint,
+	}
+}
+
+type adprimeParams struct {
```
mansinahar, line 27: Sorry we missed this in our previous review but looks like since you're using `jsonparser` to parse the `tagID` from the bidder ext, you don't need this struct. I don't see it being referenced anywhere either.

mansinahar:
@Adprime Thanks for the update. It looks good. There's just one minor thing that we missed in out earlier review. If you can address that, this should be good to go.

mansinahar:
@Adprime we merged it in for now as that extra struct won't really hurt anything and so that your adapter can go live this week. Could you please put a small follow-up PR removing that struct? We'd really appreciate it. Thanks!

Adprime:
@mansinahar https://github.com/prebid/prebid-server/pull/1432/  - here is PR for removing unnecessary struct

PR #1419, opened 2020-07-28
Video endpoint bid selection enhancements

VeronikaSolovei9:
Video endpoint bid selection enhancement: return all winning bids from all bidders per impression instead of one overall winning bid

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -470,7 +470,7 @@ func buildVideoResponse(bidresponse *openrtb.BidResponse, podErrors []PodError) 
 			if err := json.Unmarshal(bid.Ext, &tempRespBidExt); err != nil {
 				return nil, err
 			}
-			if tempRespBidExt.Prebid.Targeting[string(openrtb_ext.HbVastCacheKey)] == "" {
+			if tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbVastCacheKey, seatBid.Seat)] == "" {
 				continue
 			}
 
@@ -479,9 +479,9 @@ func buildVideoResponse(bidresponse *openrtb.BidResponse, podErrors []PodError) 
 			podId, _ := strconv.ParseInt(podNum, 0, 64)
 
 			videoTargeting := openrtb_ext.VideoTargeting{
-				HbPb:       tempRespBidExt.Prebid.Targeting[string(openrtb_ext.HbpbConstantKey)],
-				HbPbCatDur: tempRespBidExt.Prebid.Targeting[string(openrtb_ext.HbCategoryDurationKey)],
-				HbCacheID:  tempRespBidExt.Prebid.Targeting[string(openrtb_ext.HbVastCacheKey)],
+				HbPb:       tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbpbConstantKey, seatBid.Seat)],
+				HbPbCatDur: tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbCategoryDurationKey, seatBid.Seat)],
+				HbCacheID:  tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbVastCacheKey, seatBid.Seat)],
 			}
 
 			adPod := findAdPod(podId, adPods)
@@ -517,6 +517,10 @@ func buildVideoResponse(bidresponse *openrtb.BidResponse, podErrors []PodError) 
 	}
 
 	return &openrtb_ext.BidResponseVideo{AdPods: adPods}, nil
+}
+
+func formatTargetingKey(key openrtb_ext.TargetingKey, bidderName string) string {
+	return fmt.Sprintf("%s_%s", string(key), bidderName)
 }
 
 func findAdPod(podInd int64, pods []*openrtb_ext.AdPod) *openrtb_ext.AdPod {
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -520,7 +520,11 @@ func buildVideoResponse(bidresponse *openrtb.BidResponse, podErrors []PodError) 
 }
 
 func formatTargetingKey(key openrtb_ext.TargetingKey, bidderName string) string {
-	return fmt.Sprintf("%s_%s", string(key), bidderName)
+	fullKey := fmt.Sprintf("%s_%s", string(key), bidderName)
+	if len(fullKey) > exchange.MaxKeyLength {
+		return string(fullKey[0:exchange.MaxKeyLength])
+	}
+	return fullKey
 }
 
 func findAdPod(podInd int64, pods []*openrtb_ext.AdPod) *openrtb_ext.AdPod {
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -630,6 +630,7 @@ func createBidExtension(videoRequest *openrtb_ext.BidRequestVideo) ([]byte, erro
 		IncludeWinners:       true,
 		IncludeBrandCategory: inclBrandCat,
 		DurationRangeSec:     durationRangeSec,
+		IncludeBidderKeys:    true,
 	}
 
 	vastXml := openrtb_ext.ExtRequestPrebidCacheVAST{}
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -631,7 +631,6 @@ func createBidExtension(videoRequest *openrtb_ext.BidRequestVideo) ([]byte, erro
 
 	targeting := openrtb_ext.ExtRequestTargeting{
 		PriceGranularity:     priceGranularity,
-		IncludeWinners:       true,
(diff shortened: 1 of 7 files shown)


exchange/auction_test.go
```diff
@@ -245,13 +263,13 @@ func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec) {
...
-	_ = testAuction.doCache(ctx, cache, targData, &specData.BidRequest, 60, &specData.DefaultTTLs, bidCategory, &specData.DebugLog)
+	_ = testAuction.doCache(ctx, cache, targData, &specData.BidRequest, 60, &specData.DefaultTTLs, bidCategory, &specData.DebugLog, isVideo)
 
 	if len(specData.ExpectedCacheables) > len(cache.items) {
 		t.Errorf("%s:  [CACHE_ERROR] Less elements were cached than expected \n", fileDisplayName)
-	} else if len(specData.ExpectedCacheables) < len(cache.items) {
+	} else if len(specData.ExpectedCacheables) < len(cache.items) && !isVideo {
 		t.Errorf("%s:  [CACHE_ERROR] More elements were cached than expected \n", fileDisplayName)
-	} else { // len(specData.ExpectedCacheables) == len(cache.items)
+	} else if len(specData.ExpectedCacheables) == len(cache.items) && isVideo {
```
jmaynardxandr, line 272: I think you can just leave this as an 'else' here (leaving the condition commented out like it was before). In the event that isVideo and more bids than expected were cached it looks to me like the below code will still work (which just checks that all of the expected cachables were indeed cached, and that should still be the case).

exchange/targeting.go
```diff
@@ -33,11 +33,11 @@ type targetData struct {
 // The one exception is the `hb_cache_id` key. Since our APIs explicitly document cache keys to be on a "best effort" basis,
 // it's ok if those stay in the auction. For now, this method implements a very naive cache strategy.
 // In the future, we should implement a more clever retry & backoff strategy to balance the success rate & performance.
-func (targData *targetData) setTargeting(auc *auction, isApp bool, categoryMapping map[string]string) {
+func (targData *targetData) setTargeting(auc *auction, isApp bool, categoryMapping map[string]string, isVideo bool) {
 	for impId, topBidsPerImp := range auc.winningBidsByBidder {
 		overallWinner := auc.winningBids[impId]
 		for bidderName, topBidPerBidder := range topBidsPerImp {
-			isOverallWinner := overallWinner == topBidPerBidder
+			isOverallWinner := overallWinner == topBidPerBidder || isVideo
```
hhhjort, line 40: Won't setting all bids as winning bids cause key collisions between bids from different bidders? Why is setting `includeBidderKeys` to true not a solution to the problem?

VeronikaSolovei9: @hhhjort Hi Hans, just tested it with `includeBidderKeys = true`. In this case every parameter in targeting data has `_biddername` at the end, for example: `hb_pb_appnexus -> "12.00"`, `hb_uuid_appnexus -> "d81d1315-6060-4c42-bfa7-430f26f8d7d8"` instead of `hb_pb` and `hb_uuid`. With this keys we filter out such bids in video endpoint as "loosing" bids, because they don't have `hb_uuid` key: 
`if tempRespBidExt.Prebid.Targeting[string(openrtb_ext.HbVastCacheKey)] == "" {
	continue
}`
Do you mean collision in `hb_pb_cat_dur` keys? We run `applyCategoryMapping` function and deduplicate bids before auction and caching logic.

exchange/auction_test.go
VeronikaSolovei9: Right, fixed.

exchange/targeting.go
hhhjort: Ahh, yes, the keys are segregated by bids initially, which keeps them separate. Still it seems inelegant to push the `isVideo` flag down from the top in order to change the targeting key generation behavior. We should probably consider putting the flag somewhere in the request itself, the reason for the original behavior (only counting the "winning" bids) was that as the number of bidders increased, the demands on cache increased. Our initial estimates were that the cost of caching so many video bids would render the prebid video service uneconomical. So if we have a flag inside the request, we can have it on for requests that only invoke 2-3 bidders, and then turn it off for the larger requests.

hhhjort: Also, filtering out the "loosing" bids is exactly what this PR is trying to turn off, but by adding a workaround rather than removing the filtering. So that seems another count against this approach.

VeronikaSolovei9: I am also not really happy with this flag. We need to distinguish video and non-video requests in exchange somehow and we deal with openRtb Bid request only, so we cannot add any new flags in it. 
We can add this flag in targeting data only, do you want me to process with this? 
Cache usage will increase in this case, true. For bidders limiting - do we want to have max number of bidders in config? 
In case we have bids back from 5 bidders. Should we select 3 bids per imp with better prices and filter out other 2 bids?

VeronikaSolovei9: Update. Me and @jmaynardxandr just had a discussion on this and this is some points we came up with.
1) Modify "applyCategoryMapping", deduplication part of it: 
-sort bids by pod
-for the same pod deduplicate bids by category only, not by `hb_pb_cat_dur`
-instead of removing random bid, calculate bid with better price (or price per second)
2) Implement "altering auction" where we will return only 3 (max bids per imp, value from config) bids per imp and they all will be winning. In case we have 5 bids back per impression - we will return best 3 of them (by price only or by price per second). 
If we have new "alteringAuction" function, we can rely on labels and don't create any new ugly variables: 
`
if labels.RType == pbsmetrics.ReqTypeVideo{
			alteringAuction
		}else{
			auc = newAuction(adapterBids, len(bidRequest.Imp))
		}
`

hhhjort: I think we want to configure the number of bids we want to cache and return, likely in the PBS config rather than the request. We don't know what the proper balance will be between number of bids and cost ... each additional bid gives us a diminishing return, and at some point the cost in caching will outweigh the benefit gained from the extra bid. In addition, this balance point could easily be different for different PBS hosts.

Perhaps a config for the maximum amount of bidding/caching the host wants to do, and an in request parameter from the publisher for the level they would like to see. That would allow for publishers to tune the requests per the performance they see, but not let them dial it up past the level the host is willing to support.

rshivapurkar: @hhhjort Agreed, Eventually we would like to find the right balance between bid caching and the value of choosing it, however in the phase where demand and inventory are just ramping up, this limit would ideally be on high. Later we can make a decision based on data to figure out what the number would look like. For now, the idea is to limit 1 winning bid per imp for each bidder/adapter.  Ideally we need this only for long form video. Other inventory types may not need this. Please suggest, if and how we can apply this to only long-form video.

hhhjort: The straightforward way would be to have the video endpoint request `includebidderkeys` in the request it passes to the openrtb code. Then it can read out the bidder specific versions of the targeting keys (such as `hb_pb_<bidder>` in place of `hb_pb`). Another option is to not have the openrtb susbsystem generate the keys, but instead have the video endpoint read the bids and execute its own code to extract the key values. This might be best if you suspect that video needs will diverge significantly from what the openrtb endpoint does.

VeronikaSolovei9:
@hhhjort Ready for review

endpoints/openrtb2/video_auction.go
```diff
@@ -626,6 +634,7 @@ func createBidExtension(videoRequest *openrtb_ext.BidRequestVideo) ([]byte, erro
 		IncludeWinners:       true,
```
hhhjort, line 634: You may not need `IncludeWinners` now.

exchange/targeting.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
-const maxKeyLength = 20
+const MaxKeyLength = 20
```
hhhjort, line 10: If you want, you can make this a config option rather than being hardcoded. There has been some talk about making this configurable, but no one has gotten around to it yet. Not required in this PR.

hhhjort:
Two comments on things you may change, but otherwise looking very good now. Waiting on your feedback on the two points before I approve.

endpoints/openrtb2/video_auction.go
VeronikaSolovei9: Removed.

exchange/targeting.go
VeronikaSolovei9: @hhhjort I agree this should be a config option. Can we leave it for later, because this PR has first priority now?

VeronikaSolovei9:
Removed 'includeWinners:true' from default video request config, unit tests modified accordingly.
The only thing we need to have in unit tests to test old functionality works the same as before without this option: 

multiImpVideoNoIncludeBidderKeys.json has 
"targetDataIncludeWinners": true,
"targetDataIncludeBidderKeys": false

otherwise doCache test fails due to nil cached data because of this check:
`if !((bids || vast) && (includeBidderKeys || includeWinners)) {
		return nil
	}` 

About moving `MaxKeyLength` to config - can we have it in a separate PR?

exchange/targeting.go
SyntaxNode: We already have a story for this internally. Not required for this PR.

PR #1420, opened 2020-07-28
Tcf2 id support

### LINKED ISSUE #1326: Refine Prebid Server TCF2 user ID support
The committee has refined the Prebid Server support for TCF2 originally defined in issue #1177 

The full document is https://docs.google.com/document/d/1fBRaodKifv1pYsWY3ia-9K96VHUjd8kKvxZlOsozm8E/edit#

This issue tracks the following changes:

1. Remove the previously implemented enforcement for Purpose 4.
2. Replace it with Requirements 19 and 20:

```
19) User IDs must be stripped from a given vendor’s auction when any of these is true:
- GDPR is in-scope and the consent string is empty or malformed
- User has not provided legal basis for any Purpose 2,3,4,5,6,7,8,9, or 10.
User IDs are defined to be: user.buyeruid, user.id, user.ext.eids, user.ext.digitrust, device.ifa, device.macsha1, device.macmd5, device.dpidsha1, device.dpidmd5, device.didsha1, and device.didmd5.

20) If the host company has opted not to enforce a purpose in the local Prebid Server,
that configuration should not be considered in the legal basis determination for stripping
User IDs in the above requirement. i.e. if Prebid Server config says to ignore enforcement
of Purpose 2 or 7, that configuration should not automatically cause User IDs to always
be passed. 
```
3. Update the "basic enforcement" fallback to check for LITransparency for the case of Purpose 2

hhhjort:
This resolves #1326


hhhjort:
Note: after this GDPR enforcement is split into 3 booleans, base and geo as before, and ID as the new feature. I am not 100% sure the base boolean still has a function, but I think it is best to keep the privacy machinery at this more powerful and general state. The alternative is to bake into the base boolean the assumption that the ID question is the only question we care about apart from geo, trace all the code paths to verify that the assumption is correct, and then have any additional work done on gdpr recognize that this assumption is baked in, and potentially have to separate it out again when that assumption is no longer valid.

SyntaxNode:
> Note: after this GDPR enforcement is split into 3 booleans, base and geo as before, and ID as the new feature. I am not 100% sure the base boolean still has a function, but I think it is best to keep the privacy machinery at this more powerful and general state. The alternative is to bake into the base boolean the assumption that the ID question is the only question we care about apart from geo, trace all the code paths to verify that the assumption is correct, and then have any additional work done on gdpr recognize that this assumption is baked in, and potentially have to separate it out again when that assumption is no longer valid.

The "base" is still needed. For example, we still scrub the ip address even if geo and id are both allowed. Is this the intended behavior?

gdpr/impl.go
```diff
@@ -98,41 +98,42 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
 	return false, nil
 }
 
-func (p *permissionsImpl) allowPI(ctx context.Context, vendorID uint16, consent string) (bool, bool, error) {
+func (p *permissionsImpl) allowPI(ctx context.Context, vendorID uint16, consent string) (bool, bool, bool, error) {
```
SyntaxNode, line 101: To confirm, this is for TCF1 which doesn't have the same purpose breakout, so it's all or nothing here.

gdpr/impl.go
```diff
@@ -142,6 +143,12 @@ func (p *permissionsImpl) allowPITCF2(parsedConsent api.VendorConsents, vendor a
 	} else {
 		allowGeo = true
 	}
+	for i := 2; i <= 10; i++ {
+		if p.checkPurpose(consent, vendor, vendorID, tcf1constants.Purpose(i)) {
```
SyntaxNode, line 147: Does `checkPurpose` cover the following requirement:
```
20) If the host company has opted not to enforce a purpose in the local Prebid Server,
that configuration should not be considered in the legal basis determination for stripping
User IDs in the above requirement. i.e. if Prebid Server config says to ignore enforcement
of Purpose 2 or 7, that configuration should not automatically cause User IDs to always
be passed. 
```

hhhjort: The question of enforcing a purpose is checked outside of `checkPurpose`, so at this level we are essentially ignoring the enforcement configs at this point in the code. So yes, we are taking care of point 20.

gdpr/impl.go
hhhjort: Correct

hhhjort: > I didn't see removal of the Purpose 4 check.

Purpose 4 never made it in to the GDPR code. So it doesn't exist to remove.

hhhjort:
> > Note: after this GDPR enforcement is split into 3 booleans, base and geo as before, and ID as the new feature. I am not 100% sure the base boolean still has a function, but I think it is best to keep the privacy machinery at this more powerful and general state. The alternative is to bake into the base boolean the assumption that the ID question is the only question we care about apart from geo, trace all the code paths to verify that the assumption is correct, and then have any additional work done on gdpr recognize that this assumption is baked in, and potentially have to separate it out again when that assumption is no longer valid.
> 
> The "base" is still needed. For example, we still scrub the ip address even if geo and id are both allowed. Is this the intended behavior?

Yes. I had just noticed, when I was working on the tests, that I did not see any behavior linked to the "base" GDPR flag. I pondered doing a more extensive review to verify if there was anything still depending on it, and if not, refactoring it away. But I end up reasoning that even if it no longer had a use, it would still be better to keep it in until we have had some time to verify that no new GDPR logic would need to be implemented. So I did not bother to look closer, and raised the flag that it was intentional that I did not look into removing it.

gdpr/impl.go
SyntaxNode: Ah. Gotcha. It was never part of the original spec. It was added and subsequently removed before we pushed out an update.

PR #1421, opened 2020-07-29
OpenX adapter: pass optional platform (PBID-598)

laurb9:


laurb9:
> Can you please document the new field in `/static/bidder-params/openx.json`?

Will do, thanks for the quick review.

PR #1422, opened 2020-07-29
Validate External Cache Host 

guscarreon:
We recently noticed that PrebidJS adds the `http://` scheme to the `ExternalCache` URL `Host`. This pull request presents changes to validate the external cache host. We want to make sure this configuration value does not include a protocol (http, https). If validation fails the server should not start.


config/config.go
```diff
@@ -128,6 +129,46 @@ func (cfg *AuctionTimeouts) validate(errs configErrors) configErrors {
 	return errs
 }
 
+func (data *ExternalCache) validate(errs configErrors) configErrors {
+	if data.Host == "" {
+		if data.Path != "" {
+			// Throw an error because we have a blank Host nad a non-blank Path
```
SyntaxNode, line 135: Super Nitpick: Mispelled 'and' as 'nad'.

config/config.go
```diff
@@ -128,6 +129,46 @@ func (cfg *AuctionTimeouts) validate(errs configErrors) configErrors {
...
+			errs = append(errs, fmt.Errorf("Could not parse URL with empty Host: %s \n", data.Host+data.Path))
+		}
+		// Blank host and path are valid and will not throw an error
+		return errs
+	}
+
+	// Add "/" suffix to the path if missing
```
SyntaxNode, line 142: Nitpick: Please update the comment. This is actually ensuring a prefix, not a suffix.

config/config.go
```diff
@@ -128,6 +129,46 @@ func (cfg *AuctionTimeouts) validate(errs configErrors) configErrors {
...
+	if data.Path != "" && !strings.HasPrefix(data.Path, "/") && !strings.HasSuffix(data.Host, "/") {
+		data.Path = "/" + data.Path
+	}
+
+	// Add "//" suffix hosts that don't come with a scheme
+	if strings.Index(data.Host, "//") == -1 {
```
SyntaxNode, line 148: Hosts should not include a protocol/scheme or start with a "//". Not a bad idea to reverse the check and return a validation error.

config/config.go
```diff
@@ -128,6 +129,46 @@ func (cfg *AuctionTimeouts) validate(errs configErrors) configErrors {
...
+		data.Host = "//" + data.Host
+	}
+
+	u, err := url.Parse(data.Host + data.Path)
```
SyntaxNode, line 152: This is trying to correct a configuration error. It might be easier to identity incorrect input instead. Using url.parse is a good idea, but consider using it to ensure the host doesn't specify a protocol/scheme.

config/config.go
guscarreon: Corrected

config/config.go
guscarreon: I should really review this comments because they are definitely poorly written. Changed

config/config.go
guscarreon: This is just in preparation to `url.Parse()`. After some trial and error I found out that function `url.Parse()` takes a URL that looks like `//www.example.com/somepath/v1/` and will return:
```
u.Scheme | ""
u.Host   | "www.example.com"
u.Path   | "/somepath/v1"
err      | nil
```
But, and this is the sole reason to add that extra `//`, if the string we pass to `url.Parse()` looks like `www.example.com/somepath/v1/`, it will surprisingly (or maybe not so much) return:
```
u.Scheme | ""
u.Host   | ""
u.Path   | "www.example.com/somepath/v1"
err      | nil
```
Apparently `url.Parse()` needs those two characters (`//`) to delimit the host.

config/config.go
guscarreon: Ok. Let me go back to formula (Sam Raimi Spiderman reference) for this one. Maybe we can discuss it in tomorrow's stand-up.

laurb9: Could we use the IsHost() and IsPath() validation methods from the govalidator package ?

laurb9: > Could we use the IsHost() and IsPath() validation methods from the govalidator package ?

IsHost() fails because the url can contain a port, IsPath() matches Windows paths too. There are premade regexes in the validator package that can be used.

Instead of validating host and path separately, we could go the other way and validate the final composed result, something like this?
```go
u, err := url.Parse("http://" + host + path)
if err == nil && (host != u.Host || path != u.path) {
    err = fmt.Errorf("Invalid external cache config host='%s' path='%s'", host, path)
}
```

guscarreon: Updated and will still use `url.Parse` thank you @laurb9

config/config.go
```diff
@@ -128,6 +130,29 @@ func (cfg *AuctionTimeouts) validate(errs configErrors) configErrors {
 	return errs
 }
 
+func (data *ExternalCache) validate(errs configErrors) configErrors {
+	if !(data.Host == "" && data.Path == "") {
```
SyntaxNode, line 134: It's better to avoid multiple levels of indentation if it can be avoided. IMHO, I think this looks cleaner:
```
if (data.Host == "" && data.Path == "") {
 return nil
}
```

config/config.go
```diff
@@ -128,6 +130,29 @@ func (cfg *AuctionTimeouts) validate(errs configErrors) configErrors {
...
+		// Either external cache host or path not empty, validate.
+		// Return an error if any of these is true
+		if data.Host == "" && data.Path != "" || data.Host != "" && data.Path == "" {
```
SyntaxNode, line 137: Why did you change from returning the first error to returning all errors?

I think this might get confusing. For example, if `strings.ContainsAny(data.Host, "://")` fails then `url.Parse("https://" + data.Host + data.Path)` would also fail.

config/config_test.go
```diff
@@ -14,6 +14,105 @@ import (
...
+			ExternalCache{Host: "", Path: ""},
+			testOut{expectError: false, expectedHost: "", expectedPath: ""},
+		},
+		{
+			"Invalid http at the end",
+			ExternalCache{Host: "www.google.com", Path: "http://"},
+			testOut{expectError: true, expectedHost: "www.google.com", expectedPath: "/http://"},
+		},
+		{
+			"Scheme longer than host",
```
SyntaxNode, line 75: Not sure this is really testing the case of "Scheme longer than host", but rather "Unknown scheme included"

config/config_test.go
```diff
@@ -14,6 +14,105 @@ import (
...
+
+	testCases := []struct {
+		desc string
+		data ExternalCache
+		out  testOut
+	}{
+		{
+			"With http://",
+			ExternalCache{Host: "http://www.google.com", Path: "/path/v1"},
+			testOut{expectError: true, expectedHost: "www.google.com", expectedPath: "/path/v1"},
```
SyntaxNode, line 32: Why do we specify `expectedHost` and `expectedPath` when we except an error? Neither of those are checked. Consider removing in all test cases where `expectError` is `true`.

config/config_test.go
SyntaxNode: Please change this description to "Host has an unknown scheme"

config/config_test.go
```diff
@@ -14,6 +14,105 @@ import (
...
+		errs = test.data.validate(errs)
+
+		if test.out.expectError {
+			assert.Greater(t, len(errs), 0, "Test case was supposed to throw an error. Desc: %s \n", test.desc)
+			continue
+		} else {
+			assert.Equal(t, 0, len(errs), "Test case was supposed to NOT throw an error. Desc: %s errMsg = %v \n", test.desc, errs)
+		}
+
+		assert.Equal(t, test.out.expectedHost, test.data.Host, "Test case host does not match. Desc: %s  \n", test.desc)
```
SyntaxNode, line 111: Consider moving the host and path asserts to the else clause to avoid a continue.

config/config_test.go
guscarreon: They get checked at the end if we don't expect error but with the recent changes in the function they are definitely redundant. Removed

PR #1427, opened 2020-08-03
Enable geo activation of GDPR flag

hhhjort:
Not sure how to add a test for this without a major refactor. Putting this up for comments/review.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -156,8 +156,9 @@ type GDPR struct {
 	Timeouts                GDPRTimeouts `mapstructure:"timeouts_ms"`
 	NonStandardPublishers   []string     `mapstructure:"non_standard_publishers,flow"`
 	NonStandardPublisherMap map[string]int
-	TCF2                    TCF2 `mapstructure:"tcf2"`
-	AMPException            bool `mapstructure:"amp_exception"`
+	TCF2                    TCF2     `mapstructure:"tcf2"`
+	AMPException            bool     `mapstructure:"amp_exception"`
+	CountryList             []string `mapstructure:"country_list"`
 }
 
 func (cfg *GDPR) validate(errs configErrors) configErrors {
@@ -870,6 +871,10 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.enabled", true)
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.access_allowed", true)
 	v.SetDefault("gdpr.amp_exception", false)
+	v.SetDefault("gdpr.country_list", []string{"ALA", "AUT", "BEL", "BGR", "HRV", "CYP", "CZE", "DNK", "EST",
+		"FIN", "FRA", "GUF", "DEU", "GIB", "GRC", "GLP", "GGY", "HUN", "ISL", "IRL", "IMN", "ITA", "JEY", "LVA",
+		"LIE", "LTU", "LUX", "MLT", "MTQ", "MYT", "NLD", "NOR", "POL", "PRT", "REU", "ROU", "BLM", "MAF", "SPM",
+		"SVK", "SVN", "ESP", "SWE", "GBR"})
 	v.SetDefault("ccpa.enforce", false)
 	v.SetDefault("lmt.enforce", true)
 	v.SetDefault("currency_converter.fetch_url", "https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json")
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -49,6 +49,7 @@ type exchange struct {
 	UsersyncIfAmbiguous bool
 	defaultTTLs         config.DefaultTTLs
 	privacyConfig       config.Privacy
+	gdprCountries       map[string]struct{}
 }
 
 // Container to pass out response ext data from the GetAllBids goroutines back into the main thread
@@ -69,6 +70,10 @@ type bidResponseWrapper struct {
 func NewExchange(client *http.Client, cache prebid_cache_client.Client, cfg *config.Configuration, metricsEngine pbsmetrics.MetricsEngine, infos adapters.BidderInfos, gDPR gdpr.Permissions, currencyConverter *currencies.RateConverter) Exchange {
 	e := new(exchange)
 
+	var s struct{}
+	for _, c := range cfg.GDPR.CountryList {
+		e.gdprCountries[c] = s
+	}
 	e.adapterMap = newAdapterMap(client, cfg, infos, metricsEngine)
 	e.cache = cache
 	e.cacheTime = time.Duration(cfg.CacheURL.ExpectedTimeMillis) * time.Millisecond
@@ -102,9 +107,22 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 		e.me.RecordImps(impLabels)
 	}
 
+	// Make our best guess if GDPR applies
+	usersyncIfAmbiguous := e.UsersyncIfAmbiguous
+	var geo *openrtb.Geo = nil
+	if bidRequest.User != nil && bidRequest.User.Geo != nil {
+		geo = bidRequest.User.Geo
+	} else if bidRequest.Device != nil && bidRequest.Device.Geo != nil {
+		geo = bidRequest.Device.Geo
+	}
+	if geo != nil {
+		if _, found := e.gdprCountries[geo.Country]; found {
+			usersyncIfAmbiguous = false
+		}
+	}
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
-	cleanRequests, aliases, privacyLabels, errs := cleanOpenRTBRequests(ctx, bidRequest, usersyncs, blabels, labels, e.gDPR, e.UsersyncIfAmbiguous, e.privacyConfig)
+	cleanRequests, aliases, privacyLabels, errs := cleanOpenRTBRequests(ctx, bidRequest, usersyncs, blabels, labels, e.gDPR, usersyncIfAmbiguous, e.privacyConfig)
 
 	e.me.RecordRequestPrivacy(privacyLabels)
 


config/config.go
```diff
@@ -156,8 +156,9 @@ type GDPR struct {
 	Timeouts                GDPRTimeouts `mapstructure:"timeouts_ms"`
 	NonStandardPublishers   []string     `mapstructure:"non_standard_publishers,flow"`
 	NonStandardPublisherMap map[string]int
-	TCF2                    TCF2 `mapstructure:"tcf2"`
-	AMPException            bool `mapstructure:"amp_exception"`
+	TCF2                    TCF2     `mapstructure:"tcf2"`
+	AMPException            bool     `mapstructure:"amp_exception"`
+	CountryList             []string `mapstructure:"country_list"`
```
SyntaxNode, line 161: Perhaps we can name it ` eea_countries` to match what's already in PBS-Java, or else something more descriptive such as `countries_to_force_enforcement`. Ok, that last one is a bit too long.. but I think you get the idea. "list" isn't descriptive.

exchange/exchange.go
```diff
@@ -102,9 +107,22 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+	// Make our best guess if GDPR applies
+	usersyncIfAmbiguous := e.UsersyncIfAmbiguous
+	var geo *openrtb.Geo = nil
+	if bidRequest.User != nil && bidRequest.User.Geo != nil {
+		geo = bidRequest.User.Geo
+	} else if bidRequest.Device != nil && bidRequest.Device.Geo != nil {
+		geo = bidRequest.Device.Geo
+	}
+	if geo != nil {
+		if _, found := e.gdprCountries[geo.Country]; found {
```
SyntaxNode, line 119: I looks like `UsersyncIfAmbiguous` must be enabled for the Geo check to matter, otherwise it's always false. Should it instead be "if you're an EU country, enforce GDPR... otherwise fall back to UsersyncIfAmbiguous". This would better handle the case of a EU request somehow being routed to a server outside of the EU.

hhhjort:
If I create a JSON test around `HoldAuction()`, what can I test in the output to determine if geo was enforced or not? I think I need a major refactor before I can create any test to test this.

exchange/exchange.go
hhhjort: It is confusing, `UsersyncIfAmbiguous` is the opposite of enforcing the GDPR by default. If if is false, we always assume gdpr applies if no signal. So if we always assume gdpr, then it doesn't make sense to check if you are in an eu country. I should put this in an if statement to save a few cycles if we are always assuming gdpr is active.

SyntaxNode:
> If I create a JSON test around `HoldAuction()`, what can I test in the output to determine if geo was enforced or not? I think I need a major refactor before I can create any test to test this.

I thought you could test this in a manner similar to `ccpa-featureflag-on.json`. I needed to better control the configuration for that test, so I added fields to `exchangeSpec` and wired it up in `runSpec`.

The test framework is comparing expectations against outgoing adapter requests. Specifically, I test for removal of the BuyerID and I think that would work well in this case.

You'll need to control the `UsersyncIfAmbiguous` value which is currently hardcoded to false in `newExchangeForTests`. The GDPR config is already passed in as an argument.

It might be helpful to extract the logic to a method which can be tested with all edge cases separately using table driven tests, and then use this approach for the main happy and sad paths.

exchange/exchange_test.go
```diff
@@ -1019,15 +1022,25 @@ func newExchangeForTests(t *testing.T, filename string, expectations map[string]
 		}
 	}
 
+	var s struct{}
+	eeac := make(map[string]struct{})
+	for _, c := range []string{"ALA", "AUT", "BEL", "BGR", "HRV", "CYP", "CZE", "DNK", "EST",
```
SyntaxNode, line 1027: Do we need such a long list for the tests? Can we simplify this?

config/config.go
```diff
@@ -894,6 +895,10 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.enabled", true)
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.access_allowed", true)
 	v.SetDefault("gdpr.amp_exception", false)
+	v.SetDefault("gdpr.eea_countries", []string{"ALA", "AUT", "BEL", "BGR", "HRV", "CYP", "CZE", "DNK", "EST",
```
SyntaxNode, line 898: Where did you get this list from?

exchange/exchangetest/gdpr-geo-usa.json
```diff
@@ -0,0 +1,61 @@
+{
```
SyntaxNode, line 1: Consider renaming to `gdpr-geo-usa-off` to mirror `gdpr-geo-usa-on`.

exchange/exchangetest/gdpr-geo-eu.json
```diff
@@ -0,0 +1,60 @@
+{
```
SyntaxNode, line 1: Consider renaming to `gdpr-geo-eu-off` to mirror `gdpr-geo-eu-on`.

exchange/exchange.go
```diff
@@ -119,9 +124,25 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 		e.me.RecordImps(impLabels)
 	}
 
+	// Make our best guess if GDPR applies
+	usersyncIfAmbiguous := e.UsersyncIfAmbiguous
+	var geo *openrtb.Geo = nil
+	if bidRequest.User != nil && bidRequest.User.Geo != nil {
+		geo = bidRequest.User.Geo
+	} else if bidRequest.Device != nil && bidRequest.Device.Geo != nil {
+		geo = bidRequest.Device.Geo
```
SyntaxNode, line 133: The device branch is untested. Can you please add coverage?

config/config.go
```diff
@@ -156,8 +156,9 @@ type GDPR struct {
 	Timeouts                GDPRTimeouts `mapstructure:"timeouts_ms"`
 	NonStandardPublishers   []string     `mapstructure:"non_standard_publishers,flow"`
 	NonStandardPublisherMap map[string]int
-	TCF2                    TCF2 `mapstructure:"tcf2"`
-	AMPException            bool `mapstructure:"amp_exception"`
+	TCF2                    TCF2     `mapstructure:"tcf2"`
+	AMPException            bool     `mapstructure:"amp_exception"`
+	EEACountries            []string `mapstructure:"eea_countries"`
```
SyntaxNode, line 161: Please add a comment to clarify that EEA = European Economic Area and explain we will always enforce GDPR for requests with a geo which match this list.

exchange/exchange.go
```diff
@@ -119,9 +124,25 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+	}
+	if geo != nil {
+		if _, found := e.eeaCountries[geo.Country]; found {
+			usersyncIfAmbiguous = false
```
SyntaxNode, line 137: Please add a comment indicating that gdpr is always enforced if a country is in this list.

exchange/exchange.go
```diff
@@ -119,9 +124,25 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+		} else if len(geo.Country) == 3 {
+			// The country field is formatted properly as a three character country code
```
SyntaxNode, line 139: Please add a comment indicating that gdpr is never enforced if a country code is valid (3 characters) and not in the eea list.

exchange/exchange.go
```diff
@@ -119,9 +124,25 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+	// Make our best guess if GDPR applies
+	usersyncIfAmbiguous := e.UsersyncIfAmbiguous
+	var geo *openrtb.Geo = nil
+	if bidRequest.User != nil && bidRequest.User.Geo != nil {
+		geo = bidRequest.User.Geo
+	} else if bidRequest.Device != nil && bidRequest.Device.Geo != nil {
+		geo = bidRequest.Device.Geo
+	}
+	if geo != nil {
+		if _, found := e.eeaCountries[geo.Country]; found {
```
SyntaxNode, line 136: Is capitalization a concern? Should we convert the geo country to all caps before checking? .. or add both caps and lower case countries to the list?

exchange/exchange_test.go
hhhjort: I just copied what we had for the default. I am torn between having a simple setup, and having a setup that more resembles what would be seen in production.

exchange/exchangetest/gdpr-geo-usa.json
hhhjort: Done

exchange/exchangetest/gdpr-geo-eu.json
hhhjort: Done

config/config.go
hhhjort: done

exchange/exchange.go
hhhjort: done

exchange/exchange.go
hhhjort: done

exchange/exchange.go
hhhjort: That is a good question.

SyntaxNode: I escalated this question to legal. I think we're fine to leave it as upper case for now as that's whats listed in the ISO-3166-1-alpha-3 spec.

exchange/exchange_test.go
SyntaxNode: I have a stronger preference to keep things simple for the test. Makes it a bit easier to read and functionality the test is just as valid if there are 1 or 2 entries as 20+.

config/config.go
```diff
@@ -159,6 +159,11 @@ type GDPR struct {
 	TCF1                    TCF1 `mapstructure:"tcf1"`
 	TCF2                    TCF2 `mapstructure:"tcf2"`
 	AMPException            bool `mapstructure:"amp_exception"`
+	// EEACountries (EEA = European Economic Area) are a list of countries where we should assume GDPR applies.
```
SyntaxNode, line 162: Perfect! Thank you.

config/config.go
SyntaxNode: It was provided by our legal team. 👍

config/config.go
```diff
@@ -903,6 +908,10 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.enabled", true)
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.access_allowed", true)
 	v.SetDefault("gdpr.amp_exception", false)
+	v.SetDefault("gdpr.eea_countries", []string{"ALA", "AUT", "BEL", "BGR", "HRV", "CYP", "CZE", "DNK", "EST",
+		"FIN", "FRA", "GUF", "DEU", "GIB", "GRC", "GLP", "GGY", "HUN", "ISL", "IRL", "IMN", "ITA", "JEY", "LVA",
+		"LIE", "LTU", "LUX", "MLT", "MTQ", "MYT", "NLD", "NOR", "POL", "PRT", "REU", "ROU", "BLM", "MAF", "SPM",
+		"SVK", "SVN", "ESP", "SWE", "GBR"})
```
guscarreon, line 914: In a future PR we might want to have a test in `config/config_test.go` for to assert the country list parsing from the config file

exchange/exchange.go
```diff
@@ -55,6 +55,7 @@ type exchange struct {
 	UsersyncIfAmbiguous bool
 	defaultTTLs         config.DefaultTTLs
 	privacyConfig       config.Privacy
+	eeaCountries        map[string]struct{}
```
guscarreon, line 58: In a future PR, maybe we can refactor to have all GDPR related fields inside some privacy wrapper in order to simplify the `exchange` struct

exchange/exchange.go
```diff
@@ -75,6 +76,10 @@ type bidResponseWrapper struct {
 func NewExchange(client *http.Client, cache prebid_cache_client.Client, cfg *config.Configuration, metricsEngine pbsmetrics.MetricsEngine, infos adapters.BidderInfos, gDPR gdpr.Permissions, currencyConverter *currencies.RateConverter) Exchange {
 	e := new(exchange)
 
+	var s struct{}
+	for _, c := range cfg.GDPR.EEACountries {
+		e.eeaCountries[c] = s
+	}
```
guscarreon, line 82: inside `config/config.go` we build the `NonStandardPublisherMap` and `BlacklistedAppMap` in a similar fashion, maybe we can refactor in the future to set up the map in the config file instead.
```
153   type GDPR struct {
154       HostVendorID            int          `mapstructure:"host_vendor_id"`
155       UsersyncIfAmbiguous     bool         `mapstructure:"usersync_if_ambiguous"`
156       Timeouts                GDPRTimeouts `mapstructure:"timeouts_ms"`
157       NonStandardPublishers   []string     `mapstructure:"non_standard_publishers,flow"`
158       NonStandardPublisherMap map[string]int
159       TCF1                    TCF1 `mapstructure:"tcf1"`
160       TCF2                    TCF2 `mapstructure:"tcf2"`
161       AMPException            bool `mapstructure:"amp_exception"`
162       // EEACountries (EEA = European Economic Area) are a list of countries where we should assume GDPR applies.
163       // If the gdpr flag is unset in a request, but geo.country is set, we will assume GDPR applies if and only
164       // if the country matches one on this list. If both the GDPR flag and country are not set, we default
165       // to UsersyncIfAmbiguous
166       EEACountries []string `mapstructure:"eea_countries"`
    +     EEACountryMap map[string]struct{}
167   }
168   *--360 lines: func (cfg *GDPR) validate(errs configErrors) configErrors {------------------------------------
528   // New uses viper to get our server configurations.
529   func New(v *viper.Viper) (*Configuration, error) {
530       var c Configuration
531       if err := v.Unmarshal(&c); err != nil {
532           return nil, fmt.Errorf("viper failed to unmarshal app config: %v", err)
533       }
534       c.setDerivedDefaults()
535  
536   *--  8 lines: if err := c.RequestValidation.Parse(); err != nil {--------------------------------------------
544  
545       // To look for a request's publisher_id in the NonStandardPublishers list in
546       // O(1) time, we fill this hash table located in the NonStandardPublisherMap field of GDPR
547       c.GDPR.NonStandardPublisherMap = make(map[string]int)
548       for i := 0; i < len(c.GDPR.NonStandardPublishers); i++ {
549           c.GDPR.NonStandardPublisherMap[c.GDPR.NonStandardPublishers[i]] = 1
550       }
551  
    +     var s struct{}
    +     for _, c := range c.GDPR.EEACountries {
    +     	c.GDPR.EEACountryMap[c] = s
    +     }
552       // To look for a request's app_id in O(1) time, we fill this hash table located in the
553       // the BlacklistedApps field of the Configuration struct defined in this file
554       c.BlacklistedAppMap = make(map[string]bool)
555       for i := 0; i < len(c.BlacklistedApps); i++ {
556           c.BlacklistedAppMap[c.BlacklistedApps[i]] = true
557       }
config/config.go
```

exchange/exchange.go
```diff
@@ -121,9 +126,27 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+	if geo != nil {
+		// If we have a country set, and it is on the list, we assume GDPR applies if not set on the request.
+		// Otherwise we assume it does not apply as long as it appears "valid" (is 3 characters long).
+		if _, found := e.eeaCountries[strings.ToUpper(geo.Country)]; found {
+			usersyncIfAmbiguous = false
+		} else if len(geo.Country) == 3 {
+			// The country field is formatted properly as a three character country code
+			usersyncIfAmbiguous = true
+		}
+	}
```
guscarreon, line 146: Just to make it extra readable maybe we could refactor this entire block to live its own function
```
129     // Make our best guess if GDPR applies
    +   usersyncIfAmbiguous := processUserSyncEnforcement(bidRequest, e)
130 -   usersyncIfAmbiguous := e.UsersyncIfAmbiguous
131 -   var geo *openrtb.Geo = nil
132 -   if bidRequest.User != nil && bidRequest.User.Geo != nil {
133 -       geo = bidRequest.User.Geo
134 -   } else if bidRequest.Device != nil && bidRequest.Device.Geo != nil {
135 -       geo = bidRequest.Device.Geo
136 -   }
137 -   if geo != nil {
138 -       // If we have a country set, and it is on the list, we assume GDPR applies if not set on the request.
139 -       // Otherwise we assume it does not apply as long as it appears "valid" (is 3 characters long).
140 -       if _, found := e.eeaCountries[strings.ToUpper(geo.Country)]; found {
141 -           usersyncIfAmbiguous = false
142 -       } else if len(geo.Country) == 3 {
143 -           // The country field is formatted properly as a three character country code
144 -           usersyncIfAmbiguous = true
145 -       }
146 -   }
exchange/exchange.go
```

exchange/exchange_test.go
```diff
@@ -1026,15 +1029,22 @@ func newExchangeForTests(t *testing.T, filename string, expectations map[string]
...
+		eeac[c] = s
+	}
+
 	return &exchange{
 		adapterMap:          adapters,
 		me:                  metricsConf.NewMetricsEngine(&config.Configuration{}, openrtb_ext.BidderList()),
 		cache:               &wellBehavedCache{},
 		cacheTime:           0,
-		gDPR:                gdpr.AlwaysAllow{},
+		gDPR:                gdpr.AlwaysFail{},
```
guscarreon, line 1043: Hans I'm getting this output. Is it just me?
```
╭─ ~/go/src/github.com/prebid/prebid-server/exchange  ‹PR1427*›
╰─➤  go test
# github.com/prebid/prebid-server/exchange [github.com/prebid/prebid-server/exchange.test]
./exchange_test.go:1046:3: cannot use "github.com/prebid/prebid-server/gdpr".AlwaysFail literal (type "github.com/prebid/prebid-server/gdpr".AlwaysFail) as type "github.com/prebid/prebid-server/gdpr".Permissions in field value:
        "github.com/prebid/prebid-server/gdpr".AlwaysFail does not implement "github.com/prebid/prebid-server/gdpr".Permissions (wrong type for PersonalInfoAllowed method)
                have PersonalInfoAllowed(context.Context, openrtb_ext.BidderName, string, string) (bool, bool, error)
                want PersonalInfoAllowed(context.Context, openrtb_ext.BidderName, string, string) (bool, bool, bool, error)
FAIL    github.com/prebid/prebid-server/exchange [build failed]
```

exchange/exchange.go
SyntaxNode: That's why I added privacyConfig. I didn't think to comment about adding this in there though.

exchange/exchange.go
SyntaxNode: It's only used in the exchange endpoint though. I'm personally happy with its current location.

PR #1431, opened 2020-08-06
Cache refactor

laurb9:
This PR makes the caches generic and reusable. 
There should be no functionality changes visible from a fetcherWithCache user.

Reason: Cache has Fetcher-like functionality to handle both requests and
imps together, and is only usable for these objects. Internally, it still uses two 
caches configured and searched separately. 

The PR moves out the request-specific part into fetcherWithCache only,
which handles separate caches for each object type.

Keeping the req/imp code repetition in fetcher and out of cache allows for
simpler code while preserving existing fetcher functionality.

Changes in this set:

- Cache is now a simple generic id->RawMessage store called CacheJSON
- fetcherWithCache handles the separate req and imp caches
- ComposedCache handles single caches - but it does not appear to be used
- Removed cache overlap tests since they do not apply now
- Slightly less code

### DIFF
--- a/stored_requests/caches/nil_cache/nil_cache.go
+++ b/stored_requests/caches/nil_cache/nil_cache.go
@@ -8,13 +8,14 @@ import (
 // NilCache is a no-op cache which does nothing useful.
 type NilCache struct{}
 
-func (c *NilCache) Get(ctx context.Context, requestIDs []string, impIDs []string) (map[string]json.RawMessage, map[string]json.RawMessage) {
-	return nil, nil
+func (c *NilCache) Get(ctx context.Context, ids []string) map[string]json.RawMessage {
+	return make(map[string]json.RawMessage)
 }
-func (c *NilCache) Save(ctx context.Context, storedRequests map[string]json.RawMessage, storedImps map[string]json.RawMessage) {
+
+func (c *NilCache) Save(ctx context.Context, data map[string]json.RawMessage) {
 	return
 }
 
-func (c *NilCache) Invalidate(ctx context.Context, requestIDs []string, impIDs []string) {
+func (c *NilCache) Invalidate(ctx context.Context, ids []string) {
 	return
 }
--- a/stored_requests/caches/cachestest/reliable.go
+++ b/stored_requests/caches/cachestest/reliable.go
@@ -11,8 +11,6 @@ import (
 const (
 	reqCacheKey = "known-req"
 	reqCacheVal = `{"req":true}`
-	impCacheKey = "known-imp"
-	impCacheVal = `{"imp":true}`
 )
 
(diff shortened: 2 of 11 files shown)


bsardo:
Hi @laurb9, 
At first glance, this looks good and is headed in the direction I had in mind as well. I'm not sure if you saw my comment on https://github.com/prebid/prebid-server/pull/1426. I'll continue going through your two PRs this evening and I'll open up one that captures some of the thoughts I had.

Regarding caching account data, were you thinking that the cache would store the data as JSON? My thought was that we would make the cache generic so that we could store maps where the values could be of any type, in which case, for accounts they would be object instances instead of JSON.

laurb9:
Hi @bsardo,
I initially tried changing to `interface{}` instead of json, but freecache stores `[]byte`, so some serialization was needed anyway. I had separated the account branch code to make review easier, so I stuck with json for now just to see things working.

Looking forward to your PR. I wanted to take a look at the AllFetcher as well. A few of them are created, but they're only ever used for one of the interfaces they subsume (for example, requests/imps only use theirs, categories use theirs etc). I would like to understand more about why that was necessary.

Btw I have another PR in this series: #1402 the lazy file fetcher, which together with this one, was intended to support loading account configuration from filesystem with ability to refresh.

stored_requests/caches/nil_cache/nil_cache.go
```diff
@@ -8,13 +8,13 @@ import (
 // NilCache is a no-op cache which does nothing useful.
 type NilCache struct{}
 
-func (c *NilCache) Get(ctx context.Context, requestIDs []string, impIDs []string) (map[string]json.RawMessage, map[string]json.RawMessage) {
-	return nil, nil
+func (c *NilCache) Get(ctx context.Context, ids []string) map[string]json.RawMessage {
```
hhhjort, line 11: Not a problem you introduced, but I have noticed something here. If the actual cache finds no hits, it returns an existing, but empty map. The nil cache returns a nil map. There is a subtle difference in at least the fact that you can expand an empty map, but you cannot assign any entries to a nil map. Most likely not an issue, but we may want the nil cache to act more like the actual cache.

laurb9: Given this comment https://github.com/prebid/prebid-server/blob/fb386190f4491648bb1e8d1b0345a333be1c0393/stored_requests/fetcher.go#L63 it looks like the caller expects a writeable map so I'll change to return `make(map[string]json.RawMessage)`

hhhjort (approved):
Looking good to me

laurb9:
Rebased to resolve conflicts.

hhhjort (approved):

bsardo (approved):

PR #1433, opened 2020-08-07
Default TCF1 GVL in anticipation of IAB no longer hosting the v1 GVL

hhhjort:


config/config.go
```diff
@@ -180,6 +181,12 @@ func (t *GDPRTimeouts) ActiveTimeout() time.Duration {
 	return time.Duration(t.ActiveVendorlistFetch) * time.Millisecond
 }
 
+// TCF1 defines the TCF1 specific configurations for GDPR
+type TCF1 struct {
+	FetchGVL       bool   `mapstructure:"fetch_gvl"`
+	DefaultGVLPath string `mapstructure:"default_gvl_path"`
```
SyntaxNode, line 187: The word "Default" in a config file leads me to think this might be, well, a default value and not one necessarily provided by the host. Consider renaming to `FallbackGVLPath`.

gdpr/vendorlist-fetching.go
```diff
@@ -27,8 +27,27 @@ type saveVendors func(uint16, api.VendorList)
 // Nothing in this file is exported. Public APIs can be found in gdpr.go
 
 func newVendorListFetcher(initCtx context.Context, cfg config.GDPR, client *http.Client, urlMaker func(uint16, uint8) string, TCFVer uint8) func(ctx context.Context, id uint16) (vendorlist.VendorList, error) {
+	var defaultVL api.VendorList = nil
+
+	if TCFVer == tCF1 && len(cfg.TCF1.DefaultGVLPath) > 0 {
```
SyntaxNode, line 32: Consider extracting this block to it's own method.

gdpr/vendorlist-fetching.go
```diff
@@ -27,8 +27,27 @@ type saveVendors func(uint16, api.VendorList)
...
+		defaultVLbody, err := ioutil.ReadFile(cfg.TCF1.DefaultGVLPath)
+		if err != nil {
+			glog.Fatalf("Error reading from file %s: %v", cfg.TCF1.DefaultGVLPath, err)
```
SyntaxNode, line 35: Should this method return an error as well if we can't load the default GVL? Same with the result of the `ParseEagerly` call.

gdpr/vendorlist-fetching_test.go
```diff
@@ -139,6 +141,78 @@ func TestVendorListMaker(t *testing.T) {
 	assertStringsEqual(t, "https://vendorlist.consensu.org/v2/archives/vendor-list-v7.json", vendorListURLMaker(7, 2))
 }
 
+func TestDefaultVendorList(t *testing.T) {
```
SyntaxNode, line 144: This looks like this test covers two cases. One where the GVL call succeeds and one where it fails. Can you please split them apart for clarity? I know this will duplicate some of the setup, but I think the added clarity is worth it.

gdpr/vendorlist-fetching.go
SyntaxNode, line 35: Can you please add a test case for a failed file read?

hhhjort: I can't have a fatal error at this point if I want to test for that, as won't `glog.Fatalf()` mean that the test will fail at that point?

gdpr/vendorlist-fetching_test.go
hhhjort: I will split it into  two

gdpr/vendorlist-fetching_test.go
```diff
@@ -139,6 +141,101 @@ func TestVendorListMaker(t *testing.T) {
...
+		32: {
+			purposes: []int{1, 2},
+		},
+	})
+	secondVendorList := mockVendorListData(t, 2, map[uint16]*purposes{
+		12: {
+			purposes: []int{2},
+		},
+	})
+	server := httptest.NewServer(http.HandlerFunc(mockServer(1, map[int]string{
```
mansinahar, line 218: Can we please use a var to hold the list version that we use here when creating the mockServer (for this and the tests above) and then later to assert that same version? It just helps understand better what those integer values represent while reading the tests.

PR #1434, opened 2020-08-10
Eplanning adapter: Get domain from page

chino117:


adapters/eplanning/eplanning.go
```diff
@@ -114,6 +114,13 @@ func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqIn
 	var pageDomain string
 	if request.Site != nil && request.Site.Domain != "" {
 		pageDomain = request.Site.Domain
+	} else if request.Site != nil && request.Site.Page != "" {
+		u, err := url.Parse(request.Site.Page)
+		if err != nil {
+			errors = append(errors, err)
```
SyntaxNode, line 120: Can you please add a test for this error handler?

adapters/eplanning/eplanning.go
```diff
@@ -114,6 +114,13 @@ func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqIn
 	var pageDomain string
 	if request.Site != nil && request.Site.Domain != "" {
 		pageDomain = request.Site.Domain
+	} else if request.Site != nil && request.Site.Page != "" {
```
SyntaxNode, line 117: Nitpick: The `request.Site != nil` check is repeated. What do you think of pushing down the `Site.Domain` and `Site.Page` check within a single nil check?

chino117:
Perfect, no problem

chino117:
The test is cleared as the error response from url.parse differs between versions 1.13 and 1.14.2 and there is no way that works in both versions. You could leave this error handler untested but if you don't prefer it, can you think of a practical way to test it?

SyntaxNode:
> The test is cleared as the error response from url.parse differs between versions 1.13 and 1.14.2 and there is no way that works in both versions. You could leave this error handler untested but if you don't prefer it, can you think of a practical way to test it?

Ah. yes. It's an unfortunate side effect of relying on framework generated messages. We first encountered this when moving to Go 1.13 when some of the error messages changed on how they included the name of the part of the struct with a marshal error. 

We've provided a way to test these kinds of error messages using a simple regex. Most adapter json tests use the literal comparison mode, but there's also an option to use regex which we introduced to address this exact problem.

Here's an example:
https://github.com/prebid/prebid-server/blob/cce496720f13d30d45154ca7d80e81884fb9a1ac/adapters/beintoo/beintootest/supplemental/invalid-response-unmarshall-error.json#L57

PR #1438, opened 2020-08-12
Fix TCF1 Fetcher Fallback

SyntaxNode:
Fixes a bug in 0.124.0 where the TCF1 vendor list fetcher will never attempt lazy loading.

### DIFF
--- a/gdpr/vendorlist-fetching.go
+++ b/gdpr/vendorlist-fetching.go
@@ -158,7 +158,7 @@ func newVendorListCache(fallbackVL api.VendorList) (save func(id uint16, list ap
 		if ok {
 			return list.(vendorlist.VendorList)
 		}
-		return fallbackVL
+		return nil
 	}
 	return
 }
--- a/gdpr/vendorlist-fetching_test.go
+++ b/gdpr/vendorlist-fetching_test.go
@@ -173,7 +173,7 @@ func TestDefaultVendorList(t *testing.T) {
 	assert.Equal(t, false, vendor.Purpose(2))
 }
 
-func TestDefaultVendorListPassthrough(t *testing.T) {
+func TestFallbackVendorListPassthrough(t *testing.T) {
 	firstVendorList := mockVendorListData(t, 1, map[uint16]*purposes{
 		32: {
 			purposes: []int{1, 2},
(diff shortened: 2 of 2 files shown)


gdpr/vendorlist-fetching.go
```diff
@@ -158,7 +158,7 @@ func newVendorListCache(fallbackVL api.VendorList) (save func(id uint16, list ap
 		if ok {
 			return list.(vendorlist.VendorList)
 		}
-		return fallbackVL
+		return nil
```
SyntaxNode, line 161: This method needs to return `nil` in order to attempt a url lazy loading fetch. If we return `fallbackVL` here, it effectively disables lazy loading.

gdpr/vendorlist-fetching_test.go
```diff
@@ -196,15 +196,15 @@ func TestDefaultVendorListPassthrough(t *testing.T) {
 	fetcher := newVendorListFetcher(context.Background(), testcfg, server.Client(), testURLMaker(server), 1)
 	list, err := fetcher(context.Background(), 2)
 	assert.NoError(t, err, "Error with fetching existing vendorlist: %v", err)
-	assert.Equal(t, uint16(2), list.Version(), "Expected to fetch mock list version 2, got version %d", list.Version())
+	assert.Equal(t, uint16(2), list.Version(), "Expected to fetch mock list version 1, got version %d", list.Version())
```
guscarreon, line 199: Shouldn't the expected value be `uint16(1)` instead?

gdpr/vendorlist-fetching_test.go
```diff
@@ -184,7 +184,7 @@ func TestDefaultVendorListPassthrough(t *testing.T) {
 			purposes: []int{2},
 		},
 	})
-	server := httptest.NewServer(http.HandlerFunc(mockServer(2, map[int]string{
```
SyntaxNode, line 187: Both lists were originally preloaded and always pulled from cache. If we change the latest version from 2 to 1, it no longer preloads version 2 and attempts a lazy load. This exposed the bug.

gdpr/vendorlist-fetching_test.go
SyntaxNode: No,. It should be version 2. I forgot to revert my change to the comment.

SyntaxNode: Fixed.

Issue #1337, opened 2020-06-04
Add source.tid and site.publisher.id to requests


bretg:
We found that Prebid Server is not passing source.tid to bidder adapters on AMP requests, and at least one bidder requires them.

All request types should have a source.tid. If the incoming request doesn't have one, Prebid Server should generate a random UUID.

Also, we should be passing site.publisher.id as the resolved account ID. Here's an updated account resolution algorithm. Step 4 is the new part.

1. look in site.publisher.id
2. look in app.publisher.id
3. if AMP, look for the 'account' parameter on the query string
4. When we've finally resolved the account ID, make sure it's set in site.publisher.id or app.publisher.id (whichever is appropriate)

SyntaxNode:
Ok. So, if the `site` exists but doesn't have a publisher.id, try to get it from the AMP query string. Some for flow for the `app`. I believe we verify a request can't have both `site` and `app`, so cross referencing isn't a concern.

bretg:
Done in PBS-Java

SyntaxNode:
Available in PBS-Go starting with version 0.127.0.

PR #1439, opened 2020-08-13
Add ability to randomly generate source.TID if empty and set publisher.ID to resolved account ID

mansinahar:
Addresses #1337 and #1040
Add ability to randomly generate source.TID if not present
Passing publisher.ID as the resolved account ID


config/config.go
```diff
@@ -71,6 +71,8 @@ type Configuration struct {
 	Debug Debug `mapstructure:"debug"`
 	// RequestValidation specifies the request validation options.
 	RequestValidation RequestValidation `mapstructure:"request_validation"`
+	// If req.Source.TID should be filled with a randomly generated UUID or not, if it is empty
+	AutoSourceTIDFill bool `mapstructure:"auto_source_tid_fill"`
```
mansinahar, line 75: I was torn between `AutoSourceTIDFill` and `RequireSourceTID` as the name for this config. Please lemme know if `RequireSourceTID` sounds better or if you have any other suggestions for this one.

endpoints/openrtb2/amp_auction.go
```diff
@@ -262,6 +263,21 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
 
+// Sets the effective publisher ID for an AMP request
+func setEffectiveAmpPubID(req *openrtb.BidRequest, urlValues url.Values) {
+	// Try to look for the pub ID in the ususal places first
+	setEffectivePubID(req)
+
+	// For amp requests, the publisher ID could be sent via the account
+	// query string
+	if req.Site.Publisher.ID == pbsmetrics.PublisherUnknown {
+		if acc := urlValues.Get("account"); acc != "" {
```
SyntaxNode, line 274: Can you please add a check to ignore unresolved macros? If the account is resolved to "ACCOUNT_ID" then it's unresolved.

SyntaxNode, line 274: There is already a community implementation in `overrideWithParams`.

mansinahar: Could there be other such unresolved macros? And is there a way for us to know what they might be? 

Also, since PBS is pretty much just a channel to pass this value, I am wondering if this is the kind of validation that should be done either downstream in the system that is calling us or upstream in the bidders where they actually use account ID value.

mansinahar: Sorry, I missed that change as it was done before I started working on this. I will remove the duplicate code and update the PR. Thanks for pointing it out :)

SyntaxNode: There might be other unresolved macros. Not sure. This is one we've seen in the real world though and is an annoyance in the metrics.

endpoints/openrtb2/amp_auction.go
```diff
@@ -154,7 +154,8 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	} else {
 		labels.CookieFlag = pbsmetrics.CookieFlagYes
 	}
-	labels.PubID = effectivePubID(req.Site.Publisher)
+	labels.PubID = req.Site.Publisher.ID
```
hhhjort, line 157: Is there a reason you are simplifying this this way? Currently an amp request should have a site and not an app object, but I don't think it is enforced. Someone at some point might come up with a use case for an app based amp request. This will mean we will break if an amp request uses app (we don't enforce sites in amp as far as I saw). We also won't get the proper parent_id if that was set in the amp request.

endpoints/openrtb2/amp_auction.go
```diff
@@ -262,6 +263,21 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
 }
 
+// Sets the effective publisher ID for an AMP request
+func setEffectiveAmpPubID(req *openrtb.BidRequest, urlValues url.Values) {
+	// Try to look for the pub ID in the ususal places first
+	setEffectivePubID(req)
+
+	// For amp requests, the publisher ID could be sent via the account
+	// query string
+	if req.Site.Publisher.ID == pbsmetrics.PublisherUnknown {
```
hhhjort, line 273: Do we have a check somewhere that `Site.Publisher` is set in the request?

endpoints/openrtb2/auction.go
```diff
@@ -1265,21 +1289,34 @@ func writeError(errs []error, w http.ResponseWriter, labels *pbsmetrics.Labels)
...
 		}
+		pub = req.Site.Publisher
+	}
+
+	if pub.Ext != nil {
+		var pubExt openrtb_ext.ExtPublisher
+		err := json.Unmarshal(pub.Ext, &pubExt)
+		if err == nil && pubExt.Prebid != nil && pubExt.Prebid.ParentAccount != nil &&
+			*pubExt.Prebid.ParentAccount != "" {
+			pub.ID = *pubExt.Prebid.ParentAccount
```
hhhjort, line 1313: Why are you setting the publisher ID to the parent account? This is a breaking change in behavior. Parent account was created to handle the case where the account in the PBS context does not match the publisher ID agreed upon in the publisher-bidder context.

endpoints/openrtb2/auction.go
```diff
@@ -138,18 +139,19 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	}
 
 	usersyncs := usersync.ParsePBSCookieFromRequest(r, &(deps.cfg.HostCookie))
+	setEffectivePubID(req, false, r.URL.Query())
 	if req.App != nil {
 		labels.Source = pbsmetrics.DemandApp
 		labels.RType = pbsmetrics.ReqTypeORTB2App
-		labels.PubID = effectivePubID(req.App.Publisher)
+		labels.PubID = req.App.Publisher.ID
```
hhhjort, line 146: This still eliminates using the `ParentID` as our label value in metrics? This will also break being able to set our white/blacklists on accounts using the `ParentID` rather than the `Publisher.ID`.

endpoints/openrtb2/auction.go
```diff
@@ -1265,21 +1289,33 @@ func writeError(errs []error, w http.ResponseWriter, labels *pbsmetrics.Labels)
...
+	if pub.ID == "" && isAmp {
+		// For amp requests, the publisher ID could be sent via the account
+		// query string
+		if acc := urlValues.Get("account"); acc != "" && acc != "ACCOUNT_ID" {
+			pub.ID = acc
+		}
+	}
+
+	if pub.ID == "" {
+		pub.ID = pbsmetrics.PublisherUnknown
```
hhhjort, line 1317: This will cause us to add our unknown publisher value in metrics as the publisher in the request, rather than leaving it unset. This will cause confusion in bidders who do not know that this particular string is PBS metrics slang for "unknown publisher" rather than a publisher explicitly called out with an unusual ID, who may end up appearing to have a lot of diverse inventory. I think we want unknown publishers to remain with an unset ID in the request, as that is the normal signal for an unknown publisher.

hhhjort:
Seems you have replaced `EffectivePubID()` which could probably be better renamed to `getAccountID()` with `SetEffectivePubID()` that has an entirely different, AMP specific purpose.

endpoints/openrtb2/auction.go
```diff
@@ -283,6 +284,14 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) []error {
...
 	}
 
+	// If automatically filling source TID is enabled then validate that
+	// source.TID exists and If it doesn't, fill it with a randomly generated UUID
+	if deps.cfg.AutoSourceTIDFill {
+		if err := validateAndFillSourceTID(req); err != nil {
+			return []error{err}
+		}
+	}
+
```
guscarreon, line 294: Can we add a `validateRequest` test case to cover the true branch of the newly added if statement?

image

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1042,3 +1042,84 @@ func getTestBidRequest(nilUser bool, userExt *openrtb_ext.ExtUser, nilRegs bool,
...
+	}{
+		{
+			req: &openrtb.BidRequest{
+				App: &openrtb.App{
+					Publisher: nil,
+				},
+			},
+			expectedPubID: "",
+			description:   "No publisher ID provided",
+		},
```
guscarreon, line 1065: Can we list the `description` first to make tests more readable?
```
    testCases := []struct {
        req           *openrtb.BidRequest
        urlValues     url.Values
        expectedPubID string
        description   string
    }{
        {
 +          description:   "No publisher ID provided",
            req: &openrtb.BidRequest{
                App: &openrtb.App{
                    Publisher: nil,
                },
            },
            expectedPubID: "",
 -          description:   "No publisher ID provided",
        },
```

endpoints/openrtb2/auction.go
```diff
@@ -358,6 +367,20 @@ func (deps *endpointDeps) validateRequest(req *openrtb.BidRequest) []error {
...
+	}
+	if req.Source.TID == "" {
+		if rawUUID, err := uuid.NewV4(); err == nil {
+			req.Source.TID = rawUUID.String()
+		} else {
+			return errors.New("req.Source.TID missing in the req and error creating a random UID")
+		}
+	}
+	return nil
+}
```
guscarreon, line 382: I believe latter on the execution flow in function `prepareSource`, the `Source` field might get overwritten (line 216 below). Should we prevent this and also add a test that runs either `prepareSource` or even other function lower in the stack to make sure that a given bid request with a random `Source.TID` doesn't get overwritten?
```
190 func prepareSource(req *openrtb.BidRequest, bidder string, sChainsByBidder map[string]*openrtb_ext.ExtRequestPrebidSChainSChain) {
191     const sChainWildCard = "*"
192     var selectedSChain *openrtb_ext.ExtRequestPrebidSChainSChain
193
194     wildCardSChain := sChainsByBidder[sChainWildCard]
195     bidderSChain := sChainsByBidder[bidder]
196
197     // source should not be modified
198     if bidderSChain == nil && wildCardSChain == nil {
199         return
200     }
201
202     if bidderSChain != nil {
203         selectedSChain = bidderSChain
204     } else {
205         selectedSChain = wildCardSChain
206     }
207
208     // set source
209     var source openrtb.Source
210     schain := openrtb_ext.ExtRequestPrebidSChain{
211         SChain: *selectedSChain,
212     }
213     sourceExt, err := json.Marshal(schain)
214     if err == nil {
215         source.Ext = sourceExt
216         req.Source = &source //<-- gets written
217     }
218 }
exchange/utils.go
```

endpoints/openrtb2/amp_auction.go
```diff
@@ -564,3 +559,28 @@ func readConsent(url *url.URL) string {
 	// Fallback to 'gdpr_consent' for compatability until it's no longer used by AMP.
 	return url.Query().Get("gdpr_consent")
 }
+
+// Sets the effective publisher ID for amp request
+func setEffectiveAmpPubID(req *openrtb.BidRequest, urlValues url.Values) {
```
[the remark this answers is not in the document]
mansinahar: Renamed it to `urlQueryParams` for better readability

endpoints/openrtb2/auction_test.go
```diff
@@ -1269,6 +1259,62 @@ func TestSChainInvalid(t *testing.T) {
...
+		},
+		{
+			pub:           nil,
+			expectedPubID: pbsmetrics.PublisherUnknown,
+			description:   "Publisher is nil",
+		},
+	}
+
+	for _, test := range testCases {
+		getAccountID(test.pub)
```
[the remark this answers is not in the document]
mansinahar: Yes! Good catch. Thanks

endpoints/openrtb2/auction.go
mansinahar: I believe this is more of a bug in the `prepareSource` function. We should be checking if a field exists in the request before replacing it. I added that check to this function so that it is creating a new `req.Source` object only if it wasn't present in the request to start with. This way the `TID` value or any other values in the `req.Source` object won't get overridden. 

That was a great catch though! Thanks :)

endpoints/openrtb2/auction_test.go
```diff
@@ -1222,6 +1212,54 @@ func TestCCPAInvalid(t *testing.T) {
 	assert.Empty(t, req.Regs.Ext, "Invalid Consent Removed From Request")
 }
 
+func TestValidateSourceTID(t *testing.T) {
+	cfg := &config.Configuration{
+		AutoGenSourceTID: true,
+	}
```
mansinahar, line 1218: I have added a very minimal test based on the setup we have today. Ideally we'd wanna add a table driven test with multiple test cases but validateRequest is a big function and we'd have to refactor all the existing one-off validateRequest tests to convert into table driven tests. That will make this PR very crowded and so it'd be better for that to be its own separate PR.

endpoints/openrtb2/auction_test.go
```diff
@@ -1344,6 +1439,49 @@ func TestSanitizeRequest(t *testing.T) {
...
+		req           *openrtb.BidRequest
+		expectRandTID bool
+		expectedTID   string
+		description   string
+	}{
+		{
+			req:           &openrtb.BidRequest{},
+			expectRandTID: true,
+			description:   "req.Source not present. Expecting a randomly generated TID value",
+		},
```
guscarreon, line 1454: Nitpick. Can the description come first? I find it more convenient/readable.
```
34     testCases := []struct {
   +       description   string
33         req           *openrtb.BidRequest
32         expectRandTID bool
31         expectedTID   string
30 -       description   string
29     }{
28         {
   +           description:   "req.Source not present. Expecting a randomly generated TID value",
27             req:           &openrtb.BidRequest{},
26             expectRandTID: true,
25 -           description:   "req.Source not present. Expecting a randomly generated TID value",
24         },
endpoints/openrtb2/auction_test.go
```

guscarreon (approved):
LGTM

hhhjort (approved):

PR #1443, opened 2020-08-18
moving docs to website repo

bretg:
The proposal here is that the website docs over at prebid.github.io are now the master docs.

Assuming we agree on that, then this PR cleans up the code repo docs so we have only one source of truth.


SyntaxNode:
Agreed. Thank you for putting up this PR.

A few questions:
1. I see that bidders/appnexus.md is still here. Is that not represented in the docs repo?
2. I see that developers/stored-request.md is still here, but [also in the docs repo](https://docs.prebid.org/prebid-server/features/pbs-storedreqs-go.html). Are there parts of the doc not in the docs repo?
3. Instead of leaving small files around with a url, how about we add a single readme added to the docs folder to list out those deep links?
4. What's our philosophy for including new adapter and analytic module docs in the docs repo, but leaving some other version specific things like setting up stored requests here?

hhhjort:
The `README.md` in the root directory should be updated. It refers to docs that this PR removes, and should be updated to point to the new location(s)

bretg:
Good call @hhhjort - updated README in this PR

PR #1446, opened 2020-08-20
Clarify stored configuration warnings by indicating dataType

laurb9:
Add a hidden DataType field to StoredRequest, StoredRequestSlim structs, populated with the datatype ("Request", "Category",  etc). Modify warnings to include this field so we know which section they refer to. A misleading warning about ext.prebid.storedrequest being ignored in the wrong section was also fixed.

Startup log messages before the change:

> Loading Stored Requests from filesystem at path ./stored_requests/data/by_id
No Stored Request cache configured. The Fetcher backend will be used for all Stored Requests.
Loading Stored Requests from filesystem at path ./stored_requests/data/by_id
No Stored Request cache configured. The Fetcher backend will be used for all Stored Requests.
Loading Stored Requests from filesystem at path ./static/category-mapping 
No Stored Request support configured. request.imp[i].ext.prebid.storedrequest will be ignored. If you need this, check your app config
No Stored Request cache configured. The Fetcher backend will be used for all Stored Requests.

After the change:

> Loading Stored Request data from filesystem at path ./stored_requests/data/by_id
No Stored Request cache configured. The Request Fetcher backend will be used for all data requests
Loading Stored Amp Request data from filesystem at path ./stored_requests/data/by_id
No Stored Amp Request cache configured. The Amp Request Fetcher backend will be used for all data requests
Loading Stored Category data from filesystem at path ./static/category-mapping
No Stored Video support configured. If you need this, check your app config
No Stored Video cache configured. The Video Fetcher backend will be used for all data requests

### DIFF
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -14,10 +14,10 @@ import (
 type DataType string
 
 const (
-	requestDataType  DataType = "Request"
-	impDataType      DataType = "Imp"
-	categoryDataType DataType = "Category"
-	videoDataType    DataType = "Video"
+	RequestDataType    DataType = "Request"
+	CategoryDataType   DataType = "Category"
+	VideoDataType      DataType = "Video"
+	AmpRequestDataType DataType = "Amp Request"
 )
 
 // StoredRequests configures the backend used to store requests on the server.
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -11,8 +10,20 @@ import (
 	"github.com/golang/glog"
 )
 
+// DataType constants
+type DataType string
+
+const (
+	requestDataType  DataType = "Request"
+	impDataType      DataType = "Imp"
+	categoryDataType DataType = "Category"
+	videoDataType    DataType = "Video"
+)
+
 // StoredRequests configures the backend used to store requests on the server.
 type StoredRequests struct {
+	// DataType is a tag pushed from upstream indicating the type of object fetched here
+	DataType DataType
 	// Files should be true if Stored Requests should be loaded from the filesystem.
 	Files bool `mapstructure:"filesystem"`
 	//If data should be loaded from file system, path should be specified in configuration
@@ -39,6 +50,8 @@ type StoredRequests struct {
 
 // StoredRequestsSlim struct defines options for stored requests from a single endpoint
 type StoredRequestsSlim struct {
+	// DataType is a tag pushed from upstream indicating the type of object fetched here
+	DataType DataType
 	// Files should be used if Stored Requests should be loaded from the filesystem.
 	// Fetchers are in stored_requests/backends/file_system/fetcher.go
 	Files FileFetcherConfig `mapstructure:"filesystem"`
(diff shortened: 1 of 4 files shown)


config/stored_requests.go
```diff
@@ -11,8 +10,20 @@ import (
...
+	RequestDataType    DataType = "Request"
+	CategoryDataType   DataType = "Category"
+	VideoDataType      DataType = "Video"
+	AmpRequestDataType DataType = "Amp Request"
+)
+
 // StoredRequests configures the backend used to store requests on the server.
 type StoredRequests struct {
+	// DataType is a tag pushed from upstream indicating the type of object fetched here
+	DataType DataType
```
SyntaxNode, line 26: There seems to be a trend in the feedback here about the placement of the DataTypes.

The `StoredRequests` and `StoredRequestsSlim` are config structs which the host can control. It seems strange IMHO to give the host the option of configuring the `DataType` field, so this doesn't seem like the opportune spot.

I recommend changing the signature of `CreateStoredRequests` to accept a `DataType` and pass it down as hardcoded values in `NewStoredRequests`. I think it would like nice paired with the config source:

```
fetcher1, shutdown1 := CreateStoredRequests(&slimAuction, RequestDataType, metricsEngine, client, router, &dbc)
fetcher2, shutdown2 := CreateStoredRequests(&slimAmp, metricsEngine, AmpRequestDataType , client, router, &dbc)
fetcher3, shutdown3 := CreateStoredRequests(&cfg.CategoryMapping, CategoryDataType, metricsEngine, client, router, &dbc)
fetcher4, shutdown4 := CreateStoredRequests(&cfg.StoredVideo, VideoDataType, metricsEngine, client, router, &dbc)
```

... or change the order fo have the `DataType` as the first argument if you find it reads better.

config/stored_requests.go
```diff
@@ -11,8 +10,20 @@ import (
...
 )
 
+// DataType constants
+type DataType string
+
+const (
+	RequestDataType    DataType = "Request"
+	CategoryDataType   DataType = "Category"
+	VideoDataType      DataType = "Video"
+	AmpRequestDataType DataType = "Amp Request"
```
SyntaxNode, line 20: Nitpick: AMP is an acronym. Following Go naming conventions it should  be "AMPRequestDataType".

config/stored_requests.go
laurb9: I see DataType as metadata allowing any function passed the config to tell what kind it is. Passing the type separately gives us a pair of parameters meant to always be passed together, an indication to me that they should be fused.

I would have preferred to make it private so it is not exposed at all, but it needs to be readable so this implementation looked like a decent compromise.

The field itself lacks an annotation and is hardcoded, so it is not under the host control. 
It is similar to `BlackListedAppMap` or `BlackListedAcctMap`: https://github.com/prebid/prebid-server/blob/master/config/config.go#L61

I can change it to a separate parameter if that is still preferred.

SyntaxNode: That's a good point. I can appreciate representing the pair of config and data type as a struct instead of always passing them together separately. Making the field private would be better IMHO, but as you've pointed out that won't work between packages.

The annotation on the config isn't actually required for serialization. So while I don't think anyone would specify a yaml config like:
```
stored_requests:
  datatype: "foo"
```
.. it would actually work, and that doesn't give me warm feelings.

What do you think of wrapping the `StoredRequests` in another struct which includes extra data, such as the type. Something like this (but with a hopefully better name):
```
type StoredRequestsWithMetadata struct {
  StoredRequests
  DataType DataType
}
```

laurb9:
Closing in favor of #1453 which includes this change.

PR #1447, opened 2020-08-20
Update gamma adapter

gammassp:
Some response from gamma does not return VAST inlined


adapters/gamma/gamma.go
```diff
@@ -197,21 +239,33 @@ func (a *GammaAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalReq
...
+	if err := json.Unmarshal(response.Body, &gammaResp); err != nil {
 		return nil, []error{&errortypes.BadServerResponse{
 			Message: fmt.Sprintf("bad server response: %d. ", err),
 		}}
 	}
 
-	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(bidResp.SeatBid[0].Bid))
-	for _, sb := range bidResp.SeatBid {
+	//(Section 7.1 No-Bid Signaling)
+	if gammaResp.SeatBid == nil || len(gammaResp.SeatBid) == 0 {
```
jmaynardxandr, line 250: In Go, len(nil) = 0, so you could omit the first check here if you want to.

adapters/gamma/gamma.go
```diff
@@ -197,21 +239,33 @@ func (a *GammaAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalReq
...
-			})
+			mediaType := getMediaTypeForImp(gammaResp.ID, internalRequest.Imp)
+			bid := convertBid(sb.Bid[i], mediaType)
+			if bid != nil {
+				bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
+					Bid:     bid,
+					BidType: mediaType,
+				})
+			} else {
+				return nil, []error{&errortypes.BadServerResponse{
```
jmaynardxandr, line 265: Do you really want to discard all bids by returning nil here if any of them are missing their ad markup, or should you just discard the bid that's missing its ad markup but still return any other bids that were ok here?

adapters/gamma/gamma.go
```diff
@@ -16,6 +16,26 @@ import (
 type GammaAdapter struct {
 	URI string
 }
+type gammaBid struct {
```
SyntaxNode, line 19: Nitpick: Please add a leading line break to match the spacing.

adapters/gamma/gamma.go
gammassp: Oh, I almost forget it, thanks for notifying me.

adapters/gamma/gamma.go
```diff
@@ -197,24 +240,38 @@ func (a *GammaAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalReq
...
 		return nil, []error{&errortypes.BadServerResponse{
 			Message: fmt.Sprintf("bad server response: %d. ", err),
 		}}
 	}
 
-	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(bidResp.SeatBid[0].Bid))
-	for _, sb := range bidResp.SeatBid {
+	//(Section 7.1 No-Bid Signaling)
+	if len(gammaResp.SeatBid) == 0 {
+		return nil, nil
```
SyntaxNode, line 252: Please add a test for this case.

adapters/gamma/gamma.go
```diff
@@ -180,6 +201,28 @@ func (a *GammaAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapte
 	return adapterRequests, errs
 }
 
+func convertBid(gBid gammaBid, mediaType openrtb_ext.BidType) *openrtb.Bid {
+	var bid openrtb.Bid
+	bid = gBid.Bid
+
+	if mediaType == openrtb_ext.BidTypeVideo {
+		//Return inline VAST XML Document (Section 6.4.2)
```
SyntaxNode, line 209: This branch (where mediaType == openrtb_ext.BidTypeVideo) has no test coverage. Can you please add tests to cover this new logic?

PR #1449, opened 2020-08-21
fix: avoid unexpected EOF on gz writer

gpolaert:
We've spotted an issue on our side. This PR fixes it.

### DIFF
--- a/analytics/pubstack/eventchannel/eventchannel.go
+++ b/analytics/pubstack/eventchannel/eventchannel.go
@@ -94,9 +94,9 @@ func (c *EventChannel) flush() {
 	}
 
 	// finish writing gzip header
-	err := c.gz.Flush()
+	err := c.gz.Close()
 	if err != nil {
-		glog.Warning("[pubstack] fail to flush gzipped buffer")
+		glog.Warning("[pubstack] fail to close gzipped buffer")
 		return
 	}
 
--- a/analytics/pubstack/eventchannel/eventchannel.go
+++ b/analytics/pubstack/eventchannel/eventchannel.go
@@ -93,6 +93,9 @@ func (c *EventChannel) flush() {
 		return
 	}
 
+	// reset buffers and writers
+	defer c.reset()
+
 	// finish writing gzip header
 	err := c.gz.Close()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


analytics/pubstack/eventchannel/eventchannel.go
```diff
@@ -94,9 +94,9 @@ func (c *EventChannel) flush() {
 	}
 
 	// finish writing gzip header
-	err := c.gz.Flush()
+	err := c.gz.Close()
```
SyntaxNode, line 97: Makes sense. Looking at the gzip code, Flush empties the memory buffer but Close also writes a footer.

After you call Close though, the gz writer is left in a bad state. This is normally solved via `c.reset()`, but if there's an error calling Close the reset doesn't occur and the state is left in a potentially bad place.

I can guide on solving this, but first I need to know what you want to happen if the Close fails. Would you like to try again or dump the data, log an error, and move on?

gpolaert: Right now, an error log is good.
It could be good to implement another behavior in the future :)

What do you think about deferring the reset right before the close call?

gpolaert: Like that
```go
        defer c.reset()
       
        // finish writing gzip header
	err := c.gz.Close()
	if err != nil {
		glog.Warning("[pubstack] fail to close gzipped buffer")
		return
	}
```

SyntaxNode: Cool. That works for me. The behavior of multiple `defer` is documented by the language, so `c.reset()` will be run before the lock. That's a little subtle, but safe. If you're good with it, so am I.

You may also want to consider adding a metric for failed writes, so hosts can monitor and setup alerts. That can certainly be a separate PR down the line.

gpolaert:
I'm good with the stacked defer.
You've right, I can add more metrics and logs to help people to monitor the integration.

It's for us a Proof-of-concept with some of our publishers => so right now, it's okay ;)

SyntaxNode:
Would be able to add a unit test to ensure the gzip encoded bytes are exactly as you expect to prevent this issue from reoccurring in the future?

gpolaert:
Good idea!
Here is the test. The test fails if I use `flush` instead `close`.

Issue #1406, opened 2020-07-17
tests ignore expected message field. When not ignored, tests actually fail. Currently tests pass when they should not.


draksler:
tests in TestBadRequests in auction_test.go (and a few others) are ignoring the expected message field due to the `gr.description != ""` condition. 
If the condition is removed, and the messages are actually validated, the tests fail.

SyntaxNode:
Nice find. I'm looking into this. I don't know why a condition like `gr.description != ""` exists at all.

SyntaxNode:
We're working to fix this. Looks like there are two tests which aren't running and should. The other failures with removing that condition are due to unclear setup. This was probably our first pass at json based tests.

SyntaxNode:
The test setup in that package is a mix of json and code configuration. It's really difficult to follow. @guscarreon has rewritten them in a much cleaner style with all setup expressed in the json files, as we do for other parts of the code. I'll close this ticket when https://github.com/prebid/prebid-server/pull/1450 is merged.

SyntaxNode:
The test setup in that package is a mix of json and code configuration. It's really difficult to follow. @guscarreon has rewritten them in a much cleaner style with all setup expressed in the json files, as we do for other parts of the code. I'll close this ticket when

SyntaxNode:
`auction_test.go` has been completely refactored to resolve this issue.

PR #1450, opened 2020-08-21
Fix JSON tests ignore expected message field

guscarreon:
Issue [https://github.com/prebid/prebid-server/issues/1406](https://github.com/prebid/prebid-server/issues/1406) relates to the fact that tests in `endpoints/openrtb2/auction_test.go` not only read from the JSON files but also have a "code" component to them. This pull request makes the tests in `endpoints/openrtb2/auction_test.go` read from the JSON files found in `endpoints/openrtb2/sample-requests` both their test values and their respective expected values to assert. 

This PR also gets rid of most of the "code" component of `endpoints/openrtb2/auction_test.go` and eliminates unnecessary JSON files in `endpoints/openrtb2/sample-requests`.


endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
-	tests := &getResponseFromDirectory{
-		dir:           "sample-requests/aliased",
-		payloadGetter: noop,
-		messageGetter: nilReturner,
-		expectedCode:  http.StatusOK,
-		aliased:       true,
+func jsonParseTestFile(t *testing.T, fileData []byte, testFile string) *testCaseData {
+	t.Helper()
+
+	parsedTestData := &testCaseData{}
```
SyntaxNode, line 249: A reference doesn't seem necessary here. Consider changing to `parsedTestData := testCaseData{}`

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
 
-// TestAliasedRequests makes sure we handle (default) aliased bidders properly
-func TestAliasedRequests(t *testing.T) {
-	tests := &getResponseFromDirectory{
-		dir:           "sample-requests/aliased",
-		payloadGetter: noop,
-		messageGetter: nilReturner,
-		expectedCode:  http.StatusOK,
-		aliased:       true,
+func jsonParseTestFile(t *testing.T, fileData []byte, testFile string) *testCaseData {
```
SyntaxNode, line 246: Consider renaming to simply `parseTestFile` or `loadTestFile`. The json description doesn't add any value to this name, IMHO.

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
+	// Get testCaseData values
+	parsedTestData.BidRequest, _, _, err = jsonparser.Get(fileData, "mockBidRequest")
+	assert.NoError(t, err, "Error jsonparsing root.mockBidRequest from file %s. Desc: %v.", testFile, err)
+
+	parsedReturnCode, err := jsonparser.GetInt(fileData, "expectedReturnCode")
+	assert.NoError(t, err, "Error jsonparsing root.code from file %s. Desc: %v.", testFile, err)
+
+	parsedTestData.ExpectedReturnCode = int(parsedReturnCode)
+
+	if parsedTestData.ExpectedReturnCode != 200 {
```
SyntaxNode, line 261: This doesn't seem like the right place to make this check. Consider loading both the ExpectedErrorMessage and ExpectedBidResponse.

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
-		expectedCode:    http.StatusBadRequest,
-		aliased:         false,
-		disabledBidders: []string{"appnexus", "rubicon"},
-		adaptersConfig: map[string]config.Adapter{
-			"appnexus": {Disabled: true},
-			"rubicon":  {Disabled: true},
-		},
+	// Get testCaseConfig values
+	parsedTestData.Config = &testConfigValues{}
+	accReq, err := jsonparser.GetBoolean(fileData, "mockConfig", "accountRequired")
```
SyntaxNode, line 273: The term mock here is a little strange IMHO. I would recommend just calling this the 'config' section of the json file.

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
+		parsedTestData.Config.AliasJSON = aliases
 	}
-	badTests.assert(t)
-	goodTests.assert(t)
+
+	parsedTestData.Config.BlacklistedAccountArr = parseStringArray(fileData, "blacklistedAccts")
+	parsedTestData.Config.BlacklistedAppArr = parseStringArray(fileData, "blacklistedApps")
+	parsedTestData.Config.AdapterList = parseStringArray(fileData, "disabledAdapters")
+
+	return parsedTestData
```
SyntaxNode, line 287: One call to jsonparser is faster than the built in library, but the benefit becomes a bit more dubious over multiple calls. Why not just use the built in library?

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
-	tests := &getResponseFromDirectory{
-		dir:           "sample-requests/blacklisted",
-		payloadGetter: getRequestPayload,
-		messageGetter: getMessage,
-		expectedCode:  http.StatusServiceUnavailable,
-		aliased:       false,
-	}
-	tests.assert(t)
+func parseStringArray(fileData []byte, jsonField string) []string {
+	rarr := []string{}
```
SyntaxNode, line 291: `rarr` is a very confusing name. Consider calling it `results` or replace with using the standard library.

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
-		{
-			// Account is required, was provided in request and is found in the  blacklisted accounts map
-			dir:           "sample-requests/blacklisted",
-			file:          "blacklisted-acct.json",
-			payloadGetter: getRequestPayload,
-			messageGetter: getMessage,
-			expectedCode:  http.StatusServiceUnavailable,
-			accountReq:    true,
-		},
+func assertTestCaseData(t *testing.T, fileData []byte, testFile string) {
```
SyntaxNode, line 298: Since this is both loading the test case and running it, I recommend to name it from `assertTestCaseData` to `runTestCase`. The current names leads me to believe this will just contain the assert statements.

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
+	if test.ExpectedReturnCode != 200 {
+		// Assert expected error
+		assert.True(t, strings.HasPrefix(actualBidResponse, test.ExpectedErrorMessage), "Test failed. %s. Filename: %s \n", actualBidResponse, testFile)
+	} else {
+		// Assert expected response
+		diffJson(t, testFile, []byte(actualBidResponse), test.ExpectedBidResponse)
+	}
+}
+
+func parseMaps(tc *testConfigValues) *testConfigValues {
```
SyntaxNode, line 322: Please rename this method so its obvious this is building the test config.

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
+// diffJson compares two JSON byte arrays for structural equality. It will produce an error if either
+// byte array is not actually JSON.
+func diffJson(t *testing.T, description string, actual []byte, expected []byte) {
 	t.Helper()
-	var filesToAssert []string
-	if gr.file == "" {
-		// Append every file found in `gr.dir` to the `filesToAssert` array and test them all
-		for _, fileInfo := range fetchFiles(t, gr.dir) {
-			filesToAssert = append(filesToAssert, gr.dir+"/"+fileInfo.Name())
+	diff, err := gojsondiff.New().Compare(actual, expected)
```
SyntaxNode, line 348: This is different than we normally use to compare json, but it doesn't strike me as odd. If this looks good to others, I have no issue.

endpoints/openrtb2/sample-requests/invalid-whole/array.json
```diff
@@ -1,4 +1,6 @@
 {
```
SyntaxNode, line 1: `array.json`, `boolean.json`, `float.json`, `integer.json`, `null.json, and `empty-object.json` are all testing the same deserialization error. How about we keep just one (doesn't matter which) and rename it `malformed-request.json`?

endpoints/openrtb2/sample-requests/invalid-whole/format-empty-array.json
```diff
@@ -1,6 +1,6 @@
 {
-  "message": "Invalid request: request.imp[0].banner has no sizes. Define \"w\" and \"h\", or include \"format\" elements.\n",
-  "requestPayload": {
+  "description": "Banner does not define witdth nor height",
```
SyntaxNode, line 2: Nitpick: witdth is mispelled.

endpoints/openrtb2/sample-requests/invalid-whole/format-empty-object.json
```diff
@@ -1,6 +1,6 @@
 {
-  "message": "Invalid request: Request imp[0].banner.format[0] should define *either* {w, h} (for static size requirements) *or* {wmin, wratio, hratio} (for flexible sizes) to be non-zero.\n",
-  "requestPayload": {
+  "description": "Banner does not define witdth nor height",
```
SyntaxNode, line 2: Nitpick: witdth is mispelled. Consider changing this message slightly to distinguish it from `format-empty-array.json`.

endpoints/openrtb2/sample-requests/invalid-whole/format-two-widths.json
```diff
@@ -1,6 +1,6 @@
 {
-  "message": "Invalid request: Request imp[0].banner.format[0] should define *either* {w, h} *or* {wmin, wratio, hratio}, but not both. If both are valid, send two \"format\" objects in the request.\n",
-  "requestPayload": {
+  "description": "Banner comes with no width nor wratio value",
```
SyntaxNode, line 2: Description is incorrect. It comes with both specified, which is invalid.

endpoints/openrtb2/sample-requests/invalid-whole/regs-ext-gdpr-invalid.json
```diff
@@ -1,11 +1,11 @@
 {
-  "message": "Invalid request: request.regs.ext.gdpr must be either 0 or 1.\n",
-  "requestPayload": {
+  "description": "Bid request defines and invalid GDPR value",
```
SyntaxNode, line 2: Typo: and -> an

endpoints/openrtb2/auction_test.go
SyntaxNode: We can't because some of the files have invalid json. As a compromise, we agreed to explore using json.unmarshal only for the mockConfig.

endpoints/openrtb2/sample-requests/invalid-whole/alias-bidder-self.json
```diff
@@ -13,17 +13,19 @@
         },
         "ext": {
           "appnexus": {
-            "placementId": 12883451
+            "placementid": 12883451
```
SyntaxNode, line 16: Is this a typo? Looks like it was changed from `placementId` to `placementid`.

endpoints/openrtb2/auction_test.go
```diff
@@ -120,196 +191,178 @@ func TestExplicitUserId(t *testing.T) {
...
-		expectedCode:  http.StatusOK,
-		aliased:       true,
+	fileList, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Append the path of  every file found in `dir` to the `filesToAssert` array
+	for _, fileInfo := range fileList {
+		filesToAssert = append(filesToAssert, dir+"/"+fileInfo.Name())
```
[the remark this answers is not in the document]
guscarreon: Pretty good catch. Corrected

endpoints/openrtb2/auction_test.go
guscarreon: `results` sounds great. Modified.

endpoints/openrtb2/sample-requests/aliased/simple.json
```diff
@@ -1,4 +1,9 @@
 {
+  "description": "Imp extension doesn't come with valid bidder name but does come with valid bidder alias as defined in the mockAliases list.",
+  "mockConfig": {
+      "aliases": "{\"ext\":{\"prebid\":{\"aliases\":{\"test1\":\"appnexus\",\"test2\":\"rubicon\",\"test3\":\"openx\"}}}}"
```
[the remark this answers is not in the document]
guscarreon: Removed `test2` and `test3`

endpoints/openrtb2/sample-requests/account-required/with-account/required-blacklisted-acct.json
```diff
@@ -0,0 +1,91 @@
+{
```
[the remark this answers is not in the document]
guscarreon: They only differ in one line 4

endpoints/openrtb2/sample-requests/invalid-whole/alias-bidder-self.json
guscarreon: Typo. Corrected

endpoints/openrtb2/sample-requests/invalid-whole/regs-ext-gdpr-invalid.json
guscarreon: Sharp eyes. Corrected

endpoints/openrtb2/sample-requests/invalid-whole/format-two-widths.json
guscarreon: Copy-paste error. Corrected

endpoints/openrtb2/sample-requests/invalid-whole/format-empty-object.json
guscarreon: Both corrected

endpoints/openrtb2/sample-requests/invalid-whole/array.json
guscarreon: Done

endpoints/openrtb2/sample-requests/invalid-native/asset-title-no-length.json
```diff
@@ -1,9 +1,25 @@
 {
```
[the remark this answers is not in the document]
guscarreon: Done

endpoints/openrtb2/sample-requests/disabled/bad/bad-alias.json
```diff
@@ -1,6 +1,9 @@
 {
-  "message": "Invalid request: request.ext.prebid.aliases.test1 refers to unknown bidder: appnexus\n",
-  "requestPayload": {
+  "description": "Request comes with an alias to a disabled bidder, we should throw error",
```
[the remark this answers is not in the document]
guscarreon: The alias is defined in the bid request extension in line 28
```
25     "ext": {
26       "prebid": {
27         "aliases": {
28           "test1": "appnexus"
29         }
30       }
31     }
32   },
33   "expectedReturnCode": 400,
34   "expectedErrorMessage": "Invalid request: request.ext.prebid.aliases.test1 refers to unknown bidder: appnexus\n"
35 }
endpoints/openrtb2/sample-requests/disabled/bad/bad-alias.json
```

endpoints/openrtb2/sample-requests/aliased/site.json
```diff
@@ -1,4 +1,9 @@
 {
+  "description": "Imp extension comes with both valid bidder name and valid bidder alias as defined in the mockAliases list.",
+  "mockConfig": {
+      "aliases": "{\"ext\":{\"prebid\":{\"aliases\":{\"test1\":\"appnexus\",\"test2\":\"rubicon\",\"test3\":\"openx\"}}}}"
```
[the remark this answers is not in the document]
guscarreon: removed

endpoints/openrtb2/auction_test.go
guscarreon: Done. The code indeed looks cleaner by unmarshaling the `config`

endpoints/openrtb2/auction_test.go
guscarreon: How about `generateConfigMaps`? I'm not super happy about it, but I think it works. What do you think?

endpoints/openrtb2/sample-requests/account-required/with-account/required-blacklisted-acct.json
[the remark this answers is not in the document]
SyntaxNode: Ah ok. Subtle. I can't suggest any improvements here though. This one is already in the account required folder.

endpoints/openrtb2/sample-requests/disabled/bad/bad-alias.json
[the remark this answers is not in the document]
SyntaxNode: Ah, sorry. Not sure how I missed it.

endpoints/openrtb2/auction_test.go
guscarreon: Moved the check and refactored to load both `ExpectedBidResponse` and  `ExpectedErrorMessage` but I still assert error based on the return code. Let me know if that's the approach you had in mind.

endpoints/openrtb2/sample-requests/blacklisted/blacklisted-acct.json
```diff
@@ -1,88 +1,90 @@
 {
```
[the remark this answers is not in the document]
guscarreon: Done. Test `endpoints/openrtb2/sample-requests/blacklisted/blacklisted-site-publisher.json` was added

endpoints/openrtb2/sample-requests/disabled/good/partial.json
```diff
@@ -49,4 +54,11 @@
         }
       }
     }
-  }
\ No newline at end of file
+  },
+  "expectedBidResponse": {
+      "id":"some-request-id",
```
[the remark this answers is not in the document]
guscarreon: Pending..

SyntaxNode (commented):
Looks great. Relatively small list of minor comments.

endpoints/openrtb2/sample-requests/disabled/good/partial.json
```diff
@@ -1,5 +1,11 @@
 {
+  "description": "Request comes with some imps directed toward disabled adapters, but there's one non-disabled adapter and we expect a successful response",
+  "config": {
+    "disabledAdapters": ["appnexus", "rubicon"]
+  },
+  "mockBidRequest": {
     "id": "some-request-id",
+    "test": 1,
```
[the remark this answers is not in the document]
guscarreon: Leftover. Removed

endpoints/openrtb2/sample-requests/disabled/good/partial.json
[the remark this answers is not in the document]
guscarreon: Came back to this and implemented an exchange that lists "openx" in the `bidResponse.Ext` in order to prove that it was the only adapter to not get filtered out during validation. Please let me know if you approve of this approach.

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +39,216 @@ import (
...
+	test.Config = generateConfigMaps(test.Config)
+
+	// Run test
+	actualCode, actualBidResponse := doRequest(t, test)
+
+	// Assertions
+	assert.Equal(t, test.ExpectedReturnCode, actualCode, "Test failed. Filename: %s \n", testFile)
+
+	// Either assert bid response or expected error
+	if test.ExpectedReturnCode != 200 {
```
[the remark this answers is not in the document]
SyntaxNode: We discussed to remove the dependency on the return code and enforce only either the response or error message being specified. Something along the lines of:

```
assert.Equal(t, test.ExpectedReturnCode, actualCode, "Test failed. Filename: %s \n", testFile)

if len(test.ExpectedErrorMessage) > 0 {
  assert.True(t, strings.HasPrefix(actualBidResponse, test.ExpectedErrorMessage), "Actual: %s \nExpected: %s. Filename: %s \n", actualBidResponse, test.ExpectedErrorMessage, testFile)
}

if len(test.ExpectedBidResponse) > 0 {
  diffJson(t, testFile, []byte(actualBidResponse), test.ExpectedBidResponse)
}
```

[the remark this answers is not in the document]
guscarreon: Done

endpoints/openrtb2/sample-requests/disabled/good/partial.json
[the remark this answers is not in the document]
guscarreon: Modified as discussed. Now the exchange lists "openx" in the `bidResponse.SeatBid[i].Bid[j].ID` as "openx-bid" in order to assert the exact adapters that were not filtered in the validation steps enforced on `endpoints/openrtb2/auction.go`.

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,244 @@ import (
...
+	// Get the return code we expect PBS to throw back given test's bidRequest and config
+	parsedReturnCode, err := jsonparser.GetInt(fileData, "expectedReturnCode")
+	assert.NoError(t, err, "Error jsonparsing root.code from file %s. Desc: %v.", testFile, err)
+
+	// Get both bid response and error message, if any
+	parsedTestData.ExpectedBidResponse, _, _, err = jsonparser.Get(fileData, "expectedBidResponse")
+	parsedTestData.ExpectedErrorMessage, errEm = jsonparser.GetString(fileData, "expectedErrorMessage")
+
+	// Assert no error in parsing one or the other
+	if parsedReturnCode >= 200 && parsedReturnCode < 300 {
```
SyntaxNode, line 193: I don't think we need this condition anymore with the changes made in `runTestCase`. Maybe add an error check to ensure one of them exists and was retrieved without caring about which one?

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,244 @@ import (
...
+	}
+
+	if err := json.Unmarshal(expectedJson, &expectedBidResponse); err != nil {
+		t.Fatalf("Could not unmarshal expected bidResponse taken from test file.\n Test file: %s\n Error:%s\n", testFile, err)
+	}
+
+	assert.Equalf(t, actualBidResponse.ID, expectedBidResponse.ID, "BidResponse.ID doesn't match expected. Test: %s\n", testFile)
+	assert.Len(t, actualBidResponse.SeatBid, len(expectedBidResponse.SeatBid), "BidResponse.SeatBid array doesn't match expected. Test: %s\n", testFile)
+
+	//Given that bidResponses have the same lenght, compare them
```
SyntaxNode, line 247: Typo: lenght

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,244 @@ import (
...
+			}
+
+			var expectedBidsMap map[string]openrtb.Bid = make(map[string]openrtb.Bid, 0)
+			for _, bid := range expectedSeatBid.Bid {
+				expectedBidsMap[bid.ID] = bid
+			}
+
+			//Compare maps order in which bids were listed in the BidResponse.SeatBid[%d].Bid is irrelevant
+			assert.Equalf(t, actualBidsMap, expectedBidsMap, "BidResponse.SeatBid[].Bid doesn't match expected. Test: %s\n", testFile)
+		}
```
SyntaxNode, line 276: Can you please explain a bit what's going on in this compare? There is some magic in Go, where comparing maps with assert.Equals does an order-irrelevant deep compare.

endpoints/openrtb2/auction_test.go
```diff
@@ -1673,21 +1634,45 @@ func (e *nobidExchange) HoldAuction(ctx context.Context, bidRequest *openrtb.Bid
...
+
+		for _, imp := range bidRequest.Imp {
+			var bidderExts map[string]json.RawMessage
+
+			if err := json.Unmarshal(imp.Ext, &bidderExts); err != nil {
+				return nil, err
+			}
+
+			for bidderNameOrAlias := range bidderExts {
+				bidResponse.SeatBid[0].Bid = append(bidResponse.SeatBid[0].Bid, openrtb.Bid{ID: fmt.Sprintf("%s-bid", bidderNameOrAlias)})
```
SyntaxNode, line 1661: Nice!

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,244 @@ import (
...
+			"sample-requests/disabled/good",
+		},
+	}
+	for _, test := range testSuites {
+		testCaseFiles, err := getTestFiles(test.directory)
+		assert.NoError(t, err, "Test case %s. Error reading files from directory %s \n", test.description, test.directory)
+
+		for _, file := range testCaseFiles {
+			data, err := ioutil.ReadFile(file)
+			assert.NoError(t, err, "Test case %s. Error reading file %s \n", test.description, file)
```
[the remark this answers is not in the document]
guscarreon: pretty good catch. Added the check

endpoints/openrtb2/auction_test.go
guscarreon: I tried:
```
if err != nil && errEm != nil {
    t.Fatalf("One or the other should be an error, but not both")
}
```
But if `parsedReturnCode` is between 200 and 300, then `err` is the one that snouldn't be an error, and the other way around. I fear that the condition above could lead us to a couple of false non-error scenarios, unless I'm wrong. I think an improvement, could be the statement below, but do you think it's that much readable? The main flaw I find is the ambiguity of the `t.Fatalf` message.
```
statusOk := parsedReturnCode >= 200 && parsedReturnCode < 300
if (statusOk && err != nil) || (!statusOk && errEm != nil) {
    t.Fatalf("Test case error")
}
```
What do you think?

endpoints/openrtb2/auction_test.go
guscarreon: The main reason I had to create this function is accurately compare even if `[]SeatBid` array and the `[]Bid` arrays within the `SeatBid` elements come in different order. Function `assert.Equals(t, expBidResponse, actualBidResponse)` fails if the `SeatBid` elements are listed in different order even if they are identical. Simply comparing `expectedSeatBidsMap` vs `actualSeatBidsMap` with `assert.Equals(...)` could also lead to false negatives if the elements in their `[]Bid` arrays come in different order too. To clarify further I added the `TestBidRequestAssert` test case. Please let me know if you believe there's a simpler way to compare accurately independent of both the `[]SeatBid` array element order and their respective `[]Bid` array element order or if this is even desired.

I refactored this test case a bit and added more comments that probably make it a bit more readable.

SyntaxNode: Ah, ok. The normal compares fail due to an array embedded within the map. Fun. I can't think of a better way to do this.

endpoints/openrtb2/auction_test.go
SyntaxNode: Enforcing the existence of an expectation based on a response code seems like we're building a framework that is too opinionated - which could lead to its own issues down the line. I would check that both err and errEm are either nil or KeyPathNotFoundError (from the jsonparser package).

guscarreon: Good point. I rewrote the conditional to not rely on `parsedReturnCode` anymore.

SyntaxNode (approved):

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,329 @@ import (
...
+		assertBidResponseEqual(t, testFile, expectedBidResponse, actualBidResponse)
+	}
+}
+
+func parseTestFile(t *testing.T, fileData []byte, testFile string) testCase {
+	t.Helper()
+
+	parsedTestData := testCase{}
+	var err, errEm error
+
```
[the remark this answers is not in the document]
SyntaxNode: `jsonparser` doesn't perform validation and `encoding/json` does. Some of the tests currently rely on improperly formatted json to test failure paths.

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,329 @@ import (
...
+type testConfigValues struct {
+	AccountRequired       bool     `json:"accountRequired"`
+	AliasJSON             string   `json:"aliases"`
+	BlacklistedAccounts   []string `json:"blacklistedAccts"`
+	BlacklistedApps       []string `json:"blacklistedApps"`
+	AdapterList           []string `json:"disabledAdapters"`
+	ListBiddersInResponse bool     `json:"listBiddersInResponse"`
+
+	blacklistedAccountMap map[string]bool
```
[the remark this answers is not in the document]
guscarreon: Done

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,329 @@ import (
...
+		{
+			"Asserts we return 503s on requests with blacklisted accounts and apps.",
+			"sample-requests/blacklisted",
+		},
+		{
+			"Assert that requests that come with no user id nor app id return error if the `AccountRequired` field in the `config.Configuration` structure is set to true",
+			"sample-requests/account-required/no-account",
+		},
+		{
+			"Assert requests that come with a valid user id nor app id when account is not required",
```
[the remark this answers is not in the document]
guscarreon: Done

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,329 @@ import (
...
+						Seat: "seat1",
+						Bid:  []openrtb.Bid{rubiconB1, appnexusB1},
+					},
+				},
+			},
+		},
+	}
+
+	for _, test := range testSuites {
+		assertBidResponseEqual(t, test.description, test.expectedBidResponse, test.actualBidResponse)
```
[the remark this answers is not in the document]
guscarreon: I don't think it hurts

endpoints/openrtb2/sample-requests/invalid-native/eventtracker-methods-large.json
```diff
@@ -1,34 +1,25 @@
 {
-    "context": 1,
-    "plcmttype": 1,
-    "assets": [
-        {
-            "title": {
-                "len": 90
-            }
+  "description": "Bid request with empty methods array inside the native.request field in its only imp element",
```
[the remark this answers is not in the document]
guscarreon: I assume this file was meant to have a `"methods"` array with an element holding a value outside of the allowed scope  (1 or 2). File has been corrected, as well as `endpoints/openrtb2/sample-requests/invalid-native/eventtracker-type-large.json` that was testing the same functionality and now is testing for a large `"event"` field value. I believe `endpoints/openrtb2/sample-requests/invalid-native/eventtracker-event-large.json` was a much suitable name for `endpoints/openrtb2/sample-requests/invalid-native/eventtracker-type-large.json` so I renamed.

endpoints/openrtb2/sample-requests/invalid-native/eventtracker-type-large.json
```diff
@@ -1,34 +1,25 @@
...
+    "id": "req-id",
+    "site": {
+      "page": "some.page.com"
+    },
+    "tmax": 500,
+    "imp": [
+      {
+        "id": "some-imp",
+        "native": {
+          "request": "{\"context\":1,\"plcmttype\":1,\"assets\":[{\"title\":{\"len\":90}},{\"img\":{\"hmin\":30,\"wmin\":20}},{\"video\":{\"mimes\":[\"video/mp4\"],\"minduration\":5,\"maxduration\":10,\"protocols\":[1]}},{\"data\":{\"type\":2}}],\"eventtrackers\":[{\"event\":1,\"methods\":[5]}]}"
```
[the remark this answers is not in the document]
guscarreon: I believe `endpoints/openrtb2/sample-requests/invalid-native/eventtracker-type-large.json` was meant to test an out-of-scope value on the `"event"` field. I corrected the file and renamed to `endpoints/openrtb2/sample-requests/invalid-native/eventtracker-type-large.json`.

endpoints/openrtb2/sample-requests/invalid-stored/bad_incoming_1.json
```diff
@@ -1,41 +1,42 @@
...
-          "ext": {
-            "appnexus": {
-              "placementId": 12883451
-            }
+  "description": "Otherwise valid request using stored request; incoming request has a comma after the closing curly brace of the second set of dimensions to yield invalid JSON",
+
+  "mockBidRequest": {
+    "id": "some-request-id",
+    "site": {
+      "page": "prebid.org"
```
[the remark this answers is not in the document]
guscarreon: I believe this file shouldn't be moved because files in this directories seem give coverage to different parts of the code. For instance, this test covers an error scenario in the `processStoredRequests` funciton:
```
1072 func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, []error) {
1073 +-- 22 lines: Parse the Stored Request IDs from the BidRequest and Imps.--------------------------
1095     if hasStoredBidRequest {
1096         resolvedRequest, err = jsonpatch.MergePatch(storedRequests[storedBidRequestId], requestJson)
1097         if err != nil {
1098             hasErr, Err := getJsonSyntaxError(requestJson)
1099             if hasErr {
1100                 err = fmt.Errorf("Invalid JSON in Incoming Request: %s", Err)
endpoints/openrtb2/auction.go
```
Malformed Json was simplified  a bit.

endpoints/openrtb2/sample-requests/invalid-stored/bad_incoming_2.json
```diff
@@ -1,41 +1,42 @@
...
-          },
-          "ext": {
-            "appnexus": {
-              "placementId": 12883451
+  "description": "Otherwise valid request using stored request; incoming request lacks a comma after first width to yield invalid JSON",
+
+  "mockBidRequest": {
+    "id": "some-request-id",
+    "site": {
+      "page": "prebid.org"
```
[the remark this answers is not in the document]
guscarreon: Good catch, removed file

endpoints/openrtb2/sample-requests/invalid-stored/bad_incoming_imp.json
```diff
@@ -36,5 +35,7 @@
       }
     ],
     "tmax": 500
-  }
+  },
+  "expectedReturnCode": 400,
```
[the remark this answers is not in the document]
guscarreon: The stored request seems to be defined at the `imp.ext` level instead of the `root.ext` level:

endpoints/openrtb2/sample-requests/invalid-whole/banner-hmin.json
```diff
@@ -1,17 +1,19 @@
...
-          "banner": {
-            "hmin":50
-          }
+  "description": "Request comes with unsupported property 'hmin'",
+  "mockBidRequest": {
+    "id":"req-id",
+    "imp": [
+      {
+        "id":"imp-id",
+        "banner": {
```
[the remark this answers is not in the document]
guscarreon: Do you mean `endpoints/openrtb2/sample-requests/invalid-whole/banner-h-zero.json` is the same as `endpoints/openrtb2/sample-requests/invalid-whole/banner-hmin.json` ?? Files do seem to  be different. Did you mean a different file?

endpoints/openrtb2/sample-requests/invalid-whole/banner-w-zero.json
```diff
@@ -19,5 +19,7 @@
         }
       }
     ]
-  }
+  },
+  "expectedReturnCode": 400,
+  "expectedErrorMessage": "Invalid request: request.imp[0].banner has no sizes. Define \"w\" and \"h\", or include \"format\" elements.\n"
```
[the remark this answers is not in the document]
guscarreon: Good catch, removed both `endpoints/openrtb2/sample-requests/invalid-whole/banner-h-zero.json` and `endpoints/openrtb2/sample-requests/invalid-whole/banner-w-zero.json`

endpoints/openrtb2/sample-requests/invalid-whole/banner-wmax.json
```diff
@@ -1,6 +1,6 @@
 {
-  "message": "Invalid request: request.imp[0].banner uses unsupported property: \"wmax\". Use the \"format\" array instead.\n",
-  "requestPayload": {
+  "description": "Request comes with unsupported property 'wmax'",
+  "mockBidRequest": {
```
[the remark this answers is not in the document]
guscarreon: I believe we discussed this offline and agreed to keep this file.

endpoints/openrtb2/sample-requests/invalid-whole/banner-wmin.json
```diff
@@ -13,5 +13,7 @@
     "app": {
       "id": "app_001"
     }
-  }
+  },
```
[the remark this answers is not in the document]
guscarreon: We agreed to keep this file as it asserts that we don't support this `wmin` property

endpoints/openrtb2/sample-requests/invalid-whole/format-no-hratio.json
```diff
@@ -1,21 +1,23 @@
...
-			}
-		],
-        "app": {
-          "id": "app_001"
-        }
-	}
+  "description": "Banner comes with a zero hratio value",
+  "mockBidRequest": {
+  		"id": "req-id",
+  		"imp": [
```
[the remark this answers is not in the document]
guscarreon: done

endpoints/openrtb2/sample-requests/invalid-whole/interstitial-empty.json
```diff
@@ -1,43 +1,45 @@
...
-                        "minheightperc": 60
-                    }
-                }
-            }
+  "description": "Bid request device field comes with invalid interstitial value",
+  "mockBidRequest": {
+    "id": "some-request-id",
+    "site": {
+      "page": "test.somepage.com"
+    },
```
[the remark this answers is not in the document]
guscarreon: I believe the file name does make sense. The description, on the other hand, was wrong. Changed it to "Bid request banner field comes with no data to set interstitial size list"

endpoints/openrtb2/sample-requests/invalid-whole/native-empty.json
```diff
@@ -1,6 +1,6 @@
 {
-  "message": "Invalid request: request.imp[0].native missing required property \"request\"\n",
-  "requestPayload": {
+  "description": "Bid request's sole imp element has empty Native field and does not define a Banner, Video nor Audio field",
```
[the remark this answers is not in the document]
guscarreon: They are not the same and they do cover two different error scenarios, but I added a more descriptive expected error message.
```
"native": {}     |"native": {
                 |  "request": "{}"
                 |},
```

endpoints/openrtb2/sample-requests/valid-native/sample-v1.1.json
```diff
@@ -1,31 +1,41 @@
...
-    "assets": [
-        {
-            "title": {
-                "len": 90
-            }
+  "description": "Well formed native request with video asset",
+  "mockBidRequest": {
+    "id": "req-id",
+    "site": {
+      "page": "some.page.com"
```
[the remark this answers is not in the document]
guscarreon: Renamed

endpoints/openrtb2/sample-requests/valid-native/sample-v1.2.json
```diff
@@ -1,34 +1,41 @@
...
+    "id": "req-id",
+    "site": {
+      "page": "some.page.com"
+    },
+    "tmax": 500,
+    "imp": [
+      {
+        "id": "some-imp",
+        "native": {
+          "request": "{\"context\":1,\"plcmttype\":1,\"assets\":[{\"title\":{\"len\":90}},{\"img\":{\"hmin\":30,\"wmin\":20}},{\"video\":{\"mimes\":[\"video/mp4\"],\"minduration\":5,\"maxduration\":10,\"protocols\":[1]}},{\"data\":{\"type\":2}}],\"eventtrackers\":[{\"event\":1,\"methods\":[1]}]}"
```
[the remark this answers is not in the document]
guscarreon: The `request` field is not the same, notice the `\"contextsubtype\":10,` is missing in `endpoints/openrtb2/sample-requests/valid-native/sample-v1.2.json `

SyntaxNode (commented):
You got the `swp` file in here again. Consider adding it your machine's git ignore.

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,321 @@ import (
...
+
+	//Assert non-array BidResponse fields
+	assert.Equalf(t, expectedBidResponse.ID, actualBidResponse.ID, "BidResponse.ID doesn't match expected. Test: %s\n", testFile)
+	assert.Equalf(t, expectedBidResponse.BidID, actualBidResponse.BidID, "BidResponse.BidID doesn't match expected. Test: %s\n", testFile)
+	assert.Equalf(t, expectedBidResponse.NBR, actualBidResponse.NBR, "BidResponse.NBR doesn't match expected. Test: %s\n", testFile)
+
+	//Assert []SeatBid and their Bid elements independently of their order
+	assert.Len(t, actualBidResponse.SeatBid, len(expectedBidResponse.SeatBid), "BidResponse.SeatBid array doesn't match expected. Test: %s\n", testFile)
+
+	//Given that bidResponses have the same lenght, compare them in an order-independent way using maps
```
mansinahar, line 252: Typo: `have the same length`

endpoints/openrtb2/auction_test.go
```diff
@@ -121,231 +424,42 @@ func TestExplicitUserId(t *testing.T) {
...
+	endpoint, _ := NewEndpoint(
+		&mockBidExchange{},
+		newParamsValidator(t),
+		&mockStoredReqFetcher{},
+		empty_fetcher.EmptyFetcher{},
+		empty_fetcher.EmptyFetcher{},
+		&config.Configuration{
+			MaxRequestSize:  maxSize,
+			BlacklistedApps: test.Config.BlacklistedApps,
+			//BlacklistedAppMap:  test.Config.blacklistedAppMap,
```
mansinahar, line 444: Can we please remove this comments? Looks like a leftover

endpoints/openrtb2/auction_test.go
```diff
@@ -36,19 +37,321 @@ import (
...
+		{
+			"Asserts we return 503s on requests with blacklisted accounts and apps.",
+			"blacklisted",
+		},
+		{
+			"Assert that requests that come with no user id nor app id return error if the `AccountRequired` field in the `config.Configuration` structure is set to true",
+			"account-required/no-account",
+		},
+		{
+			"Assert requests that come with a valid user id nor app id when account is required",
```
mansinahar, line 90: Is the `nor` a typo here? This looks like it should be:
```
Assert requests that come with a valid user id or app id when account is required
```

endpoints/openrtb2/sample-requests/valid-native/sample-v1.2.json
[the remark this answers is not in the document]
mansinahar: Ah ok! Can you please rename this file though? `sample-v1-2.json` doesn't make much sense

mansinahar:
@guscarreon except for the very minor comments above, I think this is in a pretty good state. Really great job refactoring this 👏

endpoints/openrtb2/auction_test.go
guscarreon: I can't spell "length". Corrected

endpoints/openrtb2/sample-requests/valid-native/sample-v1.2.json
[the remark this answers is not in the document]
guscarreon: Makes sense. Renamed to `video-asset-event-tracker.json`

endpoints/openrtb2/auction_test.go
guscarreon: Yep, typo. Corrected

mansinahar (approved):

SyntaxNode (approved):

PR #1452, opened 2020-08-21
Add bidder name key support

camrice:
This PR also contains a bugfix implemented by @VeronikaSolovei9

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -621,7 +621,7 @@ func createBidExtension(videoRequest *openrtb_ext.BidRequestVideo) ([]byte, erro
 		IncludeBrandCategory: inclBrandCat,
 		DurationRangeSec:     durationRangeSec,
 		IncludeBidderKeys:    true,
-		AddBidderNames:       videoRequest.AddBidderNames,
+		AppendBidderNames:    videoRequest.AppendBidderNames,
 	}
 
 	vastXml := openrtb_ext.ExtRequestPrebidCacheVAST{}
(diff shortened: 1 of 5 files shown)


SyntaxNode:
Can you please describe the bug this is fixing?

VeronikaSolovei9:
@SyntaxNode @camrice we found a bug on bid deduplication logic. In case we find bids with same category or `pb_dur` we randomly deduplicate one of them. This can bid that was previously processed or bid on current iteration. In case we deduplicate bid from previous iteration and this is the only one bid in bidder - we need to remove the entire bidder from processing. Without this fix we removed "current iteration" bidder  instead of "previously processed" . bidder.

The fix is only  in one line: `seatBidsToRemove = append(seatBidsToRemove, dupe.bidderName)` in exchange.go. Unit test added.

SyntaxNode:
If it's a bug though, why are you adding a new Ext input?

I'm concerned about changing the public contract without first discussing in a Prebid committee. Before we add this, I'd like for you to please create an issue in this project so at least the PBS committee can review it at the next meeting. I'm not a remember of the Video Committee, but hopefully you are and can discuss it there as well.

camrice:
Understood! I can create the issue for adding the new field. We should probably split this PR up then right? The bugfix is certainly higher priority.

SyntaxNode:
> Understood! I can create the issue for adding the new field. We should probably split this PR up then right? The bugfix is certainly higher priority.

Yes, please. If you split them up we can address the bug separately from the new feature.

camrice:
Got it! I will close this one and have new PRs created.

PR #1453, opened 2020-08-21
Consolidate StoredRequest configs, add validation for all data types

laurb9:
- Includes #1446 
- Drop the "Slim" versions of stored request configuration
- Unify stored_requests, categories, stored_video_req configuration
- Add stored_amp, still built from stored_requests but moved to config. Should eventually be its own section.
- Added validation for categories and stored_video_req (previously missing).
- Net fewer lines of code than before

Testing validation for stored_video_req with this bad config
```yaml
stored_video_req:
  in_memory_cache:
    type: none
  cache_events:
    enabled: true
    endpoint: "/foo"
  postgres:
    connection:
      dbname: "blah"
    fetcher:
      query: "blah"
    poll_for_updates:
      query: "blah"
      timeout_ms: 0
```

Before the change, prebid-server starts and can crash:
```
I0821 18:22:59.943387   50156 config.go:50] Connecting to Postgres for Stored Video. DB=blah, host=, port=0, user=
F0821 18:22:59.947503   50156 config.go:282] Failed to ping Video postgres: dial tcp 127.0.0.1:5432: connect: connection refused
goroutine 1 [running]:
github.com/golang/glog.stacks(0xc00048d600, 0xc000138280, 0x81, 0xa0)
```

After the change:
```
F0821 18:19:50.035239   50048 main.go:37] Configuration could not be loaded or did not pass validation: validation errors are:

  stored_video_req: postgres.poll_for_updates.refresh_rate_seconds must be > 0
  stored_video_req: postgres.poll_for_updates.timeout_ms must be > 0
  stored_video_req: postgres.poll_for_updates.query must contain exactly one wildcard
  stored_video_req: cache_events must be disabled if in_memory_cache=none
  stored_video_req: postgres.poll_for_updates.query must be empty if in_memory_cache=none
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -103,7 +103,9 @@ func (c configErrors) Error() string {
 func (cfg *Configuration) validate() configErrors {
 	var errs configErrors
 	errs = cfg.AuctionTimeouts.validate(errs)
-	errs = cfg.StoredRequests.validate(errs)
+	errs = cfg.StoredRequests.validate("stored_requests", RequestDataType, errs)
+	cfg.CategoryMapping.DataType = CategoryDataType // CategoryMapping has no other validation
+	cfg.StoredVideo.DataType = VideoDataType        // StoredVideo has no other validation
 	errs = cfg.Metrics.validate(errs)
 	if cfg.MaxRequestSize < 0 {
 		errs = append(errs, fmt.Errorf("cfg.max_request_size must be >= 0. Got %d", cfg.MaxRequestSize))
--- a/config/config.go
+++ b/config/config.go
@@ -29,17 +29,17 @@ type Configuration struct {
 	EnableGzip  bool       `mapstructure:"enable_gzip"`
 	// StatusResponse is the string which will be returned by the /status endpoint when things are OK.
 	// If empty, it will return a 204 with no content.
-	StatusResponse  string          `mapstructure:"status_response"`
-	AuctionTimeouts AuctionTimeouts `mapstructure:"auction_timeouts_ms"`
-	CacheURL        Cache           `mapstructure:"cache"`
-	ExtCacheURL     ExternalCache   `mapstructure:"external_cache"`
-	RecaptchaSecret string          `mapstructure:"recaptcha_secret"`
-	HostCookie      HostCookie      `mapstructure:"host_cookie"`
-	Metrics         Metrics         `mapstructure:"metrics"`
-	DataCache       DataCache       `mapstructure:"datacache"`
-	StoredRequests  StoredRequests  `mapstructure:"stored_requests"`
-	StoredAmp       StoredRequests  `mapstructure:"stored_amp_req"`
-	CategoryMapping StoredRequests  `mapstructure:"category_mapping"`
+	StatusResponse    string          `mapstructure:"status_response"`
+	AuctionTimeouts   AuctionTimeouts `mapstructure:"auction_timeouts_ms"`
+	CacheURL          Cache           `mapstructure:"cache"`
+	ExtCacheURL       ExternalCache   `mapstructure:"external_cache"`
+	RecaptchaSecret   string          `mapstructure:"recaptcha_secret"`
+	HostCookie        HostCookie      `mapstructure:"host_cookie"`
+	Metrics           Metrics         `mapstructure:"metrics"`
+	DataCache         DataCache       `mapstructure:"datacache"`
+	StoredRequests    StoredRequests  `mapstructure:"stored_requests"`
+	StoredRequestsAMP StoredRequests  `mapstructure:"stored_amp_req"`
+	CategoryMapping   StoredRequests  `mapstructure:"category_mapping"`
 	// Note that StoredVideo refers to stored video requests, and has nothing to do with caching video creatives.
 	StoredVideo StoredRequests `mapstructure:"stored_video_req"`
 
@@ -105,7 +105,7 @@ func (cfg *Configuration) validate() configErrors {
 	var errs configErrors
 	errs = cfg.AuctionTimeouts.validate(errs)
 	errs = cfg.StoredRequests.validate("stored_req", errs)
-	errs = cfg.StoredAmp.validate("stored_amp_req", errs)
+	errs = cfg.StoredRequestsAMP.validate("stored_amp_req", errs)
 	errs = cfg.CategoryMapping.validate("categories", errs)
 	errs = cfg.StoredVideo.validate("stored_video_req", errs)
 	errs = cfg.Metrics.validate(errs)
--- a/config/config.go
+++ b/config/config.go
@@ -29,18 +29,19 @@ type Configuration struct {
 	EnableGzip  bool       `mapstructure:"enable_gzip"`
 	// StatusResponse is the string which will be returned by the /status endpoint when things are OK.
 	// If empty, it will return a 204 with no content.
-	StatusResponse  string             `mapstructure:"status_response"`
-	AuctionTimeouts AuctionTimeouts    `mapstructure:"auction_timeouts_ms"`
-	CacheURL        Cache              `mapstructure:"cache"`
-	ExtCacheURL     ExternalCache      `mapstructure:"external_cache"`
-	RecaptchaSecret string             `mapstructure:"recaptcha_secret"`
-	HostCookie      HostCookie         `mapstructure:"host_cookie"`
-	Metrics         Metrics            `mapstructure:"metrics"`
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -29,18 +29,19 @@ type Configuration struct {
...
+	StatusResponse  string          `mapstructure:"status_response"`
+	AuctionTimeouts AuctionTimeouts `mapstructure:"auction_timeouts_ms"`
+	CacheURL        Cache           `mapstructure:"cache"`
+	ExtCacheURL     ExternalCache   `mapstructure:"external_cache"`
+	RecaptchaSecret string          `mapstructure:"recaptcha_secret"`
+	HostCookie      HostCookie      `mapstructure:"host_cookie"`
+	Metrics         Metrics         `mapstructure:"metrics"`
+	DataCache       DataCache       `mapstructure:"datacache"`
+	StoredRequests  StoredRequests  `mapstructure:"stored_requests"`
+	StoredAmp       StoredRequests  `mapstructure:"stored_amp_req"`
```
SyntaxNode, line 41: Per Go naming convention, this should be `StoredAMP`. That does a bit strange though, so perhaps `StoredRequestsAMP`?

config/config.go
```diff
@@ -103,7 +104,10 @@ func (c configErrors) Error() string {
 func (cfg *Configuration) validate() configErrors {
 	var errs configErrors
 	errs = cfg.AuctionTimeouts.validate(errs)
-	errs = cfg.StoredRequests.validate(errs)
+	errs = cfg.StoredRequests.validate("stored_req", errs)
```
SyntaxNode, line 107: This PR is build on top of https://github.com/prebid/prebid-server/pull/1446. In that PR, I recommend moving where `DataType` lives. Applying those comments to this PR, I think `validate` would be a good spot to pass down the data types.

Due to the dependency, let's try to resolve the other PR first.

config/config.go
```diff
@@ -989,6 +998,24 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetEnvPrefix("PBS")
 	v.AutomaticEnv()
 	v.ReadInConfig()
+
+	// Migrate config settings to maintain compatibility with old configs
+	migrateConfig(v)
+}
+
+func migrateConfig(v *viper.Viper) {
+	//v.RegisterAlias("stored_requests.directorypath", "stored_requests.filesystem.directorypath")
```
SyntaxNode, line 1007: Why is this commented out?

config/config.go
```diff
@@ -989,6 +998,24 @@ func SetupViper(v *viper.Viper, filename string) {
...
+
+	// if stored_requests.filesystem is boolean in conf file, migrate to new struct to avoid breaking viper
+	switch v.Get("stored_requests.filesystem").(type) {
```
SyntaxNode, line 1010: A switch statement with a single case looks a bit odd. Consider replacement with an `if` statement.

config/config.go
```diff
@@ -989,6 +998,24 @@ func SetupViper(v *viper.Viper, filename string) {
...
+	case bool:
+		glog.Warning("stored_requests.filesystem should be changed to stored_requests.filesystem.enabled")
+		glog.Warning("stored_requests.directorypath should be changed to stored_requests.filesystem.directorypath")
+		m := v.GetStringMap("stored_requests.filesystem")
```
SyntaxNode, line 1014: Did you test if this works with the environment variable configs too?

config/stored_requests.go
```diff
@@ -107,48 +78,64 @@ type FileFetcherConfig struct {
...
 	AmpEndpoint string `mapstructure:"amp_endpoint"`
 }
 
-func (cfg *StoredRequests) validate(errs configErrors) configErrors {
+// Migrate combined stored_requests+amp configuration to separate simple config sections
+func resolvedStoredRequestsConfig(cfg *Configuration) {
+	sr := &cfg.StoredRequests
+	amp := &cfg.StoredAmp
+
+	sr.CacheEvents.Endpoint = "/storedrequests/openrtb2" // why is this here and not SetDefault ?
```
SyntaxNode, line 92: No idea. Older code perhaps? I'm fine with moving it.

config/config.go
laurb9: `StoredRequestsAMP` sounds good.

config/config.go
laurb9: I'll remove it. This would have been the natural method to migrate config settings so I tried it first, but it does not work due to `stored_requests.filesystem` having been promoted from a bool to a map.

config/config.go
laurb9: Good catch. It fails miserably with `PBS_STORED_REQUESTS_FILESYSTEM_ENABLED`.

config/stored_requests.go
```diff
@@ -11,44 +10,31 @@ import (
...
+	CategoryDataType   DataType = "Category"
+	VideoDataType      DataType = "Video"
+	AMPRequestDataType DataType = "Amp Request"
+)
+
+// StoredRequests struct defines options for stored requests for each data type
+// including some amp stored_requests options
+type StoredRequests struct {
+	// DataType is a tag pushed from upstream indicating the type of object fetched here
+	DataType DataType
```
SyntaxNode, line 27: We talked about this a bit offline, including making the DataType private and using an accessor - which is better, but I feel this is still mixing the concerns of config and context. Please consider a refactor where the host config is separated from the mix of host and hardcoded values.

stored_requests/config/config.go
```diff
@@ -269,32 +225,37 @@ func newHttpEvents(client *http.Client, timeout time.Duration, refreshRate time.
...
 // consolidate returns a single Fetcher from an array of fetchers of any size.
-func consolidate(fetchers []stored_requests.AllFetcher) stored_requests.AllFetcher {
+func consolidate(dataType config.DataType, fetchers []stored_requests.AllFetcher) stored_requests.AllFetcher {
 	if len(fetchers) == 0 {
-		glog.Warning("No Stored Request support configured. request.imp[i].ext.prebid.storedrequest will be ignored. If you need this, check your app config")
+		switch dataType {
+		case config.RequestDataType:
+			glog.Warning("No Stored Request support configured. request.imp[i].ext.prebid.storedrequest will be ignored. If you need this, check your app config")
+		default:
+			glog.Warningf("No Stored %s support configured. If you need this, check your app config", dataType)
```
SyntaxNode, line 257: This should be unreachable in practice, right?

laurb9: It does happen with the default config for video section, and the original error was confusing too as I've shown in #1446 

Original startup messages:

>...
No Stored Request support configured. request.imp[i].ext.prebid.storedrequest will be ignored. If you need this, check your app config
No Stored Request cache configured. The Fetcher backend will be used for all Stored Requests.

After the change:

>...
No Stored Video support configured. If you need this, check your app config
No Stored Video cache configured. The Video Fetcher backend will be used for all data requests

SyntaxNode: Ah, gotcha. Much improved. Thank you.

config/stored_requests.go
laurb9: I have been thinking about how to do this in a concise way. The intention is that any method looking at this config can tell what it is about, and the current code is pretty clear about that. Wrapping a struct around it can also achieve this but adds complexity to any sender which must do the wrapping.

For now, I made `dataType` private with accessor (commit a603974), as a compromise. I've updated #1426 to use the same approach (https://github.com/prebid/prebid-server/pull/1426/commits/dc8a74d1bab82b87095832e78e284dadc8e88932) and probably the blacklist maps can eventually use this too.

config/stored_requests.go
```diff
@@ -20,11 +20,15 @@ const (
 	AMPRequestDataType DataType = "Amp Request"
 )
 
+func (sr *StoredRequests) GetDataType() DataType {
```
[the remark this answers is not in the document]
laurb9: Would be my preference as well, will change

config/stored_requests.go
```diff
@@ -11,44 +10,35 @@ import (
...
-	// If non-nil, the server will use those endpoints to populate and update the cache.
-	HTTPEvents HTTPEventsConfig `mapstructure:"http_events"`
+// DataType constants
+type DataType string
+
+const (
+	RequestDataType    DataType = "Request"
+	CategoryDataType   DataType = "Category"
+	VideoDataType      DataType = "Video"
+	AMPRequestDataType DataType = "Amp Request"
```
SyntaxNode, line 20: Nitpick: Could you change the string to `"AMP Request"`?.

PR #1454, opened 2020-08-22
Delete digitrust.json

patmmccann:


SyntaxNode:
@patmmccann Why?

patmmccann:
I was looking at the supplementary examples on github and I was thinking,
feels strange and messy to be validating deprecated functionality.
Digitrust is no longer a thing,

hhhjort:
I would think it is better to remove the test when we remove the code implementing the functionality. Even if a feature is deprecated, there is some expectation that the function should continue working to support legacy implementations until the feature is officially removed. Unless all this test does is confirm that PBS doesn't randomly freak out if a digitrust ID happens to exist in a request.

SyntaxNode:
> Digitrust is no longer a thing,

I've never been familiar with Digitrust, but I didn't realize it was dead. Is it officially unsupported or no longer used in practice?

patmmccann:
It is officially unsupported by the IAB

patmmccann:
Here is the IAB PR that removed it from JS

https://github.com/prebid/Prebid.js/pull/5334

hhhjort:
There are a lot of other references to digitrust in the code. I am not worried about the adapters who do something with it, they can update as they wish when they get to it.

endpoints/amp_auction_test.go : has additional tests that include digitrust.
endpoints/auction.go : `validateUser()` has some validation around digitrust.
endpoints/openrtb2/sample-requests/valid-whole/digitrust.json : exists should be renamed and edited if it is otherwise a useful test, or removed.
exchange/exchange_test.go : a couple of tests reference digitrust, we should perhaps remove the digitrust part of the sample requests.
exchange/utils_test.go : a couple of tests reference digitrust, we should perhaps remove the digitrust part of the sample requests.
exchange/utils.go : `extractBuyerUIDs()` is digitrust aware. We may want to excise that code.
exchange/exchangetest/request-other-user-ext.json : Modify or remove test?
exchange/exchangetest/request-user-no-prebid.json : Modify or remove test?
openrtb_ext/user.go : We have the digitrust spec defined in the `user.ext` definitions.
privacy/scrubber.go : We have code handling digitrust.
privacy/scrubber_test.go : We have tests covering digitrust.

If we are removing digitrust support, we should remove the code before we remove tests that verify that the existing code works. I would also rather see it go in an organized initial attempt, rather than just picking off pieces here and there when we notice them. The latter strategy has the danger of removing something some other part of the code was indirectly depending on, and thus causing unforeseen issues when someone sends us a request with an old digitrust ID.

SyntaxNode:
@patmmccann I think you opened up a can of worms here. :) Would you like to code up the complete removal of Digitrust? .. or should we move this to an issue?

patmmccann:
Hans makes a compelling case for one shot removal; I initially thought
incremental removal would be fine. Seems reassonable to close the PR
and open a removal issue to me.

SyntaxNode:
Forked conversation to https://github.com/prebid/prebid-server/issues/1464

PR #1456, opened 2020-08-25
Fix bid dedup

VeronikaSolovei9:
In case we find bids with same category or pb_dur we randomly deduplicate one of them. This can a bid that was previously processed or bid on current iteration. In case we deduplicate bid from previous iteration and this is the only one bid in bidder - we need to remove the entire bidder from processing. Without this fix we removed "current iteration" bidder instead of "previously processed" . bidder.

### DIFF
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -660,7 +660,7 @@ func applyCategoryMapping(ctx context.Context, requestExt *openrtb_ext.ExtReques
 						// An older bid from a different seatBid we've already finished with
 						oldSeatBid := (seatBids)[dupe.bidderName]
 						if len(oldSeatBid.bids) == 1 {
-							seatBidsToRemove = append(seatBidsToRemove, bidderName)
+							seatBidsToRemove = append(seatBidsToRemove, dupe.bidderName)
 							rejections = updateRejections(rejections, dupe.bidID, "Bid was deduplicated")
 						} else {
 							oldSeatBid.bids = append(oldSeatBid.bids[:dupe.bidIndex], oldSeatBid.bids[dupe.bidIndex+1:]...)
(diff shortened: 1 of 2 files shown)


exchange/exchange.go
```diff
@@ -660,7 +660,7 @@ func applyCategoryMapping(ctx context.Context, requestExt *openrtb_ext.ExtReques
 						// An older bid from a different seatBid we've already finished with
 						oldSeatBid := (seatBids)[dupe.bidderName]
 						if len(oldSeatBid.bids) == 1 {
-							seatBidsToRemove = append(seatBidsToRemove, bidderName)
+							seatBidsToRemove = append(seatBidsToRemove, dupe.bidderName)
```
guscarreon, line 663: image

Current test cases, including the recently added `TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice`, do not seem to assert the recently modified functionality in line 663. Can we add coverage?

SyntaxNode: Good catch.

VeronikaSolovei9: Execution of this part of cade is controlled by random number generation `rand.Intn(100)`:
```
if rand.Intn(100) < 50 {
      dupeBidPrice = -1
} else {
     currBidPrice = -1
}
```
which always gives same value on the first local run. 

In CI unit tests run with race condition: `./validate.sh --nofmt --cov --race 10` and without fix in exchange.go CI build fails with error from newly added unit test: 
https://travis-ci.org/github/prebid/prebid-server/jobs/721078338

To make it more explicitly I added for loop to execute `applyCategoryMapping` in test several times in a row to "force" `rand.Intn(100)` return different values.

PR #1457, opened 2020-08-25
EMX Digital: Adding tests for validating TCF v1.1 and TCF v2 consent string

EMXDigital:
Adding better testing coverage for GDPR TCF compliance

### DIFF
--- a/adapters/emx_digital/usersync_test.go
+++ b/adapters/emx_digital/usersync_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestEMXDigitalSyncer(t *testing.T) {
+func TestEMXDigitalSyncerTCF1(t *testing.T) {
 	syncURL := "https://cs.emxdgt.com/um?ssp=pbs&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redirect=localhost%2Fsetuid%3Fbidder%3Demx_digital%26uid%3D%24UID"
 	syncURLTemplate := template.Must(
 		template.New("sync-template").Parse(syncURL),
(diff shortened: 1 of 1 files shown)


adapters/emx_digital/usersync_test.go
```diff
@@ -27,9 +27,37 @@ func TestEMXDigitalSyncer(t *testing.T) {
 		},
 	})
 
+	// Validate TCFV1 consent string
+	assert.NoError(t, gdpr.ValidateConsent("BOPVK28OVJoTBABABAENBs-AAAAhuAKAANAAoACwAGgAPAAxAB0AHgAQAAiABOADkA"))
 	assert.NoError(t, err)
```
guscarreon, line 32: Can we rename test to `TestEMXDigitalSyncerTCF1` to be consistent with the test below?

EMXDigital:
Pushed up new func name

SyntaxNode:
Thank you for taking GDPR seriously and keeping quality top of mind.

The user sync tests are only very loosely related to TCF1 and TCF2. This test is designed to ensure the user sync url can be properly resolved with the template engine. The value of the privacy related strings are inconsequential here. You could replace the GDPR consent string with the simple string `"A"` and the test would be just as valid, and perhaps that would be a good idea.

The `gdpr.ValidateConsent` method is a thin wrapper on top of go-gdpr and already has extensive test coverage. There is no need to validate consent strings as part of this test. If you'd like to enhance the TCF parsing tests, you're welcome to also contribute to https://github.com/prebid/go-gdpr.

SyntaxNode:
@EMXDigital Checking in to see if you wanted to discuss my feedback, intend to update this test to use obviously fake values, or close this PR. Thank you.

SyntaxNode (changes requested):
There is no need to call `gdpr.ValidateConsent` as that's only testing the hardcoded value. There's no need to separate TCF1 from TCF2 here, since it's just testing string replacement.

SyntaxNode:
Closing due to inactivity from author. If you get back to this in the future, please feel free to submit a new PR or reopen this one.

PR #1460, opened 2020-08-26
Add Scheme Option To External Cache URL

SyntaxNode:
Per the Prebid.org Slack thread: 

- Gives the host the option of specifying a scheme for the external cache url.
- Fixes a bug where partial urls would be generated if no host or path is provided by now no longer setting a url.

Default behavior matches current behavior where the url isn't generated with a protocol, allowing the browser and/or server to choose.


exchange/auction_test.go
```diff
@@ -298,22 +298,27 @@ type pbsBid struct {
...
-	return "", ""
+type mockCache struct {
+	scheme string
+	host   string
+	path   string
+	items  []prebid_cache_client.Cacheable
+}
+
+func (c *mockCache) GetExtCacheData() (scheme string, host string, path string) {
+	return c.scheme, c.host, c.path
```
SyntaxNode, line 315: This defaults to all empty strings, so functionality equivalent of what exists today.

prebid_cache_client/client_test.go
```diff
@@ -234,7 +234,7 @@ func TestStripCacheHostAndPath(t *testing.T) {
 	}
 	for _, test := range testInput {
 		cacheClient := NewClient(&http.Client{}, &inCacheURL, &test.inExtCacheURL, &metricsConf.DummyMetricsEngine{})
-		cHost, cPath := cacheClient.GetExtCacheData()
+		_, cHost, cPath := cacheClient.GetExtCacheData()
```
hhhjort, line 237: Shouldn't we have some test cases with a non-nil schema? in exchange_test.go, we are just using a mock cache client rather than this one.

SyntaxNode: It's really just acting as a passthrough, but I no argument about adding test coverage. Fixed.

config/config.go
```diff
@@ -136,6 +136,10 @@ func (data *ExternalCache) validate(errs configErrors) configErrors {
 		return errs
 	}
 
+	if data.Scheme != "" && data.Scheme != "http" && data.Scheme != "https" {
+		return append(errs, errors.New("External cache Scheme must be http or https if specified"))
+	}
+
```
guscarreon, line 142: The if statement in line 150 below is probably not needed now that we are including a scheme. Also, should we validate `data.Scheme` in the same manner than `data.Host` and `data.Path`?
```
133   func (data *ExternalCache) validate(errs configErrors) configErrors {
134       if data.Host == "" && data.Path == "" {
135           // Both host and path can be blank. No further validation needed
136           return errs
137       }
138  
139       if data.Scheme != "" && data.Scheme != "http" && data.Scheme != "https" {
140           return append(errs, errors.New("External cache Scheme must be http or https if specified"))
141       }
142  
143       // Either host or path or both not empty, validate.
144       if data.Host == "" && data.Path != "" || data.Host != "" && data.Path == "" {
145           return append(errs, errors.New("External cache Host and Path must both be specified"))
146       }
147       if strings.HasSuffix(data.Host, "/") {
148           return append(errs, errors.New(fmt.Sprintf("External cache Host '%s' must not end with a path separator", data.Host)))
149       }
          // Probably not needed anymore, shoulld be covered by the is statments in line 160 to 166 now,
          // right? I mean, we could leave it because it doesn't hurt to have it there
150       if strings.ContainsAny(data.Host, "://") {
151           return append(errs, errors.New(fmt.Sprintf("External cache Host must not specify a protocol. '%s'", data.Host)))
152       }
153       if !strings.HasPrefix(data.Path, "/") {
154           return append(errs, errors.New(fmt.Sprintf("External cache Path '%s' must begin with a path separator", data.Path)))
155       }
156  
157 -     urlObj, err := url.Parse("https://" + data.Host + data.Path)
    +     urlObj, err := url.Parse(data.Scheme + data.Host + data.Path)
158       if err != nil {
159           return append(errs, errors.New(fmt.Sprintf("External cache Path validation error: %s ", err.Error())))
160       }
    +     if urlObj.Scheme != data.Scheme {
    +         return append(errs, errors.New(fmt.Sprintf("External cache Scheme '%s' is invalid", data.Scheme)))
    +     }
161       if urlObj.Host != data.Host {
162           return append(errs, errors.New(fmt.Sprintf("External cache Host '%s' is invalid", data.Host)))
163       }
164       if urlObj.Path != data.Path {
165           return append(errs, errors.New("External cache Path is invalid"))
166       }
167  
168       return errs
169   }
config/config.go
```

SyntaxNode: > The if statement in line 150 below is probably not needed now that we are including a scheme.

I still think the validation is important. It's ensuring the scheme is not included as part of the host and that's still an invalid case.

> Also, should we validate data.Scheme in the same manner than data.Host and data.Path?

The validation code is testing specifically for 1 of 3 possible values. I don't believe any further validation is necessary, but please let me know if you feel differently and what you may have in mind.

PR #1465, opened 2020-08-27
CCPA Publisher No Sale Relationships

SyntaxNode:


endpoints/openrtb2/amp_auction.go
```diff
@@ -557,7 +554,27 @@ func setAmpExt(site *openrtb.Site, value string) {
 	}
 }
 
-func readConsent(url *url.URL) string {
+func readConsent(url *url.URL) (privacy.PolicyWriter, error) {
```
hhhjort, line 557: Should this be `policyFromUrl()`? I was a bit confused and though `readConsent()` without qualifications should read consent information from the request, as that is the "normal" location for this info. `readConsentFromUrl()` could also be a possibility, but you already used that below.

hhhjort (approved):
It was a lot of work, but looks good now if I didn't miss anything.

privacy/ccpa/policy.go
```diff
@@ -9,139 +9,190 @@ import (
...
+	prebidExtMap, ok := prebidExt.(map[string]interface{})
+	if !ok {
+		return nil, errors.New("request.ext.prebid is not a json object")
 	}
 
-	c = consent[3]
-	if c != 'N' && c != 'Y' && c != '-' {
-		return errors.New("must specify 'N', 'Y', or '-' for the limited service provider agreement")
+	// Remove no sale member
+	delete(prebidExtMap, "nosale")
```
bsardo, line 159: How is this modified `prebidExtMap` being reassigned to `extMap`? I see the test cases pass, particularly the one where `nosale` is removed but other fields in `prebid` remain...

bsardo (approved):

PR #1466, opened 2020-08-28
Rubicon liveramp param

DGarbar:
should translate the "liveramp.com" id from the "user.ext.eids" array to "user.ext.liveramp_idl" as follows:

```
{
  "user": {
    "ext": {
      "eids": [{
        "source": 'liveramp.com',
        "uids": [{
          "id": "T7JiRRvsRAmh88"
        }]
      }]
    }
  }
}
```

 to XAPI:
```
{
  "user": {
    "ext": {
      "liveramp_idl": "T7JiRRvsRAmh88"
    }
  }
}
```

### DIFF
--- a/adapters/rubicon/rubicon.go
+++ b/adapters/rubicon/rubicon.go
@@ -672,10 +672,7 @@ func (a *RubiconAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adap
 
 		if request.User != nil {
 			userCopy := *request.User
-			userExtRP := rubiconUserExt{
-				RP:          rubiconUserExtRP{Target: rubiconExt.Visitor},
-				LiverampIdl: "",
-			}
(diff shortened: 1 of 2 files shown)


adapters/rubicon/rubicon.go
```diff
@@ -247,6 +248,12 @@ type rubiconUserExtEidUidExt struct {
 	RtiPartner string `json:"rtiPartner,omitempty"`
 }
 
+type mappedRubiconUidsParam struct {
+	tpIds       []rubiconExtUserTpID
```
SyntaxNode, line 252: Does tp = third party? If so, please consider naming the `thirdPartyIDs` for clarity.

DGarbar: @SyntaxNode 
This name is already self-completed. It is used in Rubicon Bidder and (for my knowledge) does not mean thirdParty. 
Thanks for proposal

SyntaxNode: Alright.

Issue #1015, opened 2019-08-28
PBS video impression tracking


bretg:
As a followup to the event tracking issue #800, we're ready to implement the injection of additional tracking into VAST XML. Initially this is just to count impressions, but will be expanded to additional video events at some point. There are two scenarios, both of which are required and could be happening at the same time:

1) Server-side VAST - this scenario covers the video XML is returned in the response from a server-side adapter. If that bidder allows PBS to modify their VAST, the server then injects an <impression> tag into the VAST, pointing to a prebid server event string, and then caches the modified results.

Here's a flow diagram showing how tracking will work. Changed items are in red.

<img width="690" alt="Screen Shot 2019-08-27 at 11 34 35 AM" src="https://user-images.githubusercontent.com/4412838/63798418-e2726500-c8be-11e9-8c8e-9f6243d874e4.png">

2) Client-side VAST - this scenario occurs when the video XML is returned in the response from a client-side adapter. PBJS forwards the XML to PBS on a new endpoint that instructs PBS to update the XML and cache it. If that bidder allows PBS to modify their VAST, the server then injects an <impression> tag into the VAST, pointing to a prebid server event string, caches the modified results and responds to the client.

<img width="685" alt="Screen Shot 2019-08-27 at 9 08 02 PM" src="https://user-images.githubusercontent.com/4412838/63825224-cc8c9080-c90e-11e9-8ddf-5e92355291d7.png">

## Proposed modifications

1) Add a new bidder-level flag in PBS that allows a given bidder to turn off our ability to modify their VAST. e.g. if bidderA doesn't want us to modify their VAST, the host company won't have complete analytics.

We propose a new bidder configuration property `modifyingVastXmlAllowed`, with the default being false.

2) When PBS receives VAST from each bidder, it checks the bidder flag and whether the account supports events. If so, it inserts an <impression> tag before storing to PBC

The contents of the <impression> tag are pulled from a new event.url-template property that has macros that need to be resolved. e.g.

```
event:
    url-template: "/event?t=imp&b=%s&f=b&a=%s"
```
where b=BIDID, a=ACCOUNT

3) Algorithm for updating the VAST. (originally this was over-simplified. Changed May 13, 2021)

Scan the 'adm' on video bids:
- If adm contains <InLine> (case insensitive), it's "Inline.
- Otherwise, if adm contains <Wrapper> (case-insensitive), it's a "wrapper".
[...]

PR #1467, opened 2020-09-01
Vtrack and event endpoints

danielguedesb:
Adds /event and /vtrack endpoints

**Note:** _events_enabled_ column is added to accounts_account table:

_SELECT uuid, price_granularity, **events_enabled** FROM accounts_account where uuid = $1 LIMIT 1_

You can add the column to your _postgres_ table by running the following command (adjust accordingly depending on your relational database):

ALTER TABLE account_accounts ADD COLUMN events_enabled BOOLEAN DEFAULT FALSE;

EDIT: Implements https://github.com/prebid/prebid-server/issues/1015


endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+	ANALYTICS_PARAMETER = "x"
+
+	TRACKING_PIXEL_PNG              = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAABHNCSVQICAgIfAhkiAAAAA1JREFUCJljYGBgYAAAAAUAAYehTtQAAAAASUVORK5CYII="
+	TRACKING_PIXEL_PNG_CONTENT_TYPE = "image/png"
+)
+
+type EventRequest struct {
+	Type      Type      `json:"type,omitempty"`
+	Bidid     string    `json:"bidid,omitempty"`
+	AccountId string    `json:"account_id,omitempty"`
```
SyntaxNode, line 50: Style: AccountID per Go naming conventions.

endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+/**
+ * Parses an EventRequest from an Http request
+ */
+func ParseEventRequest(r *http.Request) (*EventRequest, error) {
+	er := &EventRequest{}
+
+	// validate type
+	err := validateType(er, r)
```
SyntaxNode, line 168: It's common in Go to group event handling with the call, such as:

```
err := validateType(er, r)
if err != nil {
  return er, err
}
```

If the func is only returning an error, it's more idiomatic to write:

```
if err := validateType(er, r); err != nil {
  return er, err
}
```

endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+	accountId, err := validateRequiredParameter(r, ACCOUNT_ID_PARAMETER)
+
+	if err != nil {
+		w.WriteHeader(http.StatusUnauthorized)
+		w.Write([]byte(fmt.Sprintf("Account '%s' is required query parameter and can't be empty", ACCOUNT_ID_PARAMETER)))
+		return
+	}
+	eventRequest.AccountId = accountId
+
+	if eventRequest.Analytics == ENABLED {
```
SyntaxNode, line 100: Style: There's a lot of code intended below this if check, please consider a reverted check with a quick return:

```
if eventRequest.Analytics != ENABLED {
  w.WriteHeader(http.StatusUnauthorized) // not sure what this should return
  return
}
```

endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+
+		// get account details
+		account, err := e.DataCache.Accounts().Get(eventRequest.AccountId)
```
SyntaxNode, line 103: Please replace with new account code.

endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+				w.WriteHeader(status)
+				w.Write([]byte(message))
+				return
+			}
+		}
+
+		// account does not support events
+		if !account.EventsEnabled {
+			w.WriteHeader(http.StatusUnauthorized)
+			w.Write([]byte(fmt.Sprintf("Account '%s' doesn't support events", eventRequest.AccountId)))
```
SyntaxNode, line 127: Are these the same responses given by the PBS-Java endpoint? It would be good to match what they already have in place.

endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+func EventRequestToUrl(externalUrl string, request *EventRequest) string {
+	s := fmt.Sprintf(TEMPLATE_URL, externalUrl, request.Type, request.Bidid, request.AccountId)
+
+	return s + optionalParameters(request)
+}
+
+/**
+ * Parses an EventRequest from an Http request
+ */
+func ParseEventRequest(r *http.Request) (*EventRequest, error) {
```
SyntaxNode, line 164: Nice. I like the pattern of this method.

endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+
+	if bidder != "" {
+		er.Bidder = bidder
+	}
+
+	return er, nil
+}
+
+func optionalParameters(request *EventRequest) string {
+	r := ""
```
SyntaxNode, line 214: Please consider using `url.Values` from the base library instead of string concatenation. This will also take care of query string encoding.

endpoints/event.go
```diff
@@ -0,0 +1,367 @@
...
+
+/**
+ * Load tracking pixel from Base64 string
+ */
+func loadTrackingPixel() *TrackingPixel {
+	reader := base64.NewDecoder(base64.StdEncoding, strings.NewReader(TRACKING_PIXEL_PNG))
+	buff := bytes.Buffer{}
+	_, err := buff.ReadFrom(reader)
+	if err != nil {
+		panic(err)
```
SyntaxNode, line 360: A panic seems ok here since it's decoding a constant. An alternative is to encode the byte array directly:

```
[]byte{0x00, 0x00, ..}
```

It's a pain, for sure.. but a one time pain. :)

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+
+const (
+	AccountParameter   = "a"
+	ImpressionCloseTag = "</Impression>"
+	ImpressionOpenTag  = "<Impression>"
+)
+
+type vtrackEndpoint struct {
+	Cfg         *config.Configuration
+	DataCache   cache.Cache
```
SyntaxNode, line 31: `DataCache` would be replaced with newer account code, possibly allowing for `PbsCache` to be renamed `Cache`.

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+type BidCacheRequest struct {
+	Puts []prebid_cache_client.Cacheable `json:"puts"`
+}
+
+type BidCacheResponse struct {
+	Responses []CacheObject `json:"responses"`
+}
+
+type CacheObject struct {
+	Uuid string `json:"uuid"`
```
SyntaxNode, line 45: Go Style Nitpick: UUID

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * Parses a BidCacheRequest from an HTTP Request
+ */
+func (v *vtrackEndpoint) parseVTrackRequest(httpRequest *http.Request) (req *BidCacheRequest, err error) {
+	req = &BidCacheRequest{}
+	err = nil
+
+	// Pull the request body into a buffer, so we have it for later usage.
+	lr := &io.LimitedReader{
```
SyntaxNode, line 174: Nice usage of the `LimitedReader`.

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	requestJson, err := ioutil.ReadAll(lr)
+	if err != nil {
+		return req, err
+	}
+
+	// If the request size was too large, read through the rest of the request body so that the connection can be reused.
+	if lr.N <= 0 {
+		if written, err := io.Copy(ioutil.Discard, httpRequest.Body); written > 0 || err != nil {
```
SyntaxNode, line 186: Is this necessary? Can the body be closed instead?

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+		err = fmt.Errorf("request body is empty")
+		return req, err
+	}
+
+	if err := json.Unmarshal(requestJson, req); err != nil {
+		return req, err
+	}
+
+	// validate request
+	if len(req.Puts) == 0 {
```
SyntaxNode, line 202: This check isn't necessary. The range will handle an empty slice.

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	return req, nil
+}
+
+/**
+ * Cache BidCacheRequest data
+ */
+func (v *vtrackEndpoint) cachePutObjects(req *BidCacheRequest, biddersAllowingVastUpdate []string, accountId string, timeout int64) ([]string, []error) {
+	var nputs []prebid_cache_client.Cacheable
```
SyntaxNode, line 225: Please consider renaming. `nputs` looks like `inputs` mis-spelled.

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+	ctx, cancel := context.WithDeadline(context.Background(), t)
+	defer cancel()
+
+	return v.PbsCache.PutJson(ctx, nputs)
+}
+
+/**
+ * Returns list of bidders that allow VAST XML modification.
+ */
+func biddersAllowingVastUpdate(req *BidCacheRequest, bidderInfos *adapters.BidderInfos, allowUnknownBidder bool) []string {
```
SyntaxNode, line 254: Consider building a `map[string]struct{}` instead to act as a hash set to get checks in O(1) instead of a O(n) search through the slice.

endpoints/vtrack.go
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * Modify BidCacheRequest element Vast XML data
+ */
+func modifyVastXml(externalUrl string, data json.RawMessage, bidid string, bidder string, accountId string, timestamp int64) json.RawMessage {
+	c := string(data)
+	ci := strings.Index(c, ImpressionCloseTag)
+
+	// no impression tag - pass it as it is
+	if ci == -1 {
+		return json.RawMessage(c)
```
SyntaxNode, line 296: Can we return `data` in this case instead?

SyntaxNode (commented):
This is awesome! Thank you for tackling such a large feature implementation! My first round of feedback is largely centered around Go specific styling and use of the legacy (read: outdated) account code.

Just to verify, you're coding against this spec?
https://docs.prebid.org/prebid-server/endpoints/pbs-endpoint-event.html

danielguedesb:
> This is awesome! Thank you for tackling such a large feature implementation! My first round of feedback is largely centered around Go specific styling and use of the legacy (read: outdated) account code.
> 
> Just to verify, you're coding against this spec?
> https://docs.prebid.org/prebid-server/endpoints/pbs-endpoint-event.html

Yes, the implementation was based on PBS's Java version.
Thanks for the feedback!

adapters/info.go
```diff
@@ -201,10 +201,11 @@ const (
...
 type BidderInfo struct {
-	Status       BidderStatus      `yaml:"status" json:"status"`
-	Maintainer   *MaintainerInfo   `yaml:"maintainer" json:"maintainer"`
-	Capabilities *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
-	AliasOf      string            `json:"aliasOf,omitempty"`
+	Status                  BidderStatus      `yaml:"status" json:"status"`
+	Maintainer              *MaintainerInfo   `yaml:"maintainer" json:"maintainer"`
+	Capabilities            *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
+	AliasOf                 string            `json:"aliasOf,omitempty"`
+	ModifyingVastXmlAllowed bool              `yaml:"modifyingVastXmlAllowed" json:"modifyingVastXmlAllowed"`
```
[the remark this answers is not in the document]
danielguedesb: The lack of a separation for the two is why I placed it here however, it also seemed to be a minor thing to be the catalyst for it.
I opted for preventing it from being exposed on the /info endpoint for now.

analytics/core.go
```diff
@@ -65,3 +67,12 @@ type CookieSyncObject struct {
 	Errors       []error
 	BidderStatus []*usersync.CookieSyncBidders
 }
+
+// Loggable object of a Notification Event
+type NotificationEvent struct {
+	Type      string         `json:"type"`
+	Bidid     string         `json:"bidid"`
+	Account   *cache.Account `json:"account"`
```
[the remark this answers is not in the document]
danielguedesb: Awesome.

analytics/core.go
```diff
@@ -65,3 +67,12 @@ type CookieSyncObject struct {
...
+	Bidder    string         `json:"bidder"`
+	Timestamp int64          `json:"timestamp"`
```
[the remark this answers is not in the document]
danielguedesb: I agree, it's likely to be more useful than an int64 for the handlers.

endpoints/event.go
danielguedesb: Yes, it is: https://github.com/rubicon-project/prebid-server-java/blob/master/src/main/java/org/prebid/server/handler/NotificationEventHandler.java#L132

endpoints/vtrack.go
danielguedesb: Nice suggestion.

endpoints/events/accounts.go
```diff
@@ -0,0 +1,41 @@
...
+	"encoding/json"
+	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/stored_requests"
+)
+
+/**
+ * Retrieves an account and merges it with AccountDefaults
+ */
+func GetAccount(ctx context.Context, cfg *config.Configuration, fetcher stored_requests.AccountFetcher, id string) (*config.Account, []error) {
```
SyntaxNode, line 14: Nice. Seems very straight forward.

Do you think this helper would make sense at perhaps `account/account.go` (or read.go, util.go, whatever...) to be shared across other endpoints? 

@laurb9 @bsardo Does this algorithm already live somewhere else? How do you fetch and resolve accounts for the other endpoints?

endpoints/events/accounts_test.go
```diff
@@ -0,0 +1,141 @@
...
+	var acc = &config.Account{
+		ID:            accountID,
+		EventsEnabled: false,
+	}
+
+	if e.EventsEnabled {
+		acc.EventsEnabled = true
+	}
+
+	s, _ := json.Marshal(acc)
```
SyntaxNode, line 34: Please never ignore an error response. I suggest accepting `t *testing.T` as an argument, add a call to `t.Helper()` to tell Go this is a util method, then do a `t.Fatal` if there's a marshal error.

endpoints/events/accounts_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return s, []error{}
+}
+
+func TestShouldReturnDefaultAccountWithSpecifiedIdWhenAccountNotFound(t *testing.T) {
+	// prepare
```
SyntaxNode, line 40: The structure of these tests all look very similar, which is a "code smell" that a [table drive test](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) might make sense here. You can look in the privacy (among others) package for further examples

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,312 @@
...
+		for _, err := range errs {
+			w.Write([]byte(fmt.Sprintf("Internal Error: %s\n", err.Error())))
+		}
+
+		return
+	}
+
+	// insert impression tracking if account allows events and bidder allows VAST modification
+	if account.EventsEnabled && v.Cache != nil {
+		biddersAllowingVastUpdate := biddersAllowingVastUpdate(req, &v.BidderInfos, v.Cfg.VTrack.AllowUnknownBidder)
```
SyntaxNode, line 98: Please consider extracting this clause to a method to improve method readability / reduce func length.

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,312 @@
...
+func biddersAllowingVastUpdate(req *BidCacheRequest, bidderInfos *adapters.BidderInfos, allowUnknownBidder bool) []string {
+	var bl = make(map[string]struct{})
+
+	for _, bcr := range req.Puts {
+		if _, ok := bl[bcr.Bidder]; isAllowVastForBidder(&bcr, bidderInfos, allowUnknownBidder) && !ok {
+			bl[bcr.Bidder] = struct{}{}
+		}
+	}
+
+	l := len(bl)
```
SyntaxNode, line 247: Please consider using a different name for `l`. It looks too much like a `1` IMHO and made line 252 hard to read / understand.

endpoints/events/accounts.go
laurb9: This looks like a simplified version of https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/auction.go#L1313-L1363 

Having two ways of looking up an Account with different behaviors can be confusing. For example, the auction one applies blacklist and disabled status which should arguably be happening everywhere. Instead, I think we can change the signature on the auction GetAccount to support both cases. It already comes with a complete set of unit tests so it should be relatively easy.

endpoints/events/accounts_test.go
laurb9: These tests are not necessary if we use the existing GetAccount, which is covered by  https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/auction_test.go#L1979-L2044

endpoints/events/accounts.go
danielguedesb: I completely agree with the idea of moving it to a package of its own since it's something that it's likely to be needed by different components of the application.

Indeed, it's a simplified version of the method written by @laurb9 - The reasoning behind it is to suggest a simpler/leaner logic to fetch an account and delegate the validations, if any, to the callers. E.g.: PBS-Java doesn't apply the validations done on the auction endpoint on the /vtrack and /event endpoints.

I wanted to open a discussion about it and hear your thoughts - As @laurb9 pointed out, those validations could (should?) arguably be done everywhere.

endpoints/events/accounts_test.go
danielguedesb: These will likely be removed due to #1483

endpoints/events/event.go
```diff
@@ -0,0 +1,363 @@
...
+type EventType string
+
+// Possible values of  vents Prebid Server can receive for an ad.
+const (
+	Win EventType = "win"
+	Imp EventType = "imp"
+)
+
+// EventFormat enumerates the values of a Prebid Server event.
+type Format string
```
laurb9, line 26: Should this be `ResponseFormat` ? it specifies what kind of response is expected.

endpoints/events/event.go
```diff
@@ -0,0 +1,363 @@
...
+
+/**
+ * Parses an EventRequest from an Http request
+ */
+func ParseEventRequest(r *http.Request) (*EventRequest, error) {
+	er := &EventRequest{}
+
+	// validate type
+	if err := validateType(er, r); err != nil {
+		return er, err
```
laurb9, line 181: this reads a little confusing, I suggest using a different name that `er`, maybe even `event`

endpoints/events/event.go
```diff
@@ -0,0 +1,363 @@
...
+func EventRequestToUrl(externalUrl string, request *EventRequest) string {
+	s := fmt.Sprintf(TemplateUrl, externalUrl, request.Type, request.Bidid, request.AccountID)
+
+	return s + optionalParameters(request)
+}
+
+/**
+ * Parses an EventRequest from an Http request
+ */
+func ParseEventRequest(r *http.Request) (*EventRequest, error) {
```
laurb9, line 176: This function returns only the first error. In other places PBS tries to return everything that went wrong.
Do we want to do that here as well ?

endpoints/events/event.go
```diff
@@ -0,0 +1,363 @@
...
+	switch request.Format {
+	case Blank:
+		r.Add(FormatParameter, string(Blank))
+	case Image:
+		r.Add(FormatParameter, string(Image))
+	}
+
+	//analytics
+	switch request.Analytics {
+	case Enabled:
```
laurb9, line 239: Should this be the default ? My understand of this flag is that we can send x=0 to disable analytics, which seems to imply that a missing parameter means that it's enabled.

endpoints/events/event.go
```diff
@@ -0,0 +1,363 @@
...
+
+	switch t {
+	case string(Imp):
+		er.Type = Imp
+		return nil
+	case string(Win):
+		er.Type = Win
+		return nil
+	default:
+		return fmt.Errorf("unknown type: '%s'", t)
```
laurb9, line 272: What do you think of using `errortypes.BadInput` as a more specific error ?

endpoints/events/accounts.go
SyntaxNode: > PBS-Java doesn't apply the validations done on the auction endpoint on the /vtrack and /event endpoints.

Do you know if they did this for a specific reason?

> As @laurb9 pointed out, those validations could (should?) arguably be done everywhere.

I agree. I don't see why we wouldn't always want to validate the account in the same way. There might be a good reason for a specific use case. I just can't think of any.

endpoints/events/event.go
danielguedesb: I agree with your reasoning but I followed PBS Java's naming convention:
https://github.com/rubicon-project/prebid-server-java/blob/master/src/main/java/org/prebid/server/events/EventRequest.java#L32

I'll leave that up to you guys.

endpoints/events/event.go
danielguedesb: That came to mind as well since I saw that pattern in other places but, that would imply an intent on doing something with all of the errors - From a user's perspective, being aware of all the syntax errors with a single call would (probably) prevent future calls using the same payload.

However, PBS Java does a sequential validation of the request returning on the first error that it encounters:
https://github.com/rubicon-project/prebid-server-java/blob/master/src/main/java/org/prebid/server/handler/NotificationEventHandler.java#L69

As such, I opted for following that same approach and keep their responses similar.
If that is unnecessary I'm happy to change it.

endpoints/events/accounts.go
danielguedesb: Not sure, nothing comes to mind besides a possible requirement that allows blacklisted, disabled or unknown accounts to speak with that endpoint - If that's the case, I'm also curious about use cases behind it.
As it is, those type of accounts will be able to trigger events as long as they have _enabled_events = true_ or the running instance of PBS allows unknown accounts.

It could be linked to a possible safety mechanism that disabled events for a given account once it was blacklisted, but that doesn't seem to be the case; at the moment, black listed accounts are loaded through local configuration: https://github.com/rubicon-project/prebid-server-java/blob/master/src/main/java/org/prebid/server/spring/config/ServiceConfiguration.java#L201

Perhaps there are plans to change that in the future and move it to an attribute of the account itself? Or support remote configs to override local configs and allow an external component manage that?

Though, instead of assuming it's probably best to reach out and ask them directly. Does anyone comes to mind?

endpoints/events/event.go
SyntaxNode: I'm good with `ResponseFormat` as it's a bit more descriptive. This name doesn't affect the API. We should certainly look at PBS-Java's implementation for guidance on this feature, but I don't see a need to keep the internal names verbatim.

endpoints/events/accounts.go
SyntaxNode: > Perhaps there are plans to change that in the future and move it to an attribute of the account itself? Or support remote configs to override local configs and allow an external component manage that?

I would expect those to be abstracted away by the accounts code. This may simply be an oversight.

> Though, instead of assuming it's probably best to reach out and ask them directly. Does anyone comes to mind?

I recommend to use the #prebid-server-java channel in Prebid.org's Slack.

endpoints/events/accounts_test.go
SyntaxNode: Sounds good. We're reviewing that today.

endpoints/events/event.go
```diff
@@ -0,0 +1,364 @@
...
+		return
+	}
+
+	// handle notification event
+	e.Analytics.LogNotificationEventObject(&analytics.NotificationEvent{
+		Type:      string(eventRequest.Type),
+		Timestamp: time.Unix(eventRequest.Timestamp, 0),
+		Bidid:     eventRequest.Bidid,
+		Bidder:    eventRequest.Bidder,
+		Account:   account,
```
laurb9, line 153: Passing the entire `EventRequest` object instead of select fields would be more consistent with the other analytics methods.

danielguedesb: The implementation follows the pattern used on both PBS Java and PBS Go: 

1. PBS Java
1.1. https://github.com/rubicon-project/prebid-server-java/blob/master/src/main/java/org/prebid/server/handler/NotificationEventHandler.java#L112

2. PBS Go
2.1. https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/amp_auction.go#L90
2.2. https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/amp_auction.go#L132

While both objects are similar at the moment, they provide a separation of concerns that gives us the ability to evolve each of them separately thus, keeping each of the components abstracted from specifics of each layer:

-  _EventRequest_ is the object within the endpoint's scope and from where we extract information to build our NotificationEvent (we could enrich our NotificationEvent with data from other sources besides _EventRequest_).

-  _NotificationEvent_ is the object from the Analytics domain and the one that should to be passed to the event handlers.

Happy to hear your thoughts on this.

SyntaxNode:
We merged in https://github.com/prebid/prebid-server/pull/1483 today, so you may take advantage of that code here.

endpoints/events/event.go
```diff
@@ -173,6 +169,24 @@ func ParseEventRequest(r *http.Request) (*analytics.EventRequest, error) {
...
+	for _, er := range errs {
+		errCode := errortypes.ReadCode(er)
+
+		switch errCode {
+		case errortypes.BlacklistedAppErrorCode, errortypes.BlacklistedAcctErrorCode:
+			return http.StatusServiceUnavailable, er.Error()
+		case errortypes.AcctRequiredErrorCode:
+			return http.StatusBadRequest, er.Error()
+		default:
+			return http.StatusInternalServerError, er.Error()
```
laurb9, line 183: With a default case, the loop only runs once.

We probably want behavior like `writeError` in https://github.com/prebid/prebid-server/blob/7b59a4bd49e502b05f81624bcc908259aec757cf/endpoints/openrtb2/auction.go#L1282 or the similar but different code the other two auction endpoints.

danielguedesb: You're right, an overview on my part - Refactored.

analytics/event.go
```diff
@@ -0,0 +1,38 @@
...
+const (
+	Win EventType = "win"
+	Imp EventType = "imp"
+)
+
+// EventFormat enumerates the values of a Prebid Server event.
+type ResponseFormat string
+
+const (
+	// BlankEventFormat describes an event which returns an HTTP 200 with an empty body.
```
SyntaxNode, line 16: Nitpick: Update comment for `BlankEventFormat` and `ImageEventFormat` to match current names.

analytics/event.go
```diff
@@ -0,0 +1,38 @@
...
+	Blank ResponseFormat = "b"
+	// ImageEventFormat describes an event which returns an HTTP 200 with a PNG body.
+	Image ResponseFormat = "i"
+)
+
+// Indicates if the notification event should be handled or not
+type Analytics string
```
SyntaxNode, line 23: Nitpick: Please rename start of the comment to match the new type name `Analytics indicates if ...`.

analytics/filesystem/file_module.go
```diff
@@ -72,6 +73,18 @@ func (f *FileLogger) LogAmpObject(ao *analytics.AmpObject) {
 	f.Logger.Flush()
 }
 
+//Logs NotificationEvent to file
+func (f *FileLogger) LogNotificationEventObject(ne *analytics.NotificationEvent) {
```
SyntaxNode, line 77: Please add unit tests for this method.

analytics/filesystem/file_module.go
```diff
@@ -176,3 +189,20 @@ func jsonifyAmpObject(ao *analytics.AmpObject) string {
 		return fmt.Sprintf("Transactional Logs Error: Amp object badly formed %v", err)
 	}
 }
+
+func jsonifyNotificationEventObject(ne *analytics.NotificationEvent) string {
```
SyntaxNode, line 193: Please add unit tests for this method.

endpoints/events/event.go
```diff
@@ -0,0 +1,324 @@
...
+	if eventRequest.Analytics != analytics.Enabled {
+		return
+	}
+
+	ctx := context.Background()
+
+	// get account details
+	account, errs := accountService.GetAccount(ctx, e.Cfg, e.Accounts, eventRequest.AccountID)
+	if len(errs) > 0 {
+		status, messages := HandleAccountServiceErrors(errs)
```
SyntaxNode, line 99: Please add unit test coverage for this new error handler condition.

endpoints/events/event.go
```diff
@@ -0,0 +1,324 @@
...
+		errs = append(errs, err)
+	}
+
+	// validate analytics (optional)
+	if err := validateAnalytics(event, r); err != nil {
+		errs = append(errs, err)
+	}
+
+	// Bidder
+	bidder := r.FormValue(BidderParameter)
```
SyntaxNode, line 169: Please use `url.Query().Get(...)` to pull values from the query string. This method will also check the request body and is more prone to introducing strange behavior / bugs.

There are a few other usages of `FormValue` in this PR.

endpoints/events/event.go
```diff
@@ -0,0 +1,324 @@
...
+
+	if bidder != "" {
```
SyntaxNode, line 171: Could this be simplified to `event.Bidder = r.URL.Query().Get(BidderParameter)`? Even if this returns an empty string, I don't see the harm of making the assignment since it defaults to an empty string anyway. Not sure the complexity of the extra `if` statement is worth it.

endpoints/events/event.go
```diff
@@ -0,0 +1,324 @@
...
+
+	if opt != "" {
+		return "&" + opt
+	}
+
+	return opt
+}
+
+// validate type
+func validateType(er *analytics.EventRequest, httpRequest *http.Request) error {
```
SyntaxNode, line 238: Nitpick: The validateXXX methods all have a side affect of updating the `EventRequest`. To me, that seems like it's primary role and it will return an error only if that fails. Consider renaming to `setType`, `setFormat`, etc.. so the setting side affect is obvious - or perhaps something like `readType` which adds the connotation of reading from the request.

This is admittedly nitpick territory. Please take this as a suggestion, not a requirement.

endpoints/events/event_test.go
```diff
@@ -0,0 +1,591 @@
...
+	"github.com/prebid/prebid-server/stored_requests"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// Mock Analytics Module
+type eventsMockAnalyticsModule struct {
```
SyntaxNode, line 18: This is fine. For the future, a preferred approach would be to use the `github.com/stretchr/testify/mock` package. End result is the same.

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,306 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte(fmt.Sprintf("Invalid request: %s\n", err.Error())))
+		return
+	}
+
+	ctx := context.Background()
+
+	// get account details
+	account, errs := accountService.GetAccount(ctx, v.Cfg, v.Accounts, accountId)
+	if len(errs) > 0 {
```
SyntaxNode, line 87: Please add unit test coverage for the errs case.

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,306 @@
...
+	req, err := ParseVTrackRequest(r, v.Cfg.MaxRequestSize+1)
+
+	// check if there was any error while parsing puts request
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte(fmt.Sprintf("Invalid request: %s\n", err.Error())))
+		return
+	}
+
+	ctx := context.Background()
```
SyntaxNode, line 83: We should probably add a configurable timeout of some kind for account lookup. This can be a follow-up PR in our backlog or for you to tackle if you'd like. Please advise.

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,306 @@
...
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(fmt.Sprintf("Error serializing pbs cache response: %s\n", err.Error())))
+
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Write(d)
+	}
+}
```
SyntaxNode, line 122: What response code is sent back to the client if the account doesn't have events enabled? I don't know what the server defaults are. Maybe add a comment?

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,306 @@
...
+func modifyVastXml(externalUrl string, data json.RawMessage, bidid string, bidder string, accountId string, timestamp int64) json.RawMessage {
+	c := string(data)
+	ci := strings.Index(c, ImpressionCloseTag)
+
+	// no impression tag - pass it as it is
+	if ci == -1 {
+		return data
+	}
+
+	vastUrlTracking := GetVastUrlTracking(externalUrl, bidid, bidder, accountId, timestamp)
```
SyntaxNode, line 291: Please add test coverage from line 291 and down. The current tests always return from the `ci == -1` check.

SyntaxNode (commented):
Looks great. Sorry for the delay in reviewing.

The flow looks great with use of the new account service. My comments are largely covering comment nitpicks and test coverage expansion requests.

endpoints/events/vtrack.go
danielguedesb: I agree, it's good practice - Refactored. 
It uses the TimeoutMS configuration variable to apply an execution time constraint to the endpoint's handler.

endpoints/events/vtrack.go
danielguedesb: Good catch, refactored. The request should still be handled even if an account doesn't have events enabled.

endpoints/events/event.go
```diff
@@ -0,0 +1,322 @@
...
+	"time"
+
+	accountService "github.com/prebid/prebid-server/account"
+	"net/http"
+	"net/url"
+	"strconv"
+)
+
+const (
+	DefaultTimeoutMS = 1000
```
SyntaxNode, line 20: Thank you. We may expand this to a config option in the future. I think this is a good starting point. We generously expect cache writes to take 50ms and account lookups should very likely be quick. I think this also matches well with our default auction timeout.

endpoints/events/event.go
```diff
@@ -0,0 +1,322 @@
...
+
+	if err != nil {
+		w.WriteHeader(http.StatusUnauthorized)
+		w.Write([]byte(fmt.Sprintf("Account '%s' is required query parameter and can't be empty", AccountIdParameter)))
+		return
+	}
+	eventRequest.AccountID = accountId
+
+	if eventRequest.Analytics != analytics.Enabled {
+		return
```
SyntaxNode, line 94: I missed this one. Please write a response with a status code before exiting this handler. Not sure if we need a body here.

endpoints/events/vtrack.go
SyntaxNode: There is still no coverage. I think the issue is with the test setup. Please look for my comment there.

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,318 @@
...
+
+// getAccountId extracts an account id from an HTTP Request
+func getAccountId(httpRequest *http.Request) string {
+	return httpRequest.URL.Query().Get(AccountParameter)
+}
+
+// modifyVastXml modifies BidCacheRequest element Vast XML data
+func modifyVastXml(externalUrl string, data json.RawMessage, bidid string, bidder string, accountId string, timestamp int64) json.RawMessage {
+	c := string(data)
+	ci := strings.Index(c, ImpressionCloseTag)
```
SyntaxNode, line 293: Does it matter that this is case sensitive? There's no easy way to add a case insensitive compare. I'm not familiar enough with VAST to know if that's going to be an issue.

endpoints/events/vtrack_test.go
```diff
@@ -0,0 +1,581 @@
...
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+
+	e = enc.Encode(req)
+
+	return buf.String(), e
+}
+
+func getVTrackRequestData(wi bool) (data []byte, e error) {
```
SyntaxNode, line 569: This test method never returns the buffer. The output `data` variable is never assigned. This is the root cause of the test coverage issues with the `modifyVastXml` func. I'm surprised, and a bit concerned, this didn't cause a test to fail - and I'm sure you are too.

endpoints/events/event.go
danielguedesb: I agree, I opted for expanding it now rather than later.

endpoints/events/event.go
danielguedesb: An oversight on my parte - Refactored. Returning 204 when there is no content to return.

endpoints/events/vtrack.go
danielguedesb: Good question, I'm not too familiar with it either. 
However, looking at how PBS Java handles this and a few samples provided by IAM, it would appear to be case sensitive: 
PBS Java - https://github.com/rubicon-project/prebid-server-java/blob/master/src/main/java/org/prebid/server/cache/CacheService.java#L520

IAM Samples - https://github.com/InteractiveAdvertisingBureau/VAST_Samples

endpoints/events/vtrack_test.go
danielguedesb: Oh my, you're right.. It was not assigning a value on the early return - Refactored.

The test doesn't fail because having no data on the cacheable object that we want to cache is a valid use case.
In this case, even if the bidder is allowed to modify vast xml, since there's no data to modify we just skip that step and and the end result is the same.

prebid_cache_client/client.go
```diff
@@ -181,6 +185,26 @@ func encodeValueToBuffer(value Cacheable, leadingComma bool, buffer *bytes.Buffe
 		buffer.WriteString(string(value.Key))
 		buffer.WriteString(`"`)
 	}
+
+	//vtrack specific
+	if len(value.BidID) > 0 {
```
SyntaxNode, line 190: Please add test coverage for the cache client's new `BidID`,` Bidder`, and `Timestamp` members.

endpoints/events/event.go
```diff
@@ -0,0 +1,327 @@
...
+	eventRequest.AccountID = accountId
+
+	if eventRequest.Analytics != analytics.Enabled {
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
+
+	ctx := context.Background()
+	if e.Cfg.Event.TimeoutMS > 0 {
+		var cancel context.CancelFunc
```
SyntaxNode, line 97: Please add test coverage for when TimeoutMS  > 0.

endpoints/events/event.go
```diff
@@ -0,0 +1,327 @@
...
+	// format
+	switch request.Format {
+	case analytics.Blank:
+		r.Add(FormatParameter, string(analytics.Blank))
+	case analytics.Image:
+		r.Add(FormatParameter, string(analytics.Image))
+	}
+
+	//analytics
+	switch request.Analytics {
```
SyntaxNode, line 224: Please add test coverage for `request.Analytics`.

endpoints/events/vtrack.go
SyntaxNode: There is good test coverage in events.go. Can you please add similar tests for the vtrack endpoint?

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// no impression tag - pass it as it is
+	if ci == -1 {
+		return data
```
SyntaxNode, line 297: Can you please add a test case with no impression tag?

SyntaxNode (commented):
Looks great. Thank you for your hard work and patience with the review. This is a major feature and I'm very excited to get this in soon, targeting for this week's release. I flagged just a few more areas to add test coverage.

bsardo:
This is looking good. Nothing new to add to @SyntaxNode's test coverage comments.

endpoints/events/event_test.go
```diff
@@ -663,3 +696,44 @@ func TestShouldParseEventCorrectly(t *testing.T) {
 	assert.Equal(t, 0, len(errs))
 	assert.EqualValues(t, expected, er)
 }
+
+func TestEventRequestToUrl(t *testing.T) {
```
bsardo, line 700: This looks good and you've got the code coverage. My personal preference would be to add the following scenarios by either altering the existing cases or adding new ones for completeness but I'm nitpicking 🙂:
- `request.Timestamp <= 0`
- `request.Bidder == ""`
- `opt == ""`

Up to you if you want to add.

endpoints/events/event.go
```diff
@@ -185,15 +186,28 @@ func HandleAccountServiceErrors(errs []error) (status int, messages []string) {
...
 
+	//grab error messages
 	for _, er := range errs {
-		errCode := errortypes.ReadCode(er)
+		messages = append(messages, er.Error())
+	}
 
+	//set appropriate error code
+	for _, er := range errs {
+		if errors.Is(er, context.DeadlineExceeded) {
```
SyntaxNode, line 196: This is good with me. It's new functionality in Go 1.13, but that's our minimum supported version.

endpoints/events/event.go
```diff
@@ -185,15 +186,28 @@ func HandleAccountServiceErrors(errs []error) (status int, messages []string) {
...
 		case errortypes.BlacklistedAppErrorCode, errortypes.BlacklistedAcctErrorCode:
 			status = http.StatusServiceUnavailable
+			break
+		case errortypes.TimeoutErrorCode:
+			status = http.StatusGatewayTimeout
+			break
 		}
-
-		messages = append(messages, er.Error())
 	}
```
SyntaxNode, line 211: I'm getting a bit of a "code smell" from the separate loops over the same data twice. I'm guessing the second loop allows you to break out early. I'm not worried about the performance. @bsardo what's your opinion here?

endpoints/events/event_test.go
```diff
@@ -424,6 +435,26 @@ func TestHandleAccountServiceErrors(t *testing.T) {
 				response: "Invalid request: Prebid-server has disabled Account ID: testacc, please reach out to the prebid server host.\n",
 			},
 		},
+		"timeout": {
+			fetcher: &mockAccountsFetcher{
+				Fail:       false,
+				DurationMS: 50,
```
SyntaxNode, line 441: Does this imply the test runs for 50ms? Is there way to perhaps complete the context immediately to simulate the timeout? Just trying to keep "sleeps" during tests minimal.

endpoints/events/vtrack.go
SyntaxNode: I don't think this has test coverage.
Screen Shot 2020-10-01 at 3 44 15 PM

endpoints/events/event_test.go
danielguedesb: Happy to do so.

endpoints/events/event_test.go
danielguedesb: Not quite, the mocked _FetchAccount_ relies on a select with two communications and blocks until one of them completes.
DurationMS indicates how long one of the two operations will take to complete, the other is controlled by the _context_ received as an argument.
The _context_ is configured with a timeout set in the configuration, which, in this case is set to **1ms**.
In essence, the _FetchAccount_ will unblock after 1ms due to our context timing out and complete soon after.

endpoints/events/event.go
bsardo: I'm ok with the two loops assuming you're trying to break out early. However, I don't think the breaks are doing anything. If you're trying to break out of the for loop early, I think you need to label the for loop and then explicitly state that you're trying to break out of the for like this: `break myForLoopLabel`.

SyntaxNode: Good point. That's breaking out of the switch case, which is implied anyway. Please don't use labels.

bsardo: 🙂In that case, it's probably best to just iterate over the errors once without stopping early for a minor performance gain. You'd be setting the status to the last error instead of the first error but I imagine it's arbitrary either way.

danielguedesb: You're right about the `break`s, force of habit.. Refactored for the sake of simplicity.

SyntaxNode (approved):

laurb9:
I have run a few requests against the event endpoint in my test server. It looks good, with a few comments:

2) Without passing `x=1` explicitly, the event is not logged.
The default should be to log the event, the examples in the doc only pass `x` when they need
to skip logging, otherwise they leave it out.

> the event url in this scenario carries an extra parameter "x=0" which tells PBS to not log this event to analytics.

```
curl -si 'http://localhost:8000/event?t=imp&b=123&a=test'
HTTP/1.1 204 No Content
```

2) With an invalid account, the server returns 500. I think it should be 400.
```
curl -si 'http://localhost:8000/event?t=imp&b=123&a=invalid_acct&x=1'
HTTP/1.1 500 Internal Server Error

Invalid request: Prebid-server could not verify the Account ID. Please reach out to the prebid server host.
```

3) The logged event has `Request` capitalized but `account` in lowercase. It should be consistent with the
other jsonified events.

```json
{"type":"/event","Request":{"type":"imp","analytics":"1","bidid":"123","account_id":"valid_acct"},"account":{"id":"VALID_ACCT","disabled":false,"cache_ttl":{"Banner":600,"Video":600,"Native":600,"Audio":3600},"events_enabled":true}}
```

4) The /event endpoint allows any (or no) bidder (see above). I think it should be required.

Will do the /vtrack after lunch and post my results here as well.

danielguedesb:
> 1. Without passing `x=1` explicitly, the event is not logged.
>    The default should be to log the event, the examples in the doc only pass `x` when they need
>    to skip logging, otherwise they leave it out.
> 

You're right - Taking a closer look at PBS [Java](https://github.com/rubicon-project/prebid-server-java/blob/0f4a78d9b7b45fb65f25ffdb747334f21203a730/src/main/java/org/prebid/server/events/EventUtil.java#L124), they default to 'x=1' unless you specify otherwise.

> 2. With an invalid account, the server returns 500. I think it should be 400.

It seems to be the returned status code in other places as well - refactored.

> 3. The logged event has `Request` capitalized but `account` in lowercase. It should be consistent with the

I agree, refactored.

> 
> 4. The /event endpoint allows any (or no) bidder (see above). I think it should be required.
> 

Java's version doesn't enforce it, any particular reason why it should/shouldn't be required?

laurb9:
I used a test request as described in #1015 and it appears /vtrack endpoint does not decode the `value` field, which causes it to fail writing to cache.
```
curl 'http://localhost:8000/vtrack?a=valid_acct' --header 'content-type: application/json' --data @- <<EOF 
{
  "puts": [
    {
      "bidid": "test-bid-id-1",
      "bidder": "openx",
      "timestamp": 123456789,
      "type": "xml",
      "value": "<?xml version=\"1.0\" encoding=\"UTF-8\"?><VAST version=\"2.0\"><Ad id=\"123456\"><Wrapper><VASTAdTagURI><![CDATA[http://localhost/]]></VASTAdTagURI><Impression><![CDATA[http://sampledomain.com/event/impression]]></Impression><Creatives></Creatives></Wrapper></Ad></VAST>",
      "ttlseconds": 3600
    }
  ]
}
EOF
```
response
```
Error(s) updating vast: Prebid Cache call to http://cache:2424/cache returned 400: Request body {"puts":[{"type":"xml","ttlseconds":3600,"value":}]} is not valid JSON.
```

I have modified the request to replace `value` with `Data` and the request succeeds but the json response is returned with `content-type: text/plain`

prebid_cache_client/client.go
```diff
@@ -45,6 +45,10 @@ type Cacheable struct {
 	Data       json.RawMessage
```
laurb9, line 45:
```suggestion
	Data       json.RawMessage    `json:"value"`
```

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+
+		cacheables = append(cacheables, *nc)
+	}
+
+	return v.Cache.PutJson(ctx, cacheables)
+}
+
+// getBiddersAllowingVastUpdate returns a list of bidders that allow VAST XML modification
+func getBiddersAllowingVastUpdate(req *BidCacheRequest, bidderInfos *adapters.BidderInfos, allowUnknownBidder bool) []string {
```
laurb9, line 251: What do you think of returning the map `bl` directly instead of a slice of keys ?
The result is used to check if a particular key is in there so a map works just as well.

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,318 @@
...
+	var bl = make(map[string]struct{})
```
laurb9, line 252:
```suggestion
	var bl = make(map[string]struct{}, len(req.Puts))
```

endpoints/events/vtrack.go
```diff
@@ -0,0 +1,318 @@
...
+	keys := make([]string, 0, length)
+	for key := range bl {
+		keys = append(keys, key)
+	}
+
+	return keys
+}
+
+// isAllowVastForBidder checks if a bidder is active and allowed to modify vast xml data
+func isAllowVastForBidder(r *prebid_cache_client.Cacheable, bidderInfos *adapters.BidderInfos, allowUnknownBidder bool) bool {
```
laurb9, line 274: nitpick: if the only field from `r` that this function uses is `r.Bidder`, it could receive that as a parameter instead of the entire structure.

prebid_cache_client/client.go
danielguedesb: I'm concerned about this change due to backwards compatibility. Thoughts?

endpoints/events/vtrack.go
danielguedesb: I like it - Refactored.

prebid_cache_client/client.go
laurb9: I think it's fine. The cache client builds the json piece by piece into a buffer so it does not need those annotations, and it does write `Data` into `"value"`. The json representation always uses `"value"`.

Tests that cover the actual inputs and outputs, like the exchange tests, are useful to detect these [de]serialization errors that passed the more focused unit tests.

bsardo (approved):

SyntaxNode:
@laurb9 Thank you for the detailed review. @danielguedesb Thank you being very patient with the review process. This is looking good to us. Do you both agree we're ready to merge in?

laurb9:
> @laurb9 Thank you for the detailed review. @danielguedesb Thank you being very patient with the review process. This is looking good to us. Do you both agree we're ready to merge in?

Looks good to me as well.

There work left to be done for event feature parity with java but that is out of scope with the description for this PR, so they can be done separately - I think wurl insertion in bids and VAST rewriting when pbs caches is remaining.

SyntaxNode (approved):

SyntaxNode:
This is largely complete in PBS-Go. @laurb9 noted two remaining things to implement:
- wurl insertion in bids 
- VAST rewriting when pbs caches

@danielguedesb Are you able to continue contributing to add those areas of functionality?

PR #1468, opened 2020-09-01
Fix Test TestEventChannel_OutputFormat

guscarreon:
`TestEventChannel_OutputFormat` found in `analytics/pubstack/eventchannel/eventchannel_test.go` is sometimes failing, specially when run by Travis, very likely because of the concurrent functionality of the `EventChannel`methods. This PR corrects said test case and adds assertions.

### DIFF
--- a/analytics/pubstack/eventchannel/eventchannel_test.go
+++ b/analytics/pubstack/eventchannel/eventchannel_test.go
@@ -161,48 +161,21 @@ func TestEventChannel_OutputFormat(t *testing.T) {
 	eventChannel := NewEventChannel(send, 15000, 10, 2*time.Minute)
 
 	eventChannel.Push([]byte("one"))
-	time.Sleep(5 * time.Millisecond)
+	time.Sleep(1 * time.Millisecond)
+
 	eventChannel.flush()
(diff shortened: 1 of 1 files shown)


analytics/pubstack/eventchannel/eventchannel_test.go
```diff
@@ -171,5 +175,34 @@ func TestEventChannel_OutputFormat(t *testing.T) {
...
+	data := make([]byte, 0)
+	send := newSender(&data)
+
+	eventChannel := NewEventChannel(send, maxByteSize, maxEventCount, maxTime)
+
+	eventChannel.Push([]byte("one"))
+	eventChannel.Push([]byte("two"))
+	eventChannel.flush()
+
+	time.Sleep(10 * time.Millisecond)
```
SyntaxNode, line 190: Consider moving the sleep before the `flush` (and `Close` call in `TestEventChannelPushAndClose`) to ensure the messages are pushed before calling `flush`.

SyntaxNode, line 190: Could we change this to 1 ms? If we need to sleep due wait for go routines, IMHO it should be a minimal amount of time.

PR #1469, opened 2020-09-02
Invibes adapter

invibes:
Add new prebid server adapter


SyntaxNode:
Thank you for your contribution. We'll review this PR early next week.

For now, could you please open a PR in https://github.com/prebid/prebid.github.io to add a doc for this new adapter. More information is found in our new adapter guide: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#document-your-adapter

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+	Kw            string
+	IsAmp         bool
+	Width         string
+	Height        string
+	GdprConsent   string
+	Gdpr          bool
+	Bvid          string
+	InvibBVLog    bool
+	VideoAdDebug  bool
+}
```
SyntaxNode, line 37: Nitpick: Per Go naming conventions, acronyms should be spelled out in uppercase. That would be:
IsAmp -> IsAMP
GdprConsent -> GDPRConsent
Gdpr -> GDPR
PlacementIds -> PlacementIDs
ImpId -> ImpID

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+	urlTemplate, err := template.New("endpointTemplate").Parse(endpointTemplate)
+	if err != nil {
+		glog.Fatal("Unable to parse endpoint url template")
+		return nil
+	}
+	return &InvibesAdapter{EndpointTemplate: *urlTemplate}
+}
+
+func (a *InvibesAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	if len(request.Imp) == 0 {
```
SyntaxNode, line 86: I know other bidders have this check. I don't believe it's necessary though. The core code will not call `MakeRequests` if there are no impressions. Given this information, I'll trust whatever choice you make here.

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+		return nil, []error{&errortypes.BadInput{
+			Message: "No imps in the bid request",
+		}}
+	}
+
+	consentString := ""
```
SyntaxNode, line 92: This method is really large. To improve readability, consider refactoring  getting the consent string and gdpr flag into a method. The call site can be reduced to something like:

```
gdprApplies, consentString := readGDPR(request)
```

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+	if request.User != nil {
+		var extUser openrtb_ext.ExtUser
+		if err := json.Unmarshal(request.User.Ext, &extUser); err == nil {
+			consentString = extUser.Consent
+		}
+	}
+	gdprApplies := "1"
```
SyntaxNode, line 99: Does Invibes only serve EU traffic? Defaulting gdprApplies to "1" and the consentString to "" will invalidate a lot of non-EU requests.

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	for _, imp := range request.Imp {
+		var bidderExt adapters.ExtImpBidder
+		if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
+			errors = append(errors, &errortypes.BadInput{
+				Message: "Error parsing bidderExt object",
+			})
+		}
+		if len(errors) == 0 {
```
SyntaxNode, line 126: I'm not sure this is doing what you intend. If there is an error with imp-1, then this statement would be false for every imp. I suspect you want to only continue parsing, if so consider adding a `continue` in the Unmarshal error handling branch above.

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+				errors = append(errors, &errortypes.BadInput{
+					Message: "Error parsing invibesExt parameters",
+				})
+			}
+			if imp.Banner == nil {
+				errors = append(errors, &errortypes.BadInput{
+					Message: "Banner not specified",
+				})
+			}
+			if len(errors) == 0 {
```
SyntaxNode, line 138: Same problem with the error check here. Consider adding a `continue` statement to the error branches for Unmarshal and Banner == nil above.

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+				currentBanner := *imp.Banner
```
SyntaxNode, line 139: Please consider extracting the adFormats code to a separate method.

adapters/invibes/invibes.go
```diff
@@ -0,0 +1,327 @@
...
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+
+	if request.Device != nil {
+		headers.Add("User-Agent", request.Device.UA)
+	}
+
+	if invibesParams.TestIp != "" {
```
SyntaxNode, line 212: Is the test ip functionality necessary in the production adapter?

adapters/invibes/params_test.go
```diff
@@ -0,0 +1,52 @@
...
+var validParams = []string{
+	`{"placementId":"123", "host": "bid.videostep.com"}`,
+	`{"placementId":"123", "host": "bid2.videostep.com", "debug":{}}`,
+	`{"placementId":"123", "host": "bid.videostep.com", "debug":{"testIp":"0.0.0.0"}}`,
+	`{"placementId":"123", "host": "bid.videostep.com", "debug":{"testBvid":"1234"}}`,
+}
+
+var invalidParams = []string{
+	``,
+	`{"placementId":"123"}`,
```
SyntaxNode, line 45: Please add a test for invalid formats, like:
```
`{"placementId":123}`,
`{"placementId":"123", "debug":"malformed"}`,
```

config/config.go
```diff
@@ -888,6 +889,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.gumgum.endpoint", "https://g2.gumgum.com/providers/prbds2s/bid")
 	v.SetDefault("adapters.improvedigital.endpoint", "http://ad.360yield.com/pbs")
 	v.SetDefault("adapters.ix.endpoint", "http://appnexus-us-east.lb.indexww.com/transbidder?p=184932")
+	v.SetDefault("adapters.invibes.endpoint", "https://{{.Host}}/bid/ServerBidAdContent")
```
SyntaxNode, line 892: Prebid Server is not a pass through service. We no longer accept using a publisher provided host and are working with existing bidders to fix their usage. There are other macros available such as account id. What part of your url is publisher dependent?

adapters/invibes/invibes.go
invibes: The documentation for GDPR property is: 
GDPR should be "1" if the caller believes the user is subject to GDPR laws, "0" if not, and undefined if it's unknown. 
Can we consider that in case of EU requests the GDPR property is always defined?

config/config.go
invibes: The subdomain can differ. We currently have two production domains plus two testing domains.
Is it ok if instead of Host parameter, we add a DomainId parameter (of type integer) and select in code one domain or the other?

adapters/invibes/invibes.go
SyntaxNode: The "core" of PBS does the best it can to infer this property. If the request doesn't explicitly set it, it'll look at the request's geo information (if available), and otherwise will fall back on host-level geo settings (a lot less precise). I believe it's possible for EU requests to still have an unknown GDPR applies at this point though.

Does that sound about right @hhhjort ?

config/config.go
SyntaxNode: Yes, I think that would be a good solution. In that case you may leave the `{{.Host}}` macro here and have the adapter choose the right one based on the DomainID parameter.

adapters/invibes/invibes.go
invibes: Thank you for the clarification.

adapters/invibes/invibes.go
invibes: No is not necessary. We have removed it

guscarreon (commented):
Adapter looks awesome. Some minor questions before approving

adapters/invibes/invibestest/exemplary/no-ad.json
```diff
@@ -0,0 +1,64 @@
...
+          "InvibBVLog": false,
+          "VideoAdDebug": false,
+          "Width": ""
+        }
+      },
+      "mockResponse": {
+        "status": 200,
+        "body": {
+          "VideoAdContentResult": {
+            "Ads": []
```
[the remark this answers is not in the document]
invibes: The response object contains many properties which can change in time. We would prefer not to declare the properties here, because we don't want to modify the adapter every time the structure of the response object changes.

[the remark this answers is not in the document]
invibes: VideoAdContentResult used in no-ad.json remained in this test file from a previos version of the adapter and make no sense in the current version (the response body doesn't contain this property anymore). I have removed it.

adapters/invibes/invibestest/exemplary/test-ad.json
```diff
@@ -0,0 +1,59 @@
...
+          "Location": "https://publisher.com/url",
+          "IsTestBid": true,
+          "InvibBVLog": false,
+          "VideoAdDebug": false,
+          "Width": ""
+        }
+      },
+      "mockResponse": {
+        "status": 200,
+        "body": {}
```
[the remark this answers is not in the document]
invibes: The "mockResponse" part of this test was just a mock. This shouldn't happen in reality. I've fixed it.

static/bidder-params/invibes.json
```diff
@@ -0,0 +1,30 @@
...
+        },
+        "testLog": {
+          "type": "boolean"
+        }
+      },
+      "description": "Parameters used for debugging porposes"
+    }
+  },
+  "required": ["placementId"]
+}
```
SyntaxNode, line 30: Please update your documentation [here](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/invibes.md) to:
- Include the new domainId field
- Include the debug field.
- Mention that the adContainerId is not supported by the server adapter.

SyntaxNode (approved):
Looks great. Thank you for your patience. The only comment I left is a request to update adapter docs.

static/bidder-params/invibes.json
invibes: Thank you for your assistance

SyntaxNode:
Docs PR: https://github.com/prebid/prebid.github.io/pull/2361

guscarreon (approved):
@invibes thank you for addressing the comments. Your adapter looks great.

PR #1471, opened 2020-09-03
Beachfront updates for multiple Imp AdM requests

muncha:
Beachfront now allows multiple impression requests to the video endpoint for AdM responses for a given exchange (ExtBeachfront.appId). These changes add a request type to take advantage of that and the logic involved in sorting and sending the different request types.


muncha:
The Travis error that I'm seeing right now seems to be that the script timed out. I don't see a way to manually re-run it, co I'm going to do a trivial change a push to get it to run again. Here's the error:

"...no output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

SyntaxNode:
> The Travis error that I'm seeing right now seems to be that the script timed out. I don't see a way to manually re-run it, so I'm going to do a trivial change and push to get it to run again. Here's the error:
> 
> "...no output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

We've been seeing timeouts frequently from Travis in pulling down modules. If we come across this kind of build failure, we'll restart the build. No worries.

adapters/beachfront/beachfront.go
```diff
@@ -103,11 +100,15 @@ type beachfrontResponseSlot struct {
...
-type beachfrontVideoBidExtension struct {
+type videoBidExtension struct {
 	Duration int `json:"duration"`
 }
 
+var one int8 = 1
+var zero int8 = 0
+
 func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var reqs = make([]*adapters.RequestData, 0)
```
hhhjort, line 111: We should size this to the max expected size, so we don't have to keep allocating larger and larger slices on the appends. Something like:
```
var reqCount = len(beachfrontRequests.ADMVideo) + len(beachfrontRequests.NurlVideo) + 1
var reqs = make([]*adapters.RequestData, 0, reqCount)
```

adapters/beachfront/beachfront.go
```diff
@@ -218,40 +216,82 @@ func preprocess(request *openrtb.BidRequest) (beachfrontReqs beachfrontRequests,
...
 		beachfrontReqs.Banner, errs = getBannerRequest(request)
 	}
 
+	/* We have video imps, so organize them into nurl imps and adm imps. The nurl imps will be
+	sent sequentially and the adm imps in sequential / parallel.
+	*/
 	if len(videoImps) > 0 {
-		var videoErrs []error
-		var videoList []beachfrontVideoRequest
+		admRequests := make(map[string]videoRequest, 0)
```
hhhjort, line 227: We might want to give go a hint to the map size with `admRequests := make(map[string]videoRequest, len(videoImps))` if the imp `AppId`s tend to be unique. If it is expected that the imps usually share the same `AppId`, then the current line should be fine.

adapters/beachfront/beachfront.go
```diff
@@ -218,40 +216,82 @@ func preprocess(request *openrtb.BidRequest) (beachfrontReqs beachfrontRequests,
 	}
 
 	if len(bannerImps) > 0 {
-		request.Imp = bannerImps
+		// request.Imp = bannerImps		// ... why was I doing this????
```
hhhjort, line 219: It looks like you were doing this to feed just the banner imps to `getBannerRequest(request)`

adapters/beachfront/beachfront.go
```diff
@@ -218,40 +216,82 @@ func preprocess(request *openrtb.BidRequest) (beachfrontReqs beachfrontRequests,
...
+*/
 func getAppId(ext openrtb_ext.ExtImpBeachfront, media openrtb_ext.BidType) (string, error) {
 	var appid string
-	var error error
-
-	if fmt.Sprintf("%s", reflect.TypeOf(ext.AppId)) == "string" &&
-		ext.AppId != "" {
+	var err error
 
+	if fmt.Sprintf("%s", reflect.TypeOf(ext.AppId)) == "string" && ext.AppId != "" {
```
hhhjort, line 294: How can `ext.AppId` be anything but a string? (including the nil string)

adapters/beachfront/beachfront.go
```diff
@@ -218,40 +216,82 @@ func preprocess(request *openrtb.BidRequest) (beachfrontReqs beachfrontRequests,
...
 		appid = ext.AppId
 	} else if fmt.Sprintf("%s", reflect.TypeOf(ext.AppIds)) == "openrtb_ext.ExtImpBeachfrontAppIds" {
```
hhhjort, line 296: Again, how can this variable be anything other than the declared type? `reflect.TypeOf()` is intended to be used on interfaces, which could have a range of concrete types implementing them.

adapters/beachfront/beachfront.go
SyntaxNode: Perhaps the `AppId` field used to be an interface and the code checked for it being either a string or an array? Just conjecture.

This looks like a good opportunity to remove it. `AppId` is typed to a string, making this check (and others like it) superfluous.

muncha: These are left over checks that I neglected to eliminate from when I was experimentally trying to have a single parameter, appId, that could be either a (json) string or a (json) object. It looks to me like you just can't do that in Go.  I have these eliminated in my current fixes, and it works.

adapters/beachfront/beachfront.go
muncha: I was concerned about making them too long, with null members at the end, but since they don't get returned, and I never call len() on any of them or iterate through them, just pull from them to append members to the slice I'm building for return, I think that's not a problem. I've identified a couple of other concerns similar to this. I'm not sure about the relative expense in one particular situation, though.  At 187 in the current commit: 

```
func preprocess(request *openrtb.BidRequest) (beachfrontReqs requests, errs []error) {
	var videoImps = make([]openrtb.Imp, 0)
	var bannerImps = make([]openrtb.Imp, 0)
```

At this point I don't know if I have any videoImps or bannerImps, so I started them at 0, then as I find Imps, I append to them. Considering that they are pretty much throw away slices, I would be better off for efficiency, as above, making some extra then assigning instead of appending, like this:

```
bannerImps = make([]openrtb.Imp, len(request.Imp))
videoImps = make([]openrtb.Imp, len(request.Imp))

```
What I'm not sure about is maybe I would be better off defining them like this:

```
	var videoImps []openrtb.Imp
	var bannerImps []openrtb.Imp
```

And only make the slice when I hit my first imp of that type, and make it len(request.Imp), something like this:

```
		if request.Imp[i].Video != nil {
			if noVideoImpsYet {
				noVideoImpsYet = false
				videoImps = make([]openrtb.Imp, 1)
			}
```
But does checking those couple of "if"s and reassigning the flags add significant overhead that would indicate I should just make them len(request.Imp) long in the first place?

adapters/beachfront/beachfront.go
muncha: So, what @SyntaxNode said.

adapters/beachfront/beachfront.go
hhhjort: You want to take advantage of the third parameter in `make()`. If you have:
```
bannerImps = make([]openrtb.Imp, 0, len(request.Imp))
videoImps = make([]openrtb.Imp, 0, len(request.Imp))
```
you will have zero length slices where you will have to use append to add any elements to the slice. So the size of the slice is 0, but the capacity of the slice is `len(request.Imp)`. In Go, each slice points to an existing, allocated array. The above makes create the slices pointing to an array that has `len(request.Imp)` items of memory already allocated. As you append items, the slice will expand into the unused elements of the array that are sitting there waiting to be used. If you try to append past the capacity, then Go will allocate a new, larger array, copy the contents of the original to the larger, and then put the new item into the next element. This is why we like to avoid these sorts of reallocations when we have some idea what size could be.

adapters/beachfront/beachfront.go
SyntaxNode: > It looks to me like you just can't do that in Go.

You can. The json deserializer doesn't know about the `ExtImpBeachfrontAppIds` type, so it'll fall back to a generic map.

Take a look:
https://play.golang.org/p/9MDzerG-flZ

hhhjort (approved):
It is looking good now, thanks.

adapters/beachfront/beachfront.go
```diff
@@ -218,65 +218,113 @@ func preprocess(request *openrtb.BidRequest) (beachfrontReqs beachfrontRequests,
...
+
+		for i := 0; i < len(videoImps); i++ {
+			var ext openrtb_ext.ExtImpBeachfront
+			ext, errs = prepVideoRequestExt(videoImps[i], errs)
+
+			if ext.VideoResponseType == "nurl" || ext.VideoResponseType == "both" {
+				requestStub := *request
+				requestStub.Imp = nil
+
+				beachfrontReqs.NurlVideo = append(beachfrontReqs.NurlVideo, videoRequest{
```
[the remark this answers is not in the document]
muncha: I get what you mean with the pointer, but would you say this also covers those issues, without introducing any new ones?
Would you maybe consider the anonymous videoRequest a mark against readability?
```go

			if ext.VideoResponseType == "nurl" || ext.VideoResponseType == "both" {
				requestStub := *request
				requestStub.Imp = make([]openrtb.Imp, 1)
				requestStub.Imp = append(requestStub.Imp, videoImps[i])

				beachfrontReqs.NurlVideo = append(beachfrontReqs.NurlVideo,
					prepVideoRequest(
						videoRequest{
							AppId:             ext.AppId,
							VideoResponseType: ext.VideoResponseType,
							Request:           requestStub,
						}),
				)
			}
```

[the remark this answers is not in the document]
muncha: I guess I'm still passing that data structure around and it would be less expensive to use a pointer.

adapters/beachfront/beachfront.go
```diff
@@ -226,50 +226,45 @@ func preprocess(request *openrtb.BidRequest) (beachfrontReqs requests, errs []er
...
 
 		for i := 0; i < len(videoImps); i++ {
 			var ext openrtb_ext.ExtImpBeachfront
 			ext, errs = prepVideoRequestExt(videoImps[i], errs)
 
 			if ext.VideoResponseType == "nurl" || ext.VideoResponseType == "both" {
 				requestStub := *request
-				requestStub.Imp = nil
+				requestStub.Imp = make([]openrtb.Imp, 1)
+				requestStub.Imp = append(requestStub.Imp, videoImps[i])
```
mansinahar, line 237: You can just merge these two lines into one as follows:
```
request.Imp = []openrtb.Imp{videoImps[i]}
```

adapters/beachfront/beachfront.go
```diff
@@ -218,65 +218,113 @@ func preprocess(request *openrtb.BidRequest) (beachfrontReqs beachfrontRequests,
...
+				beachfrontReqs.NurlVideo[j].Request.Imp = append(
+					beachfrontReqs.NurlVideo[j].Request.Imp, videoImps[i])
+				beachfrontReqs.NurlVideo[j] = prepVideoRequest(beachfrontReqs.NurlVideo[j])
+				j = j + 1
 			}
 
-			if videoList[i].VideoResponseType == "adm" || videoList[i].VideoResponseType == "both" {
-				beachfrontReqs.ADMVideo = append(beachfrontReqs.ADMVideo, videoList[i])
+			if ext.VideoResponseType == "adm" || ext.VideoResponseType == "both" {
+				admRequest, exists := admRequests[ext.AppId]
```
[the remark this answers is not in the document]
muncha: That did make sense, and I took a stab at doing as you describe, but the "both" option hung me up, so I tried this. I think  that what I have now is more efficient than what you describe? Fewer loops anyway. I've got a range loop on the map, but really that is only going to be a couple of members long if that; usually just a single member, where as the full "imps" slice, containing all video imps, will generally be significantly longer, but I'm only going through that once now.  I am now copying and replacing "videoRequest" structs in the map, which I'm sure has some overhead, but not nearly as much a nested loops, I would think. Much like [this guy](https://stackoverflow.com/questions/46492318/go-append-directly-to-slice-found-in-a-map), I would have thought that I could just append to a slice inside a map, but I'm assuming that appending to a copy, then replacing it in the map is still pretty reasonable. 

If you suspect that a) this should be AB tested or b) I'm just plain wrong, no worries. I'm sure I can work out how to handle "both" and do it as you describe.

muncha:
@SyntaxNode  : could you please run that test again?

panic: test timed out after 10m0s

hhhjort (approved):

adapters/beachfront/beachfront.go
```diff
@@ -291,195 +285,215 @@ func getBannerRequest(request *openrtb.BidRequest) (beachfrontBannerRequest, []e
...
-			continue
+			prepVideoRequest(&r)
+			nurlReqs = append(nurlReqs, r)
 		}
 
-		appid, err := getAppId(beachfrontExt, openrtb_ext.BidTypeVideo)
+		if ext.VideoResponseType == "adm" || ext.VideoResponseType == "both" {
+			_, exists := admMap[ext.AppId]
+			if !exists {
+				requestStub.Imp = make([]openrtb.Imp, 0, len(imps))
```
mansinahar, line 398: Any reason why the capacity for this slice is being set to `len(imps)`? You always append only one `imp` to that slice which is in the next line. I'd suggest just merging these two lines as the following:
```
requestStub.Imp = []openrtb.Imp{imps[i]}
```

adapters/beachfront/beachfront.go
```diff
@@ -291,195 +285,215 @@ func getBannerRequest(request *openrtb.BidRequest) (beachfrontBannerRequest, []e
...
+					VideoResponseType: ext.VideoResponseType,
+					Request:           requestStub,
+				}
+				prepVideoRequest(&admRequest)
+				admMap[ext.AppId] = admRequest
+			} else {
+				tmpRequest := admMap[ext.AppId].Request
+				tmpRequest.Imp = append(tmpRequest.Imp, imps[i])
+
+				admMap[ext.AppId] = videoRequest{
```
mansinahar, line 412: Any reason why you need to reassign the request in `admMap`? You have already appended the new imp in above two lines. All the other values for the request should already be same, no?

adapters/beachfront/beachfront.go
```diff
@@ -291,195 +285,215 @@ func getBannerRequest(request *openrtb.BidRequest) (beachfrontBannerRequest, []e
...
+	if bfReq.Request.Device != nil {
+		bfReq.Request.Device.IP = getIP(bfReq.Request.Device.IP)
+
+		if bfReq.Request.Device.DeviceType == 0 {
+			bfReq.Request.Device.DeviceType = fallBackDeviceType(&bfReq.Request)
 		}
+	}
 
+	if len(bfReq.Request.Cur) == 0 {
+		bfReq.Request.Cur = make([]string, 1)
```
mansinahar, line 494: Please merge these two lines into the following:
```
bfReq.Request.Cur = []string{"USD"}
```

adapters/beachfront/beachfront.go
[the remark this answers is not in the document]
mansinahar: https://github.com/prebid/prebid-server/pull/1471#discussion_r492354804

adapters/beachfront/beachfront.go
```diff
@@ -658,15 +675,13 @@ func getDomain(page string) string {
 
 }
 
-func isSecure(page string) int8 {
+func isPageSecure(page string) (int8, *int8) {
```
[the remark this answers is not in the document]
muncha: That would be less weird.

adapters/beachfront/beachfront.go
```diff
@@ -197,86 +186,91 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
 
-	if len(bannerImps)+len(videoImps) == 0 {
+	if len(videoImps) == 0 && !gotBanner {
 		errs = append(errs, errors.New("no valid impressions were found in the request"))
 		return
 	}
 
-	if len(bannerImps) > 0 {
-		request.Imp = bannerImps
+	if gotBanner {
```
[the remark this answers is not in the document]
muncha: It is possible to have errors from getBannerRequest and still have a valid banner, so I don't see any appropriate action that I could  take _exactly_ here, but I found several related optimizations. If all banners errored out (failed on ext or appId), the Slots will be empty, so I'm now checking that in getBannerRequest before adding anything else (ip, secure. etc), so in that case I now skip a bunch of unnecessary processing on an empty banner request.

I realized that how I had it, if I did get banner errors, they would get overwritten by the errs returned from getVideoRequests, even if the later is nil. That's no good. Previously (current master) I had merged []error responses.  I changed that to pass a []error pointer to preprocess and down the line. That creates a fairly big cascade of changes, but I think it is much better, not just because it avoids that overwrite.

I removed the test in prepVideoRequestExt to verify a video element, as it is impossible for anything that is not a video imp to get into videoImps (L211), so that was just a waste of a few cpu cycles.

muncha:
@mansinahar - please don't approve this yet. I'd like to standardize how I'm adding errors, and I'd like to see if I can use the errorTypes instead of errors.New while adding non fatal errors.

muncha:
@mansinahar Looks ok to me. I didn't do anything with the errorTypes, but improved some error formatting.

muncha:
@mansinahar - also, I considered making impsToSlots take a pointer to the []openrtb.Imps, but figured that since it's a small struct, with Video, Banner, etc just being pointers, there would be little gain there. Or would you think that would be an improvement?

adapters/beachfront/beachfront.go
```diff
@@ -103,12 +99,17 @@ type beachfrontResponseSlot struct {
...
 
-type beachfrontVideoBidExtension struct {
+type videoBidExtension struct {
 	Duration int `json:"duration"`
 }
 
 func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
-	beachfrontRequests, errs := preprocess(request)
+	var errs []error
+	beachfrontRequests := preprocess(request, &errs)
```
mansinahar, line 108: Please consider adding an error check after the `preprocess` function call. Also, you don't need to pass in the `errs` slice to the function. You can have the function return an error slice and then inspect that after the function is called or append that to `errs` slice if you'd like to move on but return the errors towards the end.

adapters/beachfront/beachfront.go
```diff
@@ -197,289 +184,327 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
 	return reqs, errs
 }
 
-func preprocess(request *openrtb.BidRequest) (beachfrontReqs beachfrontRequests, errs []error) {
-	var videoImps = make([]openrtb.Imp, 0)
-	var bannerImps = make([]openrtb.Imp, 0)
+func testForValidBanner(request *openrtb.BidRequest, index int) bool {
```
mansinahar, line 187: Please consider passing the imp under inspection to this function rather than passing the entire request and the index of the imp as this function requires only the imp to do what it needs to do

adapters/beachfront/beachfront.go
```diff
@@ -197,289 +184,327 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
+	if request.Imp[index].Banner != nil && ((request.Imp[index].Banner.Format != nil &&
+
+		len(request.Imp[index].Banner.Format) != 0 &&
```
mansinahar, line 190: You can merge the first two conditions here to be:
```
len(request.Imp[index].Banner.Format) > 0
```

adapters/beachfront/beachfront.go
```diff
@@ -197,289 +184,327 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
+			imp := request.Imp[i]
+			var secure int8 = 0
+
+			if request.Site != nil && request.Site.Page != "" {
+				secure = isPageSecure(request.Site.Page)
+				imp.Secure = &secure
+			} else {
+				// &secure = 0 at this point
+				if imp.Secure != nil && imp.Secure != &secure {
+					imp.Secure = request.Imp[i].Secure
```
mansinahar, line 222: How are `imp.Secure` and `request.Imp[i].Secure` going to be different here? L214 assigns `imp` to be `request.Imp[i]` so they should be the same, no?

adapters/beachfront/beachfront.go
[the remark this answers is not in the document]
mansinahar: @muncha Sorry but I am not sure I follow why you needed to pass the `errs` slice to `preprocess` rather than having it send back an error slice

adapters/beachfront/beachfront.go
```diff
@@ -197,289 +184,327 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
-	} else {
-		error = errors.New("unable to determine the appId(s) from the supplied extension")
 	}
 
-	return appid, error
+	if appid == "" {
+		err = errors.New("unable to determine the appId(s) from the supplied extension")
+	}
+
+	if appid == "throwTestError" {
```
mansinahar, line 272: Whats the rationale behind having a test specific app ID in the code here? Anything test specific should be in the test files. This isn't considered as a best practice

mansinahar:
> @mansinahar - also, I considered making impsToSlots take a pointer to the []openrtb.Imps, but figured that since it's a small struct, with Video, Banner, etc just being pointers, there would be little gain there. Or would you think that would be an improvement?

You'd only need to pass a pointer to that slice if you're gonna modify it which isn't what you're doing. You are just going through the imps and creating slots based on those imps so in this case passing just the slice is fine.

muncha:
@mansinahar - No worries, and thank you so much for taking the time. I'm in no rush on this.

adapters/beachfront/beachfront.go
muncha: Thank you. I'll find a better way of doing this. I need to dig into how the test files work a lot more thoroughly.

adapters/beachfront/beachfront.go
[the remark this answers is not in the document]
muncha: Right. Duh. I was looking for places that I could use pointers, but here there is really no point, as it is nil when I send it in.

In my current local changes, I've eliminated passing the []error in and have preprocess back to returning it. In preprocess, I am creating the slice and passing a pointer to that []error to getBannerRequest and getVideoRequests.

adapters/beachfront/beachfront.go
muncha: I appears there is no reason for me to do any of this anyway. The banner sizes are all verified in auction.go around L441. If the sizes are not valid, the entire request would have gotten dumped. All I really need to do is check if there is a Banner. It is already known to be valid.

In my local changes I'm now just doing this:

```Go
		if !wantsBanner {
			wantsBanner = request.Imp[i].Banner != nil
		}
```

adapters/beachfront/beachfront.go
muncha: Right. But also, if there is no value for page in the beachfront backend request, it will return null (204). RTB will allow request.Site.Domain or request.Site.Id, but beachfront won't. So now (still locally), if request.Site.Page is not set, I'm just dropping the imp.

muncha: Hum. But I gotta check for App too.

hhhjort:
> > @mansinahar - also, I considered making impsToSlots take a pointer to the []openrtb.Imps, but figured that since it's a small struct, with Video, Banner, etc just being pointers, there would be little gain there. Or would you think that would be an improvement?
> 
> You'd only need to pass a pointer to that slice if you're gonna modify it which isn't what you're doing. You are just going through the imps and creating slots based on those imps so in this case passing just the slice is fine.
> 
Passing the slice passes two ints(size and capacity) and a pointer to the underlying array. Minor overhead compared to passing a pointer to the slice. I think it is worth passing the slice to make it clear that the slice is not expected to be modified, it makes the code more readable.

adapters/beachfront/beachfront.go
hhhjort: I prefer Mansi's suggestion. It also allows you to know if the call produced new errors, or if the errors in the slice are old from elsewhere. You don't currently need to check for that, but very useful if you have the potential for a fatal error that you want to abort processing, you just need to check the new errors rather than traversing the entire array to see if a new fatal error was generated.

muncha: Here's what I've currently got locally:

muncha:
Travis failed on 1.13, but the github test on 1.13 passed. It looks like the later is using go version go1.13.15 linux/amd64, and the former go1.13.linux.amd64. The fail is a result of my "all-the-calls" test http calls running out of order. I'm installing 1.13 to see if I can reproduce that. I'm on 1.15.2

I can't imagine it actually effecting anything, the $PATH that travis is using includes 1.11, but 1.13 is first. Just seems a bit odd. 

Also, as far as the code; hhhjort's comment about how passing a slice is really just passing two ints and a pointer, my passing a pointer to the error slice in preprocess to getVideoRequests (and deeper) and getBannerRequest seems to make less sense. Mansinahar made a similar comment, but at the time I thought the critical thing was that I was passing in a pointer to what would then always be an empty slice. I'm getting now that an empty slice and a slice of arbitrary size are the same size, which is nuts and totally cool. So I'm doing a quick change to make those all just take preprocess' error slice in and return it modified, as I think that is a readability improvement.

muncha:
Locally, on 1.15.2, I ran the validation command that travis is running, and it failed on "all-the-calls" similar to the 1.13 fail on that first run. Then, changing nothing, I ran it agian and it passed. Then I pushed and it failed similarly on 1.15. So it looks like "all-the-calls" is indeterminate as far as the order of the http requests. I'm looking into that.

mansinahar:
> Locally, on 1.15.2, I ran the validation command that travis is running, and it failed on "all-the-calls" similar to the 1.13 fail on that first run. Then, changing nothing, I ran it agian and it passed. Then I pushed and it failed similarly on 1.15. So it looks like "all-the-calls" is indeterminate as far as the order of the http requests. I'm looking into that.

Hmmm.. Sounds like it could be a potential race condition. Let us know what you find or if you need help looking into it as all these json tests are going through a common `RunJSONBidderTest` function

mansinahar:
Hey @muncha! Any luck finding the cause of the failing test?

muncha:
I've been on some other work, but should be able to concentrate on this tomorrow.

muncha:
I haven't quite nailed it yet, but it looks like it is not a race condition. It has something to do with the uris getting mixed up, related to the admMap and the way I'm doing the requestStub bit. I think.

adapters/beachfront/beachfront.go
```diff
@@ -383,103 +470,56 @@ func fallBackDeviceType(request *openrtb.BidRequest) openrtb.DeviceType {
...
+func prepVideoRequest(bfReq videoRequest) videoRequest {
+	if bfReq.Request.Site != nil && bfReq.Request.Site.Domain == "" && bfReq.Request.Site.Page != "" {
+		bfReq.Request.Site.Domain = getDomain(bfReq.Request.Site.Page)
+	}
 
-		if bfReqs[i].Request.App != nil && bfReqs[i].Request.App.Domain == "" && bfReqs[i].Request.App.Bundle != "" {
-			if bfReqs[i].Request.App.Bundle != "" {
-				var chunks = strings.Split(strings.Trim(bfReqs[i].Request.App.Bundle, "_"), ".")
+	if bfReq.Request.App != nil && bfReq.Request.App.Domain == "" && bfReq.Request.App.Bundle != "" {
+		if bfReq.Request.App.Bundle != "" {
```
mansinahar, line 499: Looks like this condition is redundant. It's present in the above if condition as well and so you can skip this next one

adapters/beachfront/beachfronttest/params/race/video.json
```diff
@@ -1,4 +1,4 @@
 {
-  "bidfloor":0.01,
-  "appId":"11bc5dd5-7421-4dd8-c926-40fa653bec76"
+  "bidfloor": 0.01,
+  "appId":" 11bc5dd5-7421-4dd8-c926-40fa653bec76"
```
mansinahar, line 3: Don't you need `videoResponseType` here as well?

mansinahar (commented):
Looks fine to me except the one minor comment but also have a question, what's the rationale behind removing the banner and unmarshal related tests. They should still be useful, no?

muncha:
@mansinahar - Uhg! This is the same fail.

mansinahar:
Ughh! Sorry about that @muncha. Have you tried using the [golang race detector](https://golang.org/doc/articles/race_detector.html) to see if that helps figure what's causing this?

muncha:
No I haven't. Thanks.

adapters/beachfront/beachfront.go
```diff
@@ -197,182 +188,278 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
-	} else if fmt.Sprintf("%s", reflect.TypeOf(ext.AppIds)) == "openrtb_ext.ExtImpBeachfrontAppIds" {
-		if media == openrtb_ext.BidTypeVideo && ext.AppIds.Video != "" {
-			appid = ext.AppIds.Video
-		} else if media == openrtb_ext.BidTypeBanner && ext.AppIds.Banner != "" {
-			appid = ext.AppIds.Banner
+		bannerReq.IsMobile = 0
+	}
+
+	if request.User != nil && request.User.ID != "" {
+		if bannerReq.User.ID == "" {
```
hhhjort, line 335: Is it possible for `bannerReq.User.ID` to have a value at this point?

adapters/beachfront/beachfront.go
```diff
@@ -197,182 +188,278 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
+				W: imp.Banner.Format[j].W,
+			})
+		}
 
-			slot.Sizes = append(slot.Sizes, beachfrontSize{
-				H: request.Imp[i].Banner.Format[j].H,
-				W: request.Imp[i].Banner.Format[j].W,
+		if len(slot.Sizes) == 0 {
+			slot.Sizes = append(slot.Sizes, size{
+				W: *imp.Banner.W,
```
hhhjort, line 401: What happens here if `imp.Banner.W` is nil?

adapters/beachfront/beachfront.go
```diff
@@ -383,103 +470,54 @@ func fallBackDeviceType(request *openrtb.BidRequest) openrtb.DeviceType {
...
-		imp := request.Imp[i]
+	return beachfrontExt, err
+}
 
-		imp.Banner = nil
-		imp.Ext = nil
-		imp.Secure = &secure
+func prepVideoRequest(bfReq videoRequest) videoRequest {
+	if bfReq.Request.Site != nil && bfReq.Request.Site.Domain == "" && bfReq.Request.Site.Page != "" {
+		bfReq.Request.Site.Domain = getDomain(bfReq.Request.Site.Page)
```
hhhjort, line 495: This could be a race condition. `bfReq.Request` is a shallow copy of the incoming openrtb request. `bfReq.Request.Site` is a pointer to the `Site` object of the original openrtb request, which is shared across all adapters. So this line changes the `Site` object shared across all adapters. The same goes for modifying `App.Domain` below.

adapters/beachfront/beachfront.go
muncha: No it is no. I've fixed that locally.

adapters/beachfront/beachfront.go
muncha: len(slot.Sizes)==0 would only be true at this point (L399) if imp.Banner.Format had been length 0. If I understand correctly, imp.Banner.Format can't be nil as defined in mxmcherry banner.go, so I'm not checking that. If both imp.Banner.Format were length 0, and therefore len(slot.Sizes)==0 is true, and imp.Banner.W or imp.Banner.H were nil, it would not have passed the check at /endpoints/openrtb2/auction.go L441 (line number may have changed of course, but in the function validateBanner), and the entire request would have been discarded with an error.

If I'm not mistaken.

hhhjort: Makes sense. I was trying to figure out what could be causing your race condition, and this appeared a little suspicious. I thought we might have validation for that, but I wanted to at least flag this before diving deeper into your code. I believe I found a real issue below.

muncha: I made a quick test case for that, and I do seem to be mistaken. It threw a segmentation fault.

adapters/beachfront/beachfront.go
muncha: @hhhjort , As per @mansinahar's suggestion, I ran a few variations with the race detector. I didn't get anything definitive out of doing that directly, but it got me looking into how the validation.sh script runs the race tests. I get now that I was getting the error, same error every time, not because the validation checks were running a race test, but just because it was hitting that race condition in normal validation execution, since the race tests only include a subset of the tests, indicated by the "TestRace" prefix. I've added TestRaceJsonSamples() to beachfront_test.go and added a directory for race tests, which currently just includes a copy of my "all-the-calls" test. It seems that the race was happening about 1/10 times the test was run, though I did at one point run 20 tests in a row with no error. Adding it to the race tests run by './validate.sh --nofmt --cov --race 10' throws the error about 1/2 the time or a little more. 

From what I read on mansinahar's link above, the bit that you point out with the shallow copy is pretty much what they describe as a typical race. So I built a deep copy function. I'm pretty sure that I got that right. **I still get the error**. I'm about to go ahead and push everything, but it will fail the validation. I would appreciate it if you could have a look at my deep copy function. Maybe I'm missing something with that. Even if it is not the cause of this error, what I had before totally looks like a race condition.

Assuming that my deep copy is OK, I'm continuing to hunt this down with a better understanding of what I'm looking for and much more consistently able to reproduce the error.

adapters/beachfront/beachfront.go
muncha: @hhhjort - So do you think that I should keep my check for defined H and W, which appears to only come into play with the test specifically made for it, but not in regular execution? Or delete both the check and the test?

adapters/beachfront/beachfront.go
muncha: I think I see a problem and I'm attempting a solution which involves making multiple deep copies. I'm not sure if that's OK from a processing expense standpoint, but checking if it at least works locally.

adapters/beachfront/beachfront.go
```diff
@@ -397,10 +481,23 @@ func impsToSlots(imps []openrtb.Imp, errs []error) (bannerRequest, int8, []error
...
 		if len(slot.Sizes) == 0 {
-			slot.Sizes = append(slot.Sizes, size{
-				W: *imp.Banner.W,
-				H: *imp.Banner.H,
-			})
+			// *imp.Banner.W == nil is not valid. I have to dereference to make this comparison. Also, this does
+			// get caught in /endpoints/openrtb2/auction.go L441 when I send in the request from test case
+			// "banner-no-size-nose-dive", but it gets through on that test case.
+
+			impDeref := &imp
```
hhhjort, line 488: I don't think you need `impDeref` here. just checking `if imp.Banner.W == nil || imp.Banner.H == nil` should work. You are already doing a dereferencing in the `*imp.Banner.W` construction, this `impDeref` is a pointer to the imp structure that is then implicitly dereferenced in your code.

adapters/beachfront/beachfront.go
hhhjort: Yes, in general deep copies are expensive. Once we get a handle on all the problems, we should convert the deep cop(y/ies) into a semi-deep copy, only copying the pieces we have identified as being subject to change.

adapters/beachfront/beachfront.go
muncha: I don't need it. Removed locally. I had copy pasted the "*imp.Banner.W" from a few lines below and gotten an error that I couldn't compare that with nil, and I can't. Just needed to take the "*" off. Dumb mistake.

adapters/beachfront/beachfront.go
```diff
@@ -197,289 +188,433 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
 		}
 
 		if request.Imp[i].Video != nil {
-			videoImps = append(videoImps, request.Imp[i])
+			if request.Site != nil {
+				if request.Site.Page != "" {
+					var secure int8 = 0
+					// This will overwrite the secure value if it is included, but if
+					// a web request is not https, it's not secure, and vice versa so this seems authoritative.
+					secure = isPageSecure(request.Site.Page)
```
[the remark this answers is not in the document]
muncha: Correct. Changed locally.

adapters/beachfront/beachfront.go
```diff
@@ -197,289 +188,433 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
 	}
 
 	var t = fallBackDeviceType(request)
 
 	if t == openrtb.DeviceTypeMobileTablet {
-		bfr.Page = request.App.Bundle
+		bannerReq.Page = request.App.Bundle
 		if request.App.Domain == "" {
-			bfr.Domain = getDomain(request.App.Domain)
+			bannerReq.Domain = getDomain(request.App.Domain)
```
hhhjort, line 403: Shouldn't this be `getDomain(request.App.Bundle)` or something. As current, you are calling `getDomain("")` and should get back `""`, making this a noop.

muncha: Right. Changed to this locally:

`
		if request.App.Domain != "" {
			bannerReq.Domain = request.App.Domain
		}`

adapters/beachfront/beachfront.go
muncha: Locally I have changed getVideoRequests back to taking a pointer to request, then I'm making my deep copies when I create each videoRequest, either nurl or adm to put in the admMap. I think that is more correct, as they aren't then modifying/referencing  each other. I also changed the deepCopyRequest function to only reference requestCopy and it's dereferenced children once it is created, instead of referencing request, which I think improves readability, but nothing else as far as I can tell. 

But I stil get the error on ./validate.sh --nofmt --cov --race 10

hhhjort: Yes, we should focus on finding the race condition first, then we can tighten things like this up. Broad, expensive fixes like this can tell us if we are in the right ballpark or not, so they are certainly useful that way. My current suspicion is that having a video imp that has both adm and nurl, or just the combo of video and banner, are somehow colliding each other. They both will make use of the same imp. I just haven't found anywhere where you are writing to a shared imp that would cause the issue to appear. The search has uncovered a few nitpicks as you have seen. :)

adapters/beachfront/beachfront.go
hhhjort: I suppose I slightly lean to deleting them, as it has been shown we have verification elsewhere. I don't have a strong feeling either way.

adapters/beachfront/beachfront.go
muncha: The critical thing that got these changes going on my end is that if an adm request is made to the adapter with multiple impressions, beachfront wants a single request to their backend with all of those impressions, instead of a request for each impression as in the current master. A lot of this trying to mix things up with different types of requests is not really part of their requirement, but seemed like added functionality that could be included. Maybe it can't. If that's the case, maybe I need to take a step back and establish some rules for what requests will be accepted. As far as real life situations, this test is really an extreme edge case. I am up for a challenge though, and not giving up just yet.

Since I created the TestRaceJsonSamples() I'll come up with several tests that hammer it, but not _all_ of the possible combinations at once and see what happens with that.

Thank you for taking the time to nitpick.

adapters/beachfront/beachfront.go
```diff
@@ -197,289 +205,446 @@ func (a *BeachfrontAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *a
...
 		}
 
-		if imp.Video.H == 0 && imp.Video.W == 0 {
-			imp.Video.W = DefaultVideoWidth
-			imp.Video.H = DefaultVideoHeight
+		if ext.VideoResponseType == "nurl" || ext.VideoResponseType == "both" {
+			r := deepCopyRequest(request)
+			r.Imp = []openrtb.Imp{imps[i]}
+
+			nurlReqs = append(nurlReqs, prepVideoRequest(r))
```
hhhjort, line 582: Sounds like you are making some progress. I am not sure if this will help the problem, but it should make things more efficient in any case. Instead of running `prepVideoRequest()` for each request you generate, run it once on your deep copy of the original request, then do a shallow copy of the prepped request here in order to replace `request.imp` in this instance of your generated requests. If it is somehow `prepVideoRequest()` that is somehow stepping on its own toes, then running it only once will keep that from happening. In any case, it doesn't look like there is any dependence on the `imp` in prep, which means it should be doing the same exact thing every time you execute it. This will at least make your adapter more efficient even if it doesn't fix the race issue.

muncha: Weirdly, I don't seem to be able to comment on here any more. Unless you see this comment, in which case I now can.

muncha: It was Brave browser, blocking that which it should not block ;)

muncha:
Hi @mansinahar and @hhhjort. I _still_ haven't given up on this. I'm working from the recent changes you did cleaning up my test files (#1578), without any of the changes I made here for the support of the changes to beachfront's adm endpoint. It was just the one test file that was causing / exposing what appears to be a race of some kind, and I have determined that it was not a result of the changes in this PR, i.e., after modifying the test file (variously called "all-the-calls.json" and "sledge-hammer.json") to do per-imp adm calls and running against the current master (more or less - found a few minor bugs to fix along the way), I get the same apparent race condition. 

In testing that I have found that the error occurs when, in the httpcalls[] of the test file, I have any nurls after the adms. So if the httpcalls[] are ordered banners... adm video ... nurl video, it works. Otherwise it mixes the urls for the adms in with the bodys for the nurls. I'm sure that this is a reflection of the order that I'm building the requests in MakeRequests, in other words the order that the three types appear in httpcalls[] in the test file is tightly coupled with the order that I'm building the slice in MakeRequests. So the goal is just to decouple that. I'm working now on solving that problem independent of the changes here to support the modified endpoint.

I've pushed that branch to https://github.com/beachfront/prebid-server/tree/testSledgeHammer.

So, does anything need to be done with this PR? Just let it go stale? When I do get this worked out, I'll proceed with the modifications for the endpoint changes, but that will be from a new starting point. Maybe I'll be able to merge this branch or maybe not, but regardless, I would think that would be a new PR, right?

mansinahar:
@muncha Yeah, I think based on your plan to solve the race condition first and then do the endpoint modifications, it makes sense to just close this PR and then re-open a new one from either a new branch or this same branch (based on your preference/convenience for dev). I have no objections!

muncha:
I'm closing this PR up and addressing the preexisting race condition issue that I discovered in MakeRequests(). I will revisit this issue in the future.

PR #1476, opened 2020-09-03
github-actions validation checker for PRs and master branch

laurb9:
Add github-actions PR and master validation checker.

This works everywhere the repo was forked or mirrored without any additional configuration, unlike the Travis setup.

Reference: https://github.com/mvdan/github-actions-golang

Example results because it's not running in the main prebid repo:
https://github.com/openx/prebid-server-openx/pull/7

Screen Shot 2020-09-03 at 10 48 45 AM

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,28 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  release:
+    types:
+      - created
+name: Validate
+jobs:
+  Go:
+    strategy:
+      matrix:
+        go-version: [1.13.x, 1.14.x, 1.15.x]
+        os: [ubuntu-18.04]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Validate
+      run: |
+        ./validate.sh --nofmt --cov --race 10
+      env:
+        GO111MODULE: "on"


SyntaxNode:
Interesting. I'm not familiar with GitHub Workflows. A few questions...

Is this a proposal to replace Travis with GitHub Workflows?
Is this a service provided by Github itself? aka.. who's servers are running the validations?
Is there an equivalent badge we can use to replace the Travis Build Status icon?
How and who can trigger a rebuild?

Travis has been very flaky recently, so I'm game to try something new.

laurb9:
GH action is intended to do CI/CD and more. It's containers running on GH infrastructure, with read-only access to the repo by default. This one is very simple because all it does is greenlight the PR or branch so it definitely can replace Travis in that respect. Builds are possible too using secret tokens but I didn't do that here.

https://docs.github.com/en/actions

You can use pre-existing (standard) actions, marketplace actions or custom containers.

https://github.com/marketplace?type=actions

GH actions are enabled by default, I suspect it has not run here because I am not in the org and it's not already in the repo.

SyntaxNode:
I'm a big fan of this. I'll discuss with the team internally next week.

SyntaxNode (dismissed):
Happy to give it a try. This doesn't look like it disables Travis. We'll run both side by side for a bit until we're comfortable with this as a replacement.

.github/workflows/validate.yml
```diff
@@ -0,0 +1,28 @@
...
+  release:
+    types:
+      - created
+name: Validate
+jobs:
+  Go:
+    strategy:
+      matrix:
+        go-version: [1.13.x, 1.14.x, 1.15.x]
+        os: [ubuntu-latest]
```
hhhjort, line 15: In the docker file we build the image on ubuntu 18.04 ... should we mirror that perhaps?

laurb9: I can change it to ubuntu-18.04, or have both ubuntu-18.04 and ubuntu-20.04.

At the moment ubuntu-latest is the same as ubuntu-18.04: https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

SyntaxNode: > I can change it to ubuntu-18.04, or have both ubuntu-18.04 and ubuntu-20.04.

How about just ubuntu-18.04 and we'll keep it updated along with the docker file. Don't want this matrix to get too large.

SyntaxNode (approved):

hhhjort (approved):

Issue #879, opened 2019-04-17
First Party Data support


bretg:
[This is the Prebid Server side of https://github.com/prebid/Prebid.js/issues/3687]

## Updates

This issue has been somewhat significantly modified since the original post

1. More detail mapping the sources of the OpenRTB attributes. Providing a reference for how FPD works end-to-end. 
2. Normalize some OpenRTB attributes. We found that real-world FPD already coming in is not always in an expected form.
3. No need for imp[].ext.context.ATTRIBUTE. There aren’t any OpenRTB attributes here. Everything should go in imp[].ext.context.data.ATTRIBUTE.
4. Expand AMP first party data processing.
5. Clarify that errors in validating FPD should not cancel the auction. Prebid Server is normally pretty hard-nosed about rejecting “bad” data, but there’s a lot of FPD already flowing out there.

## Description

The goal is to provide a standard way for the page to supply first party data and control which bidders have access to it.

A number of adapters support taking key/value pairs as arguments, but they're all different. e.g.

- Rubicon takes `keywords`, `inventory` and `visitor` parameters
- AppNexus takes `keywords` and `user`
- OpenX takes `customParams`
- Index takes `firstPartyData`
- etc.

## Proposal

See https://docs.google.com/document/d/1I2eS40yMYJCgz8XvFljf4f59bEPsbWTrUpnD7a7uJEg/edit

bretg:
Re-opening this issue to track two updates to the PBJS issue https://github.com/prebid/Prebid.js/issues/3687:

1) Support for bidder-specific config in the OpenRTB (ext.prebid.bidderconfig)
2) AMP support

SyntaxNode:
@bsardo will be implementing this feature for PBS-Go.

mansinahar:
This was implemented and released in PBS-Go as part of [0.183.0](https://github.com/prebid/prebid-server/releases/tag/v0.183.0)

bretg:
Closing!

PR #1479, opened 2020-09-04
Pass Through First Party Context Data

SyntaxNode:
PBS-Go does not yet support the First Party Data feature (https://github.com/prebid/prebid-server/issues/879), but there is a conflict with how we map bidder data and it's causing issues for some publishers. We'll need to fix this eventually, so why not now?

The First Party Data feature allows for context to be provided on an imp to be shared among all bidders, such as:
```
{
    imp: [
        ...
        ext: {
            context: {
                data: {
                    // everyone sees this data
                    ADUNIT SPECIFIC CONTEXT DATA
                }
            }
         }
    ]
}
```

However, PBS-Go expects all fields within the "ext" to either be bidder data or the prebid section, such as:
```
{
    imp: [
        ...
        ext: {
            appnexus: {
                        placementId: 1
            },
            context: {
                data: {
                    // everyone sees this data
                    ADUNIT SPECIFIC CONTEXT DATA
                }
            }
        }
    ]
}
```

We attempt to validate that context is a valid bidder and return an error. We already have a special case for the prebid section. This PR also adds a special case for the context section, to pass it down to all bid requests.


endpoints/openrtb2/auction_test.go
```diff
@@ -1126,31 +1136,104 @@ func TestDisabledBidder(t *testing.T) {
...
+			requestExt:   json.RawMessage(`{"appnexus":{"placement_id":555},"context":{"data":{"keywords":"prebid server example"}}}`),
+			expectedExt:  `{"appnexus":{"placement_id":555},"context":{"data":{"keywords":"prebid server example"}}}`,
+			expectedErrs: []error{},
+		},
+		{
+			description:  "Valid Prebid Ext Bidder",
+			requestExt:   json.RawMessage(`{"prebid":{"bidder":{"appnexus":{"placement_id":555}}}}`),
+			expectedExt:  `{"prebid":{"bidder":{"appnexus":{"placement_id":555}}}}`,
+			expectedErrs: []error{},
+			// request.ext.prebid.bidder.{biddername} is only promoted/copied to request.ext.{biddername} if there is at least one disabled bidder.
```
SyntaxNode, line 1181: We need to talk about this behavior. It makes no sense to me and looks like a bug. 
- I can understand removing a disabled bidder, but that's not happening for imp.ext.bidder definitions.
- I have no idea why we're copying the bidder up a level here. That hoisting is already performed by the exchange during imp sanitization. Maybe this was added first and never removed?

endpoints/openrtb2/sample-requests/first-party-data/valid-context-allowed.json
```diff
@@ -0,0 +1,32 @@
...
+                }, {
+                    "w": 300,
+                    "h": 600
+                }]
+            },
+            "ext": {
+                "appnexus": {
+                    "placementId": 12883451
+                },
+                "context": {
```
SyntaxNode, line 24: This throws a validation exception currently.

exchange/exchangetest/first-party-data-multiple-bidders.json
```diff
@@ -0,0 +1,173 @@
...
+                            }, {
+                                "w": 300,
+                                "h": 600
+                            }]
+                        },
+                        "ext": {
+                            "bidder": {
+                                "placementId": 1
+                            },
+                            "context": {
```
SyntaxNode, line 60: This is considered a bidder today and causes all sorts of panics currently (when the validation now allows it).

exchange/utils.go
```diff
@@ -273,64 +273,76 @@ func splitImps(imps []openrtb.Imp) (map[string][]openrtb.Imp, []error) {
...
 }
 
-// sanitizedImpCopy returns a copy of imp with its ext filtered so that only "prebid" and bidder params exist.
+// sanitizedImpCopy returns a copy of imp with its ext filtered so that only "prebid", "comtext", and bidder params exist.
 // It will not mutate the input imp.
 // This function will write the new imps to the output map passed in
 func sanitizedImpCopy(imp *openrtb.Imp,
 	bidderExts map[string]json.RawMessage,
 	rawPrebidExt json.RawMessage,
+	firstPartyDataContext json.RawMessage,
```
SyntaxNode, line 309: Yes, this is hacky. This method is in need of a refactor. I wanted to keep my "new feature' hat on for now and limited the scope of this PR from a full on refactor.

exchange/utils.go
```diff
@@ -273,64 +273,76 @@ func splitImps(imps []openrtb.Imp) (map[string][]openrtb.Imp, []error) {
...
-	// in the sanitized imp
 	if err := json.Unmarshal(rawPrebidExt, &prebidExt); err == nil {
-		delete(prebidExt, "bidder")
-
-		var err error
-		if rawPrebidExt, err = json.Marshal(prebidExt); err != nil {
-			errs = append(errs, err)
+		// Remove the entire bidder field. We will already have the content we need in bidderExts. We
+		// don't want to include other demand partners' bidder params in the sanitized imp.
+		if _, hasBidderField := prebidExt["bidder"]; hasBidderField {
```
SyntaxNode, line 318: Well, except for this slight optimization. No need to marshal if there was no bidder section to remove.

endpoints/openrtb2/auction_test.go
SyntaxNode: I thought about this more and now consider it a bug. As a follow-up PR, I'd like to make the validate method only actually validate and move the disabled bidder logic to `exchange.go`, if it's not already there.

endpoints/openrtb2/auction.go
```diff
@@ -820,12 +820,20 @@ func (deps *endpointDeps) validateImpExt(imp *openrtb.Imp, aliases map[string]st
...
 	return errL
 }
 
+func isBidderToValidate(bidder string) bool {
+	// PrebidExtKey is a special case for the prebid config section and is not considered a bidder.
+
+	// ExtRequestFirstPartyDataContext is a special case for the first party data context section
+	// and is not considered a bidder.
+
+	return bidder != openrtb_ext.PrebidExtKey && bidder != openrtb_ext.ExtRequestFirstPartyDataContext
```
hhhjort, line 834: Should it be `RequestFirstPartyDataContextExtKey` to match the `PrebidExtKey` pattern?

SyntaxNode: I'll rename it to `FirstPartyDataContextExtKey`. Push incoming.

hhhjort (dismissed):
It is a bit hacky, but doing better does require a major refactor which is better off in its own PR.

endpoints/openrtb2/auction_test.go
```diff
@@ -1126,31 +1136,104 @@ func TestDisabledBidder(t *testing.T) {
...
-		map[string]string{"unknownbidder": "The bidder 'unknownbidder' has been disabled."},
-		false,
-		[]byte{},
-		openrtb_ext.BidderMap,
-		nil,
-		nil,
-		hardcodedResponseIPValidator{response: true},
+
+	for _, test := range testCases {
+		deps := &endpointDeps{
```
mansinahar, line 1207: I'd suggest moving the `deps` assignment outside the for loop as it is common for all test cases.

openrtb_ext/bidders_test.go
```diff
@@ -61,3 +61,8 @@ func TestBidderListDoesNotDefineGeneral(t *testing.T) {
 	bidders := BidderList()
 	assert.NotContains(t, bidders, BidderNameGeneral)
 }
+
+func TestBidderListDoesNotDefineContext(t *testing.T) {
```
mansinahar, line 65: If the `BidderNameContext` is only being used in this test, can we just define that as a constant here in this specific test file? The reason being it can be otherwise easily confused with `FirstPartyDataContextExtKey` which is what we're actually using when looking at the ext to separate it out from the bidders

endpoints/openrtb2/auction_test.go
SyntaxNode: Moved.

exchange/utils.go
```diff
@@ -273,64 +273,76 @@ func splitImps(imps []openrtb.Imp) (map[string][]openrtb.Imp, []error) {
...
+		if errs := sanitizedImpCopy(&imp, impExt, rawPrebidExt, firstPartyDataContext, &splitImps); errs != nil {
 			errList = append(errList, errs...)
 		}
 	}
 
 	return splitImps, nil
 }
 
-// sanitizedImpCopy returns a copy of imp with its ext filtered so that only "prebid" and bidder params exist.
+// sanitizedImpCopy returns a copy of imp with its ext filtered so that only "prebid", "comtext", and bidder params exist.
```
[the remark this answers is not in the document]
SyntaxNode: I cna't spell. Fixed.

openrtb_ext/bidders_test.go
SyntaxNode: I included `BidderNameContext` alongside the other bidder name definitions to act as documentation of expected cases while also providing a `BidderName` instance (`FirstPartyDataContextExtKey` is a string). I used the same pattern for `BidderNameGeneral` for the same reason.

The variables also refer to slightly different things:
`BidderNameContext` = A reserved bidder name due to a conflict with another imp.ext use case. Will eventually go away when we move entirely to imp.ext.prebid.bidder.xxx

`FirstPartyDataContextExtKey` = Name of the first party data context imp.ext field. Likely to be used more in the future when building out the first party data feature.

mansinahar (dismissed):

hhhjort (approved):

mansinahar (approved):

PR #1481, opened 2020-09-08
Suggestion: vscode remote container support

laurb9:
What does this do
=============

Add support for [Visual Studio Code](https://code.visualstudio.com/) with [In-Container Development](https://code.visualstudio.com/docs/remote/containers).

Why
===

- Makes it easy to start developing prebid-server on any OS that supports Docker, without having to install any other dependencies on the host. Developing in a container isolates the dev environment from the host and is easy to reset or duplicate for checking other PRs etc.

I've added a small VSCODE.md for orientation.

The rest are applicable to on-host vscode development too:
- There is an in-IDE visual debugger
- Individual tests can be run or debug directly from the IDE.
- Style is enforced by the Golang plugin (naming, comments, imports, errors etc). The Golang plugin for VSCode is maintained by the Google folks.

Here is a screenshot showing most of the features in a squeezed window.
Screen Shot 2020-09-03 at 5 43 59 PM

One drawback that I noticed is that if the container is cpu-limited, and it usually is, the tests will need a timeout longer than 10min sometimes to avoid failures.


.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
+{
```
SyntaxNode, line 1: I'm not too keen on adding this file to source control. In my view, this is veering into personal preference territory and I have no interest in trying to find a common ground amongst everyone contributing. My personal `launch.json` looks quite a bit different which I suspect will not suite anyone besides myself.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# From https://github.com/microsoft/vscode-dev-containers/blob/master/containers/go/.devcontainer/Dockerfile
+ARG VARIANT=1
+FROM mcr.microsoft.com/vscode/devcontainers/go:${VARIANT}
+
+# [Optional] Install a version of Node.js using nvm for front end dev
+ARG INSTALL_NODE="true"
```
SyntaxNode, line 6: This repo doesn't have a front end component. Not sure everyone would want to opt for this.

SyntaxNode (commented):
I think this is a cool dev setup. I also think this won't be what everyone wants to work with. How do you feel about adding this primarily as markdown with sample json file contents along with instructions?

.vscode/launch.json
laurb9: I agree, the only reason I left it in was that the default run task tries to run whatever file you're looking at, which always fails unless you go back to main.go.

I'd be curious to see your launch.json :)

.devcontainer/Dockerfile
laurb9: It is actually disabled by the option in devcontainer.json. I don't know why they did it like that, and I preferred to do straight copypasta so updating from the source listed up top would be easier.

laurb9:
> I think this is a cool dev setup. I also think this won't be what everyone wants to work with. How do you feel about adding this primarily as markdown with sample json file contents along with instructions?

Can the sample files be in `.devcontainer` ? :) I can remove the extra .vscode stuff.

I don't mind this not being merged at all, as I wrote it's just a suggestion; but from my experience, spinning up the IDE with dependencies and the right extensions to get to trying out changes, would've been a decent bonus when I started.

.gitignore
```diff
@@ -8,7 +8,6 @@ _obj
 _test
 .cover/
 .idea/
-.vscode/
```
bsardo, line 11: Can we add this back in? I think it would be best if we try to be unopinionated on this subject.

VSCODE.md
```diff
@@ -0,0 +1,28 @@
...
+- VSCode https://code.visualstudio.com/
+   - with remote development extension pack: https://aka.ms/vscode-remote/download/extension
+- (optional) add your github ssh key to agent if you want to be able to access non-public repos
+```sh
+ssh-add ~/.ssh/id_rsa # or your ssh key for github
+```
+
+#### Setup
+
+- start vscode and open repository
```
bsardo, line 18: Perhaps we should add a note here to modify `.devcontainer.json` to configure your container.

bsardo (changes requested):
I agree this is pretty cool and I think it's worth merging as long as it's opt-in.

VSCODE.md
```diff
@@ -0,0 +1,28 @@
+
```
SyntaxNode, line 1: I recommend to rename this file to something along the lines of `devcontainer.md` and link to it from the readme.

SyntaxNode (commented):
We (finally) got around to discussing this PR. We think it's great to offer guidance for devs who want to use this feature. Let's just leave the default vscode setup alone for now.

.vscode/launch.json
SyntaxNode: Nothing too crazy, just very specific.

I have two primary configs to choose between the current file or the current workspace. I find the file config works well with for running tests within a specific file and the workspace config when I need to launch the app - which is honestly rare for me.

I then have many profiles with environment configs to quickly launch with a specific config profile. I find that much easier than swapping out the yaml file.
 
```
{
    "version": "0.2.0",
    "configurations": [{
        "name": "File",
        "type": "go",
        "request": "launch",
        "mode": "auto",
        "program": "${fileDirname}",
        "env": {},
        "args": []
    },
    {
        "name": "Workspace",
        "type": "go",
        "request": "launch",
        "mode": "auto",
        "program": "${workspaceFolder}",
        "env": {},
        "args": []
    }]
}
```

.gitignore
laurb9: I forgot to undo this when I removed the .vscode files from the PR

SyntaxNode (approved):

bsardo (approved):

PR #1483, opened 2020-09-08
Refactor: move getAccount to accounts package (from openrtb2)

laurb9:
Move GetAccount() to a new accounts package, so it can be invoked from other modules.
Leaving the actual config model in config, to avoid changing too much.

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -1,4 +1,4 @@
-package accounts
+package account
 
 import (
 	"context"
--- a/account/account.go
+++ b/account/account.go
@@ -27,8 +27,10 @@ func GetAccount(ctx context.Context, cfg *config.Configuration, fetcher stored_r
 	}
 	if accountJSON, accErrs := fetcher.FetchAccount(ctx, accountID); len(accErrs) > 0 || accountJSON == nil {
 		// accountID does not reference a valid account
-		if len(accErrs) > 0 {
-			errs = append(errs, errs...)
+		for _, e := range accErrs {
+			if _, ok := e.(stored_requests.NotFoundError); !ok {
+				errs = append(errs, e)
+			}
 		}
 		if cfg.AccountRequired && cfg.AccountDefaults.Disabled {
 			errs = append(errs, &errortypes.AcctRequired{
@@ -63,5 +65,5 @@ func GetAccount(ctx context.Context, cfg *config.Configuration, fetcher stored_r
 		})
 		return nil, errs
 	}
-	return
+	return account, nil
 }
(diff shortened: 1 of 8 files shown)


SyntaxNode (commented):
Nice. I like this. Just one nitpick. The preferred naming convention for packages in Go is to use a singular form. In this case, that would be `account` vs `accounts`. I understand this project already deviates quite a bit and this is more of a guideline than a rule. 

In usage, `account.GetAccount` looks a little nicer to me than `accounts.GetAccount`, but just barely.

Thoughts?

laurb9:
> The preferred naming convention for packages in Go is to use a singular form. In this case, that would be `account` vs `accounts`.

Agree, will change it. Almost did it the first time around actually, then I saw all the other package names :)

account/account.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if cfg.AccountRequired && accountID == pbsmetrics.PublisherUnknown {
+		return nil, []error{&errortypes.AcctRequired{
+			Message: fmt.Sprintf("Prebid-server has been configured to discard requests without a valid Account ID. Please reach out to the prebid server host."),
+		}}
+	}
+	if accountJSON, accErrs := fetcher.FetchAccount(ctx, accountID); len(accErrs) > 0 || accountJSON == nil {
+		// accountID does not reference a valid account
+		if len(accErrs) > 0 {
+			errs = append(errs, errs...)
```
SyntaxNode, line 31: Caught this while reviewing. Shouldn't it be `errs = append(errs, accErrs...)`? .. and, you don't even need the `len(accErrs) > 0` check. `append` and the variadic operator handles null just fine.

bsardo:
Looks good except for @SyntaxNode's catch.

laurb9:
Updated to add back eventual fetcher errors.

account/account.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if accountJSON, accErrs := fetcher.FetchAccount(ctx, accountID); len(accErrs) > 0 || accountJSON == nil {
+		// accountID does not reference a valid account
+		for _, e := range accErrs {
+			if _, ok := e.(stored_requests.NotFoundError); !ok {
+				errs = append(errs, e)
+			}
+		}
+		if cfg.AccountRequired && cfg.AccountDefaults.Disabled {
+			errs = append(errs, &errortypes.AcctRequired{
```
SyntaxNode, line 36: If there are `FetchAccount` errors other than `NotFoundError`, then the error message here implies the publisher is sending an invalid account id where the host may be experiencing problems accessing it's accounts store. How do you feel about adding a differently worded message here when `accErrs` > 0? Perhaps something alongs the lines of:

Prebid-server could not verify the Account ID. Please reach out to the prebid server host.

laurb9: Sounds better. Best to avoid misleading messaging.
updated in cc4a174

SyntaxNode (approved):

bsardo (approved):

PR #1484, opened 2020-09-08
Update conversant adapter for new prebid-server interface

johnwier:
Update the conversant adapter to the new interface


adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
 	"github.com/prebid/prebid-server/adapters"
 	"github.com/prebid/prebid-server/errortypes"
-	"github.com/prebid/prebid-server/pbs"
-	"golang.org/x/net/context/ctxhttp"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"net/http"
 )
 
 type ConversantAdapter struct {
 	http *adapters.HTTPAdapter
```
SyntaxNode, line 14: The `http *adapters.HTTPAdapter` field is part of the old interface and should be removed. Doesn't look like you're referencing it anyway.

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
-
-func (a *ConversantAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder *pbs.PBSBidder) (pbs.PBSBidSlice, error) {
-	mediaTypes := []pbs.MediaType{pbs.MEDIA_TYPE_BANNER, pbs.MEDIA_TYPE_VIDEO}
-	cnvrReq, err := adapters.MakeOpenRTBGeneric(req, bidder, a.Name(), mediaTypes)
-
-	if err != nil {
-		return nil, err
+func (c ConversantAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errors []error
+	if len(request.Imp) == 0 {
```
SyntaxNode, line 20: Nice to have. Not strictly needed since PBS "Core" will not call `MakeRequests` if there are no imps. Consider removing the check, but I'm fine if you want to leave it in to be extra defensive.

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
-}
-
-func (a *ConversantAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder *pbs.PBSBidder) (pbs.PBSBidSlice, error) {
-	mediaTypes := []pbs.MediaType{pbs.MEDIA_TYPE_BANNER, pbs.MEDIA_TYPE_VIDEO}
-	cnvrReq, err := adapters.MakeOpenRTBGeneric(req, bidder, a.Name(), mediaTypes)
-
-	if err != nil {
-		return nil, err
+func (c ConversantAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errors []error
```
SyntaxNode, line 19: The `errors` var is unused. Please remove it.

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
-		imp.DisplayManagerVer = "1.0.1"
-		imp.BidFloor = params.BidFloor
-		imp.TagID = params.TagID
-
-		var position *openrtb.AdPosition
-		if params.Position != nil {
-			position = openrtb.AdPosition(*params.Position).Ptr()
+		if i == 0 {
+			if request.Site != nil {
+				request.Site.ID = cnvrExt.SiteID
```
SyntaxNode, line 49: `request.Site.ID` and `request.App.ID` are shared memory and should not be directly changed. Please make a copy first, something like:

```
siteCopy := *request.Site
request.Site = &siteCopy
request.Site.ID = cnvrExt.SiteID
```

... as well as for App.ID and a few other places where I've commented. This is the root cause of the race condition test failures reported by Travis (yay! they work!).

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
-				for _, protocol := range params.Protocols {
-					imp.Video.Protocols = append(imp.Video.Protocols, openrtb.Protocol(protocol))
-				}
-			}
+	var position *openrtb.AdPosition
+	if cnvrExt.Position != nil {
+		position = openrtb.AdPosition(*cnvrExt.Position).Ptr()
+	}
+	if imp.Banner != nil {
+		imp.Banner.Pos = position
```
SyntaxNode, line 92: `imp.Banner` is shared memory. Please make a shallow copy first (see code sample above for .Site).

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
 
-			if len(params.MIMEs) > 0 {
-				imp.Video.MIMEs = make([]string, len(params.MIMEs))
-				copy(imp.Video.MIMEs, params.MIMEs)
-			}
+	} else if imp.Video != nil {
+		imp.Video.Pos = position
```
SyntaxNode, line 95: `imp.Video` is shared memory. Please make a shallow copy first (see code sample above for .Site).

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
-	defer resp.Body.Close()
-
-	if req.IsDebug {
-		debug.StatusCode = resp.StatusCode
-	}
-
-	if resp.StatusCode == 204 {
-		return nil, nil
+func (c ConversantAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	var errs []error
```
SyntaxNode, line 127: The `errs` variable is unused. Please remove it.

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("bad server response: %d. ", err),
+		}}
 	}
 
-	if resp.StatusCode != 200 {
-		return nil, &errortypes.BadServerResponse{
-			Message: fmt.Sprintf("HTTP status: %d, body: %s", resp.StatusCode, string(body)),
-		}
+	if resp.SeatBid == nil {
```
SyntaxNode, line 145: This should be `len(resp.SeatBid) == 0` to protect from safely accessing `resp.SeatBid[0].Bid` below. The SeatBid slice could be non-nil, but still empty.

adapters/conversant/conversant.go
```diff
@@ -1,291 +1,177 @@
...
 		}
 
-		err := json.Unmarshal(unit.Params, &params)
-		if err != nil {
-			return nil, &errortypes.BadInput{
-				Message: err.Error(),
-			}
+		var cnvrExt openrtb_ext.ExtImpConversant
+		if err := json.Unmarshal(bidderExt.Bidder, &cnvrExt); err != nil {
+			return nil, []error{&errortypes.BadInput{
```
SyntaxNode, line 36: Please add a unit test for this branch.

router/router.go
```diff
@@ -156,7 +155,6 @@ func newExchangeMap(cfg *config.Configuration) map[string]adapters.Adapter {
 		"rubicon": rubicon.NewRubiconAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderRubicon)].Endpoint,
 			cfg.Adapters[string(openrtb_ext.BidderRubicon)].XAPI.Username, cfg.Adapters[string(openrtb_ext.BidderRubicon)].XAPI.Password, cfg.Adapters[string(openrtb_ext.BidderRubicon)].XAPI.Tracker),
 		"lifestreet": lifestreet.NewLifestreetAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderLifestreet)].Endpoint),
-		"conversant": conversant.NewConversantAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderConversant)].Endpoint),
```
SyntaxNode, line 159: We're fine with this removal. Just want you to understand that publishers still using the legacy endpoint will now receive an invalid bidder error. If you want to continue with legacy endpoint support, you can still include the old adapter code and tests in legacy.go and legacy_test.go. We would remove those for you when legacy support is retired.

static/bidder-params/conversant.json
```diff
@@ -24,10 +24,6 @@
       "type": "integer",
       "description": "Ad position on screen."
    	},
-   	"mobile": {
```
SyntaxNode, line 27: The docs for adapters are now located in the prebit.github.io repo. The link for this adapter is:
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/conversant.md

The `mobile` param isn't listed there, so I don't think any updates are needed from this PR.

SyntaxNode (approved):
Thank you for quickly addressing the feedback. Please advise on my new comment before we merge.

router/router.go
johnwier: We do want to keep the legacy support for now, so I'm going to add those back in today.

johnwier: I've added the legacy support back in under a separate file, but I still don't really follow how support for both is handled

SyntaxNode: There are different interfaces and methods for the legacy endpoint versus the openrtb endpoint.

The legacy endpoint only works with adapters which specifically implement the legacy interface.

The openrtb endpoint has a wrapper to translate between the legacy model and the openrtb model, so it can use legacy adapters as is. That was meant as a transitionary approach. Now that we're looking to remove the legacy endpoint entirely, all adapters need to specifically implement the openrtb interface and not rely on the underlying conversion.

Does that explain it any clearer?

SyntaxNode (approved):

router/router.go
johnwier: yes, thank you

SyntaxNode (approved):

exchange/adapter_map.go
```diff
@@ -183,7 +184,7 @@ func newAdapterMap(client *http.Client, cfg *config.Configuration, infos adapter
 
 	legacyBidders := map[openrtb_ext.BidderName]adapters.Adapter{
 		// TODO #267: Upgrade the Conversant adapter
-		openrtb_ext.BidderConversant: conversant.NewConversantAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderConversant)].Endpoint),
+		openrtb_ext.BidderConversant: conversant.NewConversantAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[strings.ToLower(string(openrtb_ext.BidderConversant))].Endpoint),
```
SyntaxNode, line 187: Sorry. I didn't notice this before. This needs to be removed. It's replaced by the `openrtb_ext.BidderConversant:` in the adapter map.

SyntaxNode (approved):

guscarreon (approved):
LGTM

PR #1486, opened 2020-09-09
Add metrics to capture stored data fetch all/delta durations and success/error times

bsardo:
This contains the code in #1485 plus metrics. I want to get that one merged in first and then rebase this.


bsardo:
This contains the code in #1485 plus metrics. I want to get that one merged in first and then rebase this.

bsardo:
This feature is now covered by #1515.

PR #1490, opened 2020-09-10
Added new size 640x360 (Id: 198)

hdeodhar:
Added new size 640x360 (Id: 198)


bretg:
@hdeodhar - I don't think we need to do this anymore. Hold please.

harpere (approved):

SyntaxNode:
What happened here? @bretg said to hold and then it was merged in without approval from a PBS-Go maintainer?

bretg:
I suspect there may have been confusion here about which repo this was in -- 99% of the time this kind of request is over in Prebid.js. 

It would be ok to revert this. I really don't think it's needed, especially in Prebid Server. Maybe or maybe not over in Prebid.js

PR #1492, opened 2020-09-11
Fixed TCF2 Geo Only Enforcement

SyntaxNode:
We currently remove the device ID information (ifa, mac hash, device hash) for GDPR when a TCF2 consent string only triggers Geo enforcement but allows IDs to be passed.

### DIFF
--- a/privacy/enforcement.go
+++ b/privacy/enforcement.go
@@ -8,7 +8,6 @@ import (
 type Enforcement struct {
 	CCPA    bool
 	COPPA   bool
-	GDPR    bool
 	GDPRGeo bool
 	GDPRID  bool
 	LMT     bool
@@ -16,7 +15,7 @@ type Enforcement struct {
 
 // Any returns true if at least one privacy policy requires enforcement.
 func (e Enforcement) Any() bool {
-	return e.CCPA || e.COPPA || e.GDPR || e.GDPRGeo || e.GDPRID || e.LMT
+	return e.CCPA || e.COPPA || e.GDPRGeo || e.GDPRID || e.LMT
 }
 
 // Apply cleans personally identifiable information from an OpenRTB bid request.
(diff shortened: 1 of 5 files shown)


privacy/enforcement.go
```diff
@@ -8,15 +8,14 @@ import (
 type Enforcement struct {
 	CCPA    bool
 	COPPA   bool
-	GDPR    bool
```
SyntaxNode, line 11: This bool no longer causes an effect. Removed.

privacy/enforcement.go
```diff
@@ -26,17 +25,33 @@ func (e Enforcement) Apply(bidRequest *openrtb.BidRequest, ampGDPRException bool
...
 	}
 }
 
+func (e Enforcement) getDeviceIDScrubStrategy() ScrubStrategyDeviceID {
+	if e.COPPA || e.GDPRID || e.CCPA || e.LMT {
+		return ScrubStrategyDeviceIDAll
+	}
+
+	return ScrubStrategyDeviceIDNone
+}
```
SyntaxNode, line 39: This was always set to `ScrubStrategyDeviceIDAll` before. Now it resolves to None for GDPR if only Geo is being enforced.

privacy/enforcement.go
```diff
@@ -26,17 +25,33 @@ func (e Enforcement) Apply(bidRequest *openrtb.BidRequest, ampGDPRException bool
...
+
+func (e Enforcement) getIPv4ScrubStrategy() ScrubStrategyIPV4 {
+	if e.COPPA || e.GDPRGeo || e.CCPA || e.LMT {
+		return ScrubStrategyIPV4Lowest8
+	}
+
+	return ScrubStrategyIPV4None
```
SyntaxNode, line 46: This was always set to `ScrubStrategyIPV4Lowest8 ` before.  Now it resolves to None for GDPR if only ID is being enforced. IP rounding is considered Geo enforcement.

privacy/enforcement.go
```diff
@@ -60,12 +75,11 @@ func (e Enforcement) getUserScrubStrategy(ampGDPRException bool) ScrubStrategyUs
 		return ScrubStrategyUserIDAndDemographic
 	}
 
-	if e.GDPR && ampGDPRException {
-		return ScrubStrategyUserNone
+	if e.CCPA || e.LMT {
```
SyntaxNode, line 78: This looks like a bug too? If CCPA is enforced along with GDPR for a AMP request, the user would not be scrubbed. Now the AMP exception applies only to the GDPR enforcement aspect.

privacy/enforcement_test.go
```diff
@@ -19,7 +19,6 @@ func TestAny(t *testing.T) {
 			enforcement: Enforcement{
```
SyntaxNode, line 19: This is the most important part of the PR to review to ensure policy enforcement makes sense.

hhhjort (approved):
A bit if a hack, but as you know this could use a bit of a refactor on the scrub strategy flags.

SyntaxNode:
> A bit if a hack

I don't consider this to be a hack. There may be opportunities to refactor the arguments which can certainly be explored in follow-up PRs.

mansinahar (approved):

PR #1500, opened 2020-09-16
Smarty ads adapter

KushnerykPavel:
Hello, @SyntaxNode ! This is SmartyAds adapter, please check it!


adapters/smartyads/smartyads.go
```diff
@@ -0,0 +1,209 @@
...
+	"text/template"
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/macros"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// Implements Bidder interface.
```
SyntaxNode, line 17: Please follow the Go standard for comments, such as `SmartyAdsAdapter  ...` or remove the comment entirely.

adapters/smartyads/smartyads.go
```diff
@@ -0,0 +1,209 @@
...
+
+func NewSmartyAdsBidder(endpointTemplate string) *SmartyAdsAdapter {
+	template, err := template.New("endpointTemplate").Parse(endpointTemplate)
+	if err != nil {
+		return nil
+	}
+	return &SmartyAdsAdapter{endpoint: *template}
+}
+
+func (a *SmartyAdsAdapter) CheckHasImps(request *openrtb.BidRequest) error {
```
SyntaxNode, line 30: This check isn't necessary. The "core" code will not call an adapter if there are no imps.

adapters/smartyads/smartyads.go
```diff
@@ -0,0 +1,209 @@
...
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	headers.Add("X-Openrtb-Version", "2.5")
+
+	if request.Device != nil {
+		if len(request.Device.UA) > 0 {
+			headers.Add("User-Agent", request.Device.UA)
+		}
+
+		if len(request.Device.IP) > 0 {
```
SyntaxNode, line 51: You may wish to also consider that `request.Device.IPv6` may be set when this field is empty.

adapters/smartyads/smartyads.go
```diff
@@ -0,0 +1,209 @@
...
+	if err := json.Unmarshal(bidderExt.Bidder, &smartyadsExt); err != nil {
+		return nil, &errortypes.BadInput{
+			Message: "ext.bidder not provided",
+		}
+	}
+	return &smartyadsExt, nil
+}
+
+func (a *SmartyAdsAdapter) buildEndpointURL(params *openrtb_ext.ExtSmartyAds) (string, error) {
+	endpointParams := macros.EndpointTemplateParams{Host: params.Host, SourceId: params.SourceId, AccountID: params.AccountID}
```
SyntaxNode, line 133: Adapters may not allow the publisher to specify the entire endpoint domain. I can offer more specific suggestions if you could please give us more context on why a publisher would specify the host.

openrtb_ext/imp_smartyads.go
```diff
@@ -0,0 +1,8 @@
+package openrtb_ext
+
+// ExtSmartyAds defines the contract for bidrequest.imp[i].ext.smartyads
+type ExtSmartyAds struct {
+	AccountID string `json:"accountid"`
+	SourceId  string `json:"sourceid"`
```
SyntaxNode, line 6: Nitpick: SourceId -> SourceID to keep with Go naming conventions.

SyntaxNode (commented):
Thank you for your contribution. Very solid PR with strong test coverage. I have a few comments, the most important being the publisher specifying the host name,

SyntaxNode:
Do you have a docs PR for this adapter? Please see http://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#document-your-adapter for details.

adapters/smartyads/smartyads.go
SmartyAdsSSP: Hello! @SyntaxNode , thanks for you review! Our company has different ns for different regions such as (US_EAST, EU, SGP) . We need to use for example endpoints ns1.example.com for EU and ns2.example.com for SGP. That is main reason when I use host variable.

adapters/smartyads/smartyads.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (a *SmartyAdsAdapter) MakeRequests(
+	openRTBRequest *openrtb.BidRequest,
+	reqInfo *adapters.ExtraRequestInfo,
+) (
+	requestsToBidder []*adapters.RequestData,
+	errs []error,
+) {
+
+	request := *openRTBRequest
```
hhhjort, line 75: The request that is passed to `MakeRequests()` is already a shallow copy of the original request. Additionally the imps in the request are shallow copies of the original imps. So I do not believe you need to make another copy here.

adapters/smartyads/smartyads.go
hhhjort: Should the region not be controlled by the PBS host (as in adapter config) rather than publisher, as the adapter host would control which endpoint is "closest"? Or if the region specific requirement is driven by which advertisers are active on an endpoint (east coat advertisers just want to bid on requests coming to the US_EAST node), then from your example have host = "ns1" and hardcode ".example.com" into your adapter's URL.

The reasoning behind this is we don't want PBS to be used as an open relay. For example, with an openended host setting, someone could have "host": "www.disney.com" and then flood PBS installs with requests which would then be forwarded to www.disney.com to try to overload their website. Given that Disney will see all the requests coming from PBS hosts rather than the original attacker, it will be most difficult or impossible to find the one responsible.

KushnerykPavel: @hhhjort , thanks for your answer ! I will change host to hardcode ".example.com" domain with ns1 as parameter! Thank you for clarification!

SmartyAdsSSP:
Can you help me please with CI errors? I don`t have any errors localy.

SyntaxNode:
> Can you help me please with CI errors? I don`t have any errors localy.

The build failure is:
```
# github.com/prebid/prebid-server/adapters/smartyads [github.com/prebid/prebid-server/adapters/smartyads.test]
adapters/smartyads/usersync_test.go:25:4: unknown field 'Value' in struct literal of type ccpa.Policy
```

I changed the interface on `ccpa.Policy` from `Value` to `Consent` recently. To resolve, please merge in master and rename in your adapter code.

bretg:
Docs PR

adapters/smartyads/smartyadstest/params/race/banner.json
```diff
@@ -0,0 +1,5 @@
+{
+  "host": "smartyadstest.com",
```
hhhjort, line 2: Doesn't really matter since it is just an arbitrary string, but now that the host field is actually the host prefix, would be nice to make the examples match. This way people viewing these tests will not get confused as what is expected in the host field.

SmartyAdsSSP: Hello @hhhjort ! Sure I have fixed it ! Thank you for your attention!

hhhjort (approved):
I think this is looking good now.

adapters/smartyads/smartyads.go
```diff
@@ -0,0 +1,196 @@
...
+
+	responseBody := bidderRawResponse.Body
+	var bidResp openrtb.BidResponse
+	if err := json.Unmarshal(responseBody, &bidResp); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: "Bad Server Response",
+		}}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(bidResp.SeatBid[0].Bid))
```
SyntaxNode, line 171: Is it possible for the `SeatBid` array to be empty? If so, this would cause an index out of range panic.

static/bidder-params/smartyads.json
```diff
@@ -0,0 +1,21 @@
...
+      "type": "string",
+      "description": "Network host to send request"
+    },
+    "sourceid": {
+      "type": "string",
+      "description": "Partner id"
+    },
+    "accountid": {
+      "type": "string",
+      "description": "Partner id"
```
SyntaxNode, line 17: Should the description be account id?

adapters/smartyads/smartyads.go
SmartyAdsSSP: @SyntaxNode Hello! SeatBid can`t be empty! If we send bid response it have one element in SeatBid array. Else we send no bid response .

static/bidder-params/smartyads.json
SmartyAdsSSP: @SyntaxNode, Sure, it more sutable, thank You

adapters/smartyads/smartyads.go
SmartyAdsSSP: @SyntaxNode  I think I need to prevent this situation! Thank you.

SyntaxNode (approved):

SyntaxNode:
@hhhjort is not available today to re-approve. I'm going to merge this in anyway since the changes made since his approval are minimal.

PR #1504, opened 2020-09-17
New krushmedia bid adapter

Krushmedia:
One aditional commit inside colossus bid adapter due to failing test


adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+
+func NewKrushmediaBidder(endpointTemplate string) *KrushmediaAdapter {
+	template, err := template.New("endpointTemplate").Parse(endpointTemplate)
+	if err != nil {
+		return nil
+	}
+	return &KrushmediaAdapter{endpoint: *template}
+}
+
+func (a *KrushmediaAdapter) CheckHasImps(request *openrtb.BidRequest) error {
```
mansinahar, line 29: Is there a reason why this method needs the `KrushmediaAdapter` receiver? It seems to be more of a helper method for `MakeRequests`. Also, any reason to have it as an exported/public method vs a private method?

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+	if len(request.Imp) == 0 {
+		err := &errortypes.BadInput{
+			Message: "Missing Imp Object",
+		}
+		return err
+	}
+	return nil
+}
+
+func GetHeaders(request *openrtb.BidRequest) *http.Header {
```
mansinahar, line 39: This seems like a helper method too. Any reason for it being an exported/public method vs keeping it private?

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+	openRTBRequest *openrtb.BidRequest,
+	reqInfo *adapters.ExtraRequestInfo,
+) (
+	requestsToBidder []*adapters.RequestData,
+	errs []error,
+) {
+
+	request := *openRTBRequest
+
+	if noImps := a.CheckHasImps(&request); noImps != nil {
```
mansinahar, line 76: I'd recommend replacing `noImps` with `err` here as the `CheckHasImps` function returns an error. `noImps` can be misleading as to what the function returns

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+	var bidderExt adapters.ExtImpBidder
+	if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
+		return nil, &errortypes.BadInput{
+			Message: "ext.bidder not provided",
+		}
+	}
+	var krushmediaExt openrtb_ext.ExtKrushmedia
+	if err := json.Unmarshal(bidderExt.Bidder, &krushmediaExt); err != nil {
+		return nil, &errortypes.BadInput{
+			Message: "ext.bidder not provided",
```
mansinahar, line 125: This error is a little misleading as in this case `ext.bidder` would be present but there is a problem unmarshalling its value. Can you please update this error message to be more relevant?

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+		Uri:     url,
+		Headers: *GetHeaders(&request),
+	}}, nil
+}
+
+func (a *KrushmediaAdapter) getImpressionExt(imp *openrtb.Imp) (*openrtb_ext.ExtKrushmedia, error) {
+	var bidderExt adapters.ExtImpBidder
+	if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
+		return nil, &errortypes.BadInput{
+			Message: "ext.bidder not provided",
```
mansinahar, line 119: Error here would not always mean a missing `ext.bidder`. It could also be a marshaling error. Can you please update this error message to be more relevant?

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (a *KrushmediaAdapter) buildEndpointURL(params *openrtb_ext.ExtKrushmedia) (string, error) {
+	endpointParams := macros.EndpointTemplateParams{AccountID: params.AccountID}
+	return macros.ResolveMacros(a.endpoint, endpointParams)
+}
+
+func (a *KrushmediaAdapter) CheckResponseStatusCodes(response *adapters.ResponseData) error {
+	if response.StatusCode == http.StatusNoContent {
+		return &errortypes.BadInput{Message: "No bid response"}
```
mansinahar, line 138: A `BadInput` error type in case of an empty response doesn't seem appropriate here. For example, other adapters in this case return `nil, nil` but if you prefer returning an error instead, I'd recommend using the `BadServerResponse` error type

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	if response.StatusCode == http.StatusBadRequest {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: [ %d ]", response.StatusCode),
+		}
+	}
+
+	if response.StatusCode == http.StatusServiceUnavailable {
```
mansinahar, line 147: Similar comment as above.

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,208 @@
...
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Something went wrong, please contact your Account Manager. Status Code: [ %d ] ", response.StatusCode),
+		}
+	}
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
+		return &errortypes.BadInput{
```
mansinahar, line 154: Similar comment as above. Also, wouldn't it be easier to have this if condition as `if response.StatusCode != http.StatusOK`?

adapters/krushmedia/krushmedia_test.go
```diff
@@ -0,0 +1,11 @@
+package krushmedia
+
+import (
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+)
+
```
mansinahar, line 8: Would recommend adding a simple unit test for `NewKrushmediaBidder` as well

adapters/krushmedia/krushmedia_test.go
```diff
@@ -0,0 +1,11 @@
...
+func TestJsonSamples(t *testing.T) {
+	adapterstest.RunJSONBidderTest(t, "krushmediatest", NewKrushmediaBidder("http://example.com/?c=rtb&m=req&key=2"))
```
mansinahar, line 10: You need to have a macro in here for `AccountID` otherwise it will never be able to take the account ID in the bidder ext and replace it in the URL for the tests. The `RunJSONBidderTest` basically runs all the JSON tests you have defined below.

adapters/krushmedia/krushmediatest/exemplary/banner-app.json
```diff
@@ -0,0 +1,156 @@
...
+          
+          "banner": {
+            "w":320,
+            "h":50
+          },
+          "ext": {
+            "bidder": {
+              "host": "example.com",
+              "sourceid": "partner",
+              "accountid": "hash"
```
mansinahar, line 38: Since the JSON annotation for account id is `key` here: https://github.com/prebid/prebid-server/pull/1504/files#diff-724b5d286e85be13b490de0a88acfd4dR5 you need to use `key` here rather than `accountid`. This wasn't caught as an error because, 1. You are missing the macro in the test endpoint as mentioned above and 2. You need to update the `uri` in the `expectedRequest` below to have the account ID value mentioned in this bidder ext here.

Also, any reason why you have `sourceid` and `host` defined here? From the `ExtKrushmedia` struct it looks like you only support account ID

adapters/krushmedia/krushmediatest/exemplary/banner-app.json
```diff
@@ -0,0 +1,156 @@
...
+              "123.123.123.123"
+            ],
+            "Accept-Language": [
+              "en"
+            ],
+            "Dnt": [
+              "0"
+            ]
+          },
+          "uri": "http://example.com/?c=rtb&m=req&key=2",
```
mansinahar, line 70: As mentioned above, please update this `uri` to have the account ID mentioned in the ext for this mock request

adapters/krushmedia/krushmediatest/exemplary/native-app.json
```diff
@@ -0,0 +1,159 @@
...
+          "id": "some-impression-id",
+          
+          "native": {
+            "ver":"1.1",
+            "request":"{\"adunit\":2,\"assets\":[{\"id\":3,\"img\":{\"h\":120,\"hmin\":0,\"type\":3,\"w\":180,\"wmin\":0},\"required\":1},{\"id\":0,\"required\":1,\"title\":{\"len\":25}},{\"data\":{\"len\":25,\"type\":1},\"id\":4,\"required\":1},{\"data\":{\"len\":140,\"type\":2},\"id\":6,\"required\":1}],\"context\":1,\"layout\":1,\"contextsubtype\":11,\"plcmtcnt\":1,\"plcmttype\":2,\"ver\":\"1.1\",\"ext\":{\"banner\":{\"w\":320,\"h\":50}}}"
+          },
+          "ext": {
+            "bidder": {
+              "host": "example.com",
+              "sourceid": "partner",
```
mansinahar, line 37: Please update the bidder ext and the `uri` below as mentioned above

adapters/krushmedia/krushmediatest/exemplary/banner-app.json
mansinahar: Looks like you have it defined right in most of the tests. Can you please update it for the two tests that don't have it rightly defined and update the test endpoint have a Account ID macro?

mansinahar:
Thanks for the PR! Really like how clean the code is

adapters/krushmedia/krushmediatest/exemplary/native-app.json
Krushmedia: This is invalid param set, should be object with 'key' property, this one left from another adapter used as example

Krushmedia:
@mansinahar Thank's for your deep review, we are new to golang so there is a lot of work to do. I made most of changes, except  those one "Would recommend adding a simple unit test for NewKrushmediaBidder as well", i'm not realy understand how this should be implemented , will be grateful for point me to the right direction

SyntaxNode:
Do you have a docs PR for this adapter? Please see http://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#document-your-adapter for details.

Krushmedia:
@SyntaxNode https://github.com/prebid/prebid.github.io/pull/2360

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,197 @@
...
+			Message: fmt.Sprintf("Unexpected status code: [ %d ]", bidderRawResponse.StatusCode),
+		}}
+	}
+
+	if bidderRawResponse.StatusCode == http.StatusServiceUnavailable {
+		return nil, nil
+	}
+
+	if bidderRawResponse.StatusCode != http.StatusOK {
+		return nil, []error{&errortypes.BadInput{
```
mansinahar, line 159: Please consider returning `BadServerResponse` error here rather than `BadInput` as that can be misleading

mansinahar (commented):
LGTM in general. Just one minor comment

mansinahar:
> @mansinahar Thank's for your deep review, we are new to golang so there is a lot of work to do. I made most of changes, except those one "Would recommend adding a simple unit test for NewKrushmediaBidder as well", i'm not realy understand how this should be implemented , will be grateful for point me to the right direction

So, I was suggesting adding a test function that has two test cases. One that passes a valid endpoint to the `NewKrushmediaBidder` function and other that passes an invalid endpoint but since we do `glog.Fatal` in case there's an error parsing the endpoint, the invalid endpoint test case might need a little more effort which might not be worth it so I will leave this up to you if you want to add it. If not, it's ok.

Thanks for addressing the rest of the comments. Appreciate it!

mansinahar:
@Krushmedia Looks like there are some merge conflicts. Can you please rebase the PR when you have a chance? Thanks!

SyntaxNode:
> rebase

... or just merge in master :). Whatever your level of git familiarity.

Krushmedia:
@mansinahar I've synced with main pbs repository, for now there should be no conflicts

mansinahar (dismissed):

adapters/krushmedia/krushmedia.go
```diff
@@ -0,0 +1,197 @@
...
+	var err error
+
+	for i, imp := range request.Imp {
+		krushmediaExt, err = a.getImpressionExt(&imp)
+		if err != nil {
+			errors = append(errors, err)
+			break
+		}
+		request.Imp[i].Ext = nil
+	}
```
guscarreon, line 91: Under this logic we'll keep and pass to `url, err := a.buildEndpointURL(krushmediaExt)` in line 98 the extension of the last element of the `request.Imp` array. Did we intend this?  What if the Ext's `AccountID` differs from one `Imp.Ext` element to another? 
```
request.Imp { {ID:"imp1",Ext:'{"key":"val"}'}, {ID:"imp2",Ext:'{"key":"differentValue"}'} }
```
If that is not the case and every single `Imp.Ext` element comes with the exact same `AccountID` as ilustrated in the example below:
```
request.Imp { {ID:"imp1",Ext:'{"key":"sameVal"}'}, {ID:"imp2",Ext:'{"key":"sameVal"}'} }
```
Can't we optimize by not calling `a.getImpressionExt(&imp)` on every single one of them? (And rewriting `krushmediaExt` everytime) We could just parse the first element:
```
 66   func (a *KrushmediaAdapter) MakeRequests(
 67       openRTBRequest *openrtb.BidRequest,
 68       reqInfo *adapters.ExtraRequestInfo,
 69   ) (
 70       requestsToBidder []*adapters.RequestData,
 71       errs []error,
 72   ) {
 73   *-- 10 lines: request := *openRTBRequest--------------------
 83  
    +     krushmediaExt, err = a.getImpressionExt(request.Imp[0])
    +     if err != nil {
    +         return nil, append(errors, err)
    +     }
    +
 84       for i, imp := range request.Imp {
 85 -         krushmediaExt, err = a.getImpressionExt(&imp)
 86 -         if err != nil {
 87 -             errors = append(errors, err)
 88 -             break
 89 -         }
 90           request.Imp[i].Ext = nil
 91       }
 92  
 93 -     if len(errors) > 0 {
 94 -         return nil, errors
 95 -     }
 96  
 97       url, err := a.buildEndpointURL(krushmediaExt)
 98       if err != nil {
 99           return nil, []error{err}
100       }
101   *--  6 lines: reqJSON, err := json.Marshal(request)---------
107       return []*adapters.RequestData{{
108           Method:  http.MethodPost,
109           Body:    reqJSON,
110           Uri:     url,
111           Headers: *getHeaders(&request),
112       }}, nil
113   }
adapters/krushmedia/krushmedia.go [+]
```

guscarreon (commented):
Looks pretty solid. Left a few comments

Krushmedia:
@Krushmedia i've adde most of changes but imp length check is left for testing purposes. I moved additional function and made this check inside makeBids when parsing imp extension

guscarreon (approved):
Looks pretty good to me. Thank you @Krushmedia and @Aiholkin for addressing the feedback.

PR #1506, opened 2020-09-18
Bidder Uniqueness Gatekeeping Test

SyntaxNode:


openrtb_ext/bidders_test.go
```diff
@@ -66,3 +66,55 @@ func TestBidderListDoesNotDefineContext(t *testing.T) {
...
+	//   one of them and not use both.
+	var bidders []string
+	for _, bidder := range BidderMap {
+		if bidder != BidderTripleliftNative && bidder != BidderAdkernelAdn {
+			bidders = append(bidders, string(bidder))
+		}
+	}
+
+	currentUniquenessThreshold := 6
+	assert.Equal(t, currentUniquenessThreshold, maxUniqueForPrefixLength(bidders))
```
bsardo, line 85: This test looks good but I'm assuming we aren't anticipating the bidder names changing, is that correct? I think this test would fail if one of the smart* bidder names changed such that there is a difference in the first five characters of their names instead of the difference starting at the sixth character:
```
BidderSmartadserver    BidderName = "smartadserver"
BidderSmartRTB         BidderName = "smartrtb"
```

SyntaxNode: > I'm assuming we aren't anticipating the bidder names changing.

Correct. I'm sure it could happen, but the cost of doing so is high. Such a move would invalidate a lot of stored requests  and publisher setup. Hence, appnexus is still appnexus even though the company is now part of Xandr.

> Perhaps this check should be a less than or equal to?

Sure. That's a good point. This test is meant to sound alarm if the minimum characters needed for uniqueness increases.

bsardo (dismissed):

bsardo (approved):

hhhjort (approved):
LGTM

PR #1514, opened 2020-09-28
Refactor EEAC map to be more in line with the nonstandard publisher map

hhhjort:
Resolves Gus' unresolved comment in #1427


SyntaxNode (approved):

guscarreon (approved):
Looks great.

PR #1515, opened 2020-09-30
Add metrics to capture stored data fetch all/delta durations with fetch status

bsardo:


pbsmetrics/prometheus/prometheus.go
```diff
@@ -387,6 +427,36 @@ func (m *Metrics) RecordRequestTime(labels pbsmetrics.Labels, length time.Durati
 	}
 }
 
+func (m *Metrics) RecordStoredDataFetchTime(labels pbsmetrics.StoredDataTypeLabels, length time.Duration) {
```
bsardo, line 430: One of the labels passed in here is `DataType` but that is just used to figure out which timer to update; it isn't an actual prometheus label. I figured maybe that shouldn't be a label and instead I should just pass a `StoredDataType` parameter to this function and `switch` on that. Thoughts?

hhhjort: Perhaps we should have just one timer and let that be a label as well. I don't think it increases cardinality by breaking the metrics off into the various `DataType`s, and it would allow graphing the overall performance over all the datatypes. What benefit do we get by breaking them into distinct metrics?

bsardo: That's how I originally had it 🙂. @SyntaxNode and I chatted and we figured it would be too many data points for one timer such that it might affect performance. (12 buckets + 1 count + 1 sum) * 5 data types * 2 fetch types * 2 fetch statuses = 280.
@SyntaxNode anything else to add?

stored_requests/events/postgres/database.go
```diff
@@ -65,59 +67,97 @@ func (e *PostgresEventProducer) Invalidations() <-chan events.Invalidation {
...
+func (e *PostgresEventProducer) recordError(errorType pbsmetrics.StoredDataError) {
+	e.cfg.MetricsEngine.RecordStoredDataError(
+		pbsmetrics.StoredDataLabels{
+			DataType: e.metricsStoredDataType(),
+			Error:    errorType,
+		})
+}
+
+func (e *PostgresEventProducer) metricsStoredDataType() pbsmetrics.StoredDataType {
+	return map[config.DataType]pbsmetrics.StoredDataType{
```
mansinahar, line 152: Why not define this map in a private variable global to this file and then use that? The way you currently have it will work fine but then we're defining this map every time that function is called which I don't believe is needed

stored_requests/events/postgres/database_test.go
```diff
@@ -160,24 +174,37 @@ func TestFetchAllErrors(t *testing.T) {
 				AddRow("imp-2", "", "imp").
 				CloseError(errors.New("Some close error.")),
 			wantReturnedError: false,
```
mansinahar, line 176: Looks like this test case doesn't expect an error back (`wantReturnedError: false`) so how come it expects a recorded error?

stored_requests/events/postgres/database_test.go
```diff
@@ -366,6 +404,7 @@ func TestFetchDeltaErrors(t *testing.T) {
 				AddRow("imp-2", "", "imp").
 				CloseError(errors.New("Some close error.")),
 			wantReturnedError:   false,
+			wantRecordedError:   pbsmetrics.StoredDataErrorUndefined,
```
mansinahar, line 407: Same question as above

stored_requests/events/postgres/database_test.go
```diff
@@ -375,17 +414,29 @@ func TestFetchDeltaErrors(t *testing.T) {
...
+		metricsMock.Mock.On("RecordStoredDataFetchTime", pbsmetrics.StoredDataLabels{
+			DataType:      pbsmetrics.RequestDataType,
+			DataFetchType: pbsmetrics.FetchDelta,
+		}, mock.Anything).Return()
+		metricsMock.Mock.On("RecordStoredDataError", pbsmetrics.StoredDataLabels{
+			DataType: pbsmetrics.RequestDataType,
+			Error:    tt.wantRecordedError,
+		}).Return()
+
 		eventProducer := NewPostgresEventProducer(PostgresEventProducerConfig{
```
mansinahar, line 434: You likely need a call here and other places in this file and `config_test` where you are using `MetricsEngineMock ` to one of `metricsMock.Mock.Assert*` calls to make sure your expectation were actually met i.e. the mock calls that you define here were actually made. I know the go doc doesn't have any examples on this so it's easy to miss. Check this example here: https://github.com/stretchr/testify#mock-package

pbsmetrics/prometheus/prometheus.go
bsardo: This was addressed offline. We decided to keep separate timers for each stored data type for performance reasons but to create an additional set of stored data counters for errors.

pbsmetrics/prometheus/preload.go
```diff
@@ -44,6 +46,46 @@ func preloadLabelValues(m *Metrics) {
 		requestTypeLabel: requestTypeValues,
 	})
 
+	preloadLabelValuesForHistogram(m.storedAccountFetchTimer, map[string][]string{
+		storedDataFetchTypeLabel: storedDataFetchTypeValues,
```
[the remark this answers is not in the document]
bsardo: This is how I originally had it; @hhhjort brought up the same point (see https://github.com/prebid/prebid-server/pull/1515#discussion_r499998858). We chatted offline as a group and determined we should keep them separate even with the error counters in the mix now.

stored_requests/events/postgres/database.go
```diff
@@ -65,59 +67,97 @@ func (e *PostgresEventProducer) Invalidations() <-chan events.Invalidation {
...
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	startTime := e.time.Now().UTC()
 	rows, err := e.cfg.DB.QueryContext(ctx, e.cfg.CacheInitQuery)
+	elapsedTime := time.Since(startTime)
 
 	if err != nil {
 		glog.Warningf("Failed to fetch all Stored %s data from the DB: %v", e.cfg.RequestType, err)
+		e.recordError(pbsmetrics.StoredDataErrorNetwork)
```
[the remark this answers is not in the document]
bsardo: Agreed. For now I figure we just have two types of errors: network and undefined. I updated the code to distinguish between these two types of errors when fetching the dataset.

I can open up another PR later that reclassifies the undefined errors as other known error types.

stored_requests/events/postgres/database_test.go
bsardo: 👍 I updated the code to return the row close error caught in the deferred function. I also got rid of `wantReturnedError` and removed the close error test case because it was very hard to test. You can create a `CloseError` using the DataDog SQL mock but the error only bubbles up from the SQL package and mock if `.Close` is called before `.Next` reaches the end of the row set. I spent some time on this but had some trouble; it's not easy to do given the order in which these methods are called in our code. I'm punting on writing tests for that error case given we have solid test coverage (89%) in this module already. I did manually test that the code behaves properly by forcing execution down the close error code path.

stored_requests/events/postgres/database_test.go
bsardo: See response above.

pbsmetrics/prometheus/prometheus_test.go
```diff
@@ -407,6 +407,226 @@ func TestRequestTimeMetric(t *testing.T) {
...
+		giveDataType  pbsmetrics.StoredDataType
+		giveFetchType pbsmetrics.StoredDataFetchType
+		wantCount     uint64
+		wantSum       float64
+	}{
+		{
+			description:   "Update stored account histogram with all label",
+			giveDataType:  pbsmetrics.AccountDataType,
+			giveFetchType: pbsmetrics.FetchAll,
+			wantCount:     1,
```
hhhjort, line 422: Seems like your count and sum are always the same values. Is there a reason they are part of the table structure?

bsardo: Nope, good point, removing.

stored_requests/events/postgres/database_test.go
```diff
@@ -321,81 +344,81 @@ func TestFetchDeltaSuccess(t *testing.T) {
...
 			giveFakeTime:      time.Date(2020, time.July, 1, 12, 30, 0, 0, time.UTC),
+			giveTimeoutMS:     100,
 			giveLastUpdate:    time.Date(2020, time.June, 30, 6, 0, 0, 0, time.UTC),
 			giveMockRows:      nil,
-			wantReturnedError: true,
+			wantRecordedError: pbsmetrics.StoredDataErrorUndefined,
 			wantLastUpdate:    time.Date(2020, time.June, 30, 6, 0, 0, 0, time.UTC),
 		},
 		{
 			description:    "fetch all row error",
```
mansinahar, line 380: Probably wanna update the description to `fetch delta row error`

stored_requests/events/postgres/database_test.go
```diff
@@ -321,81 +344,81 @@ func TestFetchDeltaSuccess(t *testing.T) {
...
+		description       string
+		giveFakeTime      time.Time
+		giveTimeoutMS     int
+		giveLastUpdate    time.Time
+		giveMockRows      *sqlmock.Rows
+		wantRecordedError pbsmetrics.StoredDataError
+		wantLastUpdate    time.Time
 	}{
+		{
+			description:       "fetch all timeout",
```
mansinahar, line 363: Probably wanna update the description to `fetch delta timeout`

hhhjort (approved):

mansinahar (approved):

PR #1518, opened 2020-10-01
Simplifying exchange module: bidResponseExt gets built anyway

guscarreon:
According to the current logic in `exchange/exchange.go`, function `makeExtBidResponse(adapterBids, adapterExtra, bidRequest, debugInfo, errs)` gets called under every scenario either from `HoldAuction()` or from `buildBidResponse()`. This opens the possibility of slightly simplifying these couple of functions. Please let me know what you think of this small refactoring PR.


exchange/exchange.go
```diff
@@ -189,52 +167,79 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
-		if debugLog != nil && debugLog.Enabled {
+	bidResponseExt = e.makeExtBidResponse(adapterBids, adapterExtra, bidRequest, debugInfo, errs)
+	if debugLog != nil && debugLog.Enabled {
+		if bidRespExtBytes, err := json.Marshal(bidResponseExt); err == nil {
+			debugLog.Data.Response = string(bidRespExtBytes)
+		} else {
+			debugLog.Data.Response = "Unable to marshal response ext for debugging"
+			errs = append(errs, err)
+		}
+		if !anyBidsReturned {
```
guscarreon, line 187: I'm still not sure why `debugLog.CacheKey = rawUUID.String()` only happens when `!anyBidsReturned` but I refactored to keep the exact same logic (please let me know if I you think I missed something).

exchange/exchange.go
```diff
@@ -95,6 +95,10 @@ func NewExchange(client *http.Client, cache prebid_cache_client.Client, cfg *con
 }
 
 func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, usersyncs IdFetcher, labels pbsmetrics.Labels, account *config.Account, categoriesFetcher *stored_requests.CategoryFetcher, debugLog *DebugLog) (*openrtb.BidResponse, error) {
+	var auc *auction = nil
```
SyntaxNode, line 98: There is no need to initialize variables to nil. That's the default. I'm also not sure hoisting the definitions improves readability since it removes the locality.

exchange/exchange.go
```diff
@@ -189,52 +167,79 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+		} else if len(geo.Country) == 3 {
+			// The country field is formatted properly as a three character country code
+			usersyncIfAmbiguous = true
+		}
+	}
+
+	return usersyncIfAmbiguous
+}
+
+func (e *exchange) recordImpMetrics(bidRequest *openrtb.BidRequest) {
```
SyntaxNode, line 229: IMHO, a receiver function isn't worth it to pass through the single metrics value. Consider removing the receiver and passing the metrics separately. This should also improve testability.

exchange/exchange.go
```diff
@@ -189,52 +167,79 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 				errs = append(errs, dealErrs...)
 			}
 
-			if debugLog != nil && debugLog.Enabled {
```
SyntaxNode, line 192: There's a lot going on here. Can you please describe why you moves the debug log code?

exchange/exchange.go
guscarreon: Got it. Will move these declarations back downstream where they are used

exchange/exchange.go
guscarreon: Good point. Done

exchange/exchange.go
guscarreon: In the current version of `HoldAuction` this `debugLog != nil && debugLog.Enabled` condition gets evaluated regardless of `anyBidsReturned` being `true` or `false` (lines 192 and 218 of current code). Therefore, it made sense to evaluate this expression regardless of `anyBidsReturned`. This PR is changing the order and calling `bidResponseExt = e.makeExtBidResponse(adapterBids...)` regardless of `debugLog != nil && debugLog.Enabled` because, in the end, even if `debugLog != nil && debugLog.Enabled` was `false`, `bidResponseExt = e.makeExtBidResponse(adapterBids...)` gets called inside `buildBidResponse(..)` anyway (line 503 current code).

I put toghether a truth table in my notes and, even if the **Before** and **After** tables are not identical, I believe the **After** table still makes sense:

**Before**
```
   |                 |                 | debugLog != nil && | if rawUUID, err :=  ||                        | bidRespExtBytes, err := json.Marshal(bidResponseExt) 
   | anyBidsReturned | targData != nil |  debugLog.Enabled  |       uuid.NewV4(); || e.makeExtBidResponse() | (Not called if debugLog != nil && debugLog.Enabled is false)
---|-----------------|-----------------|--------------------|---------------------||------------------------|-------------------------------------------------------------
  1|        F        |  not called     |   line 218    F    |    not called       ||     line 509           |  not called 
  2|        F        |  not called     |   line 218    F    |    not called       ||     line 509           |  not called 
  3|        F        |  not called     |   line 218    T    |    line 219 F       ||     line 509           |  not called (rawUUID, err := uuid.NewV4(); is false)
  4|        F        |  not called     |   line 218    T    |    line 219 T       ||     line 222           |  line 223
  5|        F        |  not called     |   line 218    F    |    not called       ||     line 509           |  not called 
  6|        F        |  not called     |   line 218    F    |    not called       ||     line 509           |  not called 
  7|        F        |  not called     |   line 218    T    |    line 219 F       ||     line 509           |  not called (rawUUID, err := uuid.NewV4(); is false)
  8|        F        |  not called     |   line 218    T    |    line 219 T       ||     line 222           |  line 223
  9|        T        |  line 184   F   |    not called      |    not called       ||     line 509           |  not called  
 10|        T        |  line 184   F   |    not called      |    not called       ||     line 509           |  not called  
 11|        T        |  line 184   F   |    not called      |    not called       ||     line 509           |  not called
 12|        T        |  line 184   F   |    not called      |    not called       ||     line 509           |  not called
 13|        T        |  line 184   T   |   line 192    F    |    not called       ||     line 509           |  not called  
 14|        T        |  line 184   T   |   line 192    F    |    not called       ||     line 509           |  not called  
 15|        T        |  line 184   T   |   line 192    T    |    not called       ||     line 193           |  line 194 
 16|        T        |  line 184   T   |   line 192    T    |    not called       ||     line 193           |  line 194 
```

**After**
```
   |                 | targData != nil | debugLog != nil && | if rawUUID, err :=  ||                        | bidRespExtBytes, err := json.Marshal(bidResponseExt)
   | anyBidsReturned |  IRRELEVANT     |  debugLog.Enabled  |       uuid.NewV4(); || e.makeExtBidResponse() | (Not called if debugLog != nil && debugLog.Enabled is false)
---|-----------------|-----------------|--------------------|---------------------||------------------------|-------------------------------------------------------------
  1|        F        |  not called     |   line 180    F    |    not called       ||     line 179           |   not called 
  2|        F        |  not called     |   line 180    F    |    not called       ||     line 179           |   not called 
  3|        F        |  not called     |   line 180    T    |    line 188 F       ||     line 179           | + line 181 (we simply don't debugLog.CacheKey = rawUUID.String())
  4|        F        |  not called     |   line 180    T    |    line 188 T       ||     line 179           |   line 181
  5|        F        |  not called     |   line 180    F    |    not called       ||     line 179           |   not called 
  6|        F        |  not called     |   line 180    F    |    not called       ||     line 179           |   not called 
  7|        F        |  not called     |   line 180    T    |    line 188 F       ||     line 179           | + line 181 (we simply don't debugLog.CacheKey = rawUUID.String())
  8|        F        |  not called     |   line 180    T    |    line 188 T       ||     line 179           |   line 181
  9|        T        |  line 162   F   | + line 180    F    |    not called       ||     line 179           |   not called  
 10|        T        |  line 162   F   | + line 180    F    |    not called       ||     line 179           |   not called  
 11|        T        |  line 162   F   | + line 180    T    |    not called       ||     line 179           |   line 181
 12|        T        |  line 162   F   | + line 180    T    |    not called       ||     line 179           |   line 181
 13|        T        |  line 162   T   |   line 180    F    |    not called       ||     line 179           |   not called  
 14|        T        |  line 162   T   |   line 180    F    |    not called       ||     line 179           |   not called  
 15|        T        |  line 162   T   |   line 180    T    |    not called       ||     line 179           |   line 181
 16|        T        |  line 162   T   |   line 180    T    |    not called       ||     line 179           |   line 181
```

I don't believe anything gets broken and the fact that all tests are passing, suggests it. The discrepancies in scenarios 9 through 12 of the `debugLog != nil && debugLog.Enabled` column seem to make sense IMHO, as well as rows 3 and 7 of the `bidRespExtBytes, err := json.Marshal(bidResponseExt)` column.

But please let me know your thoughts on this. Do you think this logic should be rearranged to make the tables match?

exchange/exchange.go
```diff
@@ -189,52 +166,79 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+			// The country field is formatted properly as a three character country code
+			usersyncIfAmbiguous = true
+		}
+	}
+
+	return usersyncIfAmbiguous
+}
+
+func recordImpMetrics(bidRequest *openrtb.BidRequest, metricsEngine pbsmetrics.MetricsEngine) {
+
```
SyntaxNode, line 229: Nitpick: Please remove line break at start of method.

exchange/exchange.go
SyntaxNode: Gotcha. Thank you for the background info.

exchange/exchange.go
SyntaxNode: I'm guessing there's an assumption that the CacheKey will set elsewhere if there are bids. There seems to be a lot of code checking if it's length is 0. I'd be afraid to change that assumption in this PR. If you're up for it, a separate PR would be appropriate to fix that concept.

exchange/exchange.go
```diff
@@ -500,21 +505,19 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
 
 	bidResponse.SeatBid = seatBids
 
-	if bidResponseExt == nil {
-		contextDebugValue := ctx.Value(DebugContextKey)
```
SyntaxNode, line 504: Why is the check for the DebugContextKey no longer needed?

exchange/exchange.go
guscarreon: I believe a non-nil `debugLog` object can only come from the `video_auction.go` endpoint and also see that when an error occurs in the video endpoint, a `rawUUID, err := uuid.NewV4();` also gets generated just before exiting execution. My guess is that only when an error happens, or no bids are returned, the UUID to a debug log is returned.

In either case, it'd be better that @camrice or someone from his team takes a look to this refactoring PR.

exchange/exchange.go
```diff
@@ -189,52 +166,79 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
 				errs = append(errs, err)
 			}
 		}
 	}
 
+	// Ensure caching errors are added in case auc.doCache was called and errors were returned
+	if len(cacheErrs) > 0 {
+		bidderCacheErrs := errsToBidderErrors(cacheErrs)
+		bidResponseExt.Errors[openrtb_ext.PrebidExtKey] = append(bidResponseExt.Errors[openrtb_ext.PrebidExtKey], bidderCacheErrs...)
+	}
```
bsardo, line 199: Should this block be between lines 178 and 179 where we call `makeExtBidResponse` and marshal the ext if the debugging is enabled? In that case a cache error would be in the marshaled ext.

guscarreon: I think you have a point. The current version of the code places this block after we make and marshal the `bidResponseExt`, therefore, I respected that order. But it'd probably be a good idea to reconsider this in its own PR.
```
 97 func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidRequest, usersyncs IdFetcher, labels pbsmetrics.Labels, account *config.Account, categoriesFetcher *stored_requests.CategoryFetcher, debugLog *DebugLog) (*openrtb.BidResponse, error) {
 98
 99 *-- 92 lines: requestExt, err := extractBidRequestExt(bidRequest)---------------------------------------------------------------------------
191
192             if debugLog != nil && debugLog.Enabled {
193                 bidResponseExt = e.makeExtBidResponse(adapterBids, adapterExtra, bidRequest, debugInfo, errs)
194                 if bidRespExtBytes, err := json.Marshal(bidResponseExt); err == nil {
195                     debugLog.Data.Response = string(bidRespExtBytes)
196                 } else {
197                     debugLog.Data.Response = "Unable to marshal response ext for debugging"
198                     errs = append(errs, errors.New(debugLog.Data.Response))
199                 }
200             }
201
202             cacheErrs := auc.doCache(ctx, e.cache, targData, bidRequest, 60, &account.CacheTTL, bidCategory, debugLog)
203             if len(cacheErrs) > 0 {
204                 errs = append(errs, cacheErrs...)
205             }
206             targData.setTargeting(auc, bidRequest.App != nil, bidCategory)
207
208             // Ensure caching errors are added if the bid response ext has already been created
209             if bidResponseExt != nil && len(cacheErrs) > 0 {
210                 bidderCacheErrs := errsToBidderErrors(cacheErrs)
211                 bidResponseExt.Errors[openrtb_ext.PrebidExtKey] = append(bidResponseExt.Errors[openrtb_ext.PrebidExtKey], bidderCacheErrs...)
212             }
213         }
214
215     }
216
217 *-- 17 lines: if !anyBidsReturned {---------------------------------------------------------------------------------------------------------
234     // Build the response
235     return e.buildBidResponse(ctx, liveAdapters, adapterBids, bidRequest, adapterExtra, auc, bidResponseExt, cacheInstructions.returnCreative, errs)
236 }
exchange/exchange.go
```
What do you think?

bsardo: A separate PR would be fine but I don't see the change I'm suggesting.

I figured it might be best for the following steps to happen in this order:
1. Make the ext bid response:
`bidResponseExt = e.makeExtBidResponse(adapterBids, adapterExtra, bidRequest, debugInfo, errs)`
2. Set any cache errors on the bid response:
`bidResponseExt.Errors[openrtb_ext.PrebidExtKey] = append(bidResponseExt.Errors[openrtb_ext.PrebidExtKey], bidderCacheErrs...)`
3. Marshal the bid response and then store it in the debug log data response:
```if bidRespExtBytes, err := json.Marshal(bidResponseExt); err == nil {
    debugLog.Data.Response = string(bidRespExtBytes)
````

This way any cache errors that are supposed to be added to the bid response are added to the debug log data response.

exchange/exchange.go
camrice: The cache key creation can be removed here so only the response so far is added to the `debugLog`. Before, I believe I didn't have the cache key created in `PutDebugLogError` or `handleError`, so this case would be missed.

exchange/exchange.go
guscarreon: Corrected. Pretty good catch

bsardo (approved):

SyntaxNode (approved):

bsardo (approved):

PR #1519, opened 2020-10-01
Add Account cache

laurb9:
Summary
=======

Optional cache for Accounts. This cache is not necessary for the current static file fetcher, it is intended to support upcoming http API and db fetchers.

Implementation
===========
Added [cache support for account fetcher](https://github.com/prebid/prebid-server/compare/master...openx:account-cache?expand=1#diff-c705d31b368d04ad30e87727936db388R194-R204). 
Add [new `size_bytes` parameter for `in_memory_cache`](https://github.com/prebid/prebid-server/compare/master...openx:account-cache?expand=1#diff-64d858641d4b8eb33576093f89ca1c21R400-R401) to control single cache.
Add necessary config validation so `accounts` uses `size_bytes` while the rest have `request_cache_size_bytes`
 and `imp_cache_size_bytes`.
Removed some redundant tests and refactored others a little.

Example config:
```yaml
accounts:
  in_memory_cache:
    type: lru
    ttl_seconds: 10
    size_bytes: 500000
```

### DIFF
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -434,8 +434,8 @@ func (cfg *InMemoryCache) validate(dataType DataType, errs configErrors) configE
 			if cfg.Size <= 0 {
 				errs = append(errs, fmt.Errorf("%s: in_memory_cache.size_bytes must be >= 0 when in_memory_cache.type=lru. Got %d", section, cfg.Size))
 			}
-			if cfg.RequestCacheSize > 0 || cfg.ImpCacheSize > 0 { // should this be a warning instead ? "field is ignored"
-				errs = append(errs, fmt.Errorf("%s: in_memory_cache.request_cache_size_bytes and imp_cache_size_bytes do not apply to this section", section))
+			if cfg.RequestCacheSize > 0 || cfg.ImpCacheSize > 0 {
+				glog.Warningf("%s: in_memory_cache.request_cache_size_bytes and imp_cache_size_bytes do not apply to this section and will be ignored", section)
 			}
 		} else {
 			// dual (request and imp) caches
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -445,8 +445,8 @@ func (cfg *InMemoryCache) validate(dataType DataType, errs configErrors) configE
 			if cfg.ImpCacheSize <= 0 {
 				errs = append(errs, fmt.Errorf("%s: in_memory_cache.imp_cache_size_bytes must be >= 0 when in_memory_cache.type=lru. Got %d", section, cfg.ImpCacheSize))
 			}
-			if cfg.Size > 0 { // should this be a warning instead ? "field is ignored"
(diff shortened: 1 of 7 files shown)


stored_requests/fetcher.go
```diff
@@ -190,8 +191,17 @@ func (f *fetcherWithCache) FetchRequests(ctx context.Context, requestIDs []strin
 	return
 }
 
-func (f *fetcherWithCache) FetchAccount(ctx context.Context, accountID string) (json.RawMessage, []error) {
-	return f.fetcher.FetchAccount(ctx, accountID)
+func (f *fetcherWithCache) FetchAccount(ctx context.Context, accountID string) (account json.RawMessage, errs []error) {
+	accountData := f.cache.Accounts.Get(ctx, []string{accountID})
+	// TODO: add metrics
```
bsardo, line 196: You mentioned the metrics will be a separate PR right? And you're also planning to add account HTTP and DB fetchers in their own PRs too?

bsardo (commented):
This is looking good.

stored_requests/fetcher.go
laurb9: The metrics commit is ready, I can make a PR right after this, or add another commit here.
Account http is almost ready, will make a PR separately.
I will not have time this round for the Db fetcher, but I'll try to take a stab at it.

bsardo (dismissed):

config/stored_requests.go
```diff
@@ -384,32 +397,57 @@ type InMemoryCache struct {
...
-			errs = append(errs, fmt.Errorf("%s: in_memory_cache.request_cache_size_bytes must be >= 0 when in_memory_cache.type=lru. Got %d", section, cfg.RequestCacheSize))
-		}
-		if cfg.ImpCacheSize <= 0 {
-			errs = append(errs, fmt.Errorf("%s: in_memory_cache.imp_cache_size_bytes must be >= 0 when in_memory_cache.type=lru. Got %d", section, cfg.ImpCacheSize))
+		if dataType == AccountDataType {
+			// single cache
+			if cfg.Size <= 0 {
+				errs = append(errs, fmt.Errorf("%s: in_memory_cache.size_bytes must be >= 0 when in_memory_cache.type=lru. Got %d", section, cfg.Size))
+			}
+			if cfg.RequestCacheSize > 0 || cfg.ImpCacheSize > 0 { // should this be a warning instead ? "field is ignored"
```
mansinahar, line 437: Yeah, I think this should be fine with being a warning but either way, we should keep this consistent with what we do when `size` is specified in case of stored request/imp cache which right now is a warning

mansinahar (commented):
LGTM other than the minor comment above

config/stored_requests.go
laurb9: I should have changed both at the same time, sorry.

mansinahar: No worries! Happens to all of us. Thank you for the PR :)

bsardo (approved):

PR #1520, opened 2020-10-02
Correct GetCpmStringValue's second return value

guscarreon:
Function `GetCpmStringValue` found in `exchange/price_granularity.go` has been written to return `string` and `error` but currently does not really return a non-`nil` error under any scenario.  This PR adds an error value for `GetCpmStringValue` to return when a large `config.Precision` is found to avoid triggering a Panic. It also limits said value to a maximum of 4. Do you agree? Should we have more than 4 decimal figures in the returned CPM string?

Do you believe we should also return error under any other corner case listed in `TestGetPriceBucketString`?


exchange/price_granularity.go
```diff
@@ -15,21 +16,27 @@ func GetCpmStringValue(cpm float64, config openrtb_ext.PriceGranularity) (string
 	bucketMax := 0.0
 	increment := 0.0
 	precision := config.Precision
-	// calculate max of highest bucket
+
+	// Limit the number of decimal significant figures. Very large values lead to Panics
+	if precision > 4 {
```
SyntaxNode, line 21: This needs review from @camrice and/or @VeronikaSolovei9. Can you please explain how large values lead to panics?

exchange/price_granularity.go
```diff
@@ -15,21 +16,27 @@ func GetCpmStringValue(cpm float64, config openrtb_ext.PriceGranularity) (string
 	bucketMax := 0.0
```
SyntaxNode, line 16: While we're here, consider renaming the method to `GetPriceBucket` for improved clarity.

exchange/price_granularity.go
```diff
@@ -15,21 +16,27 @@ func GetCpmStringValue(cpm float64, config openrtb_ext.PriceGranularity) (string
...
-	// calculate max of highest bucket
+
+	// Limit the number of decimal significant figures. Very large values lead to Panics
+	if precision > 4 {
+		return cpmStr, fmt.Errorf("Limit the number of precision figures to 4. Parsed value: %d", precision)
+	}
+
 	for i := 0; i < len(config.Ranges); i++ {
-		if config.Ranges[i].Max > bucketMax {
+		// calculate max of highest bucket
```
SyntaxNode, line 26: Consider removing this comment. I think the logic combined with the name `bucketMax` is self explanatory.

exchange/price_granularity_test.go
```diff
@@ -28,32 +31,111 @@ func TestGetPriceBucketString(t *testing.T) {
...
+		expectedPriceBucket string
+		expectedError       error
+	}
+	testGroups := []struct {
+		groupDesc string
+		inCpm     float64
+		testCases []aTest
+	}{
+		{
+			"cpm below the max in every price bucket",
```
SyntaxNode, line 47: The values here are not self descriptive enough. Please add field labels.

exchange/price_granularity_test.go
```diff
@@ -28,32 +31,111 @@ func TestGetPriceBucketString(t *testing.T) {
...
-
-func getOnePriceBucket(t *testing.T, name string, granularity openrtb_ext.PriceGranularity, price float64, expected string) {
-	t.Helper()
-	priceBucket, err := GetCpmStringValue(price, granularity)
-	if err != nil {
-		t.Errorf("Granularity: %s :: GetPriceBucketString: %s", name, err.Error())
+	// Define test cases
+	type aTest struct {
+		granularityId       string
+		inGranularity       openrtb_ext.PriceGranularity
```
SyntaxNode, line 37: Rename to `granularity`?

exchange/price_granularity_test.go
```diff
@@ -28,32 +31,111 @@ func TestGetPriceBucketString(t *testing.T) {
...
+		expectedPriceBucket string
+		expectedError       error
+	}
+	testGroups := []struct {
+		groupDesc string
+		inCpm     float64
```
SyntaxNode, line 43: Rename to `cpm`?

exchange/price_granularity_test.go
```diff
@@ -28,32 +31,111 @@ func TestGetPriceBucketString(t *testing.T) {
...
+				{"dense", dense, "5.70", nil},
+				{"custom1", custom1, "5.70", nil},
+			},
+		},
+		{
+			"Precision value corner cases",
+			1.876,
+			[]aTest{
+				{
+					"Negative precision defaults to number of digits already in CPM float",
```
SyntaxNode, line 75: I find the tests hard to reach with a long granularity id. I don't understand the setup.

hhhjort, line 75: Where is the code that deals with a negative precision?

exchange/price_granularity_test.go
```diff
@@ -28,32 +31,111 @@ func TestGetPriceBucketString(t *testing.T) {
...
+				},
+				{
+					"Precision value equals zero, we expect to round up to the nearest integer",
+					openrtb_ext.PriceGranularity{Ranges: []openrtb_ext.GranularityRange{{Max: 5, Increment: 0.05}}},
+					"2",
+					nil,
+				},
+				{
+					"Very large precision value: expect error and an empty string in return",
+					openrtb_ext.PriceGranularity{Precision: int(^uint(0) >> 1), Ranges: []openrtb_ext.GranularityRange{{Max: 5, Increment: 0.05}}},
```
hhhjort, line 88: Do we really want to test with about the largest possible integer when the limit is only 4? I would worry that this could potentially generate an error due to being silly large rather than violating our stated limit of 4. Would probably  generate a different error message, or error out in another way, so would still be flagged, but doesn't make sense as the number doesn't really compare to 4.

exchange/price_granularity.go
VeronikaSolovei9: I tested current upstream implementation with precision = 10 and it works. 
The only thing it looks scary in response: 
```
{
                    "hb_pb": "17.0000000000",
                    "hb_pb_cat_dur": "17.0000000000_food_30s",
                    "hb_cache_id": "37be5f83-4be9-41b4-ab3f-84883205591d"
                }
```

guscarreon: @VeronikaSolovei9 do you think we should should put a limit to `precision`? If so, what should it be?

VeronikaSolovei9: Maybe we should not return error in this case, instead just overwrite precision with maximum reasonable value?

VeronikaSolovei9: @guscarreon 2 and 4 sounds reasonable to me, but please confirm it with product owners as well.

SyntaxNode: In this case, the product owners are the Prebid Server Committee. This is already configurable by the publisher. What's the worst thing that would happen if they enter a value too large? Is this really something we need to code logic for?

VeronikaSolovei9: "hb_pb_cat_dur" is a targeting key for Ad server to select Line Items. Redirect URL also contains it. With wrong precision Ad server set up will not work.

SyntaxNode:
After further discussion, we learned the precision limit is intended to prevent the server from panicking due to a very high publisher supplied value. The check is necessary and will be moved out of this method (so it doesn't need to return an error) and included as part of the other request validation checks. This will not be configurable since the limit will be set to a high value.

exchange/price_granularity_test.go
guscarreon: Function `strconv.FormatFloat(roundedCPM, 'f', precision, 64)` deals with negative precision values. More details in the function definition itself.

exchange/price_granularity_test.go
guscarreon: You are totally right, that was too much. The latest version limits the precision value to 2^15 that both not panics and seems to not overflow the return string. I believe 2^15 = 32768 is more than enough. but let me know your take on this.

exchange/price_granularity.go
```diff
@@ -7,33 +7,32 @@ import (
...
+func GetPriceBucket(cpm float64, config openrtb_ext.PriceGranularity) string {
 	cpmStr := ""
 	bucketMax := 0.0
 	increment := 0.0
 	precision := config.Precision
-	// calculate max of highest bucket
+
 	for i := 0; i < len(config.Ranges); i++ {
-		if config.Ranges[i].Max > bucketMax {
+		if bucketMax < config.Ranges[i].Max {
```
SyntaxNode, line 18: IMHO: I preferred it the other way. It think it reads better as "if this range max is higher than the bucketMax, set bucketMax to it".

exchange/price_granularity_test.go
hhhjort: Are you testing that `strconv.FormatFloat()` is working properly, or that something is handling the case properly when `FormatFloat()` does what it is supposed to do, or that your code is subsequently handling that output correctly? I don't see any code that would react to anything particular about a negative number.

SyntaxNode: https://github.com/prebid/prebid-server/blob/550770752797f12a6336be551e991793ab89c591/openrtb_ext/request.go#L178

exchange/price_granularity.go
guscarreon: Reversed

exchange/price_granularity_test.go
guscarreon: Given that `PriceGranularity` has its own `func (pg *PriceGranularity) UnmarshalJSON(b []byte) error` inside `openrtb_ext/request.go` that sets a bunch of default values if found missing in the Json string, I thought it'd be better to cap the precision value there instead.

VeronikaSolovei9:
I have nothing to add, lgtm

openrtb_ext/request.go
```diff
@@ -192,9 +193,6 @@ func (pg *PriceGranularity) UnmarshalJSON(b []byte) error {
 			}
 			// Enforce that we don't read "min" from the request
 			pgraw.Ranges[i].Min = prevMax
-			if pgraw.Ranges[i].Min < prevMax {
-				return errors.New("Price granularity error: overlapping granularity ranges")
-			}
```
guscarreon, line 194: If we are ignoring `"min"` json field

SyntaxNode: Yeah. That condition will never be met with the `pgraw.Ranges[i].Min = prevMax` line right before it.

hhhjort (dismissed):

exchange/price_granularity_test.go
```diff
@@ -28,32 +31,105 @@ func TestGetPriceBucketString(t *testing.T) {
...
+					openrtb_ext.PriceGranularity{Precision: -1, Ranges: []openrtb_ext.GranularityRange{{Max: 5, Increment: 0.05}}},
+					"1.85",
+				},
+				{
+					"Precision value equals zero, we expect to round up to the nearest integer",
+					openrtb_ext.PriceGranularity{Ranges: []openrtb_ext.GranularityRange{{Max: 5, Increment: 0.05}}},
+					"2",
+				},
+				{
+					"Largest precision value PBS supports (2^15 = 32768)",
```
SyntaxNode, line 84: Do we want to change this to 15 from 32768 to match the other change?

guscarreon: Corrected

hhhjort (approved):

SyntaxNode (approved):

PR #1522, opened 2020-10-06
Rework pubstack module tests to remove race conditions

bsardo:


SyntaxNode:
@gpolaert This is to fix an intermittent test failure we're been seeing across PRs.

gpolaert:
> @gpolaert This is to fix an intermittent test failure we're been seeing across PRs.

Can I help?

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
-	send := func(_ []byte) error {
-		counter++
-		return nil
-	}
-	mockedEvent, err := loadJsonFromFile()
-	if err != nil {
-		t.Fail()
+func TestPubstackModuleErrors(t *testing.T) {
+	tests := []struct {
+		giveRefreshDelay string
```
SyntaxNode, line 60: IMHO the 'give' prefix isn't providing any useful meaning. Please consider removing it.

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+			giveMaxByteSize:  "90MB",
+			giveMaxTime:      "15m",
+		},
+		{
+			giveRefreshDelay: "1h",
+			giveMaxByteSize:  "90z",
+			giveMaxTime:      "15m",
+		},
+		{
+			giveRefreshDelay: "1z",
```
SyntaxNode, line 75: This was "1h" in the original test. Does that matter?

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
-	newFeatures[setUID] = true
-
-	remoteConfig = &Configuration{
-		ScopeID:  "new-scope",
-		Endpoint: "new-endpoint",
-		Features: newFeatures,
+	for _, tt := range tests {
+		server := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+			defer req.Body.Close()
+			data, _ := json.Marshal(&Configuration{})
```
SyntaxNode, line 84: Please never ignore errors. Consider returning a 400 status code if an error is encountered.

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+			res.Write(data)
+		}))
+		client := server.Client()
+
+		defer server.Close()
+
+		_, err := NewPubstackModule(client, "scope", "http://example.com", tt.giveRefreshDelay, 100, tt.giveMaxByteSize, tt.giveMaxTime)
+		assert.NotNil(t, err) // should raise an error since  we can't parse args // configRefreshDelay
```
SyntaxNode, line 92: Why not? I'm guessing because "z" isn't a valid label? Consider adding a comment in the test cases to clarify, or better yet a description per test case. Also, please include the test description in this assert statement so we know which case failed.

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+			if rootCount > 1 {
+				data, _ := json.Marshal(updatedConfig)
+				res.Write(data)
+			} else {
+				data, _ := json.Marshal(origConfig)
+				res.Write(data)
+			}
+		})
+
+		intakeCount := 0
```
SyntaxNode, line 166: The count doesn't seem to be used. Rather, the channel is used to synchronize the code. I think an empty struct channel would work just as well here.

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+		updatedConfig.Endpoint = server.URL
+
+		// instantiate module with 25ms config refresh rate
+		module, err := NewPubstackModule(client, "scope", server.URL, "25ms", 100, "1B", "10ms")
+		assert.Nil(t, err)
+
+		// allow time for the module to load the original config
+		time.Sleep(10 * time.Millisecond)
+
+		pubstack, _ := module.(*PubstackModule)
```
SyntaxNode, line 186: Ok. I think this error ignore is safe. :)

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+		tt.logObject(pubstack)
+
+		eventCount := 0
+
+		select {
+		case eventCount = <-intakeChannel:
+		case <-time.After(10 * time.Millisecond):
+		}
+
+		assert.Equal(t, 0, eventCount)
```
SyntaxNode, line 197: I think this can be written more clearly. Do we expect the channel to be triggered here? If not, please consider something like this:
```
		select {
		case eventCount = <-intakeChannel:
			t.Fail("not expected")
		case <-time.After(10 * time.Millisecond):
		}
```

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,158 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+
+		// allow time for the server to start serving the updated config
+		time.Sleep(10 * time.Millisecond)
+
+		// attempt to log; the event channel should have been created because the feature is enabled in updated config
+		tt.logObject(pubstack)
+
+		select {
```
SyntaxNode, line 205: Similarly, please consider: 
```
		select {
		case eventCount = <-intakeChannel:
		case <-time.After(10 * time.Millisecond):
			t.Fail("not expected")
		}
```

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,165 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+		assert.Nil(t, err, tt.description)
+
+		// allow time for the module to load the original config
+		time.Sleep(10 * time.Millisecond)
+
+		pubstack, _ := module.(*PubstackModule)
+		// attempt to log but no event channel was created because the feature is disabled in the original config
+		tt.logObject(pubstack)
+
+		eventCount := 0
```
SyntaxNode, line 195: The `eventCount` is no longer necessary. The test now just cares about being able (or not being able) to receive from the channel, which is taken care of with the new select cases.

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,165 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+
+		select {
+		case eventCount = <-intakeChannel:
+			assert.Fail(t, "an unexpected event was received", tt.description)
+		case <-time.After(10 * time.Millisecond):
+		}
+
+		assert.Equal(t, 0, eventCount, tt.description)
```
SyntaxNode, line 203: No longer needed.

analytics/pubstack/pubstack_module_test.go
```diff
@@ -57,130 +55,165 @@ func loadJsonFromFile() (*analytics.AuctionObject, error) {
...
+		// attempt to log; the event channel should have been created because the feature is enabled in updated config
+		tt.logObject(pubstack)
+
+		select {
+		case eventCount = <-intakeChannel:
+		case <-time.After(10 * time.Millisecond):
+			assert.Fail(t, "an event was expected but was never received", tt.description)
+		}
+
+		assert.Equal(t, 1, eventCount, tt.description)
```
SyntaxNode, line 217: No longer needed. If you want to still have an assert to make the test look 'proper', you can create a helper method for asserting the channel, ie:

```
assertChanNone(intakeChannel) / assertChanOne(intakeChannel)
```

```
func assertChanNone(t TestingT, c <-chan int, msgAndArgs ...interface{}) bool {
  select {
    case <- c:
      return testify.Fail(t, "Should NOT receive an event, but did", msgAndArgs...)
    case <-time.After(10 * time.Millisecond):
      return true
  }
}

func assertChanOne(t TestingT, c <-chan int, msgAndArgs ...interface{}) bool {
  select {
    case <- c:
      return true
    case <-time.After(10 * time.Millisecond):
      return testify.Fail(t, "Should receive an event, but did NOT", msgAndArgs...)
  }
}
```

bsardo:
@gpolaert Nice work on the pubstack module. We're planning to merge this in today to fix the intermittent test failure. I think there's an opportunity to make some changes that would improve testability so we can get UT coverage of the module itself. I'll share my thoughts shortly in case you're interested.

SyntaxNode (approved):

guscarreon (approved):
As mentioned in a private conversation, maybe in a future PR, the `assertChanNone` function could need revision. Overall the new test looks great

gpolaert:
> @gpolaert Nice work on the pubstack module. We're planning to merge this in today to fix the intermittent test failure. I think there's an opportunity to make some changes that would improve testability so we can get UT coverage of the module itself. I'll share my thoughts shortly in case you're interested.

Definitively! Let me know how can we improve this module :pray: :grin:

bsardo:
@gpolaert I'm not sure if you saw it but I pinged you in the prebid.org slack so we can chat about possible improvements whenever you get a chance. Let me know if that works for you.

Issue #1355, opened 2020-06-12
PreferDeals Flag


bretg:
Exploring the idea of providing more control over how PBS picks the overall winner when it returns ad server targeting. While it's true that PBS always sends the top bid from each bidder in the OpenRTB, some clients can't or don't look at those values, particularly AMP.

Currently, the hb_pb/hb_bidder/hb_adid/etc winning targeting values are determined by the highest bid value. Not the highest value after granularity quantizing, but the highest actual bid. If the targeting isn't in `includebidderkeys` mode, it means that a deal won't get returned unless it has the highest CPM. Which means basically that pubs who've implemented Send-Top-Bids can miss deal bids.

Proposal:
- PBS supports a new request flag: `ext.prebid.targeting.preferdeals: true` to have Prebid Server choose the highest value deal before choosing the highest value non-deal.
- No need for an account setting, since the flag is mostly relevant for AMP and SDK -- storedrequests can be created that have `preferdeals:true`.

bretg:
Approved by PBS committee

hhhjort:
Should this be on the `imp.ext` level? I just realized this is different from all the other targeting parameters that are on the request level.

bretg:
Good Q. Spoke with a publisher on this -- we think that keeping the flag at the request level is fine.

bretg:
Updated the issue to reflect the global, rather than per-imp, nature of this flag.

bretg:
Released with PBS-Java 1.66. Docs updated.

PR #1528, opened 2020-10-08
Adds preferDeals support

hhhjort:
Solves #1355


exchange/auction.go
```diff
@@ -125,6 +125,19 @@ func newAuction(seatBids map[openrtb_ext.BidderName]*pbsOrtbSeatBid, numImps int
 	}
 }
 
+// Calculate if the new bid (nbid) will win against the current winning bid (wbid) given preferDeals.
```
SyntaxNode, line 128: Nitick: Per Go's comment convention, should be something like:
```
isNewWinningBid calculates if the new bid (nbid) will win against the current winning bid (wbid) given preferDeals.
```

exchange/auction.go
```diff
@@ -125,6 +125,19 @@ func newAuction(seatBids map[openrtb_ext.BidderName]*pbsOrtbSeatBid, numImps int
...
+func isNewWinningBid(nbid, wbid *openrtb.Bid, preferDeals bool) bool {
```
SyntaxNode, line 129: Consider changing the variable names to be more obvious:
nbid -> bid (i think the 'new' part is implied)
wbid -> winningBid

exchange/auction_test.go
```diff
@@ -280,6 +280,227 @@ func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec) {
 	}
 }
 
+type testNewAuctionSpec struct {
```
SyntaxNode, line 283: This type has no use outside of the `TestNewAuction` method. Why not use an anonymous struct?
```
testCases := []struct {
	description     string
	seatBids        map[openrtb_ext.BidderName]*pbsOrtbSeatBid
	numImps         int
	preferDeals     bool
	expectedAuction auction,
}{
	{
	...
	},
}
```

SyntaxNode (dismissed):

SyntaxNode (approved):

bsardo (approved):
LGTM

PR #1529, opened 2020-10-08
Emxd 3336 add app video ctv

EMXDigital:
Adding video support to EMX Digital adapter


EMXDigital:
Juts following up here, the checks seem to be failing due to an independent/non-related part of the package `pubstack_module_test`

` FAIL: TestPubstackModuleSuccess (0.17s)
    pubstack_module_test.go:223: 
        	Error Trace:	pubstack_module_test.go:223
        	            				pubstack_module_test.go:205
        	Error:      	Should receive an event, but did NOT
        	Test:       	TestPubstackModuleSuccess
        	Messages:   	setUID events are only published when logging a setUID object with setUID feature on
FAIL`

EMX Digital adapter looks to be okay and has > 90% testing coverage

bsardo:
I'll take a look at that test failure since I recently worked on that test.

EMXDigital:
@bsardo looks like all the checks are passing now 👍

adapters/emx_digital/emx_digital.go
```diff
@@ -137,6 +137,41 @@ func buildImpBanner(imp *openrtb.Imp) error {
...
+	if imp.Video.Protocols != nil {
+		imp.Video.Protocols = cleanProtocol(imp.Video.Protocols)
+	}
+
+	return nil
+}
+
+// not supporting VAST protocol 7 (VAST 4.0);
+func cleanProtocol(protocols []openrtb.Protocol) []openrtb.Protocol {
+	newitems := []openrtb.Protocol{}
```
SyntaxNode, line 164: Consider using make to specify an initial capacity. It's slightly better memory management:
```
newitems := make([]openrtb.Protocol{}, 0, len(protocols))
```

adapters/emx_digital/emx_digital.go
```diff
@@ -137,6 +137,41 @@ func buildImpBanner(imp *openrtb.Imp) error {
 	return nil
 }
 
+func buildImpVideo(imp *openrtb.Imp) error {
+	imp.Ext = nil
```
SyntaxNode, line 141: Just curious why the Ext is removed here.

SyntaxNode (commented):
Looks good. I have a few minor comments. Please review the docs for this adapter to check if any changes are needed: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/emx_digital.md

adapters/emx_digital/emx_digital.go
EMXDigital: Makes sense to us, appreciate the suggestion

adapters/emx_digital/emx_digital.go
EMXDigital: Team had removed it as we were not supporting it for display either - now adapter supports for both display and video for future use

EMXDigital:
@SyntaxNode Updated code based on feedback and discussion with team. Ready for another round

SyntaxNode:
Is the intent to support the `/openrtb2/video` long form video endpoint? If so, please look at the "Long-Form Video Support" section of the [new adapter docs](https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html) to be sure you are setting the necessary fields.

EMXDigital:
@SyntaxNode No, we are not looking to support long form video at this time, just the vanilla video integrations

SyntaxNode:
> @SyntaxNode No, we are not looking to support long form video at this time, just the vanilla video integrations

Gotcha. Sometimes CTV = Long Form Video, so I wanted to ask. Thank you.

adapters/emx_digital/emx_digital.go
```diff
@@ -170,14 +203,22 @@ func preprocess(request *openrtb.BidRequest) []error {
...
 	if request.Site != nil && request.Site.Page != "" {
-		pageURL, err := url.Parse(request.Site.Page)
-		if err == nil && pageURL.Scheme == "https" {
-			secure = int8(1)
+		domain = request.Site.Page
+	} else if request.App != nil {
+		if request.App.Domain != "" {
+			domain = request.App.Domain
+		} else if request.App.StoreURL != "" {
+			domain = request.App.StoreURL
```
SyntaxNode, line 213: Can you please add test coverage for the `StoreURL` condition?

adapters/emx_digital/emx_digital.go
```diff
@@ -238,6 +286,19 @@ func (a *EmxDigitalAdapter) MakeBids(internalRequest *openrtb.BidRequest, extern
...
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Video != nil {
+				bidType = openrtb_ext.BidTypeVideo
+			}
+			break
+		}
+	}
+	return bidType
+}
```
SyntaxNode, line 300: Does your bid response include the bid type? Although the code you have here is common in the project, there's been issues reported (see https://github.com/prebid/prebid-server/issues/862) that this approach isn't always reliable for multiformat impressions. Is there enough information to detect the type based on the bid response?

SyntaxNode: Thank you very much for the fix here.

SyntaxNode (approved):

hhhjort (approved):

EMXDigital:
Thanks @hhhjort for merging this in - will there be an official release coming out in the coming days that we can use to message to Publishers for what version of Prebid S2S supports this new functionality?

hhhjort:
We don't have a set in stone schedule for releases, but there is likely to be a release tomorrow. If not then there should be a release in the middle of next week. The version will be 0.134.0 of course.

Issue #1429, opened 2020-08-04
Proposal: Built In Adapter Aliases


SyntaxNode:
This proposal is to include built-in adapter aliases in Prebid Server.

### What We Have Today
Hosts may currently configure adapter aliases using the default request feature. There are two limitations with this approach:

1. There is no "default" default request in the open source repository. If a publisher wants their alias to be included, they'll need to reach out to each host individually and ask for that configuration. To add further complication, we don't have a list of known hosts to provide to publishers as a starting point.

2. Aliases currently use the exact same code and configuration. If there is a specialized url to use for the alias, there's no way for a host to configure that using a default request.

We've seen a few new adapter requests which copies/pastes the original adapter to form an alias. This isn't ideal since it creates more code to maintain or (perhaps more realistically) to be forgotten about when there are updates made to the "core" adapter. We need to provide a better path forward.

### Out Of Scope
Publishers may configure adapter aliases today within their OpenRTB request for each auction or configure aliases within their stored requests. This behavior will be unaffected by this proposal.

### Proposal
Introduce enhanced adapter alias support such that aliases are part of the code. While it makes sense for aliases to inherit the configuration from the "core" adapter, we should add the ability to override some or all of the config. This would allow aliases to use a different endpoint url, a different user sync cookie, or restrict the supported auction types (app, site, banner, video, etc...).

An adapter alias would be defined in the hardcoded default configuration in `config/config.go` as follows:
```
v.SetDefault("adapters.alias.endpoint", "http://foo.com/openrtb")
v.SetDefault("adapters.alias.aliasOf", "appnexus")
```

This allows the alias to be treated the same as any other adapter from a configuration standpoint. The adapter alias may not be removed by a host (just like a "core" adapter), but may be disabled through the config either using environment variables or a yaml file.

The adapter alias would be present in the `/info` calls almost exactly as default request aliases do today, except that the info may be overwritten with configuration along the lines of:
```
v.SetDefault("adapters.alias.maintainer.email", "foo@bar.com")
v.SetDefault("adapters.alias.maintainer.capabilities.app", "banner, site")
```
Prebid Server will enforce that capabilities from the "core" adapter may be restricted, but not expanded. If the adapter is written to not support `site` requests, as an example, then an alias can't magically decide to change that. However, if a "core" adapter supports `native` ads and the alias doesn't want to for some reason, it has the ability to remove that capability.

[...]

PR #1532, opened 2020-10-10
New Adapter Initialization Framework

SyntaxNode:
Summary
=======
Replaces the current static adapter instantiation with a new dynamic framework. This is the first step in implementing [Built-In Adapter Aliases](https://github.com/prebid/prebid-server/issues/1429).

Doc PR
=======
https://github.com/prebid/prebid.github.io/pull/2495

Implementation
=======
Adapters are currently instantiated and associated to their bidder code through a static mapping in `adapter_map.go#newAdapterMap`. This approach works when the full list of adapters is known at compile time, but that will no longer be true with the [Built-In Adapter Aliases](https://github.com/prebid/prebid-server/issues/1429) feature where aliases with their own config values are allowed to be hardcoded or supplied by the host.

We need to move to a dynamic approach where multiple adapter instances can be created for potentially multiple aliases. To accomplish this, this PR replaces each adapter's unique `NewXXXAdapter` method with a new standardized builder func. 

Old Code:
```
func New33AcrossBidder(endpoint string) *TtxAdapter {
	return &TtxAdapter{
		endpoint: endpoint,
	}
}
```

New Code:
```
// Builder builds a new instance of the 33Across adapter for the given bidder with the given config.
func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
	bidder := &TtxAdapter{
		endpoint: config.Endpoint,
	}
	return bidder, nil
}
```

Adapter builders may now return an error message rather than using the hacky approach of the `MisconfiguredBidder`. Adapter build errors will prevent the application from starting up if the host specifies an invalid config. Although currently unused, the adapter is aware of which bidder code it is building - just in case that has any value down the line. I don't want to change this interface for a while after this.

I've migrated all adapters to the new approach with 100% test coverage of all new `Builder` funcs. I will continue to do so for any new adapter merged in before this work.

New Startup Behavior
=======
I've started to introduce the build loop which will eventually be extended to aliases in a follow-up PR. As a nice side affect, adapters will now only be created if they are set to active by the host. This will finally get rid of the annoying Facebook startup warnings for all of you which don't use it.

Other Changes
=======
- Several adapters were implementing parts of the legacy interface without a legacy adapter. This PR removes those unnecessary elements. 
- Legacy `NewXXXAdapter` names have been standardized.

Required Actions Before Merging
=======
- [x]  Notify all adapter maintainers via email. I consider the adapter code to be owned by each adapter maintainer, but we sometimes needed to make changes to the adapter interface. The changes in this PR are largely inconsequential to the bidding functionality, so I'm sending out an FYI for those who care or want to comment rather than seeking explicit permission.

- [x]  Update docs.prebid.org guide with the changes in this PR.


endpoints/auction.go
```diff
@@ -400,15 +400,6 @@ func setLabelSource(labels *pbsmetrics.Labels, req *pbs.PBSRequest, status *stri
 	}
 }
 
-func getAdapterValue(bidder *pbs.PBSBidder) openrtb_ext.BidderName {
```
SyntaxNode, line 403: Replaced with a simple `openrtb_ext.BidderName(bidder.BidderCode)`, as that's always the end result.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -639,22 +631,18 @@ func TestAmpBadRequests(t *testing.T) {
 		badRequests[strconv.Itoa(100+index)] = readFile(t, "sample-requests/invalid-whole/"+file.Name())
 	}
 
-	// NewMetrics() will create a new go_metrics MetricsEngine, bypassing the need for a crafted configuration set to support it.
-	// As a side effect this gives us some coverage of the go_metrics piece of the metrics engine.
-	theMetrics := pbsmetrics.NewMetrics(metrics.NewRegistry(), openrtb_ext.BidderList(), config.DisabledMetrics{})
```
SyntaxNode, line 644: Moved to a helper method and replaced all other builder code.

endpoints/openrtb2/auction.go
```diff
@@ -384,12 +396,12 @@ func validateAndFillSourceTID(req *openrtb.BidRequest) error {
...
 }
 
-func validateBidAdjustmentFactors(adjustmentFactors map[string]float64, aliases map[string]string) error {
+func (deps *endpointDeps) validateBidAdjustmentFactors(adjustmentFactors map[string]float64, aliases map[string]string) error {
 	for bidderToAdjust, adjustmentFactor := range adjustmentFactors {
 		if adjustmentFactor <= 0 {
 			return fmt.Errorf("request.ext.prebid.bidadjustmentfactors.%s must be a positive number. Got %f", bidderToAdjust, adjustmentFactor)
 		}
-		if _, isBidder := openrtb_ext.BidderMap[bidderToAdjust]; !isBidder {
+		if _, isBidder := deps.bidderMap[bidderToAdjust]; !isBidder {
```
SyntaxNode, line 404: This is a change from checking against all possible bidders to now comparing against all active bidders. I don't think this is a material change. Hopefully. 🤞

endpoints/openrtb2/auction.go
```diff
@@ -910,7 +922,7 @@ func validateUser(user *openrtb.User, aliases map[string]string) error {
 					return errors.New(`request.user.ext.prebid requires a "buyeruids" property with at least one ID defined. If none exist, then request.user.ext.prebid should not be defined.`)
 				}
 				for bidderName := range userExt.Prebid.BuyerUIDs {
-					if _, ok := openrtb_ext.BidderMap[bidderName]; !ok {
+					if _, ok := deps.bidderMap[bidderName]; !ok {
```
SyntaxNode, line 925: This is a change from checking against all possible bidders to now comparing against all active bidders. I don't think this is a material change. Hopefully. 🤞

exchange/adapter_builders.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/prebid/prebid-server/adapters/yieldmo"
+	"github.com/prebid/prebid-server/adapters/yieldone"
+	"github.com/prebid/prebid-server/adapters/zeroclickfraud"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// Adapter registration is kept in this separate file for ease of use and to aid
+// in resolving merge conflicts.
+
+func newAdapterBuilders() map[openrtb_ext.BidderName]adapters.Builder {
```
SyntaxNode, line 94: For simplicity, this is now entirely it's own file.

exchange/adapter_util.go
```diff
@@ -0,0 +1,150 @@
...
+
+			bidders[bidderName] = bidderWithInfoEnforcement
+		}
+	}
+
+	return bidders, errs
+}
+
+func buildExchangeBiddersLegacy(adapterConfig map[string]config.Adapter, infos adapters.BidderInfos) map[openrtb_ext.BidderName]adaptedBidder {
+	bidders := make(map[openrtb_ext.BidderName]adaptedBidder, 3)
```
SyntaxNode, line 93: We're actively working now to add OpenRTB adapters for these 3. Once that happens, this method can be removed and this file's code can be simplified.

exchange/exchange_test.go
```diff
@@ -2585,3 +2629,15 @@ type panicingAdapter struct{}
 func (panicingAdapter) requestBid(ctx context.Context, request *openrtb.BidRequest, name openrtb_ext.BidderName, bidAdjustment float64, conversions currencies.Conversions, reqInfo *adapters.ExtraRequestInfo) (posb *pbsOrtbSeatBid, errs []error) {
 	panic("Panic! Panic! The world is ending!")
 }
+
+func blankAdapterConfig(bidderList []openrtb_ext.BidderName) map[string]config.Adapter {
```
SyntaxNode, line 2633: Now that the exchange accepts a list of bidders directly, we can likely redesign these tests to use fake bidders. That's beyond the scope of this PR.. it's big enough as is.

openrtb_ext/bidders.go
```diff
@@ -60,7 +74,6 @@ const (
 	BidderEmxDigital       BidderName = "emx_digital"
 	BidderEngageBDR        BidderName = "engagebdr"
 	BidderEPlanning        BidderName = "eplanning"
-	BidderFacebook         BidderName = "audienceNetwork"
```
SyntaxNode, line 63: Renamed Facebook to AudienceNetwork for consistency to match its bidder code.

openrtb_ext/bidders.go
```diff
@@ -115,116 +128,130 @@ const (
 	BidderZeroClickFraud   BidderName = "zeroclickfraud"
 )
 
-// BidderMap stores all the valid OpenRTB 2.x Bidders in the project. This map *must not* be mutated.
-// The bidder name 'general' is not allowed since it has special meaning in message maps.
-var BidderMap = map[string]BidderName{
```
SyntaxNode, line 120: This is just a 1:1 mapping. No need to have new adapters add a line here. It's now automatically generated. It's also no longer shares memory, so the `This map *must not* be mutated.` note no longer applies.

openrtb_ext/bidders.go
```diff
@@ -115,116 +128,130 @@ const (
...
-	"verizonmedia":      BidderVerizonMedia,
-	"visx":              BidderVisx,
-	"vrtcal":            BidderVrtcal,
-	"yeahmobi":          BidderYeahmobi,
-	"yieldlab":          BidderYieldlab,
-	"yieldmo":           BidderYieldmo,
-	"yieldone":          BidderYieldone,
-	"zeroclickfraud":    BidderZeroClickFraud,
+// BidderNames returns all core bidders.
+func BidderNames() []BidderName {
```
SyntaxNode, line 132: The new source of truth for bidder names. Matches the model used in the metrics package.

openrtb_ext/bidders.go
```diff
@@ -115,116 +128,130 @@ const (
...
 }
 
-// BidderList returns the values of the BidderMap
-func BidderList() []BidderName {
-	bidders := make([]BidderName, 0, len(BidderMap))
-	for _, value := range BidderMap {
-		bidders = append(bidders, value)
+// BuildBidderNameLookup builds a map of string to BidderName, to remain compatbile with the
+// prebioud BidderMap variable.
+func BuildBidderNameLookup() map[string]BidderName {
```
SyntaxNode, line 225: Builds bidderMap in realtime now. I changed the name to try to make it clearer, but I'm not sure I succeeded. Naming suggestions here?

openrtb_ext/bidders.go
```diff
@@ -115,116 +128,130 @@ const (
...
+	for _, name := range BidderNames() {
+		lookup[string(name)] = name
 	}
-	return bidders
+	return lookup
 }
 
-func (name BidderName) MarshalJSON() ([]byte, error) {
-	return []byte(name), nil
+func BuildBidderNameHashSet() map[string]struct{} {
```
SyntaxNode, line 233: I'm also not convinced this makes sense here. Thoughts?

openrtb_ext/bidders.go
```diff
@@ -115,116 +128,130 @@ const (
...
+		hashSet[string(name)] = struct{}{}
+	}
+	return hashSet
 }
 
-func (name *BidderName) String() string {
-	if name == nil {
-		return ""
+// bidderNameLookup is a map of the lower case version of the bidder name to the precise BidderName value.
+var bidderNameLookup = func() map[string]BidderName {
```
SyntaxNode, line 242: Package private lookup for name normalization, since the adapter config can get an adapter name in lowercase but we're inconsistent and some adapter names are mixed case.

openrtb_ext/bidders.go
```diff
@@ -244,11 +271,13 @@ func NewBidderParamsValidator(schemaDirectory string) (BidderParamValidator, err
 		return nil, fmt.Errorf("Failed to read JSON schemas from directory %s. %v", schemaDirectory, err)
 	}
 
+	bidderMap := BuildBidderNameLookup()
```
SyntaxNode, line 274: Is this too expensive of a call here? To build a map?

router/router.go
```diff
@@ -149,17 +151,17 @@ func loadDataCache(cfg *config.Configuration, db *sql.DB) (err error) {
...
-		"lifestreet": lifestreet.NewLifestreetAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderLifestreet)].Endpoint),
-		"conversant": conversant.NewConversantAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderConversant)].Endpoint),
-		"adform":     adform.NewAdformAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderAdform)].Endpoint),
-		"sovrn":      sovrn.NewSovrnAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderSovrn)].Endpoint),
+		"lifestreet": lifestreet.NewLifestreetLegacyAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderLifestreet)].Endpoint),
+		"conversant": conversant.NewConversantLegacyAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderConversant)].Endpoint),
+		"adform":     adform.NewAdformLegacyBidder(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderAdform)].Endpoint),
+		"sovrn":      sovrn.NewSovrnLegacyAdapter(adapters.DefaultHTTPAdapterConfig, cfg.Adapters[string(openrtb_ext.BidderSovrn)].Endpoint),
```
SyntaxNode, line 164: I normalized the names of the legacy bidders for clarity.

router/router.go
```diff
@@ -228,12 +230,10 @@ func New(cfg *config.Configuration, rateConvertor *currencies.RateConverter) (r
 	}
 
 	p, _ := filepath.Abs(infoDirectory)
-	bidderInfos := adapters.ParseBidderInfos(cfg.Adapters, p, openrtb_ext.BidderList())
+	bidderInfos := adapters.ParseBidderInfos(cfg.Adapters, p, openrtb_ext.BidderNames())
 
-	disabledBidders := map[string]string{
-		"indexExchange": "Bidder \"indexExchange\" has been deprecated and is no longer available. Please use bidder \"ix\" and note that the bidder params have changed.",
```
SyntaxNode, line 234: Moved to `GetDisabledBiddersErrorMessages`. Do we really still need this though? How long ago was the rename?

router/router.go
```diff
@@ -228,12 +230,10 @@ func New(cfg *config.Configuration, rateConvertor *currencies.RateConverter) (r
...
-	}
-	activeBiddersMap := exchange.DisableBidders(bidderInfos, disabledBidders)
+	activeBidders := exchange.GetActiveBidders(bidderInfos)
+	disabledBidders := exchange.GetDisabledBiddersErrorMessages(bidderInfos)
```
SyntaxNode, line 236: `DisableBidders` is separated, since it's doing two entirely separate things which just happen to operator over the same data. We don't care about performance here since this is a run once setup when the server begins.

adapters/33across/33across_test.go
```diff
@@ -4,8 +4,17 @@ import (
...
+	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
 func TestJsonSamples(t *testing.T) {
-	adapterstest.RunJSONBidderTest(t, "33across", New33AcrossBidder("http://ssc.33across.com"))
+	bidder, buildErr := Builder(openrtb_ext.Bidder33Across, config.Adapter{
+		Endpoint: "http://ssc.33across.com"})
+
+	if buildErr != nil {
+		t.Fatalf("Builder returned expected error %v", buildErr)
```
hhhjort, line 16: I believe this should be `"Builder returned unexpected error %v"`, unless you are extremely pessimistic about the quality of adapters in this project. This typo is copy and pasted across all the adapter tests it would appear.

SyntaxNode: LOL. Nice catch. I'll fix the typo.

endpoints/info/bidders.go
```diff
@@ -13,9 +13,10 @@ import (
 
 // NewBiddersEndpoint implements /info/bidders
 func NewBiddersEndpoint(aliases map[string]string) httprouter.Handle {
-	bidderNames := make([]string, 0, len(openrtb_ext.BidderMap)+len(aliases))
-	for bidderName := range openrtb_ext.BidderMap {
-		bidderNames = append(bidderNames, bidderName)
+	bidderNames := make([]string, 0)
```
hhhjort, line 16: Should we do `bidderNames := make([]string, 0, len(openrtb_ext.CoreBidderNames()) + len(aliases))`? Perhaps also save the list from `CoreBidderNames()` so we don't have to call it twice?

endpoints/info/bidders_test.go
```diff
@@ -48,20 +48,21 @@ func testGetBidders(t *testing.T, aliases map[string]string) {
 	assert.Equal(t, http.StatusOK, r.Code, "GET /info/bidders returned bad status: %d", r.Code)
 	assert.Equal(t, "application/json", r.Header().Get("Content-Type"), "Bad /info/bidders content type. Expected application/json. Got %s", r.Header().Get("Content-Type"))
 
+	bidderMap := openrtb_ext.BuildBidderNameHashSet()
+
 	bodyBytes := r.Body.Bytes()
-	bidderSlice := make([]string, 0, len(openrtb_ext.BidderMap)+len(aliases))
+	bidderSlice := make([]string, 0)
```
hhhjort, line 54: Is there a reason you are zeroing out the capacity?

exchange/adapter_util.go
```diff
@@ -0,0 +1,150 @@
...
+		bidders[name] = addValidatedBidderMiddleware(bidder)
+	}
+}
+
+// GetActiveBidders returns a map of all active bidder names.
+func GetActiveBidders(infos adapters.BidderInfos) map[string]openrtb_ext.BidderName {
+	activeBidders := make(map[string]openrtb_ext.BidderName)
+
+	for name, info := range infos {
+		if info.Status != adapters.StatusDisabled {
```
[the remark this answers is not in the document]
hhhjort: Using `StatusDisabled` comes from wanting the default to be an adapter is active. An uninitialized boolean in go defaults to false, so going this way makes the default stance active if nothing is touched. I am open to the argument we should go the other way now given the number of adapters that now exist.

endpoints/info/bidders_test.go
SyntaxNode: It's a test. I didn't think it was worth an extra call to `CoreBidderNames` to measure the size.

endpoints/info/bidders.go
SyntaxNode: I could. This part of the code only runs once on startup. I wasn't sure if avoiding a few small memory reallocations was worth it. If you feel strongly about this, I can refactor.

hhhjort: I am not too worried, but we do make a call to `CoreBidderNames()` below so we could reuse the call. I would not have said anything except that the old code had the optimization, so seems a step backwards.

openrtb_ext/bidders.go
```diff
@@ -23,28 +23,42 @@ const BidderNameGeneral = BidderName("general")
 // BidderNameContext is reserved for first party data.
 const BidderNameContext = BidderName("context")
 
-// These names _must_ coincide with the bidder code in Prebid.js, if an adapter also exists in that project.
-// Please keep these (and the BidderMap) alphabetized to minimize merge conflicts among adapter submissions.
-// The bidder name 'general' is not allowed since it has special meaning in message maps.
+func (name BidderName) MarshalJSON() ([]byte, error) {
```
hhhjort, line 26: Why do we need these methods? Does this `MarshalJSON()` do anything different from the default?

router/router.go
hhhjort: I think I am okay with dropping this, it has been at least a year or two now.

openrtb_ext/bidders.go
SyntaxNode: Not sure. This was moved up from further down in the file. I just wanted to keep the `BidderName` receiver funcs together.

exchange/adapter_util.go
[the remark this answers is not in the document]
SyntaxNode: I wasn't sure if there was a specific intent behind this decision. I'm fine leaving precedent alone and keeping with active being 'not disabled'.  I don't want to risk introducing a breaking change.

hhhjort (approved):
LGTM

bsardo (dismissed):

hhhjort (dismissed):

lunamedia:
LGTM

TheMediaGrid (approved):

fhartwig:
Looks good to me, as far as the Pubnative adapter is concerned.

adapters/conversant/conversant_test.go
```diff
@@ -1,839 +1,19 @@
...
+)
 
-func assertNotEqual(t *testing.T, actual interface{}, expected interface{}, msg string) {
-	if expected == actual {
-		msg = fmt.Sprintf("%s: act(%v) == exp(%v)", msg, actual, expected)
-		t.Fatal(msg)
-	}
-}
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderAdtelligent, config.Adapter{})
```
johnwier, line 12: Should be BidderConversant, not BidderAdtelligent

hhhjort (approved):

bsardo (approved):

vchimishuk (approved):

PR #1535, opened 2020-10-12
Add client/AccountID support into Adoppler adapter.

vchimishuk:
Add new `client` parameter support.

**Question:** on adapters page Adoppler adapter section lists features which Adoppler suppors. `SChain Support` and `USP/CCPA Support` marked as `no`. Is it related to Prebid.js only? Otherwise, it would be great if we can mark them as `yes`, Adoppler bidder supports both of them.

[1] http://docs.prebid.org/dev-docs/pbs-bidders.html#adoppler

Doc PR: https://github.com/prebid/prebid.github.io/pull/2422


guscarreon (approved):
LGTM

adapters/adoppler/adoppler.go
```diff
@@ -28,11 +32,13 @@ type adsImpExt struct {
...
 type AdopplerAdapter struct {
-	endpoint string
+	endpoint *template.Template
 }
 
 func NewAdopplerBidder(endpoint string) *AdopplerAdapter {
-	return &AdopplerAdapter{endpoint}
+	e := endpoint + "/processHeaderBid/{{.AdUnit}}"
+
+	return &AdopplerAdapter{template.Must(template.New("endpoint").Parse(e))}
```
SyntaxNode, line 41: Please don't use `Must` in parsing the endpoint. It's better to log a descriptive message instead of causing a panic.  Here's an example from the adhese adapter:
```
func NewAdheseBidder(uri string) *AdheseAdapter {
	template, err := template.New("endpointTemplate").Parse(uri)
	if err != nil {
		glog.Fatal("Unable to parse endpoint url template")
		return nil
	}
	return &AdheseAdapter{endpointTemplate: *template}
}
```

adapters/adoppler/adoppler.go
```diff
@@ -65,8 +71,11 @@ func (ads *AdopplerAdapter) MakeRequests(
 			continue
 		}
 
-		uri := fmt.Sprintf("%s/processHeaderBid/%s",
-			ads.endpoint, url.PathEscape(ext.AdUnit))
+		uri, err := ads.bidUri(ext)
+		if err != nil {
+			errs = append(errs, &errortypes.BadInput{err.Error()})
```
SyntaxNode, line 76: Consider adding a prefix to the error to clearly delineate this as a url error.

adapters/adoppler/adoppler.go
```diff
@@ -172,6 +181,18 @@ func (ads *AdopplerAdapter) MakeBids(
 	return adsResp, nil
 }
 
+func (ads *AdopplerAdapter) bidUri(ext *openrtb_ext.ExtImpAdoppler) (string, error) {
+	params := macros.EndpointTemplateParams{}
+	params.AdUnit = url.PathEscape(ext.AdUnit)
+	if ext.Client == "" {
+		params.AccountID = DefaultClient
+	} else {
+		params.AccountID = ext.Client
```
SyntaxNode, line 190: Is this freeform text? If so, please consider running it through a UrlEscape call.

adapters/adoppler/adoppler_test.go
```diff
@@ -7,6 +7,6 @@ import (
 )
 
 func TestJsonSamples(t *testing.T) {
-	bidder := NewAdopplerBidder("http://adoppler.com")
+	bidder := NewAdopplerBidder("http://{{.AccountID}}.trustedmarketplace.com")
```
SyntaxNode, line 10: I'm a little sad to see the switch to account specific domains. This will greatly decrease the performance of your adapter on the server as the connection pool will become starved sooner. Is this really necessary?

adapters/adoppler/adopplertest/exemplary/custom-client.json
```diff
@@ -0,0 +1,15 @@
+{"mockBidRequest": {"id": "req1",
```
SyntaxNode, line 1: The json files are hard to read. Please consider auto formatting them.

openrtb_ext/imp_adoppler.go
```diff
@@ -1,5 +1,6 @@
 package openrtb_ext
 
 type ExtImpAdoppler struct {
+	Client string `json:"client"`
```
SyntaxNode, line 4: Please open a PR on the docs repository to add the new client parameter to this doc: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/adoppler.md

adapters/adoppler/adoppler_test.go
vchimishuk: I'm afraid so. There is more like a client/tenant ID. There are no many clients (only a few of them). I just reused account ID to  avoid new variable introducing.

openrtb_ext/imp_adoppler.go
vchimishuk: Openned: https://github.com/prebid/prebid.github.io/pull/2422

vchimishuk:
@SyntaxNode, addressed your comments. 
Would you like to help me also with the question in the PR description, please?

guscarreon (approved):
LGTM

adapters/adoppler/adoppler_test.go
SyntaxNode: :(

Ah well. That's fine on the repurposing of `AccountID`.

adapters/adoppler/adoppler.go
```diff
@@ -28,11 +33,18 @@ type adsImpExt struct {
 }
 
 type AdopplerAdapter struct {
-	endpoint string
+	endpoint *template.Template
 }
 
 func NewAdopplerBidder(endpoint string) *AdopplerAdapter {
-	return &AdopplerAdapter{endpoint}
+	e := endpoint + "/processHeaderBid/{{.AdUnit}}"
```
SyntaxNode, line 40: Why not include the `/processHeaderBid/{{.AdUnit}}` as part of the endpoint config? It'll give a bit more flexibility for future changes:
```
http://{{.AccountID}}.trustedmarketplace.io/ads/processHeaderBid/{{.AdUnit}}
```

adapters/adoppler/adopplertest/exemplary/multibid.json
```diff
@@ -1,60 +1,50 @@
...
-                                     "type": "banner"}]},
-                          {"currency": "USD",
-                           "bids": [{"bid": {"id": "req1-imp2-bid1",
-                                             "impid": "imp2",
-                                             "price": 0.24,
-                                             "adm": "<VAST />",
-                                             "cat": ["IAB1", "IAB2"],
-                                             "ext": {"ads": {"video": {"duration": 121}}}},
-                                     "type": "video"}]}]}
+{
```
SyntaxNode, line 1: Thank you for formatting the json tests. Much easier to read now. Hence this question.. what part of this test case is "multi"? This looks like a no-bid response.

adapters/adoppler/adoppler.go
vchimishuk: The thing is there is other "functions" except `processHeaderBid` and it can be a situation when need to use another on also in the future.

adapters/adoppler/adopplertest/exemplary/multibid.json
vchimishuk: Good catch, thanks. I've accidentally replaced this file with wrong content during formatting. It verifies multi-impression scenario. Will fix and rename it to `multiimp.json` to.

adapters/adoppler/adoppler.go
SyntaxNode: Not sure I follow. Are you saying the adapter might to call out to multiple paths in the future based on some properties of the bid request?

vchimishuk: Yes. I mean there are no such visible plans for that. Just in case. I thought about endpoint in config something more like "base URI". 
I can move `AdUnit` placeholder in config if we need, it is not big difference for me.

SyntaxNode: Gotcha. If it's not a big deal to you, then yes - please move the `AdUnit` bit to the config. This is more in line with the other adapters and gives the host a better idea of the bidder url. If you need to change it in the future as the adapter evolves, no big deal. Let's try to keep it simpler for now.

Looks great otherwise.

vchimishuk: Added commit with discussed changes.

SyntaxNode (dismissed):

adapters/adoppler/adopplertest/exemplary/no-bid.json
```diff
@@ -1,13 +1,50 @@
...
+              "id":"imp1",
+              "banner":{
+                "w":100,
+                "h":200
+              },
+              "ext":{
+                "bidder":{
+                  "adunit":"unit1"
+                }
+              }
```
mansinahar, line 36: Test cases that are testing non 200 response are expected to be added to the `supplemental` directory. Please consider moving this one there

adapters/adoppler/adopplertest/exemplary/custom-client.json
```diff
@@ -0,0 +1,52 @@
...
+        }
+      },
+      "mockResponse":{
+        "status":204,
+        "body":""
+      }
+    }
+  ],
+  "expectedBidResponses":[
+
```
mansinahar, line 50: I'd suggest adding responses for this and the next test too because that exercises the `MakeBids` code of your adapter.

SyntaxNode:
Friendly Ping @vchimishuk 
There are comments from the second reviewer which you haven't addressed.

vchimishuk:
@SyntaxNode , @mansinahar, I've addressed last review comments. 
Sorry for delay.

mansinahar (approved):

SyntaxNode (approved):

PR #1537, opened 2020-10-13
Acuity ads adapter

AcuityAdsIntegrations:
Hello! Please check new AcuityAds prebid adapter


adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,193 @@
...
+	requestsToBidder []*adapters.RequestData,
+	errs []error,
+) {
+
+	var errors []error
+	var acuityAdsExt *openrtb_ext.ExtAcuityAds
+	var err error
+
+	for i, imp := range openRTBRequest.Imp {
+		acuityAdsExt, err = a.getImpressionExt(&imp)
```
guscarreon, line 64: It seems that we are rewritting `acuityAdsExt *openrtb_ext.ExtAcuityAds` over and over again ignoring the first n-1 `ext` values.
```
[]Imp={ {...Ext:`values`}, {...Ext:`otherValues`}, {...Ext:`lastExtValues`} }
   ignored--------^--------------------^
```
Unless every `Imp` element in the array come with the exact same `ext` to which we could call `a.getImpressionExt(&imp)` only once:
```
51   func (a *AcuityAdsAdapter) MakeRequests(
52       openRTBRequest *openrtb.BidRequest,
53       reqInfo *adapters.ExtraRequestInfo,
54   ) (
55       requestsToBidder []*adapters.RequestData,
56       errs []error,
57   ) {
58  
59       var errors []error
60 -     var acuityAdsExt *openrtb_ext.ExtAcuityAds
61 -     var err error
   +     acuityAdsExt, err := a.getImpressionExt(&openRTBRequest.Imp[0])
   +     if err != nil {
   +         errors = append(errors, err)
   +     }
62  
63       for i, imp := range openRTBRequest.Imp {
64 -         acuityAdsExt, err = a.getImpressionExt(&imp)
65 -         if err != nil {
66 -             errors = append(errors, err)
67 -             break
68 -         }
69           openRTBRequest.Imp[i].Ext = nil
70       }
71  
72 -     if len(errors) > 0 {
73 -         return nil, errors
74 -     }
75       
76       url, err := a.buildEndpointURL(acuityAdsExt)
77       if err != nil {
78           return nil, []error{err}
79       }
adapters/acuityads/acuityads.go
```

adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,193 @@
...
+
+func NewAcuityAdsBidder(endpointTemplate string) *AcuityAdsAdapter {
+	template, err := template.New("endpointTemplate").Parse(endpointTemplate)
+	if err != nil {
+		return nil
+	}
+	return &AcuityAdsAdapter{endpoint: *template}
+}
+
+func GetHeaders(request *openrtb.BidRequest) *http.Header {
```
guscarreon, line 28: Nitpick: Do we need this function to be exported? If it won't be used elsewhere, maybe we could `getHeaders(request *openrtb.BidRequest) *http.Header` instead

adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+
+	if response.StatusCode == http.StatusServiceUnavailable {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Something went wrong, please contact your Account Manager. Status Code: [ %d ] ", response.StatusCode),
+		}
+	}
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
```
guscarreon, line 132: Just double checking: function `CheckResponseStatusCodes` doesn't seem to consider status codes other than: 
 - `StatusNoContent            = 204 // RFC 7231, 6.3.5`
 - `StatusBadRequest           = 400 // RFC 7231, 6.5.1`
 - `StatusServiceUnavailable   = 503 // RFC 7231, 6.6.4`
 - `StatusCreated              = 201 // RFC 7231, 6.3.2`
 - `StatusAccepted             = 202 // RFC 7231, 6.3.3`
 - `StatusNonAuthoritativeInfo = 203 // RFC 7231, 6.3.4`
 - `StatusResetContent         = 205 // RFC 7231, 6.3.6`
 - `StatusPartialContent       = 206 // RFC 7233, 4.1`
 - `StatusMultiStatus          = 207 // RFC 4918, 11.1`
 - `StatusAlreadyReported      = 208 // RFC 5842, 7.1`
 - `StatusIMUsed               = 226 // RFC 3229, 10.4.1`
 - `StatusMultipleChoices      = 300 // RFC 7231, 6.4.1`

Other adapters do:
```
			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
		}}
	}
```
To cover everything else. Would that be useful/desired in our case?

adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return &acuityAdsExt, nil
+}
+
+func (a *AcuityAdsAdapter) buildEndpointURL(params *openrtb_ext.ExtAcuityAds) (string, error) {
+	endpointParams := macros.EndpointTemplateParams{Host: params.Host, AccountID: params.AccountID}
+	return macros.ResolveMacros(a.endpoint, endpointParams)
+}
+
+func (a *AcuityAdsAdapter) CheckResponseStatusCodes(response *adapters.ResponseData) error {
```
guscarreon, line 115: Function's `CheckResponseStatusCodes` capitalization implies it is getting exported to use by outside packages, do we want/need this?

adapters/acuityads/acuityadstest/exemplary/banner-app.json
```diff
@@ -0,0 +1,151 @@
...
+          "tagid": "ogTAGID",
+          "banner": {
+            "w":320,
+            "h":50
+          },
+          "ext": {
+            "bidder": {
+              "host": "ep1",
+              "sourceid": "partner",
+              "accountid": "hash"
```
guscarreon, line 38: Is `sourceId` an `ext` field that may be of use to your adapter? It doesn't hurt to leave it there since it'll be ignored when unmarshaling `acuityAdsExt` inside `getImpressionExt(imp *openrtb.Imp)`. Just double-checking if this is not a field we want to add to the parameter list.

guscarreon (changes requested):
Overall looks pretty good. Just a few comments before approving

adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+type AcuityAdsAdapter struct {
+	endpoint template.Template
+}
+
+func NewAcuityAdsBidder(endpointTemplate string) *AcuityAdsAdapter {
+	template, err := template.New("endpointTemplate").Parse(endpointTemplate)
+	if err != nil {
+		return nil
```
SyntaxNode, line 23: Please add a glog call to let the host know if there was an error. This will be improved in the near future.

adapters/acuityads/acuityads.go
SyntaxNode: You should also just return `http.Header` since the result is immediately dereferenced anyway.

adapters/acuityads/usersync.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"text/template"
+
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/usersync"
+)
+
+func NewAcuityAdsSyncer(temp *template.Template) usersync.Usersyncer {
+	return adapters.NewSyncer("acuityads", 0, temp, adapters.SyncTypeRedirect)
```
SyntaxNode, line 11: Isn't AcuityAds assigned GVL ID 231?

config/config.go
```diff
@@ -917,6 +918,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	// for them and specify all the parameters they need for them to work correctly.
 	v.SetDefault("adapters.33across.endpoint", "http://ssc.33across.com/api/v1/hb")
 	v.SetDefault("adapters.33across.partner_id", "")
+	v.SetDefault("adapters.acuityads.endpoint", "http://{{.Host}}.admanmedia.com/bid?token={{.AccountID}}")
```
SyntaxNode, line 921: Why does AcuityAds require a per-request sub domain?

This seems to be a growing trend among adapters, which is worrying because it will reduce server performance with connection pool starving. This will impact the amount of traffic AcuityAds will receive from PBS hosts.

SyntaxNode (commented):
Thank you for your contribution. You'll also need to open a docs PR to be merged in. The instructions are detailed here: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html

Please link the PR here when you have it opened.

SyntaxNode:
The URL is pointing to AdMan. Is that a copy/paste error or is this an alias?

config/config.go
AcuityAdsIntegrations: Hello, @SyntaxNode ! Thanks for your comments! Main reason for this step is that we have some different geographical located datacenters in order to reduce latency for our partners. Currently we work for implement geo-dns for our project. I`ll send pull request as soon as we switch our current logic.

AcuityAdsIntegrations:
> The URL is pointing to AdMan. Is that a copy/paste error or is this an alias?

This is alias. We use same url with AdMan but for different projects.

AcuityAdsIntegrations:
Hello @SyntaxNode and @guscarreon ! Thank you four your review , it makes my adapter more cleaner and compact. I have made changes, which you talk about.

SyntaxNode:
> This is alias. We use same url with AdMan but for different projects.

Interesting. We generally recommend the use of a host defined default request alias instead of creating a new adapter and we're also building out new built-in alias support. The adapter parameters are different than the adman though.

Can you please provide some more context on why the adman adapter can't be aliased instead? Other than the geo specific domains.

KushnerykPavel:
> > This is alias. We use same url with AdMan but for different projects.
> 
> Interesting. We generally recommend the use of a host defined default request alias instead of creating a new adapter and we're also building out new built-in alias support. The adapter parameters are different than the adman though.
> 
> Can you please provide some more context on why the adman adapter can't be aliased instead? Other than the geo specific domains.

Sure! Even though that Acuity and Adman have similar host names there are two different project. Acuity ads can`t support ZoneID prebid parameter. We work with different partners, and have different business logic under the hood.

AcuityAdsIntegrations:
> > This is alias. We use same url with AdMan but for different projects.
> 
> Interesting. We generally recommend the use of a host defined default request alias instead of creating a new adapter and we're also building out new built-in alias support. The adapter parameters are different than the adman though.
> 
> Can you please provide some more context on why the adman adapter can't be aliased instead? Other than the geo specific domains.

Sure! AcuityAds and Adman are different project which has different partnersm business logic and teams which support it. Acuity cant support ZoneId which provide Adman , so there is main issue for reuse Adman Prebid adapter.

AcuityAdsIntegrations:
Hell @SyntaxNode , @Gaudeamus ! I have made documentation for adapter. Did you have chance for review my changes?

hhhjort:
> > Interesting. We generally recommend the use of a host defined default request alias instead of creating a new adapter and we're also building out new built-in alias support. The adapter parameters are different than the adman though.
> > Can you please provide some more context on why the adman adapter can't be aliased instead? Other than the geo specific domains.
> 
> Sure! AcuityAds and Adman are different project which has different partnersm business logic and teams which support it. Acuity cant support ZoneId which provide Adman , so there is main issue for reuse Adman Prebid adapter.

That seems like a sufficient reason to have 2 adapters. A heads up though, #1532 is the first step in a refactor aimed at providing more flexibility in this area, along with "hardcoding" aliases into the system. The new adapter API does feed the name of the current adapter or alias into the new adapter call, so the adapter can change some behavior depending on which alias it is. What may still need work to support your case is to have aliases able to have different bidder infos from the core bidder. You may want to keep an eye on that work if in the future rolling the two adapters together may reduce maintenance work for your adapter(s).

hhhjort (dismissed):

adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (a *AcuityAdsAdapter) buildEndpointURL(params *openrtb_ext.ExtAcuityAds) (string, error) {
+	endpointParams := macros.EndpointTemplateParams{Host: params.Host, AccountID: params.AccountID}
+	return macros.ResolveMacros(a.endpoint, endpointParams)
+}
+
+func (a *AcuityAdsAdapter) checkResponseStatusCodes(response *adapters.ResponseData) error {
+	if response.StatusCode == http.StatusNoContent {
+		return &errortypes.BadInput{Message: "No bid response"}
```
SyntaxNode, line 110: Please consider returning `return nil` here and refactor `MakeBids` to exit early instead of raising an error. Otherwise, this lack of a bid from this adapter will be considered an error and that's not standard in Prebid Server.

Unless you always expect a response from the server, even if there are no bids from AcuityAds?

adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,193 @@
...
+		return nil, &errortypes.BadInput{
+			Message: "ext.bidder not provided",
+		}
+	}
+	return &acuityAdsExt, nil
+}
+
+func (a *AcuityAdsAdapter) buildEndpointURL(params *openrtb_ext.ExtAcuityAds) (string, error) {
+	endpointParams := macros.EndpointTemplateParams{Host: params.Host, AccountID: params.AccountID}
+	return macros.ResolveMacros(a.endpoint, endpointParams)
```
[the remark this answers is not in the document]
SyntaxNode: I think the validation is provided by the 'core' based on the bidder-params json file. Sound about right, @hhhjort ?

adapters/acuityads/acuityads.go
```diff
@@ -0,0 +1,186 @@
...
+		if imp.ID == impId {
+			if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			} else if imp.Native != nil {
+				mediaType = openrtb_ext.BidTypeNative
+			}
+			return mediaType
+		}
+	}
+	return mediaType
```
SyntaxNode, line 185: Does your bid response include the bid type? Although the code you have here is common in the project, there's been issues reported (see https://github.com/prebid/prebid-server/issues/862) that this approach isn't always reliable for multiformat impressions. Is there enough information to detect the type based on the bid response?

SyntaxNode (commented):
Thank you for the quick updates. This is looking great. I just have a few last comments and a reminder to create and link here to your doc PR, per https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#document-your-adapter

SyntaxNode:
> @guscarreon was concerned with the unvalidated host and accountid params. I think the validation is provided by the 'core' based on the bidder-params json file. Sound about right, @hhhjort ?

Yes. We verified this morning. The checks in the bidder-param json schema is sufficient.

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2433

adapters/acuityads/acuityads.go
AcuityAdsIntegrations: @SyntaxNode  Thank, you ! I have fixed it!

adapters/acuityads/acuityads.go
AcuityAdsIntegrations: @SyntaxNode Yes, currently its enough.

hhhjort (approved):
LGTM

SyntaxNode (approved):

PR #1539, opened 2020-10-13
Add Support For imp.ext.prebid For DealTiers

SyntaxNode:
The code which implements DealTiers (largely for long form video auctions) doesn't check the newly preferred `imp.ext.prebid.bidder.{bidder}` section. This fixes that.

### DIFF
--- a/exchange/exchange_test.go
+++ b/exchange/exchange_test.go
@@ -2166,119 +2166,91 @@ func TestApplyDealSupport(t *testing.T) {
 
 func TestGetDealTiers(t *testing.T) {
 	testCases := []struct {
-		impExt       json.RawMessage
-		bidderResult map[string]bool // true indicates bidder had valid config, false indicates invalid
+		description string
+		request     openrtb.BidRequest
+		expected    map[string]openrtb_ext.DealTierBidderMap
 	}{
 		{
-			impExt: json.RawMessage(`{"validbase": {"dealTier": {"minDealTier": 5, "prefix": "tier"}, "placementId": 10433394}}`),
-			bidderResult: map[string]bool{
-				"validbase": true,
+			description: "None",
+			request: openrtb.BidRequest{
+				Imp: []openrtb.Imp{},
 			},
+			expected: map[string]openrtb_ext.DealTierBidderMap{},
 		},
 		{
-			impExt: json.RawMessage(`{"validmultiple1": {"dealTier": {"minDealTier": 5, "prefix": "tier"}, "placementId": 10433394}, "validmultiple2": {"dealTier": {"minDealTier": 5, "prefix": "tier"}, "placementId": 10433394}}`),
-			bidderResult: map[string]bool{
-				"validmultiple1": true,
-				"validmultiple2": true,
-			},
-		},
-		{
-			impExt: json.RawMessage(`{"nodealtier": {"placementId": 10433394}}`),
-			bidderResult: map[string]bool{
-				"nodealtier": false,
-			},
-		},
-		{
-			impExt: json.RawMessage(`{"validbase": {"placementId": 10433394}, "onedealTier2": {"dealTier": {"minDealTier": 5, "prefix": "tier"}, "placementId": 10433394}}`),
-			bidderResult: map[string]bool{
-				"onedealTier2": true,
-				"validbase":    false,
-			},
-		},
-	}
-
-	filledDealTier := DealTier{
-		Info: &DealTierInfo{
-			Prefix:      "tier",
-			MinDealTier: 5,
-		},
-	}
-	emptyDealTier := DealTier{}
-
-	for _, test := range testCases {
-		bidRequest := &openrtb.BidRequest{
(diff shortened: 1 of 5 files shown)


openrtb_ext/imp.go
```diff
@@ -4,29 +4,15 @@ import (
...
 // ExtImpPrebid defines the contract for bidrequest.imp[i].ext.prebid
 type ExtImpPrebid struct {
+	// StoredRequest specifies which stored impression to use, if any.
 	StoredRequest *ExtStoredRequest `json:"storedrequest"`
 
-	// Rewarded inventory signal, can be 0 or 1
+	// IsRewardedInventory is a signal intended for video impressions. Must be 0 or 1.
 	IsRewardedInventory int8 `json:"is_rewarded_inventory"`
 
-	// NOTE: This is not part of the official API, we are not expecting clients
```
SyntaxNode, line 26: It's now part of the official API. Comment removed.

SyntaxNode:
FYI: @camrice since this is code your originally wrote.

camrice:
Thank you for the notice! The changes look great. No comments from me.

openrtb_ext/deal_tier.go
```diff
@@ -0,0 +1,68 @@
...
+	for bidder, param := range impPrebidExt.Prebid.Bidders {
+		if param.DealTier != nil {
+			dealTiers[BidderName(bidder)] = normalizeDealTier(param.DealTier)
+		}
+	}
+
+	return dealTiers, nil
+}
+
+func normalizeDealTier(dealTier *DealTier) DealTier {
```
hhhjort, line 64: Was there a spec or an identified use case where normalization would be needed?

openrtb_ext/imp.go
```diff
@@ -4,29 +4,15 @@ import (
 	"encoding/json"
 )
 
-// ExtImp defines the contract for bidrequest.imp[i].ext
```
hhhjort, line 7: We no longer have any references to `ExtImp` to extract `Prebid`? A little surprised, but I guess that is good.

openrtb_ext/deal_tier.go
camrice: I don't think there is a spec that this is needed. It was originally added as some basic additional protection against bad configurations. As we are adding updates to the docs, we could either add a note about what makes a proper prefix, or perhaps this function could be removed if we don't want to enforce this.

openrtb_ext/imp.go
SyntaxNode: Correct. This looks like it was used before we moved to dynamic unmarshalling of the ext to find bidder info.

openrtb_ext/deal_tier.go
SyntaxNode: I prefer to remove the logic and update the docs. In general, PBS does not historically protect the publisher from bad config by silently trying to fix it. We like to fail loud and fail fast.

Sounds good @hhhjort @camrice ?

exchange/exchange_test.go
```diff
@@ -2278,127 +2278,99 @@ func TestApplyDealSupport(t *testing.T) {
...
+				"imp1": {openrtb_ext.BidderAppnexus: {Prefix: "tier1", MinDealTier: 5}},
+				"imp2": {openrtb_ext.BidderAppnexus: {Prefix: "tier2", MinDealTier: 8}},
 			},
 		},
 		{
-			impExt: json.RawMessage(`{"validbase": {"placementId": 10433394}, "onedealTier2": {"dealTier": {"minDealTier": 5, "prefix": "tier"}, "placementId": 10433394}}`),
-			bidderResult: map[string]bool{
-				"onedealTier2": true,
-				"validbase":    false,
+			description: "Many - Skips Malformed",
```
bsardo, line 2317: Maybe add one more test case where there are multiple bidders for a single imp?

openrtb_ext/deal_tier.go
hhhjort: If it is a bad configuration that would never work, the whole request should be rejected, so the publisher will fix the config immediately rather than limping along in a suboptimal or unintended configuration. If the error is such that a config change on the PBS host determines if the config is valid, then we just warn so that a PBS config update does not break existing configs that were working.

bsardo (approved):

openrtb_ext/deal_tier.go
camrice: I agree with removing this for this PR and replacing with a warning if possible. That can be a future enhancement as well.

SyntaxNode: Sounds good. I'll remove it. We don't have a good way of sending back warnings at the moment. I'll create an internal ticket for it.

exchange/exchange_test.go
SyntaxNode: Sure.

bsardo (approved):

hhhjort (approved):

PR #1542, opened 2020-10-14
Yieldmo app support in yaml file

Winston-Yieldmo:
add app and video support in static/bidder-info/yieldmo.yaml

### DIFF
--- a/adapters/yieldmo/yieldmo.go
+++ b/adapters/yieldmo/yieldmo.go
@@ -138,11 +138,11 @@ func NewYieldmoBidder(endpoint string) *YieldmoAdapter {
 }
 
 func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
-	//default to banner unless video exists in impression
+	//default to video unless banner exists in impression
 	for _, imp := range imps {
-		if imp.ID == impId && imp.Video != nil {
-			return openrtb_ext.BidTypeVideo
+		if imp.ID == impId && imp.Banner != nil {
+			return openrtb_ext.BidTypeBanner
 		}
 	}
-	return openrtb_ext.BidTypeBanner
+	return openrtb_ext.BidTypeVideo
 }
--- a/adapters/yieldmo/yieldmo.go
+++ b/adapters/yieldmo/yieldmo.go
@@ -123,7 +123,7 @@ func (a *YieldmoAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalR
 		for i := range sb.Bid {
 			bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
 				Bid:     &sb.Bid[i],
-				BidType: "banner",
+				BidType: getMediaTypeForImp(sb.Bid[i].ImpID, internalRequest.Imp),
 			})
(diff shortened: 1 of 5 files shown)


mansinahar (approved):

static/bidder-info/yieldmo.yaml
```diff
@@ -1,6 +1,11 @@
...
   email: "prebid@yieldmo.com"
 capabilities:
+  app:
+    mediaTypes:
+    - banner
+    - video
   site:
     mediaTypes:
     - banner
+    - video
```
SyntaxNode, line 11: The yeildmo adapter is hardcoded to support banner types only. You'll need to add auction type detection and update the bid response type.

```
bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
	Bid:     &sb.Bid[i],
	BidType: "banner",
})
```

Please also add json test cases for the new types (site video, app banner, and app video).

Winston-Yieldmo: Thank you for pointing this out. My mistake. I have made the necessary changes, and added the json test cases you recommended.

adapters/yieldmo/yieldmo.go
```diff
@@ -136,3 +136,13 @@ func NewYieldmoBidder(endpoint string) *YieldmoAdapter {
 		endpoint: endpoint,
 	}
 }
+
+func getMediaTypeForImp(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	//default to banner unless video exists in impression
+	for _, imp := range imps {
```
SyntaxNode, line 142: Does the Yeildmo bid response include the bid type? Although the code you have here is common in the project, there's been issues reported (see https://github.com/prebid/prebid-server/issues/862) that this approach isn't always reliable for multiformat impressions. Is there enough information to detect the type based on the bid response?

static/bidder-info/yieldmo.yaml
SyntaxNode: Great. Thank you. Looking good now.

adapters/yieldmo/yieldmo.go
Winston-Yieldmo: This is a good point. We unfortunately do not currently have anything in our responses that would be a good signal for determining the media type.

Our bidder doesn't support multi format requests at this time. I don't see any evidence on our end that we are receiving requests with multi format impressions, but if it were to happen, you are correct that the implementation I used would've resulted in that bug. In the event that we received both, we would've only considered the banner request, but the code in the adapter would've marked it as video, because video was present in the request.

I have flipped the logic in the adapter, so that banner is always used if banner is present. This mirrors the behavior on our end, and will prevent us from incorrectly marking bids. I recognize that in the long term, this is still a fragile way to doing this, so I have added a ticket on our end to include the media type in an extension object of the bid response. Once we have gotten that through our release process, I can make another update to our adapter to use that field instead. Will this work, for now?

SyntaxNode: Perfect. Thank you.

SyntaxNode (approved):

bsardo (approved):

Issue #1395, opened 2020-07-14
Account configuration and validation


laurb9:
## Account Configuration

We should have account configuration settings available in Prebid Server.

In addition to validating that an account (publisher id) is known to the prebid-server, account-specific configurations would be needed to support planned functionality. The Java version has the following per-account configurations, for example:

- eventsEnabled: to control win notification events such as #1202 or video impression tracking #1015
- enforceGdpr
- bannerCacheTtl
- videoCacheTtl
- analyticsSamplingFactor

PBS-Go has a simple yaml accounts list in legacy code, not used for openrtb2 endpoints and with the current definition as a list of strings, only useful for validation. PBS-Java supports a yaml file with a more complex structure, and a database source.

### Implementation

For implementation, supporting the same data sources as PBS-Java could be useful to some extent and the ability to refresh the data in flight is needed as well. The HTTP api source is well suited for this purpose, even more versatile than the database in my opinion. FileFetcher less so (I will create a ticket for separate discussion on this).

I'm willing to take a stab at implementing this. Thoughts ?

bretg:
Thanks @laurb9 - I imagine @SyntaxNode would happily take the help here.

However, expanding the schema presents its own problems - another enhancement under way in PBS-Java is to make the account query configurable.

Unlike the Stored Request queries, the Account query in PBS-Java (and I think PBS-Go too) is currently hard-coded. This isn't ideal because:

1) it causes upgrade problems between versions
2) doesn't allow host companies to define their own schemas as easily
3) it's inconsistent with the other query definitions

So what we're building:
- a new configuration parameter will be added "settings.database.account-query" for specifying SQL query that loads account data. No default value, i.e. host company must provide it in its deployment.
- certain restrictions are imposed on the SQL query and described in documentation (order, number and data types of returned attributes, placeholder for account ID in a WHERE clause, a LIMIT clause etc.)
- there will be one macro in the query which is %ACCOUNTID%

e.g.
```
settings.database.account-query =SELECT uuid, price_granularity, banner_cache_ttl, video_cache_ttl, events_enabled, enforce_ccpa, tcf_config, analytics_sampling_factor, truncate_target_attr FROM accounts_account where uuid = %ACCOUNTID% LIMIT 1
```

SyntaxNode:
I completely agree. The simple account list was "good enough" for a while, but now that we're expanding per-account settings for privacy (GDPR, CCPA, etc..) we need a much improved solution.

[...]

PR #1543, opened 2020-10-14
Add metrics for account cache

laurb9:
Follow-up on #1519, issue #1395 

Testing:
At server start:
```
$ curl -s http://localhost:8001/ | egrep ^account
account_cache_performance{cache_result="hit"} 0
account_cache_performance{cache_result="miss"} 0
```

After first request (publisher.id=valid_acct)
```
$ curl -s http://localhost:8001/ | egrep ^account
account_cache_performance{cache_result="hit"} 0
account_cache_performance{cache_result="miss"} 1
account_requests{account="valid_acct"} 1
```

After next 2 requests
```
$ curl -s http://localhost:8001/ | egrep ^account
account_cache_performance{cache_result="hit"} 2
account_cache_performance{cache_result="miss"} 1
account_requests{account="valid_acct"} 3
```

### DIFF
--- a/pbsmetrics/metrics.go
+++ b/pbsmetrics/metrics.go
@@ -368,6 +368,7 @@ type MetricsEngine interface {
 	RecordUserIDSet(userLabels UserLabels) // Function should verify bidder values
 	RecordStoredReqCacheResult(cacheResult CacheResult, inc int)
 	RecordStoredImpCacheResult(cacheResult CacheResult, inc int)
+	RecordAccountCacheResult(cacheResult CacheResult, inc int)
 	RecordStoredDataFetchTime(labels StoredDataLabels, length time.Duration)
 	RecordStoredDataError(labels StoredDataLabels)
 	RecordPrebidCacheRequestTime(success bool, length time.Duration)
--- a/pbsmetrics/prometheus/prometheus_test.go
+++ b/pbsmetrics/prometheus/prometheus_test.go
@@ -1013,8 +1013,8 @@ func TestStoredReqCacheResultMetric(t *testing.T) {
 func TestStoredImpCacheResultMetric(t *testing.T) {
 	m := createMetricsForTesting()
 
-	hitCount := 42
-	missCount := 108
+	hitCount := 41
+	missCount := 107
 	m.RecordStoredImpCacheResult(pbsmetrics.CacheHit, hitCount)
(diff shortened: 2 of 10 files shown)


pbsmetrics/prometheus/prometheus_test.go
```diff
@@ -843,6 +843,26 @@ func TestStoredImpCacheResultMetric(t *testing.T) {
 		})
 }
 
+func TestAccountCacheResultMetric(t *testing.T) {
+	m := createMetricsForTesting()
+
+	hitCount := 37
+	missCount := 92
```
SyntaxNode, line 850: Are 37 and 92 significant in any way? .. or are these test values?

pbsmetrics/metrics.go
```diff
@@ -314,6 +314,7 @@ type MetricsEngine interface {
 	RecordUserIDSet(userLabels UserLabels) // Function should verify bidder values
 	RecordStoredReqCacheResult(cacheResult CacheResult, inc int)
 	RecordStoredImpCacheResult(cacheResult CacheResult, inc int)
+	RecordAccountCacheResult(cacheResult CacheResult, inc int)
```
SyntaxNode, line 317: The `inc` parameter seems to always be 1. One on hand it's nice to keep symmetry with the other CacheResult methods, but those pass a computed value. Thoughts?

laurb9: I felt I should leave open the possibility of fetching multiple accounts if ever need to, as the metric update does not lose much in readability.

I don't have a strong preference other than this having been easy to write and test given the symmetry :)

SyntaxNode: Neither do I. Let's leave it as for the sake of symmetry.

laurb9:
Resolved conflicts with #1515 and rebased.

SyntaxNode:
Looks great. I just want to run locally and verify metric collection. I'll do that tomorrow.

SyntaxNode (approved):
Tested locally and looks good.

I realize that the cache miss counter continues to increase with an invalid account id. Beyond the scope of this PR, but we might want to consider adding a cache for invalid accounts too.

bsardo:
LGTM. I would have preferred to have seen a pure UT in `go_metrics_test.go` but you've got the influx metrics engine coverage through `metrics_test.go`.

bsardo (approved):

PR #1545, opened 2020-10-16
Add http api for fetching accounts

laurb9:
This works the same way as the API for stored requests,
preserving the ability to query multiple ids, so the
same API endpoint can be extended to support accounts.

### DIFF
--- a/stored_requests/backends/http_fetcher/fetcher.go
+++ b/stored_requests/backends/http_fetcher/fetcher.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/prebid/prebid-server/stored_requests"
@@ -19,9 +20,13 @@ import (
 //
 // This file expects the endpoint to satisfy the following API:
 //
+// Stored requests
 // GET {endpoint}?request-ids=["req1","req2"]&imp-ids=["imp1","imp2","imp3"]
 //
-// This endpoint should return a payload like:
+// Accounts
+// GET {endpoint}?account-ids=["acc1","acc2"]
+//
+// The above endpoints should return a payload like:
 //
 // {
 //   "requests": {
@@ -34,12 +39,20 @@ import (
 //     "imp3": null // If imp3 is not found
 //   }
 // }
+// or
+// {
+//   "accounts": {
+//     "acc1": { ... config data for acc1 ... },
+//     "acc2": { ... config data for acc2 ... },
+//   },
+// }
 //
 //
 func NewFetcher(client *http.Client, endpoint string) *HttpFetcher {
 	// Do some work up-front to figure out if the (configurable) endpoint has a query string or not.
 	// When we build requests, we'll either want to add `?request-ids=...&imp-ids=...` _or_
-	// `&request-ids=...&imp-ids=...`, depending.
+	// `&request-ids=...&imp-ids=...`.
+
 	urlPrefix := endpoint
 	if strings.Contains(endpoint, "?") {
 		urlPrefix = urlPrefix + "&"
@@ -47,7 +60,11 @@ func NewFetcher(client *http.Client, endpoint string) *HttpFetcher {
 		urlPrefix = urlPrefix + "?"
 	}
 
-	glog.Info("Making http_fetcher which calls GET " + urlPrefix + "request-ids=%REQUEST_ID_LIST%&imp-ids=%IMP_ID_LIST%")
+	if url, err := url.Parse(endpoint); err != nil {
+		glog.Warningf(`Invalid endpoint "%s": %v`, endpoint, err)
+	} else {
+		glog.Infof("Making http_fetcher for endpoint %v", url)
+	}
 
 	return &HttpFetcher{
 		client:   client,
@@ -81,8 +98,68 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
 	return
 }
 
-func (fetcher *HttpFetcher) FetchAccount(ctx context.Context, accountID string) (json.RawMessage, []error) {
-	return nil, []error{stored_requests.NotFoundError{accountID, "Account"}}
+// FetchAccounts retrieves account configurations
+//
+// Request format is similar to the one for requests:
+// GET {endpoint}?account-ids=["account1","account2",...]
+//
+// The endpoint is expected to respond with a JSON map with accountID -> json.RawMessage
+// {
+//   "account1": { ... account json ... }
+// }
+// The JSON contents of account config is returned as-is (NOT validated)
+func (fetcher *HttpFetcher) FetchAccounts(ctx context.Context, accountIDs []string) (accountData map[string]json.RawMessage, errs []error) {
+	if len(accountIDs) == 0 {
+		return nil, nil
+	}
+	httpReq, err := http.NewRequestWithContext(ctx, "GET", fetcher.Endpoint+"account-ids=[\""+strings.Join(accountIDs, "\",\"")+"\"]", nil)
+	if err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts "%v" via http: %v`, accountIDs, err),
+		}
+	}
+	httpResp, err := ctxhttp.Do(ctx, fetcher.client, httpReq)
+	if err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: %v`, accountIDs, err),
+		}
+	}
+	defer httpResp.Body.Close()
+	respBytes, err := ioutil.ReadAll(httpResp.Body)
+	if err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: could not read response`, accountIDs),
+		}
+	}
+	if httpResp.StatusCode != http.StatusOK {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: unexpected response status %d`, accountIDs, httpResp.StatusCode),
+		}
+	}
+	var responseData accountsResponseContract
+	if err = json.Unmarshal(respBytes, &responseData); err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: malformed response`, accountIDs),
+		}
+	}
+	errs = convertNullsToErrs(responseData.Accounts, "Account", errs)
+	return responseData.Accounts, errs
+}
+
+// FetchAccount fetchers a single accountID and returns its corresponding json
+func (fetcher *HttpFetcher) FetchAccount(ctx context.Context, accountID string) (accountJSON json.RawMessage, errs []error) {
+	accountData, errs := fetcher.FetchAccounts(ctx, []string{accountID})
+	if len(errs) > 0 {
+		return nil, errs
+	}
+	accountJSON, ok := accountData[accountID]
+	if !ok {
+		return nil, []error{stored_requests.NotFoundError{
+			ID:       accountID,
+			DataType: "Account",
+		}}
+	}
+	return accountJSON, []error{}
 }
 
 func (fetcher *HttpFetcher) FetchCategories(ctx context.Context, primaryAdServer, publisherId, iabCategory string) (string, error) {
@@ -189,4 +266,8 @@ func convertNullsToErrs(m map[string]json.RawMessage, dataType string, errs []er
 type responseContract struct {
 	Requests map[string]json.RawMessage `json:"requests"`
 	Imps     map[string]json.RawMessage `json:"imps"`
+}
+
+type accountsResponseContract struct {
+	Accounts map[string]json.RawMessage `json:"accounts"`
 }
--- a/stored_requests/backends/http_fetcher/fetcher.go
+++ b/stored_requests/backends/http_fetcher/fetcher.go
@@ -54,7 +54,7 @@ func NewFetcher(client *http.Client, endpoint string) *HttpFetcher {
 	// `&request-ids=...&imp-ids=...`.
 
 	if url, err := url.Parse(endpoint); err != nil {
-		glog.Errorf(`Invalid endpoint "%s": %v`, endpoint, err)
+		glog.Fatalf(`Invalid endpoint "%s": %v`, endpoint, err)
 	} else {
 		glog.Infof("Making http_fetcher for endpoint %v", url)
 	}
@@ -142,7 +142,7 @@ func (fetcher *HttpFetcher) FetchAccounts(ctx context.Context, accountIDs []stri
 			fmt.Errorf(`Error fetching accounts %v via http: failed to parse response: %v`, accountIDs, err),
 		}
 	}
-	errs := convertNullsToErrs(responseData.Accounts, "Account", nil)
+	errs := convertNullsToErrs(responseData.Accounts, "Account", []error{})
 	return responseData.Accounts, errs
 }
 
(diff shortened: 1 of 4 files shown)


stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -81,8 +98,68 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
+	if err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: %v`, accountIDs, err),
+		}
+	}
+	defer httpResp.Body.Close()
+	respBytes, err := ioutil.ReadAll(httpResp.Body)
+	if err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: could not read response`, accountIDs),
```
mansinahar, line 131: We should also append the error to this log message. I see you added that for the error messages above so looks like this was missed by mistake?

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -34,20 +39,32 @@ import (
...
 	urlPrefix := endpoint
 	if strings.Contains(endpoint, "?") {
 		urlPrefix = urlPrefix + "&"
 	} else {
 		urlPrefix = urlPrefix + "?"
 	}
 
-	glog.Info("Making http_fetcher which calls GET " + urlPrefix + "request-ids=%REQUEST_ID_LIST%&imp-ids=%IMP_ID_LIST%")
+	if url, err := url.Parse(endpoint); err != nil {
+		glog.Warningf(`Invalid endpoint "%s": %v`, endpoint, err)
```
mansinahar, line 64: Thanks for adding this. Although, please consider moving this before creating `urlPrefix` on L56 because if the `endpoint` isn't valid, there's not much point in appending `?` or `&` to it.

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -81,8 +98,68 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
+// }
+// The JSON contents of account config is returned as-is (NOT validated)
+func (fetcher *HttpFetcher) FetchAccounts(ctx context.Context, accountIDs []string) (accountData map[string]json.RawMessage, errs []error) {
+	if len(accountIDs) == 0 {
+		return nil, nil
+	}
+	httpReq, err := http.NewRequestWithContext(ctx, "GET", fetcher.Endpoint+"account-ids=[\""+strings.Join(accountIDs, "\",\"")+"\"]", nil)
+	if err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts "%v" via http: %v`, accountIDs, err),
```
mansinahar, line 118: Please consider changing this error message to `Error building the http request to fetch accounts "%v"` or something along those lines so that it is clear that the error was in building the request rather than fetching the account data with a valid http request

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -81,8 +98,68 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
+	}
+	if httpResp.StatusCode != http.StatusOK {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: unexpected response status %d`, accountIDs, httpResp.StatusCode),
+		}
+	}
+	var responseData accountsResponseContract
+	if err = json.Unmarshal(respBytes, &responseData); err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: malformed response`, accountIDs),
```
mansinahar, line 142: Same as above. Consider adding the actual error to this log message

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -81,8 +98,68 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
+		return nil, errs
+	}
+	accountJSON, ok := accountData[accountID]
+	if !ok {
+		return nil, []error{stored_requests.NotFoundError{
+			ID:       accountID,
+			DataType: "Account",
+		}}
+	}
+	return accountJSON, []error{}
```
mansinahar, line 162: Please consider returning `nil` her for error rather than an empty slice as that's the preferred way in Go
Ref: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

stored_requests/backends/http_fetcher/fetcher_test.go
```diff
@@ -61,14 +63,39 @@ func TestReqsAndImps(t *testing.T) {
 	assertErrLength(t, errs, 0)
 }
 
-func TestMissingValues(t *testing.T) {
```
mansinahar, line 64: Curious, what's the reason for removing this missing value test case?

stored_requests/backends/http_fetcher/fetcher_test.go
```diff
@@ -61,14 +63,39 @@ func TestReqsAndImps(t *testing.T) {
...
+	accData, errs := fetcher.FetchAccounts(context.Background(), []string{"acc-1", "acc-2"})
+	assertMapKeys(t, accData, "acc-1", "acc-2")
+	assertErrLength(t, errs, 0)
+}
+
+func TestFetchAccountsNoData(t *testing.T) {
+	fetcher, close := newFetcherBrokenBackend()
+	defer close()
+	accData, errs := fetcher.FetchAccounts(context.Background(), []string{"req-1"})
+	assertMapKeys(t, accData)
```
mansinahar, line 79: Please consider using `assert.Len` for this

stored_requests/backends/http_fetcher/fetcher_test.go
```diff
@@ -61,14 +63,39 @@ func TestReqsAndImps(t *testing.T) {
...
+	assertErrLength(t, errs, 1)
+}
+
+func TestFetchAccount(t *testing.T) {
+	fetcher, close := newTestAccountFetcher(t, []string{"acc-1"})
+	defer close()
+
+	account, errs := fetcher.FetchAccount(context.Background(), "acc-1")
+	assert.Empty(t, errs, "Unexpected error fetching existing account")
+	assert.JSONEq(t, `"acc-1"`, string(account))
```
mansinahar, line 89: Consider adding a message to this assertion in case the assertion fails

stored_requests/backends/http_fetcher/fetcher_test.go
```diff
@@ -61,14 +63,39 @@ func TestReqsAndImps(t *testing.T) {
...
+}
+
```
mansinahar, line 91: Please consider adding a test case for an invalid json response from the http endpoint.

Also, if possible consider updating the tests to a table test format as they are easier to follow and also easier to add/modify/update test cases.

stored_requests/backends/http_fetcher/fetcher.go
laurb9: Done. Should it be an error instead of warning ?

I intend to explore replacing Endpoint with URL also, perhaps in another PR.

stored_requests/backends/http_fetcher/fetcher.go
laurb9: Will change it to
```go
fmt.Errorf(`Error fetching accounts "%v" via http: build request failed with %v`, accountIDs, err),
```

stored_requests/backends/http_fetcher/fetcher.go
laurb9: Replacing with
```go
			fmt.Errorf(`Error fetching accounts %v via http: error reading response: %v`, accountIDs, err),
```

stored_requests/backends/http_fetcher/fetcher.go
laurb9: Will replace with
```go
fmt.Errorf(`Error fetching accounts %v via http: failed to parse response: %v`, accountIDs, err),
```

stored_requests/backends/http_fetcher/fetcher.go
laurb9: Will do, thanks for the ref link

stored_requests/backends/http_fetcher/fetcher_test.go
laurb9: accident, will add back

stored_requests/backends/http_fetcher/fetcher_test.go
laurb9: will replace with `assert.Empty`
will also replace `assertErrLength` with either `assert.Len` or `assert.Empty`, it does not appear to bring much to the table.

stored_requests/backends/http_fetcher/fetcher_test.go
laurb9: Added invalid json test.

laurb9: Regarding the table test: the tests are small and differ in almost every line; the table would have to include all the differences and the eventual custom text. It may not be much of a win considering the code that needs to be added to read and execute the tests will be more dense than any of them.

stored_requests/backends/http_fetcher/fetcher.go
mansinahar: That's a good point. I'd say it should be a fatal error because if the endpoint isn't valid, the fetching wouldn't work. I see the DB fetcher also doing the same in case of invalid config. Ideally these things should be part of a validate method that validates the config for stored requests

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -34,21 +39,33 @@ import (
...
 //
 //
 func NewFetcher(client *http.Client, endpoint string) *HttpFetcher {
 	// Do some work up-front to figure out if the (configurable) endpoint has a query string or not.
 	// When we build requests, we'll either want to add `?request-ids=...&imp-ids=...` _or_
-	// `&request-ids=...&imp-ids=...`, depending.
+	// `&request-ids=...&imp-ids=...`.
+
+	if url, err := url.Parse(endpoint); err != nil {
+		glog.Errorf(`Invalid endpoint "%s": %v`, endpoint, err)
```
mansinahar, line 57: This should be `golg.Fatalf`. `Errorf` will just log the error and move on vs `Fatalf` will log the error and call `os.Exit(255)` and for an error like this, we wanna fail fast.

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -81,8 +98,68 @@ func (fetcher *HttpFetcher) FetchRequests(ctx context.Context, requestIDs []stri
...
+			fmt.Errorf(`Error fetching accounts %v via http: unexpected response status %d`, accountIDs, httpResp.StatusCode),
+		}
+	}
+	var responseData accountsResponseContract
+	if err = json.Unmarshal(respBytes, &responseData); err != nil {
+		return nil, []error{
+			fmt.Errorf(`Error fetching accounts %v via http: failed to parse response: %v`, accountIDs, err),
+		}
+	}
+	errs := convertNullsToErrs(responseData.Accounts, "Account", nil)
```
mansinahar, line 145: In this case I think it will be much more clearer to pass an empty error slice to the `convertNullsToErrs` function. In case of returns a `nil` value makes more sense.

mansinahar (commented):
LGTM in general. Just some minor comments and also, please consider adding assertion messages in the assert calls in `fetcher_test.go`

laurb9:
added commit 1154006 to address feedback
- Bail out (Fatalf) if http endpoint isn't parseable.
- Add custom messages to test assertions.

Also moved error assertions before the data assertions to make it easier to understand test failures.

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -34,21 +39,33 @@ import (
...
 	// Do some work up-front to figure out if the (configurable) endpoint has a query string or not.
 	// When we build requests, we'll either want to add `?request-ids=...&imp-ids=...` _or_
-	// `&request-ids=...&imp-ids=...`, depending.
+	// `&request-ids=...&imp-ids=...`.
+
+	if url, err := url.Parse(endpoint); err != nil {
+		glog.Fatalf(`Invalid endpoint "%s": %v`, endpoint, err)
+	} else {
+		glog.Infof("Making http_fetcher for endpoint %v", url)
+	}
```
bsardo, line 60: Nitpick: I don't think this needs to be in an `else` block now given the `Fatalf`.

stored_requests/backends/http_fetcher/fetcher_test.go
```diff
@@ -9,66 +9,113 @@ import (
...
 	reqData, impData, errs := fetcher.FetchRequests(context.Background(), []string{"req-1", "req-2"}, []string{"imp-1"})
-	assertMapKeys(t, reqData)
-	assertMapKeys(t, impData)
-	assertErrLength(t, errs, 3)
+	assert.Empty(t, reqData, "Fetching unknown reqs should return no reqs")
+	assert.Empty(t, impData, "Fetching unknown imps should return no imps")
+	assert.Len(t, errs, 3, "Fetching 3 unknown reqs+imps should return 3 errors")
+}
+
+func TestFetchAccounts(t *testing.T) {
```
bsardo, line 76: The UT code coverage in this package is lacking but I think that's mostly attributed to the fact that `FetchCategories` doesn't have any tests. We'll backfill that some other time. With that said, I think we can rather easily get coverage for a lot of the `FetchAccount` and `FetchAccounts` error cases by adding a few additional tests:

```
func TestFetchAccountsNoIDsProvided(t *testing.T) {
	fetcher, close := newTestAccountFetcher(t, []string{"acc-1", "acc-2"})
	defer close()

	accData, errs := fetcher.FetchAccounts(nil, []string{})
	assert.Len(t, errs, 0)
	assert.Nil(t, accData)
}

// Force build request failure by not providing a context
func TestFetchAccountsFailedBuildRequest(t *testing.T) {
	fetcher, close := newTestAccountFetcher(t, []string{"acc-1", "acc-2"})
	defer close()

	accData, errs := fetcher.FetchAccounts(nil, []string{"acc-1"})
	assert.Len(t, errs, 1)
	assert.Nil(t, accData)
}

// Force http error via request timeout
func TestFetchAccountsContextTimeout(t *testing.T) {
	fetcher, close := newTestAccountFetcher(t, []string{"acc-1", "acc-2"})
	defer close()

	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(0))
	defer cancel()
	accData, errs := fetcher.FetchAccounts(ctx, []string{"acc-1"})
	assert.Len(t, errs, 1)
	assert.Nil(t, accData)
}

func TestFetchAccountNoIDProvided(t *testing.T) {
	fetcher, close := newTestAccountFetcher(t, nil)
	defer close()

	account, errs := fetcher.FetchAccount(context.Background(), "")
	assert.Len(t, errs, 1)
	assert.Nil(t, account)
}
```

stored_requests/backends/http_fetcher/fetcher_test.go
mansinahar: Fair enough! I understand that it will be a lot more work to refactor these tests.

mansinahar (dismissed):

stored_requests/backends/http_fetcher/fetcher_test.go
laurb9: Will cut and paste :)

stored_requests/backends/http_fetcher/fetcher.go
laurb9: It was to keep url in scope but I can do without

bsardo (approved):

mansinahar (approved):

PR #1546, opened 2020-10-16
Add missing postgres cache init config validation

bsardo:


config/stored_requests_test.go
```diff
@@ -154,6 +157,100 @@ func TestInMemoryCacheValidationSingleCache(t *testing.T) {
...
+		{
+			description:   "No connection string",
+			connectionStr: "",
+		},
+		{
+			description:   "Connection string but no queries",
+			connectionStr: "some-connection-string",
+		},
+		{
+			description:    "Valid cache init query with zero timeout",
```
SyntaxNode, line 181: Consider defining the 0 timeout since it's mentioned in the test description. I know it's defaulted to 0, but I think it would help with readability.

config/stored_requests_test.go
```diff
@@ -154,6 +157,100 @@ func TestInMemoryCacheValidationSingleCache(t *testing.T) {
 	}).validate(AccountDataType, nil))
 }
 
+func TestPostgresConfigValidation(t *testing.T) {
```
SyntaxNode, line 160: We had no tests on this before? Yikes. Thanks for fixing it.

config/stored_requests_test.go
```diff
@@ -154,6 +157,100 @@ func TestInMemoryCacheValidationSingleCache(t *testing.T) {
...
+			wantErrorCount:         1,
+		},
+		{
+			description:      "Multiple errors: valid queries missing timeouts and refresh rates plus existing error",
+			connectionStr:    "some-connection-string",
+			cacheInitQuery:   "SELECT * FROM table;",
+			cacheUpdateQuery: "SELECT * FROM table WHERE $1",
+			existingErrors:   []error{errors.New("existing error before calling validate")},
+			wantErrorCount:   4,
+		},
```
SyntaxNode, line 230: Can you please add test coverage for:
```
	if strings.Contains(cfg.Query, "$") {
		errs = append(errs, fmt.Errorf("%s: postgres.initialize_caches.query should not contain any wildcards (e.g. $1)", section))
	}
```

That looks to me as the only missing check.

SyntaxNode (approved):

config/stored_requests.go
```diff
@@ -185,7 +185,9 @@ func (cfg *PostgresConfig) validate(dataType DataType, errs configErrors) config
 		return errs
 	}
 
-	return cfg.PollUpdates.validate(dataType, errs)
+	errs = cfg.CacheInitialization.validate(dataType, errs)
```
hhhjort, line 188: That is always annoying to miss a validation routine. I guess since the type had a different name from the struct property, it was harder to catch.

hhhjort (approved):
LGTM

PR #1549, opened 2020-10-20
Add AMX RTB adapter

nickjacob:
Adding AMX RTB adapter

Server uses standard OpenRTB 2.5. This adapter adds support for overriding the publisher id using `imp.ext.amx.tagId`, although this is optional. The parameter is named to match the bid param of [our prebid.js adapter](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/amx-rtb.md)

Maintainer email: prebid@amxrtb.com


bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2437

adapters/amx/amx.go
```diff
@@ -0,0 +1,174 @@
...
+	if pub == nil {
+		pub = &openrtb.Publisher{}
+	}
+	pub.ID = publisherID
+	return pub
+}
+
+// MakeRequests creates AMX adapter requests
+func (adapter *AmxAdapter) MakeRequests(request *openrtb.BidRequest, req *adapters.ExtraRequestInfo) (reqsBidder []*adapters.RequestData, errs []error) {
+	publisherID := getPublisherID(request.Imp)
```
SyntaxNode, line 61: With the other suggested change, this would become:
```
if publisherID, ok := getPublisherID(request.Imp); ok {
	if request.App != nil {
		request.App.Publisher = ensurePublisherWithID(request.App.Publisher, publisherID)
	}
	if request.Site != nil {
		request.Site.Publisher = ensurePublisherWithID(request.Site.Publisher, publisherID)
	}
}
```

adapters/amx/amx.go
```diff
@@ -0,0 +1,174 @@
...
+func getPublisherID(imps []openrtb.Imp) *string {
+	for _, imp := range imps {
+		paramsSource := (*json.RawMessage)(&imp.Ext)
+		var params amxExt
+		if err := json.Unmarshal(*paramsSource, &params); err == nil {
+			if params.Bidder.PublisherID != "" {
+				return &params.Bidder.PublisherID
+			}
+		} else {
+			fmt.Printf("unable to decode imp.ext.bidder: %v", err)
```
SyntaxNode, line 44: There is no consequence to this print statement, neither in the response or the in the log. The chance of this happening is also likely 0% due to preprocessing by the "core" logic. I recommend removing the print statement since the publisher id is optional for your adapter.

adapters/amx/amx.go
```diff
@@ -0,0 +1,174 @@
...
+	if publisherID != nil {
+		if request.App != nil {
+			request.App.Publisher = ensurePublisherWithID(request.App.Publisher, *publisherID)
+		}
+		if request.Site != nil {
+			request.Site.Publisher = ensurePublisherWithID(request.Site.Publisher, *publisherID)
+		}
+	}
+
+	encoded, err := json.Marshal(request)
```
SyntaxNode, line 71: Nitpick: In Go, it's common to remove the line break after the `json.Marshal` call to group it with the error handling code.

adapters/amx/amx.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	reqBidder := &adapters.RequestData{
+		Method:  "POST",
+		Uri:     adapter.endpoint + "?v=pbs1.0",
```
SyntaxNode, line 82: Please consider adding this query parameter safely using Go's url methods within the `NewAmxBidder` method.

adapters/amx/amx.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	if http.StatusBadRequest == response.StatusCode {
+		return nil, []error{&errortypes.BadInput{
+			Message: fmt.Sprint("Invalid request: 400"),
+		}}
+	}
+
+	if http.StatusOK != response.StatusCode {
+		return nil, []error{&errortypes.BadInput{
```
SyntaxNode, line 105: Since this error would likely mean the server responded back with a bad response, it's common to use `BadServerResponse` here instead.

adapters/amx/amx.go
```diff
@@ -0,0 +1,174 @@
...
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			return mediaType, nil
+		}
+	}
+
+	return "", &errortypes.BadInput{
+		Message: fmt.Sprintf("No impression for: %s", impID),
+	}
+}
```
SyntaxNode, line 152: Does your bid response include the bid type? Although the code you have here is common in the project, there's been issues reported (see https://github.com/prebid/prebid-server/issues/862) that this approach isn't always reliable for multiformat impressions. Is there enough information to detect the type based on the bid response?

adapters/amx/usersync.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"text/template"
+
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/usersync"
+)
+
+// NewAmxSyncer produces an AMX RTB usersyncer
+func NewAmxSyncer(temp *template.Template) usersync.Usersyncer {
+	return adapters.NewSyncer("amx", 737, temp, adapters.SyncTypeRedirect)
```
SyntaxNode, line 12: GVL ID is registered to `Monet Engine Inc`. Is that correct for AMX?

adapters/amx/usersync_test.go
```diff
@@ -0,0 +1,24 @@
...
+func TestDmxSyncer(t *testing.T) {
+	temp := template.Must(template.New("sync-template").Parse(testSyncEndpoint))
+	syncer := NewAmxSyncer(temp)
+	syncInfo, err := syncer.GetUsersyncInfo(privacy.Policies{})
+	assert.NoError(t, err)
+	assert.Equal(t, "redirect", syncInfo.Type)
+	assert.Equal(t, testSyncEndpoint, syncInfo.URL)
+	assert.Equal(t, false, syncInfo.SupportCORS)
+	assert.EqualValues(t, 737, syncer.GDPRVendorID())
+}
```
SyntaxNode, line 24: Please follow the format form the other syncer tests.

usersync/usersyncers/syncer.go
```diff
@@ -99,6 +100,7 @@ func NewSyncerMap(cfg *config.Configuration) map[openrtb_ext.BidderName]usersync
 	insertIntoMap(cfg, syncers, openrtb_ext.BidderAdtelligent, adtelligent.NewAdtelligentSyncer)
 	insertIntoMap(cfg, syncers, openrtb_ext.BidderAdvangelists, advangelists.NewAdvangelistsSyncer)
 	insertIntoMap(cfg, syncers, openrtb_ext.BidderAJA, aja.NewAJASyncer)
+	insertIntoMap(cfg, syncers, openrtb_ext.BidderAmx, amx.NewAmxSyncer)
```
SyntaxNode, line 103: Per Go naming convention, this should be `BidderAMX`, `NewAMXSyncer`, etc...

adapters/amx/amx.go
```diff
@@ -0,0 +1,174 @@
...
+func NewAmxBidder(endpoint string) *AmxAdapter {
+	return &AmxAdapter{endpoint: endpoint}
+}
+
+type amxExt struct {
+	Bidder amxParams `json:"bidder"`
+}
+
+type amxParams struct {
+	PublisherID string `json:"tagId,omitempty"`
```
SyntaxNode, line 32: Please move `amxParams` to the `openrtb_ext` package in the file openrtb_ext/imp_amx.go with the name `ExtImpAmx` to match our standard convention.

SyntaxNode (commented):
Thank you for your contribution. Very glad to see the docs PR already open. :)

nickjacob:
Sorry for some reason github won't let me resolve the comments or reply 
here—but thanks for the great CR!

I made the changes you noted. Our bids do always have `startdelay` in 
bid.ext when video, so I used that instead to determine the type.

For the GVL—737 is correct thanks

I renamed everything to AMX (from Amx). Let me know if there's anything 
else

Thanks!

adapters/amx/amx.go
```diff
@@ -0,0 +1,158 @@
...
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(5)
+
+	for _, sb := range bidResp.SeatBid {
+		for _, bid := range sb.Bid {
+			var bidExt amxBidExt
+			if err := json.Unmarshal(bid.Ext, &bidExt); err == nil {
```
SyntaxNode, line 119: There's a problem here because if `bid.Ext` is empty the `json.Unmarshal` call will return an error and not produce a bid response. This doesn't fail the exemplary tests because you're missing `expectedBidResponses` from those files.

This will need to be changed to handle a missing `bid.Ext`. Perhaps something like:
```
		for _, bid := range sb.Bid {

			bidExt, bidExtErr := getBidExt(bid.Ext)
			if bidExtErr != nil {
				errs = append(errs, bidExtErr)
				continue
			}

			bidType := getMediaTypeForBid(bidExt)

			b := &adapters.TypedBid{
				Bid:     &bid,
				BidType: bidType,
			}
			if b.BidType == openrtb_ext.BidTypeVideo {
				b.Bid.AdM = interpolateImpressions(bid, bidExt)
			}

		}
	}
```

```
func getBidExt(ext json.RawMessage) (amxBidExt, error) {
	if len(ext) == 0 {
		return amxBidExt{}, nil
	}

	var bidExt amxBidExt
	err := json.Unmarshal(ext, &bidExt)

	return bidExt, err
}
```

adapters/amx/amx_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+	// check that the publisher ID overrides the site.publisher.id if provided
+	var body openrtb.BidRequest
+	if err := json.Unmarshal(actualAdapterRequests[0].Body, &body); err != nil {
+		t.Errorf("failed to read bid request")
+	}
+	if body.Site.Publisher.ID != "publisher_id_example" {
+		t.Errorf("invalid publisher id: %s", body.Site.Publisher.ID)
+	}
+}
```
SyntaxNode, line 74: Please consider using testify assert statements in your tests. Not strictly required. A bit easier to read though and you don't need to be concerned with edge cases.

```
assert.NoError(t, actualAdapterRequests, "request")
assert.Empty(t, err, "request errors")
...
assert.Equal(t, "publisher_id_example", body.Site.Publisher.ID, "publisher")
```

adapters/amx/amx_test.go
```diff
@@ -0,0 +1,354 @@
...
+	inputRequest := openrtb.BidRequest{
+		Imp: []openrtb.Imp{imp1},
+		Site: &openrtb.Site{
+			Publisher: &openrtb.Publisher{
+				ID: "1234567",
+			},
+		},
+		App: &openrtb.App{ID: "cansanuabnua", Publisher: &openrtb.Publisher{ID: "whatever"}},
+		ID:  "1234",
+	}
```
SyntaxNode, line 102: It's not valid for a BidRequest to contain both a `Site` and `App` section. This will be caught earlier on by the framework. For this test, please only use the `App`.

adapters/amx/amx_test.go
```diff
@@ -0,0 +1,354 @@
...
+            "adm": "<?xml version=\"1.0\" encoding=\"UTF-8\" ?><VAST version=\"2.0\"><Ad id=\"128a6.44d74.46b3\"><InLine><Error><![CDATA[http://example.net/hbx/verr?e=]]></Error><Impression><![CDATA[http://example.net/hbx/vimp?lid=test&aid=testapp]]></Impression><Creatives><Creative sequence=\"1\"><Linear><Duration>00:00:15</Duration><TrackingEvents><Tracking event=\"firstQuartile\"><![CDATA[https://example.com?event=first_quartile]]></Tracking></TrackingEvents><VideoClicks><ClickThrough><![CDATA[http://example.com]]></ClickThrough></VideoClicks><MediaFiles><MediaFile delivery=\"progressive\" width=\"16\" height=\"9\" type=\"video/mp4\" bitrate=\"800\"><![CDATA[https://example.com/media.mp4]]></MediaFile></MediaFiles></Linear></Creative></Creatives></InLine></Ad></VAST>",
+            "adomain": ["amxrtb.com"],
+            "iurl": "https://assets.a-mo.net/300x250.v2.png",
+            "cid": "1",
+            "crid": "1",
+            "h": 600,
+            "w": 300,
+            "ext": {
+							"himp": ["https://example.com/imp-tracker/pixel.gif?param=1&param2=2"],
+							"startdelay": 0
```
SyntaxNode, line 206: Please fix the whitespace on these lines for consistency.

adapters/amx/amx_test.go
```diff
@@ -0,0 +1,354 @@
...
+	inputRequest := openrtb.BidRequest{
+		Imp: []openrtb.Imp{imp1, imp1, imp1},
+		Site: &openrtb.Site{
+			Publisher: &openrtb.Publisher{
+				ID: "10007",
+			},
+		},
+		User: &openrtb.User{Ext: json.RawMessage(`{"eids": [{
+						"source": "adserver.org",
+							"uids": [{
```
SyntaxNode, line 254: Same here. Please use consistent whitespace for the json blobs.

adapters/amx/amx_test.go
```diff
@@ -0,0 +1,354 @@
...
+		t.Errorf("Invalid bid.ext: %v", err)
+	}
+
+	data := interpolateImpressions(bid, bidExt)
+	find := strings.Index(data, "example2.com/nurl")
+	if find == -1 {
+		t.Errorf("String was not found")
+	}
+
+	if strings.Index(data, "example.com/pixel.png") == -1 {
```
SyntaxNode, line 351: assert.Contains(t, data, "example.com/pixel.png", "interpolated into vast")

adapters/amx/amxtest/exemplary/app.simple.json
```diff
@@ -0,0 +1,152 @@
...
+              "h": 300,
+              "w": 250
+            }]
+          }],
+          "cur": "USD"
+        }
+      }
+    }
+  ]
+}
```
SyntaxNode, line 152: Please add expectedBidResponses to the json test files.

SyntaxNode (commented):
Thank you for the quick update from our feedback. I took a closer look into the tests and found a few items to address.

nickjacob:
Thanks @SyntaxNode  -- I made the requested CR changes; refactored to use assert which did really improve the tests, adding some more test coverage and "supplemental" JSON tests, expectedBidResponse & expectedMakeBidsErrors in JSON tests

adapters/amx/amx_test.go
```diff
@@ -0,0 +1,335 @@
...
+	actualAdapterRequests, _ := adapter.MakeRequests(&inputRequest, &adapters.ExtraRequestInfo{})
+	assert.Len(t, actualAdapterRequests, 1, "there should be 1 request")
+
+	body := getRequestBody(actualAdapterRequests)
+	assert.NotNil(t, body, "the generated OpenRTB request is not valid JSON")
+
+	userExt := getUserExt(body.User)
+	assert.NotNil(t, userExt, "the generated user.ext is invalid")
+
+	assert.Len(t, userExt.Eids, 2)
```
mansinahar, line 281: Not exactly sure what exactly this test is testing because I don't see any modifications around user ext in your `MakeRequests` function. This to me looks more like making sure that user exts are not getting modified by `MakeRequests`. Is that correct? 

Also, I'd suggest adding these tests in this file as part of the JSON tests rather than unit tests as that will help remove a lot of setup for your tests which is covered by `RunJSONBidderTest` and will also provide consistency in tests with other adapters. In fact some of these test cases are already covered with the JSON tests.

static/bidder-params/amx.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "AMX RTB Adapter Params",
+  "description": "A schema to validate params accepted by the AMX adapter",
+  "type": "object",
+  "properties": {
+    "publisher_id" : {
```
mansinahar, line 7: Bidder params should have the params expected by your adapter which in your case should be `tagid`. You will have to update the `adapters/amx/amxtest/params/race/display.json` and `adapters/amx/amxtest/params/race/video.json` file accordingly as well.

nickjacob:
Thanks @mansinahar -- I made those changes & rebased against master

adapters/amx/amx_test.go
```diff
@@ -0,0 +1,335 @@
...
+)
+
+const (
+	amxTestEndpoint = "http://pbs-dev.amxrtb.com/auction/openrtb"
+	defaultImpExt   = "{\"bidder\":{\"tagId\":\"publisher_id_example\"}}"
+)
+
+func TestJsonSamples(t *testing.T) {
+	t.Logf("TESTING JSON SAMPLES!!")
+	adapterstest.RunJSONBidderTest(t, "amxtest", new(AMXAdapter))
```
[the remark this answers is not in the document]
mansinahar: @nickjacob Could you please address this? looks like this was missed in the update. Thanks!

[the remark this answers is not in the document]
nickjacob: sorry about that missed this -- just pushed another commit + with the table-driven test for the publisher ID logic

mansinahar (commented):
Please add a `params_test.go` file as well to test the bidder params. For reference you can check this one for 33across: https://github.com/prebid/prebid-server/blob/master/adapters/33across/params_test.go

adapters/amx/amx.go
```diff
@@ -0,0 +1,197 @@
...
+
+// NewAMXBidder creates an AMXAdapter
+func NewAMXBidder(endpoint string) *AMXAdapter {
+	endpointURL, err := url.Parse(endpoint)
+	if err != nil {
+		glog.Fatalf("invalid endpoint provided to AMX: %s, error: %v", endpoint, err)
+		return nil
+	}
+
+	qs, _ := url.ParseQuery(endpointURL.RawQuery)
```
SyntaxNode, line 35: Please handle the error from `url.ParseQuery` in a similar manner as you did with `url.Parse`.

nickjacob:
added params_test.go, improved testing of interpolation function + better error handling in `NewAMXBidder`. Also added default sync endpoint

also rebased off of master just to be safe

mansinahar (dismissed):

SyntaxNode:
@nickjacob Can you please address my comment on the docs PR?

SyntaxNode (approved):
I'm good to merge this in. The docs PR is open with one remaining comment, but I don't see a reason for that to be a blocker here.

nickjacob:
@SyntaxNode  -- for the docs PR; we don't want to have all those same parameters for PBS that we have in PBJS (e.g. testMode is the same as test: 1, so not necessary). Should I make a second table in the doc that describes the PBS parameters separate from PBJS?

SyntaxNode:
> @SyntaxNode -- for the docs PR; we don't want to have all those same parameters for PBS that we have in PBJS (e.g. testMode is the same as test: 1, so not necessary). Should I make a second table in the doc that describes the PBS parameters separate from PBJS?

Yes, please. The `testMode` makes sense. 

Before I git the merge button... what about `tagId` vs `publisher_id` and the `adUnitId`?

nickjacob:
I ended up changing the parameter here to be "tagId" to match "tagId" in PBJS. adUnitId is probably a good idea you're right. I'll add that now so the PBS version better matches PBJS

SyntaxNode:
> I ended up changing the parameter here to be "tagId" to match "tagId" in PBJS. adUnitId is probably a good idea you're right. I'll add that now so the PBS version better matches PBJS

Ah, right. Sorry. Forgot about the `tagId` change was already made. I'll hold off on merging for the `adUnitId` then. Thank you.

SyntaxNode (approved):

mansinahar (approved):

PR #1550, opened 2020-10-21
[Invibes] remove user sync for invibes

invibes:
Remove usersync endpoint


invibes:
We need to remove the usersync because we receive a large number of request on this endpoint althow we haven't integrated with any publishers yet (we have integrations only with prebid.js but not with prebid server)
When is the usersync url called?

SyntaxNode:
Thank you for the contribution. We'll look at this PR soon.

bsardo (approved):

hhhjort (approved):

PR #1553, opened 2020-10-22
Add Accounts to http cache events

laurb9:
I just added Accounts alongside Requests and Imps in the events for minimal impact change.

We should consider replacing the event with a more generic `{dataType, data}` message in the future though, to support other data types through the same mechanism.

PR #1556, opened 2020-10-27
Add accounts to and refactor http cache events

laurb9:
Add account support for refresh events for #1395, after #1545 and #1519 were merged. 
Potentially supersedes #1553 if we choose to go this route.

Refactored events:
- Make Save and Invalidate events generic and single-typed: `{DataType, Data}`. They were previously a `{RequestsData, ImpData}` combo.
- Add ImpDataType to DataTypes (to avoid creating yet another selector key for these types of data)
- Add tests for Account
- Refactored http startup tests into a list of testcases.

### DIFF
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -15,6 +15,7 @@ type DataType string
 
 const (
 	RequestDataType    DataType = "Request"
+	ImpDataType        DataType = "Imp"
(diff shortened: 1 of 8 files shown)


laurb9:
Closing. This needs more work so it does not change the HTTP API for on-demand event invalidation/updates

Issue #1323, opened 2020-06-01
Controlling GDPR behavior for AMP


bretg:
We discussed in the Prebid Server committee that publishers sometimes implement GDPR differently for AMP and Web due to differences in the CMPs available. The request has come up to enable accounts to turn off GDPR enforcement at the integration type level: AMP/Web/App.

The TCF2 requirements at https://docs.google.com/document/d/1fBRaodKifv1pYsWY3ia-9K96VHUjd8kKvxZlOsozm8E/edit# have been updated with proposed config. See requirement 21 in the Prebid Server section.

The proposed account config is:

- account.A.gdpr.integration_enabled.amp: true/false // default to account.A.gdpr.enabled
- account.A.gdpr.integration_enabled.app: true/false // default to account.A.gdpr.enabled
- account.A.gdpr.integration_enabled.web: true/false // default to account.A.gdpr.enabled
- account.A.gdpr.integration_enabled.video: true/false // default to account.A.gdpr.enabled

We don't believe there's a need to control the integration behavior at the purpose level.

Notes:
- /cookie_sync and /setuid are not separately controllable -- they're used across all of the integration types, so depend only on the status of account.A.gdpr.enabled.
- there's no cross-account mechanism for turning off GDPR enforcement for one integration type. This is purely a per-account control. 
- if account.A.gdpr.enabled is not defined, then the host-level gdpr.enabled is used as a default for integration_enabled

This is related to Issue #1312

bretg:
This feature is in PBS-Java 1.41

SyntaxNode:
This is implemented in PBS-Go 0.137.0.

PR #1564, opened 2020-11-03
Add account/ host GDPR enabled flags & account per request type GDPR enabled flags

bsardo:
The order of precedence for determining whether GDPR is enabled is:
1) Use account-level request type setting if specified
2) Use account-level general setting if specified
3) Use host-level general setting

Related issue: https://github.com/prebid/prebid-server/issues/1323


exchange/utils.go
```diff
@@ -94,7 +95,12 @@ func cleanOpenRTBRequests(ctx context.Context,
 	privacyLabels.COPPAEnforced = privacyEnforcement.COPPA
 	privacyLabels.LMTEnforced = lmtEnforcer.ShouldEnforce(unknownBidder)
 
-	if gdpr == 1 {
+	GDPREnabled := accountLevelGDPREnabled(account, labels.RType)
```
SyntaxNode, line 98: Nitpick: Variable names within a method should be lower case. `GDPREnabled` -> `gdprEnabled`

exchange/utils.go
```diff
@@ -127,6 +133,36 @@ func cleanOpenRTBRequests(ctx context.Context,
 	return
 }
 
+// GDPRAccountEnabled returns the GDPR setting for the request type in the account config if defined or
+// the general GDPR setting in the account config if defined; otherwise it returns nil
+func accountLevelGDPREnabled(account *config.Account, requestType pbsmetrics.RequestType) *bool {
```
SyntaxNode, line 138: Would this be better situated as a receiver of config.Account.GDPR?

```
account.GDPR.EnabledForRequestType(requestType pbsmetrics.RequestType) *bool
```

.. or with my refactor comment above, this could still be simplified to:

```
func accountLevelGDPREnabled(accountConfig *config.Account, privacyConfig config.Privacy, requestType pbsmetrics.RequestType) bool {
    if x := accountConfig.GDPR.EnabledForRequestType(requestType); x != nil {
        return *x
    }
    return privacyConfig.GDPR.Enabled
}
```

config/accounts.go
```diff
@@ -6,4 +6,19 @@ type Account struct {
...
+
+// AccountGDPR represents account-specific GDPR configuration
+type AccountGDPR struct {
+	Enabled            *bool                  `mapstructure:"enabled" json:"enabled,omitempty"`
+	IntegrationEnabled AccountGDPRIntegration `mapstructure:"integration_enabled" json:"integration_enabled"`
+}
+
+// AccountGDPRIntegration indicates whether GDPR is enabled for each request type
+type AccountGDPRIntegration struct {
+	AMP   *bool `mapstructure:"amp"   json:"amp,omitempty"`
```
SyntaxNode, line 20: Why is there a mapstructure on `AccountGDPR` but not `AccountGDPRIntegration`? Is the `mapstructure` required if there's already a json attribute? I really don't know. Just looking to reduce duplication if we can.

exchange/utils.go
```diff
@@ -94,7 +95,12 @@ func cleanOpenRTBRequests(ctx context.Context,
 	privacyLabels.COPPAEnforced = privacyEnforcement.COPPA
 	privacyLabels.LMTEnforced = lmtEnforcer.ShouldEnforce(unknownBidder)
 
-	if gdpr == 1 {
+	GDPREnabled := accountLevelGDPREnabled(account, labels.RType)
+	if GDPREnabled == nil {
+		GDPREnabled = hostLevelGDPREnabled(privacyConfig)
+	}
+
+	if gdpr == 1 && *GDPREnabled {
```
SyntaxNode, line 103: This made me scratch my head for what the `if` statement would do if the `GDPREnabled` is nil. Segfault? That's not possible with the code path, so perhaps a bit of a refactor can eliminate that uncertainty. Consider moving the host level check into the helper func and have the func return a bool instead of a bool pointer.

exchange/utils.go
bsardo: Ah yep I was just working on CCPA thinking about possibly refactoring that so it's a receiver; I'll move forward with that change.

config/accounts.go
bsardo: The `mapstructure` isn't needed so I removed it.

config/accounts.go
```diff
@@ -1,9 +1,61 @@
...
+// RequestType : Request type enumeration
+type RequestType string
+
+// The request types
+const (
+	RequestTypeAMP   RequestType = "AMP"
+	RequestTypeApp   RequestType = "app"
+	RequestTypeVideo RequestType = "video"
+	RequestTypeWeb   RequestType = "web"
+)
```
bsardo, line 12: I added this `RequestType` type otherwise there would be a circular dependency if this package `config` was dependent on `pbsmetrics`. I added a map from the `pbsmetrics` `RequestType` to this type elsewhere. Ideally this wouldn't live here. We can find a better place for this but I think there's a larger related issue here regarding package responsibilities that led to the circular dependency issue and this solution. For example, I think the `config` package does too much but perhaps that's a larger team discussion.

hhhjort: Good question on where this should live. Perhaps we need a "request constants" or "auction constants" package to resolve.

config/accounts.go
```diff
@@ -1,9 +1,61 @@
...
+	ID            string      `json:"id"`
+	Disabled      bool        `json:"disabled"`
+	CacheTTL      DefaultTTLs `json:"cache_ttl"`
+	EventsEnabled bool        `json:"events_enabled"`
+	GDPR          AccountGDPR `json:"gdpr"`
+}
+
+// AccountGDPR represents account-specific GDPR configuration
+type AccountGDPR struct {
+	Enabled            *bool                  `json:"enabled,omitempty"`
```
laurb9, line 25: I am curious why these fields are pointers, unlike the other "enabled" fields which are just bool

config/accounts.go
```diff
@@ -1,9 +1,61 @@
...
+	RequestTypeVideo RequestType = "video"
+	RequestTypeWeb   RequestType = "web"
+)
+
 // Account represents a publisher account configuration
 type Account struct {
-	ID            string      `mapstructure:"id" json:"id"`
-	Disabled      bool        `mapstructure:"disabled" json:"disabled"`
-	CacheTTL      DefaultTTLs `mapstructure:"cache_ttl" json:"cache_ttl"`
-	EventsEnabled bool        `mapstructure:"events_enabled" json:"events_enabled"`
```
laurb9, line 8: `mapstructure` was intended for parsing `account_defaults` section in the yaml config with viper. Does it still work without it ?

config/accounts_test.go
```diff
@@ -0,0 +1,114 @@
...
+		giveVideoGDPREnabled *bool
+		giveWebGDPREnabled   *bool
+		wantEnabled          *bool
+	}{
+		{
+			description:        "GDPR AMP integration enabled, general GDPR disabled",
+			giveRequestType:    RequestTypeAMP,
+			giveGDPREnabled:    &[]bool{false}[0],
+			giveAMPGDPREnabled: &[]bool{true}[0],
+			wantEnabled:        &[]bool{true}[0],
```
laurb9, line 25: what about this ?
```go
True, False := true, false
...
wantEnabled:        &True,
```

exchange/exchange_test.go
```diff
@@ -2456,6 +2457,7 @@ func TestUpdateHbPbCatDur(t *testing.T) {
 }
 
 type exchangeSpec struct {
+	Enabled           bool                   `json:"enabled"`
```
laurb9, line 2460: minor, shouldn't this be GDPREnabled / gdpr_enabled ?

config/accounts.go
```diff
@@ -1,9 +1,61 @@
...
+
+	switch requestType {
+	case RequestTypeAMP:
+		integrationEnabled = a.IntegrationEnabled.AMP
+	case RequestTypeApp:
+		integrationEnabled = a.IntegrationEnabled.App
+	case RequestTypeVideo:
+		integrationEnabled = a.IntegrationEnabled.Video
+	case RequestTypeWeb:
+		integrationEnabled = a.IntegrationEnabled.Web
```
laurb9, line 42: This could be just `return` ?

config/accounts.go
```diff
@@ -1,9 +1,61 @@
...
+// AccountGDPR represents account-specific GDPR configuration
+type AccountGDPR struct {
+	Enabled            *bool                  `json:"enabled,omitempty"`
+	IntegrationEnabled AccountGDPRIntegration `json:"integration_enabled"`
+}
+
+// EnabledForRequestType indicates whether GDPR is turned on at the account level for the specified request type
+// by using the request type setting if defined or the general GDPR setting if defined; otherwise it returns nil
+func (a *AccountGDPR) EnabledForRequestType(requestType RequestType) *bool {
+	var integrationEnabled *bool
```
laurb9, line 32: I feel this additional setting has potential for ambiguous situations. For example, [enabled=true, all others=false], or [enabled=false, all others=true]. With account defaults layered in, activation of the individual settings is even less clear.

What are your thoughts on removing this fallback altogether ? We do not lose any configurability as the behavior can be fully specified by the four settings in the same section, and overrides apply straight on each field.

config/accounts.go
bsardo: I don't think so because the value could for a given request type could be `nil`, in which case we want to check the general account level GDPR flag on lines 48-50 and return that.

config/accounts_test.go
```diff
@@ -0,0 +1,114 @@
...
+		giveAMPGDPREnabled   *bool
+		giveAppGDPREnabled   *bool
+		giveVideoGDPREnabled *bool
+		giveWebGDPREnabled   *bool
+		wantEnabled          *bool
+	}{
+		{
+			description:        "GDPR AMP integration enabled, general GDPR disabled",
+			giveRequestType:    RequestTypeAMP,
+			giveGDPREnabled:    &[]bool{false}[0],
```
hhhjort, line 23: So you are setting `giveGDPREnabled` to the address (`&`) of the first element (`[0]`) of a slice of booleans (`[]bool`) with just one element, set to false (`{false}`)? Would it be simpler to set `bfalse := false` and then `&bfalse` here? And likewise for true values?

bsardo: 👍 @laurb9 raised a similar point. I did consider that but chose this so everything was in line but I understand it's a bit more complicated. I'm more than happy to change it to what you're both suggesting as the consensus appears to favor this approach.

config/accounts.go
bsardo: @hhhjort @SyntaxNode My take was that the committee already agreed on the configuration settings so it would be preferred to follow that, though I understand the two versions sometimes deviate from one another. I recognize that potential ambiguity @laurb9 pointed out. How do you guys feel about this?

Regarding account defaults, I figured the defaults for all of these is `nil`; you either need to explicitly set these on the account or fall back to whatever the host settings are.

config/accounts.go
SyntaxNode: Indeed. This is seems to be fairly standard Go convention to workaround the issue. I agree though, requiring use of a pointer to express "is present in JSON" is a fairly large code smell.

config/accounts_test.go
SyntaxNode: .. or truePointer / falsePointer since it's weird for method scoped variables to start with a capital letter. Same idea.

config/accounts.go
SyntaxNode: Agreed that those cases are ambiguous. The intent is to provide an easy on/off switch with the ability to fine tune it if needed per request type. I suppose this is geared toward expressiveness in config. If an account doesn't care about each integration type, it's easier to set one enabled flag instead of four. The fallback also better defines future compatibility if a new integration type is added in the future.

Since this is implementing an already approved spec, I recommend moving this conversation to https://github.com/prebid/prebid-server/issues/1323.

I'd prefer to keep the same settings as PBS-Java currently for simplicity's sake.

config/accounts.go
bsardo: Given that we currently only allow setting the account default `disabled` field on the config and it's a bool and not a struct, it doesn't look like we need the `mapstructure` annotation on the account structures. I ran some manual tests to verify you only need that annotation for non-primitive types; it wasn't needed to map bool and int config values to the structs. I'm ok with either adding or removing the annotations. @SyntaxNode @hhhjort any preference?

hhhjort: I think the reason `mapstructure` isn't needed on primitives is that viper will just fall back on the property name in the struct, and it does not care about capitalization. So it is looking for "Disabled", sees "disabled" in the yaml or "DISABLED" in the env variables, and calls it a match. So the mapstructure can document the "preferred" format of the name, and help protect against a newer version of viper becoming more sensitive to capitalization.

bsardo: 👍 I'll re-add those annotations.

config/accounts.go
SyntaxNode: What do you think of calling it 'IntegrationType'? Also.. why is the "AMP" value not lower case like the others?

config/accounts.go
```diff
@@ -1,9 +1,61 @@
...
+	CacheTTL      DefaultTTLs `mapstructure:"cache_ttl"      json:"cache_ttl"`
 	EventsEnabled bool        `mapstructure:"events_enabled" json:"events_enabled"`
+	GDPR          AccountGDPR `mapstructure:"gdpr"           json:"gdpr"`
+}
+
+// AccountGDPR represents account-specific GDPR configuration
+type AccountGDPR struct {
+	Enabled            *bool                  `mapstructure:"enabled"             json:"enabled,omitempty"`
+	IntegrationEnabled AccountGDPRIntegration `mapstructure:"integration_enabled" json:"integration_enabled"`
+}
```
SyntaxNode, line 27: Did GoFmt organize the attributes this way?

hhhjort: "AMP" is "AMP" because it is an acronym, "Accelerated Mobile Pages" or something like that. The others are actual words.

SyntaxNode (commented):
Looks great. Just left two minor comment nitpicks.

hhhjort (commented):
I can't find any other issues than Scott's nitpicks.

config/accounts.go
```diff
@@ -1,9 +1,61 @@
 package config
 
+// IntegrationType : Integration type enumeration
+type IntegrationType string
+
+// The integration types
```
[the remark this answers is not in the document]
bsardo: Are we going to use a linter? If so, golint says that this section should have a comment because the constants are exported.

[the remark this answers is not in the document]
hhhjort: Interesting, normally the linting rules say that the comment should start with the variable/function name. In this case should each lntegrationType have a comment? In any case, the comment above really doesn't say much beyond "This is here to make the linter happy" and it seems doubtful it would make a linter happy as written.

bsardo:
Manually verified for web, app and amp

SyntaxNode (approved):

hhhjort (approved):

Issue #1312, opened 2020-05-21
GDPR AMP Special Case


SyntaxNode:
In reviewing our privacy policy enforcement recently, I discovered that there was a special GDPR case for AMP auctions. This outdated logic didn't remove the BuyerID for AMP auctions because initially AMP did not support passing through a consent string. This legacy behavior was removed in 0.111.0.

Turns out, this affected a lot of publishers that weren't configured to send GDPR consent strings through AMP. In today's 0.112.0 release, we added back this behavior as a config option (`gdpr. amp_exception`) for hosts. It defaults to `false`, so you'll need to add config on your servers to re-enable the behavior. 

This is designed as a temporary measure to give time for publishers to add GDPR content strings to their AMP requests. I don't have a proposed date for removal, but I also don't want to set the expectation that this is a permanent option.

SyntaxNode:
We discussed implementing the proposal in https://github.com/prebid/prebid-server/issues/1323 as a permanent replacement of this temporary solution.

SyntaxNode:
The proposed long term solution described by https://github.com/prebid/prebid-server/issues/1323 has been implemented. As discussed earlier in this thread, we plan to remove this special case code in an upcoming release. Please migrate to account level controls if you still rely on the hardcoded AMP special case.

SyntaxNode:
Following the processes decided by the Prebid.org Server Committee, we are removing this feature in today's 0.142.0 release. Hosts who have this value configured to true (the default is false) will receive a fatal startup warning.

PR #1565, opened 2020-11-05
Remove legacy GDPR AMP config flag used to prevent buyer ID scrub on AMP requests

bsardo:
Related issue: https://github.com/prebid/prebid-server/issues/1312

The replacement strategy in https://github.com/prebid/prebid-server/pull/1564 should be merged in before this.


SyntaxNode:
We'll wait to merge this until we build the final release this year. We had publisher complaints the first time around so this is deemed higher risk than normal, and we're not keen on doing this in Q4 2020. Also gives more time for hosts to discover the new GDPR account level config.

config/config.go
```diff
@@ -1045,7 +1044,6 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("gdpr.tcf2.special_purpose1.enabled", true)
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.enabled", true)
 	v.SetDefault("gdpr.tcf2.purpose_one_treatement.access_allowed", true)
-	v.SetDefault("gdpr.amp_exception", false)
```
SyntaxNode, line 1048: What do you think of providing a warning to the host if this is enabled to guide them to understanding the change?

guscarreon (commented):
Looks pretty good. I just have a couple of questions.

SyntaxNode:
Let's discuss with the Prebid.org Committee how we want to handle feature deprecation.

config/config.go
```diff
@@ -231,6 +231,10 @@ func (cfg *GDPR) validate(errs configErrors) configErrors {
 	if cfg.HostVendorID < 0 || cfg.HostVendorID > 0xffff {
 		errs = append(errs, fmt.Errorf("gdpr.host_vendor_id must be in the range [0, %d]. Got %d", 0xffff, cfg.HostVendorID))
 	}
+
+	if cfg.AMPException == true {
+		glog.Warning("GDPR AMP Exception is enabled but it is no longer supported. If you need to disable GDPR for AMP, you can do so on a per-account basis or disable it for all request types at the host level.")
```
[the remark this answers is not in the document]
bsardo: We recently added a warning in the last release indicating that this config flag is deprecated and will be removed in a future version. This commit will now throw a fatal error as discussed.

config/config.go
```diff
@@ -214,7 +214,7 @@ func (cfg *GDPR) validate(errs []error) []error {
 		glog.Warning("gdpr.host_vendor_id was not specified. Host company GDPR checks will be skipped.")
 	}
 	if cfg.AMPException == true {
-		glog.Warning("gdpr.amp_exception is deprecated and will be removed in a future version. If you need to disable GDPR for AMP, you may do so per-account (gdpr.integration_enabled.amp) or at the host level for the default account (account_defaults.gdpr.integration_enabled.amp).")
+		errs = append(errs, fmt.Errorf("gdpr.amp_exception has been discontinued and must be set to false. If you need to disable GDPR for AMP, you may do so per-account (gdpr.integration_enabled.amp) or at the host level for the default account (account_defaults.gdpr.integration_enabled.amp)"))
```
SyntaxNode, line 217: I'd say the end result is we want hosts to remove the value in their config. I suggest a slight word change to something along the lines of:

"gdpr.amp_exception has been discontinued and must be removed from your config. If you need ..."

SyntaxNode (approved):

privacy/enforcement_test.go
```diff
@@ -264,7 +213,7 @@ func TestApply(t *testing.T) {
 		m.On("ScrubDevice", req.Device, test.expectedDeviceID, test.expectedDeviceIPv4, test.expectedDeviceIPv6, test.expectedDeviceGeo).Return(replacedDevice).Once()
 		m.On("ScrubUser", req.User, test.expectedUser, test.expectedUserGeo).Return(replacedUser).Once()
 
-		test.enforcement.apply(req, test.ampGDPRException, m)
+		test.enforcement.apply(req, m)
```
mansinahar, line 216: Since we're not using `ampGDPRException` any longer, any reason to still keep it as part of the `testCase` struct?

mansinahar (commented):
We probably wanna remove `AmpException()` from the following places as well:
- https://github.com/prebid/prebid-server/blob/master/endpoints/cookie_sync_test.go#L261
- https://github.com/prebid/prebid-server/blob/master/endpoints/setuid_test.go#L444
- https://github.com/prebid/prebid-server/blob/master/endpoints/auction_test.go#L427
- https://github.com/prebid/prebid-server/blob/master/exchange/utils_test.go#L37

privacy/enforcement_test.go
bsardo: Nope, I'll remove. Good catch.

mansinahar (approved):

SyntaxNode (approved):

PR #1566, opened 2020-11-05
Add account CCPA enabled and per-request-type enabled flags

bsardo:
Similar to what we are doing for GDPR, as described in issue #1323 and PR #1564, this adds account-level CCPA configuration.

The order of precedence for determining whether CCPA is enabled is the same as for GDPR:

1. Use account-level request type setting if specified
2. Use account-level general setting if specified
3. Use host-level general setting

### DIFF
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -33,11 +33,7 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
 	if integrationEnabled := a.IntegrationEnabled.GetByIntegrationType(integrationType); integrationEnabled != nil {
 		return integrationEnabled
 	}
-	if a.Enabled != nil {
-		return a.Enabled
-	}
-
-	return nil
+	return a.Enabled
 }
 
 // AccountGDPR represents account-specific GDPR configuration
@@ -53,11 +49,7 @@ func (a *AccountGDPR) EnabledForIntegrationType(integrationType IntegrationType)
 	if integrationEnabled := a.IntegrationEnabled.GetByIntegrationType(integrationType); integrationEnabled != nil {
 		return integrationEnabled
 	}
-	if a.Enabled != nil {
-		return a.Enabled
-	}
-
-	return nil
+	return a.Enabled
 }
 
 // AccountIntegration indicates whether a particular privacy policy (GDPR, CCPA) is enabled for each integration type
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -23,8 +23,8 @@ type Account struct {
 
 // AccountCCPA represents account-specific CCPA configuration
 type AccountCCPA struct {
-	Enabled            *bool                  `mapstructure:"enabled" json:"enabled,omitempty"`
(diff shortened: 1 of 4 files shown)


bsardo:
Manually verified for web, app and amp

hhhjort (approved):
LGTM

config/accounts.go
```diff
@@ -17,9 +17,50 @@ type Account struct {
...
+	}
+	if a.Enabled != nil {
+		return a.Enabled
+	}
+
+	return nil
+}
+
+// AccountCCPAIntegration indicates whether CCPA is enabled for each request type
+type AccountCCPAIntegration struct {
```
SyntaxNode, line 57: Eh. I think `AccountCCPA` makes sense, but I wouldn't mind having just one unified `AccountIntegration`, or some such. The only downside is it becomes harder to add a CCPA or GDPR specific value.. but we could always split the types then.

exchange/utils_test.go
```diff
@@ -93,72 +93,114 @@ func TestCleanOpenRTBRequests(t *testing.T) {
...
 func TestCleanOpenRTBRequestsCCPA(t *testing.T) {
+	trueValue, falseValue := true, false
+
 	testCases := []struct {
 		description         string
 		reqExt              json.RawMessage
 		ccpaConsent         string
-		enforceCCPA         bool
+		ccpaAccountEnabled  *bool
+		ccpaHostEnabled     bool
```
SyntaxNode, line 103: Nitpick: Consider placing host above account because it's the highest level.

hhhjort (dismissed):

config/accounts.go
```diff
@@ -17,32 +17,40 @@ type Account struct {
...
+	IntegrationEnabled AccountIntegration `mapstructure:"integration_enabled" json:"integration_enabled"`
+}
+
+// EnabledForIntegrationType indicates whether CCPA is turned on at the account level for the specified integration type
+// by using the integration type setting if defined or the general CCPA setting if defined; otherwise it returns nil
+func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType) *bool {
+	if integrationEnabled := a.IntegrationEnabled.GetByIntegrationType(integrationType); integrationEnabled != nil {
+		return integrationEnabled
+	}
+	if a.Enabled != nil {
```
SyntaxNode, line 36: Nitpick: You can just return a.Enabled here. Does't matter if it's nil or not. Comment is great and accurate.

config/accounts.go
```diff
@@ -17,32 +17,40 @@ type Account struct {
...
-		integrationEnabled = a.IntegrationEnabled.Video
-	case IntegrationTypeWeb:
-		integrationEnabled = a.IntegrationEnabled.Web
-	}
 
-	if integrationEnabled != nil {
+	if integrationEnabled := a.IntegrationEnabled.GetByIntegrationType(integrationType); integrationEnabled != nil {
 		return integrationEnabled
 	}
 	if a.Enabled != nil {
```
SyntaxNode, line 56: Same nitpick here. Sorry. Seems I missed this the first time around
.

SyntaxNode (approved):

hhhjort (approved):

PR #1557, opened 2020-10-28
33Across: Add video support in adapter

curlyblueeagle:

PR #1567, opened 2020-11-06
Add test coverage

curlyblueeagle:
Based on CR feedback in https://github.com/prebid/prebid-server/pull/1557


curlyblueeagle:
Filed by mistake

PR #1569, opened 2020-11-07
IX: Implement Bidder interface, update endpoint.

ix-prebid-support:


robhazan:
@bsardo - please let us know if anything is required for your review here. We're ready for you guys to merge this into the next release as soon as it's reviewed.

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
-	width       uint64
-	height      uint64
-}
+func (a *IxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	logAsJSON("bid-request: %s", request)
 
-func isValidIXSize(f openrtb.Format, s [2]uint64) bool {
-	if f.W != s[0] || f.H != s[1] {
-		return false
+	nImp := len(request.Imp)
```
SyntaxNode, line 25: You don't need this check. The core framework won't call `MakeRequests` is there are no impressions.

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
 	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
-// maximum number of bid requests
-const requestLimit = 20
+var maxRequests = 20 // it's not a const for the unit test convenience
 
 type IxAdapter struct {
 	http *adapters.HTTPAdapter
```
SyntaxNode, line 18: The `http *adapters.HTTPAdapter` value is not valid for modern bidders. That's a legacy model specific thing. Please remove it.

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
-	return bids, nil
-}
-
-func (a *IxAdapter) callOne(ctx context.Context, reqJSON bytes.Buffer) (ixBidResult, error) {
-	var result ixBidResult
+	// 'adapters/bidder.go': Bidder implementations may safely assume that this
+	// JSON has been validated by their static/bidder-params/{bidder}.json file.
+	var bidderExt adapters.ExtImpBidder
+	var ixExt openrtb_ext.ExtImpIx
+	json.Unmarshal(request.Imp[0].Ext, &bidderExt)
```
SyntaxNode, line 98: It is safe to assume the json is already validation, but you still need to bubble up the errors. There are still other situations where an error can occur.

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
-	Error        error
-}
-
-type callOneObject struct {
-	requestJSON bytes.Buffer
-	width       uint64
-	height      uint64
-}
+func (a *IxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	logAsJSON("bid-request: %s", request)
```
SyntaxNode, line 23: Adapters should not write to the log. It can overwhelm the host's logging backend.

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
+	"github.com/golang/glog"
 	"github.com/mxmCherry/openrtb"
 	"github.com/prebid/prebid-server/adapters"
 	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
-// maximum number of bid requests
-const requestLimit = 20
+var maxRequests = 20 // it's not a const for the unit test convenience
```
SyntaxNode, line 15: Please relocate this variable to the `IxAdapter` to change it for testing. Then you can still have a defaultRequestLimit constant here.

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
-						Message: fmt.Sprintf("Unknown ad unit code '%s'", result.Bid.AdUnitCode),
+		if request.Imp[0].Banner != nil {
+			banner := *request.Imp[0].Banner
+			request.Imp[0].Banner = &banner
+			formats := getBannerFormats(&banner)
+			for iFmt := range formats {
+				banner.Format = formats[iFmt : iFmt+1]
+				banner.W = openrtb.Uint64Ptr(banner.Format[0].W)
+				banner.H = openrtb.Uint64Ptr(banner.Format[0].H)
+				if requestData, err := createRequestData(a, request, &headers); err == nil {
```
SyntaxNode, line 61: It doesn't look like `createRequestData` errors are sent back to the publisher. Please consider adding the err to the errs slice.

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
+					if iFmt == 0 {
+						requests = append(requests, requestData)
+					} else {
+						multiSizeRequests = append(multiSizeRequests, requestData)
```
SyntaxNode, line 65: Do you think keeping track of a multiSizeRequests counter would be easier than using two separate slices?

adapters/ix/ix.go
```diff
@@ -1,267 +1,189 @@
...
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("JSON parsing error: %v", err),
+		}}
+	}
+
+	// Until the time we support multi-format ad units, we'll use a bid request impression media type
+	// as a bid response bid type. They are linked by the impression id.
+	impMediaType := map[string]openrtb_ext.BidType{}
+	for _, imp := range internalRequest.Imp {
+		if imp.Video != nil {
```
SyntaxNode, line 149: How does your server handle a request for both video and banner?

adapters/ix/ix_test.go
```diff
@@ -1,643 +1,19 @@
...
-
-	if err != nil {
-		t.Fatalf("Should not have gotten an error: %v", err)
-	}
-
-	if len(bids) != requestLimit {
-		t.Fatalf("Should have received %d bid", requestLimit)
-	}
+	maxRequests = 2
+	adapterstest.RunJSONBidderTest(t, "ixtest", NewIxBidder(nil, "http://ib.adnxs.com/openrtb2"))
```
SyntaxNode, line 18: Please use a more obvious test url, such as `http://fakehost/openrtb2` or the like.

SyntaxNode (commented):
Thank you very much for updating the IX bid adapter. I'm fine if you'd like to remove the legacy implementation entirely, but please understand publishers still using the legacy endpoint will no longer have access to IX demand. 

We plan to remove the legacy endpoint soon-ish. We don't have a definite date yet.

adapters/ix/ix.go
ix-prebid-support: I've realized that removing this check will not affect the safety of this function regardless of the core framework guaranties.

adapters/ix/ix.go
ix-prebid-support: I'll be putting back the legacy interface.

SyntaxNode: Sounds good. You can throw it in legacy.go and legacy_test.go to keep it isolated. I think that's what Conversant did.

adapters/ix/ix.go
ix-prebid-support: It'll write to the log only when started with "-v 3". I saw other adapters doing  that.

adapters/ix/ix.go
ix-prebid-support: Our server will respond with banner. Fixed.

adapters/ix/ix.go
ix-prebid-support: `json.Marshal(*openrtb.BidRequest)` can never fail:
https://github.com/golang/go/blob/master/src/encoding/json/encode.go

 That's why everyone has a problem with the unit test coverage of the "failed" branch.

adapters/ix/ix.go
ix-prebid-support: I'm using two slices as a simple way to implement the logic described in the comments. If `createRequestData` could return an error it would be complicated to create one continuous slice in one loop.

ix-prebid-support:
@SyntaxNode Thank you very much for your review and comments.
One thing we need to agree on: is it okay to write to the log when prebid-server is started with "-v 3"?

One more thing. We've decided to put the legacy adapter code back. Would you prefer it as a separate commit?
Yuri.

adapters/ix/ix.go
SyntaxNode: I don't see where in the documentation it says `json.Marshal` can never fail.

It can fail for a few reasons, such as if the struct contains unmarshallable types (chan, func pointer, etc...), if there are cyclical references, or if there is memory corruption. The latter is what I'm more concerned with happening as we've seen it before.

We understand it is not feasible to reproduce a `json.Marshal` error for tests and we don't comment on lack of coverage for those parts.

adapters/ix/ix.go
SyntaxNode: I discussed this with the team. We don't want adapters to write the application log, as it could quickly escalate. The v level does help with that, but we don't see why it's necessary. PBS already has debug functionality to return this exact value when running an auction.

You're right. There are other adapters writing to the log. Most are during application startup and is resolved with https://github.com/prebid/prebid-server/pull/1532. The others though which we haven't noticed (Adhese, Gamoshi, Pubmatic, and Rubicon Legacy) and will work with them to remove those calls.

adapters/ix/ix.go
ix-prebid-support: What I meant is not that `json.Marshal` can never fail (it can fail for the reasons you've mentioned), but that `json.Marshal(*openrtb.BidRequest)` can never fail because it doesn't have these cases. If we have a memory corruption it's a much bigger problem that can't be solved with an err flag.

ix-prebid-support:
@SyntaxNode We have a deployment freeze until January 4, 2021. Our senior management asked us not to deploy our new adapter code until the freeze is over. Please hold off merging this PR until then. Thank you.

ix-prebid-support:
@SyntaxNode Did you have a chance to review my new commits to see if I've addressed all of your concerns? I understand that with your recent refactoring I'll have to rebase my work, but it'll be mostly mechanical.

adapters/ix/ix.go
```diff
@@ -247,23 +236,183 @@ func (a *IxAdapter) callOne(ctx context.Context, reqJSON bytes.Buffer) (ixBidRes
...
+				Bid:     &bid,
+				BidType: bidType,
+			})
+		}
+	}
+
+	return bidderResponse, errs
+}
+
+func NewIxAdapter(config *adapters.HTTPAdapterConfig, endpoint string) *IxAdapter {
```
SyntaxNode, line 407: We have since merged in https://github.com/prebid/prebid-server/pull/1532 which conflicts with this method. We now have new `Builder` methods with an updated json test method template for OpenRTB adapters. Should be straight forward to resolve, either by looking at other adapters or checking out the [very recently update bid adapter docs](https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html).

config/config.go
```diff
@@ -967,7 +967,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.gumgum.endpoint", "https://g2.gumgum.com/providers/prbds2s/bid")
 	v.SetDefault("adapters.improvedigital.endpoint", "http://ad.360yield.com/pbs")
 	v.SetDefault("adapters.inmobi.endpoint", "https://api.w.inmobi.com/showad/openrtb/bidder/prebid")
-	v.SetDefault("adapters.ix.endpoint", "http://appnexus-us-east.lb.indexww.com/transbidder?p=184932")
+	v.SetDefault("adapters.ix.endpoint", "http://exchange.indexww.com/pbs?p=192919")
```
SyntaxNode, line 970: I'd just like to verify the `192919` does not uniquely identify AppNexus, as many other hosts run PBS.

adapters/ix/ix.go
```diff
@@ -247,23 +236,183 @@ func (a *IxAdapter) callOne(ctx context.Context, reqJSON bytes.Buffer) (ixBidRes
...
+
+	// Until the time we support multi-format ad units, we'll use a bid request impression media type
+	// as a bid response bid type. They are linked by the impression id.
+	impMediaType := map[string]openrtb_ext.BidType{}
+	for _, imp := range internalRequest.Imp {
+		if imp.Banner != nil {
+			impMediaType[imp.ID] = openrtb_ext.BidTypeBanner
+		} else if imp.Video != nil {
+			impMediaType[imp.ID] = openrtb_ext.BidTypeVideo
+		}
```
SyntaxNode, line 383: Your yaml file indicates support for audio and native formats as well. Would you like add detection for those here too?

static/bidder-info/ix.yaml
```diff
@@ -1,6 +1,15 @@
...
+      - banner
+      - video
+      - native
+      - audio
   site:
     mediaTypes:
       - banner
+      - video
+      - native
+      - audio
```
SyntaxNode, line 15: Could you please add race condition tests for native and audio formats?

SyntaxNode (commented):
I think it looks great. I commend you on the comprehensive code coverage. I have just a few relatively minor comments.

I would still recommend to return an error if `createRequestData` fails to be consistent with how you handle `setSitePublisherId` errors. This will provide the publisher with information about why their impression was dropped.

I'm going to hold off on giving an official approval to ensure this is not merged in before the end of your code freeze, as you requested.

config/config.go
ix-prebid-support: No, it does not.

adapters/ix/ix.go
```diff
@@ -247,23 +236,183 @@ func (a *IxAdapter) callOne(ctx context.Context, reqJSON bytes.Buffer) (ixBidRes
...
+		return err
+	}
+
+	request.Site.Publisher.ID = ixExt.SiteId
+	return nil
+}
+
+func getBannerFormats(banner *openrtb.Banner) []openrtb.Format {
+	if len(banner.Format) == 0 && banner.W != nil && banner.H != nil {
+		banner.Format = []openrtb.Format{{W: *banner.W, H: *banner.H}}
```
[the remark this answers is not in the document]
ix-prebid-support: In this particular case it would work, but I prefer to fix the inconsistency so that it'll work in all cases. It's just a one pointer assignment.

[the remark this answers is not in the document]
SyntaxNode: Alright.

ix-prebid-support:
@SyntaxNode Happy New Year! We are ready.

adapters/ix/ix.go
```diff
@@ -99,29 +82,35 @@ func (a *IxAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder *pbs.P
...
-			// Not supporting multi-media-type adunit yet
+			// Only grab this ad unit. Not supporting multi-media-type adunit yet.
 			thisImp := indexReqImp[i]
 
 			thisImp.TagID = unit.Code
-			thisImp.Banner.Format = []openrtb.Format{format}
-			thisImp.Banner.W = &format.W
-			thisImp.Banner.H = &format.H
+			if thisImp.Banner != nil {
+				thisImp.Banner.Format = []openrtb.Format{format}
```
SyntaxNode, line 90: Thank you for adding the nil check. Although we plan to remove the legacy code later this year, it's nice to fix the "broken windows" until then. In our newer adapters we would ask for the banner object to be shallow copied first due to pointer sharing, but this object is created by `MakeOpenRTBGeneric` earlier in this adapter so I expect we're fine.

SyntaxNode:
Looks fantastic. Thank you very much.

bsardo (approved):
LGTM

PR #1570, opened 2020-11-09
Refactored HoldAuction Arguments

SyntaxNode:


exchange/exchange.go
```diff
@@ -200,7 +210,32 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
 	// Build the response
-	return e.buildBidResponse(ctx, liveAdapters, adapterBids, bidRequest, adapterExtra, auc, bidResponseExt, cacheInstructions.returnCreative, errs)
+	return e.buildBidResponse(ctx, liveAdapters, adapterBids, r.BidRequest, adapterExtra, auc, bidResponseExt, cacheInstructions.returnCreative, errs)
+}
+
+func buildLabels(r AuctionRequest) pbsmetrics.Labels {
+	labels := pbsmetrics.Labels{
+		RType:         r.RequestType,
+		PubID:         r.Account.ID,
+		Browser:       pbsmetrics.BrowserOther, // Should we just remove this entirely? we don't have an 'unknown' option.
```
SyntaxNode, line 220: We do in https://github.com/prebid/prebid-server/pull/1571

laurb9:
Nice, and long overdue :) This should fix the signature and help avoid having to update dozens of unit tests with new parameters, which I'd have to do for #1015 . This object can be passed down to the various MakeBid* methods.

PBS-Java has a similar object called [AuctionContext](https://github.com/rubicon-project/prebid-server-java/blob/b7698bef12612de7c360734196d2f84c4882e93e/src/main/java/org/prebid/server/auction/model/AuctionContext.java#L21) whose naming I find more intuitive save for the fact that `context` has a specific meaning in Go.

router/router.go
```diff
@@ -242,21 +243,21 @@ func New(cfg *config.Configuration, rateConvertor *currencies.RateConverter) (r
 
 	exchanges = newExchangeMap(cfg)
 	cacheClient := pbc.NewClient(cacheHttpClient, &cfg.CacheURL, &cfg.ExtCacheURL, r.MetricsEngine)
-	theExchange := exchange.NewExchange(generalHttpClient, cacheClient, cfg, r.MetricsEngine, bidderInfos, gdprPerms, rateConvertor)
+	theExchange := exchange.NewExchange(generalHttpClient, cacheClient, cfg, r.MetricsEngine, bidderInfos, gdprPerms, rateConvertor, categoriesFetcher)
```
laurb9, line 246: Maybe accounts fetcher could also go in here, since it's not endpoint-specific either.

exchange/exchange.go
```diff
@@ -108,21 +117,22 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
+	recordImpMetrics(r.BidRequest, e.me)
 
 	// Make our best guess if GDPR applies
-	usersyncIfAmbiguous := e.parseUsersyncIfAmbiguous(bidRequest)
+	usersyncIfAmbiguous := e.parseUsersyncIfAmbiguous(r.BidRequest)
 
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
+	labels := buildLabels(r)
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
-	cleanRequests, aliases, privacyLabels, errs := cleanOpenRTBRequests(ctx, bidRequest, requestExt, usersyncs, blabels, labels, e.gDPR, usersyncIfAmbiguous, e.privacyConfig, account)
```
hhhjort, line 125: This is rather curious. I see we have a new line defining `labels` here, yet I do not see an old line defining `labels` that this new line is replacing. That would seem to indicate that `labels` is new to this PR, but the old line just above has it as an argument to `cleanOpenRTBRequests()`.  I must be missing something.

hhhjort: Ahh, found it. It is being passed in the arguments from the endpoint, and the endpoint also builds the labels. So it seems we are tossing out that work and repeating it here. Should we just get rid of this new `buildLables()` line?

SyntaxNode:
> Nice, and long overdue :) This should fix the signature and help avoid having to update dozens of unit tests with new parameters, which I'd have to do for #1015 . This object can be passed down to the various MakeBid* methods.
> 
> PBS-Java has a similar object called [AuctionContext](https://github.com/rubicon-project/prebid-server-java/blob/b7698bef12612de7c360734196d2f84c4882e93e/src/main/java/org/prebid/server/auction/model/AuctionContext.java#L21) whose naming I find more intuitive save for the fact that `context` has a specific meaning in Go.

Indeed. We're in the process of refactoring the entire auction code path. I'd like to wait until we're happy with the restructured code before making changes to the adapter `MakeBid` interface. That's very expensive to change.

I would expect the name of the `AuctionRequest` to evolve over time. You're right. The term 'context' has other meaning in Go's HTTP handlers. I also think the term 'context' is too generic and equivalent to 'stuff' or 'thing'. Make sense in some cases, but I'd like to be more specific with this object.

exchange/exchange.go
SyntaxNode: I don't like the concept of the label structs. The code is organizing core information around the metrics system, whereas the reverse makes sense IMHO. Can't change it all overnight. We had discussed this as a first step in the metrics refactoring.

hhhjort: We can move it here as well, if that is the direction we want to move. It just doesn't make sense to do the same work twice. I thought perhaps you started to move it, and then realized that it was still needed at the higher level, and forgot to remove this code when the plan fell through.

router/router.go
SyntaxNode: Perhaps. Let's explore that in a follow-up PR. The endpoints are each performing the account lookup right now with some, such as AMP, doing non-trivial things. I started on the refactor for this PR but it got a bit complicated.

exchange/exchange.go
SyntaxNode: Alright. I removed `buildLabels` and pass down the labels from the endpoints until we can factor it away entirely.

exchange/exchange.go
```diff
@@ -200,7 +210,32 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
...
 
 	// Build the response
-	return e.buildBidResponse(ctx, liveAdapters, adapterBids, bidRequest, adapterExtra, auc, bidResponseExt, cacheInstructions.returnCreative, errs)
+	return e.buildBidResponse(ctx, liveAdapters, adapterBids, r.BidRequest, adapterExtra, auc, bidResponseExt, cacheInstructions.returnCreative, errs)
+}
+
+func buildLabels(r AuctionRequest) pbsmetrics.Labels {
+	labels := pbsmetrics.Labels{
+		RType:         r.RequestType,
+		PubID:         r.Account.ID,
```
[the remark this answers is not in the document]
SyntaxNode: This code is gone now, but `Account.ID` is set to the result of the Pub.ID/ParentAccount checks done by the endpoints.

laurb9:
> I'd like to wait until we're happy with the restructured code before making changes to the adapter `MakeBid` interface. That's very expensive to change.

My bad, I was thinking of the bid response builders (`makeExtBidResponse`, `buildBidResponse`, `makeSeatBid`), and come to think of it, also `newAuction`. As far as bidders, I assumed they shouldn't have access to the original request anyway.

hhhjort (approved):
LGTM

mansinahar (approved):

Issue #1301, opened 2020-05-14
Possibly Incorrect GDPR Vendor IDs


SyntaxNode:
I went through all adapters to check on the Vendor IDs for possible mistakes or omissions, as a follow-up to https://github.com/prebid/prebid-server/issues/1299.

**Adapters With No Vendor ID**
- Adtelligent: Looks like they might be vendor id 410.
- Beachfront: Looks like they might be vendor id 335.
- ePlanning: Looks like they might be vendor id 90.

**Adapters With Different GVL Names**
It's very possible these are correct business relationships. I'm just looking to verify these are intentional.
- Adpone uses vendor id 16, which is listed as RTB House.
- Advangelists uses vendor id 61, which is listed as Gum Gum.
- Datablocks uses vendor id 14, which is listed as Adverline.
- Engagebdr uses vendor id 62, which is listed as Justpremium BV.

bretg:
I've sent email to the contact addresses for these 7 bid adapters. Just to make sure everyone knows what we're referring to, SyntaxNode confirmed all the adapters against the IAB's 'Global Vendor List' at https://vendorlist.consensu.org/v2/vendor-list.json

trchandraprakash:
@bretg thanks for reporting this issue, we will update adapter soon.

muncha:
@bretg thanks. I'm looking into Beachfront.

SyntaxNode:
Thank yo @muncha. I updated the description to indicate Beachfront has addressed this concern.

SyntaxNode:
We haven't had any movement on this recently. @bretg, could you please send a follow-up email on behalf of PrebidOrg to these adapters?

bretg:
- Adtelligent - @aradilov - please confirm Adtelligent is GVL ID 410
- AdPone - @seergiioo6 - found from the latest list that AdPone is GVL ID 799. Please update your usersync.go
- ePlanning - @Nistenf , @chino117 - please confirm ePlanning is GVL ID 90
- Datablocks - @htang555 - if Datablocks doesn't have a GVL ID, please remove the line of code in your usersync.go -- we assume this was a copy-paste error
- Engagebdr - @yannaingkyaw1986 - if EngageBdr doesn't have a GVL ID, please remove the line of code in your usersync.go -- we assume this was a copy-paste error

htang555:
Hey @bretg here is the PR:

https://github.com/prebid/prebid-server/pull/1572

GeneGenie:
[...]

PR #1572, opened 2020-11-09
update Datablocks usersync.go

htang555:
Removing the GVL id for Datablocks as per:

https://github.com/prebid/prebid-server/issues/1301


SyntaxNode (approved):

hhhjort (approved):
LGTM

PR #1573, opened 2020-11-10
adform bidder video bid response support

braizhas:
docs: https://github.com/prebid/prebid.github.io/pull/2488


guscarreon:
@braizhas thank you for submitting this PR. I've been looking at it and looks pretty good so far. One question though: Are you looking to also add support for Long Form Video / Instream Video with this PR?

braizhas:
No, Adform currently does not support long-form video. Is there a need in additional validation?

mansinahar:
@braizhas Looks like Adform adapter is supported on both the legacy endpoint `/auction` and the openrtb endpoint `/openrtb2/auction` and this PR adds video support for only the latter so wanna confirm if that's intentional?

Other than that, I'd suggest please add some JSON tests as well for video requests, multi-imp requests with both video and banner and multi-format requests with single imp requesting both video and banner, for better test coverage. Please also add `adform/adformtest/params/race/video.json`

mansinahar:
Please also update the documentation for your adapter to reflect support for video: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/adformPbs.md

braizhas:
@mansinahar, somehow thought that legacy auction endpoint is going to be deprecated, so there is no point in updating it. But maybe it's better to keep the same functionality unless it's really deprecated.

guscarreon (dismissed):
LGTM

adapters/adform/adformtest/exemplary/multiformat-impression.json
```diff
@@ -0,0 +1,99 @@
...
+                "deal_id": "DID-123-22",
+                "win_crid": "20078831"
+              }
+          ]
+        }
+      }
+    ],
+    "expectedBidResponses": [
+      {
+        "currency": "EUR",
```
mansinahar, line 69: Shouldn't the expected currency here be `USD`? Looks like we don't assert the response currencies in our JSON tests and which is why the tests didn't flag here. I will work to correct that and make sure the tests assert that too but for now, can you please update this to right expected value? Thank you!

adapters/adform/adformtest/exemplary/single-banner-impression.json
```diff
@@ -0,0 +1,64 @@
...
+            "deal_id": null,
+            "win_crid": "20078830"
+          }
+        ]
+      }
+    }
+  ],
+  "expectedBidResponses": [
+    {
+      "currency": "EUR",
```
mansinahar, line 47: Same as above for the currency value

adapters/adform/adformtest/exemplary/single-video-impression.json
```diff
@@ -0,0 +1,60 @@
...
+            "deal_id": null,
+            "win_crid": "20078830"
+          }
+        ]
+      }
+    }
+  ],
+  "expectedBidResponses": [
+    {
+      "currency": "EUR",
```
mansinahar, line 43: Same as above for the currency value

adapters/adform/adformtest/exemplary/multiformat-impression.json
braizhas: Yeah, you are right, copy/pasting mistake.

braizhas:
@mansinahar, thanks for a thorough review.

mansinahar:
@braizhas Of course! Happy to! Thanks for addressing all the comments. LGTM now :)

mansinahar (approved):

guscarreon (approved):
LGTM

PR #1579, opened 2020-11-11
DMX Bidfloor fix

stevealliance:
added bidfloor from original JSON request


guscarreon (dismissed):
Addition looks great.

adapters/dmx/dmx.go
```diff
@@ -221,21 +222,40 @@ func (adapter *DmxAdapter) MakeBids(request *openrtb.BidRequest, externalRequest
...
 func fetchParams(params dmxExt, inst openrtb.Imp, ins openrtb.Imp, imps []openrtb.Imp, banner *openrtb.Banner, video *openrtb.Video, intVal int8) []openrtb.Imp {
+
 	if params.Bidder.TagId != "" {
 		ins = openrtb.Imp{
-			ID:     inst.ID,
-			TagID:  params.Bidder.TagId,
-			Ext:    inst.Ext,
-			Secure: &intVal,
+			ID:                inst.ID,
+			TagID:             params.Bidder.TagId,
```
mansinahar, line 229: Wouldn't it be easier to just create a copy of `inst` as `ins` and then manipulate the fields that you need to manipulate? And you can do that outside of these two if conditions as everything else except the `tagID` value is the same in either case.

mansinahar (commented):
Please add a JSON test to make sure these fields are propagated to your service as expected

adapters/dmx/dmx.go
stevealliance: @mansinahar Added test json for imp other expected value for banner and video

guscarreon (approved):
LGTM

mansinahar (approved):

Issue #1470, opened 2020-09-02
Prebid Server Event Support - server-side analytics


bretg:
The next revision of the Prebid Events design is at https://docs.google.com/document/d/1ry0X4C2EV-R0pMrm1IQk9BstxaT395UCl3KKqTGa5c8/edit#

The driver for this round of updates is the need to have server-side-only tracking of auctions for the web 'channel'. In the previous iteration, the assumption was that Prebid.js analytics adapters would be responsible for auction events and most bidsWon events.

However, there are two use cases where that's not enough:

- Prebid Server host company doesn't have an analytics adapter on the page
- The server-side analytics system is different from the client-side analytics.

The changes to the doc are:

- Added scenarios for Web/Banner/Server-Side-with-PBS-Analytics and Web/Instream/Server-Side-with-PBS-Analytics
- Fleshed out all the flow diagrams with numbered descriptions
- Added the “int” (integration type) parameter to event URLs (related to https://github.com/prebid/prebid-server/issues/1428)
- Refined the Prebid Server Event Algorithm

This issue tracks the review and implementation of the changes to the Prebid Server Event Algorithm

Specifically, here's what changed:
- New account+channel analytics configuration. e.g account 111 gets AMP/App events, but not web events. Account 222 gets events for all 3 channels: web/AMP/App.
- Server-side analytics adapters need access to this new account-level event config so they can determine what to do.
- Clarified the algorithm around when and how to create the 3 different event URLs. Includes how to deal with Programmatic Guaranteed, which is out of scope for PBS-Go.

bretg:
As part of this improvement, we realized that the definition of "channel" is quite important to the proper functioning of events and that it really needs to be consistent between PBS-Core that's dropping bidsWon/imp URLs into responses and the analytics adapter.

However in https://github.com/prebid/prebid-server/issues/1428 we defined PBS-core to only set AMP and App. But  as defined in the [PBS event algorithm](https://docs.google.com/document/d/1ry0X4C2EV-R0pMrm1IQk9BstxaT395UCl3KKqTGa5c8/edit#), that's not enough to trigger bidsWon for "web". So we need to extend what core does to determine channel. Options:

1) support a configurable default channel (e.g. web)
2) can we assume that the existence of a "site" object is always channel "web"?
3) make channel determination very configurable (e.g. regex map in config)

Another aspect of the account/channel configuration needs to be confirmed as well:
[...]

PR #1584, opened 2020-11-13
Add timestamp to analytics and response.ext.prebid.auctiontimestamp l…

laurb9:
…estamp (like PBS-Java)

In addition to analytics, this timestamp will be used later for generating event urls for #1470 and #1015 

```json
  "ext": {
    "prebid": {
      "auctiontimestamp": 1605231385254
    }
  }
```

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -91,18 +91,17 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	ao := analytics.AmpObject{
 		Status: http.StatusOK,
 		Errors: make([]error, 0),
+		// Prebid Server interprets request.tmax to be the maximum amount of time that a caller is willing
+		// to wait for bids. However, tmax may be defined in the Stored Request data.
+		//
+		// If so, then the trip to the backend might use a significant amount of this time.
+		// We can respect timeouts more accurately if we note the *real* start time, and use it
+		// to compute the auction timeout.
+		Timestamp: time.Now(),
 	}
-
-	// Prebid Server interprets request.tmax to be the maximum amount of time that a caller is willing
-	// to wait for bids. However, tmax may be defined in the Stored Request data.
-	//
-	// If so, then the trip to the backend might use a significant amount of this time.
-	// We can respect timeouts more accurately if we note the *real* start time, and use it
-	// to compute the auction timeout.
 
 	// Set this as an AMP request in Metrics.
 
-	start := time.Now()
 	labels := pbsmetrics.Labels{
 		Source:        pbsmetrics.DemandWeb,
 		RType:         pbsmetrics.ReqTypeAMP,
@@ -113,7 +112,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	}
 	defer func() {
 		deps.metricsEngine.RecordRequest(labels)
-		deps.metricsEngine.RecordRequestTime(labels, time.Since(start))
+		deps.metricsEngine.RecordRequestTime(labels, time.Since(ao.Timestamp))
 		deps.analytics.LogAmpObject(&ao)
 	}()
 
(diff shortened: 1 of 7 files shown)


laurb9:
This should probably wait for #1570

exchange/exchange.go
```diff
@@ -176,6 +176,11 @@ func (e *exchange) HoldAuction(ctx context.Context, bidRequest *openrtb.BidReque
 	}
 
 	bidResponseExt := e.makeExtBidResponse(adapterBids, adapterExtra, bidRequest, debugInfo, errs)
+	if start != nil {
+		bidResponseExt.Prebid = &openrtb_ext.ExtResponsePrebid{
+			AuctionTimestamp: start.UnixNano() / 1e+6,
+		}
+	}
```
guscarreon, line 183: Consider moving lines 179 to 183 inside `makeExtBidResponse`.

guscarreon (changes requested):
Looks pretty good, one quick comment

exchange/exchange.go
laurb9: I agree this code belongs in `makeExtBidResponse`. I wanted to avoid changing more signatures with the associated impact on unit tests... I'll take another look after #1570 settles.

guscarreon: Hi @laurb9 friendly reminder that #1570 merged. No hurry though

exchange/exchange.go
```diff
@@ -102,6 +102,7 @@ type AuctionRequest struct {
 	Account     config.Account
 	UserSyncs   IdFetcher
 	RequestType pbsmetrics.RequestType
+	Timestamp   *time.Time
```
SyntaxNode, line 105: Yay! This is proving useful already.

1. Do you think naming it 'StartTime' would be an improvement?
2. Is there a particular reason you're making this a pointer? Doesn't seem necessary IMHO.

laurb9: The naming was after the json attribute `auctiontimestamp` and event parameter `ts` where this would be used. I agree that `ReceivedTime` is a more apt description. Should I change it both here and in analytics ?

As for the pointer - I didn't want a copy with the source of truth elsewhere. Incidentally, the nil test also avoided breaking all the tests by returning `auctiontimestamp: 0`. It's also a minor size optimization, and I was thinking maybe Account should also be a reference.

laurb9: Hmm, the docs for time.Time say
> Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time.
so I will change it.

SyntaxNode: > I agree that ReceivedTime is a more apt description (edit: but StartTime sounds better).

I was on the fence there too.

> Should I change it both here and in analytics ?

Yes please. I think we should try to keep consistent. To be clear though, let's keep the "auctiontimestamp" json name for compatibility with PBS-Java.

> It's also a minor size optimization

In Go, a pointer forces a heap allocation whereas a value will avoid it and use the stack. At some point the value copy operation can take a long time, but for smaller objects pass by value is typically preferred. A notable exception being JSON serialization which forces the use of a pointer to determine default vs not specified... and that bugs me to no end. Grrr.

mansinahar (commented):
LGTM in general except for the one comment. Thanks @laurb9 as usual for your contributions :)

endpoints/openrtb2/amp_auction.go
```diff
@@ -91,18 +91,17 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	ao := analytics.AmpObject{
 		Status: http.StatusOK,
 		Errors: make([]error, 0),
+		// Prebid Server interprets request.tmax to be the maximum amount of time that a caller is willing
+		// to wait for bids. However, tmax may be defined in the Stored Request data.
+		//
+		// If so, then the trip to the backend might use a significant amount of this time.
+		// We can respect timeouts more accurately if we note the *real* start time, and use it
+		// to compute the auction timeout.
+		StartTime: time.Now(),
```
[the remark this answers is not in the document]
laurb9: I thought that the standalone `start` variable became redundant and just takes stack space once the same value is in the ao object. I did not consider `ao` a dependency as it is not passed down.

I will add the startTime back.

mansinahar (approved):

endpoints/openrtb2/amp_auction.go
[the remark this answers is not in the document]
mansinahar: Yeah, I agree that currently it's not being passed as a dependency downstream but we are looking at refactoring this code and when we do (it might take a while to be honest), this might change and even in general defining the start time should be independent of the analytics object.

guscarreon (approved):
LGTM

PR #1585, opened 2020-11-14
New Adapter: Mobfox

mobfxoHB:
https://github.com/prebid/prebid.github.io/pull/2501


guscarreon (commented):
Thank you for submitting your adapter @mobfxoHB . After a quick review, I realize your adapter's source code is almost identical to other adapters. (Particualrly the Adprime adapter). Did you intend to create an alias instead of an entire new adapter?

If you are looking for the exact same functionality as other adapter you previously coded and you only mean to change the bidder name, the bidding server URL, and/or the sync URL, we can probably just create an alias and call it a day. 

Please let me know your thoughts.

mobfxoHB:
@guscarreon  The Adprime adapter was not created by us and it is not related to our company. I have reviewed multiple adapters and realised that their implementation is pretty similar to what is required, so I would like to create new one that might look related to already existing ones. Please let me know how we should proceed.

guscarreon (commented):
> The Adprime adapter was not created by us and it is not related to our company. I have reviewed multiple adapters and realised that their implementation is pretty similar to what is required, so I would like to create new one that might look related to already existing ones. Please let me know how we should proceed.

Awesome @mobfxoHB. Your adapter looks pretty good. Just one small nitpick from my end: Could we format all JSON files with a 2 space tabs? Some files seem already formatted, but others aren't.

static/bidder-params/mobfoxpb.json
```diff
@@ -0,0 +1,14 @@
...
+
+    "type": "object",
+    "properties": {
+      "TagID": {
+        "type": "string",
+        "description": "An ID which identifies the mobfox ad tag"
+      }
+    },
+    "required" : [ "TagID" ]
+  }
```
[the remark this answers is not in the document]
mobfxoHB: Done

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,137 @@
...
+			adapterRequests = append(adapterRequests, adapterReq)
+		}
+		errs = append(errs, errors...)
+	}
+	return adapterRequests, errs
+}
+
+func (a *MobfoxpbAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
+
+	var errs []error
```
mansinahar, line 57: Does this method really needs to return an error slice? The only place it potentially returns an error is from the JSON Marshal and that's a single error instance. It will be easier to just return `error` rather than a slice of `error`

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type MobfoxpbAdapter struct {
+	URI string
+}
+
+// NewMobfoxpbBidder Initializes the Bidder
+func NewMobfoxpbBidder(endpoint string) *MobfoxpbAdapter {
+	return &MobfoxpbAdapter{
```
mansinahar, line 21: Please consider adding validation of the endpoint URL here before returning the adapter instance. Something like:
```
endpointURL, err := url.ParseRequestURI(endpoint)
if err != nil {
	glog.Fatalf("invalid endpoint provided for Mobfox: %s, error: %v", endpoint, err)
	return nil
}
return &MobfoxpbAdapter{
	URI: endpointURL.String(),
}
```

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,137 @@
...
+
+// MakeBids makes the bids
+func (a *MobfoxpbAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	var errs []error
+
+	if response.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if response.StatusCode == http.StatusNotFound {
```
mansinahar, line 85: Since you're returning the same error message for `StatusNotFound` as when the response status is not `StatusOK`, I'd say this is already covered by the `if` condition below and so you can safely remove this one

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,137 @@
...
+	var bidResp openrtb.BidResponse
+
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(1)
+
+	for _, sb := range bidResp.SeatBid {
+		for i := range sb.Bid {
```
mansinahar, line 106: Please consider changing this for loop to `for _, bid := range sb.Bid` and then use `bid` directly inside the loop. This will simplify accessing it in the loop. For example your next line can then become:
```
bidType, err := getMediaTypeForImp(bid.ImpID, internalRequest.Imp)
```

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return bidResponse, errs
+}
+
+func getMediaTypeForImp(impID string, imps []openrtb.Imp) (openrtb_ext.BidType, error) {
+	mediaType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner == nil && imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
```
mansinahar, line 127: Does this mean that for multi-format requests i.e. requests that would have an imp with both banner and video defined, your server is always guaranteed to return a banner response?

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+			return mediaType, nil
+		}
+	}
+
+	// This shouldnt happen. Lets handle it just incase by returning an error.
+	return "", &errortypes.BadInput{
```
mansinahar, line 134: `BadInput` is a misleading errortype here because `BadInput` suggests the request had something invalid which is not the case here. Please consider using `BadServerResponse` error type as the issue is with the response sent by the server

adapters/mobfoxpb/mobfoxpbtest/params/banner.json
```diff
@@ -0,0 +1,3 @@
+{
+  "TagID": "6"
```
mansinahar, line 2: Only the `params/race/*.json` files are needed. You can safely remove this one. I noticed that a few other adapters do this. I will work on getting them removed but can you please do it for yours? Thanks!

adapters/mobfoxpb/mobfoxpbtest/params/video.json
```diff
@@ -0,0 +1,3 @@
+{
```
mansinahar, line 1: Same as above.

adapters/mobfoxpb/mobfoxpbtest/supplemental/status-204.json
```diff
@@ -0,0 +1,81 @@
...
+            "ifa": "sdjfksdf-dfsds-dsdg-dsgg"
+          }
+        }
+      },
+      "mockResponse": {
+        "status": 204,
+        "body": {}
+      }
+    }
+  ]
```
mansinahar, line 80: Please add an empty `expectedBidResponses` here as that indicates your adapter returns an empty bid response back with no error. We are in the process of doing that for other adapters as well. See #1589

adapters/mobfoxpb/mobfoxpbtest/supplemental/string_imp_ext.json
```diff
@@ -0,0 +1,38 @@
+{
```
mansinahar, line 1: Is this one any different from `imp_ext_string.json`? Perhaps it was duplicated by mistake?

adapters/mobfoxpb/params_test.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+var validParams = []string{
+	`{"TagID": "6"}`,
+}
+
+var invalidParams = []string{
+	`{"id": "123"}`,
+	`{"tagid": "123"}`,
+	`{"TagID": 16}`,
```
mansinahar, line 45: Please consider adding another case with the right key but and empty string value

openrtb_ext/imp_mobfoxpb.go
```diff
@@ -0,0 +1,6 @@
+package openrtb_ext
+
```
mansinahar, line 2: Since you're using JSON parser to parse the `tagID` value, you don't really need this struct here as it's not being used anywhere

usersync/usersyncers/syncer_test.go
```diff
@@ -99,6 +99,7 @@ func TestNewSyncerMap(t *testing.T) {
 		openrtb_ext.BidderInMobi:       true,
 		openrtb_ext.BidderKidoz:        true,
 		openrtb_ext.BidderKubient:      true,
+		openrtb_ext.BidderMobfoxpb:     true,
```
mansinahar, line 102: Just wanna confirm that your adapter doesn't support usersync at this time, correct?

mobfxoHB: Yes, we are currently working on it

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: Not sure how to handle this case properly, CI, when tests prebid-server it also tests adapters with empty config, so if i'm returning nil from adapter end printing error message - testing stops with failed status

mansinahar: @mobfxoHB Can I ask which CI are you referring to here?

adapters/mobfoxpb/mobfoxpb.go
mansinahar: Does that mean your system discards requests that have multiple formats or that you always just consider the `banner` part in that request and return bids just for that?

adapters/mobfoxpb/mobfoxpbtest/supplemental/string_imp_ext.json
mansinahar: Umm.. I see both of them having `ext` set as an empty string.

Ref: https://github.com/prebid/prebid-server/pull/1585/files#diff-068bd5de743e84f0ea42bf24b78266a42b133d1c56746f07f05e29250e53b9c6R20, https://github.com/prebid/prebid-server/pull/1585/files#diff-b3db1ea0f32285de8f539bf8eab4dfce6a8ccbba76a29d79a1fc9756616428caR20

`imp_ext_empty_object.json` and `empty_imp_ext.json` also look similar. They are the ones that have `bid.Ext` set as an empty object rather than an empty string

Ref: https://github.com/prebid/prebid-server/pull/1585/files#diff-fcc47bc4b52242c11553e49496206a1c37641d1c3e62d97bce2c6d130b38499dR20, https://github.com/prebid/prebid-server/pull/1585/files#diff-827dd6469890439efd8f99990038cfd8ca56cb10ac6d886423b5b99d83bb49ceR20

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: Github circle ci, after failed commit - i've reverted changes

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,130 @@
...
+// MakeRequests create bid request for mobfoxpb demand
+func (a *MobfoxpbAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errs []error
+	var err error
+	var tagID string
+
+	var adapterRequests []*adapters.RequestData
+
+	reqCopy := *request
+	for _, imp := range request.Imp {
```
guscarreon, line 36: Some thoughts on the for loop inside `MakeRequests`: I read above that you expect arrays with a single `Imp`. Even if that's always the case, I left some comments below:
```
28   func (a *MobfoxpbAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
29   *--  5 lines: var errs []error--------------------------------------------------------------------------
34  
35       reqCopy := *request
36       for _, imp := range request.Imp {
   +         // No need to manually create a local copy, the foreach statement creates a local copy of the current element
37 -         reqCopy.Imp = []openrtb.Imp{imp}
38  
39 -         tagID, err = jsonparser.GetString(reqCopy.Imp[0].Ext, "bidder", "TagID")
   +         // Unmarshalling the current element is safe
   +         tagID, err = jsonparser.GetString(imp.Ext, "bidder", "TagID")
40           if err != nil {
41               errs = append(errs, err)
42               continue
43           }
44  
45 -         reqCopy.Imp[0].TagID = tagID
   +         // In golang, the foreach statement creates a local copy of the elements of the array
   +         // it is traversing. Therefore, it's safe to overwrite `TagID`
   +         imp.TagID = tagID
   +         // It might be better to move the `Imp` slice creation down here in order to do it only after after we
   +         // made sure `jsonparser.GetString(reqCopy.Imp[0].Ext, "bidder", "TagID")` did not return an error
   +         reqCopy.Imp = []openrtb.Imp{imp}
46  
47           adapterReq, err := a.makeRequest(&reqCopy)
48           if err != nil {
49               errs = append(errs, err)
50           }
51           if adapterReq != nil {
52               adapterRequests = append(adapterRequests, adapterReq)
   +             // Given that we succeeded, we should break here, isn't it? Unless, we want 
   +             // to append more than one "valid" `adapterReq` in the `adapterRequests` array
   +             break
53           }
54       }
55       return adapterRequests, errs
56   }
adapters/mobfoxpb/mobfoxpb.go
```

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: first bid.imp (bid.imp[0]) will be taken into account

adapters/mobfoxpb/mobfoxpbtest/supplemental/string_imp_ext.json
mobfxoHB: Yep, really there was two duplicates

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: Done

adapters/mobfoxpb/mobfoxpb.go
mansinahar: @mobfxoHB That looks like something specific to your CI configuration. Perhaps you can change that on your end? Because in case of invalid endpoints, it is better to validate and fail early than run into issues while processing requests.

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,129 @@
...
+
+		imp.TagID = tagID
+		reqCopy.Imp = []openrtb.Imp{imp}
+		adapterReq, err := a.makeRequest(&reqCopy)
+		if err != nil {
+			errs = append(errs, err)
+		}
+		if adapterReq != nil {
+			adapterRequests = append(adapterRequests, adapterReq)
+			break
```
mansinahar, line 51: Why `break` only after the first imp? Are you sure you want to discard all other imps in the request if there are more than one? If you're bidder only supports 1 imp per request, you can create multiple `adapterRequests` for each imp and they will all be sent as separate requests to your bidder

adapters/mobfoxpb/mobfoxpb.go
mansinahar: Do you mean `request.Imp[0]`? Even that single `imp` can have both `banner` and `video` set for multi-format requests. In such a case if you always only consider the banner part of the imp and bid only on that then this code is fine but if you consider both banner and video and have some logic to determine which one to bid on then you could be sending a video bid back but marking it as `banner` here. I hope that makes sense.

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: For now , there wouldn't be multiple imressions. We will work on multy bidding in next iteration

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: We have no CI on our side - test are beeing done only by this one:
Validate / Go (1.13.x, ubuntu-18.04) (pull_request)
Validate / Go (1.14.x, ubuntu-18.04) (pull_request)
Validate / Go (1.15.x, ubuntu-18.04) (pull_request)
And there , i think can not be invalid URL - we use only default one

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: Something should be set as default, media-type can't be empty, and even in case some one will send us multiformat impression - key, passed in extension will lead to predefined endpoint with predefined media type, so response will have expected type

guscarreon:
@mobfxoHB. Your adapter looks pretty good, but: Starting our next release on December 3rd, adapter source code will need to be written in a slightly different way. Pull request #1532 describes this changes. Therefore, before we merge your adapter please take a look to PR #1532 and modify your adapter accordingly. If you have any questions, we're glad to help.

SyntaxNode:
We also just published a [new adapter doc](https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html) today which will guide you through the new approach.

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,131 @@
+package mobfoxpb
```
SyntaxNode, line 1: Why did you choose `mobfoxpb` instead of simply `mobfox`?

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// MobfoxpbAdapter adapter structure
+type MobfoxpbAdapter struct {
+	URI string
+}
+
+// Builder builds a new instance of the Adman adapter for the given bidder with the given config.
```
SyntaxNode, line 21: Copy/Paste Mistake: Please change Adman to your bidder name.

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/buger/jsonparser"
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// MobfoxpbAdapter adapter structure
+type MobfoxpbAdapter struct {
```
SyntaxNode, line 17: We prefer the name `adapter` here. No need to export it since it's not referenced directly outside of this package. No need to duplicate the mobfox name.

adapters/mobfoxpb/mobfoxpb.go
SyntaxNode: I think that's fine. Please be clear in your docs that only the first imp is used and the others are ignored. I'd also like to see removal of the for loop since you're only looking at the first one. Something like:

```
imp := request.Imp[0]

tagID, err = jsonparser.GetString(imp.Ext, "bidder", "TagID")
if err != nil {
            Message: fmt.Sprintf("unable to read imp[0].ext.bidder.TagID (%v)", err)}}
}

reqCopy := *request
reqCopy.Imp = []openrtb.Imp{imp[0]}

adapterReq, err := a.makeRequest(&reqCopy)
if err != nil {
    return nil, []error{err}
}

return []*adapters.RequestData{adapterRequest}, nil
```

adapters/mobfoxpb/mobfoxpb.go
SyntaxNode: What we're asking for here is to either read the type of the format from your bidding server's response or perfectly duplicate the logic from your server here. If your server gets an imp for both a banner and video, which will it choose?

adapters/mobfoxpb/mobfoxpb_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, _ := Builder(openrtb_ext.BidderMobfoxpb, config.Adapter{
+		Endpoint: "http://example.com/?c=o&m=ortb"})
+	adapterstest.RunJSONBidderTest(t, "mobfoxpbtest", bidder)
+}
```
SyntaxNode, line 15: I understand that it's not possible for your adapter to throw an error, but I'd like to ask for you to still please include the error check for uniformity with other adapter and in case that changes in the future. 

```
	bidder, buildErr := Builder(openrtb_ext. BidderMobfoxpb, config.Adapter{
		Endpoint: "http://example.com/?c=o&m=ortb"})

	if buildErr != nil {
		t.Fatalf("Builder returned unexpected error %v", buildErr)
	}

	adapterstest.RunJSONBidderTest(t, "mobfoxpbtest", bidder)
}
}
```

SyntaxNode (commented):
Nice update. I left a few clarifying comments. What did you think of the new adapter docs?

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: We are not expecting to get multiformat impression. Our backend logic looks something like that:
```
if(bid.banner) return 'banner'
if(bid.video) return 'vide'
```

mobfxoHB: so if imp has banner - that would be banner traffic, even if video there inside bid

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: this comes from Prebid.js adapter, current adapters are made for new solution, but there is still working previous one which integrated through `mobfox` prebid.js adapter, so we decided to give new name to prebid.js and prebid-server adapters

adapters/mobfoxpb/mobfoxpb_test.go
mobfxoHB: done

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: done

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: done

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: I've made changes, which docs do you mean? prebid.github.io dev-docs? maybe there is some formal parameter which shows that adapter excepts multi imp or single one ?

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: actualy the code in adapter do the same

mobfxoHB:
@guscarreon About docs - i think now it's more clear what should be done, but one thing would be helpful - minimal structure of files, which should be added or edited at the beginning of the docs as the reffernce to prevent scrolling back and forth.  About this comment https://github.com/prebid/prebid-server/pull/1585/#discussion_r534364812 , which docs do you mean?  github.io dev-docs ? maybe there is some formal parameter for multi impression support

adapters/mobfoxpb/mobfoxpb.go
SyntaxNode: Gotcha. Sounds good for matching with a Prebid.js bidder name,

SyntaxNode:
> About this comment #1585 , which docs do you mean? github.io dev-docs ? maybe there is some formal parameter for multi impression support

We were referring to the github.io dev-docs. Please include a mention there that your adapter will only forward the first impression to your bidding server and ignore the rest.

I think you already mentioned this, but please consider supporting multiple impressions in the future. Each impression can be broken down to individual requests.

mobfxoHB:
> > About this comment #1585 , which docs do you mean? github.io dev-docs ? maybe there is some formal parameter for multi impression support
> 
> We were referring to the github.io dev-docs. Please include a mention there that your adapter will only forward the first impression to your bidding server and ignore the rest.
> 
> I think you already mentioned this, but please consider supporting multiple impressions in the future. Each impression can be broken down to individual requests.

I've mentioned this in the docs

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,127 @@
...
+func (a *adapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errs []error
+	var err error
+	var tagID string
+
+	var adapterRequests []*adapters.RequestData
+
+	reqCopy := *request
+	imp := request.Imp[0]
+	tagID, err = jsonparser.GetString(imp.Ext, "bidder", "TagID")
```
guscarreon, line 38: Do we expect `tagID` values within a certain range? Should we error out if less or equal to zero?

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner == nil && imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			return mediaType, nil
+		}
+	}
+
+	// This shouldnt happen. Lets handle it just incase by returning an error.
```
guscarreon, line 123: nitpick: "in case it returns an error"

guscarreon (commented):
Two very minor comments

adapters/mobfoxpb/mobfoxpb.go
guscarreon: Given that we are not null-checking `imp.Banner` nor `imp.Video` anywhere else in this adapter code, I believe the truth table below describes the scenarios we may find. Right?
```
  | imp.Banner  | imp.Video   | other (imp.Audio OR imp.Native)  ||  RESULT
--|-------------|-------------|----------------------------------||---------------------------
 1| nil         | nil         | nil                              || openrtb_ext.BidTypeBanner
 2| nil         | nil         | not-nil                          || openrtb_ext.BidTypeBanner
 3| nil         | not-nil     | nil                              || openrtb_ext.BidTypeVideo
 4| nil         | not-nil     | not-nil                          || openrtb_ext.BidTypeVideo
 5| not-nil     | nil         | nil                              || openrtb_ext.BidTypeBanner
 6| not-nil     | nil         | not-nil                          || openrtb_ext.BidTypeBanner
 7| not-nil     | not-nil     | nil                              || openrtb_ext.BidTypeBanner
 8| not-nil     | not-nil     | not-nil                          || openrtb_ext.BidTypeBanner
```

If so, are we ok with this logic? Or should we modify and set to `BidTypeVideo` in scenarios 7, and 8? Should we error out in the somewhat unlikely case scenarios 1 or 2 happen?

adapters/mobfoxpb/mobfoxpb.go
SyntaxNode: Looks like their bidders params json schema is enforcing a min length of 1 already.

mobfxoHB: @guscarreon Yes, we really expecting tag id to be at least one character length , in not always will be a valid number so we do not check numeric ranges

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: I've made some changes to be closer to our backend logic, so for now we check mediatypes in such way (on backend and in pbs adapter) first we check banner, if no - we check video, if no - we check native, at least one media type will be found, if we recieve unsuported media type - bid would be rejected.  About #1 - pbs will reject any request without at least one media type.

guscarreon (approved):
Some other adapters throw an `errortypes.BadInput` error when `resp.StatusCode == http.StatusBadRequest`. While desirable, it is by no means required and I believe I'm ready to approve. Thank you for addressing our feedback.
```
 74   func (a *adapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
 75       var errs []error
 76  
 77       if response.StatusCode == http.StatusNoContent {
 78           return nil, nil
 79       }
    +
    +     if resp.StatusCode == http.StatusBadRequest {
    +         return nil, []error{&errortypes.BadInput{"bad request"}}
    +     }
 80  
 81       if response.StatusCode != http.StatusOK {
 82           return nil, []error{&errortypes.BadServerResponse{
 83               Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
 84           }}
 85       }
 86  
 87       var bidResp openrtb.BidResponse
 adapters/mobfoxpb/mobfoxpb.go
```

mobfxoHB:
@guscarreon Thank you for your patience, I think we would live adapter without any changes for now and add more precise status codes handling and multi bidding in next iteration

openrtb_ext/imp_mobfoxpb.go
```diff
@@ -0,0 +1,4 @@
...
+// ExtImpMobfoxpb defines mobfox specifiec param
```
mansinahar, line 3: Can you please remove this struct definition as it is not being used anywhere?

mansinahar (commented):
LGTM except for the comment below. Can you please address that and I should be good to approve

openrtb_ext/imp_mobfoxpb.go
mobfxoHB: done

mansinahar (approved):

config/config.go
```diff
@@ -970,7 +970,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.lunamedia.endpoint", "http://api.lunamedia.io/xp/get?pubid={{.PublisherID}}")
 	v.SetDefault("adapters.marsmedia.endpoint", "https://bid306.rtbsrv.com/bidder/?bid=f3xtet")
 	v.SetDefault("adapters.mgid.endpoint", "https://prebid.mgid.com/prebid/")
-	v.SetDefault("adapters.mobilefuse.endpoint", "http://mfx.mobilefuse.com/openrtb?pub_id={{.PublisherID}}")
+	v.SetDefault("adapters.mobfoxpb.endpoint", "http://bes.mobfox.com/?c=o&m=ortb")
+	v.SetDefault("adapters.mobilefuse.endpoint", "http://mfx-us-east.mobilefuse.com/openrtb?pub_id={{.PublisherID}}")
```
SyntaxNode, line 974: Looks like a merge error. Can you please fix the conflict such that this PR isn't affecting mobilefuse's endpoint?

static/bidder-params/mobfoxpb.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Mobfox Adapter Params",
+  "description": "A schema which validates params accepted by the Mobfox adapter",
+  "type": "object",
+  "properties": {
+    "TagID": {
```
SyntaxNode, line 7: Are you sure you want to use `TagID` instead of `tagID` or `tagid`? It's common in JavaScript/JSON to start with a lowercase letter and it's common in OpenRTB to use all lowercase letters. Either is fine in my view, as I see this is decision owned by the adapter.

Just be aware that changes to this after merging are disastrous to publishers and should be avoided.

SyntaxNode (commented):
Looks good to me too. I noticed a merge conflict which needs to address before we can merge.

I also left a warning of using a non-standard naming convention for the bidder parameters. You can proceed as is if you wish, but wanted to provide a few more notes before doing so.

static/bidder-params/mobfoxpb.json
mobfxoHB: This made to show that  this property and property from bidder extension are the same, we have no experienced golang developer, so this brings little better understanding. Let's left it untouched

SyntaxNode: The name here doesn't have to match the field name on the programming language. If you'd like to keep it TagID, then sure. Would you want to name it `placementId` to match with your existing PBJS adapter? A lot publishers prefer to have the same variables on both PBS and PBJS so they can easily retarget, and using different parameter names messes with that a bit.

config/config.go
mansinahar: Looks like you still need to address @SyntaxNode's comment here ^^

static/bidder-params/mobfoxpb.json
mobfxoHB: Let's keep it with current name

config/config.go
mobfxoHB: @mansinahar @SyntaxNode  I've checked - github doesn't show merge conflicts for PR, also, to be sure, that's it's really so - i've cloned prebid-server repo, added mobfox repo as remote upstream and merged locally upstream/master of mobfox to prebid-server master successfully

SyntaxNode: @mobfxoHB Thank you for checking. I'll be more precise with my wording. You had a merge conflict earlier and resolved it incorrectly. You can see here that your PR changes the "adapters.mobilefuse.endpoint" line and it should not. If we were to merge this now, the mobilefuse adapter would stop working.

Can you please change this line:
```
v.SetDefault("adapters.mobilefuse.endpoint", "http://mfx-us-east.mobilefuse.com/openrtb?pub_id={{.PublisherID}}")
```

To this line:
```
v.SetDefault("adapters.mobilefuse.endpoint", "http://mfx.mobilefuse.com/openrtb?pub_id={{.PublisherID}}")
```

In this PR? When you're done, you'll only see the "adapters.mobfoxpb.endpoint" entry in this GitHub Files Changed.

mobfxoHB: Oh, i see, no this should be ok

mobfxoHB:
@mansinahar @SyntaxNode Good day, i've made last changes to resolve merge conflict, so, can we please finish with this PR ?

SyntaxNode (approved):

hhhjort (approved):
LGTM

PR #1590, opened 2020-11-17
Improve GitHub Actions Validation

SyntaxNode:
- Formatting updated.
- Validate workflow no longer runs on release creation. I don't see a need for this.
- Validate workflow now validates just the branch to match what TravisCI did.
- Validate Merge workflow added to validate things are still good after the auto-merge occurs.
- Dropped Go 1.13 build targets / support.
- Upgraded go.mod to 1.14.

### DIFF
--- a/.github/workflows/validate-merge.yml
+++ b/.github/workflows/validate-merge.yml
@@ -1,0 +1,24 @@
+name: Validate Merge
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  validate-merge:
+    runs-on: ubuntu-18.04
+    
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14.2
+
+    - name: Checkout Merged Branch
+      uses: actions/checkout@v2
+
+    - name: Validate
+      run: |
+        ./validate.sh --nofmt
+      env:
+        GO111MODULE: "on"
--- a/.github/workflows/validate-merge.yml
+++ b/.github/workflows/validate-merge.yml
@@ -19,6 +19,6 @@ jobs:
 
     - name: Validate
       run: |
-        ./validate.sh --nofmt
+        ./validate.sh --nofmt --cov --race 10
       env:
         GO111MODULE: "on"
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Checkout Branch
       uses: actions/checkout@v2
       with:
-        ref: ${{ github.event.milestone.title }}
+        ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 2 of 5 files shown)


.github/workflows/validate.yml
```diff
@@ -1,26 +1,30 @@
...
     - name: Install Go
       uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go-version }}
-    - name: Checkout code
+
+    - name: Checkout Branch
       uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
```
laurb9, line 26: With this, the PR checker can no longer guard against a failed merge. Also, `event.pull_request` is not available on push events.

 I suggest keeping the default checkout action, as it handles both cases as expected.

SyntaxNode: Understood. I've created a new action, validate-merge, to handle the case of a failed merge and repurposed this one for branch code only. My intention with separating these out is to clearly express to the contributor why a build is failing. @hhhjort was understandably confused in a previous PR why his tests passed locally but failed on the server. Took us all a bit to figure it out, especially since Travis behaves differently.

The validate-merge task is also simpler since it just runs on one Go version (the prod one) and omits race detection and code coverage reports.

.github/workflows/validate-merge.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  validate-merge:
+    runs-on: ubuntu-18.04
+    
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14.2
```
mansinahar, line 15: LGTM in general! I'd just suggest that we run the `validate-merge` action with both the supported versions rather than just one because that's going to be the actual state of the code when the PR gets merged and we wanna make sure that looks good with all the supported versions whereas for the PR branch validation, running with only one of the Go versions should be fine IMO because that is more to verify for the author that their tests pass on their PR as they do locally.

guscarreon (dismissed):
Because the command `./validate.sh --nofmt --cov --race 10`, is more complete than simply `./validate.sh --nofmt`, I thought either `.github/workflows/validate-merge.yml` or `.github/workflows/validate.yml` should run and not both if not needed. But given the fact that they both run in parallel and specially, that this first run after committing this PR itself was successful, LGTM.

.github/workflows/validate.yml
SyntaxNode: > Also, event.pull_request is not available on push events.

Correct. I suspected the ref would resolve to an empty string if not present and default to the HEAD for push requests. I just ran a series of experiments which proved that's what is happening.

mansinahar (approved):

bsardo (approved):

PR #1591, opened 2020-11-18
Fix GDPR consent assumption when gdpr req signal is unambiguous and s…

bsardo:
…et to 1 and consent string is blank

To address this issue, I pulled `UserSyncIfAmbiguous` out of the `extractGDPR` method so that method's sole responsibility now is to extract the GDPR field from the request. The logic that determines whether GDPR should be enforced by considering `UserSyncIfAmbiguous` has been moved into `CleanOpenRTBRequests`. If GDPR should be enforced and it’s enabled at either the host or account level, we check if personal info is allowed by considering the GDPR flag in the request along with the consent string.

Most of the test changes are conversions to a table-based format with a couple of new test cases.


gdpr/impl.go
```diff
@@ -42,21 +51,26 @@ func (p *permissionsImpl) BidderSyncAllowed(ctx context.Context, bidder openrtb_
...
-	if ok {
+func (p *permissionsImpl) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, gdpr GDPRState, consent string) (allowPI bool, allowGeo bool, allowID bool, err error) {
+	if _, ok := p.cfg.NonStandardPublisherMap[PublisherID]; ok {
 		return true, true, true, nil
 	}
 
-	id, ok := p.vendorIDs[bidder]
-	if ok {
-		return p.allowPI(ctx, id, consent)
+	if gdpr == NoGDPR {
```
hhhjort, line 59: If no GDPR, that is GDPR does not apply to this request, should we not always allow PI to pass (true, true, true, nil) rather than scrub all PI (false, false, false, nil)?

gdpr/impl_test.go
```diff
@@ -177,6 +177,73 @@ func TestMalformedConsent(t *testing.T) {
...
+		{
+			description: "Allow PI - Non standard publisher",
+			bidderName:  openrtb_ext.BidderAppnexus,
+			publisherID: "appNexusAppID",
+			gdpr:        YesGDPR,
+			consent:     "BOS2bx5OS2bx5ABABBAAABoAAAABBwAA",
+			allowPI:     true,
+		},
+		{
+			description: "Don't allow PI - No GDPR",
```
hhhjort, line 197: If GDPR does not apply, why would we not allow PI?

gdpr/impl_test.go
```diff
@@ -177,6 +177,73 @@ func TestMalformedConsent(t *testing.T) {
...
+		},
+		{
+			description: "Don't allow PI - default vendor with non-empty consent",
+			bidderName:  openrtb_ext.BidderRubicon,
+			gdpr:        YesGDPR,
+			consent:     "BOS2bx5OS2bx5ABABBAAABoAAAABBwAA",
+			allowPI:     false,
+		},
+		{
+			description: "PI allowed according to host settings - default vendor with empty consent",
```
hhhjort, line 225: What is the host setting that allows PI when GDPR applies and no consent is given?

bsardo: `UserSyncIfAmbiguous` is the setting I was referring to.

gdpr/impl.go
bsardo: Oops, yeah that should be `true, true, true, nil`.

gdpr/impl_test.go
bsardo: Not sure where my head was at here. This mistake is related to the one you pointed out above.

gdpr/impl_test.go
hhhjort: `UserSyncIfAmbiguous` should only come into play if `AmbiguousGDPR`. If the GDPR setting is not ambiguous, then we should not be using the "IfAmbiguous" setting for anything.

bsardo: Makes sense. I didn't change the default vendor permissions logic but it seems I need to.

If the bidder is not in the vendor list, we fall back to the default vendor permissions. Should the permissions be modified to include the `&& gdpr == AmbiguousGDPR` condition as shown here?
```if consent == "" && gdpr == AmbiguousGDPR {
	return p.cfg.UsersyncIfAmbiguous, p.cfg.UsersyncIfAmbiguous, p.cfg.UsersyncIfAmbiguous, nil
}
return false, false, false, nil```

Comment (hhhjort) [reply gdpr/impl_test.go]: that seems correct

**Review by hhhjort — APPROVED** [review]
LGTM

**Review by SyntaxNode — COMMENTED** [review]
Nice refactor and fix. I like this direction.

File: endpoints/auction_test.go [hunk 4]
```diff
@@ -420,7 +420,7 @@ func (m *auctionMockPermissions) BidderSyncAllowed(ctx context.Context, bidder o
 	return m.allowBidderSync, nil
 }
 
-func (m *auctionMockPermissions) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, consent string) (bool, bool, bool, error) {
+func (m *auctionMockPermissions) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, gdpr gdpr.Signal, consent string) (bool, bool, bool, error) {
```
SyntaxNode, line 423: Consider renaming `gdpr` -> `gdprSignal` for clarity in the interface and impelemtators.

exchange/utils.go
```diff
@@ -79,8 +79,10 @@ func cleanOpenRTBRequests(ctx context.Context,
 		return
 	}
 
-	gdpr := extractGDPR(orig, usersyncIfAmbiguous)
+	//TODO(bfs): if an error is encountered while extracing GDPR or Consent, should user syncing be considered ambiguous or don't sync?
```
SyntaxNode, line 82: Let's create a follow-up story for tracking rather than this comment, please.

exchange/utils.go
```diff
@@ -104,7 +106,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 
 	gdprEnabled := gdprEnabled(account, privacyConfig, integrationTypeMap[labels.RType])
```
SyntaxNode, line 107: Consider using just one variable and combining it as:
```
gdprEnforced = gdprEnforced && gdprEnabled(account, privacyConfig, integrationTypeMap[labels.RType])
```

 and then later on

```
if gdprEnforced {
  ...
```

exchange/utils_test.go
```diff
@@ -1124,6 +1126,31 @@ func TestCleanOpenRTBRequestsGDPR(t *testing.T) {
...
 			},
 		},
+		{
+			description:        "Enforce - Ambiguous signal, don't sync user if ambiguous",
+			gdprAccountEnabled: nil,
+			gdprHostEnabled:    true,
+			gdpr:               "null",
+			gdprConsent:        "BONV8oqONXwgmADACHENAO7pqzAAppY",
+			gdprScrub:          true,
+			expectPrivacyLabels: pbsmetrics.PrivacyLabels{
```
SyntaxNode, line 1136: Please add `userSyncIfAmbiguous: false` here for ease of reading.

gdpr/impl.go
```diff
@@ -19,6 +19,14 @@ import (
 //
 // Nothing in this file is exported. Public APIs can be found in gdpr.go
 
+type Signal int
+
+const (
+	AmbiguousGDPR Signal = -1
+	NoGDPR        Signal = 0
+	YesGDPR       Signal = 1
+)
```
SyntaxNode, line 28: No need to repeat the GDPR name here since it's already package scoped. How about something like:
```
const (
	SignalAmbiguous Signal = -1
	SignalNo        Signal = 0
	SignalYes       Signal = 1
)
```

gdpr/impl.go
```diff
@@ -98,11 +111,13 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
 	return false, nil
 }
 
-func (p *permissionsImpl) allowPI(ctx context.Context, vendorID uint16, consent string) (bool, bool, bool, error) {
-	// If we're not given a consent string, respect the preferences in the app config.
-	if consent == "" {
+func (p *permissionsImpl) allowPI(ctx context.Context, vendorID uint16, gdpr Signal, consent string) (bool, bool, bool, error) {
+	if consent == "" && gdpr == AmbiguousGDPR {
 		return p.cfg.UsersyncIfAmbiguous, p.cfg.UsersyncIfAmbiguous, p.cfg.UsersyncIfAmbiguous, nil
 	}
```
SyntaxNode, line 117: If `allowPI` is only ever called from `PersonalInfoAllowed`, consider moving up the `consent == "" && gdpr == AmbiguousGDPR` check before the vendor check, since it's done first by either possible branch.

bsardo: I confirmed this is only called from `PersonalInfoAllowed` so I moved this and the `if consent == "" && gdprSignal == SignalYes {` check immediately after it before the vendor check.

SyntaxNode:
Removing stale label to end of year holidays.

gdpr/impl_test.go
```diff
@@ -177,6 +177,80 @@ func TestMalformedConsent(t *testing.T) {
...
+			consent:     "",
+			allowPI:     false,
+		},
+		{
+			description: "Don't allow PI - default vendor with ambiguous GDPR and non-empty consent",
+			bidderName:  openrtb_ext.BidderRubicon,
+			gdpr:        SignalAmbiguous,
+			consent:     "BOS2bx5OS2bx5ABABBAAABoAAAABBwAA",
+			allowPI:     false,
+		},
```
SyntaxNode, line 244: I'm a bit confused by this case. What do I have wrong in my head?

The above test case should allow PI because we have an ambiguous signal with UsersyncIfAmbiguous set to true (which means do not enforce GDPR?).

gdpr/impl_test.go
```diff
@@ -178,12 +178,13 @@ func TestMalformedConsent(t *testing.T) {
...
-		description string
-		bidderName  openrtb_ext.BidderName
-		publisherID string
-		gdpr        Signal
-		consent     string
-		allowPI     bool
+		description         string
+		bidderName          openrtb_ext.BidderName
+		publisherID         string
+		userSyncIfAmbiguous bool
```
SyntaxNode, line 184: I like this change. Could you please add `userSyncIfAmbiguous: false` to the tests which have it disabled? I know this is the default value and not strictly necessary, but it makes the test cases more self documenting and easier to read.

SyntaxNode (approved):

hhhjort (approved):

Issue #1428, opened 2020-08-04
Integration Type and Channel


bretg:
Our hosted Prebid Server has been getting more complicated with events, analytics, special cases, and additional business models. We'd like to propose some changes to support distinguishing between various traffic sources and business arrangements.

# Integration Type

In support of analytics and billing, Rubicon's fork of PBS-Java supports a concept called "integration type". It's a host-company-specific set of values used to indicate the source of the request, which is turn is used in events and billing calculations. For example, we use one value for our "fully managed service" offering, and other value for our "hosted service" offering, which have different financial characteristics.

Example use cases:
- host company getting PBS traffic from their own managed PBJS: "mpbjs" (managed)
- host company getting PBS traffic from their a publisher-managed PBJS: "shpbjs" (self-hosted)
- host company getting PBS traffic from a 3rd-party managed PBJS: "3pbjs"

We'd like to take this feature out of our private fork and formalize it with the community. Here's what it impacts:

1) Account configuration can define a default `integration` 
2) OpenRTB requests can contain `ext.prebid.integration`. Allowed values are dependent on the host company.
3) The `integration` value is added to event URLs generated by Prebid Server and is available to analytics adapters.

# Channel

Also in support of analytics, we'd like to formalize a value that defines how the request came in. It's not used by PBS-core, but is available to analytics adapters. Valid values could vary by host company, but for example:

- pbjs
- app
- amp
- s2s

Proposal:

1. Channel is an object that contains a name (required) and a version (optional)
2. Update the PBJS pbsBidAdapter to pass ext.prebid.channel: {name: "pbjs", version: "4.1"} unless overridden by s2sConfig.extPrebid
3. PBS gets the channel name out of the request if it's there
4. If the channel isn't present:
    4a. set channel name to 'amp' if the request came in on /openrtb2/amp
    4b. set channel name to 'app' if the request contains an 'app' object
    4c. otherwise leave channel blank. Analytics adapters can default if they'd like.

albertmolinermrf:
Would it be possible to make the proposal the default behaviour, but allow the ability to:

1. Skip the addition from Prebid.js (i.e., not pass anything at all).
1. Pass any channel name from Prebid.js.

bretg:
[...]

PR #1597, opened 2020-11-24
Adding Events support in bid responses

laurb9:
Implemented wurl insertion in cached bids, vast rewriting and bid.ext.prebid.events for #1015 
Updated and refactored auction_test unit tests for cached bids
Updated exchange_test for bid.ext.prebid.events and vast rewriting

Test coverage
>ok  	github.com/prebid/prebid-server/exchange	6.039s	coverage: 90.5% of statements
>ok  	github.com/prebid/prebid-server/endpoints/events	0.022s	coverage: 97.4% of statements

PG is deferred to later PR (PBS-Java does not implement it either)

CC @bretg


laurb9:
Rebased after #1532

endpoints/events/vtrack.go
```diff
@@ -269,25 +270,34 @@ func getAccountId(httpRequest *http.Request) string {
...
-func modifyVastXml(externalUrl string, data json.RawMessage, bidid string, bidder string, accountId string, timestamp int64) json.RawMessage {
-	c := string(data)
+// modifyVastXmlString rewrites and returns the string vastXML or an empty string if it was left unmodified
+func ModifyVastXmlString(externalUrl string, c string, bidid string, bidder string, accountId string, timestamp int64) string {
 	ci := strings.Index(c, ImpressionCloseTag)
 
 	// no impression tag - pass it as it is
 	if ci == -1 {
-		return data
+		return ""
```
SyntaxNode, line 279: Why the change to an empty string from passing it through? aka.. why not continue returning `c`?

endpoints/events/vtrack.go
```diff
@@ -269,25 +270,34 @@ func getAccountId(httpRequest *http.Request) string {
 	return httpRequest.URL.Query().Get(AccountParameter)
 }
 
-// modifyVastXml modifies BidCacheRequest element Vast XML data
-func modifyVastXml(externalUrl string, data json.RawMessage, bidid string, bidder string, accountId string, timestamp int64) json.RawMessage {
-	c := string(data)
+// modifyVastXmlString rewrites and returns the string vastXML or an empty string if it was left unmodified
+func ModifyVastXmlString(externalUrl string, c string, bidid string, bidder string, accountId string, timestamp int64) string {
```
SyntaxNode, line 274: Nitpick: There's no need to repeat type declarations of arguments. This can be simplified to:
```
func ModifyVastXmlString(externalUrl, c, bidid, bidder, accountId string, timestamp int64) string {
```

endpoints/events/vtrack.go
```diff
@@ -269,25 +270,34 @@ func getAccountId(httpRequest *http.Request) string {
...
 	if ci-oi == len(ImpressionOpenTag) {
-		return json.RawMessage(strings.Replace(c, ImpressionOpenTag, ImpressionOpenTag+impressionUrl, 1))
+		return strings.Replace(c, ImpressionOpenTag, ImpressionOpenTag+impressionUrl, 1)
 	}
 
-	return json.RawMessage(strings.Replace(c, ImpressionCloseTag, ImpressionCloseTag+ImpressionOpenTag+impressionUrl+ImpressionCloseTag, 1))
+	return strings.Replace(c, ImpressionCloseTag, ImpressionCloseTag+ImpressionOpenTag+impressionUrl+ImpressionCloseTag, 1)
+}
+
+// modifyVastXml modifies BidCacheRequest element Vast XML data
```
SyntaxNode, line 293: Nitpick: Comment is outdated. `modifyVastXml` -> `ModifyVastXmlJSON`

endpoints/events/vtrack.go
```diff
@@ -269,25 +270,34 @@ func getAccountId(httpRequest *http.Request) string {
...
+func ModifyVastXmlJSON(externalUrl string, data json.RawMessage, bidid string, bidder string, accountId string, timestamp int64) json.RawMessage {
+	c := string(data) // FIXME: should decode json, or escapes like "\u003e" will not be recognized as "<"
```
SyntaxNode, line 295: `FIXME` Ah! Can you fix it before merging it in please?

exchange/events.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/analytics"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/endpoints/events"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/pbsmetrics"
+)
+
+// eventsData has configuration fields needed for adding event tracking to an auction response
+type eventsData struct {
```
SyntaxNode, line 18: Nitpick: Consider renaming to `eventTracking`?

exchange/events.go
```diff
@@ -0,0 +1,89 @@
...
+	bidderInfos        adapters.BidderInfos
+	externalURL        string
+}
+
+// getExtEventsData creates an eventsData object from the different configuration sources
+func getExtEventsData(requestExtPrebid *openrtb_ext.ExtRequestPrebid, ts time.Time, account *config.Account, bidderInfos adapters.BidderInfos, externalURL string) *eventsData {
+	return &eventsData{
+		accountID:          account.ID,
+		enabledForAccount:  account.EventsEnabled,
+		enabledForRequest:  requestExtPrebid != nil && requestExtPrebid.Events != nil, // || account.analytics.auction-events.<web|app|amp>
```
SyntaxNode, line 33: What's with the comment here?

openrtb_ext/request.go
```diff
@@ -23,6 +23,7 @@ type ExtRequestPrebid struct {
 	SChains              []*ExtRequestPrebidSChain `json:"schains,omitempty"`
 	StoredRequest        *ExtStoredRequest         `json:"storedrequest,omitempty"`
 	Targeting            *ExtRequestTargeting      `json:"targeting,omitempty"`
+	Events               json.RawMessage           `json:"events,omitempty"`
```
SyntaxNode, line 26: Can this be a strongly typed struct instead?

exchange/exchange.go
```diff
@@ -52,11 +52,13 @@ type IdFetcher interface {
 
 type exchange struct {
 	adapterMap          map[openrtb_ext.BidderName]adaptedBidder
+	bidderInfo          adapters.BidderInfos
```
SyntaxNode, line 55: Do we really need the whole info, or perhaps just a boolean lookup for if vast is enabled for the bidders?

SyntaxNode (commented):
I didn't review the tests in detail yet. Have some other comments first.

endpoints/events/vtrack.go
laurb9: Nice, didn't know that

endpoints/events/vtrack.go
laurb9: Empty string indicates no change was needed, which for `ModifyVastXmlJSON` means to skip the conversion back to RawMessage. I felt this was ok given that empty string is not a possible outcome from doing the replacement on any non-empty input.

exchange/events.go
laurb9: Will do. I had named it `eventsData` to parallel `targetingData` which has similar interface.

exchange/events.go
laurb9: This should be on the account line. It's a note I left regarding the algorithm making a mention to a channel event configuration for account, which I decided to leave out for now, to add in the next iteration.

> A new account-level configuration is proposed to define whether an account needs server-side events by channel. e.g account 111 gets AMP/App events, but not web events. Account 222 gets events for all 3 channels: Web/AMP/App.
If an account doesn't specify the channel config, the default is that AMP and App are "true" if events are enabled.

openrtb_ext/request.go
laurb9: It can, but it would be an empty struct, as the only possible value so far is `events: {}`.

exchange/exchange.go
laurb9: I thought to keep it generic, so if we add per-channel settings later, or need to look up other bidder settings, we would not need to go back and modify the `exchange` struct and all the unit tests.

Performance-wise, it looks equivalent, only one map lookup is done either way to reach the value.
Let me know your thoughts.

endpoints/events/vtrack.go
laurb9: I was debating fixing this as a case of doing it fast vs doing it right, since only the Go serializer insists on escaping html - but perhaps I should.

laurb9: vscode nags me about the naming of these functions. It says they should be `ModifyVastXMLJSON` and `ModifyVastXMLString` respectively (probably because it doesn't know VAST is also an acro). Thoughts ?

endpoints/events/vtrack.go
SyntaxNode: Gotcha. Do you think it would be clearer to add an "ok" boolean return, like Go does for map retrievals?

endpoints/events/vtrack.go
SyntaxNode: I don't have a strong opinion on the name. You're right. It should technically be called ModifyVASTXMLJSON.. but I mean... my eyes! I would just leave it as is and not worry.

Could you please fill me in on why escaping the HTML is a problem here? I just don't have the context. Anyway, you can customize the Go json encoder to turn this feature off:
https://play.golang.org/p/C79fK-_OKe5

exchange/events.go
SyntaxNode: Gotcha. Could you perhaps leave this out or better explain it? It looks like a mistake without that context.

openrtb_ext/request.go
SyntaxNode: Ah. Is this like caching where the presence is the trigger?

exchange/exchange.go
SyntaxNode: I wasn't worried about performance. Thought it might be simpler to only pass the information needed. I feel 50/50 on it. I imagine we'll refactor the bidderInfo map to some kind of bidderInfoService in the future anyway. I think we might need more context after I add in hardcoded aliases.

openrtb_ext/request.go
laurb9: `ext.prebid.cache` needs a value though (`vastxml` and/or `bids`). 
I expect that in time, so will `ext.prebid.events`, but for now it works as you said.

endpoints/events/vtrack.go
laurb9: Sorry, html escapes in Go are a red herring here and probably worth a separate discussion.
Regarding the issue at hand, PBS has no control over how adm was encoded, so it has to be able to handle it, so I replaced `string()` with a proper `UnMarshal()`

laurb9:
I implemented part of the spec change mentioned here https://github.com/prebid/prebid-server/issues/1015#issuecomment-740138755 - the modified VAST creative should be returned in bid response, not only cache.

I moved the VAST rewriting from auction/doCache to exchange/HoldAuction with the associated unit test migration, as they use two different sets of json specs. There are still some unanswered questions regarding the requirement that targeting be enabled for VAST rewriting because only the winning bids are modified, or even if we should modify all the bids instead.

For now, this implementation requires targeting, until the requirements are clarified. This should be ok if most requests use GAM or some other network that needs the targeting keys anyway, so targeting would be enabled.

laurb9:
@bretg Added 6797b3a (unit tests) and 5f3018a (implementation) for updated functionality, sans PG or channels, which can be added in a later PR.

Updates:
- all VIDEO bids, cached and returned, have modified VAST if account and bidder enabled - not just winning cached bids.
- all NON-VIDEO returned bids have `ext.prebid.events` if account or request enabled
- all cached NON-VIDEO bids have `wurl` patched in if account or request enabled.

ok  	github.com/prebid/prebid-server/exchange	6.039s	coverage: 90.5% of statements

laurb9:
I'm going to hold off resolving conflicts from #1626 until all the other module renames I see as pending got in, or this PR gets the go-ahead, whichever comes first :)

laurb9:
Updated to new spec (ext.prebid.events now overrides server config for both vast and banner)
Resolved conflicts merging the latest master branch.

exchange/events_test.go
```diff
@@ -0,0 +1,120 @@
+package exchange
+
+import (
+	"testing"
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/stretchr/testify/assert"
+)
+
```
bsardo, line 10: You've put together some nice integration tests to obtain coverage of `modifyBidsForEvents` and helper methods `isModifyingVASTXMLAllowed` and `modifyBidVAST`. Thoughts on whether we should add a unit test for `modifyBidsForEvents`? I imagine you figured it would be a pain to test via UT because of the complexity around creating the seat bid function inputs.

laurb9: I've left it out because its behavior and parameters changed with the spec changes and I started leaning on the exchangeJSON tests to validate them instead.
Since those cover the code anyway, I feel the specialized UT would add less value now.

exchange/exchangetest/events-vast-account-off-request-off.json
```diff
@@ -0,0 +1,160 @@
...
+                      "hb_pb": "0.70",
+                      "hb_size": "200x250"
+                  }
+                }
+              }
+            },
+            {
+              "id": "losing-bid",
+              "nurl": "http://domain.com/losing-bid",
+              "impid": "my-imp-id",
```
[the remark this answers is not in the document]
laurb9: Correct. The adm would be generated only to modify it, so if we do not need to modify then it is left alone.
This felt a little inconsistent at first (if it's ok to generate sometimes, why not always do it) but I've made peace with it.

exchange/exchangetest/events-bid-account-off-request-off.json
```diff
@@ -0,0 +1,100 @@
...
+                }
+              }
+            }
+          ]
+        }
+      ]
+    },
+    "prebid": {
+      "auctiontimestamp": 1234567890
+    }
```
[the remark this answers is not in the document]
laurb9: These seem to have been added in 4fac314 and are indeed not used. Will remove.

endpoints/events/vtrack.go
```diff
@@ -269,32 +269,37 @@ func getAccountId(httpRequest *http.Request) string {
...
+		// failed to decode json, fall back to string
+		vast = string(data)
 	}
-
-	i := sort.SearchStrings(s, e)
-	return i < len(s) && s[i] == e
+	var ok bool
+	vast, ok = ModifyVastXmlString(externalUrl, vast, bidid, bidder, accountId, timestamp)
+	if !ok {
+		return data
```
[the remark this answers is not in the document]
laurb9: Thought we can just return the original `data` unmodified as it's already json.RawMessage.

exchange/eventscachetest/json-account-off-request-on.json
```diff
@@ -0,0 +1,49 @@
...
+            "price": 5.64
+        },
+        "bidType": "banner",
+        "bidder": "pubmatic"
+    }],
+    "expectedCacheables": [
+        {
+            "type": "json",
+            "ttlseconds": 660,
+            "value":{ "id": "bidOne", "impid": "oneImp", "price": 7.64, "exp": 600, "wurl": "http://localhost/event?t=win\u0026b=bidOne\u0026a=ACCOUNT_ID\u0026bidder=appnexus\u0026ts=1234567890"}
```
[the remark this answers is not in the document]
laurb9: Works either way because they're the same in json, but I will make it `&` so it is consistent.

bsardo (approved):
This is looking good. I'll get a review from another team member this week.

exchange/auction_test.go
```diff
@@ -245,7 +208,14 @@ func runCacheSpec(t *testing.T, fileDisplayName string, specData *cacheSpec) {
 		winningBidsByBidder: winningBidsByBidder,
 		roundedPrices:       roundedPrices,
 	}
-	_ = testAuction.doCache(ctx, cache, targData, &specData.BidRequest, 60, &specData.DefaultTTLs, bidCategory, &specData.DebugLog)
+	evTracking := &eventTracking{
+		accountID:          "ACCOUNT_ID",
```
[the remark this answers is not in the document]
laurb9: Better to rename ACCOUNT_ID to TEST_ACC_ID to avoid confusion.

endpoints/events/vtrack.go
```diff
@@ -269,32 +269,37 @@ func getAccountId(httpRequest *http.Request) string {
...
+	var vast string
+	if err := json.Unmarshal(data, &vast); err != nil {
+		// failed to decode json, fall back to string
+		vast = string(data)
 	}
-
-	i := sort.SearchStrings(s, e)
-	return i < len(s) && s[i] == e
+	var ok bool
+	vast, ok = ModifyVastXmlString(externalUrl, vast, bidid, bidder, accountId, timestamp)
```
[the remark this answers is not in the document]
laurb9: Looks like it could work;  I must have written it originally as a single `if` where the other one was shadowed, and it ended like this

exchange/events.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// getEventTracking creates an eventTracking object from the different configuration sources
+func getEventTracking(requestExtPrebid *openrtb_ext.ExtRequestPrebid, ts time.Time, account *config.Account, bidderInfos adapters.BidderInfos, externalURL string) *eventTracking {
+	return &eventTracking{
+		accountID:          account.ID,
+		enabledForAccount:  account.EventsEnabled,
+		enabledForRequest:  requestExtPrebid != nil && requestExtPrebid.Events != nil,
+		auctionTimestampMs: ts.UnixNano() / 1e+6,
+		integration:        "", // FIXME
```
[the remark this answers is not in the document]
laurb9: This was meant for integration support, then I found out it is not yet implemented #1428 
I can remove it, but I would prefer to keep it to avoid integration popping up as a parameter later when it could be here. Thoughts ?

exchange/events.go
```diff
@@ -0,0 +1,112 @@
...
+	if !ev.enabledForAccount && !ev.enabledForRequest || pbsBid.bidType == openrtb_ext.BidTypeVideo {
+		return jsonBytes
+	}
+	var winEventURL string
+	if pbsBid.bidEvents != nil { // All bids should have already been updated with win/imp event URLs
+		winEventURL = pbsBid.bidEvents.Win
+	} else {
+		winEventURL = ev.makeEventURL(analytics.Win, pbsBid, bidderName)
+	}
+	// wurl attribute is not in the schema, so we have to patch
```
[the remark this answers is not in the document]
laurb9: It does not appear to be. It also means cached bids look different with respect to events than returned bids (cached bids have `wurl`, returned bids have`ext.prebid.events`).

exchange/events.go
```diff
@@ -0,0 +1,112 @@
...
+	if patch, err := json.Marshal(map[string]string{"wurl": winEventURL}); err == nil {
+		if modifiedJSON, err := jsonpatch.MergePatch(jsonBytes, patch); err == nil {
+			jsonBytes = modifiedJSON
+		}
+	}
+	return jsonBytes
```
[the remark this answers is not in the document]
laurb9: I thought it shouldn't happen because jsonBytes is valid json (it was marshalled earlier).
But I agree that ignored errors can waste time later, so I can add it to the error list.

exchange/events.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// makeBidExtEvents make the data for bid.ext.prebid.events if needed, otherwise returns nil
+func (ev *eventTracking) makeBidExtEvents(pbsBid *pbsOrtbBid, bidderName openrtb_ext.BidderName) *openrtb_ext.ExtBidPrebidEvents {
+	if !ev.enabledForAccount && !ev.enabledForRequest || pbsBid.bidType == openrtb_ext.BidTypeVideo {
```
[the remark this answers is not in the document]
laurb9: `&&` takes precedence and the expression is correct. I suppose I can also write it like this, if it's more readable:
```go
if !(ev.enabledForAccount || ev.enabledForRequest) || pbsBid.bidType == openrtb_ext.BidTypeVideo {
```

laurb9:
Addressed comments and resolved conflicts from #1644

exchange/events.go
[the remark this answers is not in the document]
SyntaxNode: Ah ok. Gotcha. I think that's fine.

exchange/events.go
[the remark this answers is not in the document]
SyntaxNode: Thanks. The extra parenthesis makes it much easier to follow, IMHO.

SyntaxNode (approved):

bsardo (approved):
LGTM

PR #1598, opened 2020-11-25
Fix appnexus adapter not setting currency in the bid response

guscarreon:
It has been reported that appnexus may not be behaving properly in foreign currency contexts. It appears that the adapter does not set the `curr` (currency), field in the bid response. This pull request makes this correction and adds a JSON test file but proper assertion of the `Currency` field of the `BidderResponse` inside `adapters/adapterstest/test_json.go` will be implemented in a separate PR being developed by @mansinahar


adapters/appnexus/appnexus.go
```diff
@@ -595,6 +595,9 @@ func (a *AppNexusAdapter) MakeBids(internalRequest *openrtb.BidRequest, external
 						BidVideo:     impVideo,
 						DealPriority: bidExt.Appnexus.DealPriority,
 					})
+					if len(internalRequest.Cur) > 0 {
+						bidResponse.Currency = internalRequest.Cur[0]
```
mansinahar, line 599: Shouldn't this be set to `bidResp.Cur`? What's the rationale behind setting it to the currency in the request? Is the Appnexus bidder always guaranteed to bid in the same currency as in the request? Even then I'd say we should be setting it to `bidResp.Cur` rather than `internalRequest.Cur[0]`

guscarreon: Great catch. I'm also moving the `if` statement to a place that makes more sense. Test was modified accordingly.

SyntaxNode:
That merge doesn't look right. All stories merged since this one was created are showing up in the diff.

.github/workflows/validate-merge.yml
```diff
@@ -0,0 +1,24 @@
+name: Validate Merge
```
hhhjort, line 1: Is this file really part of the PR, or is it in to resolve merge conflicts?

hhhjort (approved):
LGTM, assuming all the merge conflict stuff hasn't been touched since it came in from master.

SyntaxNode:
I've seen GitHub get confused before on merges. I'm going to close and reopen.

SyntaxNode:
That didn't work. I'm just going to recreate this PR.

PR #1602, opened 2020-11-30
Fix 33Across App Handling

SyntaxNode:
@curlyblueeagle We observed a bug with the 33Across adapter for handling app requests. This is a fix. Look good to you?

I fixed the bug and added app tests


curlyblueeagle:
@SyntaxNode looks good to me. Thanks for the fix.

guscarreon (approved):
LGTM

mansinahar (approved):

PR #1606, opened 2020-12-02
Update MobileFuse Endpoint

dtbarne:
We (MobileFuse) are going to start using our global endpoint, which uses the nearest data center automatically. This PR is to update the endpoint accordingly.


SyntaxNode (approved):

guscarreon (approved):
LGTM

Issue #1604, opened 2020-12-01
Question: Multi-Imp requests


curlyblueeagle:
@SyntaxNode In a prior PR you had suggested for 33Across adapter to handle multi-imp requests by breaking them up into multiple http requests https://github.com/prebid/prebid-server/pull/1557#discussion_r520008392

We are in the process of implementing this suggestion. However, I have a question for you. How would Prebid Server handle the individual bid responses and associate them with the incoming mult-imp bid request? The adapter will of course pass the Imp ID in each of the bid response during `MakeBids`. Will this be sufficient? Does the adapter need to do anything more?

bretg:
Not sure how this works technically. @SyntaxNode ?

SyntaxNode:
@curlyblueeagle Yes, that will be sufficient.

Prebid Server will call your adatper's `MakeBids` method once for each response received and combine the bids from all of those calls together as a single OpenRTB Seat Bid for your adapter. I recommend you perform end-to-end testing to make sure it looks right when all is said and done, but I don't anticipate any problems.

FYI: The code for this behavior is found in `exchange/bidder.go` in the `requestBid` method if you'd like to explore the logic further. Look for the `for i := 0; i < len(bidResponse.Bids); i++ {` loop.

curlyblueeagle:
@SyntaxNode Thanks! 

We did an end-to-end test and all looks well. So I am going to close this issue since my question was answered.

Issue #1042, opened 2019-09-24
Support request level Bidder Params


PubMatic-OpenWrap:
Reopening issue for #626 

A few Bidders now have defined params that aren't really "per-Imp" params. Appnexus has `member`, and Pubmatic has `wrapper` param.

There may be others too. It might be a good idea to generalize support for bidder params in the `request.ext`, rather than forcing them all to be in the `imp`.

PubMatic-OpenWrap:
This is the approach that we propose to add request specific bidder parameters in the oRTB request.

Currently, the Ext object in the BidRequest is openrtb_ext.ExtRequest
// ExtRequest defines the contract for bidrequest.ext
type ExtRequest struct {
Prebid ExtRequestPrebid json:"prebid"
}

The proposal is to add another parameter 'BidderParams', which would contain bidder specific data, to ExtRequestPrebid

```
// ExtRequestPrebid defines the contract for bidrequest.ext.prebid 
type ExtRequestPrebid struct { 
  Aliases               map[string]string       json:"aliases,omitempty"
  BidAdjustmentFactors  map[string]float64      json:"bidadjustmentfactors,omitempty"
  Cache                 *ExtRequestPrebidCache  json:"cache,omitempty"
  StoredRequest         *ExtStoredRequest       json:"storedrequest,omitempty"
  Targeting             *ExtRequestTargeting    json:"targeting,omitempty"
  Debug                 int                     json:"debug,omitempty"
  BidderParams          interface{}             json:"bidderparams,omitempty"   ==> new parameter 
}
```

BidderParams interface would be a map, with key as bidderName and value would be another map containing bidder specific data.

Example:   
```
var BidderParams map[BidderName]BidderSpecificParams 
type BidderSpecificParams map[string]interface{}
```
Each adapter can/should define what the bidder specific data (BidderSpecificParams) would be in order for them to retrieve that information from the request
so BidderParams in the request that prebid would receive would look like


```
"bidderparams": { 
  "adapter1": { 
    "key1": "data specific to key1", 
    "key2": "  { 
      "version": 3.3, 
       "profileid": 1234 
      } 
    }, 
  "adapter2": { 
   "key3": "data specific to key3" 
  }, 
  "adapter2": { 
    "key4": [ 
     "data" 
    ] 
  } 
} 

```
This data can be filtered when the request is split for each bidder so that request that is sent to the adapter only has the data specific to that adapter


So request going to adapter1 will have  
```
request: { 
[...]

PR #1609, opened 2020-12-03
33Across: Add support for multi-imp requests

curlyblueeagle:
We have the following changes made:

1. The 33Across exchange server does not support SRA (requests with multiple imp objects) and therefore the adapter, until now, operated under the assumption of only a single imp object per request. This resulted in incomplete validations and multi-imp requests generating a 400 response. 
In https://github.com/prebid/prebid-server/pull/1557#discussion_r520008392 splitting up such requests into individual http calls was suggested. This PR implements that suggestion.

2. Thanks to https://github.com/prebid/prebid-server/pull/1602 we realized that although our exchange supports only `site` based requests the adapter lets `app` requests. Updated the YAML to define our support only for site.

3. Also added caller as `Prebid Server` in our ext field to help us keep track of requests.


curlyblueeagle:
Validate failing because of AppNexus.

bretg:
Looks like this is related to https://github.com/prebid/prebid-server/issues/1604

curlyblueeagle:
> Looks like this is related to #1604

@bretg yeah I asked that question to ensure that our implementation in this PR is correct.

adapters/33across/33across.go
```diff
@@ -38,39 +51,62 @@ func (a *TtxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters
...
 	var adapterRequests []*adapters.RequestData
 
-	adapterReq, errors := a.makeRequest(request)
-	if adapterReq != nil {
-		adapterRequests = append(adapterRequests, adapterReq)
+	// Break up multi-imp request into multiple external requests since we don't
+	// support SRA in our exchange server
+	for i := 0; i < len(request.Imp); i++ {
+		adapterReq, err := a.makeRequest(request, request.Imp[i])
+		if adapterReq != nil {
```
SyntaxNode, line 58: It's more common to check for errors before proceeding, rather than checking for nil. Ideally should be closer to:
```
if adapterReq, err := a.makeRequest(request, request.Imp[i]); err != nil {
  adapterRequests = append(adapterRequests, adapterReq)
} else {
  errs = append(errs, err)
}


```

adapters/33across/33across.go
```diff
@@ -38,39 +51,62 @@ func (a *TtxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters
...
 
-	if reqCopy.Imp[0].Banner == nil && reqCopy.Imp[0].Video == nil {
-		errs = append(errs, &errortypes.BadInput{
-			Message: "At least one of [banner, video] formats must be defined in Imp. None found",
-		})
+	reqCopy.Imp = append(imps, impCopy)
 
-		return nil, errs
+	// Add info about caller
+	caller := TtxCaller{"Prebid-Server", "n/a"}
```
SyntaxNode, line 82: Consider making this a constant / static value to avoid memory allocations per call.

adapters/33across/33across.go
```diff
@@ -81,63 +117,64 @@ func (a *TtxAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.Request
...
 		Headers: headers,
-	}, errs
+	}, nil
 }
 
-// Mutate the request to get it ready to send to ttx.
-func preprocess(request *openrtb.BidRequest) error {
-	var imp = &request.Imp[0]
+func makeImps(imp openrtb.Imp) (openrtb.Imp, error) {
+	impCopy := imp
```
SyntaxNode, line 124: This line isn't necessary. The copy is made implicitly by passing by value to the func.

adapters/33across/33across.go
curlyblueeagle: I am new to Golang. How would I make this into a const? Declare it as a top-level variable?

adapters/33across/33across.go
curlyblueeagle: assuming you mean `err == nil` check in the if statement?

SyntaxNode: Ha. Yes. Absolutely. :)

adapters/33across/33across.go
SyntaxNode: Yup. It can't literally be a const because it's a struct, but a package/top level variable which you only read from is how you would do it.

adapters/33across/33across.go
```diff
@@ -38,39 +55,59 @@ func (a *TtxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters
...
-func (a *TtxAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
-	var errs []error
+func (a *TtxAdapter) makeRequest(request *openrtb.BidRequest, imp openrtb.Imp) (*adapters.RequestData, error) {
+	var imps []openrtb.Imp
 
-	// Make a copy as we don't want to change the original request
 	reqCopy := *request
-	if err := preprocess(&reqCopy); err != nil {
-		errs = append(errs, err)
+	impCopy, err := makeImps(imp)
```
mansinahar, line 76: I'd suggest extracting this logic related to `reqExt` into it's own function and calling it before calling `makeRequest` for each imp because right now this code is executed for each imp but that's not really needed as the `reqExt` is gonna be the same and so you don't need to unmarshal and marshal it for each imp.

Once you do that, you can then only pass the `reqExt` and the `imp` under consideration to the `makeRequest` function.

JSON marshaling and unmarshaling is an expensive operation and we wanna make sure we avoid doing that multiple times unless it is absolutely needed.

curlyblueeagle: Done.

SyntaxNode (commented):
Code looks great. I left some small nitpicks, but the logic looks good to me and the test coverage for the code added in this PR is perfect.

adapters/33across/33across.go
```diff
@@ -38,39 +55,45 @@ func (a *TtxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters
...
 
-	if reqCopy.Imp[0].Banner == nil && reqCopy.Imp[0].Video == nil {
-		errs = append(errs, &errortypes.BadInput{
-			Message: "At least one of [banner, video] formats must be defined in Imp. None found",
-		})
+	reqCopy.Imp = append(imps, impCopy)
 
-		return nil, errs
-	}
+	reqCopy.Ext = TTXRequestExt
```
mansinahar, line 91: Since the `ttxRequestExt` doesn't really change based on imps, I'd suggest moving this outside of this function on L64. You can safely set this as the requestExt for the request sent to your adapter. This request is already a (shallow) copy that is made specifically for your adapter so you can safely modify it unless you're modifying one of the reference/pointer fields

adapters/33across/33across.go
```diff
@@ -38,39 +55,45 @@ func (a *TtxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters
...
 
-// Update the request object to include custom value
-// site.id
-func (a *TtxAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
-	var errs []error
+func (a *TtxAdapter) makeRequest(request *openrtb.BidRequest, imp openrtb.Imp, TTXRequestExt []byte) (*adapters.RequestData, error) {
+	var imps []openrtb.Imp
 
-	// Make a copy as we don't want to change the original request
 	reqCopy := *request
```
mansinahar, line 82: Nitpick: This is fine but if you are using just the copy in this method, I'd suggest passing the request to this method as value than as reference. That'd have the same effect

adapters/33across/33across.go
```diff
@@ -81,63 +104,84 @@ func (a *TtxAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.Request
...
+	if imp.Banner == nil && imp.Video == nil {
+		return openrtb.Imp{}, &errortypes.BadInput{
+			Message: fmt.Sprintf("Imp ID %s must have at least one of [Banner, Video] defined", imp.ID),
+		}
+	}
+
 	var bidderExt adapters.ExtImpBidder
 	if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
-		return &errortypes.BadInput{
+		return openrtb.Imp{}, &errortypes.BadInput{
```
mansinahar, line 119: Why not just return an imp pointer in this function and accordingly return `nil` here and below when you encounter an error? You can then just dereference it above when appending it to the imp array. Right now, we're creating these empty imp instances which aren't really gonna be used and will just take up memory until they are garbage collected

adapters/33across/33across.go
```diff
@@ -81,63 +104,84 @@ func (a *TtxAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.Request
...
-
 	// Validate Video if it exists
 	if imp.Video != nil {
 		videoCopy, err := validateVideoParams(imp.Video, impExt.Ttx.Prod)
 
 		imp.Video = videoCopy
 
 		if err != nil {
-			return &errortypes.BadInput{
+			return imp, &errortypes.BadInput{
```
mansinahar, line 156: Should this return statement be returning an empty `imp` object/nil? In case of error, you anyway discard the imp and move on to the next one

adapters/33across/33across.go
```diff
@@ -38,39 +55,45 @@ func (a *TtxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters
 	var errs []error
 	var adapterRequests []*adapters.RequestData
 
-	adapterReq, errors := a.makeRequest(request)
-	if adapterReq != nil {
-		adapterRequests = append(adapterRequests, adapterReq)
+	// Construct request extension common to all imps
+	// NOTE: not blocking e
```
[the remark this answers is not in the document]
curlyblueeagle: Yes sorry not sure how it got mangled, I will fix it.

adapters/33across/33across.go
```diff
@@ -38,39 +55,45 @@ func (a *TtxAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters
...
+	TTXRequestExt, err := makeReqExt(request)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	// Break up multi-imp request into multiple external requests since we don't
+	// support SRA in our exchange server
+	for i := 0; i < len(request.Imp); i++ {
+		// makeRequest will either return an adapterRequest or err
```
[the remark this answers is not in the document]
curlyblueeagle: Oh interesting, why?

adapters/33across/33across.go
curlyblueeagle: Ah gotcha, wasn't sure how to return nil values with structs (sorry, new to Golang). Will do this.

adapters/33across/33across.go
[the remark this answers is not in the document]
SyntaxNode: That's the default expectation in Go. Line 69 has a very standard `if` statement, getting back a value and an error from the method call. I see that you're only using the value if there is no error.. which is perfectly idiomatic. You can leave the comment in if you'd like. I just didn't find it add anything to my understanding of the code.

curlyblueeagle:
@SyntaxNode @mansinahar I have made the changes recommended. As a newbie to Golang (coming from JS land :)) , I appreciate your CR feedback and patience, very helpful. Please let me know if there's anything more.

SyntaxNode (approved):

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2557

adapters/33across/33across.go
mansinahar: No worries at all :) Thanks for addressing it

mansinahar (approved):

bretg:
@curlyblueeagle - we notice that your adapter accepts parameters on ext.33across

1) How would values get placed here? It's not possible with Prebid.js. Possibly some PBS host companies might be able to set those values, but not all of them. (e.g. Magnite cannot)
2) The standard location we'd proposed in #1042 for setting cross-request values would be ext.prebid.bidderparams.33across

PR #1611, opened 2020-12-04
Add a BidderRequest struct to hold bidder specific request info

mansinahar:
Adds a new BidderRequest struct to hold the bidder specific request information and modifies methods downstream to refactor the arguments passed to them


exchange/exchange.go
```diff
@@ -107,6 +109,15 @@ type AuctionRequest struct {
 	LegacyLabels pbsmetrics.Labels
 }
 
+// BidderRequest holds the bidder specific request and all other
+// information needed to process that bidder request.
+type BidderRequest struct {
```
mansinahar, line 114: I went back and forth on this name a few times. I don't really like the redundancy in the names here such as `BidderRequest.BidRequest`, `BidderRequest.BidderName` and such but the name `Bidder` doesn't convey much information and the other name that I played with was `AuctionRequestBidder` but didn't like that much either so went back to `BidderRequest`. Happy to consider any other suggestions if there are.

SyntaxNode: I like the `BidderRequest` request. You could remove the Bid/Bidder prefixes on the fields to reduce the naming redundancy, like:

```
type BidderRequest struct {
	Request     *openrtb.BidRequest
	Name     openrtb_ext.BidderName
	CoreName openrtb_ext.BidderName
	Labels   pbsmetrics.AdapterLabels
}
```

exchange/exchange.go
```diff
@@ -134,12 +145,12 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
 	blabels := make(map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels)
-	cleanRequests, aliases, privacyLabels, errs := cleanOpenRTBRequests(ctx, r.BidRequest, requestExt, r.UserSyncs, blabels, r.LegacyLabels, e.gDPR, usersyncIfAmbiguous, e.privacyConfig, &r.Account)
+	bidderRequests, privacyLabels, errs := cleanOpenRTBRequests(ctx, r, requestExt, blabels, e.gDPR, usersyncIfAmbiguous, e.privacyConfig)
```
SyntaxNode, line 148: See my comment just after this one. We no longer need `blabels` since `bidderRequests` comtains the blabel for each bidder.

exchange/exchange.go
```diff
@@ -322,37 +333,39 @@ func (e *exchange) makeAuctionContext(ctx context.Context, needsCache bool) (auc
 }
 
 // This piece sends all the requests to the bidder adapters and gathers the results.
-func (e *exchange) getAllBids(ctx context.Context, cleanRequests map[openrtb_ext.BidderName]*openrtb.BidRequest, aliases map[string]string, bidAdjustments map[string]float64, blabels map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels, conversions currencies.Conversions) (map[openrtb_ext.BidderName]*pbsOrtbSeatBid, map[openrtb_ext.BidderName]*seatResponseExtra, bool) {
+func (e *exchange) getAllBids(ctx context.Context, bidderRequests []BidderRequest,
+	bidAdjustments map[string]float64, blabels map[openrtb_ext.BidderName]*pbsmetrics.AdapterLabels,
```
SyntaxNode, line 337: `blabels` is now unused.

exchange/exchange.go
mansinahar: I considered this idea but the issue with that though is then the field names don't really make sense based on the object name i.e. the field `Name` or `CoreName` doesn't belong to a bidder _request_. It belongs to a _bidder_. And using just `Name` and `CoreName` with `BidderRequest` as the object name doesn't convey that. This is where I had moved towards using just `Bidder` for the object name but then that didn't make enough sense for this object name either

exchange/exchange.go
mansinahar: Oops looks like I missed removing that. Good catch. Will update the PR.

exchange/exchange.go
SyntaxNode: Gotcha. Let's just fo with this then. It's an internal data structure which we can always rename again in the future if we think of something better.

SyntaxNode (approved):

bsardo (approved):
LGTM

PR #1612, opened 2020-12-04
Add config gdpr.amp_exception deprecation warning

bsardo:


SyntaxNode (approved):

config/config.go
```diff
@@ -214,6 +214,9 @@ func (cfg *GDPR) validate(errs []error) []error {
 	if cfg.HostVendorID < 0 || cfg.HostVendorID > 0xffff {
 		errs = append(errs, fmt.Errorf("gdpr.host_vendor_id must be in the range [0, %d]. Got %d", 0xffff, cfg.HostVendorID))
 	}
+	if cfg.AMPException == true {
+		glog.Warning("gdpr.amp_exception is deprecated and will be removed in a future version. If you need to disable GDPR for AMP, you may do so on a per-account basis or disable it for all request types at the host level.")
```
mansinahar, line 218: Can we please include where they can set this as you did in the deprecation comment above?

mansinahar (commented):
LGTM other than the minor comment below

config/config.go
SyntaxNode: How about:
`glog.Warning("gdpr.amp_exception is deprecated and will be removed in a future version. If you need to disable GDPR for AMP, you may do so per-account (gdpr.integration_enabled.amp) or at the host level for the default account (account_defaults.gdpr.integration_enabled.amp).")`

mansinahar: Yup! That looks great 👍

SyntaxNode (approved):

mansinahar (approved):

PR #1613, opened 2020-12-04
Add TLS Handshake connection metrics

mansinahar:


pbsmetrics/prometheus/prometheus.go
```diff
@@ -257,6 +258,11 @@ func NewMetrics(cfg config.PrometheusMetrics, disabledMetrics config.DisabledMet
 		"Seconds to resolve DNS",
 		standardTimeBuckets)
 
+	metrics.tlsHandhakeTimer = newHistogram(cfg, metrics.Registry,
+		"tls_handshake_time",
+		"Seconds to perform TLS Handshake",
+		standardTimeBuckets)
```
SyntaxNode, line 264: This time bucket profile maxes out at 1 second. Does that make sense for the values you expect for this metric? Can we get away with less buckets perhaps? Not a big deal since this isn't per-bidder.

mansinahar: That's a good question. So for the total connection wait time we have the same buckets and I expect the TLS Handshake time to be less than the total connection time and so I kept them as the same. Your question makes me wonder what happens if the value ever surpasses the 1sec time. I am guessing it gets put in the last bucket and that I think is okay because at that point I think it suffices for us to know that it is taking 1sec or more which is bad and we need to figure out ways to improve it.

About the fewer buckets, as you mentioned that since this not at a per bidder level, I think it would help to have that granularity to get the values as close to actual as possible. Also, the other reason is that I want to keep the buckets across all connection metrics consistent because otherwise comparing those values can be confusing. Like for example  if we had fewer buckets for TLS Handshake then I can imagine there being times where the TLSHandshake time for a connection might fall in a bigger bucket and reflect a value more than the total connection time which can be misleading.

bsardo (approved):

SyntaxNode (approved):

PR #1617, opened 2020-12-09
Deepintent: Params normalization

sourabhg:
Normalized Param names as per prebid js (requirement from client)

Changed the maintainer email address to DL (requirement from Magnite)


adapters/deepintent/deepintent.go
```diff
@@ -21,7 +21,7 @@ type DeepintentAdapter struct {
 }
 
 type deepintentParams struct {
-	TagID string `json:"tagId"`
+	tagId string `json:"tagId"`
```
SyntaxNode, line 24: This change isn't really necessary. I'm fine if you want to keep it this way or how it was before.

The capitalization of the first character of a field in Go is the access modifier. Upper case = exported for other packages. Lower case = package private. Since the struct is already package private, the access modifier doesn't really matter.

The json serializer will use the name `tagId` from the json attribute.

static/bidder-info/deepintent.yaml
```diff
@@ -1,5 +1,5 @@
 maintainer:
-  email: "sourabh@deepintent.com"
+  email: "prebiddev@deepintent.com"
```
SyntaxNode, line 2: Not sure how we missed that one. Thank you for changing this to a mailing group instead of an individual contact.

static/bidder-params/deepintent.json
```diff
@@ -5,10 +5,10 @@
...
   "properties": {
-    "TagID": {
+    "tagId": {
       "type": "string",
       "description": "An ID which identifies the deepintent ad tag"
     }
   },
-  "required" : [ "TagID" ]
+  "required" : [ "tagId" ]
 }
```
SyntaxNode, line 14: This is a potentially dangerous change to make as it will invalidate the configuration of any publisher currently using your adapter. Such publishers would have their entire request rejected by Prebid Server and unable to use any demand partner until they update.

This is a relative new adapter though, so perhaps it's possible no one is using it yet?

SyntaxNode: FYI: I don't see any traffic yet for this adapter, so you're probably still in the "safe zone" unless you know of publishers using it currently.

sourabhg: Yes the changes have been made to make params same as prebid js params, and as per requirement of a client itself. So its safe to merge

adapters/deepintent/deepintent.go
sourabhg: This is changed to keep in sync the standard names in adapter. I think this should be fine for now.

static/bidder-info/deepintent.yaml
sourabhg: No Problem 😄

SyntaxNode (approved):

bsardo:
LGTM. I missed hitting the approved button before the merge.

bsardo (commented):
LGTM. I missed hitting the approved button before the merge.

PR #1622, opened 2020-12-11
New Adapter: Revcontent

jcamp-revc:
Hello!

I'm looking to add support for Revcontent ORTB to prebid-server. We really just need a [generic adapter](https://github.com/prebid/prebid-server/issues/1429), but from the discussion I don't think that's a thing yet.

I'm fairly new to go and this project, so just let me know what I need to fix. :)


adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"net/http"
+)
+
+type RevcontentAdapter struct {
```
SyntaxNode, line 14: Nitpick: We prefer the name `adapter`, as the object is referenced by the `adapters.Bidder` interface outside of this package. There's not need to duplicate the name of the package or export it (start with an uppercase letter).

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,124 @@
...
+// Builder builds a new instance of the Revcontent adapter for the given bidder with the given config.
+func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
+	bidder := &RevcontentAdapter{
+		endpoint: config.Endpoint,
+	}
+	return bidder, nil
+}
+
+func (adapter *RevcontentAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var adapterRequests []*adapters.RequestData
```
SyntaxNode, line 27: This method is really just calling the `makeRequest` and adds no new logic. I recommend moving the content of the `makeRequest` method here, such as:

```
func (a *adapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
	reqBody, err := json.Marshal(request)

	if err != nil {
		return nil, []error{err}
	}

	headers := http.Header{}

	req := &adapters.RequestData{
		Method:  "POST",
		Uri:     a.endpoint,
		Body:    reqBody,
		Headers: headers,
	}
	return []*adapters.RequestData{req}, nil
}
```

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,124 @@
...
+
+// MakeBids make the bids for the bid response.
+func (a *RevcontentAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if response.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if response.StatusCode == http.StatusBadRequest {
+		return nil, []error{&errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: %d.", response.StatusCode),
```
SyntaxNode, line 65: Nitpick: In Go, it's preferable for error messages to be start a lower case letter and not end with punctation. 
`"unexpected status code: %d"`

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	var bidResp openrtb.BidResponse
+
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+
+	count := getBidCount(bidResp)
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(count)
```
SyntaxNode, line 82: It's not worth counting the bids to use an exact number here, IMHO. The effort of counting likely won't offset the small memory performance increase by specifying a guess. Set this to the number of bids your server sends back on average.

usersync/usersyncers/syncer_test.go
```diff
@@ -109,6 +109,7 @@ func TestNewSyncerMap(t *testing.T) {
 		openrtb_ext.BidderTappx:        true,
 		openrtb_ext.BidderYeahmobi:     true,
 		openrtb_ext.BidderAdprime:      true,
+		openrtb_ext.BidderRevcontent:   true,
```
SyntaxNode, line 112: Please add your adapter alphabetically to this list. Although to be fair, it seems to be in need of sorting.

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func getBidType(impId string, imps []openrtb.Imp) openrtb_ext.BidType {
+	bidType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Banner != nil {
+				break
+			}
+			if imp.Video != nil {
```
SyntaxNode, line 112: Your bidder doesn't support video, so this case doesn't seem possible.

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,124 @@
...
+				break
+			}
+			if imp.Native != nil {
+				bidType = openrtb_ext.BidTypeNative
+				break
+			}
+
+		}
+	}
+	return bidType
```
SyntaxNode, line 123: It would be ideal to use a response from your bidder to set the media type. Does your bidder include any ext values to use in that determination? If not, what would your bidder respond back with if the impression is for both banner and native?

SyntaxNode (commented):
Looks great. Thank you for your contribution. I have one comment on determining bid type and the rest are all syntax nitpicks. Test coverage looks great.

Seems like you followed the new docs. I'm curious how that worked out for you or if you recommend any improvements.

jcamp-revc:
Thanks for the feedback. I think this should all be addressed in the latest commit. The new docs were very helpful, thank you!

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,99 @@
...
+				Bid:     &sb.Bid[i],
+				BidType: mediaType,
+			})
+		}
+	}
+	return bidResponse, nil
+
+}
+
+func getBidCount(bidResponse openrtb.BidResponse) int {
```
mansinahar, line 84: Looks like this function is unused. Can you please remove it?

adapters/revcontent/revcontenttest/params/race/banner.json
```diff
@@ -0,0 +1,3 @@
+{
+}
```
mansinahar, line 2: If your adapter doesn't support any parameters then you don't need these params files. You can remove these.

mansinahar (commented):
Except for the following minor comments, LGTM!

jcamp-revc:
> Except for the following minor comments, LGTM!

Thanks, these should be addressed now.

static/bidder-params/revcontent.json
```diff
@@ -0,0 +1,9 @@
+
```
SyntaxNode, line 1: Nitpick: Please remove the leading line break.

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"net/http"
+)
+
+type Adapter struct {
```
SyntaxNode, line 14: This doesn't need to be exported and should therefore start with a lower case letter. Please rename to `adapter`.

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,91 @@
...
+
+}
+
+func getBidType(bidAdm string) openrtb_ext.BidType {
+	// native: {"ver":"1.1","assets":...
+	// banner: <div id='rtb-widget...
+	if bidAdm != "" && bidAdm[:1] == "<" {
+		return openrtb_ext.BidTypeBanner
+	}
+	return openrtb_ext.BidTypeNative
```
SyntaxNode, line 90: Nice. Thank you.

static/bidder-params/revcontent.json
```diff
@@ -0,0 +1,9 @@
...
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Revcontent Adapter Params",
+  "description": "A schema which validates params accepted by the Revcontent adapter",
+  "type": "object",
+  "properties": {},
```
SyntaxNode, line 7: Does your adapter truly have no bidder parameters? If so, not a problem.

If you do have bidder parameters but aren't defining them here because you pass the OpenRTB request exactly as is, please consider adding them here to allow PBS to enforce the parameters are provided and prevent outgoing traffic to your bidding server if the request is invalid.

SyntaxNode (commented):
Looks great with ideal test coverage. Just a few more nitpicks and a question.

static/bidder-params/revcontent.json
jcamp-revc: whoops.

static/bidder-params/revcontent.json
jcamp-revc: Do you mean things like `bidfloor`? Then yes, I suppose I'll have to look at this. Thanks!

jcamp-revc: We really just need to look at:
(request.App.name OR request.Site.domain) AND request.Device, and everything else is optional on our end.

Do we need to enforce these in a check following this example: https://github.com/prebid/prebid-server/blob/master/adapters/pubnative/pubnative.go#L80-L88

Or is there a way to use the params.json to check those? We don't currently need any ext params or anything.

SyntaxNode: I wouldn't say you need to, but I do highly recommend it. The earlier we can disqualify a bid request the less resources we all have to use to accomodate it. In this case, PBS hosts wouldn't need to reach out to your bidding servers for requests which are known to be invalid.

This must be coded in the adapter, as you linked to for pubnative. The params.json file is just to check bidder extensions not define within the OpenRTB spec.

Thank you for the details. This json file is fine as is. I do encourage you to add the checks you mentioned along with new json tests cases to cover the new logic.

SyntaxNode:
Removing stale label to end of year holidays.

jcamp-revc:
Thanks, I'll get those checks added and then hopefully we can merge soon.

adapters/revcontent/revcontent.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func checkRequest(request *openrtb.BidRequest) error {
+	if (request.App == nil || len(request.App.Name) == 0) && (request.Site == nil || len(request.Site.Domain) == 0) {
+		return &errortypes.BadInput{
+			Message: "Impression is missing app name or site domain, and must contain one.",
+		}
+	}
+
+	if request.App != nil && request.Site != nil {
```
SyntaxNode, line 56: This check is performed by the "core" and isn't necessary here. See: https://github.com/prebid/prebid-server/blob/a709baa4f842b11d4711b642c68ee245639dde07/endpoints/openrtb2/auction.go#L336

SyntaxNode (commented):
Thank you for adding the validation. One the checks is already covered by "core" code and doesn't need to be repeated.

adapters/revcontent/revcontent.go
jcamp-revc: sorry, not sure how that got missed. Fixing.

SyntaxNode:
Thank you for the changes. Looks like it's failing on `app_and_site.json` and that's ok. The adapter json unit tests bypass the core checks. You can resolve by removing that test. We have coverage elsewhere.

jcamp-revc:
> Thank you for the changes. Looks like it's failing on `app_and_site.json` and in retrospec that's expected. The adapter json unit tests bypass the core checks. You can resolve by removing that test. We have coverage elsewhere.

That's what I get for not running the tests, even on a simple change. Sorry about that.

SyntaxNode (dismissed):

SyntaxNode:
LGTM. Thank you. I'll seek a second review tomorrow and there's a very good chance this can be incorporated into the next release. Before we can merge, we just need you to open the documentation PR. For details, please see the bottom of the new adapter docs: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#user-documentation

bsardo (dismissed):
LGTM

SyntaxNode:
@jcamp-revc We're ready to merge in once you have a PR to update your adapter docs.

jcamp-revc:
Thanks, [PR is opened](https://github.com/prebid/prebid.github.io/pull/2614).

SyntaxNode:
@jcamp-revc Thank you.

I noticed that "Revcontent will provide you with a custom endpoint that will contain your account details". With this in mind, does the default URL provided in this PR work? .. or should this adapter be disabled by default until hosts work with Revcontent to set it up?

jcamp-revc:
I tried to disable it, but was having trouble with getting tests to pass. Let me try again.

config/config.go
```diff
@@ -980,6 +980,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.pubnative.endpoint", "http://dsp.pubnative.net/bid/v1/request")
 	v.SetDefault("adapters.pulsepoint.endpoint", "http://bid.contextweb.com/header/s/ortb/prebid-s2s")
 	v.SetDefault("adapters.revcontent.endpoint", "https://trends.revcontent.com/rtb")
+	v.SetDefault("adapters.revcontent.disabled", "true")
```
bsardo, line 982: The second parameter here should be a boolean instead a string.

jcamp-revc: Thanks. Sorry about that.

SyntaxNode: No worries. Thanks for fixing it. There's just a pesky merge conflict now.

jcamp-revc: Is that for me to fix, or you?

SyntaxNode: That's for you to fix. Please merge in the master branch and resolve the conflict. Looks like a super minor issue with the user sync tests.

SyntaxNode (approved):

bsardo (approved):

terrapop:
Does this adapter work? Seems it sends out ad requests, but we don't get any response from Revconent at all. Our setup looks ok, and we used the correct values for the API key and account ID.

SyntaxNode:
@terrapop If you don't receive a response from @jcamp-revc here, please reach out to their support email address developers@revcontent.com.

We approved the adapter based on the sample requests included in this PR. We don't have test accounts or test placements with Revcontent to be able to verify further. It's the responsibility of bidders to perform end to end testing on their submissions.

jcamp-revc:
> Does this adapter work? Seems it sends out ad requests, but we don't get any response from Revconent at all. Our setup looks ok, and we used the correct values for the API key and account ID.

@terrapop  Hi. I'm happy to take a look at this. Can we move over to email at jon.camp@revcontent.com and I'll help you look into it.

Issue #1621, opened 2020-12-09
eids permissions


bretg:
The Prebid Server side of https://github.com/prebid/Prebid.js/issues/5814 -- we need to support permissioning of eids data.

1. PBS could receive ext.prebid.data.eidPermissions which is expected to be an array of objects
```
{
    ext: {
        prebid: {
            data: {
                eidpermissions: [   // prebid server will use this to filter user.ext.eids
                   {"source": "sharedid.org", "bidders": ["*"]},  // * is the default
                   {"source": "neustar.biz", "bidders": ["bidderB"]},
                   {"source": "id5-sync.com", "bidders": ["bidderA","bidderC"]}
                ]
            }
        }
    }
}
```
2. If eidPermissions is not an array, ignore it and emit a 400 error with a reasonable description in the body
3. If there's no "source" attribute in the object in each array entry, emit a 400 error with a reasonable description in the body
4. If there's no "bidders" attribute in the object in each array entry, emit a 400 error with a reasonable description in the body
5. If the value of the "bidder" attribute is not an array of strings, emit a 400 error with a reasonable description in the body
6. If one of the bidder values is an asterisk, it means this source should be passed to all bidders
7.  If a bidder value is an unrecognized biddercode, emit a 400 error with a reasonable description in the body
8. If there are no valid biddercodes in the array, emit a 400 error with a reasonable description in the body
9. If the same source is listed more than once in the array, emit a 400 error with a reasonable description in the body
10. If there are any valid eidPermissions entries, PBS-core needs to handle them by filtering user.ext.eids before passing to each bidder. Loop through user.ext.eids
    1. If user.ext.eids.source doesn't exist in eidPermissions, all bidders can receive it
    2. If user.ext.eids.source exists in eidPermissions and the bidder value is "*", all bidders can receive it
    3. If user.ext.eids.source exists in eidPermissions with specific bidders, only those bidders can receive that entry

SyntaxNode:
> If a bidder value is an unrecognized biddercode, drop that biddercode and emit a warning

PBS currently returns an error for an unrecognized bidder code and we did the same for the CCPA no sale bidders array. The idea being the PBS will "fail fast and loud" at the sign of trouble and an unrecognized bidder code is either an invalid request setup or a missing alias.

Should we be consistent and return an error in this case too?

bretg:
[...]

PR #1633, opened 2020-12-21
Implement EID Permissions

SyntaxNode:
Implements: https://github.com/prebid/prebid-server/issues/1621


exchange/utils.go
```diff
@@ -445,6 +453,100 @@ func copyWithBuyerUID(user *openrtb.User, buyerUID string) *openrtb.User {
...
+		return err
+	}
+
+	eidsRaw, eidsSpecified := userExt["eids"]
+	if !eidsSpecified {
+		return nil
+	}
+
+	var eids []openrtb_ext.ExtUserEid
+	if err := json.Unmarshal(eidsRaw, &eids); err != nil {
```
guscarreon, line 480: Given that `extractBuyerUIDs` unmarshals the `User.Ext` into an `openrtb_ext.ExtUser` struct that already reads the `eids` into an `[]ExtUserEid` array, maybe we can work with this field and avoid additional unmarshalling inside `removeUnpermissionedEids`. Both `extractBuyerUIDs` and `removeUnpermissionedEids` use the unmarshalled `User.Ext`, therefore we could maybe:
```
174   func getAuctionBidderRequests(req AuctionRequest, requestExt *openrtb_ext.ExtRequest, impsByBidder map[string][]openrtb.I
175
176       bidderRequests := make([]BidderRequest, 0, len(impsByBidder))
177
    +     userExt := unmarshalUserExt(req.BidRequest.User)
    +
178 -     explicitBuyerUIDs, err := extractBuyerUIDs(req.BidRequest.User)
    +     explicitBuyerUIDs, err := extractBuyerUIDs(userExt)
179       if err != nil {
180           return nil, []error{err}
181       }
182
183   *-- 11 lines: var sChainsByBidder map[string]*openrtb_ext.ExtRequestPrebidSChainSChain---
194
195       var errs []error
196       for bidder, imps := range impsByBidder {
197           coreBidder := resolveBidder(bidder, aliases)
198
199   *--  5 lines: reqCopy := *req.BidRequest-------------------------------------------------
204
205           if err := removeUnpermissionedEids(&reqCopy, bidder, requestExt); err != nil {
    +         if err := removeUnpermissionedEids(&reqCopy, bidder, requestExt, userExt); err != nil {
206               errs = append(errs, fmt.Errorf("unable to enforce request.ext.prebid.data.eidpermissions because %v", err))
207               continue
208           }
exchange/utils.go
```
and modify `removeUnpermissionedEids` accordingly.

endpoints/openrtb2/auction.go
```diff
@@ -419,6 +423,51 @@ func validateSChains(req *openrtb_ext.ExtRequest) error {
 	return err
 }
 
+func (deps *endpointDeps) validateEidPermissions(req *openrtb_ext.ExtRequest, aliases map[string]string) error {
+	if req == nil || req.Prebid.Data == nil {
+		return nil
+	}
```
guscarreon, line 429: According to issue #1621: "If eidPermissions is not an array, ignore it and emit a 400 error with a reasonable description in the body". Unless I overlooked it, are we missing this validation? Should we do:
```
if len(req.Prebid.Data.EidPermissions) == 0{
    return fmt.Errorf(`....`)
}
endpoints/openrtb2/auction.go
```

guscarreon (commented):
A couple of comments

exchange/utils_test.go
```diff
@@ -1408,3 +1408,254 @@ func TestBidderToPrebidChainsZeroLengthSChains(t *testing.T) {
...
+			expectedUserExt: json.RawMessage(`{"eids":[],"other":42}`),
+		},
+		{
+			description:     "Allowed By Nil Permissions",
+			userExt:         json.RawMessage(`{"eids":[{"source":"source1","id":"anyID"}]}`),
+			eidPermissions:  []openrtb_ext.ExtRequestPrebidDataEidPermission{},
+			expectedUserExt: json.RawMessage(`{"eids":[{"source":"source1","id":"anyID"}]}`),
+		},
+		{
+			description:     "Allowed By Empty Permissions",
```
hhhjort, line 1468: What is the difference between nil permissions and empty permissions?

SyntaxNode: As of now, nothing. Good catch. This was a copy/paste error. Fixed.

hhhjort (approved):
LGTM

guscarreon (approved):
LGTM

PR #1636, opened 2020-12-28
Audit tests

muncha:

If, when I submit this PR, it fails validation, I will immediately close it and re-examine. I have run .`/validation.sh --cov --nofmt --race 10` (and --race 20) several times without a fail. I've also rebuilt my dev environment from intellij / running on my system (arch linux) to the Visual Studio / docker setup described in the developer documentation (and canceled my intellij subscription renewal as Visual Studio beats the pants off it).

So, fingers crossed, here it goes...


muncha:
I sure do have a big mess of commits dating back to 2018 in this. Really these changes start at 9facef5. Should I do a a squash of some kind or otherwise clean this up before preceding with a review?

muncha:
That commit mess really bothers me. I'm going to redo this PR with a new branch from the fresh master. BRB.

PR #1637, opened 2020-12-29
Audit tests2

muncha:

- I have removed the check for Imp[n].banner.Size.H and W as it appears that option has been deprecated in favor of Imp[n].banner.format, as long anticipated, and checking for it now throws an error. 


muncha:
I'm still not clear on why I have a bunch of old commits in here. I switched to my most recent master and did a pull from the beachfront fork, then pulled master from here, and copied by changes (b73acca) over from a copy I had made. Maybe the beachfront fork had some stuff in master that I had never PR'd to the primary? Does this create any problems? I could just do a fresh clone of your master, i guess that would be a fresh fork, and copy these changes in.

Yeah, I'm going to do that. I notice that fde9619, for example, references docs/bidders/beachfront.md, which doesn't even exist any more. So I'm going to close this and start from a fresh fork. I've accumulated a lot of junk in here.

PR #1638, opened 2020-12-29
Audit beachfront tests and change some videoResponseType details

muncha:

- I have removed the check for Imp[n].banner.Size.H and W as it appears that option has been deprecated in favor of Imp[n].banner.format, as long anticipated, and checking for it now throws an error. 


muncha:
This looks like the clean PR I was looking for. Ready for review.

SyntaxNode:
> This looks like the clean PR I was looking for. Ready for review.

Fantastic! We're out this week on vacation. We'll review early next week. Happy New Year.

hhhjort (dismissed):
looks good to me

adapters/beachfront/beachfronttest/noop/six-nine-combo--response-order--banner-adm-nurl-adm--FAIL.json
```diff
@@ -0,0 +1,458 @@
+{
```
bsardo, line 1: I think the `noop` directory and the test files in it should be removed since they aren't working tests. You can always add them to the `supplemental` folder in another PR when you have them passing.

bsardo (commented):
Nice job. This is looking good. I just have the one request to remove the `noop` test directory.

Aside from that, I just want to point out that if you ever feel inspired to boost your test coverage, you could do so in another PR at your convenience by adding the following JSON test cases:
- 400 bad request response from bidder
- 500 internal server error response from bidder
- bidder response unmarshal error
- PBS request with no imps
- PBS request with banner imp but missing `ext.bidder.appId`
- PBS request with video imp but missing `ext.bidder.appId`
- `ext.bidder.bidfloor` < `minBidFloor`
There's also missing coverage of default values when constructing a bid request in `getBidRequest` and `getVideoRequests`.

adapters/beachfront/beachfronttest/noop/six-nine-combo--response-order--banner-adm-nurl-adm--FAIL.json
muncha: Will do. The big thing that I want to get working in this is the ability for the bidder to send requests to beachfront's video endpoint as a single request with multiple imps. As it is it sends one request for each video imp, which is obviously a waste of resources. That is what I'm build up to. Before that, I'll push with the noop directory removed for this PR, then another PR with your suggested tests above.

hhhjort (approved):

bsardo (approved):
LGTM

PR #1642, opened 2021-01-06
Fix appnexus adapter not setting currency in the bid response

SyntaxNode:
It has been reported that appnexus may not be behaving properly in foreign currency contexts. It appears that the adapter does not set the curr (currency), field in the bid response. This pull request makes this correction and adds a JSON test file but proper assertion of the Currency field of the BidderResponse inside adapters/adapterstest/test_json.go will be implemented in a separate PR being developed by @mansinahar

NOTE: Recreated from https://github.com/prebid/prebid-server/pull/1598 to solve strange merge state issue.


hhhjort (approved):
LGTM

bsardo (approved):
LGTM

PR #1646, opened 2021-01-08
Fix Shared Memory Corruption In EMX_Digital

SyntaxNode:
The adapter changes a shared video pointer without first making a copy. We have reason to believe this is causing occasional memory corruption issues when marshaling a request targeting multiple adapters.

@EMXDigital Can you please review?

### DIFF
--- a/adapters/emx_digital/emx_digital.go
+++ b/adapters/emx_digital/emx_digital.go
@@ -152,8 +152,10 @@ func buildImpVideo(imp *openrtb.Imp) error {
 		}
 	}
 
-	if imp.Video.Protocols != nil {
-		imp.Video.Protocols = cleanProtocol(imp.Video.Protocols)
+	if len(imp.Video.Protocols) > 0 {
+		videoCopy := *imp.Video
+		videoCopy.Protocols = cleanProtocol(imp.Video.Protocols)
+		imp.Video = &videoCopy
 	}
 
 	return nil


guscarreon (commented):
Pretty good catch. Just to unit test this change, should we include something like:
```
func TestBuildImpVideo(t *testing.T) {
	testVideoImp := &openrtb.Imp{
		Video: &openrtb.Video{
			MIMEs: []string{"a_mime"},
			W:     480,
			H:     320,
			Protocols: []openrtb.Protocol{
				openrtb.ProtocolVAST10,
				openrtb.ProtocolVAST20,
				openrtb.ProtocolVAST30,
			},
		},
	}
	err := buildImpVideo(testVideoImp)

	//assert no error
	assert.NoError(t,err,"error message")

	//assert testVideoImp was not modified
	assert.Equal(/*...*/)
}
```
I realize it doesn't add coverage, but it does explicitly unit-tests this modification. What do you think?

guscarreon (approved):

SyntaxNode:
> I realize it doesn't add coverage, but it does explicitly unit-tests this modification. What do you think?

I like the idea. We have the race condition tests to try and catch these issues, but the coverage is too low. I think it would be better to add assertions to the main adapter json framework to identify if shared memory is altered. They already have high test coverage so that would likely do much better... once time allows, of course.

hhhjort (approved):
LGTM

SyntaxNode:
Don't merge yet. Let's give @EMXDigital some time to see this PR.

SyntaxNode:
I've been unable to reach @EMXDigital through GitHub mentions and their support email. I'm going to proceed to merge this since it fixes a panic scenario.

EMXDigital:
Hey @SyntaxNode , appreciate you finding this issue and proposing the solution as well. For some reason this particular Git thread didn't make it to our attention until today but I'm glad to see everything both looks good from an engineering standpoint as well as Prebid flow. 

LGTM

PR #1648, opened 2021-01-11
Eplanning: new prioritization metric for adunit sizes

chino117:
### Type of change:
 Feature
### Description of change:
New prioritization metric for adunit sizes

### contact email of the adapter’s maintainer:
chino117@hotmail.com

### DIFF
--- a/adapters/eplanning/eplanning.go
+++ b/adapters/eplanning/eplanning.go
@@ -27,7 +27,6 @@ const sec = "ROS"
 const dfpClientID = "1"
 const requestTargetInventory = "1"
 
-var isMobile = -1
 var priorityOrderForMobileSizesAsc = []string{"1x1", "300x50", "320x50", "300x250"}
 var priorityOrderForDesktopSizesAsc = []string{"1x1", "970x90", "970x250", "160x600", "300x600", "728x90", "300x250"}
 
@@ -66,21 +65,7 @@ type hbResponseAd struct {
 	Height       uint64 `json:"h,omitempty"`
 }
 
-type byPriority []openrtb.Format
-
-func (a byPriority) Len() int      { return len(a) }
-func (a byPriority) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
-func (a byPriority) Less(i, j int) bool {
-	var priorityOrderForSizesAsc []string
-	if isMobile > 0 {
-		priorityOrderForSizesAsc = priorityOrderForMobileSizesAsc
-	} else {
-		priorityOrderForSizesAsc = priorityOrderForDesktopSizesAsc
-	}
-	size1 := fmt.Sprintf("%dx%d", a[i].W, a[i].H)
-	size2 := fmt.Sprintf("%dx%d", a[j].W, a[j].H)
-	index1 := indexOf(size1, priorityOrderForSizesAsc)
-	index2 := indexOf(size2, priorityOrderForSizesAsc)
+func isPriority(index1 int, index2 int) bool {
 	if index1 > -1 {
 		if index2 > -1 {
 			if index1 < index2 {
@@ -100,17 +85,42 @@ func (a byPriority) Less(i, j int) bool {
 	}
 }
 
+func lessByPriority(w1, h1, w2, h2 uint64, priorityOrderForSizesAsc []string) bool {
+	size1 := fmt.Sprintf("%dx%d", w1, h1)
+	size2 := fmt.Sprintf("%dx%d", w2, h2)
+	index1 := indexOf(size1, priorityOrderForSizesAsc)
+	index2 := indexOf(size2, priorityOrderForSizesAsc)
+
+	return isPriority(index1, index2)
+}
+
+type byPriorityMobile []openrtb.Format
+
+func (a byPriorityMobile) Len() int      { return len(a) }
+func (a byPriorityMobile) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a byPriorityMobile) Less(i, j int) bool {
+	return lessByPriority(a[i].W, a[i].H, a[j].W, a[j].H, priorityOrderForMobileSizesAsc)
+}
+
+type byPriorityDesktop []openrtb.Format
+
+func (a byPriorityDesktop) Len() int      { return len(a) }
+func (a byPriorityDesktop) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a byPriorityDesktop) Less(i, j int) bool {
+	return lessByPriority(a[i].W, a[i].H, a[j].W, a[j].H, priorityOrderForDesktopSizesAsc)
+}
+
 func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
 	errors := make([]error, 0, len(request.Imp))
 	totalImps := len(request.Imp)
 	spacesStrings := make([]string, 0, totalImps)
 	totalRequests := 0
 	clientID := ""
-	isMobile = isMobileDevice(request)
+	isMobile := isMobileDevice(request)
 
 	for i := 0; i < totalImps; i++ {
 		imp := request.Imp[i]
-		extImp, err := verifyImp(&imp)
+		extImp, err := verifyImp(&imp, isMobile)
 		if err != nil {
 			errors = append(errors, err)
 			continue
@@ -227,18 +237,8 @@ func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqIn
 	return requests, errors
 }
 
-func isMobileDevice(request *openrtb.BidRequest) int {
-	if request.Device != nil {
-		if request.Device.DeviceType == openrtb.DeviceTypePersonalComputer {
-			return 0
-		}
-		if request.Device.DeviceType == openrtb.DeviceTypeMobileTablet ||
-			request.Device.DeviceType == openrtb.DeviceTypePhone ||
-			request.Device.DeviceType == openrtb.DeviceTypeTablet {
-			return 1
-		}
-	}
-	return -1
+func isMobileDevice(request *openrtb.BidRequest) bool {
+	return request.Device != nil && (request.Device.DeviceType == openrtb.DeviceTypeMobileTablet || request.Device.DeviceType == openrtb.DeviceTypePhone || request.Device.DeviceType == openrtb.DeviceTypeTablet)
 }
 
 func cleanName(name string) string {
@@ -248,7 +248,7 @@ func cleanName(name string) string {
 	return name
 }
 
-func verifyImp(imp *openrtb.Imp) (*openrtb_ext.ExtImpEPlanning, error) {
+func verifyImp(imp *openrtb.Imp, isMobile bool) (*openrtb_ext.ExtImpEPlanning, error) {
 	var bidderExt adapters.ExtImpBidder
 
 	if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
@@ -271,7 +271,7 @@ func verifyImp(imp *openrtb.Imp) (*openrtb_ext.ExtImpEPlanning, error) {
 		}
 	}
 
-	width, height := getSizeFromImp(imp)
+	width, height := getSizeFromImp(imp, isMobile)
 
 	if width == 0 && height == 0 {
 		impExt.SizeString = nullSize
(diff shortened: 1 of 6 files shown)


adapters/eplanning/eplanning.go
```diff
@@ -26,6 +27,10 @@ const sec = "ROS"
 const dfpClientID = "1"
 const requestTargetInventory = "1"
 
+var isMobile = -1
```
SyntaxNode, line 30: Adapters must not modify package scope variables. Due to the highly concurrent nature of PBS, this will most certainly result in race conditions and panics. To solve this, I recommend splitting the sorter into two entites such as `byPriorityMobile` and `byPriorityDesktop` created by passing down the `isMobile` flag.

adapters/eplanning/eplanning.go
```diff
@@ -61,12 +66,47 @@ type hbResponseAd struct {
 	Height       uint64 `json:"h,omitempty"`
 }
 
+type byPriority []openrtb.Format
```
SyntaxNode, line 69: I recommend moving the sorter to its own file due to its non-trivial logic. I'd also recommend coded unit tests (not using the json framework) for this sorting behavior specifically.

adapters/eplanning/eplanning.go
```diff
@@ -61,12 +66,47 @@ type hbResponseAd struct {
...
+			return false
+		}
+	} else {
+		if index2 > -1 {
+			return true
+		} else {
+			return false
+		}
+	}
+}
```
SyntaxNode, line 101: Sorting algorithms are always fun to read. Could you please describe the algorithm to assist with the review?

I'm a bit concerned with the performance of building the `size1` and `size2` strings on each compare. Perhaps that burden can be reduced with memoization.. although before I perhaps prematurely worry about performance, how many formats do you typically see in requests?

adapters/eplanning/eplanning.go
```diff
@@ -187,6 +227,20 @@ func (adapter *EPlanningAdapter) MakeRequests(request *openrtb.BidRequest, reqIn
...
+			return 0
+		}
+		if request.Device.DeviceType == openrtb.DeviceTypeMobileTablet ||
+			request.Device.DeviceType == openrtb.DeviceTypePhone ||
+			request.Device.DeviceType == openrtb.DeviceTypeTablet {
+			return 1
+		}
+	}
+	return -1
+}
```
SyntaxNode, line 242: This method returns 3 states: mobile, desktop, or unknown. However, the sorter only cares if its mobile and groups desktop and unknown together. Consider keeping things simple and returning a boolean here.

adapters/eplanning/eplanning.go
```diff
@@ -232,13 +286,29 @@ func verifyImp(imp *openrtb.Imp) (*openrtb_ext.ExtImpEPlanning, error) {
...
+		if element == v {
+			return k
+		}
+	}
+	return -1
+}
+
+func getSizesSortedByPriority(sizeArray []openrtb.Format) []openrtb.Format {
+	sort.Sort(byPriority(sizeArray))
+	return sizeArray
```
SyntaxNode, line 300: Since `sort.Sort` is a "destructive" call to the array, there's no need to return it. Consider inlining `sort.Sort(byPriority(sizeArray))`.

adapters/eplanning/eplanning.go
```diff
@@ -232,13 +286,29 @@ func verifyImp(imp *openrtb.Imp) (*openrtb_ext.ExtImpEPlanning, error) {
...
 func getSizeFromImp(imp *openrtb.Imp) (uint64, uint64) {
 	if imp.Banner.W != nil && imp.Banner.H != nil {
 		return *imp.Banner.W, *imp.Banner.H
 	}
 
 	if imp.Banner.Format != nil {
-		for _, format := range imp.Banner.Format {
+		sizesSortedByPriority := getSizesSortedByPriority(imp.Banner.Format)
+		for _, format := range sizesSortedByPriority {
+			fmt.Println(format)
```
SyntaxNode, line 311: Was this print line for debugging purposes? Please remove it before merging in.

adapters/eplanning/eplanning.go
```diff
@@ -61,16 +65,62 @@ type hbResponseAd struct {
...
+			return false
+		}
+	} else {
+		if index2 > -1 {
+			return true
+		} else {
+			return false
+		}
+	}
+}
```
SyntaxNode, line 86: Some of the comparisons can be simplified:

```
func isPriority(index1 int, index2 int) bool {
	if index1 > -1 {
		if index2 > -1 {
			return index1 >= index2
		} else {
			return false
		}
	}
	return index2 > -1
}
```

adapters/eplanning/eplanning.go
```diff
@@ -61,16 +65,62 @@ type hbResponseAd struct {
 	Height       uint64 `json:"h,omitempty"`
 }
 
+func isPriority(index1 int, index2 int) bool {
+	if index1 > -1 {
+		if index2 > -1 {
+			if index1 < index2 {
```
SyntaxNode, line 71: This is essentially the return for `less`. Why is it returning false when index1 is less than index2? Could you please explain the algorithm and/or perhaps add comments?

adapters/eplanning/eplanning.go
```diff
@@ -61,16 +65,62 @@ type hbResponseAd struct {
...
+		if index2 > -1 {
+			return true
+		} else {
+			return false
+		}
+	}
+}
+
+func lessByPriority(w1, h1, w2, h2 uint64, priorityOrderForSizesAsc []string) bool {
+	size1 := fmt.Sprintf("%dx%d", w1, h1)
```
SyntaxNode, line 89: I'm a bit concerned there are 2 `Sprintf` calls occurring for each compare. Might not be a big deal if there aren't many formats though. How many formats are common on requests to your adapter?

chino117:
https://github.com/prebid/prebid-server/pull/1648#discussion_r560801779 :
The code can be simplified, it's true, I did it like this for a more descriptive logic, I change it if you want to as you suggest.

https://github.com/prebid/prebid-server/pull/1648#discussion_r560802446 :
The priorityOrderForMobileSizesAsc and priorityOrderForDesktopSizesAsc arrays are sorted in ascending order of priority, so the last item is the highest priority.
If the size is in the array, the indexOf function gives the index of its position within the array, if it is not defined, it returns -1.
What you want to achieve with the ordering is that the one with the highest priority is in the first position of the sizesSortedByPriority array.
So:
-if both sizes are defined, we take the one with the highest index as the smallest
-in case that one of the two sizes is not defined, the one that is defined is taken as smaller
-in case none is defined, the order is maintained

https://github.com/prebid/prebid-server/pull/1648#discussion_r560803368 :
As for the formats, they are usually 3 to 4

adapters/eplanning/eplanning.go
```diff
@@ -61,16 +65,62 @@ type hbResponseAd struct {
...
+	return isPriority(index1, index2)
+}
+
+type byPriorityMobile []openrtb.Format
+
+func (a byPriorityMobile) Len() int      { return len(a) }
+func (a byPriorityMobile) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a byPriorityMobile) Less(i, j int) bool {
+	return lessByPriority(a[i].W, a[i].H, a[j].W, a[j].H, priorityOrderForMobileSizesAsc)
+}
```
guscarreon, line 103: I just want to double-check on the sorting logic. The first element in `priorityOrderForMobileSizesAsc` has the highest priority and should be placed last in the sorted list, right? Just for the sake of understanding the changes better I wrote the test cases shown below. Is the behavior they unveil correct?  Is this the intended logic? (I find test case #4 particularly interesting)
```
// Some test I wrote just to help me understand the sorting logic better
func TestLessByPriorityMobile(t *testing.T) {
	testCases := []struct {
		desc          string
		inFormatList  []openrtb.Format
		expectedOrder []openrtb.Format
	}{
		//priorityOrderForMobileSizesAsc = []string{"1x1", "300x50", "320x50", "300x250"}
		{
			desc: "1) No WxH is listed in priorityOrderForMobileSizesAsc list doesn't change",
			inFormatList: []openrtb.Format{
				{W: 640, H: 480},
				{W: 320, H: 240},
			},
			expectedOrder: []openrtb.Format{
				{W: 640, H: 480},
				{W: 320, H: 240},
			},
		},
		{
			desc: "2) One entry listed in priorityOrderForMobileSizesAsc, that entry will be swapped to the last place",
			inFormatList: []openrtb.Format{
				{W: 1, H: 1},
				{W: 1080, H: 780},
				{W: 640, H: 480},
			},
			expectedOrder: []openrtb.Format{
				{W: 1080, H: 780},
				{W: 640, H: 480},
				{W: 1, H: 1},
			},
		},
		{
			desc: "3) Two entries found in priorityOrderForMobileSizesAsc and will be swapped to appear in order",
			inFormatList: []openrtb.Format{
				{W: 300, H: 250},
				{W: 1, H: 1},
			},
			expectedOrder: []openrtb.Format{
				{W: 300, H: 250},
				{W: 1, H: 1},
			},
		},
		{
			desc: "4) List exactly like priorityOrderForMobileSizesAsc. Entries will be reversed.",
			inFormatList: []openrtb.Format{
				{W: 1, H: 1},
				{W: 300, H: 50},
				{W: 320, H: 50},
				{W: 300, H: 250},
			},
			expectedOrder: []openrtb.Format{
				{W: 300, H: 250},
				{W: 320, H: 50},
				{W: 300, H: 50},
				{W: 1, H: 1},
			},
		},
	}
	for i, test := range testCases {

		actualOrder := test.inFormatList
		sort.Sort(byPriorityMobile(actualOrder))

		assert.Equal(t, test.expectedOrder, actualOrder, "Test %d not ok", i+1)
	}
}
adapters/eplanning/eplanning_test.go
```

The main reason behind wanting to double-check this is because in line 311 shown below it seems to me that the very first non-zero element (that can be the one with lowest priority) would be selected:
```
310         for _, format := range sizesSortedByPriority {
311             if format.W != 0 && format.H != 0 {
312                 return format.W, format.H
313             }
314         }
315     }
316
317     return 0, 0
318 }
adapters/eplanning/eplanning.go
```

adapters/eplanning/eplanning.go
```diff
@@ -232,13 +286,28 @@ func verifyImp(imp *openrtb.Imp) (*openrtb_ext.ExtImpEPlanning, error) {
...
-		for _, format := range imp.Banner.Format {
+		sizesSortedByPriority := imp.Banner.Format
+		if isMobile {
+			sort.Sort(byPriorityMobile(sizesSortedByPriority))
+		} else {
+			sort.Sort(byPriorityDesktop(sizesSortedByPriority))
+		}
+		for _, format := range sizesSortedByPriority {
 			if format.W != 0 && format.H != 0 {
 				return format.W, format.H
```
guscarreon, line 312: If at the end, only one format is going to be included in the `ImpExt`, do we need to sort the entire list? Assuming the format list comes with N entries, sorting comes at a cost of O(NlogN). But, if we just compare the format list and select a match from the `priorityOrderFor<DEVICE>Sizes` list, we could, at the cost of a little memory that Prebid Server could spare, do this in O(N+K) time instead. This seems like a good optimization opportunity given that the sorting algorithm you are currently using swaps to the end only the `priorityOrder` matches (see test case number two of the `TestLessByPriorityMobile(t *testing.T)` I wrote).

Maybe we could do something similar to:

```
298   func getSizeFromImp(imp *openrtb.Imp, isMobile bool) (uint64, uint64) {
299       if imp.Banner.W != nil && imp.Banner.H != nil {
300           return *imp.Banner.W, *imp.Banner.H
301       }
302  
303       if imp.Banner.Format != nil {
304 -         sizesSortedByPriority := imp.Banner.Format
305 -         if isMobile {
306 -             sort.Sort(byPriorityMobile(sizesSortedByPriority))
307 -         } else {
308 -             sort.Sort(byPriorityDesktop(sizesSortedByPriority))
309 -         }
310 -         for _, format := range sizesSortedByPriority {
311 -             if format.W != 0 && format.H != 0 {
312 -                 return format.W, format.H
313 -             }
314 -         }
    +         // Space O(n)
    +         hashedFormats := make(map[string]int,0)
    +
    +         // Traverse format array in O(n) time
    +         for i, format := range imp.Banner.Format {
    +             if format.W != 0 && format.H != 0 {
    +                 hashedFormats[fmt.Sprintf("%dx%d", format.W, format.H)] = i
    +             }
    +         }
    +         // Traverse prioritized formats in O(k) time
    +         for i:=len(priorityOrderFor__MobileOrDesktop__SizesAsc)-1; i>=0; i-- {
    +             //   ^--- Start with the highest priority in priorityOrderFor__MobileOrDesktop__SizesAsc
    +
    +             if formatIndex, wasFound := hashedFormats[priorityOrderFor__MobileOrDesktop__SizesAsc]; wasFound {
    +                 return imp.Banner.Format[formatIndex].W, imp.Banner.Format[formatIndex].H
    +             }
    +         }
315       }
316  
317       return 0, 0
318   }
adapters/eplanning/eplanning.go
```

guscarreon (changes requested):
Overall looks pretty good. Left a couple of comments

chino117:
https://github.com/prebid/prebid-server/pull/1648#discussion_r562218502
No, the logic is the last element of priorityOrderForMobileSizesAsc has the highest priority and must be placed first in the ordered list.

https://github.com/prebid/prebid-server/pull/1648#discussion_r562222704
I liked your suggestion and I have taken your code and made some modifications to it so that it complies with the logic that my implementation had.

guscarreon (approved):
LGTM. Thank you for addressing our comments.

adapters/eplanning/eplanning.go
```diff
@@ -232,17 +240,34 @@ func verifyImp(imp *openrtb.Imp) (*openrtb_ext.ExtImpEPlanning, error) {
...
+}
+
+func getSizeFromImp(imp *openrtb.Imp, isMobile bool) (uint64, uint64) {
 	if imp.Banner.W != nil && imp.Banner.H != nil {
 		return *imp.Banner.W, *imp.Banner.H
 	}
 
 	if imp.Banner.Format != nil {
-		for _, format := range imp.Banner.Format {
+		hashedFormats := make(map[string]int, 0)
```
SyntaxNode, line 258: Small Memory Optimization: Can you please provide a capacity to the map? This will eliminate the need for extra allocations in the case of longer format lists.
```
hashedFormats := make(map[string]int, len(imp.Banner.Format))
```

SyntaxNode (commented):
Looks great. Thank you for working with us on the feedback. Just one small optimization comment.

chino117:
Yes, no problem

SyntaxNode:
> Yes, no problem

Fantastic! Thank you.

SyntaxNode (approved):

guscarreon (approved):
LGTM

PR #1651, opened 2021-01-13
Bubble up GDPR signal/consent errors while applying privacy policies

bsardo:


hhhjort (approved):
LGTM

exchange/utils.go
```diff
@@ -75,14 +75,19 @@ func cleanOpenRTBRequests(ctx context.Context,
...
+	consent, err := extractConsent(req.BidRequest)
+	if err != nil {
+		errs = append(errs, err)
+	}
 	gdprEnforced := gdprSignal == gdpr.SignalYes || (gdprSignal == gdpr.SignalAmbiguous && !usersyncIfAmbiguous)
 
 	ccpaEnforcer, err := extractCCPA(req.BidRequest, privacyConfig, &req.Account, aliases, integrationTypeMap[req.LegacyLabels.RType])
 	if err != nil {
 		errs = append(errs, err)
-		return
```
SyntaxNode, line 85: Why does an early return no longer make sense?

bsardo: I removed this so we apply all other applicable privacy policies before returning since we don't fail fast (the cleaned openrtb requests are still processed downstream despite the error with requests being made to the bidders, etc).

SyntaxNode: Ah. Makes sense.

That's because of `bidderRequests` getting set a few lines before. I found the named return types confusing in this context. Something we can fix in the future.

SyntaxNode (approved):

PR #1656, opened 2021-01-14
Always sync when GDPR globally enabled and allow host cookie sync …

bsardo:
…when host vendor ID is 0

### DIFF
--- a/endpoints/auction_test.go
+++ b/endpoints/auction_test.go
@@ -349,7 +349,7 @@ func TestCacheVideoOnly(t *testing.T) {
 		t.Fatal(err.Error())
 	}
 	syncers := usersyncers.NewSyncerMap(cfg)
-	gdprPerms := gdpr.NewPermissions(nil, config.GDPR{
+	gdprPerms := gdpr.NewPermissions(context.TODO(), config.GDPR{
 		HostVendorID: 0,
 	}, nil, nil)
 	prebid_cache_client.InitPrebidCache(server.URL)
--- a/endpoints/auction_test.go
+++ b/endpoints/auction_test.go
@@ -349,7 +349,7 @@ func TestCacheVideoOnly(t *testing.T) {
 		t.Fatal(err.Error())
 	}
 	syncers := usersyncers.NewSyncerMap(cfg)
-	gdprPerms := gdpr.NewPermissions(context.TODO(), config.GDPR{
+	gdprPerms := gdpr.NewPermissions(context.Background(), config.GDPR{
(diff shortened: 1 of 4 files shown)


endpoints/auction_test.go
```diff
@@ -349,7 +349,7 @@ func TestCacheVideoOnly(t *testing.T) {
 		t.Fatal(err.Error())
 	}
 	syncers := usersyncers.NewSyncerMap(cfg)
-	gdprPerms := gdpr.NewPermissions(nil, config.GDPR{
+	gdprPerms := gdpr.NewPermissions(context.TODO(), config.GDPR{
```
hhhjort, line 352: Perhaps add a comment to say that `context.TODO()` is being used to generate a dummy context rather than as a placeholder for an improved permissions generator or an improvement of the testing code.

hhhjort: Actually should just use `context.Background()`, unless there is some subtle detail that makes background different from TODO, and not work here.

hhhjort (commented):
Looks good aside from the use of the TODO context.

gdpr/impl.go
```diff
@@ -223,6 +223,26 @@ func (p *permissionsImpl) parseVendor(ctx context.Context, vendorID uint16, cons
 	return
 }
 
+// DisallowHostCookies represents a GDPR permissions policy with host cookie syncing disallowed
+type DisallowHostCookies struct {
+	permissionsImpl Permissions
```
SyntaxNode, line 228: Can be simplified using wrapping.

hhhjort (approved):
LGTM

gdpr/impl.go
```diff
@@ -223,6 +223,16 @@ func (p *permissionsImpl) parseVendor(ctx context.Context, vendorID uint16, cons
...
 }
 
+// DisallowHostCookies represents a GDPR permissions policy with host cookie syncing disallowed
+type DisallowHostCookies struct {
+	*permissionsImpl
+}
+
+// HostCookiesAllowed always returns false
+func (p *DisallowHostCookies) HostCookiesAllowed(ctx context.Context, consent string) (bool, error) {
+	return false, nil
```
SyntaxNode, line 233: Re-reading the spec, I think this should return true. We want to disable the host cookie check if the host gvl id is not configured. Whereas before, the GVL code would check for consent given to host 0 and always return false.

hhhjort (approved):

SyntaxNode (approved):

PR #1665, opened 2021-01-21
Create usersync.go

prebidtappx:
We accept now cookie sync.
Some pull request today.
This is the first.


SyntaxNode:
Part of https://github.com/prebid/prebid-server/pull/1674.

PR #1671, opened 2021-01-21
Handle empty consent string during cookie sync and setuid

bsardo:
The main purpose of this commit is to properly handle an empty consent during a cookie sync. The logic is similar to changes recently made to the openrtb endpoints.
- If the GDPR signal is determined to be 0 indicating GDPR does not apply (either because it was explicitly set to 0 in the request or was ambiguous and normalized to 0), cookie sync is allowed.
- Else If the GDPR signal is determined to be 1 indicating GDPR applies (either because it was explicitly set to 1 in the request or was ambiguous and normalized to 1), and the consent string is empty, cookie sync is not allowed.
- Else the GDPR signal was determined to be 1 with a non-empty consent string so process the consent string to determine whether cookie sync is allowed

Other things to note:
- You’ll notice that the three methods that implement the Permissions interface all generally behave in the same way now for clarity. They attempt to normalize the GDPR signal, early return if the signal is No, early return if the consent string is empty and then process the consent string.
- the GDPR processing switch statements in `endpoints/auction.go` and `endpoints/setuid.go` were refactored to flatten the logic with early returns.
- I didn’t bother to tabularize the new tests I added in `impl_test.go`. I’ll handle that in another PR dedicated to that test file soon.


endpoints/setuid.go
```diff
@@ -143,32 +143,40 @@ func checkChromeBrowserVersion(ua string, index int, chromeStrLength int) bool {
...
-		}
-	default:
+func preventSyncsGDPR(gdprEnabled string, gdprConsent string, perms gdpr.Permissions) (shouldReturn bool, status int, body string) {
+
+	if gdprEnabled != "" && gdprEnabled != "0" && gdprEnabled != "1" {
 		return true, http.StatusBadRequest, "the gdpr query param must be either 0 or 1. You gave " + gdprEnabled
 	}
+
+	// TODO: keep or delete this logic? We could delete since this check is performed in HostCookiesAllowed, which is
+	// called below, but we won't get an error message back to include in the response like we see here.
```
bsardo, line 153: Thoughts on this?

SyntaxNode: I vote to keep it. Changing external behavior can have side effects. If no one's asking, let it be. IMHO.

endpoints/auction.go
```diff
@@ -189,21 +189,17 @@ func (a *auction) recoverSafely(inner func(*pbs.PBSBidder, metrics.AdapterLabels
...
-		fallthrough
-	default:
-		if canSync, err := a.gdprPerms.HostCookiesAllowed(ctx, gdprPrivacyPolicy.Consent); !canSync || err != nil {
-			return false
-		}
-		canSync, err := a.gdprPerms.BidderSyncAllowed(ctx, bidder, gdprPrivacyPolicy.Consent)
-		return canSync && err == nil
+	gdprSignal := gdpr.SignalAmbiguous
+	// invalid data is treated as an ambiguous signal
+	if i, err := strconv.Atoi(gdprPrivacyPolicy.Signal); err == nil && (i == 0 || i == 1) {
```
SyntaxNode, line 194: Do we need the `(i == 0 || i == 1)` check or can `gdpr.Signal(i)` take care of that logic?

endpoints/auction_test.go
```diff
@@ -397,11 +397,22 @@ func TestShouldUsersync(t *testing.T) {
 			t.Errorf("Expected syncs: %t, allowed syncs: %t", expectAllow, allowSyncs)
 		}
 	}
-	doTest("0", "", false, false, true)
-	doTest("1", "", true, true, false)
-	doTest("1", "a", true, false, false)
-	doTest("1", "a", false, true, false)
-	doTest("1", "a", true, true, true)
+
+	doTest("0", "", false, false, false)
```
SyntaxNode, line 401: If you're willing to convert this to a table driven test, I find this:
```
 {signal: "0", consent: "", allowBidderSync: true, allowHostCookies: false, expectAllow: true 
```
much easier to read.

endpoints/setuid.go
```diff
@@ -143,32 +143,38 @@ func checkChromeBrowserVersion(ua string, index int, chromeStrLength int) bool {
...
+		// Since we try to fetch new versions as requests come in for them, PBS *should* self-correct
+		// rather quickly, meaning that most of these will be malformed strings.
+		return true, http.StatusBadRequest, "No global vendor list was available to interpret this consent string. If this is a new, valid version, it should become available soon."
+	}
+
+	if allowed {
+		return false, 0, ""
+	}
+
+	return true, http.StatusOK, "The gdpr_consent string prevents cookies from being saved"
```
SyntaxNode, line 179: FYI: We'll revisit in this in the future. We decided on a 451 error code for these situations.

gdpr/impl.go
```diff
@@ -82,10 +84,22 @@ func (p *permissionsImpl) defaultVendorPermissions() (allowPI bool, allowGeo boo
 	return false, false, false, nil
 }
 
+func (p *permissionsImpl) normalizeGDPR(gdprSignal *Signal) {
```
SyntaxNode, line 87: Why is signal a pointer? Would you consider:
```
gdprSignal = p.normalizeGDPR(gdprSignal)
```
to avoid the heap allocation?

endpoints/auction.go
bsardo: I think we need that logic. I originally intended to use `gdpr.Signal(i)` to handle other integer values as you're suggesting but it results in `gdprSignal` being set to that integer instead of one of the three signal types.

SyntaxNode: Gotcha.

hhhjort (dismissed):
Looks good to me.

endpoints/auction.go
SyntaxNode: Looking at this again, I dislike that 0 and 1 are hardcoded here far away from the context of 0 and 1 being known signals in the gdpr package. How about something like:
```
gdprSignal := gdpr.SignalAmbiguous
if i, err := strconv.Atoi(gdprPrivacyPolicy.Signal); err == nil && gdpr.IsKnownSignal(i) {
  gdprSignal = gdpr.Signal(i)
}
```
or
```
gdprSignal := gdpr.SignalAmbiguous
if signal, err := gdpr.SignalParse(gdprPrivacyPolicy.Signal); err == nil {
  gdprSignal = signal
}
```

endpoints/auction.go
```diff
@@ -189,21 +189,17 @@ func (a *auction) recoverSafely(inner func(*pbs.PBSBidder, metrics.AdapterLabels
...
+		gdprSignal = gdpr.Signal(i)
+	}
+
+	if canSync, err := a.gdprPerms.HostCookiesAllowed(ctx, gdprSignal, gdprPrivacyPolicy.Consent); !canSync || err != nil {
```
SyntaxNode, line 198: How about we check the error first? `err != nil || !canSync`.

endpoints/auction_test.go
```diff
@@ -376,32 +378,134 @@ func TestCacheVideoOnly(t *testing.T) {
...
+		allowBidderSync  bool
+		wantAllow        bool
+	}{
+		{
+			description:      "Don't sync - GDPR off, host cookies allows and bidder sync disallows",
+			signal:           "0",
+			allowHostCookies: true,
+			allowBidderSync:  false,
+			wantAllow:        false,
+		},
```
SyntaxNode, line 394: If the gdpr signal is no (don't enforce) and the host allows cookies, shouldn't we allow syncing? The bidder sync wouldn't apply since gdpr is not enforced?

This might be a test setup error since we're using s mock. The real `BidderSyncAllowed` would return true if the signal is no.

endpoints/auction_test.go
```diff
@@ -376,32 +378,134 @@ func TestCacheVideoOnly(t *testing.T) {
...
+		{
+			description:      "Don't sync - GDPR ambiguous, host cookies allows and bidder sync disallows",
+			signal:           "",
+			allowHostCookies: true,
+			allowBidderSync:  false,
+			wantAllow:        false,
+		},
```
SyntaxNode, line 408: Same comment as earlier since ambiguous == "no" (well, depending on the `UsersyncIfAmbiguous` setting, but that's not fully in play here since it's a mock.

endpoints/auction_test.go
```diff
@@ -376,32 +378,134 @@ func TestCacheVideoOnly(t *testing.T) {
...
+		{
+			description:      "Don't sync - GDPR off, host cookies disallows and bidder sync allows",
+			signal:           "0",
+			allowHostCookies: false,
+			allowBidderSync:  true,
+			wantAllow:        false,
+		},
```
SyntaxNode, line 415: Should we always allow if signal is no / don't enforce?

SyntaxNode (commented):
I have some nitpick comments and test comments. Because we're mocking out parts of the gdpr checks, some of the test cases don't make sense to me.

endpoints/auction_test.go
bsardo: Do you think I should just remove the cases that are unrealistic in this test given the internals of `HostCookiesAllowed` and `BidderSyncAllowed`? My thought here was to exercise the conditionals in the `shouldUserSync` method ignoring the real implementation of those methods by using the mocks that were already in place.

bsardo: @SyntaxNode mentioned offline that he thinks we should remove the unrealistic cases.

hhhjort: We shouldn't test conditions that only the mock would pass (test relies on the mock behavior being different from the real behavior). In these cases we should check if the real behavior is already being tested for in the test for the object being mocked. If so we should get rid of the test here. If not, we should consider either adding a test on the original object or improving the mock here so that the code paths can be properly tested.

bsardo: That makes sense. I'm remembering now that this is testing the legacy auction endpoint and this will all be deleted soon so I don't think it's worth our time to improve the mock, plus the original object has pretty good coverage already with a story in my queue to further improve it. I removed a bunch of the unrealistic test cases and just focused on testing how `HostCookiesAllowed` and `BidderSyncAllowed` affect the return value.

gdpr/gdpr.go
```diff
@@ -65,3 +67,17 @@ type ErrorMalformedConsent struct {
...
+
+// SignalParse parses a raw signal and returns
+func SignalParse(rawSignal string) (Signal, error) {
+	i, err := strconv.Atoi(rawSignal)
+
+	if err != nil {
+		return SignalAmbiguous, err
+	}
+	if i != 0 && i != 1 {
+		return SignalAmbiguous, &errortypes.BadInput{"GDPR signal should be integer 0 or 1"}
```
SyntaxNode, line 79: Linters will warn you against using positional/unkeyed values. Please add the `Message:` key in the initializer for the message content.

gdpr/gdpr_test.go
```diff
@@ -48,3 +48,61 @@ func TestNewPermissions(t *testing.T) {
...
+			rawSignal:   "1",
+			wantSignal:  SignalYes,
+			wantError:   false,
+		},
+		{
+			description: "Invalid signal - raw signal is empty",
+			rawSignal:   "",
+			wantSignal:  SignalAmbiguous,
+			wantError:   true,
+		},
```
SyntaxNode, line 76: Should an empty signal really emit an error? I would think it would return SignalAmbiguous without an error since in other parts of code we treat an empty signal as SignalAmbiguous.

gdpr/gdpr_test.go
```diff
@@ -48,3 +48,61 @@ func TestNewPermissions(t *testing.T) {
...
+			rawSignal:   "-1",
+			wantSignal:  SignalAmbiguous,
+			wantError:   true,
+		},
+		{
+			description: "Invalid signal - raw signal is 2",
+			rawSignal:   "-1",
+			wantSignal:  SignalAmbiguous,
+			wantError:   true,
+		},
```
SyntaxNode, line 88: Copy/paste error. This is the same as the previous test with `"-1"`. However, I don't think it's necessary as -1 is already testing a valid int outside of the acceptable range.

gdpr/gdpr.go
bsardo: Good catch. Not sure how I missed that one.

gdpr/gdpr_test.go
bsardo: Oops. Refactored.

SyntaxNode (approved):

hhhjort (approved):

PR #1675, opened 2021-01-22
AMX Bid Adapter: Loop Variable Bug

nickjacob:
This is a bugfix for an issue affecting the AMX Bid adapter. When the AMX RTB server responds with multiple (N) bids, the adapter will parse this as the first bid, repeated N times.

This is due to the fact that we store a pointer to the `bid` variable in the `TypedBid` which is changed with each iteration of the loop.

- added test `display-multiple.json` to reproduce
- refactored to add temporary loop variable
- bumped amx version string

### DIFF
--- a/adapters/amx/amx.go
+++ b/adapters/amx/amx.go
@@ -147,10 +147,9 @@ func (adapter *AMXAdapter) MakeBids(request *openrtb.BidRequest, externalRequest
 
 	bidResponse := adapters.NewBidderResponseWithBidsCapacity(5)
 
-	for _, seatbid := range bidResp.SeatBid {
-		sb := seatbid
-		for _, bidExternal := range sb.Bid {
-			bid := bidExternal
+	for _, sb := range bidResp.SeatBid {
+		for _, bid := range sb.Bid {
+			bid := bid
 			bidExt, bidExtErr := getBidExt(bid.Ext)
 			if bidExtErr != nil {
 				errs = append(errs, bidExtErr)
--- a/adapters/amx/amx.go
+++ b/adapters/amx/amx.go
@@ -17,7 +17,7 @@ import (
 const vastImpressionFormat = "<Impression><![CDATA[%s]]></Impression>"
 const vastSearchPoint = "</Impression>"
 const nbrHeaderName = "x-nbr"
-const adapterVersion = "pbs1.0"
+const adapterVersion = "pbs1.1"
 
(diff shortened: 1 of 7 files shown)


adapters/amx/amx.go
```diff
@@ -147,8 +147,10 @@ func (adapter *AMXAdapter) MakeBids(request *openrtb.BidRequest, externalRequest
 
 	bidResponse := adapters.NewBidderResponseWithBidsCapacity(5)
 
-	for _, sb := range bidResp.SeatBid {
-		for _, bid := range sb.Bid {
+	for _, seatbid := range bidResp.SeatBid {
+		sb := seatbid
+		for _, bidExternal := range sb.Bid {
+			bid := bidExternal
 			bidExt, bidExtErr := getBidExt(bid.Ext)
```
SyntaxNode, line 154: Nice catch. We'll need to keep a better eye out for these issues in the future.

The `sb := seatbid` local assignment isn't necessary since you're not taking the address of it.

A common pattern for the local assignment is to use the same variable name and rely on shadowing. You may leave the names as you wish, but just wanted to throw it out there:

```
for _, bid := range sb.Bid {
  bid := bid
```

SyntaxNode (commented):
Nice test coverage. One small nitpick comment.

adapters/amx/amx.go
nickjacob: Thanks -- that makes sense. Just pushed a commit removing that sb := assignment & changing to bid := bid

adapters/amx/amxtest/exemplary/display-multiple.json
```diff
@@ -0,0 +1,307 @@
...
+                "adm": "<i></i><script>_c=[];</script><script src=\"https://assets.a-mo.net/js/c.js\"></script>",
+                "adomain": [
+                  "dell.com.au"
+                ],
+                "cid": "668",
+                "crid": "253510977",
+                "ext": {
+                  "dp": 0.475762,
+                  "lid": 142443,
+                  "sbid": "668"
```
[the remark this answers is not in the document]
nickjacob: Thanks for pointing that out @guscarreon  -- we're removing all of those ext properties from the server response (they're not used client-side). I updated the test to take those out / remove some of the other unnecessary data

guscarreon (approved):
LGTM

SyntaxNode (approved):

PR #1677, opened 2021-01-22
Debug disable feature implementation:

VeronikaSolovei9:
- Added Bidder level debug variable
- Added Account level debug variable

### DIFF
--- a/adapters/info.go
+++ b/adapters/info.go
@@ -206,7 +206,7 @@ type BidderInfo struct {
 	Capabilities            *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
 	AliasOf                 string            `json:"aliasOf,omitempty"`
 	ModifyingVastXmlAllowed bool              `yaml:"modifyingVastXmlAllowed" json:"-" xml:"-"`
-	DebugInfo               *DebugInfo        `yaml:"debug,omitempty" json:"debug,omitempty"`
+	Debug                   *DebugInfo        `yaml:"debug,omitempty" json:"-" xml:"-"`
 }
 
 type DebugInfo struct {
--- a/adapters/info.go
+++ b/adapters/info.go
@@ -206,7 +206,11 @@ type BidderInfo struct {
 	Capabilities            *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
 	AliasOf                 string            `json:"aliasOf,omitempty"`
 	ModifyingVastXmlAllowed bool              `yaml:"modifyingVastXmlAllowed" json:"-" xml:"-"`
-	DebugInfo               *config.DebugInfo `yaml:"debug,omitempty" json:"debug,omitempty"`
+	DebugInfo               *DebugInfo        `yaml:"debug,omitempty" json:"debug,omitempty"`
+}
+
+type DebugInfo struct {
+	Allow bool `yaml:"allow" json:"allow"`
 }
 
 type MaintainerInfo struct {
--- a/adapters/info.go
+++ b/adapters/info.go
@@ -206,6 +206,7 @@ type BidderInfo struct {
 	Capabilities            *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
 	AliasOf                 string            `json:"aliasOf,omitempty"`
 	ModifyingVastXmlAllowed bool              `yaml:"modifyingVastXmlAllowed" json:"-" xml:"-"`
+	DebugInfo               *config.DebugInfo `yaml:"debug,omitempty" json:"debug,omitempty"`
 }
 
 type MaintainerInfo struct {
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -19,7 +19,7 @@ type Account struct {
 	EventsEnabled bool        `mapstructure:"events_enabled" json:"events_enabled"`
 	CCPA          AccountCCPA `mapstructure:"ccpa" json:"ccpa"`
 	GDPR          AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
-	DebugAllowed  bool        `mapstructure:"debug_allowed" json:"debug_allowed"`
+	DebugAllow    bool        `mapstructure:"debug_allow" json:"debug_allow"`
 }
 
 // AccountCCPA represents account-specific CCPA configuration
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -19,7 +19,7 @@ type Account struct {
 	EventsEnabled bool        `mapstructure:"events_enabled" json:"events_enabled"`
 	CCPA          AccountCCPA `mapstructure:"ccpa" json:"ccpa"`
 	GDPR          AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
-	DebugAllowed  bool        `mapstructure:"debugallowed" json:"debugallowed"`
+	DebugAllowed  bool        `mapstructure:"debug_allowed" json:"debug_allowed"`
 }
 
 // AccountCCPA represents account-specific CCPA configuration
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -19,6 +19,7 @@ type Account struct {
 	EventsEnabled bool        `mapstructure:"events_enabled" json:"events_enabled"`
 	CCPA          AccountCCPA `mapstructure:"ccpa" json:"ccpa"`
 	GDPR          AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
+	DebugAllowed  bool        `mapstructure:"debugallowed" json:"debugallowed"`
 }
 
 // AccountCCPA represents account-specific CCPA configuration
--- a/config/config.go
+++ b/config/config.go
@@ -446,10 +446,6 @@ type Debug struct {
 	TimeoutNotification TimeoutNotification `mapstructure:"timeout_notification"`
 }
 
-type DebugInfo struct {
-	Allow bool `yaml:"allow" json:"allow"`
-}
-
 func (cfg *Debug) validate(errs []error) []error {
 	return cfg.TimeoutNotification.validate(errs)
 }
--- a/config/config.go
+++ b/config/config.go
@@ -937,7 +937,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("blacklisted_accts", []string{""})
 	v.SetDefault("account_required", false)
 	v.SetDefault("account_defaults.disabled", false)
-	v.SetDefault("account_defaults.debugallowed", true)
+	v.SetDefault("account_defaults.debug_allowed", true)
 	v.SetDefault("certificates_file", "")
 	v.SetDefault("auto_gen_source_tid", true)
 
--- a/config/config.go
+++ b/config/config.go
@@ -937,7 +937,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("blacklisted_accts", []string{""})
 	v.SetDefault("account_required", false)
 	v.SetDefault("account_defaults.disabled", false)
-	v.SetDefault("account_defaults.debug_allowed", true)
+	v.SetDefault("account_defaults.debug_allow", true)
 	v.SetDefault("certificates_file", "")
 	v.SetDefault("auto_gen_source_tid", true)
 
--- a/config/config.go
+++ b/config/config.go
@@ -446,6 +446,10 @@ type Debug struct {
 	TimeoutNotification TimeoutNotification `mapstructure:"timeout_notification"`
 }
 
+type DebugInfo struct {
+	Allow bool `yaml:"allow" json:"allow"`
+}
+
 func (cfg *Debug) validate(errs []error) []error {
 	return cfg.TimeoutNotification.validate(errs)
 }
@@ -937,6 +941,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("blacklisted_accts", []string{""})
 	v.SetDefault("account_required", false)
 	v.SetDefault("account_defaults.disabled", false)
+	v.SetDefault("account_defaults.debugallowed", true)
 	v.SetDefault("certificates_file", "")
 	v.SetDefault("auto_gen_source_tid", true)
 
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -274,17 +274,6 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	ctx = context.WithValue(ctx, exchange.DebugAllowedContextKey, account.DebugAllowed)
-
-	/*if debugLog.Enabled && account.DebugAllowed {
-		debugLog.Data.Request = string(requestJson)
-		if headerBytes, err := json.Marshal(r.Header); err == nil {
-			debugLog.Data.Headers = string(headerBytes)
-		} else {
-			debugLog.Data.Headers = fmt.Sprintf("Unable to marshal headers data: %s", err.Error())
-		}
-	}*/
-
 	auctionRequest := exchange.AuctionRequest{
 		BidRequest:   bidReq,
 		Account:      *account,
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -274,6 +274,17 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	ctx = context.WithValue(ctx, exchange.DebugAllowedContextKey, account.DebugAllowed)
+
+	/*if debugLog.Enabled && account.DebugAllowed {
+		debugLog.Data.Request = string(requestJson)
+		if headerBytes, err := json.Marshal(r.Header); err == nil {
+			debugLog.Data.Headers = string(headerBytes)
+		} else {
+			debugLog.Data.Headers = fmt.Sprintf("Unable to marshal headers data: %s", err.Error())
+		}
+	}*/
+
 	auctionRequest := exchange.AuctionRequest{
 		BidRequest:   bidReq,
 		Account:      *account,
--- a/exchange/adapter_util.go
+++ b/exchange/adapter_util.go
@@ -39,7 +39,7 @@ func buildExchangeBidders(cfg *config.Configuration, infos adapters.BidderInfos,
 
 	exchangeBidders := make(map[openrtb_ext.BidderName]adaptedBidder, len(bidders))
 	for bidderName, bidder := range bidders {
-		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName, infos[bidderName.String()].DebugInfo)
+		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName, infos[bidderName.String()].Debug)
 	}
 
 	return exchangeBidders, nil
--- a/exchange/adapter_util.go
+++ b/exchange/adapter_util.go
@@ -39,7 +39,12 @@ func buildExchangeBidders(cfg *config.Configuration, infos adapters.BidderInfos,
 
 	exchangeBidders := make(map[openrtb_ext.BidderName]adaptedBidder, len(bidders))
 	for bidderName, bidder := range bidders {
-		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName, infos[bidderName.String()].Debug)
+		info, infoFound := infos[string(bidderName)]
+		if !infoFound {
+			errs = append(errs, fmt.Errorf("%v: bidder info not found", bidder))
+			continue
+		}
+		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName, info.Debug)
 	}
 
 	return exchangeBidders, nil
--- a/exchange/adapter_util.go
+++ b/exchange/adapter_util.go
@@ -39,7 +39,7 @@ func buildExchangeBidders(cfg *config.Configuration, infos adapters.BidderInfos,
 
 	exchangeBidders := make(map[openrtb_ext.BidderName]adaptedBidder, len(bidders))
 	for bidderName, bidder := range bidders {
-		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName)
+		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName, infos[bidderName.String()].DebugInfo)
 	}
 
 	return exchangeBidders, nil
--- a/exchange/bidder.go
+++ b/exchange/bidder.go
@@ -93,7 +93,9 @@ type pbsOrtbSeatBid struct {
 //
 // The name refers to the "Adapter" architecture pattern, and should not be confused with a Prebid "Adapter"
 // (which is being phased out and replaced by Bidder for OpenRTB auctions)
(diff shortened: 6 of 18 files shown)


VeronikaSolovei9:
@jmaynardxandr @camrice Guys please review debug disable feature PR.

@camrice Cam, can you please pay attention to this commit: https://github.com/prebid/prebid-server/pull/1677/commits/4c7d53afcef10cc07239cf3ed01c663560cd66f3
This is possible cache fix, mostly about adding response to logs in case of bids returned and not.

endpoints/openrtb2/video_auction.go
```diff
@@ -274,6 +274,17 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
+	ctx = context.WithValue(ctx, exchange.DebugAllowedContextKey, account.DebugAllowed)
+
+	/*if debugLog.Enabled && account.DebugAllowed {
+		debugLog.Data.Request = string(requestJson)
+		if headerBytes, err := json.Marshal(r.Header); err == nil {
+			debugLog.Data.Headers = string(headerBytes)
+		} else {
+			debugLog.Data.Headers = fmt.Sprintf("Unable to marshal headers data: %s", err.Error())
+		}
+	}*/
```
camrice, line 286: This block can be removed

exchange/bidder_test.go
```diff
@@ -1401,6 +1422,23 @@ func TestTimeoutNotificationOn(t *testing.T) {
 	assert.EqualValues(t, logExpected, logActual)
 }
 
+func TestParseDebugInfoTrue(t *testing.T) {
+	debugInfo := &config.DebugInfo{Allow: true}
+	resDebugInfo := parseDebugInfo(debugInfo)
+	assert.True(t, resDebugInfo.Allow, "Debug Allow value should be false")
```
camrice, line 1428: The failure message here should mention true instead of false

exchange/bidder_test.go
```diff
@@ -1401,6 +1422,23 @@ func TestTimeoutNotificationOn(t *testing.T) {
...
+
+func TestParseDebugInfoFalse(t *testing.T) {
+	debugInfo := &config.DebugInfo{Allow: false}
+	resDebugInfo := parseDebugInfo(debugInfo)
+	assert.False(t, resDebugInfo.Allow, "Debug Allow value should be false")
+}
+
+func TestParseDebugInfoIsNil(t *testing.T) {
+	resDebugInfo := parseDebugInfo(nil)
+	assert.True(t, resDebugInfo.Allow, "Debug Allow value should be false")
```
camrice, line 1439: Same as above, just need to change failure message to true

exchange/exchange.go
```diff
@@ -194,36 +203,57 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+
+			cacheErrs = auc.doCache(ctx, e.cache, targData, evTracking, r.BidRequest, 60, &r.Account.CacheTTL, bidCategory, debugLog)
 			if len(cacheErrs) > 0 {
 				errs = append(errs, cacheErrs...)
 			}
+			/*// Ensure caching errors are added in case auc.doCache was called and errors were returned
+			if len(cacheErrs) > 0 {
+				bidderCacheErrs := errsToBidderErrors(cacheErrs)
+				bidResponseExt.Errors[openrtb_ext.PrebidExtKey] = append(bidResponseExt.Errors[openrtb_ext.PrebidExtKey], bidderCacheErrs...)
+			}*/
```
camrice, line 225: This block can be removed

endpoints/openrtb2/video_auction.go
camrice, line 286: Should this be removed? Are the request and header fields added elsewhere?

exchange/exchange.go
```diff
@@ -194,36 +203,57 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+
+			if jsonBytes, err := json.Marshal(debugLog.CacheString); err == nil {
+				toCache = append(toCache, prebid_cache_client.Cacheable{
+					Type:       debugLog.CacheType,
+					Data:       jsonBytes,
+					TTLSeconds: debugLog.TTL,
+					Key:        "log_" + debugLog.CacheKey,
+				})
+			}
+			_, err := e.cache.PutJson(ctx, toCache)
```
camrice, line 253: I think this would cause two calls to the cache in a no-bid case. In video_auction.go, there is a check around line 306 for if the adpod length is 0, and then it calls the PutDebugLogError() function. This also adds the cache key to the response. I think this PutJson call is not needed.

adapters/info.go
```diff
@@ -206,6 +206,7 @@ type BidderInfo struct {
 	Capabilities            *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
 	AliasOf                 string            `json:"aliasOf,omitempty"`
 	ModifyingVastXmlAllowed bool              `yaml:"modifyingVastXmlAllowed" json:"-" xml:"-"`
+	DebugInfo               *config.DebugInfo `yaml:"debug,omitempty" json:"debug,omitempty"`
```
SyntaxNode, line 209: The json must be `json:"-" xml:"-"` to play nicely with the /info/bidders endpoint.

SyntaxNode, line 209: Would calling it `Debug` instead of `DebugInfo` cause a naming conflict?

config/accounts.go
```diff
@@ -19,6 +19,7 @@ type Account struct {
 	EventsEnabled bool        `mapstructure:"events_enabled" json:"events_enabled"`
 	CCPA          AccountCCPA `mapstructure:"ccpa" json:"ccpa"`
 	GDPR          AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
+	DebugAllowed  bool        `mapstructure:"debugallowed" json:"debugallowed"`
```
SyntaxNode, line 22: Let's call the JSON `debug_allowed` to match the other naming conventions.

endpoints/openrtb2/video_auction.go
```diff
@@ -274,6 +274,17 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	ctx = context.WithValue(ctx, exchange.DebugAllowedContextKey, account.DebugAllowed)
```
SyntaxNode, line 277: I really don't like the context. Is there a way we can do without this?

config/config.go
```diff
@@ -446,6 +446,10 @@ type Debug struct {
 	TimeoutNotification TimeoutNotification `mapstructure:"timeout_notification"`
 }
 
+type DebugInfo struct {
+	Allow bool `yaml:"allow" json:"allow"`
```
hhhjort, line 450: The default in Go for a boolean is false, so that will result in every adapter being disallowed unless all the BidderInfo yamls are updated to allow this.

exchange/bidder.go
```diff
@@ -102,10 +102,22 @@ func adaptBidder(bidder adapters.Bidder, client *http.Client, cfg *config.Config
 		config: bidderAdapterConfig{
 			Debug:              cfg.Debug,
 			DisableConnMetrics: cfg.Metrics.Disabled.AdapterConnectionMetrics,
+			DebugInfo:          parseDebugInfo(debugInfo),
 		},
 	}
 }
 
+func parseDebugInfo(info *config.DebugInfo) config.DebugInfo {
```
hhhjort, line 110: Minor point, but perhaps we just want to return the final boolean value here, rather than a full `config.DebugInfo`. Is there any benefit in keeping it packed inside of the the struct?

exchange/exchangetest/append-bidder-names.json
```diff
@@ -166,9 +166,6 @@
     },
     "ext": {
       "debug": {
-        "httpcalls": {
```
hhhjort, line 169: Do we still have test coverage somewhere that `httpcalls` are included if debug is allowed?

endpoints/openrtb2/video_auction.go
VeronikaSolovei9: Yes, this commented block of code is moved to line 160. It's not removed.

endpoints/openrtb2/video_auction.go
VeronikaSolovei9: Refactored. We have this variable in account. To have this in `bidder -> requestBid` I passed it through function param from `HoldAuction`

exchange/exchangetest/append-bidder-names.json
VeronikaSolovei9: Yes, exchange_test.go, `TestDebugBehaviour` covers all the cases

exchange/bidder.go
VeronikaSolovei9: Yes, fixed

config/config.go
VeronikaSolovei9: @SyntaxNode please advice if we want to have` DebugRestricted` instead of `DebugInfo.Allow`

adapters/info.go
VeronikaSolovei9: Right, didn't know about this, fixed

VeronikaSolovei9: Renamed, no conflicts

config/accounts.go
VeronikaSolovei9: Modified!

exchange/bidder_test.go
VeronikaSolovei9: Fixed, thank you!

exchange/bidder_test.go
VeronikaSolovei9: Fixed

exchange/exchange.go
VeronikaSolovei9: Yes, removed

exchange/exchange.go
VeronikaSolovei9: You are right! We only need to build response and put it to debugLog. Fixed. Great great great catch!

exchange/exchange.go
```diff
@@ -194,34 +198,39 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+				bidderCacheErrs := errsToBidderErrors(cacheErrs)
+				bidResponseExt.Errors[openrtb_ext.PrebidExtKey] = append(bidResponseExt.Errors[openrtb_ext.PrebidExtKey], bidderCacheErrs...)
+			}*/
 			targData.setTargeting(auc, r.BidRequest.App != nil, bidCategory)
 
 		}
-	}
-
-	bidResponseExt := e.makeExtBidResponse(adapterBids, adapterExtra, r, debugInfo, errs)
+		bidResponseExt = e.makeExtBidResponse(adapterBids, adapterExtra, r, debugInfo, errs)
```
bsardo, line 224: Why aren't we marshaling the ext and setting `debugLog.Data.Response` here if the `debugLog` is not nil and enabled?

config/config.go
hhhjort: I think you missed my point. I was thinking of moving this flag out of the `BidderInfo` struct, and thus out of the bidder/info endpoint, and into the actual server configuration in the `Adapter` struct, where we actually do sever level configuration of the adapter. I am not aware of anything in the /static/bidder-info/ directory that a PBS host should touch, all configuration should be driven by the config file and/or environment variables. Spreading configuration out across multiple files will just lead to confusion.

VeronikaSolovei9: I'm confused. 
We want to have bidder level debug flag and store it in static/bidder-info/{bidder}.yaml files. The only place where we load it is in adapters/info.go -> func ParseBidderInfos. We Unmarshal file data to BidderInfo struct. 
Unless I miss something... 
Do you mean I need to  move it to  config/adapter.go -> Adapter struct after loading? 
What is "the actual server configuration in the Adapter struct"?

adapters/info.go
```diff
@@ -206,6 +206,11 @@ type BidderInfo struct {
 	Capabilities            *CapabilitiesInfo `yaml:"capabilities" json:"capabilities"`
 	AliasOf                 string            `json:"aliasOf,omitempty"`
 	ModifyingVastXmlAllowed bool              `yaml:"modifyingVastXmlAllowed" json:"-" xml:"-"`
+	Debug                   *DebugInfo        `yaml:"debug,omitempty" json:"-" xml:"-"`
+}
+
+type DebugInfo struct {
+	Allow bool `yaml:"allow" json:"allow"`
```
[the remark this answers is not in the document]
SyntaxNode: > Should debug be allowed by default?

Yes. The entire `DebugInfo` structure is a pointer currently. I think that's fine.

As you said, another option is to make the `Debug` field in `BidderInfo` not a pointer and add a pointer here. That would make it more future proof, but the current approach solves the problem at hand. tomato/tomato IMHO.

config/config.go
SyntaxNode: > @SyntaxNode please advice if we want to have DebugRestricted instead of DebugInfo.Allow

I don't have a strong opinion. Bret introduced this config as debug allow and that makes sense to me, but yes - it does play against the boolean zero value. We'd need to ask in the GitHub issues to swap it to something like restricted.

> I was thinking of moving this flag out of the BidderInfo struct, and thus out of the bidder/info endpoint, and into the actual server configuration in the Adapter struct,

I think both have valid arguments. I had spec'd it for the bidder yaml since I didn't expect a host to override this value. I thought if the bidder knows its url will include sensitive information, then it could specify such here for all hosts to follow.

We've also had discussions about in the future using bidder yaml as the base config for each adapter as is done in PBS-Java and then using the config system to override those values if needed. This would work well for reducing the code length of config.go, making it easier to port adapters to PBS-Java, and would provide a nice way to override yaml information for aliases.

exchange/bidder.go
```diff
@@ -155,8 +166,16 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
 		httpInfo := <-responseChannel
 		// If this is a test bid, capture debugging info from the requests.
+		// Write debug data to ext in case if:
+		// - debugContextKey (url param) in true
+		// - account debug is allowed
+		// - bidder debug is allowed
 		if debugInfo := ctx.Value(DebugContextKey); debugInfo != nil && debugInfo.(bool) {
-			seatBid.httpCalls = append(seatBid.httpCalls, makeExt(httpInfo))
+			if accountDebugAllowed {
+				if bidder.config.DebugInfo.Allow {
```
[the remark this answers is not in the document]
VeronikaSolovei9: In case we bid more than one bidder in one request, some of those bidders may have debug.allowed=true and others may have debug.allowed=false. For this case we need to make 2 checks and write logs for those bidders where debug.allowed=true. Hope this make sense.

exchange/exchange.go
VeronikaSolovei9: We do this earlier in code before `doCache` function, line 204.
If no bids returned we marshal it in `else`, line 230.

exchange/bidder.go
```diff
@@ -93,7 +93,9 @@ type pbsOrtbSeatBid struct {
 //
 // The name refers to the "Adapter" architecture pattern, and should not be confused with a Prebid "Adapter"
 // (which is being phased out and replaced by Bidder for OpenRTB auctions)
-func adaptBidder(bidder adapters.Bidder, client *http.Client, cfg *config.Configuration, me metrics.MetricsEngine, name openrtb_ext.BidderName) adaptedBidder {
+func adaptBidder(bidder adapters.Bidder, client *http.Client, cfg *config.Configuration, me metrics.MetricsEngine, name openrtb_ext.BidderName, debugInfo *adapters.DebugInfo) adaptedBidder {
+	bidderDebugInfo := adapters.DebugInfo{parseDebugInfo(debugInfo)}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Done!

exchange/bidder.go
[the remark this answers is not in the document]
VeronikaSolovei9: Wait... did my comment disappear? 
Correct. If account level debug disabled, we don't need to check bidder level config. We just skip it. 
But if account debug is allowed, we need to check every bidder in request. 

We may bid more than one bidder in one request and those bidders may have different debug options. To handle this case we need to check each of them and log data only for those bidders where debug.allow = true. 

Hope this makes sense and answers your question.

config/config.go
hhhjort: Oh, if PBS-Java is already using it for config, then it is not such a big change to for us to do the same. I have no opinion on if this struct should be in config/config.go or adapters/info.go

hhhjort (dismissed):
Looks good to me as long as everyone else is happy.

exchange/adapter_util.go
```diff
@@ -39,7 +39,7 @@ func buildExchangeBidders(cfg *config.Configuration, infos adapters.BidderInfos,
 
 	exchangeBidders := make(map[openrtb_ext.BidderName]adaptedBidder, len(bidders))
 	for bidderName, bidder := range bidders {
-		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName)
+		exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName, infos[bidderName.String()].Debug)
```
SyntaxNode, line 42: There is a guarantee that `infos[bidderName.String()]` will be found, but only because there's a check in `buildBidders`. I don't think that relationship is obvious enough. I recommend to add protections for the lookup here as well. Something like:

```
for bidderName, bidder := range bidders {
  info, infoFound := infos[string(bidderName)]
  if !infoFound {
    errs = append(errs, fmt.Errorf("%v: bidder info not found", bidder))
    continue
  }

  exchangeBidders[bidderName] = adaptBidder(bidder, client, cfg, me, bidderName, info.Debug)
}

return exchangeBidders, errs
```

exchange/bidder.go
```diff
@@ -155,8 +164,16 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+		// - debugContextKey (url param) in true
+		// - account debug is allowed
+		// - bidder debug is allowed
 		if debugInfo := ctx.Value(DebugContextKey); debugInfo != nil && debugInfo.(bool) {
-			seatBid.httpCalls = append(seatBid.httpCalls, makeExt(httpInfo))
+			if accountDebugAllowed {
+				if bidder.config.DebugInfo.Allow {
+					seatBid.httpCalls = append(seatBid.httpCalls, makeExt(httpInfo))
+				}
+			}
```
SyntaxNode, line 176: Could you combine the `if` clauses to improve readability?
```
if accountDebugAllowed && bidder.config.DebugInfo.Allow {
```

exchange/exchange.go
```diff
@@ -136,6 +135,10 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 	}
 
 	debugInfo := getDebugInfo(r.BidRequest, requestExt)
+
+	debugInfo = debugInfo && r.Account.DebugAllowed
+	debugLog.Enabled = debugLog.Enabled && r.Account.DebugAllowed
```
SyntaxNode, line 140: `debugLog` can be nil. This needs a nil check, or perhaps incorporate the `r.Account.DebugAllowed` condition in line 203 and 228. Could you please add a unit test to cover when `debugLog` is nil, because that doesn't seem to have coverage.

exchange/exchange.go
SyntaxNode: I think the comment block is till here.

config/accounts.go
```diff
@@ -19,6 +19,7 @@ type Account struct {
 	EventsEnabled bool        `mapstructure:"events_enabled" json:"events_enabled"`
 	CCPA          AccountCCPA `mapstructure:"ccpa" json:"ccpa"`
 	GDPR          AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
+	DebugAllowed  bool        `mapstructure:"debug_allowed" json:"debug_allowed"`
```
SyntaxNode, line 22: Should be DebugAllow / debug_allow per the spec and to coordinate with debug.allow in the bidder info.

exchange/exchange_test.go
```diff
@@ -292,10 +326,99 @@ func TestDebugBehaviour(t *testing.T) {
...
+
+	// Set up test
+	noBidServer := func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(204)
+	}
+	server := httptest.NewServer(http.HandlerFunc(noBidServer))
+	defer server.Close()
+
+	categoriesFetcher, error := newCategoryFetcher("./test/category-mapping")
+	if error != nil {
```
guscarreon, line 348: Can we rename `error` to `err` instead? Just so it doesn't get confused with golang's `error` interface.

guscarreon (commented):
A couple of comments

exchange/adapter_util.go
VeronikaSolovei9: Done!

exchange/bidder.go
VeronikaSolovei9: Yes, was going to to this too, done! :)

exchange/exchange.go
VeronikaSolovei9: Good catch! Also modified existing test to make sure `debugLog` is properly handled in Auction function.

config/accounts.go
VeronikaSolovei9: Done

exchange/exchange_test.go

exchange/exchange.go
```diff
@@ -194,34 +198,39 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 				dealErrs := applyDealSupport(r.BidRequest, auc, bidCategory)
 				errs = append(errs, dealErrs...)
 			}
-			cacheErrs := auc.doCache(ctx, e.cache, targData, evTracking, r.BidRequest, 60, &r.Account.CacheTTL, bidCategory, debugLog)
+
+			bidResponseExt = e.makeExtBidResponse(adapterBids, adapterExtra, r, debugInfo, errs)
```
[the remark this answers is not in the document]
VeronikaSolovei9: I need to test it more before pushing. Working on it...

exchange/exchange.go
```diff
@@ -163,10 +166,11 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 	// Get currency rates conversions for the auction
 	conversions := e.currencyConverter.Rates()
 
-	adapterBids, adapterExtra, anyBidsReturned := e.getAllBids(auctionCtx, bidderRequests, bidAdjustmentFactors, conversions)
+	adapterBids, adapterExtra, anyBidsReturned := e.getAllBids(auctionCtx, bidderRequests, bidAdjustmentFactors, conversions, r.Account.DebugAllowed)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Makes sense. Will push it shortly.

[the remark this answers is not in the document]
SyntaxNode: I think this is veering out of scope of this work. I don't feel this needs to be included in this PR. If you already did it, great. Otherwise I think this comment can be worked on separately.

hhhjort (approved):

exchange/exchange.go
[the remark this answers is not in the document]
VeronikaSolovei9: This is fix for problem we had before - when we put data in cache in case everything is ok - there is no response data in debug logs. We don't have explicit test for this, we only check we put "something" in cache. 

I understand your concern, but cannot come up with better solution.

guscarreon (approved):
LGTM

SyntaxNode (approved):

bsardo (approved):

PR #1680, opened 2021-01-26
New Adapter: Epom

epomrnd:


adapters/epom/epom.go
```diff
@@ -0,0 +1,172 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type EpomAdapter struct {
```
SyntaxNode, line 15: Please use the name `adapter`. There is no need to repeat the adapter name from the package and there is no need to export the type (it should start with a lower case letter to be package private).

adapters/epom/epom.go
```diff
@@ -0,0 +1,172 @@
...
+		endpoint: config.Endpoint,
+	}
+	return bidder, nil
+}
+
+func (a *EpomAdapter) MakeRequests(request *openrtb.BidRequest, reqInfo *adapters.ExtraRequestInfo) (requests []*adapters.RequestData, errors []error) {
+	rq, errs := a.makeRequest(request)
+
+	if len(errs) > 0 {
+		errors = append(errors, errs...)
```
SyntaxNode, line 31: You can return the `errs` directly, as its not possible for `errors` to have anything else in it:
```
if len(errs) > 0 {
  return nil, errors
}
```

adapters/epom/epom.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, errors
+	}
+
+	if rq != nil {
+		requests = append(requests, rq)
+	}
+
+	return requests, errors
```
SyntaxNode, line 39: Errors will always be empty here. Consider returning `nil` instead.
```
return requests, nil
```

adapters/epom/epom.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (a *EpomAdapter) makeRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
+	var errs []error
+	var validImps []openrtb.Imp
+
+	if len(request.Imp) == 0 {
+		return nil, []error{&errortypes.BadInput{
+			Message: "No impressions in request",
+		}}
+	}
```
SyntaxNode, line 50: The imp length check is performed by PBS-Core. It does not need to be repeated here.

adapters/epom/epom.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, []error{&errortypes.BadInput{
+			Message: "Site or app field shouldn't be empty",
+		}}
+	}
+
+	if request.Site != nil && request.App != nil {
+		return nil, []error{&errortypes.BadInput{
+			Message: "Site and app can't be filled simultaneously",
+		}}
+	}
```
SyntaxNode, line 68: The site / app check is performed by PBS-Core. Both `request.Site == nil && request.App == nil` and `request.Site != nil && request.App != nil` do not need to be repeated here.

adapters/epom/epom.go
```diff
@@ -0,0 +1,172 @@
...
+			if imp.Banner != nil {
+				return openrtb_ext.BidTypeBanner
+			} else if imp.Video != nil {
+				return openrtb_ext.BidTypeVideo
+			} else if imp.Native != nil {
+				return openrtb_ext.BidTypeNative
+			}
+		}
+	}
+	return openrtb_ext.BidTypeBanner
```
SyntaxNode, line 171: Does your bidding server include the bid format in the response?

SyntaxNode (commented):
Could you please also a documentation PR and link to it here. See: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#user-documentation

adapters/epom/epom.go
epomrnd: Renamed

adapters/epom/epom.go
epomrnd: You are right. Fixed!

adapters/epom/epom.go
epomrnd: Done

adapters/epom/epom.go
epomrnd: Ok. Removed!

adapters/epom/epom.go
epomrnd: Removed

adapters/epom/epom.go
epomrnd: Epom bidding server responds according to OpenRtb 2.5 spec. So, bid doesn't have any bid format info.

epomrnd:
PR into prebid documentation project is here:
https://github.com/prebid/prebid.github.io/pull/2667

adapters/epom/epom.go
```diff
@@ -0,0 +1,159 @@
...
+	var errs []error
+	var validImps []openrtb.Imp
+
+	if request.Device == nil || request.Device.IP == "" {
+		return nil, []error{&errortypes.BadInput{
+			Message: "No ip address in request",
+		}}
+	}
+
+	if request.Site == nil && request.App == nil {
```
SyntaxNode, line 51: This check is performed by PBS-Core. You can remove it here. You can assume you'll always have either a Site or App defined here, not both and not neither.

adapters/epom/epom.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	return requests, nil
+}
+
+func (a *adapter) makeRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
+	var errs []error
+	var validImps []openrtb.Imp
+
+	if request.Device == nil || request.Device.IP == "" {
```
SyntaxNode, line 45: Do you specifically need an IPv4 address? There's also `Device.IPv6`. PBS-Core will determine the type of ip address and set it in the appropriate field.

adapters/epom/epom.go
```diff
@@ -0,0 +1,159 @@
...
+	headers.Add("Accept", "application/json")
+	return &adapters.RequestData{
+		Method:  "POST",
+		Uri:     a.endpoint,
+		Body:    reqJSON,
+		Headers: headers,
+	}, errs
+}
+
+func preprocess(imp *openrtb.Imp) error {
```
SyntaxNode, line 88: I actually don't think you need this check. PBS-Core has you covered here too. See: https://github.com/prebid/prebid-server/blob/ef06fac6494f7e91f9d1d90ab07f02f8499b5efc/endpoints/openrtb2/auction.go#L472

adapters/epom/epom.go
SyntaxNode: It's preferable to include metadata in your bidding server response to be sure PBS has the right creative type to ensure proper rendering. If we can't receive a hint from your server, please ensure the order of preference matches your server. For example, banner is always chosen over video.

SyntaxNode (commented):
It looks like we can simply the adapter even further my removing a few more checks already enforced by PBS-Core.

adapters/epom/epom.go
epomrnd: Ok. Removed

adapters/epom/epom.go
epomrnd: Yes, Epom Bidder always choose media type these way. Banner type is the first priority, then - video and then - native.

adapters/epom/epom.go
epomrnd: Currently we support only IPv4 address

adapters/epom/epom.go
epomrnd: Ok. Here I removed id check and added check if impression media type is not audio (we don't support audio)

epomrnd:
Hello @SyntaxNode  @bsardo 
It seems like I have fixed and answered all your pr reviews. 
Feel free to let me know if I need to check/fix something else. 
Thank you!

adapters/epom/epom.go
```diff
@@ -0,0 +1,153 @@
...
+	if imp.Audio != nil {
+		return &errortypes.BadInput{
+			Message: "Audio is not supported",
+		}
+	}
```
SyntaxNode, line 87: This check is also covered by PBS-Core, since audio excluded as an acceptable type in the yaml file. You can remove the entire `preprocess` method, it's tests, and the `for _, imp := range request.Imp {` loop in ` MakeRequests`. The only check you need to keep is for the ipv4.

In the end, I think this would be sufficient:
```
func (a *adapter) makeRequest(request *openrtb.BidRequest) (*adapters.RequestData, []error) {
	if request.Device == nil || request.Device.IP == "" {
			Message: "ipv4 address is required field",
		}}
	}

	reqJSON, err := json.Marshal(request)
	if err != nil {
		return nil, []error{err}
	}

	headers := http.Header{}
	return &adapters.RequestData{
		Method:  "POST",
		Uri:     a.endpoint,
		Body:    reqJSON,
		Headers: headers,
	}, nil
}
```

adapters/epom/epom.go
```diff
@@ -0,0 +1,153 @@
...
+		}}
+	}
+
+	var bidResp openrtb.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+
+	//additional no content check
+	if len(bidResp.SeatBid) == 0 || len(bidResp.SeatBid[0].Bid) == 0 {
```
SyntaxNode, line 121: Could you please add a test for this condition.

SyntaxNode (commented):
Thank you for the update. I have two more comments. If you reply today or early tomorrow, there's a very good chance you'll make it in this week's release.

bsardo (commented):
LGTM other than @SyntaxNode's latest comments.

adapters/epom/epom.go
epomrnd: Tests with empty seatbid and bid objects added

adapters/epom/epom.go
epomrnd: Thank you! Used your code. Check please.

SyntaxNode (approved):

bsardo (approved):
LGTM

agilfix:
Hi epom team, I was just curious if properties is meant to be empty or if there are any optional/required parameters that epom needs to get inventory information to bid on (https://github.com/prebid/prebid-server/blob/master/static/bidder-params/epom.json)?

PR #1681, opened 2021-01-26
Marsmedia - add bid floor price

vladi-mmg:
Add bid floor price


adapters/marsmedia/marsmedia.go
```diff
@@ -67,6 +67,10 @@ func (a *MarsmediaAdapter) MakeRequests(requestIn *openrtb.BidRequest, reqInfo *
 		} else if request.Imp[i].Video != nil {
 			validImpExists = true
 		}
+
+		if marsmediaExt.BidFloor > 0 {
+			request.Imp[i].BidFloor = marsmediaExt.BidFloor
```
SyntaxNode, line 72: We prefer adapters to use data from the standard OpenRTB request. Please see our docs: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#bidder-parameters

Why do are you adding a bidder parameter for a bid floor and setting it in the impression, rather than using what's specified in the impression?

vladi-mmg: HI @SyntaxNode 
Thanks for your feedback.
I get it now, and I will use the data from the request.
So, in this case, the version update is not necessary and can be canceled.

PR #1685, opened 2021-02-01
Pubmatic: Trimming publisher ID before passing

sachin-pubmatic:


adapters/pubmatic/pubmatic_test.go
```diff
@@ -728,3 +730,36 @@ func TestGetBidTypeForUnsupportedCode(t *testing.T) {
 		t.Errorf("Expected Bid Type value was: %v, actual value is: %v", openrtb_ext.BidTypeBanner, actualBidTypeValue)
 	}
 }
+
+func TestMakeRequestsTrimsPubID(t *testing.T) {
```
guscarreon, line 734: We've been trying to make the adapter tests to be JSON only. We don't need to convert all the tests found in `pubmatic_test.go` to JSON tests but I believe we could make the newly added `TestMakeRequestsTrimsPubID(t *testing.T)` into a JSON test by simply adding spaces to a working test like:
```
diff --git a/adapters/pubmatic/pubmatictest/exemplary/simple-banner.json b/adapters/pubmatic/pubmatictest/exemplary/simple-banner.json
index 1eb8a212..2efab8fc 100644
--- a/adapters/pubmatic/pubmatictest/exemplary/simple-banner.json
+++ b/adapters/pubmatic/pubmatictest/exemplary/simple-banner.json
@@ -12,7 +12,7 @@
             "ext": {
                 "bidder": {
                     "adSlot": "AdTag_Div1@300x250",
-                    "publisherId": "999",
+                    "publisherId": "   999    ",
                     "keywords": [{
                         "key": "pmZoneID",
                         "value": ["Zone1", "Zone2"]
```
If not, maybe we can make a `adapters/pubmatic/pubmatictest/supplemental/` test too. Thoughts?

guscarreon (commented):
Looks good but I was wondering whether we could add a JSON test instead.

SyntaxNode:
@sachin-pubmatic This looks good to us if you could just please express the test using the json framework, which includes more robust assertions.

sachin-pubmatic:
Thanks @guscarreon & @SyntaxNode for catching this. I have updated the unit test to use JSON framework.

mansinahar (approved):

guscarreon (approved):
LGTM

PR #1695, opened 2021-02-04
New Adapter: Onetag

onetag-dev:


adapters/onetag/onetag.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/macros"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type onetagAdapter struct {
```
SyntaxNode, line 17: There's no reason to repeat the package name. Please prefer the simpler name `adapter`.

adapters/onetag/onetag.go
```diff
@@ -0,0 +1,169 @@
...
+	bidder := &onetagAdapter{
+		endpointTemplate: *template,
+	}
+
+	return bidder, nil
+}
+
+func (a *onetagAdapter) MakeRequests(request *openrtb.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+
+	if len(request.Imp) <= 0 {
```
SyntaxNode, line 36: This check is already performed by the core framework. It's not necessary to duplicate it here.

adapters/onetag/onetag.go
```diff
@@ -0,0 +1,169 @@
...
+			if impression.Banner != nil {
+				return openrtb_ext.BidTypeBanner, nil
+			}
+			if impression.Video != nil {
+				return openrtb_ext.BidTypeVideo, nil
+			}
+			if impression.Native != nil {
+				return openrtb_ext.BidTypeNative, nil
+			}
+		}
```
SyntaxNode, line 163: We consider this an anti-pattern. Does your bidding server indicate the type of the bid in its response?

openrtb_ext/imp_onetag.go
```diff
@@ -0,0 +1,10 @@
+package openrtb_ext
+
+import (
+	"encoding/json"
+)
+
+type ExtImpOnetag struct {
+	PubId string          `json:"pubId"`
```
SyntaxNode, line 8: OpenRTB already includes a publisher id field. Why do you need to use a separate one?

SyntaxNode (commented):
Thank you for following the guidance of the new bidder doc. 👍

openrtb_ext/imp_onetag.go
onetag-dev: This is intended as an Account ID, which uniquely identifies a Publisher on our system and it has a different meaning from the request.{site,app}.publisher.id

adapters/onetag/onetag.go
onetag-dev: I renamed it to adapter!

adapters/onetag/onetag.go
onetag-dev: Fixed.

adapters/onetag/onetag.go
onetag-dev: Onetag server doesn't include the bid format because it responds according to OpenRtb 2.5 spec.

SyntaxNode:
Doc PR: https://github.com/prebid/prebid.github.io/pull/2685

adapters/onetag/onetag_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	assert.Error(t, buildErr)
+}
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderOneTag, config.Adapter{
+		Endpoint: "https://example.com/prebid-server/{{.PublisherID}}"})
+
+	if buildErr != nil {
+		t.Fatalf("Builder returned unexpected error %v", buildErr)
```
mansinahar, line 24: NitPick: Consider using `assert.NoError` here

adapters/onetag/onetagtest/exemplary/simple-native.json
```diff
@@ -0,0 +1,120 @@
...
+                  "seat": "onetag"
+                }
+              ],
+              "cur": "USD"
+            }
+          }
+        }
+      ],
+      "expectedBidResponses": [
+        {
```
mansinahar, line 102: Please consider adding the expected currency as well for the bid response like you have in the previous test

adapters/onetag/params_test.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}`,
+	`{"pubId": "386276e072"}`,
+}
+
+var invalidParams = []string{
+	`{"ext": {
+		"key1": "value1",
+		"key2": "value2"
+	}`,
```
mansinahar, line 50: Please consider adding another the following cases here as well:
1. A case with empty params i.e. with neither `ext` or `pubId` defined
2. A case with `pubId` defined to be an empty string
3. A case with `pubId` defined to be a number rather than a string

adapters/onetag/onetag.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	return "", &errortypes.BadServerResponse{
+		Message: fmt.Sprintf("The impression with ID %s is not present into the request", bid.ImpID),
```
mansinahar, line 160: Can you please add a JSON test to cover this error case as well?

mansinahar:
Looks pretty good in general @onetag-dev! Just left a few minor comments.

adapters/onetag/onetag.go
onetag-dev: I added this test case.

adapters/onetag/params_test.go
onetag-dev: I added this test cases.

adapters/onetag/onetagtest/exemplary/simple-native.json
onetag-dev: Fixed.

adapters/onetag/onetag_test.go
onetag-dev: Fixed.

mansinahar (approved):
Thanks for addressing the comments @onetag-dev! Looks Good 👍

adapters/onetag/onetagtest/exemplary/simple-native.json
```diff
@@ -0,0 +1,121 @@
...
+          "cur": "USD",
```
mansinahar, line 103: Oops looks like I missed this but this should really be `currency` and not `cur`. I know, it's confusing. We should really just be sticking to either one but anyway, can you please update this for now?

onetag-dev: Thank you! Fixed.

mansinahar (dismissed):
Re-approved!

adapters/onetag/onetag.go
```diff
@@ -0,0 +1,162 @@
...
+	return &onetagExt, nil
+}
+
+func (a *adapter) buildEndpointURL(pubID string) (string, error) {
+	endpointParams := macros.EndpointTemplateParams{PublisherID: pubID}
+	return macros.ResolveMacros(a.endpointTemplate, endpointParams)
+}
+
+func (a *adapter) MakeBids(request *openrtb.BidRequest, requestData *adapters.RequestData, responseData *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if responseData.StatusCode == http.StatusNoContent {
```
SyntaxNode, line 103: Could you please add supplemental json tests to cover these status code conditions (204, 400, not 200, 204, 400)?

adapters/onetag/onetag.go
```diff
@@ -0,0 +1,162 @@
...
+	for idx, imp := range request.Imp {
+		onetagExt, err := getImpressionExt(imp)
+		if err != nil {
+			return nil, []error{err}
+		}
+		if onetagExt.PubId != "" {
+			if pubID == "" {
+				pubID = onetagExt.PubId
+			}
+			if pubID != onetagExt.PubId {
```
SyntaxNode, line 45: Could this be an `else` condition? If the pubID is empty, then it can't also equal onetagExt.PubId.

adapters/onetag/onetagtest/supplemental/required-publisher-id.json
```diff
@@ -0,0 +1,66 @@
+{
```
SyntaxNode, line 1: Please consider adding a publisher id test where there's only one unique publisher id mixed with other imps which define an empty publisher id.

SyntaxNode (commented):
Looks great. Just a few more requests for test coverage then I think we're all good.

adapters/onetag/onetag.go
onetag-dev: I added a test to cover 204 and a test to cover not 200. I deleted the check for status code 400 because it is never returned by the server.

adapters/onetag/onetag.go
onetag-dev: Fixed.

adapters/onetag/onetagtest/supplemental/required-publisher-id.json
onetag-dev: I added this test.

mansinahar (approved):

SyntaxNode (approved):

PR #1702, opened 2021-02-16
Mobfox: Add rout to adexcange

mobfxoHB:
Allows ssp to connect without direct inventory through oRTB adexchange


adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -28,21 +39,40 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
 
+	requestURI := a.URI
 	reqCopy := *request
 	imp := request.Imp[0]
-	tagID, err = jsonparser.GetString(imp.Ext, "bidder", "TagID")
-	if err != nil {
-		errs = append(errs, err)
+	tagID, err_tag = jsonparser.GetString(imp.Ext, "bidder", "TagID")
+	key, err_key = jsonparser.GetString(imp.Ext, "bidder", "key")
+	if err_tag != nil && err_key != nil {
```
hhhjort, line 55: This will only trigger if both errors out. I think you want to use or (||) here.

SyntaxNode: If this wasn't cause by a test case, there may be a testing gap.

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -125,7 +155,5 @@ func getMediaTypeForImp(impID string, imps []openrtb.Imp) (openrtb_ext.BidType,
 	}
 
 	// This shouldnt happen. Lets handle it just incase by returning an error.
-	return "", &errortypes.BadServerResponse{
-		Message: fmt.Sprintf("Failed to find impression \"%s\" ", impID),
-	}
+	return "", nil
```
mansinahar, line 158: Curious, why remove this error message?

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -28,21 +39,40 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
 	reqCopy := *request
 	imp := request.Imp[0]
-	tagID, err = jsonparser.GetString(imp.Ext, "bidder", "TagID")
-	if err != nil {
-		errs = append(errs, err)
+	tagID, err_tag = jsonparser.GetString(imp.Ext, "bidder", "TagID")
+	key, err_key = jsonparser.GetString(imp.Ext, "bidder", "key")
+	if err_tag != nil && err_key != nil {
+		errs = append(errs, &errortypes.BadInput{
+			Message: fmt.Sprintf("Invalid or non existing key or tagId, atleast one should be present"),
```
mansinahar, line 57: From the code it seems like you expect either a valid key or tag ID and so you might want the error message to be:
```
Invalid or non existing key and tagId
```

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -28,21 +39,40 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
+	var method string
 	var adapterRequests []*adapters.RequestData
 
+	requestURI := a.URI
 	reqCopy := *request
 	imp := request.Imp[0]
-	tagID, err = jsonparser.GetString(imp.Ext, "bidder", "TagID")
-	if err != nil {
-		errs = append(errs, err)
+	tagID, err_tag = jsonparser.GetString(imp.Ext, "bidder", "TagID")
```
mansinahar, line 53: **Nitpick**: Please consider using the variable declare and define shorthand `:=` here rather than defining the variables explicitly above. Meaning, you can change these lines to be:
```
tagID, err_tag := jsonparser.GetString(imp.Ext, "bidder", "TagID")
key, err_key := jsonparser.GetString(imp.Ext, "bidder", "key")
```
And then remove the `err_tag`, `err_key` and `key` variable declarations from L43-45.

adapters/mobfoxpb/params_test.go
```diff
@@ -37,6 +37,7 @@ func TestInvalidParams(t *testing.T) {
 
 var validParams = []string{
 	`{"TagID": "6"}`,
+	`{"key": "1"}`,
 }
 
 var invalidParams = []string{
```
mansinahar, line 43: Please consider adding invalid param values for `key` param as well for the `invalidParams` test cases

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: No that's valid, we except etleast one, and fail only if there is now neither tag nor key

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: I can't find a way to test this keys, so looks like there is no way to really produce this error

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: Not really sure what do you mean, to remove this part from message `, atleast one should be present` ?

mansinahar: Nope! Sorry that wasn't super clear. I meant changing the `key or tagID` part to `key and tagID` because you will get to this error only if both `key` and `tagId` are invalid. Hope that makes sense

adapters/mobfoxpb/mobfoxpb.go
mansinahar: You can add a JSON test where the mock response from the bidder has a diff imp ID than what was in the request. That will trigger this case.  I'd suggest keeping that error message though so that you know if your users ever come across such an issue.

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: We really expect atleast one parameter, and fail only if both not present or invalid

adapters/mobfoxpb/mobfoxpb.go
mobfxoHB: Removed and renamed to follow single format

adapters/mobfoxpb/mobfoxpb.go
mansinahar: Yeah which is why that should be `Invalid or non existing key and tagID` because that'd indicate that both are either not present or invalid and therefore it's failing. `Invalid or non existing key or tagID` indicates that you need both those values to be present and valid.

adapters/mobfoxpb/mobfoxpb.go
```diff
@@ -142,6 +137,7 @@ func (a *adapter) MakeBids(internalRequest *openrtb.BidRequest, externalRequest
 func getMediaTypeForImp(impID string, imps []openrtb.Imp) (openrtb_ext.BidType, error) {
 	mediaType := openrtb_ext.BidTypeBanner
 	for _, imp := range imps {
+		fmt.Println(impID, imp.ID)
```
mansinahar, line 140: This looks like a leftover log line for local debugging. Can you please remove it?

hhhjort (approved):
Looking good.

mansinahar:
@mobfxoHB PR looks good. Thanks for addressing the feedback. Can you please also open a doc PR adding the new `key` param in this file: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/mobfoxpb.md

mobfxoHB:
@mansinahar https://github.com/prebid/prebid.github.io/pull/2724

Issue #1699, opened 2021-02-09
Accommodate Apple iOS LMT bug


bszekely1:
## Type of issue

Intent to implement

## Description

Apple introduced a bug in iOS 14.0 that persisted until 14.2 improperly setting LMT = 1 in all scenarios. The impact to this bug is an increase LMT=1 traffic for those iOS versions. This ticket is to resolve the bug, moving future LMT logic to PBS core. The Prebid SDK is not able to keep up to date with the changes fast enough, plus any devices with an older version Prebid SDK (prior to 1.8.1) and newer versions of iOS will exhibit problems.

Below is the current LMT behavior matrix with the associated Prebid SDK change to always pass LMT in version 1.20:

<img width="767" alt="Screen Shot 2021-02-09 at 1 35 25 PM" src="https://user-images.githubusercontent.com/49168870/107431741-e155e780-6af4-11eb-8f72-9caf08028bcd.png">

Below is the future behavior matrix of LMT as it pertains to Prebid SDK, assuming the change in PBS for LMT logic and Prebid SDK changing to always pass LMT in PBS:

<img width="777" alt="Screen Shot 2021-02-09 at 1 35 35 PM" src="https://user-images.githubusercontent.com/49168870/107431758-e7e45f00-6af4-11eb-8ec2-e8db2cab70f9.png">

## Goals

- Move LMT logic to PBS for iOS versions where Apple LMT behavior is in flux

## Proposed Design

- If PBS receives a request with the following attributes: app object exists, device.os = “iOS”(lower case value before matching), device.osv = “14.0" or “14.1” the following requirements apply:
    - If device.ifa = null or “” (empty string) or “00000000-0000-0000-0000-000000000000”, set device.lmt = 1
    - Else, set device.lmt=0, if not already set,  in the request to bid adapters
- Else if  PBS receives a request with the following attributes: app object exists, device.os = “iOS”, device.osv = “14.2” or greater the following requirements apply:
    - If device.ext.atts = 0 or 3, set device.lmt = 0, if not already set, in the request to bid adapters. ATTS 0 and 3 are "unspecified" and "authorized" respectively
    - Else if device.ext.atts = 1 or 2, set device.lmt = 1, if not already set, in the request to bid adapters. ATTS 1 or 2 are "restricted" and "denied" respectively
- Else, use device.lmt in the request

PBS should set any LMT changes **BEFORE** applying LMT anonymization.

### Prebid SDK Changes
[...]

PR #1718, opened 2021-02-19
Accommodate Apple iOS LMT bug

SyntaxNode:
Implements: https://github.com/prebid/prebid-server/issues/1699


endpoints/openrtb2/auction_test.go
```diff
@@ -2076,6 +2077,245 @@ func TestValidateBidders(t *testing.T) {
 	}
 }
 
+func TestWorkaroundIOS14PrivacyBug(t *testing.T) {
```
SyntaxNode, line 2080: There's a LOT of possible test case combinations. I likely already included too many, but I cut myself off at testing all initial values of the LMT flag. In most cases, it's set to `nil` at the start. There are a few test cases which ensure the LMT is overridden when set to a different non-nil value.

endpoints/openrtb2/auction_test.go
```diff
@@ -2076,6 +2077,245 @@ func TestValidateBidders(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range testCases {
+		workaroundIOS14PrivacyBug(test.givenRequest)
+		assert.Equal(t, test.expectedLMT, test.givenRequest.Device.Lmt, test.description)
+	}
+}
+
+func TestWorkaroundIOS14PrivacyBugEndToEnd(t *testing.T) {
```
SyntaxNode, line 2291: This is testing that the workaround logic is being hit during an end-to-end test.

I think the combinatorial details covered by the `TestWorkaroundIOS14PrivacyBug` are sufficient to prove the logic works and this test proves the logic is hit.

openrtb_ext/device.go
```diff
@@ -12,14 +14,58 @@ const PrebidExtKey = "prebid"
 
 // ExtDevice defines the contract for bidrequest.device.ext
 type ExtDevice struct {
+	// Attribute:
```
SyntaxNode, line 17: Trying out a new comment format, mimicking what exists in the OpenRTB library. Whatchya think?

util/iosutil/iosutil.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// EqualOrGreater returns true if iOS device version is equal or greater to the desired version, using semantic versioning.
+func (v IOSVersion) EqualOrGreater(major, minor int) bool {
+	if v.Major == major {
+		return v.Minor >= minor
+	}
+
+	return v.Major > major
+}
```
SyntaxNode, line 47: The Go standard library doesn't have simple semantic versioning helpers and popular third party libraries provide way more functionality than needed. I opted for a simple implementation instead.

endpoints/openrtb2/auction.go
```diff
@@ -264,6 +265,8 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb.
 		return
 	}
 
+	lmt.ModifyForIOS(req)
```
guscarreon, line 268: Nitpick: `lmt.ModifyForIOS(req)` comes after `processInterstitials(req)` in line 263. This `processInterstitials(req)` function unmarshalls the `Device.Ext` making the `jsonparser.GetInt(deviceExt, "atts")` call that `lmt.ModifyForIOS(req)` eventually makes inside `ParseDeviceExtATTS(deviceExt json.RawMessage)` to be unnecessary. Actually, that 261 to 268 neighborhood makes use of `Device` a lot 
```
260     // Populate any "missing" OpenRTB fields with info from other sources, (e.g. HTTP request headers).
261     deps.setFieldsImplicitly(httpRequest, req)  // <--- uses `Device`
262
263     if err := processInterstitials(req); err != nil {  // <--- unmarshalls `Device.Ext` and checks for `Device!=nil` again
264         errs = []error{err}
265         return
266     }
267
268     lmt.ModifyForIOS(req)  // <--- jsonparses `Device.Ext`
269
270     errL := deps.validateRequest(req)
endpoints/openrtb2/auction.go
```
Do you think it'd be a good idea to maybe consolidate inside `deps.setFieldsImplicitly(httpRequest, req)`?
```
260       // Populate any "missing" OpenRTB fields with info from other sources, (e.g. HTTP request headers).
    +     if err := deps.setFieldsImplicitly(httpRequest, req); err != nil {
261 -     deps.setFieldsImplicitly(httpRequest, req)
262 -  
263 -     if err := processInterstitials(req); err != nil {
264           errs = []error{err}
265           return
266       }
267 -  
268 -     lmt.ModifyForIOS(req)
269
270       errL := deps.validateRequest(req)
             .
             .
             .
1059 - func (deps *endpointDeps) setFieldsImplicitly(httpReq *http.Request, bidReq *openrtb.BidRequest) {
1059 + func (deps *endpointDeps) setFieldsImplicitly(httpReq *http.Request, bidReq *openrtb.BidRequest) error {
1060 -     sanitizeRequest(bidReq, deps.privateNetworkIPValidator) // <-- READS Device and checks Device != nil
1061 -
1062 -     setDeviceImplicitly(httpReq, bidReq, deps.privateNetworkIPValidator) // <-- if Device==nil, creates one bidReq.Device = &openrtb.Device{}
     +     if bidReq.Device != nil {
     +         &openrtb.Device{}
     +     }
     +     sanitizeRequest(bidReq, deps.privateNetworkIPValidator) // <-- MODIFY ACCORDINGLY
     +     setDeviceImplicitly(httpReq, bidReq, deps.privateNetworkIPValidator) // <-- MODIFY ACCORDINGLY
     +     
     +     //unmarshal device.Ext if any     
     +     if req.Device.Ext != nil {
     +         err := json.Unmarshal(req.Device.Ext, &devExt)
     +         if err == nil {
     +             // Do the iOS and interstitials thing
     +             if err := processInterstitials(req, devExt); err != nil { // <-- MODIFY TO TAKE UNMARSHALED Device.Ext
     +                 return err // I guess we'll have to return error from setFieldsImplicitly now
     +             }
     +             lmt.ModifyForIOS(req, devExt) // <-- MODIFY TO TAKE EITHER UNMARSHALED Device.Ext OR THE ATTS VALUE
     +         }              
     +     }
1063  
1064       // Per the OpenRTB spec: A bid request must not contain both a Site and an App object.
1065       if bidReq.App == nil {
1066           setSiteImplicitly(httpReq, bidReq)
1067       }
1068       setImpsImplicitly(httpReq, bidReq.Imp)
1069  
1070       setAuctionTypeImplicitly(bidReq)
     +     return nil
1071   }
endpoints/openrtb2/auction.go
```

guscarreon: On a second thought, I realize that this is more a refactoring effort than a simple nitpick and, given the need for this feature, we should probably sort it on a separate PR. 

Thoughts?

guscarreon: Nevermind my comment above, sometimes I tend to overcomplicate things. A far better approach could be to simply return from `processInterstitials(req)` the unmarshaled `Device.Ext` and modify `lmt.ModifyForIOS(req)` accordingly. What do you think? Given the time constraint, do you feel it's feasible?
```
260     // Populate any "missing" OpenRTB fields with info from other sources, (e.g. HTTP request headers).
261     deps.setFieldsImplicitly(httpRequest, req)
262
263 -   if err := processInterstitials(req); err != nil {
    +   if devExt, err := processInterstitials(req); err != nil {  // MODIFY TO RETURN `devExt`
264         errs = []error{err}
265         return
266     }
267
268 -   lmt.ModifyForIOS(req)
    +   lmt.ModifyForIOS(req, devExt)  // <--- MODIFY so there's no need to jsonparse `Device.Ext` now
269
270     errL := deps.validateRequest(req)
endpoints/openrtb2/auction.go
```

privacy/lmt/ios.go
```diff
@@ -0,0 +1,67 @@
...
+
+func modifyForIOS(req *openrtb.BidRequest, modifiers map[iosutil.VersionClassification]modifier) {
+	if !isRequestForIOS(req) {
+		return
+	}
+
+	versionClassification := iosutil.DetectVersionClassification(req.Device.OSV)
+	if modifier, ok := modifiers[versionClassification]; ok {
+		modifier(req)
+	}
```
bsardo, line 34: My first thought here was null object pattern but perhaps what you have is better.

bsardo (commented):
The logic is looking good; I have just a few comments there. I'll review the tests shortly.

privacy/lmt/ios.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"strings"
+
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/util/iosutil"
+)
+
+var (
+	int8Zero int8 = 1
```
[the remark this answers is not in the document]
SyntaxNode: jiFfM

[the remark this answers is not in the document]
SyntaxNode: ... fixed.

util/iosutil/iosutil.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// Version specifies the version of an iOS device.
+type Version struct {
+	Major int
+	Minor int
+}
+
+// ParseVersion parses the major.minor version for an iOS device.
+func ParseVersion(v string) (Version, error) {
```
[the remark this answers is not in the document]
SyntaxNode: I was using it from another package in an earlier draft. I don't see the harm in leaving it as exported even though its currently now just used by the iosutil's `DetectVersionClassification` method. If you feel strongly it should not be exported, I'll change it.

privacy/lmt/ios.go
SyntaxNode: I love that the null object pattern was in your mind. :)

I don't think it applies in this situation since the modifier func isn't a return value passed between methods. I consider the range coverage of > 14.0 to be a coincidence which may change in the future. I did not intend for the map to be a comprehensive of all possible iOS versions.

privacy/lmt/ios.go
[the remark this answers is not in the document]
SyntaxNode: I also updated the tests to use independent 0/1 values. The tests passed because I reused the same int8Zero reference.

endpoints/openrtb2/auction_test.go
SyntaxNode: I very much disliked my original solution and rewrote it.

privacy/lmt/ios.go
```diff
@@ -0,0 +1,67 @@
...
+
+	switch *atts {
+	case openrtb_ext.IOSAppTrackingStatusNotDetermined:
+		req.Device.Lmt = &int8Zero
+	case openrtb_ext.IOSAppTrackingStatusRestricted:
+		req.Device.Lmt = &int8One
+	case openrtb_ext.IOSAppTrackingStatusDenied:
+		req.Device.Lmt = &int8One
+	case openrtb_ext.IOSAppTrackingStatusAuthorized:
+		req.Device.Lmt = &int8Zero
```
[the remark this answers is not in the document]
SyntaxNode: I see. In general I agree with removing duplicate code. For this case, the two switch statements are doing very different things. I worry that combining them would lead to harder to read code since now `IsKnownIOSAppTrackingStatus` would both validate it's a known value and return the LMT override value, both of which are in different pacakges.

endpoints/openrtb2/auction.go
SyntaxNode: I'm unsure if this is worth fixing now, or if it's better to wait on the work @hhhjort is spearheading to solve this problem on a larger scale. Both `processInterstitials` and `ModifyForIOS` only unmarshal the device ext in certain circumstances, so I can't rely on `processInterstitials` actually performing the unmarshal.

privacy/lmt/ios.go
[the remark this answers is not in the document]
guscarreon: 😮 good catch!

util/iosutil/iosutil.go
```diff
@@ -0,0 +1,78 @@
...
+	Version142OrGreater
+)
+
+// DetectVersionClassification detects the iOS version classification.
+func DetectVersionClassification(v string) VersionClassification {
+	// exact comparisons first. no parsing required.
+	if v == "14.0" {
+		return Version140
+	}
+	if v == "14.1" {
```
guscarreon, line 66: Are we positive the `req.Device.OSV` will come with iOS versions written exactly like `"14.0"` and `"14.1"`? Do you think it'd be safer to try to match with a regexp instead? Or maybe use `string` library functions such as `string.Contains()` or `string.HasPrefix()`

privacy/lmt/ios.go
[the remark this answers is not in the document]
guscarreon: Got it. How do you feel about grouping the `case`s that return an `&int8One` and an `&int8Zero`
```
51   func modifyForIOS142OrGreater(req *openrtb.BidRequest) {
52       atts, err := openrtb_ext.ParseDeviceExtATTS(req.Device.Ext)
53       if err != nil || atts == nil {
54           return
55       }
56  
57       switch *atts {
58 -     case openrtb_ext.IOSAppTrackingStatusNotDetermined:
   +     case openrtb_ext.IOSAppTrackingStatusNotDetermined, openrtb_ext.IOSAppTrackingStatusAuthorized:
59           req.Device.Lmt = &int8Zero
60 -     case openrtb_ext.IOSAppTrackingStatusRestricted:
   +     case openrtb_ext.IOSAppTrackingStatusRestricted, openrtb_ext.IOSAppTrackingStatusDenied:
61           req.Device.Lmt = &int8One
62 -     case openrtb_ext.IOSAppTrackingStatusDenied:
63 -         req.Device.Lmt = &int8One
64 -     case openrtb_ext.IOSAppTrackingStatusAuthorized:
65 -         req.Device.Lmt = &int8Zero
66       }
67   }
privacy/lmt/ios.go
```

openrtb_ext/device.go
```diff
@@ -12,14 +14,58 @@ const PrebidExtKey = "prebid"
...
+	case IOSAppTrackingStatusNotDetermined:
+		return true
+	case IOSAppTrackingStatusRestricted:
+		return true
+	case IOSAppTrackingStatusDenied:
+		return true
+	case IOSAppTrackingStatusAuthorized:
+		return true
+	default:
+		return false
```
guscarreon, line 59: Since `IOSAppTrackingStatusNotDetermined`, `IOSAppTrackingStatusRestricted`, `IOSAppTrackingStatusDenied`, and `IOSAppTrackingStatusAuthorized` are equal to ints from 0 to 3, `IsKnownIOSAppTrackingStatus` could probably be simplified to:
```
48   func IsKnownIOSAppTrackingStatus(v int64) bool {
49 -     switch IOSAppTrackingStatus(v) {
50 -     case IOSAppTrackingStatusNotDetermined:
51 -         return true
52 -     case IOSAppTrackingStatusRestricted:
53 -         return true
54 -     case IOSAppTrackingStatusDenied:
55 -         return true
56 -     case IOSAppTrackingStatusAuthorized:
57 -         return true
58 -     default:
59 -         return false
60 -     }
   +     if v < IOSAppTrackingStatusNotDetermined || v > IOSAppTrackingStatusAuthorized {
   +         return false
   +     }
   +     return true
61   }
openrtb_ext/device.go
```

util/iosutil/iosutil_test.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		result := test.givenVersion.EqualOrGreater(givenMajor, givenMinor)
+		assert.Equal(t, test.expected, result, test.description)
+	}
+}
+
+func TestDetectVersionGroup(t *testing.T) {
```
bsardo, line 113: Nitpick: Can we change this test name to `TestDetectVersionClassification` to more closely match the function under test?

bsardo (commented):
Looks good, just one nitpick.

endpoints/openrtb2/auction.go
guscarreon: Agree

openrtb_ext/device.go
```diff
@@ -49,3 +95,26 @@ func (edi *ExtDeviceInt) UnmarshalJSON(b []byte) error {
...
 	return nil
 }
+
+// ParseDeviceExtATTS parses the ATTS value from the request.device.ext OpenRTB field.
+func ParseDeviceExtATTS(deviceExt json.RawMessage) (*IOSAppTrackingStatus, error) {
+	v, err := jsonparser.GetInt(deviceExt, "atts")
+
+	// node not found error
+	if err == jsonparser.KeyPathNotFoundError {
+		return nil, nil
```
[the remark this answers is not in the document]
SyntaxNode: Correct. I don't believe its fair to call the request invalid if the ATTS is missing and I'm not sure what the implications of that would be. In this PR I'm using its existence as criteria for applying the LMT override. If the ATTS field is not present, the code would do nothing.

privacy/lmt/ios_test.go
```diff
@@ -0,0 +1,271 @@
...
+		{
+			description: "Overwrites Existing",
+			givenDevice: openrtb.Device{Ext: json.RawMessage(`{"atts":3}`), Lmt: openrtb.Int8Ptr(1)},
+			expectedLMT: openrtb.Int8Ptr(0),
+		},
+		{
+			description: "Invalid Value - Unknown",
+			givenDevice: openrtb.Device{Ext: json.RawMessage(`{"atts":4}`), Lmt: nil},
+			expectedLMT: nil,
+		},
```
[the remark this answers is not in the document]
SyntaxNode: Sure. Added.

privacy/lmt/ios.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	versionClassification := iosutil.DetectVersionClassification(req.Device.OSV)
+	if modifier, ok := modifiers[versionClassification]; ok {
+		modifier(req)
+	}
+}
+
+func isRequestForIOS(req *openrtb.BidRequest) bool {
+	return req != nil && req.App != nil && req.Device != nil && strings.EqualFold(req.Device.OS, "ios")
```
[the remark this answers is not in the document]
SyntaxNode: We are not missing further checks. This covers the specification "app object exists" in the linked issue.

util/iosutil/iosutil.go
SyntaxNode: Yes. I asked in the linked issue if that was a valid assumption and the author confirmed it was, including a link back to the iOS version docs.

openrtb_ext/device.go
SyntaxNode: If it's alright with you, I'd prefer to keep it as a switch. The fact that the enumeration values are sequential are not a guaranteed feature of the value. I don't believe this would have a performance impact.

util/iosutil/iosutil_test.go
SyntaxNode: Ah, yes. Good eye. I renamed the method during a refactor and forgot to update the corresponding test.

privacy/lmt/ios.go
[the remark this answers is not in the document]
SyntaxNode: I was also concerned with readability. I don't expect this would impact performance.

guscarreon (approved):
LGTM

bsardo (approved):
LGTM

PR #1719, opened 2021-02-22
New Adapter: UNICORN

faithnh:
Also fixed doc: https://github.com/prebid/prebid.github.io/pull/2710


adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		secure := int8(1)
+		imp.Secure = &secure
+		imp.TagID = placementID
+	}
+	return request.Imp, nil
+}
+
+func getStoredRequestImpID(imp *openrtb.Imp) (string, error) {
```
SyntaxNode, line 170: The built in Go library isn't ideal for pulling back a single value from a JSON object. While this is valid and working code, please consider an abbreviated approach using the jsonparser library. Something like:

```
func getStoredRequestImpID(imp *openrtb.Imp) (string, error) {
	v, err := jsonparser.GetString(imp.Ext, "prebid", "storedrequest", "id")

	if err != nil && err != jsonparser.KeyPathNotFoundError {
		return "", fmt.Errorf("stored request id not found: %s", err)
	}

	return v, nil
}
```

EDIT: Ah, I see you're already using jsonparser in `setExt`.

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+	siteExt, err := json.Marshal(unicornSourceExt{
+		Stype:  "prebid_server_uncn",
+		Bidder: "unicorn",
+	})
+	if err != nil {
+		return nil, &errortypes.BadInput{
+			Message: fmt.Sprintf("Error while encoding source.ext, err: %s", err),
+		}
+	}
+	return siteExt, nil
```
SyntaxNode, line 210: It seems wasteful to marshal this for every call when it contains all static content. Consider hardcoding a json raw message or moving the marshal to do the builder if you'd like to continue doing this dynamically.

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(request.Imp))
+	bidResponse.Currency = response.Cur
+	for _, seatBid := range response.SeatBid {
+		for _, bid := range seatBid.Bid {
+			var bidType openrtb_ext.BidType
+			for _, imp := range request.Imp {
+				if imp.ID == bid.ImpID {
+					if imp.Banner != nil {
```
SyntaxNode, line 269: This approach for determining bid type is an anti-pattern. Does your bidding server return any extra information you can use instead?

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+	var response openrtb.BidResponse
+	if err := json.Unmarshal(responseData.Body, &response); err != nil {
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(request.Imp))
+	bidResponse.Currency = response.Cur
+	for _, seatBid := range response.SeatBid {
+		for _, bid := range seatBid.Bid {
+			var bidType openrtb_ext.BidType
```
SyntaxNode, line 266: We've recently been made aware of odd (IMHO) behavior in Go range loops. Since we take the address of bid below, we need to "pin" the value to this loop iteration using a new variable, ie:

```
  for _, bid := range seatBid.Bid {
    bid := bid
    var bidType openrtb_ext.BidType
```

This creates a new bid variable which shadows the reference from the loop iterator.

adapters/unicorn/unicorn_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderUnicorn, config.Adapter{
+		Endpoint: "http://localhost:4000"})
```
SyntaxNode, line 13: This doesn't need to be a real endpoint. Why are you specifying a port?

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		if len(request.Device.IPv6) > 0 {
+			headers.Add("X-Forwarded-For", request.Device.IPv6)
+		}
+
+		if len(request.Device.IP) > 0 {
+			headers.Add("X-Forwarded-For", request.Device.IP)
+		}
+	}
```
SyntaxNode, line 124: Please add test coverage for the existence of UA, IPv6, and IP.

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// MakeBids unpacks the server's response into Bids.
+func (a *UnicornAdapter) MakeBids(request *openrtb.BidRequest, requestData *adapters.RequestData, responseData *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+
+	if responseData.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if responseData.StatusCode == http.StatusBadRequest {
```
SyntaxNode, line 243: Please add test coverage for this condition.

SyntaxNode (commented):
Thank you for reading the our new bidder docs.

adapters/unicorn/unicorn.go
faithnh: I don't have the information from my bidding server need to determine the BidType, so I took this approach that other bidder adapters are doing.

By the way, most of the bidder adapters do not follow this mapping rule. What do you think?

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/buger/jsonparser"
+	"github.com/mxmCherry/openrtb"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// UnicornAdapter describes a UNICORN prebid server adapter.
+type UnicornAdapter struct {
```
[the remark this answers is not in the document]
faithnh: Fixed.

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+	Bidder string `json"bidder"`
+}
+
+// unicornExt is ext for UNICORN
+type unicornExt struct {
+	Prebid    *openrtb_ext.ExtImpPrebid `json:"prebid,omitempty"`
+	AccountID int64                     `json:"accountId,omitempty"`
+}
+
+// Builder builds a new instance of the Foo adapter for the given bidder with the given config.
```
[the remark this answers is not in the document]
faithnh: Fixed. Thank you!

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,285 @@
...
+
+		if len(request.Device.IP) > 0 {
+			headers.Add("X-Forwarded-For", request.Device.IP)
+		}
+	}
+
+	return headers
+}
+
+func setImps(request *openrtb.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]openrtb.Imp, error) {
```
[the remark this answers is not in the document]
faithnh: Fixed.

adapters/unicorn/unicorn.go
faithnh: I used above approach. fixed.

adapters/unicorn/unicorn.go
faithnh: I use hardcoding a json raw message. fixed.

adapters/unicorn/unicorn.go
faithnh: Fixed. Thank you!!!

adapters/unicorn/unicorn_test.go
faithnh: There is no reason. Changed to `https://ds.uncn.jp` for now.

adapters/unicorn/unicorn.go
faithnh: Fixed.

adapters/unicorn/unicorn.go
faithnh: I mistaked on 400.json. Fixed.

adapters/unicorn/unicorn.go
SyntaxNode: It's not a deal breaker. I always like to ask because publisher ad rendering may not work correctly if the bid type is incorrect. Please just try to match the behavior of your bidding server the best you can. For example, if you receive an imp with both banner and native sections defined, will your server prefer native? .. as this code would currently do.

faithnh: Certainly, there is a problem with the current conditional branching. understood.

Changed from "if" to "else if".

SyntaxNode (approved):

guscarreon (changes requested):
Looks pretty good. A  couple of comments before approving

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,244 @@
...
+			}
+		}
+	}
+
+	err := modifyImps(request, requestInfo)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	request.Source.Ext = setSourceExt()
```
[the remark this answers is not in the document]
faithnh: Fixed it. Thank you so much!

c27da8b

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,244 @@
...
+
+		if len(request.Device.IP) > 0 {
+			headers.Add("X-Forwarded-For", request.Device.IP)
+		}
+	}
+
+	return headers
+}
+
+func modifyImps(request *openrtb.BidRequest, requestInfo *adapters.ExtraRequestInfo) error {
```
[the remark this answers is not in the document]
faithnh: Certainly, `requestInfo *adapters.ExtraRequestInfo` is not used. Fixed it. Thank you!

c27da8b

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,251 @@
...
+		return nil, []error{err}
+	}
+
+	var modifiableSource *openrtb.Source
+	if request.Source != nil {
+		modifiableSource = request.Source
+	} else {
+		modifiableSource = &openrtb.Source{}
+	}
+	modifiableSource.Ext = setSourceExt()
```
SyntaxNode, line 78: This code does not actually make a copy of the source. Since the variable is defined as a pointer on line 72, it's still pointing to shared memory. You want to something more like this:

```
var modifiableSource openrtb.Source
if request.Source != nil {
	modifiableSource = *request.Source
}
modifiableSource.Ext = setSourceExt()
request.Source = &modifiableSource
```

I mocked out this situation in the Go Playground if you'd like to explore further: https://play.golang.org/p/rVCn1RzHs4g

faithnh: Oh, Sorry.I not understood it and mistaked...

Fixed it.

https://github.com/prebid/prebid-server/pull/1719/commits/fb58a7ee99d689075713ebb414a0f6cfd7b1412c

adapters/unicorn/unicorn.go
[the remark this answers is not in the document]
faithnh: Fixed again because there is have copying source problem.

https://github.com/prebid/prebid-server/commit/fb58a7ee99d689075713ebb414a0f6cfd7b1412c

adapters/unicorn/unicorn.go
SyntaxNode: No worries. :)

SyntaxNode (approved):

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,251 @@
...
+		} else {
+			placementID, err = getStoredRequestImpID(imp)
+			if err != nil {
+				return &errortypes.BadInput{
+					Message: fmt.Sprintf("Error get StoredRequestImpID from imp[%d]: %s", i, err),
+				}
+			}
+		}
+
+		ext.Bidder.PlacementID = placementID
```
guscarreon, line 149: The fact that there's a scenario where `ext.Bidder.PlacementID` does not get modified at all makes for an opportunity to maybe refactor the `if-else` statement in line 138 to maybe something like:
```
128           var ext unicornImpExt
129           err := json.Unmarshal(imp.Ext, &ext)
130
131           if err != nil {
132               return &errortypes.BadInput{
133                   Message: fmt.Sprintf("Error while decoding imp[%d].ext: %s", i, err),
134               }
135           }
136
137 -         var placementID string
138 -         if ext.Bidder.PlacementID != "" {
139 -             placementID = ext.Bidder.PlacementID
140 -         } else {
    +         if ext.Bidder.PlacementID == "" {
141 -             placementID, err = getStoredRequestImpID(imp)
    +             ext.Bidder.PlacementID, err = getStoredRequestImpID(imp)
142               if err != nil {
143                   return &errortypes.BadInput{
144                       Message: fmt.Sprintf("Error get StoredRequestImpID from imp[%d]: %s", i, err),
145                   }
146               }
147           }
148
149 -         ext.Bidder.PlacementID = placementID
150
151           imp.Ext, err = json.Marshal(ext)
152           if err != nil {
153               return &errortypes.BadInput{
154                   Message: fmt.Sprintf("Error while encoding imp[%d].ext: %s", i, err),
155               }
156           }
157
158           secure := int8(1)
159           imp.Secure = &secure
160 -         imp.TagID = placementID
    +         imp.TagID = ext.Bidder.PlacementID
adapters/unicorn/unicorn.go
```

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,251 @@
...
+	for _, seatBid := range response.SeatBid {
+		for _, bid := range seatBid.Bid {
+			bid := bid
+			var bidType openrtb_ext.BidType
+			for _, imp := range request.Imp {
+				if imp.ID == bid.ImpID {
+					if imp.Banner != nil {
+						bidType = openrtb_ext.BidTypeBanner
+					} else if imp.Native != nil {
+						bidType = openrtb_ext.BidTypeNative
```
guscarreon, line 239: The `true` branch of this `if` statement is not covered in the tests cases. Could we add a native JSON test that covers it?

guscarreon (commented):
Looks pretty solid so far. Have a couple more questions.

adapters/unicorn/unicorn.go
faithnh: Native will support it in the future, so I left it, but since it is not currently supported, I will remove this branch.

adapters/unicorn/unicorn.go
```diff
@@ -0,0 +1,251 @@
...
+		imp.Ext, err = json.Marshal(ext)
+		if err != nil {
+			return &errortypes.BadInput{
+				Message: fmt.Sprintf("Error while encoding imp[%d].ext: %s", i, err),
+			}
+		}
+
+		secure := int8(1)
+		imp.Secure = &secure
+		imp.TagID = placementID
```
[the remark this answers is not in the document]
faithnh: No, `modifyImps` is rewriting imp to include the required data.
For example, it is implemented to put `placementId` or `storedRequestImpId` for `Imp[i].TagID`.

adapters/unicorn/unicorn.go
faithnh: Fixed it

deaf6c0

adapters/unicorn/unicorn.go
faithnh: Certainly, fixed it. Thanks you.

deaf6c0

SyntaxNode (approved):

adapters/unicorn/unicorn.go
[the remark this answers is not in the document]
guscarreon: Ohh you are right. Misread line 126 as a local copy of the imp (`imp := *request.Imp[i]`) instead of `imp := &request.Imp[i]`

guscarreon (approved):
@faithnh thank you for addressing the feedback. Your adapter looks pretty solid

PR #1723, opened 2021-02-24
AdForm: Fix Regs Parsing

SyntaxNode:
The standard Go JSON parser will return an error if the input is nil. Adding a nil check and a test case.

@braizhas Could you please review?

### DIFF
--- a/adapters/adform/adform.go
+++ b/adapters/adform/adform.go
@@ -453,7 +453,7 @@ func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []erro
 
 	gdprApplies := ""
 	var extRegs openrtb_ext.ExtRegs
-	if request.Regs != nil {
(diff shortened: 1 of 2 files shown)


adapters/adform/adform.go
```diff
@@ -466,7 +466,7 @@ func openRtbToAdformRequest(request *openrtb.BidRequest) (*adformRequest, []erro
 
 	eids := ""
 	consent := ""
-	if request.User != nil {
+	if request.User != nil && request.User.Ext != nil {
```
SyntaxNode, line 469: Not strictly required, but adding as a slight performance boost. No reason to try unmarshaling when we know it will fail.

braizhas (approved):
LGTM

mansinahar (approved):

guscarreon (approved):
LGTM

Issue #1650, opened 2021-01-12
Turn Off Debugging By Publisher, By Bidder


SyntaxNode:
_This issue now incorporates the requirements from https://github.com/prebid/prebid-server/issues/510, combing proposals from @hhhjort and @bretg._

## Description

PBS has a [debug feature](https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#test) that publishers can enable to get extra details in their response from Prebid Server, including the resolved stored request and requests/responses to/from bidding servers (url, status code, body, headers).

It's possible for malicious users to submit a request to Prebid Server with the intent of pulling back stored requests for another publisher or for retrieving secret tokens or ids from requests to bidding servers. We'd like to point out that Prebid.js has no possibility of hiding this information since all data is proxied through the user's browser. It's only with Prebid Server that we can consider enhanced protections for hosts and publishers.

For example, the host specific values for the new Recontent https://github.com/prebid/prebid-server/pull/1622 adapter might want to be secured.

## Proposal

Build new options in Prebid Server to allow publishers to disable debugging for their requests using account level settings and to allow hosts to disable debugging of specific bidders. However, disabling the debug information has a very real side effect of making it hard to track down problems. Hosts may run a separate debug instance on their servers or local machines with the debug restrictions removed, or this proposal also includes an authorization header token which will bypass any debug disabled settings. It's my intention for authorization tokens to be used by the host to debug issues on behalf of their publishers.

### Publisher Controls

Add a new account level setting `debug-allow` which defaults to true. I don't expect further fine grain controls added to this feature at the account level, so the setting is a single boolean value. When false, no debugging information at all is sent back to the publisher and a warning is emitted in the response:

Account JSON For PBS-Go:
```
{
  "debug_allow": true
}
```

Response:
```
{
  "ext": {
    "warnings":  {
      "general": [{code: 10002, message: "debug turned off for account"}]
    }
  }
}
```

Warning codes in PBS-Go begin with 10000 and the next available value is 10002. The code allows publishers to look for specific type of warnings without relying on string compare.

### Bidder Controls

Add a new bidder level settings `debug.allow` which defaults to true. I can imagine fine grain controls added in the future to possibly restrict just specific kinds of the debug information such as the url, body, or headers. Therefore, I'm proposing a debug object for forward compatibility. When false, no debugging information for this specific bidder is sent back to the publisher, except for a warning in the response:

Bidder YAML:
```
debug:
  allow: true
[...]

PR #1724, opened 2021-02-25
Debug warnings

VeronikaSolovei9:


errortypes/code.go
```diff
@@ -17,6 +17,8 @@ const (
 const (
 	UnknownWarningCode               = 10999
 	InvalidPrivacyConsentWarningCode = iota + 10000
+	AccountLevelDebugDisabledWarningCode
+	BidderLevelDebugDisabledWarningCode
```
VeronikaSolovei9, line 21: These values are 10002 and 10003. According to https://github.com/prebid/prebid-server/issues/1650 these values should be 10001 and 10002. Is it ok?

exchange/bidder.go
```diff
@@ -87,6 +87,8 @@ type pbsOrtbSeatBid struct {
 	// if len(bids) > 0, this will become response.seatbid[i].ext.{bidder} on the final OpenRTB response.
 	// if len(bids) == 0, this will be ignored because the OpenRTB spec doesn't allow a SeatBid with 0 Bids.
 	ext json.RawMessage
+
+	warnings map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderError
```
VeronikaSolovei9, line 91: Warnings are actually Errors. Should I use errortypes.Warning instead? If yes - I need to add Code in it:
```
type Warning struct {
	Message string
	Code int
}
```

SyntaxNode: I agree it's a bit strange to see warnings as a collection of errors. This only a concern in code with the type name. I'm not sure its worth having two separate but identical data structures with different names, but I don't feel too strongly if you prefer to see `ExtBidderWarning`.

errortypes/code.go
hhhjort: Does that mean `InvalidPrivacyConsentWarningCode` is set to 10001?

exchange/bidder.go
hhhjort: It is reasonable to me that warnings are a subset of go "errors". If you need to define `Warning` in more than one package, then I would just add it to errortypes. If it appears in just one place, I don't have a strong feeling on it.

hhhjort: Too be clear, I don't have an issue with `warnings map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderError` appearing in multiple places, just the idea of defining a `Warning` struct multiple places.

errortypes/code.go
VeronikaSolovei9: Yes, `InvalidPrivacyConsentWarningCode` is now 10001. 
If I comment out `//UnknownWarningCode               = 10999` then `InvalidPrivacyConsentWarningCode` is  10000. Looks like it depends on a position of all declared constants.

exchange/bidder.go
VeronikaSolovei9: This new(or updated) Warning should also have json properties mapping.
```
type ExtBidderError struct {
	Code    int    `json:"code"`
	Message string `json:"message"`
}
```
Please advice. I'll do what you guys think is better.

VeronikaSolovei9: Another thing. Do we want to separate "models" from "views"? 
By model I mean `Warning` struct without json properties mapping, we use it "internally" only.
By "views" I mean another `Warning` struct we will Marshal to response. 
Now I use same `ExtBidderError` for everything.

errortypes/code.go
hhhjort: I suspect that it is line 18 that is occupying the iota=0 spot, so at line 19 iota=1 ... would fix it, and probably cleaner, to move the `UnknownWarningCode` to the bottom of the list ... would be where it belongs sequentially anyway.

exchange/bidder.go
hhhjort: I lean towards just using `ExtBidderError` and using the `Code` value to determine the type. A "warning" can also be called a "non-fatal error", it is a problem, just not a problem severe enough to require aborting the processing. A new struct that is identical to an old struct should really just be used when type safety is an important concern. There is a large channel (communicating issues out to the requester) where we want to treat them interchangeably, to a degree, and just a few places where we want to separate them out. I am open to being convinced otherwise.

errortypes/code.go
VeronikaSolovei9: Done

exchange/bidder.go
VeronikaSolovei9: I'm confused. 
`Code` in `ExtBidderError` is an error number like `UnknownWarningCode = 10999`, we need to set these values.
`Code` in `Warning` is Severity level. 
Do you mean I need to "consolidate" these structures?

hhhjort: I reviewed again and I see the warnings are their own thing now in the code. I suppose it depends on future plans, if we want to reuse some of the error machinery for warnings in the future, then we should keep them consolidated. If this is the extent we want to have warnings, then we can consider making them their own struct.

VeronikaSolovei9: Per morning discussion I renamed ExtBidderError to more generic ExtBidderMessage. 
Should we leave it all as it is now?

errortypes/code.go
VeronikaSolovei9: Reverted back per morning discussion with team.

exchange/bidder.go
hhhjort: I am good with that.

hhhjort (dismissed):
LGTM

exchange/bidder.go
```diff
@@ -87,6 +87,8 @@ type pbsOrtbSeatBid struct {
 	// if len(bids) > 0, this will become response.seatbid[i].ext.{bidder} on the final OpenRTB response.
 	// if len(bids) == 0, this will be ignored because the OpenRTB spec doesn't allow a SeatBid with 0 Bids.
 	ext json.RawMessage
+
+	warnings map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage
```
SyntaxNode, line 91: Is the idea to create a new pathway for reporting on warnings? Right now, warnings are error types with a lower severity code. This looks to change that approach and I'm unsure how that will fit in with our current error handling infrastructure.

SyntaxNode, line 91: The `pbsOrtbSeatBid` object is already bidder specific. The mapping from bidder to a map a bidder errors is make in exchange/exchange.go#makeExtBidResponse.

```
if len(responseExtra.Errors) > 0 {
	bidResponseExt.Errors[bidderName] = responseExtra.Errors
}
if len(errList) > 0 {
	bidResponseExt.Errors[openrtb_ext.PrebidExtKey] = errsToBidderErrors(errList)
}
```

I was originally thinking this would look something like:
```
if errs := errortypes.FatalOnly(responseExtra.Errors); len(errs) > 0 {
	bidResponseExt.Errors[openrtb_ext.PrebidExtKey] = errsToBidderErrors(errs)
}

if warnings := errortypes.WarningOnly(responseExtra.Errors); len(warnings) > 0 {
	bidResponseExt.Warnings[openrtb_ext.PrebidExtKey] = errsToBidderWarnigs(warnings)
}
```

endpoints/openrtb2/amp_auction.go
```diff
@@ -35,10 +35,10 @@ import (
...
 
 type AmpResponse struct {
-	Targeting map[string]string                                       `json:"targeting"`
-	Debug     *openrtb_ext.ExtResponseDebug                           `json:"debug,omitempty"`
-	Errors    map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderError `json:"errors,omitempty"`
-	Warnings  map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderError `json:"warnings,omitempty"`
+	Targeting map[string]string                                         `json:"targeting"`
+	Debug     *openrtb_ext.ExtResponseDebug                             `json:"debug,omitempty"`
+	Errors    map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage `json:"errors,omitempty"`
+	Warnings  map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage `json:"warnings,omitempty"`
```
SyntaxNode, line 41: Nice. I like the new choice of `ExtBidderMessage` name.

exchange/bidder.go
```diff
@@ -168,10 +171,17 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb.Bi
...
+			if accountDebugAllowed && bidder.config.DebugInfo.Allow {
 				seatBid.httpCalls = append(seatBid.httpCalls, makeExt(httpInfo))
 			}
+			if accountDebugAllowed && !bidder.config.DebugInfo.Allow {
+				debugDisabledWarning := openrtb_ext.ExtBidderMessage{
+					Code:    errortypes.BidderLevelDebugDisabledWarningCode,
+					Message: "debug turned off for bidder",
+				}
+				seatBid.warnings[bidder.BidderName] = append(seatBid.warnings[bidder.BidderName], debugDisabledWarning)
+			}
```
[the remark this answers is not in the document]
SyntaxNode: With other comment about representing warnings as error types, this would be:
```
errs = append(errs, <new debug warning>)
```

exchange/exchange.go
```diff
@@ -66,7 +66,8 @@ type exchange struct {
 // Container to pass out response ext data from the GetAllBids goroutines back into the main thread
 type seatResponseExtra struct {
 	ResponseTimeMillis int
-	Errors             []openrtb_ext.ExtBidderError
+	Errors             []openrtb_ext.ExtBidderMessage
+	Warnings           map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage
```
SyntaxNode, line 70: There wouldn't be a need to add `Warnings` here. They would still be represented as `Errors` at this point.

VeronikaSolovei9: `openrtb_ext.ExtBidderMessage` doesn't support severity level. To consolidate Errors and Warnings we need to modify it and have `message`, `code` and `severity level` properties in it. We discussed this with @hhhjort here: https://github.com/prebid/prebid-server/pull/1724#discussion_r582575208

exchange/bidder.go
VeronikaSolovei9: Good point about bidder specific warning. I modified it to list.
To use one list for both Errors and Warning I need to use `error` (or consolidated struct) instead of  `openrtb_ext.ExtBidderMessage`. Should I process with this?

hhhjort (dismissed):
LGTM

SyntaxNode (commented):
I successfully performed end to end testing with the auction endpoint for account and bidder debug warnings. Does this include AMP and Video endpoint warning wire-up?

I wasn't able to get a CCPA warning for the auction endpoint. Do you want to wire that up now or in a follow-up PR?

exchange/exchange_test.go
```diff
@@ -393,7 +405,9 @@ func TestTwoBiddersDebugDisabledAndEnabled(t *testing.T) {
 	e.currencyConverter = currency.NewRateConverter(&http.Client{}, "", time.Duration(0))
 	e.categoriesFetcher = categoriesFetcher
 
-	debugLog := DebugLog{}
+	//ctx := context.Background()
+	//ctx = context.WithValue(ctx, DebugContextKey, true)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Removed

exchange/exchange_test.go
```diff
@@ -424,6 +438,7 @@ func TestTwoBiddersDebugDisabledAndEnabled(t *testing.T) {
 		}
 		// Run test
 		outBidResponse, err := e.HoldAuction(context.Background(), auctionRequest, &debugLog)
+		fmt.Println(string(outBidResponse.Ext))
```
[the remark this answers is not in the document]
VeronikaSolovei9: Removed

VeronikaSolovei9:
I tested it for video and auction endpoints, but not for AMP. 
What is the use case for CCPA warning?

SyntaxNode:
> What is the use case for CCPA warning?

The CCPA string is invalid. It is removed from the request and a warning is returned to the publisher. An example invalid is the literal string "invalid".

VeronikaSolovei9:
Still need to add more tests for AMP endpoint

VeronikaSolovei9:
Ready for review

hhhjort (dismissed):
LGTM

endpoints/openrtb2/auction_test.go
```diff
@@ -2104,6 +2106,54 @@ func TestIOS14EndToEnd(t *testing.T) {
...
+
+	deps.Auction(recorder, req, nil)
+
+	if recorder.Code != http.StatusOK {
+		t.Errorf("Endpoint should return a 200")
+	}
+	warnings := deps.ex.(*warningsCheckExchange).auctionRequest.Warnings
+	assert.Len(t, warnings, 1, "One warning should be returned from exchange")
+
+	actualWarning := warnings[0].(*errortypes.Warning)
```
SyntaxNode, line 2140: This will panic if the warnings don't exist. I recommend protecting this logic with modifying assert.Len above:

```
if !assert.Len(t, warnings, 1, "One warning should be returned from exchange") {
  t.FailNow()
}
```

exchange/exchange.go
```diff
@@ -236,6 +238,27 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+		accountDebugDisabledWarning := openrtb_ext.ExtBidderMessage{
+			Code:    errortypes.AccountLevelDebugDisabledWarningCode,
+			Message: "debug turned off for account",
+		}
+		bidResponseExt.Warnings[generalWarningKey] = append(bidResponseExt.Warnings[generalWarningKey], accountDebugDisabledWarning)
+	}
+
+	if len(r.Warnings) != 0 {
+		for _, warning := range r.Warnings {
+			typedWarning := warning.(*errortypes.Warning)
```
SyntaxNode, line 253: We shouldn't code against a specific struct here. Please just rely on the interfaces:
```
generalWarning := openrtb_ext.ExtBidderMessage{
  Code:    errortypes.ReadCode(warning),
  Message: warning.Error()
}
```

endpoints/openrtb2/auction.go
```diff
@@ -179,6 +180,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		RequestType:  labels.RType,
 		StartTime:    start,
 		LegacyLabels: labels,
+		Warnings:     warnings,
```
SyntaxNode, line 183: To Other Reviewers: This is necessary evil right now since HoldAuction is also composing the endpoint response. We'll need to separate these behaviors in the future, but that's too much work to do here.

exchange/exchange_test.go
```diff
@@ -180,59 +181,68 @@ func TestDebugBehaviour(t *testing.T) {
...
 	type aTest struct {
-		desc      string
-		in        inTest
-		out       outTest
-		debugData debugData
+		desc            string
+		in              inTest
+		out             outTest
+		debugData       debugData
+		generalWarnings bool
```
SyntaxNode, line 188: Should this be `generateWarnings`? Looks like it will add a warning if true.

exchange/exchange_test.go
```diff
@@ -305,6 +315,13 @@ func TestDebugBehaviour(t *testing.T) {
 			UserSyncs:  &emptyUsersync{},
 			StartTime:  time.Now(),
 		}
+		if test.generalWarnings {
+			var errL []error
+			errL = append(errL, &errortypes.Warning{
+				Message:     fmt.Sprintf("CCPA consent is invalid and will be ignored."),
```
SyntaxNode, line 321: Please use an obviously fake error message for tests.

exchange/exchange.go
```diff
@@ -236,6 +238,27 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 		}
 	}
 
+	generalWarningKey := openrtb_ext.BidderName(openrtb_ext.BidderReservedGeneral)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good point, done!

exchange/exchange.go
```diff
@@ -236,6 +238,27 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+
+	if !r.Account.DebugAllow && requestDebugInfo {
+		accountDebugDisabledWarning := openrtb_ext.ExtBidderMessage{
+			Code:    errortypes.AccountLevelDebugDisabledWarningCode,
+			Message: "debug turned off for account",
+		}
+		bidResponseExt.Warnings[generalWarningKey] = append(bidResponseExt.Warnings[generalWarningKey], accountDebugDisabledWarning)
+	}
+
+	if len(r.Warnings) != 0 {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Right, fixed

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -395,103 +395,107 @@ func TestCCPAConsent(t *testing.T) {
...
-		analyticsConf.NewPBSAnalytics(&config.Analytics{}),
-		map[string]string{},
-		[]byte{},
-		openrtb_ext.BuildBidderMap(),
-	)
+		// Build Request
+		var regs *openrtb_ext.ExtRegs
+		if !testCase.nilRegs {
+			regs = &openrtb_ext.ExtRegs{USPrivacy: testCase.regs}
+		}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, now looks better, done

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -395,103 +395,107 @@ func TestCCPAConsent(t *testing.T) {
...
-		t.Fatalf("Failed to marshal the complete openrtb.BidRequest object %v", err)
+func TestConsentWarnings(t *testing.T) {
+	type inputTest struct {
+		nilRegs           bool
+		regs              string
+		invalidConsentURL bool
+		bidderWarnings    bool
+	}
+	testData := []inputTest{
+		{true, "", false, false},
```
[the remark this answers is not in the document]
VeronikaSolovei9: Done

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -395,103 +395,107 @@ func TestCCPAConsent(t *testing.T) {
...
+		} else {
+			if !testCase.invalidConsentURL {
+				assert.Len(t, response.Warnings["appnexus"], 1, "1 bidder warning should be present")
+				assert.Len(t, response.Warnings["general"], 0, "General warnings should not be present")
+
+			} else {
+				assert.Len(t, response.Warnings["appnexus"], 1, "1 bidder warning should be present")
+				assert.Len(t, response.Warnings["general"], 2, "2 General warnings should be present")
+			}
+		}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Done!

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: Done

exchange/exchange.go
VeronikaSolovei9: Good point, fixed!

exchange/exchange_test.go

exchange/exchange_test.go
VeronikaSolovei9: Done

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -962,6 +966,24 @@ func (m *mockAmpExchange) HoldAuction(ctx context.Context, r exchange.AuctionReq
...
+	//m.lastRequest = r.BidRequest
+	response := &openrtb.BidResponse{
+		SeatBid: []openrtb.SeatBid{{
+			Bid: []openrtb.Bid{{
+				AdM: "<script></script>",
+				Ext: json.RawMessage(`{ "prebid": {"targeting": { "hb_pb": "1.20", "hb_appnexus_pb": "1.20", "hb_cache_id": "some_id"}}}`),
+			}},
+		}},
+		Ext: json.RawMessage(`{ "warnings": {"appnexus": [{"code": 10003, "message": "debug turned off for bidder"}] }}`),
+	}
```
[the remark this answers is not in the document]
VeronikaSolovei9: This is expected warning that will be compared with real warning. I cannot change it to test looking warning.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -962,6 +966,24 @@ func (m *mockAmpExchange) HoldAuction(ctx context.Context, r exchange.AuctionReq
 	return response, nil
 }
 
+type mockAmpExchangeWarnings struct {
+	//lastRequest *openrtb.BidRequest
```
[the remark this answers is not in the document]
VeronikaSolovei9: Removed

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -395,103 +395,106 @@ func TestCCPAConsent(t *testing.T) {
...
-		analyticsConf.NewPBSAnalytics(&config.Analytics{}),
-		map[string]string{},
-		[]byte{},
-		openrtb_ext.BuildBidderMap(),
-	)
+	testData := []inputTest{
+		{nil, false, nil},
+		{nil, true, tc2Wrnings},
+		{&openrtb_ext.ExtRegs{USPrivacy: "invalid"}, true, tc3Wrnings},
+		{&openrtb_ext.ExtRegs{USPrivacy: "1NYN"}, false, tc4Wrnings},
```
SyntaxNode, line 429: Nice. I think we can go one step further to using named field setters and embedding the expected warnings:
```
testData := []inputTest{
	{
		regs:              nil,
		invalidConsentURL: false,
		expectedWarnings:  nil,
	},
	{
		regs:              nil,
		invalidConsentURL: true,
		expectedWarnings: map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage{
			openrtb_ext.BidderReservedGeneral: {invalidCCPAWarning},
		},
	},
	{
		regs:              &openrtb_ext.ExtRegs{USPrivacy: "invalid"},
		invalidConsentURL: true,
		expectedWarnings: map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage{
			openrtb_ext.BidderReservedGeneral: {invalidCCPAWarning, invalidConsentWarning},
			openrtb_ext.BidderAppnexus:        {bidderWarning},
		},
	},
	{
		regs:              &openrtb_ext.ExtRegs{USPrivacy: "1NYN"},
		invalidConsentURL: false,
		expectedWarnings: map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage{
			openrtb_ext.BidderAppnexus: {bidderWarning},
		},
	},
}
```

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -962,6 +965,22 @@ func (m *mockAmpExchange) HoldAuction(ctx context.Context, r exchange.AuctionReq
 	return response, nil
 }
 
+type mockAmpExchangeWarnings struct{}
+
+func (m *mockAmpExchangeWarnings) HoldAuction(ctx context.Context, r exchange.AuctionRequest, debugLog *exchange.DebugLog) (*openrtb.BidResponse, error) {
+	//m.lastRequest = r.BidRequest
```
SyntaxNode, line 971: Please remove commented out code.

exchange/exchange.go
```diff
@@ -236,6 +238,23 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+	if !r.Account.DebugAllow && requestDebugInfo {
+		accountDebugDisabledWarning := openrtb_ext.ExtBidderMessage{
+			Code:    errortypes.AccountLevelDebugDisabledWarningCode,
+			Message: "debug turned off for account",
+		}
+		bidResponseExt.Warnings[openrtb_ext.BidderReservedGeneral] = append(bidResponseExt.Warnings[openrtb_ext.BidderReservedGeneral], accountDebugDisabledWarning)
+	}
+
+	for _, warning := range r.Warnings {
+		//typedWarning := warning.(*errortypes.Warning)
```
SyntaxNode, line 250: Please remove commented out code.

hhhjort (approved):

SyntaxNode (approved):

Issue #1728, opened 2021-03-01
Beachfront: Empty array returned from endpoint should result in 204 response


muncha:
As per email from Karen Lim <karen.lim@xandr.com>:

Reaching out regarding Status 200 with empty responsebody being returned for Gourmet Ads’ bid request. Wondering if it is possible for Beachfront to return Status 204 instead of 200 with error since this isn’t an error but an empty response using [https://github.com/prebid/prebid-server/blob/master/adapters/beachfront/beachfront.go#L488]? Thank you!

muncha:
I'll put in a PR for this today. I'm not sure if I really should have made an issue for it, as I can't assign it to myself.

SyntaxNode:
Thank you for looking into the issue. I think the key is for Beachfront to not return an error if there are no bids.

muncha:
Yep, that works as far as not throwing an error , but changing the response code that is reported seems a bit trickier than I naively assumed. I'm going to get with the maintainer of that endpoint and see If I can get a 204 instead of the current response.

muncha:
Beachfront is setting up an optional parameter on the endpoint that will send the 204.

SyntaxNode:
@muncha I believe this work is complete. May we close this issue?

muncha:
Yes please.

muncha:
I've got it

PR #1729, opened 2021-03-01
Corrected response for empty array return from banner endpoint (Issue #1728)

muncha:
This covers issue #1728


muncha:
I'm closing this, as further testing looks like it causes problems when both banner and video requests are included.

Issue #1732, opened 2021-03-02
Remove validation for native enumerations


YuriyVelichkoOpenX:
## Issue

According to the Open Measurement (OM) spec, the ad request should contain the **event tracker with custom type 555** to signal the demand side that SDK supports OM. See section Native Bid Response in OpenRTB p.28 of [Onboarding Guide for Integration Partners](https://s3-us-west-2.amazonaws.com/omsdk-files/docs/Open+Measurement+SDK+Onboarding_version_1.2.pdf)

```
{\"event\":555,\"methods\":[2]}
```

The bid requests with such a tracker leads to the server error:

```
Invalid request: json: cannot unmarshal number 555 into Go struct field EventTracker.eventtrackers.event of type native.EventType
```

## Objective

Need to provide publishers an ability to request native ads with any exchange-specific event tracker including the Open Measurement. 

## Additional info

The [IAB spec](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf) for native ads:
4.1 Native Markup Request Object
4.7 Event Trackers Request Object
7.6 **Event Types Table** - the most important section for our case

Example of native markup in the request:

```
"native": {
        "request": "{\"assets\":[{\"required\":1,\"title\":{\"len\":90}},{\"img\":{\"hmin\":50,\"type\":1,\"wmin\":50},\"required\":1},{\"img\":{\"hmin\":50,\"type\":3,\"wmin\":150},\"required\":1},{\"data\":{\"type\":2},\"required\":1},{\"data\":{\"type\":12},\"required\":1},{\"data\":{\"type\":1},\"required\":1}],\"context\":2,\"contextsubtype\":20,\"eventtrackers\":[{\"event\":1,\"methods\":[1,2]},{\"event\":555,\"methods\":[2]}],\"plcmttype\":1,\"ver\":\"1.2\"}",
        "ver": "1.2"
      }
```

The fast troubleshooting search has shown that the issue is related to  OpenRTB library
https://github.com/mxmCherry/openrtb/blob/v11.0.0/native/event_type.go#L4

SyntaxNode:
Thank you for the detailed report.

We use a third party library to represent the OpenRTB data structures. They have the `native.EventType` as an 8-bit integer, but more recent updates change that to a larger variable type. We are currently using an older version. I'll put out a PR to bring us up to date.

laurb9:
[...]

PR #1733, opened 2021-03-02
Update OpenRTB Library

SyntaxNode:
Updates mxmCherry/openrtb from 11.0.0 to 14.0.0. We were several years out of date. :-/

Breaking changes were introduced in the versions since, namely switching all `uint` to `int`. Consequently, this means our validation layers will now allow negative follows to the following OpenRTB objects:

- Audio: sequence, minbitrate, maxbitrate, maxseq
- Banner: w, h, wmax, hmax, wmin, hmin
- Bid: bid w, h, wratio, hratio
- Content: episode
- Device: h, w, ppi
- Format: w, h, wratio, hratio, wmin
- Geo: accuracy
- Video: w, h, minbitrate, maxbitrate

How do we want to handle these? Should we add new validations to continue enforcing positive values only?

There are also changes to the native image object's w, h, wmin, hmin values. I included validation for those in the PR already since there are specific tests for those validations.

For now, I also left the adapter models alone as `uint` and cast them to `int` to properly fit into the open rtb data model now. This could overflow, but the practical values we use in open rtb requests shouldn't be a concern.

Resolves: https://github.com/prebid/prebid-server/issues/1732


adapters/33across/33across.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/mxmCherry/openrtb"
+	"github.com/mxmCherry/openrtb/v14/openrtb2"
```
guscarreon, line 8: This import change makes us go from `openrtb.AnObject` to `openrtb2.AnObject` everywhere in the codebase. But, given that overtime we'll keep the OpenRTB library up to date, eventually we might get to `"github.com/.../openrtb3"` and we'll have to change every single `openrtb3.AnObject` again. Do you think we could prevent this (and simplify this PR) if we keep naming the import as `openrtb` instead?
```
  -     "github.com/mxmCherry/openrtb"
  +     openrtb "github.com/mxmCherry/openrtb/v14/openrtb2"
```

SyntaxNode:
So far, the prevailing preference is to add validation in PBS-Core to continue ensuring non-negative values are provided in the request for the fields listed above. For adapters using `uint`s in their ext data models, we'll leave those as-is and cast them to `int` when interacting with the updated openrtb data models. This does leave the potential for an overflow, but we don't expect any of the values to approach the overflow limits in practical use.

adapters/33across/33across.go
laurb9: Looks like there are some [structural differences between openrtb2 and openrtb3](https://github.com/InteractiveAdvertisingBureau/openrtb3-converter/blob/master/openrtb-3-converter-mappings.pdf), so I see both modules living together for a long time after openrtb3 is introduced.

I don't have an opinion on doing the naming change now or later, but I do think the change will be needed eventually.

laurb9:
I am curious about the reasoning behind the library making those obviously positive quantities signed.
Is there additional meaning for the negative values or it is simply a literal take on the spec ?

One reason to skip the additional validation right now is that all the existing configurations live must be using positive values or they wouldn't work, so negative values would have an obvious impact.

SyntaxNode:
I'm curious for the reasoning too. I'm not aware of any valid cases for negative integers. I don't see an explanation in their relevant PRs, so I opened https://github.com/mxmCherry/openrtb/issues/43.

> One reason to skip the additional validation right now is that all the existing configurations live must be using positive values or they wouldn't work, so negative values would have an obvious impact.

To be honest, I'm a bit scared of what will happen when PBS changes its current validation (implicitly) and starts sending negative values to adapters. I can't be sure if there's anyone relying on the non-negative validation check today. Why do you think the impact of negative values would be obvious? As a PBS host, I've encountered a lot more subtle issues related to publisher input than obvious ones.

adapters/33across/33across.go
SyntaxNode: I agree with @laurb9 for renaming the package to `openrtb2` now rather than aliasing back to `openrtb`, causing us trouble in the future when we take the plunge to add OpenRTB 3 support.

SyntaxNode:
PBS-Java validates positive integers for:

Banner: w, h
Format: w, h, wratio, hratio, wmin

PBS-Java *does not* validate positive integers for:

Audio: sequence, minbitrate, maxbitrate, maxseq
Banner: wmax, hmax, wmin, hmin (these values are deprecated)
Bid: w, h, wratio, hratio
Content: episode
Device: h, w, ppi
Geo: accuracy
Video: w, h, minbitrate, maxbitrate

endpoints/openrtb2/auction.go
```diff
@@ -729,21 +731,37 @@ func validateNativeAssetTitle(title *nativeRequests.Title, impIndex int, assetIn
...
+		return fmt.Errorf("request.imp[%d].native.request.assets[%d].img.h must be a positive integer", impIndex, assetIndex)
+	}
+	if img.WMin < 0 {
+		return fmt.Errorf("request.imp[%d].native.request.assets[%d].img.wmin must be a positive integer", impIndex, assetIndex)
+	}
+	if img.HMin < 0 {
+		return fmt.Errorf("request.imp[%d].native.request.assets[%d].img.hmin must be a positive integer", impIndex, assetIndex)
+	}
+	return nil
+}
```
guscarreon, line 763: Good catch

guscarreon (commented):
Looks pretty good. A couple of questions:

1) Does it make sense to also modify these? I realize that, in context, it is implied that they refer to openRtb2 but, if in a near future openrtb3 coexists with openrtb2 in our codebase, does it make sense to explicitly name them, even in the comments, to avoid any ambiguity?
```
./exchange/exchange.go:// Return an openrtb seatBid for a bidder
./exchange/gdpr.go:// ExtractGDPR will pull the gdpr flag from an openrtb request
./exchange/gdpr.go:// ExtractConsent will pull the consent string from an openrtb request
./openrtb_ext/user.go:	// For more info, see: https://github.com/digi-trust/dt-cdn/wiki/OpenRTB-extension#openrtb-2x
./adapters/adot/adot.go:		return nil, []error{fmt.Errorf("unable to marshal openrtb request (%v)", err)}
./adapters/pubnative/pubnative.go:// make sure that banner has openrtb 2.3-compatible size information
./adapters/dmx/dmx_test.go:		t.Errorf("openrtb type should be an Array when it's an App")
./adapters/dmx/dmx_test.go:		t.Errorf("openrtb type should be an Array when it's an App")
./adapters/dmx/dmx_test.go:		t.Errorf("openrtb type should be empty")
./adapters/appnexus/appnexus_test.go:// Code below this line tests the legacy, non-openrtb code flow. It can be deleted after we
./adapters/sovrn/sovrn_test.go:// Code below this line tests the legacy, non-openrtb code flow. It can be deleted after we
./adapters/consumable/adtypes.go:/* Turn array of openrtb formats into consumable's code*/
./adapters/conversant/cnvr_legacy_test.go:// Helper to create a test http server that receives and generate openrtb requests and responses
./adapters/ix/ix_test.go:// Tests for the legacy, non-openrtb code.
./adapters/pubmatic/pubmatic_test.go:// Code below this line tests the legacy, non-openrtb code flow. It can be deleted after we
./adapters/pulsepoint/pulsepoint_test.go: * Verify the openrtb request sent to Pulsepoint endpoint.
./static/bidder-params/dmx.json:      "description": "The minimum price acceptable for a bid, this is optional since we do get the one from the original openrtb request"
./router/router.go:		glog.Fatalf("Failed to create the openrtb endpoint handler. %v", err)
```

2) Are we still using OpenRtb 2.5? I assume yes but just double-checking:
```
./adapters/gamma/gamma.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/advangelists/advangelists.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/nanointeractive/nanointeractive.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/lockerdome/lockerdome.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/adpone/adpone.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/verizonmedia/verizonmedia.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/ninthdecimal/ninthdecimal.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/smartrtb/smartrtb.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/brightroll/brightroll.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/somoaudience/somoaudience.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/kidoz/kidoz.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/marsmedia/marsmedia.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/adkernel/adkernel.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/adkernelAdn/adkernelAdn.go:	headers.Add("x-openrtb-version", "2.5")
./adapters/lunamedia/lunamedia.go:	headers.Add("x-openrtb-version", "2.5")
```

SyntaxNode:
> 1. Does it make sense to also modify these? I realize that, in context, it is implied that they refer to openRtb2 but, if in a near future openrtb3 coexists with openrtb2 in our codebase, does it make sense to explicitly name them, even in the comments, to avoid any ambiguity?

Thank you for reviewing our usage of openrtb throughout the code base. I agree with modifying some of them. I don't want to modify any adapter specific error message to avoid subtle breaking changes. I think a lot of these references are meant to distinguish between the legacy custom PBS model versus the standardized OpenRTB model, in which case the change of the mxmCherry package from `openrtb` to `openrtb2` is irrelevant.

> 2. Are we still using OpenRtb 2.5? I assume yes but just double-checking:

Yes. We are. The library we use for OpenRTB 2.5 has restructured, but the version of the spec we target is still the same.

SyntaxNode:
Updated to add non-negative validation to all data structures except for the bid response. Adapters currently aren't able to set negative bid values due to the units.

adapters/audienceNetwork/facebook.go
```diff
@@ -10,19 +10,18 @@ import (
...
-	"github.com/buger/jsonparser"
-	"github.com/mxmCherry/openrtb"
 )
 
-var supportedBannerHeights = map[uint64]bool{
-	50:  true,
-	250: true,
+var supportedBannerHeights = map[int64]struct{}{
+	50:  {},
+	250: {},
```
SyntaxNode, line 24: This should've always been a hashset. Fixed.

adapters/audienceNetwork/facebook.go
```diff
@@ -204,15 +197,14 @@ func (this *FacebookAdapter) modifyImp(out *openrtb.Imp) error {
 			}
 		}
 
-		/* This will get overwritten post-serialization */
-		out.Banner.W = openrtb.Uint64Ptr(0)
+		out.Banner.W = openrtb2.Int64Ptr(-1)
```
SyntaxNode, line 200: This doesn't have to be a hack now. It's ok to use a negative number here. We're past bid request validation.

adapters/audienceNetwork/facebook.go
```diff
@@ -359,6 +337,8 @@ func (this *FacebookAdapter) MakeBids(request *openrtb.BidRequest, adapterReques
 
 	for _, seatbid := range bidResp.SeatBid {
 		for _, bid := range seatbid.Bid {
+			bid := bid
```
mansinahar, line 340: What's the reason for assigning `bid` to `bid` here? This should have no effect, right?

SyntaxNode: It has an affect. This fixes a memory issue with passing a reference to a loop iterator variable outside the loop. Contributors have pointed out the issue in recent PRs. The new adapter docs now include the corrected code and many adapters have been updated, but we should do a review to ensure they're all fixed.

The root issue is passing a reference to the `bid` loop variable outside of the loop. The `bid` variable is re-used with each loop iteration and is overwritten with a shallow copy of the current item. The fix is to either create a new shallow copy or to use the index and access the bid item directly from the slice. This is the former approach. Since Go allows for variable shadowing, this line is creating a new variable `bid` as a shallow copy of the loop iterator variable `bid`.  An alternative is to avoid using `range` here altogether.

guscarreon (commented):
Changes and validations look pretty good. I'm ready to approve once the merge conflicts have been resolved.

adapters/audienceNetwork/facebook.go
SyntaxNode: Rewritten to use index instead of shadow copying the bid loop variable.

mansinahar (approved):

guscarreon (approved):
LGTM

PR #1737, opened 2021-03-05
beachfront: Changes to support real 204

muncha:
An optional parameter has been added to beachfront's banner endpoint to support returning a 204 response instead of an empty array on no content. These changes support that.


adapters/beachfront/beachfronttest/supplemental/banner-and-adm-video-by-explicit-204-response-on-banner.json
```diff
@@ -0,0 +1,147 @@
...
+                }
+              ],
+              "seat": "bfb-io-s1"
+            }
+          ],
+          "bidid": "5d839458f73decdc1572b7f6",
+          "cur": "USD"
+        }
+      }
+    }
```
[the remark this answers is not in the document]
muncha: @mansinahar - This gets into some deeper questions and concerns that I've had for a while and I think I need to get to the bottom of them.

Many of the test cases do not behave like the real requests. Specifically, the "expectedBidResponses" works if I break it up into multiple "bids" objects, each with an array of "bid" objects (see banner-and-adm-video.json), wrapped in an object, but for the same request sent to a running server, I get back a single "seatbid" with an array of "bid" objects, wrapped in an object. I'll send you some gists in a bit.

For the test above, if I send an identical request (substituting "beachfront" for "bidder", of course) to a running server with a bad appid for the banner, therefore getting a 204, I still get back a video response, and it works just fine. However, the only "expectedBidResponses" that I've been able to make work in the test is an empty array, which does not reflect what bid responses I actually expect. If I include anything like what I actually expect, like just the video bid, or like the banner-and-adm-video.json, but with an empty array for the banner bid, I get a runtime error.

So, I suspect that this is happening because the beachfront bidder uses multiple endpoints. I've gotten a better understanding of how the validate.sh tests work through my last couple PRs (I think). I've also dug through that code thinking maybe this should be called an issue with the test system. I'm now thinking that's the wrong  thinking. I think that the prebid-server system assumes a bidder calls one endpoint, and I've hacked around that. **I think that the beachfront bidder should be broken up into 3 separate bidders.** Does that make sense?

**As far as you know - is anyone else using multiple endpoints in a single bidder?**

Also, I think that this is related to the crazy mess that I had with trying to incorporate multiple impressions for the AdM video requests. That was working fine when I did mixed requests to a running server, but failed on tests with runtime errors.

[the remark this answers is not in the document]
muncha: Here are some requests and responses from running the test json to a running instance:

banner and adm gist:
https://gist.github.com/muncha/f18360946e6ddf8d6a3205adf8dc8913

banner(204) and adm gist:
https://gist.github.com/muncha/74b4bd025fa07ce12b1e7e344db5aba5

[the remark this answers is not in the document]
muncha: Done. 

The change that I made in adapters/adapterstest/test_json.go got this test to work, and makes sense to me. Without that the test throws a runtime [error](https://gist.githubusercontent.com/muncha/74b4bd025fa07ce12b1e7e344db5aba5/raw/aa1e4967ff33b8c6ff01c4afb970cd38726890c7/runtime_error.txt).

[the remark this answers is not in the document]
muncha: As I mentioned in some now deleted comments where I babbled on quite a bit, sending the same request to a running instance does not raise a runtime error and works just fine.

https://gist.github.com/muncha/74b4bd025fa07ce12b1e7e344db5aba5

adapters/adapterstest/test_json.go
```diff
@@ -123,7 +123,9 @@ func runSpec(t *testing.T, filename string, spec *testSpec, bidder adapters.Bidd
 	diffErrorLists(t, fmt.Sprintf("%s: MakeBids", filename), bidsErrs, spec.MakeBidsErrors)
 
 	for i := 0; i < len(spec.BidResponses); i++ {
-		diffBidLists(t, filename, bidResponses[i].Bids, spec.BidResponses[i].Bids)
+		if bidResponses[i] != nil {
```
mansinahar, line 126: I think the issue with this check is that if `bidResponses[i]` is `nil` but `spec.BidResponses[i]` is not `nil` then we are just skipping over that BidResponse and passing the test even though the test should be marked failed. How about changing the `diffBidLists` method to accept `[]*adapters.BidResponse` as an argument rather than `[]*adapters.TypedBid` and adding an assertion in that method to make sure that they are both either nil or not nil. What do you think?

I realize that this a gap in our test framework so If this is not too much for you to do and if you can then go ahead but if it is then lemme know and I can. put up a quick PR myself.

muncha: Yeah, that makes sense. I'm in the middle of other work right now, but should be able to jump on it shortly.

mansinahar: Perfect! Yeah, take your time. Thank you @muncha

mansinahar:
Hey @muncha! Just a friendly follow-up to see if you had a chance to address the test issue?

adapters/adapterstest/test_json.go
muncha: @mansinahar - I am looking at this, and I've tried a couple things along the lines of what you suggest above. I'm not quite there, but I think I get how it works. I should be updating the PR in an hour or two.

muncha: But looking at this, I think maybe I should just return an empty adapters.NewBidderResponseWithBidsCapacity() instead of nil on 204.
```

// diffOrtbBids compares the actual Bid made by the adapter to the expectation from the JSON file.
func diffOrtbBids(t *testing.T, description string, actual *openrtb.Bid, expected json.RawMessage) {
	if actual == nil {
		t.Errorf("Bidders cannot return nil Bids. %s was nil.", description)
		return
	}
```

muncha:
@mansinahar - Note that intentionally failed commit 30d9eb8 above demonstrates the sanity check failing.

bsardo (approved):

mansinahar (approved):

PR #1743, opened 2021-03-08
validateNativeContextTypes function test cases

guscarreon:
This PR adds test case coverage to function `validateNativeContextTypes` found in `endpoints/openrtb2/auction.go`


mansinahar:
Looks like you have a `.swp` file here by mistake. Can you please remove that?

bsardo (approved):
LGTM

endpoints/openrtb2/sample-requests/invalid-native/context-type-social-incompatible-subtype.json
```diff
@@ -0,0 +1,27 @@
...
+        "ext": {
+          "appnexus": {
+            "placementId": 12883451
+          }
+        }
+      }
+    ]
+  },
+  "expectedReturnCode": 400,
+  "expectedErrorMessage": "Invalid request"
```
mansinahar, line 24: I see the error messages returned in the `validateNativeContextTypes` method to be different that what is expected in these tests. Is the error message getting overridden somewhere to this?? If so, they shouldn't be as the error messages in the `validateNativeContextTypes` are more informative of the problem

bsardo: I noticed this as well but I played with one of the tests a bit and saw that the actual error message returned by `validateNativeContextTypes` has `Invalid Request: ` prepended to it and the test framework is just checking that the error message begins with whatever `expectedErrorMessage` is rather than doing a full string compare.

guscarreon: Exactly

mansinahar (approved):

Issue #1140, opened 2019-12-11
Feature: pub-supplied currency rates


LukaszSawicki:
Hi,

We have a problem with currency exchanging on prebid-server biddings. In our project we have combined server and client side requests and we have also included `Currency module` (https://github.com/prebid/Prebid.js/blob/master/modules/currency.js) in our Prebid.js build.

On client-side requests it works perfectly, but we found that Prebid Server is not including bidder currency param in any response, and because of that  `PrebidServerBidAdapter module` (https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js) sets currency always to `USD`, no matter what currency was given in response by bidder.

Here's an example on AdForm bid request. From https://prebid.adnxs.com/pbs/v1/openrtb2/auction we got that response:
```
{
  'id': '92871ddc-4e43-45bd-a2c8-f44870f091fc',
  'seatbid': [
    {
      'bid': [
        {
          'id': '9_112572',
          'impid': '9_112572',
          'price': 1.3734983170865096,
[...]

PR #1753, opened 2021-03-15
Request Provided Currency Rates

guscarreon:
This pull request implements feature described in: https://github.com/prebid/prebid-server/issues/1140

Until now, Prebid Server uses currency conversion rates fetched periodically from a service configured in package `currency`. This pull request allows a given bid request to use its own bid-request-level defined currency conversion rates. These bid-level currency rates come inside a newly created `bidRequest.ext.prebid.currency` field that looks like this:
```
"ext": {
  "prebid": {
    "currency": {
      "rates": {
        "USD": { "UAH": 24.47, "ETB": 32.04 },
        "GBP": { "USD": 1.21, "EUR": 1.1 }
      },
      "usepbsrates": false
    }
  }
}
```
`request.ext.prebid.currency` contains two fields: `rates` that is a map of currency conversion rates, and `usePbsRates` that determines whether or not the custom rates defined in `rates` complement those of Prebid Server's currency rate service (default behavior) or substitutes them entirely.


currency/rates.go
```diff
@@ -63,10 +63,10 @@ func (r *Rates) GetRate(from string, to string) (float64, error) {
 	if r.Conversions != nil {
 		if conversion, present := r.Conversions[fromUnit.String()][toUnit.String()]; present {
 			// In case we have an entry FROM -> TO
-			return conversion, err
+			return conversion, nil
```
guscarreon, line 66: it is easier to read if we explicitly return `nil`

endpoints/openrtb2/sample-requests/custom-currency/errors/non-usd-cur-custom-rates-usepbs-false.json
```diff
@@ -0,0 +1,53 @@
+{
+  "description": "BidRequest defines its own custom rates in the root.ext and sets the usepbsrates to false, not allowing for PBS to use its default converter and defaulting to a price of 0",
+  "config": {
+    "currencyRates":{
+      "USD": {
+        "MXN": 5.09
+      }
+    },
+    "mockBidder": {
```
guscarreon, line 9: This pull request also implements the ability to define a mock bid within the JSON test

endpoints/openrtb2/sample-requests/custom-currency/errors/non-usd-cur-custom-rates-usepbs-false.json
```diff
@@ -0,0 +1,53 @@
+{
+  "description": "BidRequest defines its own custom rates in the root.ext and sets the usepbsrates to false, not allowing for PBS to use its default converter and defaulting to a price of 0",
+  "config": {
+    "currencyRates":{
```
guscarreon, line 4: mock Prebid Server currency rate service values added to test configuration

endpoints/openrtb2/sample-requests/account-required/no-account/not-required-no-acct.json
```diff
@@ -73,7 +73,7 @@
             {
               "id": "appnexus-bid",
               "impid": "",
-              "price": 0
+              "price": 1
```
guscarreon, line 76: Newly implemented mock bidder makes a bid of $1.00 dollar

endpoints/openrtb2/sample-requests/custom-currency/valid/custom-rates-override-usepbsrates-default.json
```diff
@@ -0,0 +1,68 @@
+{
+  "description": "Despite being more expensive, custom conversion rate overrides that of the currency conversion service. usepbsrates defaults to true",
```
guscarreon, line 2: This test case describes the functionality added to this PR in a nutshell.

exchange/exchange.go
```diff
@@ -865,6 +865,59 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
 	return
 }
 
+// getAuctionCurrencyRates returns the constant currency conversion rates that will be used to process this
+// auction. Parameter customRates contains, if any, the bidRequest custom currency conversion rates
+// defined in bidRequest.Ext.  If usePbsRates flag is set to false, all previous Conversions values are discarded
+// and this function merely substitutes Conversions with customRates. But if usePbsRates was ommitted or set to
+// true, customRates' values update or add to the Conversions map
+func (e *exchange) getAuctionCurrencyRates(customRates *openrtb_ext.ExtRequestCurrency) currency.Conversions {
+	pbsConversions := e.currencyConverter.Rates()
```
mansinahar, line 874: Any reason why we are doing the default conversions first and then discarding them if custom conversion rates are present? I think it would make more sense and be more efficient to apply the custom conversion rates first if they exist and fallback to default only if custom conversion rates don't exist or don't apply

openrtb_ext/request.go
```diff
@@ -35,6 +35,13 @@ type ExtRequestPrebid struct {
 	// passing of personally identifiable information doesn't constitute a sale per CCPA law.
 	// The array may contain a single sstar ('*') entry to represent all bidders.
 	NoSale []string `json:"nosale,omitempty"`
+
+	BidReqConversions *ExtRequestCurrency `json:"currency,omitempty"`
+}
+
+type ExtRequestCurrency struct {
+	ConversionRates map[string]map[string]float64 `json:"rates"`
+	UsePBSRates     *bool                         `json:"usepbsrates"`
```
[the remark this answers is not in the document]
guscarreon: If we don't define it as a pointer, it would always default to `false` and the requirements in #1140 estate that we should default to true. For those `bidRequest.ext` fields that don't explicitly set `"usepbsrates"` to `true`, we still want that field to default to true so we can also use PBS currency rates. In other words defining it as a pointer gives us the ability to tell whether it:
- Was explicitly listed in `bidRequest.ext` and it came with a `false` value
- Was explicitly listed in `bidRequest.ext` and it came with a `true` value
- Was not listed in `bidRequest.ext` and it we don't know what value it would've had, but we want it to default to `true`.
 
Whereas if a regular `bool` JSON field:
- Was explicitly listed in `bidRequest.ext` and it came with a `false` value
- Was explicitly listed in `bidRequest.ext` and it came with a `true` value
- Was not listed listed in `bidRequest.ext`, defaults to `false` when unmarshaled.

endpoints/openrtb2/sample-requests/custom-currency/errors/custom-rates-currency-missing-usepbs-false.json
```diff
@@ -0,0 +1,54 @@
+{
+  "description": "False usepbsrates forces BidRequest use custom currency rates but bidRequest currency is not found in bidRequest.ext.prebid.currency.rates field",
+  "description": "BidRequest defines its own custom rates in the root.ext and sets the usepbsrates to false, not allowing for PBS to use its default converter and defaulting to a price of 0",
```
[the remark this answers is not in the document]
guscarreon: It is, sorry. Removed

endpoints/openrtb2/sample-requests/custom-currency/errors/custom-rates-empty-usepbs-false.json
```diff
@@ -0,0 +1,49 @@
+{
+  "description": "usepbsrates set to false forces BidRequest to use custom currency rates but empty bidRequest.ext.prebid.currency.rates field is found",
```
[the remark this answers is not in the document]
guscarreon: yep, reworded.

exchange/exchange.go
guscarreon: Good point. Also refactored the code a bit and provided a couple more test cases.

endpoints/openrtb2/sample-requests/custom-currency/valid/custom-rates-override-usepbsrates-default.json
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+  },
+  "expectedBidResponse": {
+      "id":"some-request-id",
+      "bidid":"test bid id",
+      "nbr":0,
+      "seatbid": [
+        {
+          "bid": [
```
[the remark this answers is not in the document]
guscarreon: I did consider adding it to the `mockExchanges`'s implementation of `HoldAuction` but I changed my mind because at the end I wasn't super happy with these tests overall because I had to write a bit more features into the `mockExchange`  we use in `endpoints/openrtb2/auction_test.go`. I can go ahead and add the currency here but, what do you think? Do you agree with these `custom-currency/valid` tests? Or should I discard them along with the `mockExchange`'s added complexity. Strictly speaking, only the `custom-currency/errors` test cases are covering the error scenarios of the recently added `validateCustomRates()` function found in `endpoints/openrtb2/auction.go`. 

Let me know your take. I was enthusiastic about the `custom-currency/valid` folder at first but not so much now.

endpoints/openrtb2/auction.go
```diff
@@ -434,6 +440,53 @@ func validateSChains(req *openrtb_ext.ExtRequest) error {
 	return err
 }
 
+// validateCustomRates discards invalid 3-digit currency codes found in the bidRequest.ext.prebid.currency
+// field. If all of them were discarded and bidRequest.ext.prebid.currency was set to false, it means
+// we have no currencies to work with and this function returns a fatal BadInput error
+func validateCustomRates(bidReqCurrencyRates *openrtb_ext.ExtRequestCurrency) error {
+
+	if bidReqCurrencyRates == nil {
```
SyntaxNode, line 448: Nitpick: Please remove leading line break.

endpoints/openrtb2/auction_test.go
```diff
@@ -2170,7 +2350,63 @@ func (e *nobidExchange) HoldAuction(ctx context.Context, r exchange.AuctionReque
...
+		for toCurrency, rate := range rates {
+			// If it was found in the custom conversions, don't modify because custom rates take precedence
+			if _, customRateFound := auctionRates[fromCurrency][toCurrency]; !customRateFound {
+				auctionRates[fromCurrency][toCurrency] = rate
+			}
+		}
+	}
+
+	return currency.NewRates(time.Now(), auctionRates)
 }
```
SyntaxNode, line 2410: This is a lot of code in mock. Why can't we use the real implementation?

exchange/exchange.go
```diff
@@ -903,6 +903,66 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
 }
 
+// getAuctionCurrencyRates returns the constant currency conversion rates that will be used to process this
+// auction. Parameter customRates contains, if any, the bidRequest custom currency conversion rates
+// defined in bidRequest.Ext.  If usePbsRates flag is set to false, all previous Conversions values are discarded
+// and this function merely substitutes Conversions with customRates. But if usePbsRates was ommitted or set to
+// true, customRates' values update or add to the Conversions map
+func (e *exchange) getAuctionCurrencyRates(customRates *openrtb_ext.ExtRequestCurrency) currency.Conversions {
+	// No bidRequest.ext-defined, use PBS rates
+	if customRates == nil || len(customRates.ConversionRates) == 0 {
```
SyntaxNode, line 913: There's no need for the nil check. Can just be:

```
if len(customRates.ConversionRates) == 0 {
```

exchange/exchange.go
```diff
@@ -903,6 +903,66 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
+		return e.currencyConverter.Rates()
+	}
+
+	var usePbsRates bool = true
```
SyntaxNode, line 917: Nitpick: I like how you wrote this in other parts of the code:
```
usePbsRates := customRates.UsePBSRates != nil && *customRates.UsePBSRates
```

exchange/exchange.go
```diff
@@ -903,6 +903,66 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
+	if customRates.UsePBSRates != nil {
+		usePbsRates = *customRates.UsePBSRates
+	}
+
+	// when usePbsRates is false, only use bidRequest.ext-defined rates
+	if !usePbsRates {
+		return currency.NewRates(time.Now(), customRates.ConversionRates)
```
SyntaxNode, line 924: Why is the time needed for constant rates from the request?

endpoints/openrtb2/auction.go
```diff
@@ -434,6 +440,53 @@ func validateSChains(req *openrtb_ext.ExtRequest) error {
...
+	}
+
+	var badCurrencyCodes []string
+
+	for fromCurrency, rates := range bidReqCurrencyRates.ConversionRates {
+
+		// Check if fromCurrency is a valid 3-letter currency code
+		_, err := currency.ParseISO(fromCurrency)
+		if err != nil {
+			badCurrencyCodes = append(badCurrencyCodes, fromCurrency)
```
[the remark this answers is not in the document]
guscarreon: This PR is throwing a fatal error only if all of the custom currency codes were found to be wrong. The reason this approach made more sense to me is that we ignore malformed or inaccurate 3-digit currency codes inside `requestBid()` anyways.
```
197                 var conversionRate float64
198                 var err error
199                 for _, bidReqCur := range request.Cur {
200                     if conversionRate, err = conversions.GetRate(bidResponse.Currency, bidReqCur); err == nil {
201                         seatBid.currency = bidReqCur
202                         break
203                     }
204                 }
exchange/bidder.go
```
But, I'm open to ideas here.

endpoints/openrtb2/auction.go
```diff
@@ -434,6 +440,53 @@ func validateSChains(req *openrtb_ext.ExtRequest) error {
...
+		if len(rates) == 0 {
+			delete(bidReqCurrencyRates.ConversionRates, fromCurrency)
+		}
+	}
+
+	// If we are retricted to only use custom currency rates but the currency rate map is empty, return error
+	customRatesOnly := bidReqCurrencyRates.UsePBSRates != nil && !(*bidReqCurrencyRates.UsePBSRates)
+	if customRatesOnly && len(bidReqCurrencyRates.ConversionRates) == 0 {
+		if len(badCurrencyCodes) > 0 {
+			return &errortypes.BadInput{Message: fmt.Sprintf("Required custom currency rates are all invalid. %s", strings.Join(badCurrencyCodes, ","))}
```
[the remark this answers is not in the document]
guscarreon: But to reach this branch `bidReqCurrencyRates.ConversionRates` must be empty which means either it was empty from the beginning, which leads to returning the error in line 484, or it got emptied after finding all of the 3-letter currency codes to be wrong that resolves in returning error in 482. Do you agree with this approach? We could still reword the error message or even refactor this part if needed.

endpoints/openrtb2/auction.go
```diff
@@ -37,6 +38,7 @@ import (
 	"github.com/prebid/prebid-server/util/httputil"
 	"github.com/prebid/prebid-server/util/iputil"
 	"golang.org/x/net/publicsuffix"
+	"golang.org/x/text/currency"
```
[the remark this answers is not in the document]
guscarreon: Upgraded to v0.3.6

endpoints/openrtb2/auction_test.go
```diff
@@ -1183,6 +1197,172 @@ func TestContentType(t *testing.T) {
...
+			desc: "empty custom currency rates but UsePBSRates is set to false, return error",
+			inBidReqCurrencies: &openrtb_ext.ExtRequestCurrency{
+				ConversionRates: map[string]map[string]float64{},
+				UsePBSRates:     &boolFalse,
+			},
+			outFilteredCurrencies: map[string]map[string]float64{},
+			outCurrencyError:      &errortypes.BadInput{Message: "Required custom currency rates field is empty"},
+		},
+		{
+			desc: "bidExt fromCurrency is fake, expect no errors because UsePBSRates is nil and defaults to true",
```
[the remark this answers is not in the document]
guscarreon: Good point. Changed

endpoints/openrtb2/auction.go
guscarreon: Done

endpoints/openrtb2/auction_test.go
guscarreon: Totally agree. I'd love to go straight into using the `exchange/exchange.go`'s implementation of `HoldAuction()` and have end to end tests with a mock bidder that can be json defined too. Should we code those changes in this PR? Should we wait and code them separately in a future PR instead?

exchange/exchange.go
guscarreon: I believe we need that extra check because `customRates` is a pointer

exchange/exchange.go
guscarreon: Does look a lot better. Done

exchange/exchange.go
guscarreon: Although these custom rates are constant values that will not change, in terms of the logic inside our `currency` package, they are not considered `ConstantRates` or at least they are not named that. This return value is meant to return the `currency.Conversions` implementation found in `currency/rates.go`. While "constant" rates are implemented in `currency/constant_rates.go`. They widely differ and in my opinion `ConstantRates` should be renamed to something else. Probably `SameCurrencyRates` or `UnitaryRates` or something. Kept that part out of the scope of this PR but I could always modify it if you disagree.

Because line 924 returns all-custom currency conversions object that doesn't go into a `RateConverter` dating them with `time.Now()` as the time they were `fetched` seemed more accurate but you raise a good point. Given that we don't actually do anything with that time stamp from here on I'll change to a less expensive `time.Time{}` instead.

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
SyntaxNode: Let's seek out @hhhjort @mansinahar @bsardo for additional thoughts here.

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
SyntaxNode: I don't think that's necessarily true. As the code is currently, a `ConversionRates` entry will be removed if the base currency is wrong even if the mapped currencies are correct. For example:
```
ConversionRates: map[string]map[string]float64{
    "FOO": {
        "GBP": 1.2,
        "MXN": 0.05,
        "JPY": 0.95,
    },
}
```
If I'm reading the code correctly, this will cause the error message to be reached when `UsePBSRates` is true. While the base currency of FOO is invalid, the mapped currencies of GBP, MXN, and JPY are valid.

endpoints/openrtb2/auction_test.go
SyntaxNode: Do you think that would be a lot of work? If so, perhaps its best to do that in a separate PR.

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
SyntaxNode: I don't think that's necessarily true. As the code is currently, a `ConversionRates` entry will be removed if the base currency is wrong even if the mapped currencies are correct. For example:
```
ConversionRates: map[string]map[string]float64{
    "FOO": {
        "GBP": 1.2,
        "MXN": 0.05,
        "JPY": 0.95,
    },
}
```
If I'm reading the code correctly, this will cause the error message to be reached when `UsePBSRates` is true. While the base currency of FOO is invalid, the mapped currencies of GBP, MXN, and JPY are valid.

exchange/exchange.go
SyntaxNode: Yes. You're right.

exchange/exchange.go
SyntaxNode: Ideally, I think we should remove the reference to time entirely.

It seems this field only exists to fully parse the auto generated PBS currency file. We don't actually do anything with that timestamp at all. Perhaps we can remove it entirely.

Thoughts? @guscarreon @bsardo

exchange/exchange.go
```diff
@@ -936,6 +936,88 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
+	// No bidRequest.ext custom rates defined, use PBS rates as usual
+	if customRates == nil || len(customRates.ConversionRates) == 0 {
+		return e.currencyConverter.Rates()
+	}
+
+	// bidRequest.ext defines custom rates of its own, figure out whether to use a mix of
+	// Pbs rates and custom rates or custom rates only
+	if customRates.UsePBSRates != nil && !(*customRates.UsePBSRates) {
+		// Use custom rates only
+		includeInverseRates(customRates.ConversionRates)
```
mansinahar, line 954: In case when only custom rates are in play, I believe we don't need to store the inverse rate conversions as well. The `GetRate` function in `rates.go` should take care of it when it does the currency conversion between two currencies

exchange/exchange.go
```diff
@@ -936,6 +936,88 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
+
+		return currency.NewRates(time.Time{}, customRates.ConversionRates)
+	} else {
+		// Use both custom and PBS rates
+		pbsConversions := e.currencyConverter.Rates()
+		serverRates := pbsConversions.GetRates()
+		auctionRates := make(map[string]map[string]float64, len(*serverRates))
+
+		copyCustomRates(auctionRates, customRates.ConversionRates)
```
mansinahar, line 963: Wouldn't it be easier here if we just looped through all the custom rates and override the values in PBS rates (or a copy of PBS rates) that need to be overridden? Something like:
```
// Go through all the custom rates
for fromCur, rates := range customRates {
	for toCur, val := range rates {
		// If the rate conversion exists in PBS rates then override the value
		if _, ok := pbsRates[fromCur]; ok {
			pbsRates[fromCur][toCur] = val
		} else if _, ok := pbsRates[toCur]; ok {
			pbsRates[toCur][fromCur] = val
		} else {
			// If it doesn't exist then, add the conversion
			pbsRates[fromCur][toCur] = val
		}
	}
}
```

(This code snippet is just to give you an idea of what I was thinking. May or may not compile as is)

endpoints/openrtb2/auction.go
```diff
@@ -434,6 +440,53 @@ func validateSChains(req *openrtb_ext.ExtRequest) error {
...
+
+	var badCurrencyCodes []string
+
+	for fromCurrency, rates := range bidReqCurrencyRates.ConversionRates {
+
+		// Check if fromCurrency is a valid 3-letter currency code
+		_, err := currency.ParseISO(fromCurrency)
+		if err != nil {
+			badCurrencyCodes = append(badCurrencyCodes, fromCurrency)
+			delete(bidReqCurrencyRates.ConversionRates, fromCurrency)
```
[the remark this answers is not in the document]
guscarreon: Good catch. Added a `continue` to skip to next entry

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
guscarreon: Modified this function to throw fatal bad input error at the first invalid or malformed currency code found in the map

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
guscarreon: This part of the code has been removed because we are now immediately returning a bad input error when the first non-valid code is found.

endpoints/openrtb2/auction_test.go
guscarreon: Do you agree if we try this in a separate PR?

exchange/exchange.go
guscarreon: True. Good catch, corrected.

exchange/exchange.go
guscarreon: I considered that option at first but discarded in favor of the new code found inside the newly created `currency/rate_engines.go`. Let me know your take on this.

exchange/exchange.go
guscarreon: I didn't remove it in today's PR update but I can do so if you want to. @bsardo @SyntaxNode please let me know.

exchange/exchange.go
```diff
@@ -936,6 +936,40 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
 	return
 }
 
+// getAuctionCurrencyRates returns the a currency conversions implementation that will be used in the rest of this
```
SyntaxNode, line 939: IMHO this comment isn't necessary.

exchange/exchange.go
```diff
@@ -936,6 +936,40 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
+// auction. Parameter customRates contains, if any, the bidRequest custom currency conversion rates
+// defined in bidRequest.Ext. If usePbsRates flag is set to false, the currency conversions provided by Prebid Server's
+// currency conversion service get discarded in favor of the bidRequest.ext.currency-defined conversion rates.
+func (e *exchange) getAuctionCurrencyRates(customRates *openrtb_ext.ExtRequestCurrency) currency.Conversions {
```
SyntaxNode, line 943: Consider renaming `customRates` to `requestRates` to make the source of the information clearer. Thoughts?

currency/rate_engines.go
```diff
@@ -0,0 +1,42 @@
...
+		return rate, err
+	}
+
+	// because the custom rates' GetRate() call returned an error other than "conversion
+	// rate not found", there's nothing wrong with the 3 letter currency code so let's
+	// try the PBS rates instead
+	return re.conversions[PBSRates].GetRate(from, to)
+}
+
+// No need to call GetRates on RateEngines
```
SyntaxNode, line 39: Why?

currency/rate_engines.go
```diff
@@ -0,0 +1,42 @@
...
+
+import "strings"
+
+const (
+	CustomRates = iota
+	PBSRates
+)
+
+type RateEngines struct {
+	conversions []Conversions
```
SyntaxNode, line 11: Consider having `customRates` and `serverRates` fields instead of using the slice. I don't see why `re.conversions[CustomRates]` wouldn't be clearer as `re.customRates`.

currency/rate_engines.go
```diff
@@ -0,0 +1,42 @@
+package currency
+
+import "strings"
+
+const (
+	CustomRates = iota
+	PBSRates
+)
+
+type RateEngines struct {
```
SyntaxNode, line 10: Love the concept, not the name. The term 'Engine' is meaningless in this context. Consider names along the lines of `AggregateConversions` or `requestRates`.

exchange/exchange.go
guscarreon: I agree. Done

currency/rate_engines.go
guscarreon: This function is only ever used (outside of test cases) to put up the `ConverterInfo` the `/currency/rates` endpoint returns. At the `exchange` level, to return a map with all of the currencies does not seem useful. A way to implement it, would be to copy the PBS rates to a new map and overwrite the custom rates on top (something similar to what a previous version of the `getAuctionCurrencyRates(requestRates *openrtb_ext.ExtRequestCurrency)` did). Under the new `RatesEngines` approach I see no need. I don't think it would ever be used. Thoughts?

I can still implement it if you think we need to:
```
// Implementing GetRates on RateEngines
func (re *RateEngines) GetRates() *map[string]map[string]float64 {
	serverRatesMap := re.conversions[PBSRates].GetRates()
	customRatesMap := re.conversions[CustomRates].GetRates()

	returnRatesMap := make(map[string]map[string]float64, len(*serverRatesMap))

	//Copy server rates into returnRatesMap
	for fromCurrency, rates := range (*serverRatesMap) {
		returnRatesMap[fromCurrency] = make(map[string]float64, len(rates))
		for toCurrency, rate := range rates {
			returnRatesMap[fromCurrency][toCurrency] = rate
		}
	}

	//Overwrite with customRates
	for fromCurrency, rates := range (*customRatesMap) {
		if _, defined := returnRatesMap[fromCurrency]; !defined {
			returnRatesMap[fromCurrency] = make(map[string]float64, len(rates))
		}

		for toCurrency, rate := range rates {
			returnRatesMap[fromCurrency][toCurrency] = rate

			// Write entries to inverse of conversion rates if needed
			if _, inverseIncluded := customRates.ConversionRates[toCurrency][fromCurrency]; !inverseIncluded {
				if _, defined := returnRatesMap[toCurrency]; !defined {
					returnRatesMap[toCurrency] = map[string]float64{fromCurrency: 1.00 / rate}
				} else {
					returnRatesMap[toCurrency][fromCurrency] = 1.00 / rate
				}
			}
		}
	}

	//Return pointer
	return *returnRatesMap
}
```

currency/rate_engines.go
guscarreon: Good point. Done

currency/rate_engines.go
guscarreon: Do you like `ConversionsGroup`? How about `GroupedConversions` or `GroupedCurrencyConversions`? I'm not great at naming.

SyntaxNode: `GroupedCurrencyConversions` too wordy IMHO. If you don't like `AggregateConversions`, then `GroupedConversions` is similar or perhaps also `RequestConversions`?

currency/rate_engines.go
SyntaxNode: > This function is only ever used (outside of test cases) to put up the `ConverterInfo` the `/currency/rates` endpoint returns.

Ah, ok. I understand. Seems fine for now. I think we should revisit this design in the future such that the `GetRate` for the bidder loop doesn't have a dependency on the `GetRates` for the currency conversion endpoint.

currency/aggregate_conversions_test.go
```diff
@@ -0,0 +1,79 @@
...
+			"EUR": 0.80,
+		},
+	})
+
+	pbsRates := NewRates(time.Now(), map[string]map[string]float64{
+		"USD": {
+			"GBP": 0.50,
+			"MXN": 10.31,
+		},
+	})
```
mansinahar, line 25: I'd also suggest having a test case where a reverse currency conversion is present as an override in the custom. currency rates. So for example, here maybe you can add `MXN` to `USD` in custom rates and make sure that works as expected as well?

exchange/exchange.go
```diff
@@ -936,6 +936,36 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
+		return e.currencyConverter.Rates()
+	}
+
+	// If bidRequest.ext.currency.usepbsrates is nil, we understand its value as true. It will be false
+	// only if it's explicitly set to false
+	usePbsRates := requestRates.UsePBSRates == nil || *requestRates.UsePBSRates
+
+	if !usePbsRates {
+		// If PBS Rates cannot be used, use either constant rates or custom rates only
+		if len(requestRates.ConversionRates) == 0 {
```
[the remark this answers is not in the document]
guscarreon: True. Good catch. Maybe this logic, or at least a truth table and some other notes I had, is a leftover from a couple of commits ago when or error handling in `validateCustomRates(bidReqCurrencyRates *openrtb_ext.ExtRequestCurrency)` was different. Refactored.

currency/aggregate_conversions_test.go
guscarreon: I believe we already have that covered in the test cases. Did you have any othre scenario un mind? 
```
36     testGroups := []struct {
37         expectError bool
38         testCases   []aTest
39     }{
40         {
41             expectError: false,
42             testCases: []aTest{
43                 {"Found in both, return custom rate", "USD", "GBP", 0.77208},
44 /*OVERRIDE-->*/ {"Found in both, return inverse custom rate", "GBP", "USD", 1 / 0.77208},
45                 {"Found in custom rates only", "USD", "EUR", 0.80},
46                 {"Found in PBS rates only", "USD", "MXN", 10.31},
47                 {"Found in PBS rates only, return inverse", "MXN", "USD", 1 / 10.31},
48                 {"Same currency, return unitary rate", "USD", "USD", 1},
49             },
50         },
currency/aggregate_conversions_test.go
```

currency/aggregate_conversions_test.go
```diff
@@ -0,0 +1,79 @@
...
+	rateEngines := NewAggregateConversions(customRates, pbsRates)
```
SyntaxNode, line 26: Nitpick: Please update `rateEngines` variable name.

exchange/exchange.go
```diff
@@ -936,6 +936,34 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
+		return currency.NewRates(time.Time{}, requestRates.ConversionRates)
+	}
+
+	// Both PBS and custom rates can be used, check if ConversionRates is not empty
+	if len(requestRates.ConversionRates) == 0 {
+		// Custom rates map is empty, use PBS rates only
+		return e.currencyConverter.Rates()
+	}
+
+	// Return a RateEngines object that includes both custom and PBS currency rates but will
```
SyntaxNode, line 962: Nitpick: Please update `RateEngines` name usage in comment to `AggregateConversions`.

endpoints/openrtb2/sample-requests/custom-currency/valid/non-usd-bid.json
```diff
@@ -0,0 +1,66 @@
...
+      "bidid":"test bid id",
+      "nbr":0,
+      "cur": "USD",
+      "seatbid": [
+        {
+          "bid": [
+            {
+              "id": "appnexus-bid",
+              "impid": "",
+              "price": 2
```
SyntaxNode, line 58: Where does the price of 2 come from? Is it hardcoded in the test framework?

currency/aggregate_conversions.go
```diff
@@ -0,0 +1,38 @@
+package currency
+
+import "strings"
+
+type AggregateConversions struct {
+	serverRates, customRates Conversions
+}
+
+// NewRates creates a new Rates object holding currencies rates
```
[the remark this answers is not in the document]
guscarreon: Done

currency/aggregate_conversions.go
```diff
@@ -0,0 +1,38 @@
...
+func NewAggregateConversions(customRates, pbsRates Conversions) *AggregateConversions {
```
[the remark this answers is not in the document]
guscarreon: Given that `AggregateConversions` prioritizes `customRates` over `serverRates` may swap the `AggregateConversions` fields instead?

endpoints/openrtb2/auction_test.go
```diff
@@ -248,6 +259,7 @@ func assertBidResponseEqual(t *testing.T, testFile string, expectedBidResponse o
 	assert.Equalf(t, expectedBidResponse.ID, actualBidResponse.ID, "BidResponse.ID doesn't match expected. Test: %s\n", testFile)
 	assert.Equalf(t, expectedBidResponse.BidID, actualBidResponse.BidID, "BidResponse.BidID doesn't match expected. Test: %s\n", testFile)
 	assert.Equalf(t, expectedBidResponse.NBR, actualBidResponse.NBR, "BidResponse.NBR doesn't match expected. Test: %s\n", testFile)
+	assert.Equalf(t, expectedBidResponse.Cur, actualBidResponse.Cur, "BidResponse.Cur doesn't match expected. Test: %s\n", testFile)
```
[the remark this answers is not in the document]
guscarreon: Has to do with this PR. The currency assertion that we've been wanting to add and hasn't been added yet is on the adapter json test front.

endpoints/openrtb2/auction_test.go
```diff
@@ -2590,7 +2731,63 @@ func (e *nobidExchange) HoldAuction(ctx context.Context, r exchange.AuctionReque
...
+	}
+}
+
+// getAuctionCurrencyRates copies the logic of the exchange package for testing purposes
+func (e *mockBidExchange) getAuctionCurrencyRates(customRates *openrtb_ext.ExtRequestCurrency) currency.Conversions {
+	if customRates == nil || len(customRates.ConversionRates) == 0 {
+		return currency.NewRates(time.Now(), e.pbsRates)
+	}
+
+	var usePbsRates bool = true
```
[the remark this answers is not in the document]
guscarreon: Changed

endpoints/openrtb2/sample-requests/custom-currency/errors/custom-rates-currency-missing-usepbs-false.json
```diff
@@ -0,0 +1,53 @@
+{
+  "description": "BidRequest defines its own custom rates in the root.ext and sets the usepbsrates to false not allowing for PBS to use its default converter. We expect to default to a price of 0 because bid request currency is not found in custom currency map.",
```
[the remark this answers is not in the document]
guscarreon: JSON doesn't allow breaking lines for readability so I shortened the description instead

endpoints/openrtb2/sample-requests/custom-currency/valid/non-usd-bid.json
guscarreon: The expected price of $2.00 is the converted amount of $20.00 MXN to US dollars. The `mockBidder` placed a $20 mexican pesos bid and the custom conversion rate of $10.00 MXN per USD takes precedence over PBS' currency rate or $8.00 MXN per USD.

currency/aggregate_conversions.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+}
+
+// GetRate returns the conversion rate between two currencies prioritizing
+// the customRates currency rate over that of the PBS currency rate service
+// returns an error if both Conversions objects return error.
+func (re *AggregateConversions) GetRate(from string, to string) (float64, error) {
+
+	rate, err := re.customRates.GetRate(from, to)
+	if err == nil || !strings.HasPrefix(err.Error(), `Currency conversion rate not found`) {
```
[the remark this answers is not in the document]
guscarreon: done

endpoints/openrtb2/auction_test.go
```diff
@@ -2590,7 +2731,63 @@ func (e *nobidExchange) HoldAuction(ctx context.Context, r exchange.AuctionReque
...
+
+		for toCurrency, rate := range rates {
+			// If it was found in the custom conversions, don't modify because custom rates take precedence
+			if _, customRateFound := auctionRates[fromCurrency][toCurrency]; !customRateFound {
+				auctionRates[fromCurrency][toCurrency] = rate
+			}
+		}
+	}
+
+	return currency.NewRates(time.Now(), auctionRates)
```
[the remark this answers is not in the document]
guscarreon: Good point. Done

currency/aggregate_conversions.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// GetRate returns the conversion rate between two currencies prioritizing
+// the customRates currency rate over that of the PBS currency rate service
+// returns an error if both Conversions objects return error.
+func (re *AggregateConversions) GetRate(from string, to string) (float64, error) {
+	rate, err := re.customRates.GetRate(from, to)
+
+	if err == nil {
+		if _, isMissingRateErr := err.(*errortypes.NoConversionRate); !isMissingRateErr {
```
mansinahar, line 28: Wouldn't this line lead to a panic? In the previous line you're checking if `err` is `nil` and entering this block only if it is and here you're trying to see if you can cast `err` to `*errortypes.NoConversionRate`

I'd suggest changing this to be more along the lines of:
```
if err == nil {
	return rate, nil
} else if _, isMissingRateErr := err.(*errortypes.NoConversionRate); !isMissingRateErr {
	return nil, err
}
```

endpoints/openrtb2/auction.go
```diff
@@ -434,6 +439,36 @@ func validateSChains(req *openrtb_ext.ExtRequest) error {
...
+// the bidRequest.ext.prebid.currency field is invalid, malfomed or does not represent any actual
+// currency. No error is thrown if bidRequest.ext.prebid.currency is invalid or empty.
+func validateCustomRates(bidReqCurrencyRates *openrtb_ext.ExtRequestCurrency) error {
+	if bidReqCurrencyRates == nil {
+		return nil
+	}
+
+	customRatesOnly := bidReqCurrencyRates.UsePBSRates != nil && !(*bidReqCurrencyRates.UsePBSRates)
+	if customRatesOnly && len(bidReqCurrencyRates.ConversionRates) == 0 {
+		return &errortypes.BadInput{Message: "Required custom currency rates field is empty"}
```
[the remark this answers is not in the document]
guscarreon: In case `usePbseRates` is set to `false` and the customRates map is empty, PBS-Java uses PBS rates anyways:
```
176     /**
177      * Converts price from bidCurrency to adServerCurrency using rates and usepbsrates flag defined in request.
178      * If usepbsrates is true it takes rates from prebid server, if false from request. Default value of usepbsrates
179      * is true.
180      * Throws {@link PreBidException} in case conversion is not possible.
181      */
182     private BigDecimal convertCurrency(BigDecimal price, Map<String, Map<String, BigDecimal>> requestCurrencyRates,
183                                       String adServerCurrency, String bidCurrency, Boolean usepbsrates) {
184         // use Default USD currency if bidder left this field empty. After, when bidder will implement multi currency
185         // support it will be changed to throwing PrebidException.
186         final String effectiveBidCurrency = bidCurrency != null ? bidCurrency : DEFAULT_BID_CURRENCY;
187 +--  7 lines: if (Objects.equals(adServerCurrency, effectiveBidCurrency)) {-------------------------------------------------------------------------------------------------------------------------------------------------------------
194
195         if (BooleanUtils.isFalse(usepbsrates)) {
196             firstPriorityRates = requestCurrencyRates;   // <-- what we call custom currency rates
197             secondPriorityRates = externalCurrencyRates; // <-- PBS rates
198         } else {
199             firstPriorityRates = externalCurrencyRates;
200             secondPriorityRates = requestCurrencyRates;
201         }
202
203         final BigDecimal conversionRate = getConversionRateByPriority(firstPriorityRates, secondPriorityRates,
204                 adServerCurrency, effectiveBidCurrency);
205
206         if (conversionRate == null) {
207             throw new PreBidException(
208                     String.format("Unable to convert from currency %s to desired ad server currency %s",
209                             effectiveBidCurrency, adServerCurrency));
210         }
211
212         return price.divide(conversionRate, DEFAULT_PRICE_PRECISION, RoundingMode.HALF_EVEN);
213     }
src/main/java/org/prebid/server/currency/CurrencyConversionService.java
```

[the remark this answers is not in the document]
guscarreon: I'll go ahead and add the logic for disabling rate conversion when `usePbsRates` is `false` and the custom currency rate map is empty.

currency/aggregate_conversions.go
guscarreon: Agree, done

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
guscarreon: Done. Added a couple of test cases when conversions were disabled

exchange/exchange_test.go
```diff
@@ -507,6 +507,516 @@ func TestTwoBiddersDebugDisabledAndEnabled(t *testing.T) {
...
+		e := new(exchange)
+		e.currencyConverter = mockCurrencyConverter
+
+		// Run test
+		auctionRates := e.getAuctionCurrencyRates(tc.given.bidExtCurrency)
+
+		// When fromCurrency and toCurrency are the same, a rate of 1.00 is always expected
+		rate, err := auctionRates.GetRate("USD", "USD")
+		assert.NoError(t, err, tc.desc)
+		assert.Equal(t, rate, rate, tc.desc)
```
mansinahar, line 1001: Shouldn't the expected rate here be `1`?

exchange/exchange_test.go
```diff
@@ -507,6 +507,516 @@ func TestTwoBiddersDebugDisabledAndEnabled(t *testing.T) {
...
+		if tc.expected.constantRates {
+			rate, err := auctionRates.GetRate("USD", "MXN")
+			assert.Error(t, err, tc.desc)
+			assert.Equal(t, float64(0), rate, tc.desc)
+		} else {
+			for fromCurrency, rates := range tc.expected.resultingRates {
+				for toCurrency := range rates {
+					rate, err := auctionRates.GetRate(fromCurrency, toCurrency)
+					assert.NoError(t, err, tc.desc)
+					assert.Equal(t, rate, rate, tc.desc)
```
mansinahar, line 1013: Shouldn't the expected rate here be the rate that you're getting from the `resultingRates` map? Comparing `rate` with `rate` is always gonna be equal. We need to compare the actual value with the expected value in `resultingRates`

endpoints/openrtb2/sample-requests/custom-currency/valid/custom-rates-override-usepbsrates-true.json
```diff
@@ -0,0 +1,70 @@
+{
```
mansinahar, line 1: please consider changing the directory name from `custom-currency` to `currency-conversion` as these tests are generally testing out currency conversions i.e. we also have tests in this directory where we use just PBS rates and not custom currency, right?

exchange/exchange_test.go
```diff
@@ -507,6 +507,516 @@ func TestTwoBiddersDebugDisabledAndEnabled(t *testing.T) {
...
+	oneDollarBidBidder := &goodSingleBidder{
+		httpRequest: &adapters.RequestData{
+			Method:  "POST",
+			Uri:     mockAppnexusBidService.URL,
+			Body:    []byte("{\"key\":\"val\"}"),
+			Headers: http.Header{},
+		},
+	}
+
+	e := new(exchange)
```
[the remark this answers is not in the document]
SyntaxNode: We need integration tests to ensure the lower level code is hit, but I think the unit test coverage in the lower level code is sufficient.

currency/aggregate_conversions_test.go
```diff
@@ -0,0 +1,86 @@
...
+		},
+	}
+
+	for _, group := range testGroups {
+		for _, tc := range group.testCases {
+			// Execute:
+			rate, err := aggregateConversions.GetRate(tc.from, tc.to)
+
+			// Verify:
+			assert.Equal(t, tc.expectedRate, rate, "rate should be 0: %s\n", tc.desc)
```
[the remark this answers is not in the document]
guscarreon: Agree. Done

currency/aggregate_conversions_test.go
```diff
@@ -0,0 +1,86 @@
...
+			assert.Equal(t, group.expectedError, err, "err shouldn't be nil: %s\n", tc.desc)
```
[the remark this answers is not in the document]
guscarreon: done

currency/rates.go
```diff
@@ -45,8 +45,8 @@ func (r *Rates) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-// GetRate returns the conversion rate between two currencies
-// returns an error in case the conversion rate between the two given currencies is not in the currencies rates map
+// GetRate returns the conversion rate between two currencies returns a MissingConversionRate error
```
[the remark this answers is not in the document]
guscarreon: I listed the types of errors this function is able to return and I think it makes for a clearer description. Let me know your thoughts on it.

endpoints/openrtb2/sample-requests/custom-currency/valid/custom-rates-override-usepbsrates-true.json
guscarreon: All test files in this directory are about bid requests that come with a `requet.ext.prebid.currency` field that defines custom currency conversion rates. I believe the name `custom-currency` is better in here, don't you?

exchange/exchange_test.go
guscarreon: Good catch. copy/paste error.

exchange/exchange_test.go
guscarreon: Yes. Good catch

exchange/exchange.go
```diff
@@ -939,6 +939,34 @@ func (e *exchange) getBidCacheInfo(bid *pbsOrtbBid, auction *auction) (cacheInfo
...
 
+func (e *exchange) getAuctionCurrencyRates(requestRates *openrtb_ext.ExtRequestCurrency) currency.Conversions {
+	if requestRates == nil {
+		// No bidRequest.ext.currency field was found, use PBS rates as usual
+		return e.currencyConverter.Rates()
+	}
+
+	// If bidRequest.ext.currency.usepbsrates is nil, we understand its value as true. It will be false
+	// only if it's explicitly set to false
+	usePbsRates := requestRates.UsePBSRates == nil || *requestRates.UsePBSRates
```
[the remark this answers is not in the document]
guscarreon: We discussed this assignment with @SyntaxNode and we found it elegant and clean. Do you disagree?

openrtb_ext/request.go
```diff
@@ -41,6 +41,13 @@ type ExtRequestPrebid struct {
 	// passing of personally identifiable information doesn't constitute a sale per CCPA law.
 	// The array may contain a single sstar ('*') entry to represent all bidders.
 	NoSale []string `json:"nosale,omitempty"`
+
+	BidReqConversions *ExtRequestCurrency `json:"currency,omitempty"`
```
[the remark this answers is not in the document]
guscarreon: done

errortypes/errortypes.go
```diff
@@ -165,6 +167,32 @@ func (err *BidderTemporarilyDisabled) Severity() Severity {
 	return SeverityWarning
 }
 
+// NoConversionRate is thrown by the currency.Conversions GetRate(from string, to string) method
+// whenever the conversion rate between the two currencies passed as arguments, nor its reciprocal, is found
+// in the currencies rates map nor its reciprocal rate.
+type NoConversionRate struct {
```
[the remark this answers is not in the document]
guscarreon: Totally agree, it's wordy and the last part makes no sense. I added the `"nor its reciprocal"` part.

endpoints/openrtb2/sample-requests/custom-currency/errors/custom-rates-empty-usepbs-false.json
```diff
@@ -0,0 +1,49 @@
+{
+  "description": "usepbsrates set to false forces BidRequest to use custom currency rates but bidRequest.ext.prebid.currency.rates field is empty",
```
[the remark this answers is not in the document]
guscarreon: They differ in line 34. This file sets `"MXN"` as the sole element in the `request.cur` array. Therefore, because PBS currency rates cannot be used, this test returns an error whereas in `endpoints/openrtb2/sample-requests/custom-currency/valid/conversion-disabled.json` we default to `"USD"`, no conversion is needed and a valid one-dollar-bid response is generated without errors.

endpoints/openrtb2/sample-requests/custom-currency/valid/non-usd-bid.json
```diff
@@ -0,0 +1,66 @@
+{
+  "description": "USD BidRequest gets converted because mockbidder bids in foreign currency, custom conversion rate is used",
```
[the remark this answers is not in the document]
guscarreon: Isn't `non-usd-bid.json` more descriptive? This is the sole JSON test where the `mockBidder` bids in a currency other than US dollars.

exchange/exchange_test.go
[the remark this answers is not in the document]
guscarreon: Reduced table test to only two test cases to make sure `HoldAuction` calls `e.getAuctionCurrencyRates`

endpoints/openrtb2/auction.go
```diff
@@ -437,6 +442,38 @@ func validateSChains(req *openrtb_ext.ExtRequest) error {
...
+// currency. No error is thrown if bidRequest.ext.prebid.currency is invalid or empty.
+func validateCustomRates(bidReqCurrencyRates *openrtb_ext.ExtRequestCurrency) error {
+	if bidReqCurrencyRates == nil {
+		return nil
+	}
+
+	customRatesOnly := bidReqCurrencyRates.UsePBSRates != nil && !(*bidReqCurrencyRates.UsePBSRates)
+	if customRatesOnly && len(bidReqCurrencyRates.ConversionRates) == 0 {
+		return &errortypes.Warning{
+			Message:     "currency conversion disabled. cannot use pbs rates and custom currency rates map is empty",
```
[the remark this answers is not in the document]
guscarreon: done

exchange/exchange.go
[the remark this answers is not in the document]
mansinahar: I personally find the above more easier to read and understand but sure we can keep it as is too

endpoints/openrtb2/sample-requests/custom-currency/valid/non-usd-bid.json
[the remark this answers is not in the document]
mansinahar: The reason I suggested something like `reverse-currency-conversion` is because that is the functionality we're trying to test. I agree that this is the only test that where the `mockBidder` bids in a currency that is not USD but ultimately `USD` is just one of the different currencies that we could use for these tests and with this name for someone to understand what this is exactly testing, we're expecting them to have the additional context that all other tests are bidding only in `USD`. If they don't have that additional context, it will be hard for them to understand what we're testing in this file based on the filename. We should try to name the tests as close to the functionality that we're testing as possible.

currency/rates.go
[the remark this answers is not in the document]
mansinahar: I agree that listing the types of error the function returns is helpful. My suggestion was to just move it in a diff line. Right now it all seems like one single sentence. The description of what the function is doing and what type of error it is returning. My suggestion was to consider changing it to the following for better readability:
```
// GetRate returns the conversion rate between two currencies
// Returns a MissingConversionRate error
```

endpoints/openrtb2/sample-requests/custom-currency/errors/custom-rates-empty-usepbs-false.json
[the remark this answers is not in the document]
mansinahar: Ah ok! I somehow missed that. Thanks for the clarification

endpoints/openrtb2/sample-requests/custom-currency/valid/custom-rates-override-usepbsrates-true.json
mansinahar: Ah! I missed that. In that case, can we please add some test cases where the custom currency rates are not defined i.e.`ext.prebid.currency` doesn't exits but PBS rates are defined and we're using those to do the conversions? I'd have assumed we already had these tests but from a quick look I didn't find them in the `sample-requests` JSON tests. And we can have all these currency related tests then under a `currency-conversion` directory

currency/aggregate_conversions_test.go
```diff
@@ -79,8 +79,12 @@ func TestGroupedGetRate(t *testing.T) {
 			rate, err := aggregateConversions.GetRate(tc.from, tc.to)
 
 			// Verify:
-			assert.Equal(t, tc.expectedRate, rate, "rate should be 0: %s\n", tc.desc)
-			assert.Equal(t, group.expectedError, err, "err shouldn't be nil: %s\n", tc.desc)
+			assert.Equal(t, tc.expectedRate, rate, "conversion rate doesn't match the expected rate: %s\n", tc.desc)
+			if group.expectedError != nil {
+				assert.Error(t, group.expectedError, err, "error doesn't match expected: %s\n", tc.desc)
```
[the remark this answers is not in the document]
guscarreon: true. Good catch :)

endpoints/openrtb2/auction_test.go
```diff
@@ -1213,12 +1213,12 @@ func TestValidateCustomRates(t *testing.T) {
...
 		},
 		{
-			desc: "empty custom currency rates but UsePBSRates is set to false, return error",
+			desc: "empty custom currency rates but UsePBSRates is set to false, we don't return error nor warning",
 			inBidReqCurrencies: &openrtb_ext.ExtRequestCurrency{
 				ConversionRates: map[string]map[string]float64{},
 				UsePBSRates:     &boolFalse,
 			},
-			outCurrencyError: &errortypes.BadInput{Message: "Required custom currency rates field is empty"},
+			outCurrencyError: nil,
```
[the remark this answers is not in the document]
guscarreon: Originally that function returned a `BadInput` error when `UsePBSRates` was set to `false` and the custom rates were empty. After feedback from a previous review, we changed the `BadInput` error for a warning but I believe I didn't update this test case accordingly. The reason the test was still passing is that warnings are an implementation of the `error` interface.

endpoints/openrtb2/auction_test.go
```diff
@@ -1184,6 +1198,128 @@ func TestContentType(t *testing.T) {
...
+				ConversionRates: map[string]map[string]float64{
+					"USD": {
+						"FOO": 10.0,
+						"MXN": 0.05,
+						"BAR": 10.95,
+					},
+				},
+				UsePBSRates: &boolFalse,
+			},
+			outCurrencyError: &errortypes.BadInput{Message: "currency code FOO is not recognized or malformed"},
```
SyntaxNode, line 1270: This test is failing due to a subtly where the iteration order for a map is indeterminate. That doesn't affect production, but does cause this test to sometimes fail because you can't be certain if "FOO" or "BAR" will fail first. You can solve this with having just one invalid currency.

Please review other tests to ensure they don't have the same problem.

endpoints/openrtb2/auction_test.go
```diff
@@ -2590,7 +2731,63 @@ func (e *nobidExchange) HoldAuction(ctx context.Context, r exchange.AuctionReque
...
+	return &mockBidExchange{
+		mockBidder: bidder,
+		pbsRates:   mockCurrencyConversionRates,
+	}
+}
+
+// getAuctionCurrencyRates copies the logic of the exchange package for testing purposes
+func (e *mockBidExchange) getAuctionCurrencyRates(customRates *openrtb_ext.ExtRequestCurrency) currency.Conversions {
+	if customRates == nil || len(customRates.ConversionRates) == 0 {
+		return currency.NewRates(time.Now(), e.pbsRates)
```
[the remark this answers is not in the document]
SyntaxNode: I think you forgot to add this comment.

endpoints/openrtb2/auction_test.go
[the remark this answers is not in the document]
SyntaxNode: I don't see this change either. The commit may have gotten lost?

errortypes/errortypes.go
```diff
@@ -165,6 +167,31 @@ func (err *BidderTemporarilyDisabled) Severity() Severity {
...
+func (err *NoConversionRate) Error() string {
+	return fmt.Sprintf("Currency conversion rate not found: '%s' => '%s'", err.fromCur, err.toCur)
+}
+
+func (err *NoConversionRate) Code() int {
+	return NoConversionRateErrorCode
+}
+
+func (err *NoConversionRate) Severity() Severity {
+	return SeverityUnknown
```
SyntaxNode, line 192: Please choose a severity of either Fatal (the auction is stopped when the error is encountered) or Warning (the auction continues and a warning is written to the debug section of the response). A vanilla Go error is considered fatal, so that might make the most sense here - especially since you've defined the code as an error code as opposed to a warning code.

guscarreon: The error gets ignored in `exchange/bidder.go` so the severity doesn't really matter:

endpoints/openrtb2/auction_test.go
guscarreon: Yep, this was also a problem in the JSON tests. Let me know if you agree with the correction

SyntaxNode (dismissed):

mansinahar (approved):

endpoints/openrtb2/sample-requests/custom-currency/valid/non-usd-bid.json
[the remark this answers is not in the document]
guscarreon: Sorry I missed this comment Mansi. Done

endpoints/openrtb2/sample-requests/custom-currency/valid/custom-rates-override-usepbsrates-true.json
guscarreon: Hadn't seen this comment before Mansi. Done. Thanks for your careful review

SyntaxNode (approved):

PR #1754, opened 2021-03-16
Replace TravisCI With GitHub Actions

SyntaxNode:
We are already using GitHub Actions, but still rely on TravisCI for release generation. This PR moves us entirely to GitHub Actions, with a release kicked off by a new tag. Further improvements to the build process will be made this year.


hhhjort (commented):
So github actions prefer the `.yml` extention rather than `.yaml`?

SyntaxNode:
> So github actions prefer the `.yml` extention rather than `.yaml`?

Correct. I don't know why. I just kept with their convention.

hhhjort (dismissed):
LGTM

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Version
+        id: get_version
+        run: echo ::set-output name=version::${GITHUB_REF/refs\/tags\//}
+
+      - name: Create & Publish Release
+        uses: release-drafter/release-drafter@v5.12.1
+        with:
+          config-name: release-drafter-config.yml
```
guscarreon, line 21: Isn't the name of this file `release-drafter.yml` instead?

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+          name: ${{ steps.get_version.outputs.version }}
+          tag: ${{ steps.get_version.outputs.version }}
```
guscarreon, line 23: Should this `tag` line differ from the previous `name` line and look like `${{ steps.get_tag.outputs.tag }}` instead?
The following [example](https://github.com/marketplace/actions/get-latest-tag) fetches the latest tag using: `${{ steps.previoustag.outputs.tag }}`, but I'm still new to this tool though.

.github/workflows/release.yml
SyntaxNode: Good catch. I started with a copy/paste of PBS-Java's tasks and forgot to change this back. Fixed.

.github/workflows/release.yml
SyntaxNode: I'm not using the 'Get Latest Tag' action. That operates on the git repository. I simplified this task to just operate on GitHub event metadata. My custom `get_version` task is correct for the name, tag, and version. In our case, we're using the same value for all 3. I'm passing them explicitly as parameters to avoid using the Release Drafter's auto version increement.

.github/workflows/release.yml
guscarreon: Don't we still need the `config-name` entry? Should we just change the file name?
```
20     with:
21 -     config-name: release-drafter-config.yml
   +     config-name: release-drafter.yml
22       name: ${{ steps.get_version.outputs.version }}
23       tag: ${{ steps.get_version.outputs.version }}
.github/workflows/release.yml 
```

guscarreon (approved):
LGTM

.github/workflows/release.yml
hhhjort: That is what I am wondering as well

hhhjort (approved):

PR #1756, opened 2021-03-17
Test for data race conditions in adapters

guscarreon:
In order to test bidder adapter code for data race conditions, this pull request adds a shared memory race test feature inside file `adapters/adapterstest/test_json.go`. In a nutshell, all that changes is: 
```
102   func runSpec(t *testing.T, filename string, spec *testSpec, bidder adapters.Bidder, isAmpTest, isVideoTest bool) {
103       reqInfo := adapters.ExtraRequestInfo{}
104       if isAmpTest {
105           // simulates AMP entry point
106           reqInfo.PbsEntryPoint = "amp"
107       } else if isVideoTest {
108           reqInfo.PbsEntryPoint = "video"
109       }
    +
    +     // Save original bidRequest values to assert no data races occur inside MakeRequests latter
    +     deepBidReqCopy := openrtb.BidRequest{}
    +     copier.Copy(&deepBidReqCopy, &spec.BidRequest)
    +     shallowBidReqCopy := spec.BidRequest
    +     
    +     // Save original []Imp elements to assert no data races occur inside MakeRequests latter
    +     deepImpCopies := make([]openrtb.Imp, len(spec.BidRequest.Imp))
    +     shallowImpCopies := make([]openrtb.Imp, len(spec.BidRequest.Imp))
    +     for i := 0; i < len(spec.BidRequest.Imp); i++ {
    +     	deepImpCopy := openrtb.Imp{}
    +     	copier.Copy(&deepImpCopy, &spec.BidRequest.Imp[i])
    +     	deepImpCopies = append(deepImpCopies, deepImpCopy)
    +     
    +     	shallowImpCopy := spec.BidRequest.Imp[i]
    +     	shallowImpCopies = append(shallowImpCopies, shallowImpCopy)
    +     }
    +     
110       actualReqs, errs := bidder.MakeRequests(&spec.BidRequest, &reqInfo)
111       diffErrorLists(t, fmt.Sprintf("%s: MakeRequests", filename), errs, spec.MakeRequestErrors)
112       diffHttpRequestLists(t, filename, actualReqs, spec.HttpCalls)
113  
    +     // Assert no data races occur using original bidRequest copies of references and values
    +     assertNoDataRace(t, &deepBidReqCopy, &shallowBidReqCopy, deepImpCopies, shallowImpCopies)
    +     
114       bidResponses := make([]*adapters.BidderResponse, 0)
115  
116       var bidsErrs = make([]error, 0, len(spec.MakeBidsErrors))
117       for i := 0; i < len(actualReqs); i++ {
118           thisBidResponse, theseErrs := bidder.MakeBids(&spec.BidRequest, spec.HttpCalls[i].Request.ToRequestData(t), spec.HttpCalls[i].Response.ToResponseData(t))
119           bidsErrs = append(bidsErrs, theseErrs...)
120           bidResponses = append(bidResponses, thisBidResponse)
121       }
122  
123       diffErrorLists(t, fmt.Sprintf("%s: MakeBids", filename), bidsErrs, spec.MakeBidsErrors)
124  
125       for i := 0; i < len(spec.BidResponses); i++ {
126           diffBidLists(t, filename, bidResponses[i].Bids, spec.BidResponses[i].Bids)
127       }
128   }
adapters/adapterstest/test_json.go
```
The rest is simply the implementation of `assertNoDataRace` and a small refactoring effort on the `runSpec` function. Tests will fail since data races were found in some adapters.


adapters/adapterstest/test_json.go
```diff
@@ -316,3 +362,52 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
...
+
+	// Assert slice fields were not modified by bidder adapter MakeRequests implementation
+	assert.ElementsMatch(t, bidRequestBefore.WSeat, bidRequestAfter.WSeat, "Data race in BidRequest.[]WSeat array")
+	assert.ElementsMatch(t, bidRequestBefore.BSeat, bidRequestAfter.BSeat, "Data race in BidRequest.[]BSeat array")
+	assert.ElementsMatch(t, bidRequestBefore.Cur, bidRequestAfter.Cur, "Data race in BidRequest.[]Cur array")
+	assert.ElementsMatch(t, bidRequestBefore.WLang, bidRequestAfter.WLang, "Data race in BidRequest.[]WLang array")
+	assert.ElementsMatch(t, bidRequestBefore.BCat, bidRequestAfter.BCat, "Data race in BidRequest.[]BCat array")
+	assert.ElementsMatch(t, bidRequestBefore.BAdv, bidRequestAfter.BAdv, "Data race in BidRequest.[]BAdv array")
+	assert.ElementsMatch(t, bidRequestBefore.BApp, bidRequestAfter.BApp, "Data race in BidRequest.[]BApp array")
+	assert.ElementsMatch(t, bidRequestBefore.Ext, bidRequestAfter.Ext, "Data race in BidRequest.[]Ext array")
```
SyntaxNode, line 388: I think this looks right. I don't recall exactly how Go handles slice fields during a shallow copy. Why wouldn't `assert.Equal` work correctly here?

adapters/adapterstest/test_json.go
```diff
@@ -107,23 +109,67 @@ func runSpec(t *testing.T, filename string, spec *testSpec, bidder adapters.Bidd
...
+	deepImpCopies := make([]openrtb.Imp, len(spec.BidRequest.Imp))
+	shallowImpCopies := make([]openrtb.Imp, len(spec.BidRequest.Imp))
+	for i := 0; i < len(spec.BidRequest.Imp); i++ {
+		deepImpCopy := openrtb.Imp{}
+		copier.Copy(&deepImpCopy, &spec.BidRequest.Imp[i])
+		deepImpCopies = append(deepImpCopies, deepImpCopy)
+
+		shallowImpCopy := spec.BidRequest.Imp[i]
+		shallowImpCopies = append(shallowImpCopies, shallowImpCopy)
+	}
```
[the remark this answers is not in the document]
guscarreon: With the deep copy option the `nil` slice fields found in either `Imp[i]` or any of its children (`*Banner`, `*Video`, `*Native` and so on) would be initialized to an empty slice as discussed in the answer above.

adapters/adapterstest/test_json.go
```diff
@@ -107,23 +109,67 @@ func runSpec(t *testing.T, filename string, spec *testSpec, bidder adapters.Bidd
...
+
+	testMakeBidsImpl(t, filename, spec, bidder, actualReqs)
+}
+
+func testMakeRequestsImpl(t *testing.T, filename string, spec *testSpec, bidder adapters.Bidder, reqInfo *adapters.ExtraRequestInfo) []*adapters.RequestData {
+	t.Helper()
+
+	// Save original bidRequest values to assert no data races occur inside MakeRequests latter
+	deepBidReqCopy := openrtb.BidRequest{}
+	copier.Copy(&deepBidReqCopy, &spec.BidRequest)
```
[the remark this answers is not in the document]
guscarreon: After running some tests, the difference I found in results between `copier.CopyWithOption`, and `copier.Copy` is that is that `copier.CopyWithOption(&deepBidReqCopy, &spec.BidRequest, copier.Option{DeepCopy: true})` initializes slices even if they were nil in the original object. For instance types `Cat []string`, `SectionCat []string`, `PageCat []string` part of `BidRequest.Site` get initialized as empty non-nil slices:
```
   Diff:
   --- Expected
   +++ Actual
   @@ -4,8 +4,5 @@
     Domain: (string) "",
   - Cat: ([]string) {
   - },
   - SectionCat: ([]string) {
   - },
   - PageCat: ([]string) {
   - },
   + Cat: ([]string) <nil>,
   + SectionCat: ([]string) <nil>,
   + PageCat: ([]string) <nil>,
     Page: (string) "",
```
It also initializes nil `Ext` objects to empty `json.RawMessage`
```
                Diff:
                --- Expected
                +++ Actual
                @@ -31,4 +31,3 @@
                  MACMD5: (string) "",
                - Ext: (json.RawMessage) {
                - }
                + Ext: (json.RawMessage) <nil>
                 })
Test:           TestJsonSamples
Messages:       Data race in BidRequest.Device field
```
I believe this is not the functionality that we are looking for. First idea that comes to mind is simply to compare slice elements only if not nil but I guess it doesn't exactly look better: 
```
	if shallowCopy.WSeat != {
        // Compare only if originally was something other than `nil`
	    assert.ElementsMatch(t, deepCopy.WSeat, shallowCopy.WSeat, "Data race in BidRequest.[]WSeat array")
    }
```

adapters/adapterstest/test_json.go
```diff
@@ -107,23 +109,67 @@ func runSpec(t *testing.T, filename string, spec *testSpec, bidder adapters.Bidd
...
+
+	// Run MakeRequests
+	actualReqs, errs := bidder.MakeRequests(&spec.BidRequest, reqInfo)
+
+	// Compare MakeRequests actual output versus expected values found in JSON file
+	assertErrorList(t, fmt.Sprintf("%s: MakeRequests", filename), errs, spec.MakeRequestErrors)
+	assertMakeRequestsOutput(t, filename, actualReqs, spec.HttpCalls)
+
+	// Assert no data races occur using original bidRequest copies of references and values
+	assertNoDataRace(t, &deepBidReqCopy, &shallowBidReqCopy, deepImpCopies, shallowImpCopies)
```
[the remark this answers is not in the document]
guscarreon: The first thing that brought me to write `assertNoDataRace` was to have descriptive fail messages such as `"Data race in BidRequest.Site field"` whenever a race condition was found so it would be easier to locate the problem in the adapter source code. But now that we opted to copy with `deepcopy.Copy()`, that avoids the initialization of the zero-lenght slices, I believe assert.Equal()` simplifies the code and yields descriptive-enough test-fail messages

adapters/adapterstest/test_json.go
guscarreon: You are right, it does work. Changed.

guscarreon: The `assert.ElementsMatch()` was removed

SyntaxNode: How about giving https://github.com/mohae/deepcopy a try?

The `github.com/getlantern/deepcopy` library is not worth referencing, IMHO. Consider writing a helper method instead of importing the library. The following is the full source code of the library:

```
func Copy(dst interface{}, src interface{}) error {
	if dst == nil {
		return fmt.Errorf("dst cannot be nil")
	}
	if src == nil {
		return fmt.Errorf("src cannot be nil")
	}
	bytes, err := json.Marshal(src)
	if err != nil {
		return fmt.Errorf("Unable to marshal src: %s", err)
	}
	err = json.Unmarshal(bytes, dst)
	if err != nil {
		return fmt.Errorf("Unable to unmarshal into dst: %s", err)
	}
	return nil
}
```

adapters/adapterstest/test_json.go
```diff
@@ -331,3 +324,110 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
...
+	// Shallow copy PBS core provides to adapters
+	shallowReqCopy := *original
+
+	// Save original []Imp elements to assert no data races occur inside MakeRequests latter
+	deepReqCopy.Imp = make([]openrtb2.Imp, len(original.Imp))
+	shallowReqCopy.Imp = make([]openrtb2.Imp, len(original.Imp))
+	for i := 0; i < len(original.Imp); i++ {
+		deepImpCopy := openrtb2.Imp{}
+		copier.Copy(&deepImpCopy, original.Imp[i])
+		deepReqCopy.Imp = append(deepReqCopy.Imp, deepImpCopy)
```
SyntaxNode, line 363: Sorry, I think you explained this before. Why do we need to do make deep copies of the impression objects? Does `copier.Copy(&deepReqCopy, original)` not already do that?

adapters/adapterstest/test_json.go
```diff
@@ -331,3 +324,110 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
...
+	}
+
+	// Assert actual errors thrown by MakeBids implementation versus expected JSON-defined spec.MakeBidsErrors
+	assertErrorList(t, fmt.Sprintf("%s: MakeBids", filename), bidsErrs, spec.MakeBidsErrors)
+
+	// Assert MakeBids implementation BidResponses with expected JSON-defined spec.BidResponses[i].Bids
+	for i := 0; i < len(spec.BidResponses); i++ {
+		assertMakeBidsOutput(t, filename, bidResponses[i], spec.BidResponses[i].Bids)
+	}
+}
```
SyntaxNode, line 433: Did we not have this functionality before?

adapters/adapterstest/test_json.go
```diff
@@ -107,24 +116,10 @@ func runSpec(t *testing.T, filename string, spec *testSpec, bidder adapters.Bidd
 	} else if isVideoTest {
 		reqInfo.PbsEntryPoint = "video"
 	}
-	actualReqs, errs := bidder.MakeRequests(&spec.BidRequest, &reqInfo)
-	diffErrorLists(t, fmt.Sprintf("%s: MakeRequests", filename), errs, spec.MakeRequestErrors)
-	diffHttpRequestLists(t, filename, actualReqs, spec.HttpCalls)
 
-	bidResponses := make([]*adapters.BidderResponse, 0)
+	actualReqs := testMakeRequestsImpl(t, filename, spec, bidder, &reqInfo)
```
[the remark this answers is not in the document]
guscarreon: done

adapters/adapterstest/test_json.go
```diff
@@ -227,10 +222,10 @@ func diffErrorLists(t *testing.T, description string, actual []error, expected [
 	}
 }
 
-func diffBidLists(t *testing.T, filename string, response *adapters.BidderResponse, expected []expectedBid) {
+func assertMakeBidsOutput(t *testing.T, filename string, actualBidderResp *adapters.BidderResponse, expected []expectedBid) {
```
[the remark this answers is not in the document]
guscarreon: I thought the `actual` part would make it more explicit that those are the values we want to assert. Removed

adapters/adapterstest/test_json.go
guscarreon: Because the deep copy version of this library was throwing false positives when comparing the `Ext` fields at every level. Switched to github.com/mohae/deepcopy to avoid it. The for loop could not be removed though because we still have to create the shallow copies of the `[]Imp` array elements.

adapters/adapterstest/test_json.go
```diff
@@ -331,3 +317,109 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
...
+	if assert.Len(t, impsAfter, len(impsBefore), "Original []Imp array was modified and length is not equal to original after MakeRequests was called. File:%s", filename) {
+		// Assert no data races occured in individual Imp elements
+		for i := 0; i < len(impsBefore); i++ {
+			assert.Equal(t, impsBefore[i].Banner, impsAfter[i].Banner, "Data race in bidRequest.Imp[%d].Banner field. File:%s", i, filename)
+			assert.Equal(t, impsBefore[i].Video, impsAfter[i].Video, "Data race in bidRequest.Imp[%d].Video field. File:%s", i, filename)
+			assert.Equal(t, impsBefore[i].Audio, impsAfter[i].Audio, "Data race in bidRequest.Imp[%d].Audio field. File:%s", i, filename)
+			assert.Equal(t, impsBefore[i].Native, impsAfter[i].Native, "Data race in bidRequest.Imp[%d].Native field. File:%s", i, filename)
+			assert.Equal(t, impsBefore[i].PMP, impsAfter[i].PMP, "Data race in bidRequest.Imp[%d].PMP field. File:%s", i, filename)
+			assert.Equal(t, impsBefore[i].Secure, impsAfter[i].Secure, "Data race in bidRequest.Imp[%d].Secure field. File:%s", i, filename)
+			assert.ElementsMatch(t, impsBefore[i].Metric, impsAfter[i].Metric, "Data race in bidRequest.Imp[%d].[]Metric array. File:%s", i)
```
SyntaxNode, line 396: Should we include the other slices in the data race tests too?

adapters/adapterstest/test_json.go
```diff
@@ -331,3 +317,109 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
 		}
 	}
 }
+
+// testMakeBidsImpl asserts the results of the bidder MakeRequests implementation against the
```
SyntaxNode, line 321: Nitpick: Copy/paste error on the func name in this comment.

SyntaxNode (commented):
Looks great! This is very straight forward and easy to follow. Thank you for expanding on the comments for the data race condition tests. Just one comment left.

adapters/adapterstest/test_json.go
SyntaxNode: We decided to add all slices from the BidRequest and the Imp using an assert.Equal. Slices are shared memory which also may not be mutated.

adapters/adapterstest/test_json.go
```diff
@@ -375,6 +375,16 @@ func assertNoDataRace(t *testing.T, bidRequestBefore *openrtb2.BidRequest, bidRe
...
 
+	// Assert slice fields were not modified by bidder adapter MakeRequests implementation
+	assert.Equal(t, bidRequestBefore.WSeat, bidRequestAfter.WSeat, "Data race in BidRequest.[]WSeat array")
+	assert.Equal(t, bidRequestBefore.BSeat, bidRequestAfter.BSeat, "Data race in BidRequest.[]BSeat array")
+	assert.Equal(t, bidRequestBefore.Cur, bidRequestAfter.Cur, "Data race in BidRequest.[]Cur array")
+	assert.Equal(t, bidRequestBefore.WLang, bidRequestAfter.WLang, "Data race in BidRequest.[]WLang array")
+	assert.Equal(t, bidRequestBefore.BCat, bidRequestAfter.BCat, "Data race in BidRequest.[]BCat array")
+	assert.Equal(t, bidRequestBefore.BAdv, bidRequestAfter.BAdv, "Data race in BidRequest.[]BAdv array")
+	assert.Equal(t, bidRequestBefore.BApp, bidRequestAfter.BApp, "Data race in BidRequest.[]BApp array")
+	assert.Equal(t, bidRequestBefore.Ext, bidRequestAfter.Ext, "Data race in BidRequest.[]Ext array")
```
SyntaxNode, line 386: This is a string, not a slice. It's ok for the bidder to modify it since strings in Go are immutable.

adapters/adapterstest/test_json.go
```diff
@@ -375,6 +375,16 @@ func assertNoDataRace(t *testing.T, bidRequestBefore *openrtb2.BidRequest, bidRe
 	assert.Equal(t, bidRequestBefore.Source, bidRequestAfter.Source, "Data race in BidRequest.Source field in file %s", filename)
 	assert.Equal(t, bidRequestBefore.Regs, bidRequestAfter.Regs, "Data race in BidRequest.Regs field in file %s", filename)
 
+	// Assert slice fields were not modified by bidder adapter MakeRequests implementation
```
SyntaxNode, line 378: Please add in this comment that the `Imp[]` slice is purposefully excluded.

adapters/adapterstest/test_json.go
```diff
@@ -331,3 +317,84 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
...
+//  2) A deep copy that will preserve the original values of all the fields. This copy remains untouched
+//     by the adapters' processes and serves as reference of what the shared memory values should still
+//     be after the `MakeRequests()` call.
+//
+// The original values stored in the deep copy will be compared against the shared memory values and
+// discrepancies will point to data race conditions in an adapter's `MakeRequests()` implementation.
+//
+// Because neither the shallow nor the deep copies are passed to `MakeRequests()` we know static fields
+// in openrtb2.BidRequest nor openrtb2.Imp elements will change, therefore, we can reliably compare
+// reference values using the `assert.Equal()` method.
```
guscarreon, line 335: Didn't end up liking this huge comment. It probably needs rewording. Suggestions?

adapters/adapterstest/test_json.go
```diff
@@ -331,3 +317,84 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
...
+
+	// Run MakeRequests
+	requests, errs := bidder.MakeRequests(&spec.BidRequest, reqInfo)
+
+	// Compare MakeRequests actual output versus expected values found in JSON file
+	assertErrorList(t, fmt.Sprintf("%s: MakeRequests", filename), errs, spec.MakeRequestErrors)
+	assertMakeRequestsOutput(t, filename, requests, spec.HttpCalls)
+
+	// Assert no data races occur using original bidRequest copies of references and values
+	assert.Equal(t, deepBidReqCopy, shallowBidReqCopy, "Data race found. Test: %s \n", filename)
```
guscarreon, line 350: After many local tests, local adapter modifications to purposefully create race conditions, and trying out many deep copy libraries available out there: we finally found the mix to simply `assert.Equal()` instead of writing an assert function of our own. The only downside of not using our previously implemented `assertNoDataRaces(...)` function is the less descriptive `"Data race found. Test: %s \n"` message compared to something like: `"Data race in bidRequest.Imp[0].Native field. File: %s \n"`.  The latter would point to the exact shared memory field being modified.

I still think it's worth it though.

SyntaxNode (commented):
The new algorithm looks good to me. I ran through a bunch of forced data race conditions and they were caught correctly.

adapters/adapterstest/test_json.go
```diff
@@ -331,3 +317,84 @@ func diffJson(t *testing.T, description string, actual []byte, expected []byte)
...
+	if err != nil {
+		return nil, nil, err
+	}
+	deepReqCopy := cpy.(*openrtb2.BidRequest)
+
+	shallowReqCopy := *original
+
+	// If we call `copy` on a nil []Imp, the shallowReqCopy.Imp slice will initialize to openrtb2.Imp{}
+	// and not nil, which will lead to false positives.
+	if original.Imp != nil {
```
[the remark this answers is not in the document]
SyntaxNode: I'd like to see the comment updated as to why the `Imp` slice of the shallow copy needs to have this special behavior. The current comment is only part of the reason.

SyntaxNode (approved):

mansinahar (approved):

PR #1758, opened 2021-03-17
Applogy: Fix Shared Memory Overwriting

guscarreon:
Copy of reference field `Banner` needed to be made inside `adapters/applogy/applogy.go` in order to avoid a data race on shared memory. Authors of source code should be notified and encourage to make any changes to this PR if they disagree with the proposed approach.


adapters/applogy/applogytest/exemplary/simple-banner.json
```diff
@@ -49,9 +49,7 @@
 						"format": [{
 							"h": 250,
 							"w": 300
-						}],
-						"h": 250,
-						"w": 300
+						}]
```
SyntaxNode, line 52: Why did the test case change?

adapters/applogy/applogy.go
```diff
@@ -45,6 +45,7 @@ func (a *ApplogyAdapter) MakeRequests(request *openrtb.BidRequest, _ *adapters.E
 				banner.W = &format.W
 				banner.H = &format.H
 			}
+			impression.Banner = &banner
```
SyntaxNode, line 48: You should move the banner copy and reference inside the `if banner.W == nil || banner.H == nil || *banner.W == 0 || *banner.H == 0 {` statement, because otherwise the copy isn't necessary.

SyntaxNode (commented):
Please email the Applogy through their bidder info email address to bring this change to their attention.

adapters/applogy/applogy.go
hhhjort: Also, `impression` is a copy of an element in the `Imp` slice. The assignment modifies the the `Banner` pointer of the copy, not the actual impression within the `Imp` array. The `Imp` slice is already a shallow copy of the original request, so it is safe to modify the `Banner` pointer in one of the elements.

SyntaxNode: Ah, that's a good point @hhhjort. It shouldn't have failed the new memory tests then though?

guscarreon: Did we throw a false positive here then?

hhhjort: It doesn't modify the existing Banner (which would be bad as that is pointing to the shared original request Banner). In line 36 (banner := *impression.Banner), you make `banner` a local shallow copy of the original `Banner`. You then modify that local copy. Line 48 you assign `impression.Banner`, where `impression` is the loop copy of the actual element in `Imp` to point to the new local copy of `banner`. So the temporary `impression` imp object gets it's `Banner` pointer assigned to a new, local `banner` object, and the object in the `Imp` slice that we are iterating over is unchanged. Thus when the loop ends, `request.Imp[n].Banner` is unchanged.

guscarreon: I committed an update, please let me know if we got it right this time

guscarreon:
@IncSW,
We recently coded pull request #1756 which is meant to enable Prebid Server to find data race conditions in adapter bidder source code. According to this newly coded feature, your adapter might be writing to shared memory and this PR implements an approach to prevent it. Please let us know whether or not you agree with the changes proposed in this pull request or if you rather want to code the fix yourself. 

adapters/applogy/applogy.go
```diff
@@ -70,7 +70,7 @@ func (a *ApplogyAdapter) MakeRequests(request *openrtb.BidRequest, _ *adapters.E
 			errs = append(errs, errors.New("Applogy token required"))
 			continue
 		}
-		request.Imp = impressions[i : i+1]
+		request.Imp = []openrtb.Imp{impression}
```
SyntaxNode, line 73: I think this the best approach for this adapter. Based on the `request.Imp = impressions` line later in this adapter, it seems their intent is to preserve the bid request. Any other approach would not accomplish that goal.

IncSW:
@guscarreon,
I'm agree with these changes.

SyntaxNode (approved):

hhhjort (approved):
LGTM

PR #1759, opened 2021-03-17
Pubmatic: Fix Shared Memory Overwriting

guscarreon:
Copy of reference field Banner needed to be made inside `adapters/pubmatic/pubmatic.go` in order to avoid a data race on shared memory. Authors of source code should be notified and encourage to make any changes to this PR if they disagree with the proposed approach.

### DIFF
--- a/adapters/pubmatic/pubmatic.go
+++ b/adapters/pubmatic/pubmatic.go
@@ -433,7 +433,7 @@ func assignBannerSize(banner *openrtb.Banner) (*openrtb.Banner, error) {
 	}
 
 	if len(banner.Format) == 0 {
-		return banner, errors.New(fmt.Sprintf("No sizes provided for Banner %v", banner.Format))
+		return nil, errors.New(fmt.Sprintf("No sizes provided for Banner %v", banner.Format))
 	}
 
 	return assignBannerHeightAndWidth(banner, banner.Format[0].H, banner.Format[0].H), nil
--- a/adapters/pubmatic/pubmatic.go
+++ b/adapters/pubmatic/pubmatic.go
@@ -158,7 +158,8 @@ func (a *PubmaticAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder 
 					}
 
 					pbReq.Imp[i].TagID = strings.TrimSpace(adSlot[0])
-					pbReq.Imp[i].Banner = assignBannerHeightAndWidth(pbReq.Imp[i].Banner, uint64(height), uint64(width))
+					pbReq.Imp[i].Banner.H = openrtb.Uint64Ptr(uint64(height))
+					pbReq.Imp[i].Banner.W = openrtb.Uint64Ptr(uint64(width))
 
 					if len(params.Keywords) != 0 {
 						kvstr := prepareImpressionExt(params.Keywords)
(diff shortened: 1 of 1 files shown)


SyntaxNode:
Please contact PubMatic using their bidder info email address to inform them of this change.

adapters/pubmatic/pubmatic.go
```diff
@@ -158,8 +158,7 @@ func (a *PubmaticAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder
 					}
 
 					pbReq.Imp[i].TagID = strings.TrimSpace(adSlot[0])
-					pbReq.Imp[i].Banner.H = openrtb.Uint64Ptr(uint64(height))
-					pbReq.Imp[i].Banner.W = openrtb.Uint64Ptr(uint64(width))
+					pbReq.Imp[i].Banner = assignBannerHeightAndWidth(pbReq.Imp[i].Banner, uint64(height), uint64(width))
```
SyntaxNode, line 161: Don't bother with changes in the old interfaces, such as `Call`. That shouldn't be caught by the json test framework.

adapters/pubmatic/pubmatic.go
```diff
@@ -428,25 +426,27 @@ func validateAdSlot(adslot string, imp *openrtb.Imp) error {
...
 }
 
-func assignBannerSize(banner *openrtb.Banner) error {
-	if banner == nil {
-		return nil
-	}
+func assignBannerHeightAndWidth(banner *openrtb.Banner, h uint64, w uint64) *openrtb.Banner {
+	bannerCopy := *banner
+
+	bannerCopy.W = new(uint64)
```
SyntaxNode, line 432: Please use the `openrtb.Uint64Ptr()` helper method. It's a bit cleaner:
```
bannerCopy.W = openrtb.Uint64Ptr(w)
bannerCopy.H = openrtb.Uint64Ptr(h)
```

guscarreon:
@sachin-pubmatic,

adapters/pubmatic/pubmatic.go
guscarreon: Rolled back

adapters/pubmatic/pubmatic.go
```diff
@@ -428,25 +427,25 @@ func validateAdSlot(adslot string, imp *openrtb.Imp) error {
...
-
+func assignBannerSize(banner *openrtb.Banner) (*openrtb.Banner, error) {
 	if banner.W != nil && banner.H != nil {
-		return nil
+		return banner, nil
 	}
 
 	if len(banner.Format) == 0 {
-		return errors.New(fmt.Sprintf("No sizes provided for Banner %v", banner.Format))
+		return banner, errors.New(fmt.Sprintf("No sizes provided for Banner %v", banner.Format))
```
SyntaxNode, line 436: I think this should return a `nil` banner pointer if there's an error.

guscarreon: Good catch. Removed

hhhjort (approved):
LGTM

SyntaxNode (approved):

SyntaxNode:
We received approval via email for this change.

bjorn-lw:
I just downloaded and tested the latest version of Prebid server (0.155) and noticed that with this version Pubmatic is starting to respond with unexpected banner sizes.

bjorn-lw:
I just downloaded and tested the latest version of Prebid server (0.155) and noticed that with that version Pubmatic is starting to respond with unexpected banner sizes.

For instance, a request that requests only 160x600 as size, Pubmatic may respond with 600x600.

If I use an earlier version of Prebid Server (sorry, I can currently only test 0.139.1 easily) this problem does not occur.

But after looking at what has changed in the pubmatic adapter, this looks like something that maybe is causing this. I have no idea why though, and I may be completely off.

I will compile earlier versions and test until the problem goes away to see exactly in which version it happens.

bjorn-lw:
The issue is NOT present in 0.152.0 so this is most likely related to this change.

Sorry for not being too helpful in locating the source of the issue, GO is not my first language, but I will make an honest try.

PR #1762, opened 2021-03-17
Beachfront: Fix Shared Memory Overwriting

guscarreon:
The Beachfront adapter seems to be writting to shared memory inside the `getVideoRequests` function, particulartly to the `bfReqs[i].Request.Device` and the `bfReqs[i].Request.Site` fields. This pull request proposes changes to stop the race conditions, but authors of the beachfront adapter source code should be notified and encouraged to make any changes to this PR if they disagree with the proposed approach.

### DIFF
--- a/adapters/beachfront/beachfront.go
+++ b/adapters/beachfront/beachfront.go
@@ -407,12 +407,11 @@ func getVideoRequests(request *openrtb.BidRequest) ([]beachfrontVideoRequest, []
 		bfReqs[i].Request = *request
 		var secure int8
 
-		var deviceCopy *openrtb.Device
+		var deviceCopy openrtb.Device
 		if bfReqs[i].Request.Device == nil {
-			deviceCopy = &openrtb.Device{}
+			deviceCopy = openrtb.Device{}
 		} else {
-			device := *bfReqs[i].Request.Device
-			deviceCopy = &device
+			deviceCopy = *bfReqs[i].Request.Device
 		}
 
 		if beachfrontExt.VideoResponseType == "nurl" {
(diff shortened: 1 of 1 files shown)


adapters/beachfront/beachfront.go
```diff
@@ -437,10 +443,11 @@ func getVideoRequests(request *openrtb.BidRequest) ([]beachfrontVideoRequest, []
 			}
 		}
 
-		if bfReqs[i].Request.Device != nil && bfReqs[i].Request.Device.DeviceType == 0 {
+		if deviceCopy != nil && deviceCopy.DeviceType == 0 {
```
SyntaxNode, line 446: It's not possible for `deviceCopy` to be `nil`.

guscarreon: Yep, I noticed, left the check anyways but I can remove

guscarreon:
@muncha,

When it comes to adapters' source code improvements, we usually wait a week for owners to come back to us. If we get no feedback, we would go ahead and review and merge this pull request ourselves.

adapters/beachfront/beachfront.go
SyntaxNode: Cool. Might as well clean that up while we're here. Likely left over from an older implementation.

muncha:
@guscarreon - Thanks, this looks great. I've got nothing to add.

SyntaxNode (approved):

adapters/beachfront/beachfront.go
```diff
@@ -407,22 +407,27 @@ func getVideoRequests(request *openrtb.BidRequest) ([]beachfrontVideoRequest, []
 		bfReqs[i].Request = *request
 		var secure int8
 
+		var deviceCopy openrtb.Device
 		if bfReqs[i].Request.Device == nil {
-			bfReqs[i].Request.Device = &openrtb.Device{}
+			deviceCopy = openrtb.Device{}
```
hhhjort, line 412: Doesn't line 410 already do this? I think you only need one of these two lines.

hhhjort (approved):
LGTM

PR #1764, opened 2021-03-18
33across: Fix Shared Memory Overwriting

guscarreon:
@curlyblueeagle,

### DIFF
--- a/adapters/33across/33across.go
+++ b/adapters/33across/33across.go
@@ -228,7 +228,7 @@ func (a *TtxAdapter) MakeBids(internalRequest *openrtb.BidRequest, externalReque
 }
 
 func validateVideoParams(video *openrtb.Video, prod string) (*openrtb.Video, error) {
-	videoCopy := video
+	videoCopy := *video
 	if videoCopy.W == 0 ||
 		videoCopy.H == 0 ||
 		videoCopy.Protocols == nil ||
@@ -252,7 +252,7 @@ func validateVideoParams(video *openrtb.Video, prod string) (*openrtb.Video, err
 		}
 	}
 
-	return videoCopy, nil
+	return &videoCopy, nil
 }
 
 func getBidType(ext bidExt) openrtb_ext.BidType {


SyntaxNode (approved):

hhhjort (approved):

curlyblueeagle:
I am okay with the proposed changes, can you explain where a race condition could potentially occur in the code?

guscarreon:
> I am okay with the proposed changes, can you explain where a race condition could potentially occur in the code?

@curlyblueeagle thank you for setting aside time to review this change. The `openrtb.BidRequest` parameter that `MakeRequests` receives is a "shallow" copy. A copy is shallow when the exact values of basic types such as `string`s and `int`s where copied from the original to the destination but as far as reference types: new copies of these objects where **not** created. Therefore, the shallow copy preserves the pointer values of the orginal which are pointing to shared memory that, given the parallel nature of Prebid Server core, might be rewritten by a different thread. Out of the `bidRequest` fields we have to carefully watch the reference types `bidRequest.Site`, `bidRequest.Imp[i]`, `bidRequest.App`, `bidRequest.Device`, `bidRequest.User`, `bidRequest.Source`, and `bidRequest.Regs`.
```
type BidRequest struct {
      .
      .
    ID string `json:"id"`  <-- Basic type, can be read and modified without a problem
      .
      .
    Imp []Imp `json:"imp"` <-- Reference type, every `Imp` element can be read BUT NOT modified
      .                        because it points to memory that might be shared with other parallel processes
      .
    Site *Site `json:"site,omitempty"` <-- Reference type, every `Imp` element can be read BUT NOT modified
      .                                    because it points to memory that might be accessed by other parallel processes
      .                                    The only safe way to modify is to create a copy, modify the copy and assign the pointer 
      .                                    to the local pointer value: 
      .                                       siteLocalCopy := *bidRequest.Site //Creates a shallow copy, we can modify its basic types now
      .                                       siteLocalCopy.Name = "anyOtherValue" //Perfectly safe
      .                                       bidRequest.Site = &siteLocalCopy //perfectly safe because the original space in memory pointed
      .                                                                        //by bidRequest.Site only got read and not modified
    Test int8 `json:"test,omitempty"` <-- Basic type, can be read and modified without a problem
      .
      .
    Ext json.RawMessage `json:"ext,omitempty"` <-- Reference type but these fields usually get Unmarshalled which is a read
                                                   operation so usually there's no problem with these 
}
```
Same goes for reference fields found in `bidRequest.Site`, `bidRequest.Imp[i]`, `bidRequest.App`, `bidRequest.Device`, `bidRequest.User`, `bidRequest.Source`, and `bidRequest.Regs`. That's the reason we needed to create a copy of `bidRequest.Imp[i].Video` in your adapter. `Video` is a reference type of the `Imp` struct.

I hope this clarifies a bit.

curlyblueeagle:
@guscarreon thanks for the explanation, super helpful! We will definitely bear this mind in future.

SyntaxNode:
> bidRequest.Imp[i]

You don't need to worry about changes to those elements. Like the bid request, this is also a deep copy. This is the only exception within the request object.

PR #1765, opened 2021-03-19
FPD: Allow imp.ext.data To Passthrough To Adapters

SyntaxNode:
All json fields within `request.imp[].ext` are expected to be bidder names. Exemptions need to be made for special cases, such as for First Party Data. Other parts of code enforce the bidder name "data" is forbidden making this a safe operation.

Implements https://github.com/prebid/prebid-server/issues/1768


hhhjort (approved):
A nice simple addition to the list of exceptions. Looks good to me.

bsardo (approved):
LGTM

PR #1766, opened 2021-03-19
Stored Request A/B Testing POC #1645

### LINKED ISSUE #1645: Stored Request A/B testing support for PBS
Description
=========

We need a way for PBS-Go to randomly choose "experimental" values for any stored request, as a way to validate potential changes in a statistically relevant way and without impacting the bulk of the revenue. These experiments could be:

- adding a new bidder
- changing floors
- ad sizes 
- etc

How it works
=========

A list of experiments such as the one below is saved in the stored request under `ext.prebid.storedrequest.ab_config`. Each experiment has a code that identifies it for analytics purposes, the desired split ratio, the id of the stored request to use instead of the requested one and a map to replace eventual stored imp ids.

stored request <stored-req-id-1>
```json
{
  "ext": {
    "prebid": {
      "storedrequest": {
        "ab_config": [
	  {
            "code": "add_new_bidder_test_1",
            "ratio": 10.0,
            "request_id": "<new-stored-req-id>",
            "imp_ids": {
              "imp-id-1": "<new-stored-imp-id-1>"
              ...
            }
	  },
	  ...
       ]
```

When PBS receives an incoming auction request that uses stored request `stored-req-id-1`  and stored imp `stored-imp-id-1`, for 10% of such requests it will use the data from`replacement-stored-req-id` and `replacement-stored-req-id-1` instead. When the substitution happens, the field `ext.prebid.storedrequest.ab_code` will be set to `add_new_bidder_test_1` for analytics modules. No changes are done to the control group (the one that uses the original ids).

If the replacement stored requests cannot be found, it is an error, just as if they had been specified in the original request.

Alternative implementation
====================

The ab_config rules could be set in the Account object instead, as a map of stored-request-id to ab_config. This has the advantage that all ids can be remapped before fetching so only one fetch is required to fill them:

account "Publisher"
```json
{
  "id": "Publisher",
  "disabled": false,
  "ab_config": {
    "stored-req-id-1": [ <config above> ],
    "other-stored-req-id": [ <other config> ],
    ...
  }
```

Was:
>the requirements would be something like this (implementation details aside):

- Customer can define additional secondary values for a given stored request/imp, possibly identified by a tag
- Customer can define a distribution ratio of tags for a stored request/imp, adding up to 100%. 
  For example: primary/tag1/tag2=80/15/5. By default the ratio is primary=100 and the untagged value is always chosen, as it works today.
- PBS will randomly select the primary or one of the tags respecting the ratio, for each request.
- The selection information (selected tags and distribution configuration) must be made available in the analytics object. This should not depend on the Customer adding extra fields in the stored data, but rather be a function of PBS itself.
- Extra credit: synchronize the tag selection between the stored request and all stored imps. This does impose on the Customer to define the same tags for all stored items used in a request, and creates additional handling problem for PBS when they're not all present.

laurb9:
This is a functional proof of concept for stored request A/B testing via remapping rules provided in the stored request.
It is a separate AllFetcher component intended to be placed in front of the rest of the fetchers in a composite
configuration, which means it can work for all auction endpoints and requires no changes to any of them.

If the default config is: `FetcherWithCache` → `Backend fetchers`
With A/B we will have: `ABFetcher` → `FetcherWithCache` → `Backend fetchers`

Some pieces missing from this POC are
-  fetcher unit testing until the implementation solidifies, as it is dependent on it.
- config - would be driven by a flag which goes here: https://github.com/prebid/prebid-server/compare/master...openx:ab?expand=1#diff-7abdb4226bb8e2194a2c424477150835fd9a8e6730e088af7dc5f1c88eb79b43R124
- metrics

The remapping rule would be provided in the stored request under `ext.prebid.storedrequest.ab_config` as a list of "experiment" configs:
```json
[
	{
		"code": "add_new_bidder_test_1",
		"ratio": 15.0,
		"request_id": "ABCD-0123-4567-111111", ⬅ this is the replacement stored request id
		"imp_ids": {
			"DCBA-3333-4444-012345": "DCBA-3333-4444-111111", ⬅ replacement stored imp id
			"DCBA-4444-5555-666666": "DCBA-4444-9999-222222", ⬅ replacement stored imp id
			...
		}
	},
	...
]
```

Fixes #1645

### DIFF
--- a/stored_requests/ab_fetcher.go
+++ b/stored_requests/ab_fetcher.go
@@ -12,8 +12,9 @@ import (
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
-// ABFetcher is a request-aware AllFetcher implementing the A/B experiment selection rules
+// ABFetcher is an AllFetcher implementing the A/B experiment selection rules for stored requests only
 // The results from this fetcher MUST NOT be cached for the distribution rules to work.
+// The composite fetcher layout is ABFetcher -> FetcherWithCache -> Backend Fetcher (Files, HTTP, Db)
 // This fetcher SHOULD be in front of a caching fetcher to take advantage of caching on refetch.
 type ABFetcher struct {
 	fetcher       AllFetcher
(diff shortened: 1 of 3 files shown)


stored_requests/ab_fetcher.go
```diff
@@ -0,0 +1,212 @@
...
+// triggered by ext.prebid.storedrequest.ab_config existence as a list of ABConfig objects
+func (f *ABFetcher) FetchRequests(ctx context.Context, requestIDs []string, impIDs []string) (requestData map[string]json.RawMessage, impData map[string]json.RawMessage, errs []error) {
+	if requestData, impData, errs = f.fetcher.FetchRequests(ctx, requestIDs, impIDs); len(errs) > 0 || len(requestIDs) != 1 || len(requestData) == 0 {
+		return
+	}
+	requestID := requestIDs[0]
+	storedRequest := requestData[requestID]
+
+	// Extract ab_config json if present
+	abConfigs, dataType, _, err := jsonparser.Get(storedRequest, "ext", openrtb_ext.PrebidExtKey, "storedrequest", "ab_config")
```
SyntaxNode, line 42: Interesting approach. This wouldn't work for A/B Tests defined in the request sent to Prebid Server though, would it? Looks like it only works if the stored request itself defines the A/B Test. Would we want to allow publishers to define experiments in their Prebid.js config too?

stored_requests/ab_fetcher.go
```diff
@@ -0,0 +1,212 @@
...
+	// This can be a little confusing (because the original ids are still present)
+	// but avoids rewriting the main request json to replace the ids
+	var analyticsInfo json.RawMessage
+	if value, found := newRequestData[abConfig.RequestID]; found {
+		storedRequest = value
+		// patch loaded stored request with original ab_config info to preserve it,
+		// and add ab_code for analytics
+		analyticsInfo = json.RawMessage(
+			fmt.Sprintf(`{"ext":{"%s":{"storedrequest":{"ab_config":%s,"ab_code":"%s"}}}}`,
+				openrtb_ext.PrebidExtKey, abConfigs, abConfig.Code))
```
SyntaxNode, line 95: I understand this is a POC. For production readiness, we'll need to replace hardcoded JSON strings with a proper encoder to prevent potential attack vectors, such as manipulations of the AB Code.

stored_requests/ab_fetcher.go
```diff
@@ -0,0 +1,212 @@
...
+	} else {
+		// patch original stored request with just ab_code, the ab_config is already there
+		analyticsInfo = json.RawMessage(
+			fmt.Sprintf(`{"ext":{"%s":{"storedrequest":{"ab_code":"%s"}}}}`,
+				openrtb_ext.PrebidExtKey, abConfig.Code))
+	}
+	if enrichedRequest, err := jsonpatch.MergePatch(storedRequest, analyticsInfo); err == nil {
+		requestData[requestID] = enrichedRequest
```
SyntaxNode, line 103: How expensive do you expect the MergePatch to be in terms of performance, when A/B Tests are enabled?

stored_requests/ab_fetcher.go
```diff
@@ -0,0 +1,212 @@
...
+			...
+		}
+	},
+	...
+]
+*/
+type ABConfig struct {
+	// Code is a string identifying this experiment.
+	// The control set must not be specified (its ratio inferred and it makes no replacements)
+	Code string `json:"code"`
```
SyntaxNode, line 204: Could we call this an ID instead? I think that would better follow the existing OpenRTB object model.

stored_requests/ab_fetcher.go
```diff
@@ -0,0 +1,212 @@
...
+	// Ratio is the percentage of requests that should have this experiment applied (0-100)
+	// The sum of Ratio values for all the experiments in the group must be <100
+	Ratio float64 `json:"ratio"`
```
SyntaxNode, line 207: Needing the ratios to sum to 100% might become annoying and error prone. What do you think of a "Weight" instead of a "Ratio". Where the ratio is computed as ABConfig.Weight / Sum Of All ABConfig.Weights?

It would still work the same if someone wants to sum up to 100, but otherwise would gracefully handle other situations.

stored_requests/ab_fetcher.go
```diff
@@ -0,0 +1,212 @@
...
+	// RequestID is the replacement stored request id used in this experiment
+	RequestID string `json:"request_id"`
+	// ImpIDs maps request stored imp ids to replacement stored imp ids
+	ImpIDs map[string]string `json:"imp_ids"`
+}
```
SyntaxNode, line 212: Would you consider the names StoredRequestID and StoredImpID to make it clearer these are a mapping of replacement values to use if the experiment wins?

SyntaxNode (commented):
Thank you for the proof of concept. I left a few comments. I'm also curious if you think Prebid Server should remove the A/B Test config in the requests to adapters?

stored_requests/ab_fetcher.go
laurb9: We limited the scope intentionally to just stored request A/B testing, to help with app config testing with its longer deploy timelines. Prebid.js can handle switching itself with better control that includes all bidders not just s2s.

stored_requests/ab_fetcher.go
laurb9: Good reminder.

stored_requests/ab_fetcher.go
laurb9: PBS uses this MergePatch to merges stored request, and then each stored imp, where many keys match; this won't be slower than one of those... yeah we should bench it.

I'd prefer the code to go directly in the analytics context, but staying with the confines of a fetcher this was the best I could do.

stored_requests/ab_fetcher.go
laurb9: We want to control the exact ratio for analytics. With weights, we have the opposite problem that adding a weight changes the distribution of the rest.

The 100% requirement shouldn't be a problem, because most of the time tests involve a fraction of the traffic, leaving at least 80-90% unchanged.

laurb9:
After trying it out, we think this approach is left wanting from the perspective of managing the configurations in UI or manually, and even moving the mapping into Account has the same issue. 

We've decided to put this on hold for now.

Thank you @SyntaxNode for your timely review, it was very helpful.

Issue #1171, opened 2020-01-22
Prebid Server generated bidId


bretg:
Followup to https://github.com/prebid/prebid-server/issues/1015

An assumption of the video/amp/sdk event tracking feature was that bidders generated reasonably unique IDs for each seatbid[].bid[].id. We've discovered that assumption to be false: bidders are all over the map with regards to that ID:
- some treat is as an incremental bid ID like 1,2,3,...
- some copy the request.id for each seatbid.bid
- some generate it uniquely

To be useful for joining truncated events back to the auction ID, we need to make sure that bidIds are unique, and it's become clear that bidder adapters aren't aligned with that goal.

The proposal to address this:
- Prebid Server gains a configuration that will cause it to generate a unique UUID for each successful bid and place it in seatbid[].bid[].ext.prebid.bidid.
- When creating event URLs, use this generated bidid instead of the bidder-created id
- PBJS would also be modified to look for seatbid[].bid[].ext.prebid.bidid and place it in the javascript bidResponse object for use in analytics adapters.
- The bidder-supplied bidid will be sent through as normal for those bidders who can use it to join back to their server logs.

SyntaxNode:
I don't have an objection, just some follow-up questions:

> An assumption of the video/amp/sdk event tracking feature
Would this apply to all bid requests or be restricted to the video/amp/sdk requests?

> The bidder-supplied bidid will be sent through as normal for those bidders who can use it to join back to their server logs.
To be clear, this is stating the existing bidid will still flow through the system as it does today and PBS will only place its generated uuid in extension object?

bretg:
Discussed in the Prebid Server meeting. Bryan brought up the interesting idea of using cache ID instead creating a new ID. We discussed this afterwards and while the idea has solid merit, it ties two concepts together that are not necessarily tied: bids aren’t always cached, and it could be confusing to have a fake cacheId in a scenario where we only needed the ID to be a bid. 

> Would this apply to all bid requests or be restricted to the video/amp/sdk requests?

Seems easier to just apply it to all bid requests, right? Prebid.js doesn't have to use it.

> this is stating the existing bidid will still flow through the system as it does today and PBS will only place its generated uuid in extension object?

Right.

laurb9:
Overwriting bid.id could be controlled by an adapter flag instead of an application-wide one, since this is a bidder issue.
Bidders that are known (or declared) to return unique bid ids can keep some trace continuity through PBS this way.

bretg:
[...]

PR #1772, opened 2021-03-23
Generate seatbid[].bid[].ext.prebid.bidid

VeronikaSolovei9:
Implements https://github.com/prebid/prebid-server/issues/1171


VeronikaSolovei9:
Unit tests in progress

exchange/events.go
```diff
@@ -41,6 +45,9 @@ func (ev *eventTracking) modifyBidsForEvents(seatBids map[openrtb_ext.BidderName
 	for bidderName, seatBid := range seatBids {
 		modifyingVastXMLAllowed := ev.isModifyingVASTXMLAllowed(bidderName.String())
 		for _, pbsBid := range seatBid.bids {
+			if ev.generateBidId {
+				pbsBid.generatedBidId = generateBidId()
```
laurb9, line 49: How about 
```suggestion
				pbsBid.generatedBidId = uuid.NewString()
```
using `github.com/google/uuid`

VeronikaSolovei9: @laurb9 @SyntaxNode  
Should I reuse existing `uuid.NewV4()` from "github.com/gofrs/uuid" like we do in cache (action.go)?

SyntaxNode: I don't care which library is used. Both are actively maintained and the function signatures are identical.

SyntaxNode: Either way, please consider having a bid id generator interface so it can be swapped out for predictable values for test.

exchange/events.go
```diff
@@ -62,7 +74,11 @@ func (ev *eventTracking) modifyBidVAST(pbsBid *pbsOrtbBid, bidderName openrtb_ex
 		return
 	}
 	vastXML := makeVAST(bid)
-	if newVastXML, ok := events.ModifyVastXmlString(ev.externalURL, vastXML, bid.ID, bidderName.String(), ev.accountID, ev.auctionTimestampMs); ok {
+	bidID := bid.ID
+	if ev.generateBidId {
```
SyntaxNode, line 78: Why do we need the check both here and in `modifyBidsForEvents`?

config/config.go
```diff
@@ -80,6 +80,8 @@ type Configuration struct {
 	RequestValidation RequestValidation `mapstructure:"request_validation"`
 	// When true, PBS will assign a randomly generated UUID to req.Source.TID if it is empty
 	AutoGenSourceTID bool `mapstructure:"auto_gen_source_tid"`
+	//When true, new bid id will be generated in seatbid[].bid[].ext.prebid.bidid and used in event urls instead
+	GenerateBidID bool `mapstructure:"generate_bid_id"`
```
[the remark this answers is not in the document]
laurb9: Good question. The answer could point to the best place for the config option.

I can see it as a feature independent from events. I remember @bretg mentioning how prebid.js looks for it to sub the original bid id.

This implementation does appear co-dependent on the events code, but that could be to avoid another nested iteration over the bids - it doesn't actually need events to be enabled. Another option to separate it could be at bid collection time - actually could be nice if this could be also controllable by bidder.

exchange/events.go
laurb9: how about something like this, to make it less dependent on the unrelated config flag:
```go
func getBidID(pbsBid *pbsOrtbBid) (bidID string) {
        if len(pbsBid.generatedBidId) > 0 {
                bidID = pbsBid.generatedBidId
        else {
                bidID = bid.ID
        }
        return
```

exchange/events.go
VeronikaSolovei9: @SyntaxNode interface created. Looks cleaner and now it's decoupled from events. Good point!
I. reused existing "github.com/gofrs/uuid" to generate UUID, is it ok?

config/config.go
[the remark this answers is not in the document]
VeronikaSolovei9: It's now decoupled from events, please see previous comment for details. 
I'm confused. Do you mean I need to overwrite existing bidId returned from bidder with generated bidId?

exchange/events.go
VeronikaSolovei9: Yes. Refactored.

config/config.go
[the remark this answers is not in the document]
laurb9: No need to overwrite, sorry for the misleading link. I had added it to show how prebid.js uses the ext.prebid.bidid field.

exchange/events.go
SyntaxNode: That's fine. I don't care which library we use. We've seen the gofrs work just fine at scale in Prebid Cache. Thanks for the update. This does look much nicer.

config/config.go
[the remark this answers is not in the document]
SyntaxNode: I'm not sure how it's implemented in PBS-Java. Based on this line in the spec:

> Prebid Server gains a configuration that will cause it to generate a unique UUID for each successful bid and place it in seatbid[].bid[].ext.prebid.bidid.

I believe we should assign the value in the bid collection loop. The `modifyBidsForEvents` should read the `prebid.bidid` value which would've been already populated.

> actually could be nice if this could be also controllable by bidder.

Could be. Sure. Let's keep this PR focused on a global host config and bring the per-bidder config up with the Prebid Server Committee if you think it has value (certainly may).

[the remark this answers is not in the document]
VeronikaSolovei9: Refactored

exchange/bidder.go
```diff
@@ -69,6 +70,7 @@ type pbsOrtbBid struct {
 	bidEvents         *openrtb_ext.ExtBidPrebidEvents
 	dealPriority      int
 	dealTierSatisfied bool
+	generatedBidId    string
```
SyntaxNode, line 73: Nitpick: Rename to `generatedBidID`

exchange/events.go
```diff
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"time"
 
-	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/evanphx/json-patch"
```
SyntaxNode, line 7: I didn't expect this change. Was this automatically done by GoLand?

exchange/exchange.go
```diff
@@ -79,6 +81,24 @@ type bidResponseWrapper struct {
 	bidder       openrtb_ext.BidderName
 }
 
+type UUIDGenerator interface {
```
SyntaxNode, line 84: Nitpick: Please change the name to be specific to BidID, such as `BidIDGenerator`.

exchange/exchange.go
```diff
@@ -79,6 +81,24 @@ type bidResponseWrapper struct {
...
+	NewUUID() string
```
SyntaxNode, line 85: Nitpick: This can just be `New`. The fact that's it a UUID is an implementation detail.

exchange/exchange.go
```diff
@@ -79,6 +81,24 @@ type bidResponseWrapper struct {
...
+type bidIDGenerator struct {
+	enabled bool
+}
+
+func (big *bidIDGenerator) Enabled() bool {
+	return big.enabled
+}
+
+func (big *bidIDGenerator) NewUUID() string {
+	rawUuid, _ := uuid.NewV4()
```
SyntaxNode, line 98: Please don't ignore errors. If it's possible to have an error from `uuid.NewV4()`, we'll need to bubble that up the call chain.

exchange/exchangetest/append-bidder-names.json
```diff
@@ -137,6 +138,7 @@
               "crid": "creative-3",
               "ext": {
                 "prebid": {
+                  "bidid": "mock_uuid",
                   "targeting": {
                     "hb_bidder": "appnexus",
                     "hb_bidder_appnexus": "appnexus",
```
SyntaxNode, line 144: Thoughts on if this should be disabled by default on the exchange tests to avoid changing all of them?

exchange/exchange.go
```diff
@@ -79,6 +81,24 @@ type bidResponseWrapper struct {
 	bidder       openrtb_ext.BidderName
 }
 
+type UUIDGenerator interface {
+	NewUUID() string
+	Enabled() bool
```
mansinahar, line 86: What's the rationale behind creating this interface? Why not just add a `generateBidID` flag to the `exchange` object and do a check directly like: `if e.GenerateBidID` on L216 and if it is then directly call `uuid.NewV4().String()`?

If the purpose of this is for unit test coverage then I am wondering if we need a mock ID generator for the unit tests? Can't we just assert that the `GeneratedBidID` value is not an empty string?

exchange/events.go
```diff
@@ -62,7 +62,11 @@ func (ev *eventTracking) modifyBidVAST(pbsBid *pbsOrtbBid, bidderName openrtb_ex
 		return
 	}
 	vastXML := makeVAST(bid)
```
mansinahar, line 64: I understand that unit tests for this function already don't exist but do you think we can incorporate them as part of this PR if it's not a lot of effort?

config/config.go
mansinahar, line 84: Can you please add assertions for this new flag in the `TestDefaults` and `TestFullConfig` methods in `configs_test.go`?

exchange/exchangetest/append-bidder-names.json
VeronikaSolovei9: Default is false. Reason why it's in all tests is because inserting bidID logic is placed directly in HoldAction. So I modified `newExchangeForTests` function and hardcoded mock bidId generator to true and this applies to all tests in exchange test. This unit test is the only one good place to test it.

exchange/exchange.go
VeronikaSolovei9: This is how it was initially, but then I modified it to interface due to this comment: https://github.com/prebid/prebid-server/pull/1772#discussion_r599699867

exchange/events.go
VeronikaSolovei9: There is no specific unit test for this function, but it's covered by these test cases as a part of `TestExchangeJSON`: 
exchange/exchangetest/events-vast-account-off-request-on.json
exchange/exchangetest/events-vast-account-on-request-off.json
There are some other functions like this: `makeEventURL`, `modifyBidVAST`, `ModifyVastXmlString`

config/config.go
VeronikaSolovei9: Good point, added!

exchange/bidder.go
VeronikaSolovei9: Renamed

exchange/events.go
VeronikaSolovei9: Probably go fmt or IntellijIdea autofix. It works without it. Should I revert it back?

exchange/exchange.go

exchange/exchange.go

exchange/exchange.go
VeronikaSolovei9: Yes, added

exchange/exchangetest/append-bidder-names.json
VeronikaSolovei9: Added separate unit test for this.

VeronikaSolovei9: Also reverted back json files and added new for new tests

exchange/events.go
mansinahar: I see. Thanks! Honestly, it'd be good to have some unit test coverage for these as well but that's out of scope for this PR

exchange/events_test.go
```diff
@@ -71,6 +80,7 @@ func Test_eventsData_modifyBidJSON(t *testing.T) {
 		enabledForAccount bool
 		enabledForRequest bool
 		bidType           openrtb_ext.BidType
+		generateBidId     string
```
mansinahar, line 83: Can you please rename this to `generatedBidID` as `generateBidId` indicates the `generateBidId` boolean flag rather than the generated BidID value?

exchange/exchange_test.go
```diff
@@ -1440,7 +1442,7 @@ func buildImpExt(t *testing.T, jsonFilename string) json.RawMessage {
 	return json.RawMessage(toReturn)
 }
 
-func TestPanicRecoveryHighLevel(t *testing.T) {
+func NoTestPanicRecoveryHighLevel(t *testing.T) {
```
mansinahar, line 1445: What's the reason behind changing this test name to have `No` in front?

exchange/exchange_test.go
```diff
@@ -1585,6 +1587,53 @@ func TestExchangeJSON(t *testing.T) {
...
+}
+
+func TestInvalidBidID(t *testing.T) {
+	fileName := "./test/bidid/invalid-bid-id.json"
+	bid, spec := testBidIdHelper(t, fileName, true)
+	diffJson(t, "bid.SeatBid[0].Bid[0].Ext modified", bid.SeatBid[0].Bid[0].Ext, spec.Response.Bids.SeatBid[0].Bid[0].Ext)
+	diffJson(t, "bid.Ext info modified", bid.Ext, spec.Response.Ext)
+}
+
+func testBidIdHelper(t *testing.T, fileName string, bidIDError bool) (*openrtb2.BidResponse, *exchangeSpec) {
```
mansinahar, line 1603: Why not use the existing JSON test framework for these tests as well? You can just enable/disable generating Bid ID by setting the `generatBidID` flag in the JSON test it self. See this debug log disabled test for example: https://github.com/prebid/prebid-server/blob/master/exchange/exchangetest/debuglog_disabled.json

exchange/events_test.go
```diff
@@ -13,6 +13,7 @@ func Test_eventsData_makeBidExtEvents(t *testing.T) {
 		enabledForAccount bool
 		enabledForRequest bool
 		bidType           openrtb_ext.BidType
+		generatedBidId    string
```
mansinahar, line 16: **Nitpick**: Can you please update this to `generatedBidID` as well for consistency?

config/config_test.go
```diff
@@ -415,6 +416,7 @@ func TestFullConfig(t *testing.T) {
 	cmpStrings(t, "request_validation.ipv4_private_networks", cfg.RequestValidation.IPv4PrivateNetworks[0], "1.1.1.0/24")
 	cmpStrings(t, "request_validation.ipv6_private_networks", cfg.RequestValidation.IPv6PrivateNetworks[0], "1111::/16")
 	cmpStrings(t, "request_validation.ipv6_private_networks", cfg.RequestValidation.IPv6PrivateNetworks[1], "2222::/16")
+	cmpBools(t, "generate_bid_id", cfg.GenerateBidID, false)
```
mansinahar, line 419: Would suggest setting this as `true` in the above `fullConfig` variable and asserting it here to `true` also as this test is making sure the config is read correctly and false is anyway the default

exchange/exchange_test.go
VeronikaSolovei9: Oh sorry, I just wanted to disable it temporarily, I'll fix it back!

config/config_test.go
VeronikaSolovei9: Done, good point!

exchange/events_test.go
VeronikaSolovei9: Yes, done!

exchange/exchange_test.go
VeronikaSolovei9: This a genius thing I didn't know about! Modified!

mansinahar (dismissed):

SyntaxNode (dismissed):

exchange/events.go
```diff
@@ -62,7 +62,11 @@ func (ev *eventTracking) modifyBidVAST(pbsBid *pbsOrtbBid, bidderName openrtb_ex
...
-	if newVastXML, ok := events.ModifyVastXmlString(ev.externalURL, vastXML, bid.ID, bidderName.String(), ev.accountID, ev.auctionTimestampMs); ok {
+	bidID := bid.ID
+	if len(pbsBid.generatedBidID) > 0 {
+		bidID = pbsBid.generatedBidID
```
SyntaxNode, line 67: Could you please add test coverage for this condition? Look to `events-vast-account-off-request-on` as an example, but just focus on the bid id generation.

exchange/exchange_test.go
```diff
@@ -1772,7 +1777,27 @@ func newExchangeForTests(t *testing.T, filename string, expectations map[string]
 		categoriesFetcher:   categoriesFetcher,
 		bidderInfo:          bidderInfos,
 		externalURL:         "http://localhost",
+		bidIDGenerator:      bidIDGenerator,
+	}
+}
+
+type mockBidIDGenerator struct {
+	GenerateBidID bool
```
SyntaxNode, line 1785: Nitpick: Please add json attributes to use `generateBidID` and `returnError` in the test files for naming consistency.

exchange/exchange_test.go
```diff
@@ -1772,7 +1777,27 @@ func newExchangeForTests(t *testing.T, filename string, expectations map[string]
...
 		bidderInfo:          bidderInfos,
 		externalURL:         "http://localhost",
+		bidIDGenerator:      bidIDGenerator,
+	}
+}
+
+type mockBidIDGenerator struct {
+	GenerateBidID bool `json:"GenerateBidID"`
+	ReturnError   bool `json:"ReturnError"`
+}
```
SyntaxNode, line 1787: I had intended for the json names to use a lower case starting letter to match the rest of our existing json names:
```
type mockBidIDGenerator struct {
	GenerateBidID bool `json:"generateBidID"`
	ReturnError   bool `json:"returnError"`
}
```

VeronikaSolovei9: Oh I see, will do!

mansinahar (approved):

exchange/exchange_test.go
SyntaxNode: Thank you.

SyntaxNode (approved):

PR #1774, opened 2021-03-24
New Adapter: Zemanta

rokostik:
Hey!

This PR adds the Zemanta adapter.

- contact email of the adapter’s maintainer: prog-ops-team@outbrain.com

- docs repo PR prebid/prebid.github.io#2796


SyntaxNode:
We recently merged a PR which upgrades the openrtb package. Please merge with the master branch and change this import to `"github.com/mxmCherry/openrtb/v14/openrtb2"`. You will also need to change the code to use the new `openrtb2` package name instead of the previous `openrtb`.

adapters/zemanta/zemanta.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		imp.TagID = zemantaExt.TagId
+		reqCopy.Imp[i] = imp
+	}
+
+	publisher := &openrtb.Publisher{
+		ID:     zemantaExt.Publisher.Id,
+		Name:   zemantaExt.Publisher.Name,
+		Domain: zemantaExt.Publisher.Domain,
+	}
```
SyntaxNode, line 52: Why does this adapter need to define it's own publisher object, instead of using what's provided on the request. Same question for bcat and badv.

adapters/zemanta/zemanta.go
```diff
@@ -0,0 +1,147 @@
...
+		if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		if err := json.Unmarshal(bidderExt.Bidder, &zemantaExt); err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		imp.TagID = zemantaExt.TagId
+		reqCopy.Imp[i] = imp
```
SyntaxNode, line 45: It looks to me that you don't want to modify the original request, since you're making modifications on a shallow `reqCopy` copy. Not a problem.

I'd like to point out this code is modifying the original request at this line. The `reqCopy.Imp` is a Go Slice which uses the same underlying memory as the original `request.Imp`. The `Imp` copy is unique to each adapter, so there's no issue. I just wanted you to be aware. If you'd like to leave the original request alone, you'll want to clone the `reqCopy.Imp` slice before the loop.

adapters/zemanta/zemanta.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(request.Imp))
+	bidResponse.Currency = response.Cur
+
+	var errs []error
+	for _, seatBid := range response.SeatBid {
+		for _, bid := range seatBid.Bid {
+			bidType, err := getMediaTypeForImp(bid.ImpID, request.Imp)
```
SyntaxNode, line 115: There's a subtle bug in this loop with taking a reference of the bid loop iterator value. Please consider this alternative:

```
	for i := range seatBid.Bid {
		bid := seatbid.Bid[i]
		
		bidType, err := getMediaTypeForImp(bid.ImpID, request.Imp)

		if err != nil {
			errs = append(errs, err)
			continue
		}

		b := &adapters.TypedBid{
			Bid:     &bid,
			BidType: bidType,
		}
	}
}
```

adapters/zemanta/zemanta.go
```diff
@@ -0,0 +1,147 @@
...
+func getMediaTypeForImp(impID string, imps []openrtb.Imp) (openrtb_ext.BidType, error) {
+	mediaType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner == nil {
+				mediaType = openrtb_ext.BidTypeNative
+			}
+			return mediaType, nil
+		}
+	}
```
SyntaxNode, line 141: This is an anti-pattern. Does your bidding server provide metadata in the response describing if the bid is a native or banner ad?

adapters/zemanta/zemanta.go
rokostik: This is mostly regarding support for s2s in Prebid.js. We already support the publisher param in our Prebid.js and are now adding support for bcat and badv in https://github.com/prebid/Prebid.js/pull/6468. Is there any other way to support these params from Prebid.js? AFAIK they come from the bidder ext field in the request.

adapters/zemanta/zemanta.go
rokostik: Unfortunately no, we don't send any info about the media type in the response.

adapters/zemanta/zemanta.go
rokostik: Isn't it prohibited to change the original request? 
The documentation in https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html states:
> If your adapter needs to alter any data referenced by a pointer then you must first make a shallow copy. The only exception is for request.Imp and its elements, as these are already shallow copies.

adapters/zemanta/zemanta.go
rokostik: Oof thanks for catching this, learned something new :) will fix.

adapters/zemanta/zemanta.go
SyntaxNode: The request provided to your adapter is already a shallow copy. You may change anything at the root level of the request or anything at the root level of an imp. If you want to change a variable within a pointer, then that's shared memory and you need to make a copy first.. as you're doing correctly with site and app.

The consequence is the request given to MakeRequests is also given to MakeBids, so if you modify the request in MakeRequests then MakeBids will work off the modified request, and that is often times ok.

I'll update the doc to explain that the request is also a shallow copy. Thank you for pointing out that inconsistency.

adapters/zemanta/zemanta.go
SyntaxNode: Gotcha. Does this logic best mimic the behavior of your server? Will it prefer banner over native?

rokostik: Thanks for pointing this out, it actually wasn't. Fixed now.

adapters/zemanta/zemanta.go
SyntaxNode: I'm not familiar with the details of the Prebid.js interaction in this case. @AlexBVolcy is investigating. Does Prebid.js not allow publishers to specify request level `badv` and `bcat`?

AlexBVolcy: So after some investigating, it seems that this work around that you have implemented is necessary for the time being! There's nothing currently in the works with Prebid.js that allows for a more general way of supporting `badv` and `bcat`. Hopefully in the future this changes!

bretg:
Discussed in Prebid Server committee last week

- It's possible to supply bcat/badv in the new Prebid.js ortb2 config, but the pbsBidAdapter doesn't forward anything but site and user. Issue https://github.com/prebid/Prebid.js/issues/6528 tracks this enhancement.
- But even when that's done, it's not ideal to specify badv in Prebid.js because A) it can be a large list and B) it can differ per bidder. So we need to continue to allow bid adapters to specify these values either as bidder-specific parameters or in their server-side config.
- I don't see a problem with zemanta overwriting site.publisher.id/name with their values.  Not clear they need to ask the pub to supply the domain since PBS supplies that, but I suppose it's their call. Maybe there are scenarios where the pub domain they require is different than the page domain?

Bottom line: this seems ok to me. Someday a bcat/badv module can inject this stuff from host-company account config.

SyntaxNode:
> * I don't see a problem with zemanta overwriting site.publisher.id/name with their values.  Not clear they need to ask the pub to supply the domain since PBS supplies that, but I suppose it's their call. Maybe there are scenarios where the pub domain they require is different than the page domain?

@rokostik Look like we're good with the badv/bcat section. What's your take on the publisher section? re: Bret's comment.

rokostik:
Hey @SyntaxNode and @bretg thanks for investigating this. Regarding the publisher domain: prebid.js server adapter only sends the publisher id in the request [here](https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L311). What did you mean with "since PBS supplies that"? Maybe I'm misunderstanding something here.

hhhjort:
PBS will fill in some data into the request from the request context. See the function `setFieldsImplicitly()` in exchange/auction.go, and more specifically `setSiteImplicitly()`. This fills in the site page/domain from the referrer URL if the referrer exists, and the request does not already supply the information.

rokostik:
@hhhjort thanks for explaining. Since PBS only fills the site domain I suggest we leave our adapter publisher object as is. Do you agree @SyntaxNode?

SyntaxNode (approved):

AlexBVolcy (approved):

PR #1777, opened 2021-03-25
Tappx changes - Backward compatible change of version

prebidtappx:
Hi! This is a backward compatible change for a future feature that our IT team is developing for our partners.
We are going to improve our endpoint.
Thanks for all.


prebidtappx:
Ok, we have to modify tappx_test for this change.

prebidtappx:
For our side all is ok! :)

adapters/tappx/tappx.go
```diff
@@ -126,7 +127,9 @@ func (a *TappxAdapter) buildEndpointURL(params *openrtb_ext.ExtImpTappx, test in
 		}
 	}
 
-	thisURI.Path += params.Endpoint
+	if !(strings.Contains(strings.ToLower(thisURI.Host), strings.ToLower(params.Endpoint))) {
```
SyntaxNode, line 130: Could you please add a new json test for this condition?

prebidtappx: Hi! I have add now the commit of a new json test that is think for this condition.
Also, the idea is when all our partners change to new system, we will update all tests and remove the lines (130-132) that don't be needed more.
Thanks for all!

SyntaxNode: Nice. Thank you.

SyntaxNode (approved):

hhhjort (approved):
Not necessarily for this PR, but would be nice to get some validation on the host of your URL. As it stands, a malicious actor could craft a PBS request with any arbitrary hostname, and have a PBS host send a Tappx bid request to that hostname. PBS is not a good proxy for a DOS attack, but would be nice to close up that possibility.

SyntaxNode:
@hhhjort We have an email chain started for that issue. I owe a response to Tappx on that issue to share my initial thoughts and to see what they recommend.

PR #1778, opened 2021-03-25
DMX: Enforcing w and h in imp

stevealliance:
Include DMX Ortb rule for imp.w and imp.h to be present on each request


adapters/dmx/dmx.go
```diff
@@ -100,6 +100,13 @@ func (adapter *DmxAdapter) MakeRequests(request *openrtb.BidRequest, req *adapte
 		if dmxReq.App.ID != "" {
 			anyHasId = true
 		}
+		if anyHasId == false {
+			idfa, valid := getIdfa(request)
```
SyntaxNode, line 104: Nitpick: May be simplified as:
```
if idfa, valid := getIdfa(request); valid {
  dmxReq.App.ID = idfa
  anyHasId = true
}
```

adapters/dmx/dmx.go
```diff
@@ -186,6 +189,9 @@ func (adapter *DmxAdapter) MakeRequests(request *openrtb.BidRequest, req *adapte
 	}
 
 	dmxReq.Imp = imps
+	if anyHasId == false {
```
SyntaxNode, line 192: Nitpick: Please consider adding a line break before this condition for readability.

adapters/dmx/dmx.go
```diff
@@ -271,10 +278,15 @@ func fetchParams(params dmxExt, inst openrtb.Imp, ins openrtb.Imp, imps []openrt
 		tempimp.Secure = &intVal
 	}
 	if banner != nil {
+		if banner.H == checkValue || banner.W == checkValue {
```
SyntaxNode, line 281: I'm unclear what you're looking to accomplish with this check. The `checkValue` is never initialized, so isn't this the same as `if banner.W == nil || banner.W == nil {`?

adapters/dmx/dmx.go
```diff
@@ -271,10 +278,15 @@ func fetchParams(params dmxExt, inst openrtb.Imp, ins openrtb.Imp, imps []openrt
...
+			banner.H = &banner.Format[0].H
```
SyntaxNode, line 282: Banner is shared memory. You need to make a copy of it first before modifying.
```
bannerCopy := *banner
bannerCopy.H = &banner.Format[0].H
bannerCopy.W = &banner.Format[0].W
banner = &bannerCopy
```

adapters/dmx/dmx.go
```diff
@@ -257,6 +263,7 @@ func (adapter *DmxAdapter) MakeBids(request *openrtb.BidRequest, externalRequest
 
 func fetchParams(params dmxExt, inst openrtb.Imp, ins openrtb.Imp, imps []openrtb.Imp, banner *openrtb.Banner, video *openrtb.Video, intVal int8) []openrtb.Imp {
```
SyntaxNode, line 264: While we're here, `ins openrtb.Imp` is unused. Please remove this argument.

adapters/dmx/dmx.go
```diff
@@ -257,6 +263,7 @@ func (adapter *DmxAdapter) MakeBids(request *openrtb.BidRequest, externalRequest
...
 	var tempimp openrtb.Imp
+	var checkValue *uint64
 	tempimp = inst
```
SyntaxNode, line 267: While we're here, this assignment isn't necessary. `inst` is already a copy of the imp since it's passed by value. Please consider using `inst` directly.

adapters/dmx/dmx.go
```diff
@@ -271,10 +278,15 @@ func fetchParams(params dmxExt, inst openrtb.Imp, ins openrtb.Imp, imps []openrt
...
 	if banner != nil {
+		if banner.H == checkValue || banner.W == checkValue {
+			banner.H = &banner.Format[0].H
+			banner.W = &banner.Format[0].W
+		}
 		tempimp.Banner = banner
 	}
 
 	if video != nil {
+		video.Protocols = checkProtocols(video)
```
SyntaxNode, line 289: Video is shared memory too:
```
videoCopy := * videp
videoCopy.Protocols = checkProtocols(video)
tempimp.Video = &video
```

adapters/dmx/dmx.go
```diff
@@ -327,3 +339,17 @@ func isDmxParams(t interface{}) bool {
...
 	}
 }
+
+func getIdfa(request *openrtb.BidRequest) (string, bool) {
+	device := request.Device
+	if device.IFA != "" {
+		return device.IFA, true
+	}
+	return "empty", false
+}
```
SyntaxNode, line 349: It's possible for `Device` to be `nil`. Please add a protection check. There's also no need to use the literal "empty" since it's discarded. Consider the following:
```
func getIDFA(request *openrtb.BidRequest) (string, bool) {
  if request.Device == nil {
    return nil, false
  }

  return request.Device.IFA, request.Device.IFA != ""
}
```

adapters/dmx/dmx.go
```diff
@@ -327,3 +339,17 @@ func isDmxParams(t interface{}) bool {
...
+func checkProtocols(imp *openrtb.Video) []openrtb.Protocol {
+	if len(imp.Protocols) > 0 {
+		return imp.Protocols
+	}
+	return []openrtb.Protocol{2, 3, 5, 6, 7, 8}
```
SyntaxNode, line 354: Small Optimization: Please consider defining the protocol list at the package level. It's safe because it's read only within the adapter. This will slightly reduce memory overheard and slightly improve performance.

adapters/dmx/dmx.go
stevealliance: Forgot it was a shared in-memory

adapters/dmx/dmx.go
stevealliance: Forgot it was a shared in-memory

adapters/dmx/dmx.go
```diff
@@ -100,6 +102,12 @@ func (adapter *DmxAdapter) MakeRequests(request *openrtb.BidRequest, req *adapte
 		if dmxReq.App.ID != "" {
 			anyHasId = true
 		}
+		if anyHasId == false {
+			if idfa, valid := getIdfa(request); valid {
```
SyntaxNode, line 106: Looks good. Please add a json test for the `idfa, valid := getIdfa(request); valid` condition.

adapters/dmx/dmx.go
```diff
@@ -271,10 +277,15 @@ func fetchParams(params dmxExt, inst openrtb.Imp, ins openrtb.Imp, imps []openrt
 		tempimp.Secure = &intVal
 	}
 	if banner != nil {
+		if banner.H == nil || banner.W == nil {
+			banner.H = &banner.Format[0].H
+			banner.W = &banner.Format[0].W
+		}
```
SyntaxNode, line 283: Please add json test coverage for the `if banner.H == nil || banner.W == nil` condition.

SyntaxNode (commented):
Changes look good. Thank you. Please add test coverage using json files as the next step.

SyntaxNode:
We recently merged a PR which upgrades the openrtb package. Please merge with the master branch and change this import to `"github.com/mxmCherry/openrtb/v14/openrtb2"`. You will also need to change the code to use the new `openrtb2` package name instead of the previous `openrtb`.

hhhjort (approved):
LGTM

SyntaxNode (approved):

Issue #1776, opened 2021-03-25
Authorization leak through headers


bjorn-lw:
If "test" is specified in the call to PBS, the debug info coming back contains all headers sent to the endpoints.

e.g. There can be a basic authentication key exposed. Even though it's not the actual user/password it feels like not something that should be publicly exposed.

bretg:
Thanks for the bug report @bjorn-lw

SyntaxNode:
Good catch. We missed that the Rubicon adapter makes use of the authorization header.

Are you ok with us keeping the header debugging feature and replacing the content of the auth header with a literal `<HIDDEN BY PREBID SERVER>` value?

PR #1779, opened 2021-03-25
Remove Authorization Headers From Debug Response

SyntaxNode:
Fixes https://github.com/prebid/prebid-server/issues/1776


hhhjort (approved):
LGTM

exchange/bidder.go
```diff
@@ -327,23 +327,39 @@ func getAssetByID(id int64, assets []nativeRequests.Asset) (nativeRequests.Asset
...
 		}
 	}
+
+	return &openrtb_ext.ExtHttpCall{}
+}
+
+var authorizationHeader = http.CanonicalHeaderKey("authorization")
+
+func filterHeader(h http.Header) map[string][]string {
+	clone := h.Clone()
```
mansinahar, line 360: Wondering if we need to clone the http header map here? Do we expect the authorization header to be used anywhere else outside of this function? If so then it makes sense to not modify the original object but if not then we can just delete it from the original object

exchange/bidder.go
```diff
@@ -327,23 +327,39 @@ func getAssetByID(id int64, assets []nativeRequests.Asset) (nativeRequests.Asset
...
+				RequestBody:    string(httpInfo.request.Body),
+				RequestHeaders: filterHeader(httpInfo.request.Headers),
+			}
 		}
 	}
+
+	return &openrtb_ext.ExtHttpCall{}
+}
+
+var authorizationHeader = http.CanonicalHeaderKey("authorization")
```
mansinahar, line 357: We should make this a constant variable as we don't expect any code to change it

exchange/bidder.go
mansinahar: I checked the code and I don't see any other usage of the Authorization header so it should be safe to remove it from the original object here

exchange/bidder.go
SyntaxNode: In Go, constants must be a hardcoded value. The function call to `CanonicalHeaderKey` prevents me from marking it as a constant.

mansinahar (approved):

hhhjort (approved):

PR #1782, opened 2021-03-29
Basic GDPR enforcement for specific publisher-vendors.

### LINKED ISSUE #1750: Add conditional GDPR Vendor override for publishers.
Publishers may have out of band reasons to allow certain user's requests to bypass the vendor checks. They or their affiliates may have special contracts with technology providers (bidders) that keeps all user data siloed as the property of the publisher, rather than the bidder. Or they may have other signals and conditions in place to believe consent for a vendor is in place that is not properly reflected in the consent string. This issue proposes a signal in the request to bypass the consent checking on the GVL side of the equation in the request.

The signal in the request would look something like this:

`{ "ext":{ "prebid": {"skip_gvl_bidders": [ "bidder1", "bidder2" ]
 }}}
`
In the GDPR processing, checks for that bidder would essentially drop to basic enforcement, and skip the check for the user consenting to the bidder, disabling that bidder from GDPR enforcement just for the context of this one request.

hhhjort:
Resolves #1750


endpoints/openrtb2/amp_auction.go
```diff
@@ -191,7 +191,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 		LegacyLabels: labels,
 	}
 
-	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
+	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil, account)
```
SyntaxNode, line 194: The account is already available in the auctionRequest. Could you please use that instead of adding a new parameter?

gdpr/impl.go
```diff
@@ -74,7 +79,16 @@ func (p *permissionsImpl) PersonalInfoAllowed(ctx context.Context, bidder openrt
...
+		basicEnforce := false
+		if account != nil {
+			for _, vendor := range account.GDPR.BasicEnforcementVendors {
+				if vendor == string(bidder) {
+					basicEnforce = true
+					break
+				}
+			}
+		}
+		return p.allowPI(ctx, id, consent, basicEnforce)
```
SyntaxNode, line 91: I recommend passing in a `basicEnforce` flag as a parameter and moving the vendor lookup logic up a level. This will make it easier for us to add a global level basic mode.

endpoints/openrtb2/amp_auction.go
hhhjort: Oops, I missed that. Thanks.

gdpr/impl.go
hhhjort: Seems reasonable.

SyntaxNode (commented):
Looks good to me. @bsardo could you please verify this follows the definition of basic enforcements as defined in the spec? 

Spec: https://docs.google.com/document/d/1fBRaodKifv1pYsWY3ia-9K96VHUjd8kKvxZlOsozm8E/edit#heading=h.3kby792a2sqy

gdpr/impl.go
```diff
@@ -194,16 +199,16 @@ const pubRestrictNotAllowed = 0
...
 		return p.cfg.TCF2.PurposeOneTreatment.AccessAllowed
 	}
 	if consent.CheckPubRestriction(uint8(purpose), pubRestrictNotAllowed, vendorID) {
 		return false
 	}
 
-	purposeAllowed := vendor.Purpose(purpose) && consent.PurposeAllowed(purpose) && consent.VendorConsent(vendorID)
-	legitInterest := vendor.LegitimateInterest(purpose) && consent.PurposeLITransparency(purpose) && consent.VendorLegitInterest(vendorID)
+	purposeAllowed := consent.PurposeAllowed(purpose) && (basicEnforce || (vendor.Purpose(purpose) && consent.VendorConsent(vendorID)))
+	legitInterest := consent.PurposeLITransparency(purpose) && (basicEnforce || (vendor.LegitimateInterest(purpose) && consent.VendorLegitInterest(vendorID)))
```
bsardo, line 211: Should these be:
```
purposeAllowed := consent.PurposeAllowed(purpose) && consent.VendorConsent(vendorID) && (basicEnforce || (vendor.Purpose(purpose)))
legitInterest := consent.PurposeLITransparency(purpose) && consent.VendorLegitInterest(vendorID) && (basicEnforce || vendor.LegitimateInterest(purpose))
```
The basic enforcement requirements indicate if we are enforcing the purpose and the vendor, then the user needs to consent to both. My understanding is we are always enforcing vendor consent; I don't see a config flag to disable it. In that case, my code snippet requires `consent.VendorConsent(vendorID)` return true rather than ORing it with the `baseEnforce` flag.

bsardo: As @bretg pointed out in #1750, we don't want to enforce any of the per-vendor checks and the linked spec will be updated to reflect this. This is good as is.

bsardo (dismissed):

bsardo (approved):
LGTM

SyntaxNode (approved):

Issue #1783, opened 2021-03-29
usersync_if_ambiguous -> default_value. require default_value to be configured by host.


Jobeso:
Hey Prebid Team,
I updated our self hosted prebid-server from 0.138.0 to 0.153.0. Since that change I see an impression drop of 80%. Rolling back this update seems to be a temporary fix.

The drop is not os related and is an issue on iOS and Android. Also, it's not EEA related, also happens for countries outside of the EU. All bidders are effected except for OpenX.

I checked the changelogs between the versions but didn't see any breaking changes. Did I miss anything there? How would you approach debugging this?

Thanks in advance for your help.

SyntaxNode:
HI @Jobeso,

That's no good. We didn't intend to introduce a breaking change, but it seems something slipped by us. Sorry for that. Let's figure it out and fix it.

These sort of issues are almost always related to privacy policies. We tweaked some of the GDPR settings at the start of the year to better align with the specs. I suspect one of these changes had unintended side affects.

To start, what are your GDPR settings? Specifically: `gdpr.enabled`, `gdpr.host_vendor_id`, and `gdpr.usersync_if_ambiguous`.

Jobeso:
Hey @SyntaxNode,
thanks for the quick answer. I suspect that whatever is breaking might only be for our setup so I suppose rather a mistake from our side.

We run a rather basic setup with no huge modifications of the default values. This is true for your above mentioned flags as well, we do not set them explicitly, default values seem to be the following and they did not change from 0.138.0 to 0.153.0.

```
gdpr.enabled: true
gdpr.host_vendor_id: 0
gdpr.usersync_if_ambiguous: false
```

It's a bit hard for me to find documentation on these to understand what's missing from our end. I found your [comment](https://github.com/prebid/prebid-server/pull/1775#issuecomment-808525543) and this [discussion](https://github.com/prebid/prebid-server/issues/1393) that explains this kinda but I'm still struggling with understanding what we should change here. 

With our prebid-server we only enable bidding for our own app (iOS & Android). We gather the consent with a cmp and the consent string is automatically added with prebid-mobile to the requests.

I really appreciate the work you and your team are doing and thanks for your help with this issue.

SyntaxNode:
Thank you sharing your GDPR configuration. I've identified the issue as an unintended side effect of [this change](https://github.com/prebid/prebid-server/pull/1656).

[...]

PR #1786, opened 2021-03-31
Flip usersync_if_ambiguous Default

SyntaxNode:
Addresses: https://github.com/prebid/prebid-server/issues/1783

The `usersync_if_ambiguous` forces TCF enforcement on any bid request this doesn't explicitly set the `regs.ext.gdpr` field. This is not a good default and has caused several complaints. Setting this to `true` is a better default value which only enforces TCF if the bid request includes `regs.ext.gdpr` set to `1`.

We'll follow-up with a separate PR to rename the `usersync_if_ambiguous` config value to something more descriptive, as it controls much more than user sync.


SyntaxNode:
This proposed change would introduce too much risk in GDPR enforcement. I'll discuss further options on the linked issue.

PR #1790, opened 2021-04-05
Fix shared memory issue when stripping authorization header from bid requests

bsardo:


exchange/bidder_test.go
```diff
@@ -141,48 +141,6 @@ func TestSingleBidder(t *testing.T) {
 	}
 }
 
-func TestRequestBidRemovesSensitiveHeaders(t *testing.T) {
```
SyntaxNode, line 144: I recommend to keep this end to end test.

exchange/bidder_test.go
```diff
@@ -974,6 +932,28 @@ func TestMakeExt(t *testing.T) {
 				Status:         999,
 			},
 		},
```
SyntaxNode, line 934: Could you please keep the original `"Request & Response - No Error"` test in addition to the new `"Request & Response - No Error with Authorization removal"` test? I think it's useful to track down possible implementation errors.

exchange/bidder_test.go
```diff
@@ -974,6 +932,28 @@ func TestMakeExt(t *testing.T) {
...
+					StatusCode: 999,
+				},
+			},
+			expected: &openrtb_ext.ExtHttpCall{
+				Uri:            "requestUri",
+				RequestBody:    "requestBody",
+				RequestHeaders: nil,
+				ResponseBody:   "responseBody",
+				Status:         999,
+			},
```
SyntaxNode, line 955: Nice. Thank you for adding this test.

exchange/bidder_test.go
bsardo: Oops yeah that was my intention. I had temporarily removed this test and was trying to add something similar that verified the shared memory error was resolved by making multiple bid requests but that seems to be more effort than it's worth so I'll just add this back in.

SyntaxNode: Sounds good. I wouldn't bother to try to reproduce the shared memory conflict either. It's just too hard to simulate reliably.

SyntaxNode (approved):

mansinahar (approved):

PR #1792, opened 2021-04-07
AppNexus: Make Ad Pod Id Optional

VeronikaSolovei9:
Docs PR: https://github.com/prebid/prebid.github.io/pull/2836

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
@@ -291,7 +292,6 @@ func (a *AppNexusAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder 
 func (a *AppNexusAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
 	memberIds := make(map[string]bool)
 	errs := make([]error, 0, len(request.Imp))
-	adPodId := false
 
 	// AppNexus openrtb2 endpoint expects imp.displaymanagerver to be populated, but some SDKs will put it in imp.ext.prebid instead
 	var defaultDisplayManagerVer string
@@ -302,12 +302,20 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *ad
 			defaultDisplayManagerVer = fmt.Sprintf("%s-%s", source, version)
 		}
 	}
+	var adPodId *bool
+
 	for i := 0; i < len(request.Imp); i++ {
 		memberId, impAdPodId, err := preprocess(&request.Imp[i], defaultDisplayManagerVer)
 		if memberId != "" {
 			memberIds[memberId] = true
 		}
-		adPodId = adPodId || impAdPodId
+		if adPodId == nil {
+			adPodId = &impAdPodId
+		} else if *adPodId != impAdPodId {
+			errs = append(errs, errors.New("generate ad pod option should be same for all pods in request"))
+			return nil, errs
+		}
+
 		// If the preprocessing failed, the server won't be able to bid on this Imp. Delete it, and note the error.
 		if err != nil {
 			errs = append(errs, err)
@@ -370,7 +378,7 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *ad
 	// For this all impressions in  request should belong to the same pod
 	// If impressions number per pod is more than maxImpsPerReq - divide those imps to several requests but keep pod id the same
 	// If  adpodId feature disabled and impressions number per pod is more than maxImpsPerReq  - divide those imps to several requests but do not include ad pod id
-	if isVIDEO == 1 && adPodId {
+	if isVIDEO == 1 && *adPodId {
 		podImps := groupByPods(imps)
 
 		requests := make([]*adapters.RequestData, 0, len(podImps))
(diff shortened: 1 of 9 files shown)


openrtb_ext/imp_appnexus.go
```diff
@@ -17,6 +17,7 @@ type ExtImpAppnexus struct {
 	UsePmtRule              *bool                   `json:"use_pmt_rule"`
 	// At this time we do no processing on the private sizes, so just leaving it as a JSON blob.
 	PrivateSizes json.RawMessage `json:"private_sizes"`
+	AdPodId      bool            `json:"adPodId"`
```
SyntaxNode, line 20: Recommend renaming to `generate_ad_pod_id`

adapters/appnexus/appnexus.go
```diff
@@ -302,10 +303,11 @@ func (a *AppNexusAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *ad
 		}
 	}
 	for i := 0; i < len(request.Imp); i++ {
-		memberId, err := preprocess(&request.Imp[i], defaultDisplayManagerVer)
+		memberId, impAdPodId, err := preprocess(&request.Imp[i], defaultDisplayManagerVer)
 		if memberId != "" {
 			memberIds[memberId] = true
 		}
+		adPodId = adPodId || impAdPodId
```
SyntaxNode, line 310: Should this produce an error or assume it's enabled for all imps? Does this adapter have an existing pattern to follow if there's missing or conflicting data in an imp?

openrtb_ext/imp_appnexus.go
VeronikaSolovei9: Renamed

adapters/appnexus/appnexus.go
VeronikaSolovei9: There is no consistent pattern, it now throws bidder level error.

hhhjort: I think the ideal situation is, if this flag is on in at least one Imp, and either missing or on in the rest, then we generate Ids. If some have it on and others have it off, then we throw an error. If it is just off or missing, then no Ids. This really should be a request level flag rather than an Imp level flag, but we don't have request level bidder parameters.

Actually, now that I think of it, there was a proposal for request level bidder params, that would be merged into all the imp level bidder params. I don't recall that it was ever implemented.

hhhjort (approved):
Looks good. This version will error out unless all Imps have `generate_ad_pod_id` set to true, or no imps have `generate_ad_pod_id` set to true. It would be valid for some to be false and others missing, but not for some to be true and others missing.

SyntaxNode:
> Looks good. This version will error out unless all Imps have `generate_ad_pod_id` set to true, or no imps have `generate_ad_pod_id` set to true.

It's also be valid for no imps to have `generate_ad_pod_id` set to true.

adapters/appnexus/appnexus_test.go
```diff
@@ -207,6 +205,37 @@ func TestVideoTwoPods(t *testing.T) {
 	assert.NotEqual(t, adPodId1, adPodId2, "AdPod id should be different for different pods")
 }
 
+func TestVideoTwoPodsDifferentAdPodId(t *testing.T) {
```
SyntaxNode, line 208: Please write this test using the json framework. I think it would be a good ides to add json tests to cover all true, all false, all missing, mix of true/missing, mix of false/missing, mix of true/false.

adapters/appnexus/appnexus.go
VeronikaSolovei9: `if this flag is on in at least one Imp, and either missing or on in the rest, then we generate Ids. If some have it on and others have it off, then we throw an error. If it is just off or missing, then no Ids.` - yes, this is is how it works now.

I don't think we have this functionality. `request level bidder params` - sounds like a new feature.

adapters/appnexus/appnexus_test.go
VeronikaSolovei9: Added json files to unit test cases you mentioned. I cannot find a good way to test 2 pods with adpodid enabled for both pods because I don't have a control on random number generation in unit tests. This specific case is covered in `TestVideoTwoPods` and `TestVideoTwoPodsManyImps`

VeronikaSolovei9: I suddenly started to get this error:
```
FAIL: TestVideoSinglePod (0.00s)
    appnexus_test.go:95: 
        	Error Trace:	appnexus_test.go:95
        	Error:      	Should be true
        	Test:       	TestVideoSinglePod
        	Messages:   	AdPod id doesn't present in Appnexus extension or has incorrect format
FAIL
```
It works if I run single test and doesn't if I run it using "validate.sh". RegExp matcher returns false. 
Please help!

adapters/appnexus/appnexus.go
hhhjort: Yes, I don't think it was ever implemented. But as I noted in my approval, you need all Imps set to true to generate ad pod IDs. If at least one imp is set to generate ad pod IDs, and another is left blank, this will error out.

adapters/appnexus/appnexus_test.go
hhhjort: regMatch, matchErr := regexp.Match(`[0-9]19`, []byte(reqDataExt.Appnexus.AdPodId))
	assert.NoError(t, matchErr, "Regex match error should be nil")
	assert.True(t, regMatch, "AdPod id doesn't present in Appnexus extension or has incorrect format")

We are matching for an adpod ID that contains the digits '19', but aren't the first two digits of a `rand.Int63()` value. Why would we expect the adpod ID contain a one followed by a nine? It seems reasonably likely given the length of a 64 bit number, but certainly not expected.

VeronikaSolovei9: Regexp modified to `^[0-9]+$`

hhhjort (approved):

AlexBVolcy (approved):

PR #1794, opened 2021-04-07
IX: Set bidVideo when category and duration is available

ixjohnny:
- Update IX adapter to set bidVideo if category and duration is available from response for video supply.

### DIFF
--- a/adapters/ix/ix.go
+++ b/adapters/ix/ix.go
@@ -402,9 +402,23 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
 			if !ok {
 				errs = append(errs, fmt.Errorf("Unmatched impression id: %s.", bid.ImpID))
 			}
+
+			var bidExtVideo *openrtb_ext.ExtBidPrebidVideo
+			var bidExt openrtb_ext.ExtBid
+			if bidType == openrtb_ext.BidTypeVideo {
+				unmarshalExtErr := json.Unmarshal(bid.Ext, &bidExt)
+				if unmarshalExtErr == nil && bidExt.Prebid != nil && bidExt.Prebid.Video != nil {
+					bidExtVideo = &openrtb_ext.ExtBidPrebidVideo{
+						Duration:        bidExt.Prebid.Video.Duration,
+						PrimaryCategory: bidExt.Prebid.Video.PrimaryCategory,
+					}
+				}
+			}
+
 			bidderResponse.Bids = append(bidderResponse.Bids, &adapters.TypedBid{
-				Bid:     &bid,
-				BidType: bidType,
+				Bid:      &bid,
+				BidType:  bidType,
+				BidVideo: bidExtVideo,
 			})
 		}
 	}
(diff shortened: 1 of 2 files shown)


adapters/ix/ix.go
```diff
@@ -402,9 +402,23 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
...
 				errs = append(errs, fmt.Errorf("Unmatched impression id: %s.", bid.ImpID))
 			}
+
+			var bidExtVideo *openrtb_ext.ExtBidPrebidVideo
+			var bidExt openrtb_ext.ExtBid
+			if bidType == openrtb_ext.BidTypeVideo {
+				unmarshalExtErr := json.Unmarshal(bid.Ext, &bidExt)
+				if unmarshalExtErr == nil && bidExt.Prebid != nil && bidExt.Prebid.Video != nil {
+					bidExtVideo = &openrtb_ext.ExtBidPrebidVideo{
+						Duration:        bidExt.Prebid.Video.Duration,
```
grossjo, line 412: can these be nil?

adapters/ix/ix.go
```diff
@@ -402,9 +402,23 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
 			if !ok {
 				errs = append(errs, fmt.Errorf("Unmatched impression id: %s.", bid.ImpID))
 			}
+
+			var bidExtVideo *openrtb_ext.ExtBidPrebidVideo
+			var bidExt openrtb_ext.ExtBid
+			if bidType == openrtb_ext.BidTypeVideo {
+				unmarshalExtErr := json.Unmarshal(bid.Ext, &bidExt)
```
grossjo, line 409: Do we need to log on errors?

adapters/ix/ix.go
```diff
@@ -402,9 +402,23 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
...
+					}
+				}
+			}
+
 			bidderResponse.Bids = append(bidderResponse.Bids, &adapters.TypedBid{
-				Bid:     &bid,
-				BidType: bidType,
+				Bid:      &bid,
+				BidType:  bidType,
+				BidVideo: bidExtVideo,
```
grossjo, line 421: could be unassigned.

adapters/ix/ix.go
ixjohnny: PrimaryCategory/Duration? no, they will default to their zero value if they're not set.

adapters/ix/ix.go
ixjohnny: It'll be nil in that case.

mlb7687 (approved):
Change LGTM

adapters/ix/ix.go
ixjohnny: Done.

VeronikaSolovei9:
LGTM

ixjohnny:
Thanks for the review @VeronikaSolovei9 ! Is there anything else you need from me to approve the PR?

bsardo (approved):
LGTM. Just for awareness, we usually recommend using the JSON test framework to achieve adapter test coverage but that's not necessary here because our test framework doesn't currently support the `video` field on bid responses. We will add support in the future.

VeronikaSolovei9 (approved):

Issue #1787, opened 2021-04-01
Conversant adapter dropping bidfloor


patmmccann:
We recently updated to prebid 4.28+ which incorporated https://github.com/prebid/Prebid.js/pull/6273/files. For most of the request bodies with debug on I can see bidfloor being passed to adapter endpoints. Conversant however drops it from its endpoint, but stil passes currency eg `requestbody: "{"id":"46feca51-4311-4f57-acbe-7387ebe56c15","imp":[{"id":"AdThrive_Content_2_desktop","banner":{"format":[{"w":728,"h":90},{"w":300,"h":250},{"w":320,"h":50},{"w":970,"h":250},{"w":336,"h":280},{"w":970,"h":90},{"w":300,"h":50},{"w":320,"h":100},{"w":468,"h":60},{"w":250,"h":250},{"w":120,"h":240},{"w":1,"h":1},{"w":300,"h":300},{"w":552,"h":334},{"w":728,"h":250}]},"displaymanager":"prebid-s2s","displaymanagerver":"2.0.0","tagid":"8fdfa3de","bidfloorcur":"USD","secure":1,"ext":{"bidder":{"site_id":"203587","tag_id":"8fdfa3de"}}}],"site":{"id":"203587","domain":"firstquarterfinance.com","page":"https://firstquarterfinance.com/?pbjs_debug=true","publisher":{"id":"72409f4d-9858-41ac-bf45-be255b82de89"},"ext":{"amp":0}},"device":{"ua":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36","ip":"100.35.114.240","h":920,"w":1848,"ext":{"prebid":{"interstitial":{"minwidthperc":50,"minheightperc":40}}}},`

Test page:
https://firstquarterfinance.com/?pbjs_debug=true

SyntaxNode:
> I seem to recall all adapters were fixed to read from the standard location.

We did not update adapters to use the standard bid floor location. I believe we asked adapter maintainers to update and we look for it in new PRs. 

@johnwier Can you please look into this for the Conservant adapter?

patmmccann:
[...]

PR #1795, opened 2021-04-07
Openx: use bidfloor if set - prebid.js adapter behavior 

laurb9:
- Fixes bug that set bidfloor to 0 when customFloor not given (so almost always)
- Implements behavior in https://github.com/prebid/Prebid.js/blob/master/modules/openxBidAdapter.js#L495
- for #1787


hhhjort (approved):
LGTM

mansinahar (approved):

PR #1802, opened 2021-04-12
Update openrtb library to v15

rokostik:
Updates the openrtb library to v15, which fixes an issue described here https://github.com/mxmCherry/openrtb/issues/44.

The changes are mostly import paths except some explicit pointer handling.

Not sure why there's so many changes in go.mod and go.sum, I just ran `make deps`. Please let me know if I should do something differently about that.


SyntaxNode:
> Not sure why there's so many changes in go.mod and go.sum, I just ran `make deps`. Please let me know if I should do something differently about that.

`go.sum` does its own thing. The `make deps` command already includes a run of `go run tidy`. Try running it again or directly run `go mod tidy` in the project folder to be sure we only have the required references.

exchange/bidder.go
```diff
@@ -280,27 +280,27 @@ func addNativeTypes(bid *openrtb2.Bid, request *openrtb2.BidRequest) (*nativeRes
 
 func setAssetTypes(asset nativeResponse.Asset, nativePayload nativeRequests.Request) error {
 	if asset.Img != nil {
-		if tempAsset, err := getAssetByID(asset.ID, nativePayload.Assets); err == nil {
+		if tempAsset, err := getAssetByID(*asset.ID, nativePayload.Assets); err == nil {
```
SyntaxNode, line 283: The `asset.ID` now needs a `nil` check. Consider returning an error right after `asset.Img != nil` if the `asset.ID` is `nil`. To ensure the code behaves correctly, please add a test for when the id is omitted from the native json.

exchange/bidder.go
```diff
@@ -280,27 +280,27 @@ func addNativeTypes(bid *openrtb2.Bid, request *openrtb2.BidRequest) (*nativeRes
...
+				return fmt.Errorf("Response has an Image asset with ID:%d present that doesn't exist in the request", *asset.ID)
 			}
 		} else {
 			return err
 		}
 	}
 
 	if asset.Data != nil {
-		if tempAsset, err := getAssetByID(asset.ID, nativePayload.Assets); err == nil {
+		if tempAsset, err := getAssetByID(*asset.ID, nativePayload.Assets); err == nil {
```
SyntaxNode, line 297: Need to add a `nil` check and tests here too.

exchange/bidder.go
```diff
@@ -280,27 +280,33 @@ func addNativeTypes(bid *openrtb2.Bid, request *openrtb2.BidRequest) (*nativeRes
 
 func setAssetTypes(asset nativeResponse.Asset, nativePayload nativeRequests.Request) error {
 	if asset.Img != nil {
-		if tempAsset, err := getAssetByID(asset.ID, nativePayload.Assets); err == nil {
+		if asset.ID == nil {
+			return fmt.Errorf("Response Image asset doesn't have an ID")
```
SyntaxNode, line 284: Nitpick: There's no need to use `fmt.Errorf` since there is no dynamic content. Please use `errors.New()` here and for the Data asset.

SyntaxNode (commented):
Thank you for identifying the issue and coordinating between mxmcherry and prebid. A left a small nitpick, but otherwise looks good. I'll assign a second reviewer today.

bsardo (commented):
LGTM other than Scott's nitpick on `fmt.Error`

exchange/bidder.go
rokostik: Oh yeah of course, didn't notice that. Will fix.

bsardo (approved):

SyntaxNode (approved):

PR #1809, opened 2021-04-14
Bidmachine: Update JSON Schema Reference To Type String

agilfix:
Remove definitions object from schema and define types and other parameters directly in properties objects to ensure compatibility with more downstream systems that use this schema.

### DIFF
--- a/static/bidder-params/bidmachine.json
+++ b/static/bidder-params/bidmachine.json
@@ -5,22 +5,19 @@
     "type": "object",
     "properties": {
         "host": {
-            "$ref": "#/definitions/non-empty-string",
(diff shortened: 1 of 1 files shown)


SyntaxNode:
We'll assign PBS reviewers once you have approval from Bidmachine.

hhhjort:
Both forms seem to be valid. As Scott says, we would want approval from Bidmachine (@meeeeeepo ?) before making such a change.

meeeeeepo (approved):

hhhjort (approved):

SyntaxNode (approved):

PR #1810, opened 2021-04-15
Allow Native Ad Exchange Specific Types

SyntaxNode:
Resolves https://github.com/prebid/prebid-server/issues/1732

The Native 1.2 Spec allows for various exchange specific types. Prebid Server previously considered these invalid and failed validation. This PR allows custom 500+ exchange specific type values in all places allowed by the spec.

Introduces unit tests for most of the native validation code. This was sparsely covered by json tests previously.


endpoints/openrtb2/auction.go
```diff
@@ -657,7 +657,7 @@ func validateNativeContextTypes(cType native1.ContextType, cSubtype native1.Cont
 		// Context is only recommended, so none is a valid type.
```
SyntaxNode, line 657: I would appreciate feedback on the context and context sub type validation. The spec includes combinations to enforce, but also allows for exchange specific values without accompanying validation guidance. I'm interpreting this to mean exchange specific context types and context sub types are allowed in any combination.

endpoints/openrtb2/auction.go
```diff
@@ -666,28 +666,27 @@ func validateNativeContextTypes(cType native1.ContextType, cSubtype native1.Cont
 	if cSubtype == 0 {
 		return nil
 	}
-
-	if cSubtype >= 500 {
-		return fmt.Errorf("request.imp[%d].native.request.contextsubtype can't be greater than or equal to 500. See https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=39", impIndex)
-	}
```
SyntaxNode, line 672: I'm a little nervous this was explicitly forbidden before. I also have no idea why this was enforced when the spec allows for it. Thoughts?

endpoints/openrtb2/sample-requests/invalid-native/contextsubtype-greater-than-max.json
```diff
@@ -1,26 +0,0 @@
-{
```
SyntaxNode, line 1: This test no longer represents a valid condition.

endpoints/openrtb2/sample-requests/valid-native/eventtracker-exchange-specific.json
```diff
@@ -0,0 +1,35 @@
+{
```
SyntaxNode, line 1: This test specifically verifies end to end functionally for https://github.com/prebid/prebid-server/issues/1732

endpoints/openrtb2/auction.go
SyntaxNode: I like the extrapolation. I'm not sure it works that way in the wild. The linked issue is using event code 555.

endpoints/openrtb2/auction.go
SyntaxNode: Perhaps. I don't know why we wouldn't support exchange-specific type since we're just passing through data. The validation only exists to eliminate obviously wrong native responses.

endpoints/openrtb2/auction.go
bsardo: Ah ok I figured we might not have free rein to define how this works 🙂 . Event code 555 is the context type right? I don't see the associated subtypes in the linked Open Measurement spec. I guess it's best not to limit the subtypes for exchange-specific types since this is already in the wild then.

mansinahar: Yeah, I like the extrapolation as well but don't think that would work for the exchange specific context types and subtypes. There doesn't seem to be any validation guidelines for those in the spec so I agree with Scott that it looks like that they are allowed in any combination

endpoints/openrtb2/auction.go
mansinahar: I am not aware of any specific reasons for not allowing exchange specific types either

endpoints/openrtb2/auction.go
```diff
@@ -852,7 +851,7 @@ func validateNativeAssetVideo(video *nativeRequests.Video, impIndex int, assetIn
 }
 
 func validateNativeAssetData(data *nativeRequests.Data, impIndex int, assetIndex int) error {
-	if data.Type < native1.DataAssetTypeSponsored || data.Type > native1.DataAssetTypeCTAText {
+	if data.Type < native1.DataAssetTypeSponsored || (data.Type > native1.DataAssetTypeCTAText && data.Type < 500) {
```
mansinahar, line 854: I am not sure why we don't have the asset type check for the image asset as well but can we add that too in the `validateNativeAssetImage` method? From the spec, it looks like Image can have exchange specific types as well

endpoints/openrtb2/auction_test.go
```diff
@@ -2145,6 +2146,425 @@ func TestAuctionWarnings(t *testing.T) {
...
+		},
+		{
+			description:      "Content - Sub Type In Range",
+			givenContextType: 1,  // Content constant
+			givenSubType:     10, // Content range is currently 10-15
+			expectedError:    "",
+		},
+		{
+			description:      "Content - Sub Type In Range - Context Type Exchange Specific Boundary",
+			givenContextType: 500,
```
mansinahar, line 2208: Should `500` be allowed? The spec doesn't mention if 500 is included or not for context types and subtypes but I do see the following wording for image and data asset types:
```
Reserved	for Exchange specific usage numbered above 500
```

I am reading that as only values above 500 to be included. Also, the `500+` suggests the same to me. Thoughts?

bsardo: Hm good question. I read `500+` as inclusive but I see the confusion with that description. If I were creating custom values I would start with 501...

SyntaxNode: Good point. I took it to mean 500 inclusive, but I can understand from the quote above how others might interpret it to be exclusive. We chatted about this and we don't see a problem including 500, even if that's not the intent of the spec, to best handle the situation where others might read it as 500 inclusive.

endpoints/openrtb2/auction.go
SyntaxNode: It can, yes. I also don't know why we don't validate the image asset type. I'll add that in a follow-up PR.

endpoints/openrtb2/auction.go
```diff
@@ -803,14 +802,14 @@ func validateNativeAssetTitle(title *nativeRequests.Title, impIndex int, assetIn
...
 	}
 	if len(tracker.Methods) < 1 {
 		return fmt.Errorf("request.imp[%d].native.request.eventtrackers[%d].method is required. See section 7.7: https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=43", impIndex, eventIndex)
 	}
 	for methodIndex, method := range tracker.Methods {
-		if method < native1.EventTrackingMethodImage || method > native1.EventTrackingMethodJS {
+		if method < native1.EventTrackingMethodImage || (method > native1.EventTrackingMethodJS && method < openrtb_ext.NativeExchangeSpecificLowerBound) {
 			return fmt.Errorf("request.imp[%d].native.request.eventtrackers[%d].methods[%d] is invalid. See section 7.7: https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf#page=43", impIndex, eventIndex, methodIndex)
 		}
 	}
```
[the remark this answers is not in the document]
SyntaxNode: There is no type in the `validateNativeAssetVideo` object. :)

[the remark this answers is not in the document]
bsardo: Ha not sure what happened there 🤦

bsardo (approved):
LGTM

mansinahar (approved):

PR #1811, opened 2021-04-15
Outbrain: transform native eventtrackers to imptrackers and jstracker

rokostik:
Since the PUC native-trk library doesn't support eventtrackers, this PR transforms them to the deprecated imptrackers and jstracker for the outbrain adapter.

### DIFF
--- a/adapters/outbrain/outbrain.go
+++ b/adapters/outbrain/outbrain.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/mxmCherry/openrtb/v15/native1"
+	nativeResponse "github.com/mxmCherry/openrtb/v15/native1/response"
 	"github.com/mxmCherry/openrtb/v15/openrtb2"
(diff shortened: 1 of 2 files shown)


guscarreon (approved):
LGTM

SyntaxNode:
> Since the PUC native-trk library doesn't support eventtrackers

I'm not familar with that library. Is it used by all of your publishers? Should you consider making this behavior optional for publishers using a different library (if applicable)?

rokostik:
> I'm not familar with that library. Is it used by all of your publishers? Should you consider making this behavior optional for publishers using a different library (if applicable)?

Oh no, that's the library generated from the [PUC repo](https://github.com/prebid/prebid-universal-creative/blob/master/gulpfile.js#L122). It's the one referenced in the [native adops tutorial](https://docs.prebid.org/prebid-mobile/adops-native-setup.html), so I'm guessing all publishers use this library for native ads?

hhhjort (approved):
LGTM

PR #1814, opened 2021-04-16
AppNexus: reform bid floor handling

hhhjort:
Set Appnexus to prefer the openrtb bid floor if it is supplied.
RE: https://github.com/prebid/prebid-server/issues/1787

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -495,7 +495,7 @@ func preprocess(imp *openrtb2.Imp, defaultDisplayManagerVer string) (string, err
 	if appnexusExt.InvCode != "" {
 		imp.TagID = appnexusExt.InvCode
 	}
-	if appnexusExt.Reserve > 0 {
+	if imp.BidFloor <= 0 && appnexusExt.Reserve > 0 {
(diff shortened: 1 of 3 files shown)


adapters/appnexus/appnexus.go
```diff
@@ -495,7 +495,7 @@ func preprocess(imp *openrtb2.Imp, defaultDisplayManagerVer string) (string, err
 	if appnexusExt.InvCode != "" {
 		imp.TagID = appnexusExt.InvCode
 	}
-	if appnexusExt.Reserve > 0 {
+	if imp.BidFloor <= 0 && appnexusExt.Reserve > 0 {
```
SyntaxNode, line 498: Tests or didn't happen. :)

SyntaxNode, line 498: We don't validate the BidFloor in the original request. Would it be better to just check for `0` to be sure it omitted rather than explicitly set to a nonsensical value?

mansinahar: I think since a negative value would anyway be invalid, this check makes sense to me

SyntaxNode (approved):

PR #1815, opened 2021-04-19
New Adapter: adf (adformOpenRTB)

braizhas:
Hello,
PR adds adformOpenRTB adapter. 
Docs: https://github.com/prebid/prebid.github.io/pull/2887

The adapter name does not comply with a naming convention. Is it ok to add one more exception https://github.com/prebid/prebid-server/compare/master...adform:adfom-openrtb-adapter?expand=1#diff-5e8919c9857c3a2fd497d7c49c5dfe37f18651fa7e8627f55629cdde51f2c96bR53 ?


SyntaxNode:
> The adapter name does not comply with a naming convention. Is it ok to add one more exception?

The check exists to ensure there is enough uniqueness in the bidder name for the publisher to work with targeting keys without fear of collision.

What's the difference between this adapter and the existing adform adapter? Would a publisher ever in be in a position where both would be enabled?

braizhas:
In the future, this adapter should replace adform adapter. Currently, this adapter supports only native ads while adform adapter supports banner and video ads. 
Prebid.js adformOpenRTB already exists for some time and there were no issues with targeting key collision. Seems there is no use case in adform bids separation according to the adapter.

adapters/adformOpenRTB/adformOpenRTB.go
```diff
@@ -0,0 +1,107 @@
...
+func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
+	bidder := &adapter{
+		endpoint: config.Endpoint,
+	}
+	return bidder, nil
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errors []error
+	requestCopy := *request
```
hhhjort, line 29: This makes a shallow copy of `request`. It does not appear that you refer to this `request` again. The original `request` is already a shallow copy of the incoming request unique to your adapter, so making another shallow copy is not needed.

braizhas: updated

SyntaxNode:
> Currently, this adapter supports only native ads while adform adapter supports banner and video ads.

It seems then a publisher wouldn't include both adform adapters for the same impression. If that does happen though, please understand I'm not clear on the consequences of targeting key collisions.

SyntaxNode (dismissed):

adapters/adformOpenRTB/adformOpenRTB.go
```diff
@@ -0,0 +1,106 @@
+package adformOpenRTB
```
SyntaxNode, line 1: Package names in Go are preferred to be lower case. Please consider naming your package and folder adformopenrtb.

openrtb_ext/bidders.go
```diff
@@ -76,6 +76,7 @@ const (
 	Bidder33Across         BidderName = "33across"
 	BidderAcuityAds        BidderName = "acuityads"
 	BidderAdform           BidderName = "adform"
+	BidderAdformOpenRTB    BidderName = "adformOpenRTB"
```
SyntaxNode, line 79: Should the bidder name also be lower case? Thoughts @hhhjort?

hhhjort: There is only one exception to all lower case, probably best to not make it two.

braizhas:
Renamed adapter to be compliant. Will also rename Prebid.js adapter adding adformOpenRTB as an alias name. This should solve naming issue in a backwards-compatible manner.

hhhjort (approved):
LGTM

config/config.go
```diff
@@ -568,6 +568,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	externalURL := cfg.ExternalURL
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.Bidder33Across, "https://ic.tynt.com/r/d?m=xch&rt=html&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&ru="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3D33across%26uid%3D33XUSERID33X&id=zzz000000000002zzz")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderAcuityAds, "https://cs.admanmedia.com/sync/prebid?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dacuityads%26uid%3D%5BUID%5D")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderAdf, "https://cm.adform.net/cookie?redirect_url="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dadform%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26us_privacy%3D{{.USPrivacy}}%26uid%3D%24UID")
```
SyntaxNode, line 571: There's a copy/paste typo in the redirect url. It still has bidder "adform" instead of "adf".

I also see you added the USPrivacy macro the redirect url. PBS makes no use of this field. If a user sync is prevented due to CCPA the cookie sync endpoint will never direct them to the setuid callback. Did you intend to add the USPrivacy macro as an Adform's query param?

braizhas: thanks for the note - fixed.

SyntaxNode (approved):

hhhjort (approved):

PR #1827, opened 2021-04-26
InMobi: Updating support for sites

cormorinth:
Adding site to static yaml, and exemplary tests


SyntaxNode (approved):

mansinahar (approved):

mansinahar:
Can you please also open a doc PR to include `pbs_site_supported: true` in the table here: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/inmobi.md as well?

SyntaxNode:
Thank you @mansinahar. I didn't realize the docs include app and site support. @cormorinth We'll wait to merge this in until the docs PR is created.

cormorinth:
I've done as asked, although I don't see reference of that string anywhere in the docs (https://docs.prebid.org/prebid-server/developers/add-new-bidder-java.html), or within code searches of the entire https://github.com/prebid/prebid.github.io repo.

SyntaxNode:
> The mismatch between the yaml configuration and the docs is quite curious as pbs_app_supported seems to imply web as a given, but my (poorly informed) presumption is that it's a holdover from prebid originally being prebid.js.

Thank you. Indeed, that is curious. You very may well be correct. We'll have that discussion in the prebid.github.io PR. We're all good here.

PR #1835, opened 2021-05-04
New Adapter: Interactive Offers

IOTiagoFaria:


SyntaxNode:
The test are failing because you need to add your bidder name to the list in `func CoreBidderNames() []BidderName {`. On a side note, I realize this is annoying and we are exploring alternative approaches.

bsardo (approved):
LGTM

PR #1837, opened 2021-05-04
IX: Set category in bid.cat

mlb7687:
From context in https://prebid.slack.com/archives/C7UCUKKB8/p1619639305041100, we should be setting bid.cat, not PrimaryCategory in order to utilize category translation.
Update adapter to accept both for compatibility as we fully migrate to bid.cat in exchange response.

### DIFF
--- a/adapters/ix/ix.go
+++ b/adapters/ix/ix.go
@@ -411,7 +411,7 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
 					bidExtVideo = &openrtb_ext.ExtBidPrebidVideo{
 						Duration: bidExt.Prebid.Video.Duration,
 					}
-					if bid.Cat == nil || len(bid.Cat) == 0 {
+					if len(bid.Cat) == 0 {
 						bid.Cat = []string{bidExt.Prebid.Video.PrimaryCategory}
 					}
 				}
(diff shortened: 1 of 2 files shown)


hhhjort (approved):
LGTM

adapters/ix/ix.go
```diff
@@ -409,8 +409,10 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
 				unmarshalExtErr := json.Unmarshal(bid.Ext, &bidExt)
 				if unmarshalExtErr == nil && bidExt.Prebid != nil && bidExt.Prebid.Video != nil {
 					bidExtVideo = &openrtb_ext.ExtBidPrebidVideo{
-						Duration:        bidExt.Prebid.Video.Duration,
-						PrimaryCategory: bidExt.Prebid.Video.PrimaryCategory,
+						Duration: bidExt.Prebid.Video.Duration,
+					}
+					if bid.Cat == nil || len(bid.Cat) == 0 {
```
SyntaxNode, line 414: Nitpick: Go has graceful handling of `nil`. It's safe to just use `len(bid.Cat) == 0` since `len` will return 0 for `nil`.

mlb7687: Good catch. Changed

bsardo (approved):

hhhjort (approved):

PR #1838, opened 2021-05-06
New Adapter : Kayzen

raghuteja:


adapters/kayzen/kayzen.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return bidder, nil
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) (requests []*adapters.RequestData, errors []error) {
+	var kayzenExt *openrtb_ext.ExtKayzen
+	var err error
+
+	if len(request.Imp) > 0 {
+		kayzenExt, err = a.getImpressionExt(&(request.Imp[0]))
```
hhhjort, line 39: This only pays attention to the Kayzen ext on the first Imp. If someone tries to get fancy and put in a single large request where not all the Imps have the same parameters, you may miss something. Then again, you are only zeroing out the ext on the first Imp, so you have access to the other parameters on your bidder. This is not necessarily an issue, it depends on your bidder setup I suppose. I just wanted to flag this to make sure you are aware.

raghuteja: Thanks for raising this and Yes, this was intentional to consider only the first impression ext params

config/config.go
```diff
@@ -862,6 +862,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.interactiveoffers.endpoint", "https://rtb.ioadx.com/bidRequest/?partnerId=d9e56d418c4825d466ee96c7a31bf1da6b62fa04")
 	v.SetDefault("adapters.ix.disabled", true)
 	v.SetDefault("adapters.jixie.endpoint", "https://hb.jixie.io/v2/hbsvrpost")
+	v.SetDefault("adapters.kayzen.endpoint", "https://bids-{{.ZoneID}}.bidder.kayzen.io/?exchange={{.AccountID}}")
+	v.SetDefault("adapters.kayzen.disabled", true)
```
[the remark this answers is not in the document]
raghuteja: Thanks for clarifying, I have enabled by default

raghuteja (commented):
I have addressed comments related to minLength as well
Yes, It should not pass empty strings
So I have added that minLength condition

hhhjort (approved):
LGTM

adapters/kayzen/kayzen.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	var bidResp openrtb2.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: "Bad Server Response",
+		}}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(1)
```
guscarreon, line 126: Prebid server core assigns `"USD"` as default currency if the bidResponse's `Currency` field is found to have an empty value. (Line 189 of `exchange/bidder.go` below)
```
187             if bidResponse != nil {
188                 // Setup default currency as `USD` is not set in bid request nor bid response
189                 if bidResponse.Currency == "" {
190                     bidResponse.Currency = defaultCurrency
191                 }
exchange/bidder.go
```
Will your bid server always bid in `"USD"`? If not, some adapters that use multi-currencies explicitly assign a currency value in their `MakeBids()` implementation. This is not a requirement is only a suggestion in case it makes sense in the context of your adapter bidder.
```
100   func (a *adapter) MakeBids(internalRequest *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
101       if response.StatusCode == http.StatusNoContent {
102           return nil, nil
103       }
104   *-- 21 lines: if response.StatusCode == http.StatusBadRequest {--------------------------------------------------------------------------------------------------------------------------------------------------------------------
125  
126       bidResponse := adapters.NewBidderResponseWithBidsCapacity(1)
    +     bidResponse.Currency = response.Cur
127  
128       var errs []error
129       for _, seatBid := range bidResp.SeatBid {
130   *--  7 lines: for _, bid := range seatBid.Bid {------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
137       }
138       return bidResponse, errs
139   }
adapters/kayzen/kayzen.go
```

guscarreon (commented):
Looks pretty good. One last question before approving

adapters/kayzen/kayzen.go
raghuteja: Thanks for raising this
Yes our bidder always bid in USD, So that should not be an issue

guscarreon: Thank you for addressing our feedback.

guscarreon (approved):
LGTM

PR #1839, opened 2021-05-06
Refactor: Removed unused GDPR return value

bsardo:


endpoints/setuid_test.go
```diff
@@ -439,8 +439,8 @@ func (g *mockPermsSetUID) BidderSyncAllowed(ctx context.Context, bidder openrtb_
 	return false, nil
 }
 
-func (g *mockPermsSetUID) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal gdpr.Signal, consent string, weakVendorEnforcement bool) (bool, bool, bool, error) {
-	return g.allowPI, g.allowPI, g.allowPI, nil
+func (g *mockPermsSetUID) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal gdpr.Signal, consent string, weakVendorEnforcement bool) (bool, bool, error) {
+	return g.allowPI, g.allowPI, nil
```
SyntaxNode, line 443: Should we consider renaming `g.allowPI` with removal of the `allowPI` return value? I think the `personalInfoAllowed` variable used below looks a bit more straightforward.

exchange/utils.go
```diff
@@ -133,7 +133,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 				}
 			}
 			var publisherID = req.LegacyLabels.PubID
-			_, geo, id, err := gDPR.PersonalInfoAllowed(ctx, bidderRequest.BidderCoreName, publisherID, gdprSignal, consent, weakVendorEnforcement)
+			geo, id, err := gDPR.PersonalInfoAllowed(ctx, bidderRequest.BidderCoreName, publisherID, gdprSignal, consent, weakVendorEnforcement)
```
SyntaxNode, line 136: Nice! This looks a lot cleaner.

endpoints/auction_test.go
```diff
@@ -454,8 +453,8 @@ func (m *auctionMockPermissions) BidderSyncAllowed(ctx context.Context, bidder o
 	return m.allowBidderSync, nil
 }
 
-func (m *auctionMockPermissions) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal gdpr.Signal, consent string, weakVendorEnforcement bool) (bool, bool, bool, error) {
-	return m.allowPI, m.allowGeo, m.allowID, nil
+func (m *auctionMockPermissions) PersonalInfoAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal gdpr.Signal, consent string, weakVendorEnforcement bool) (bool, bool, error) {
```
SyntaxNode, line 456: I see in other methods you named the return arguments to `allowGeo bool, allowID bool, err error`. Could you please do that for all implementations of `PersonalInfoAllowed`?

gdpr/impl_test.go
```diff
@@ -286,10 +286,11 @@ func TestAllowPersonalInfo(t *testing.T) {
 	for _, tt := range tests {
 		perms.cfg.UsersyncIfAmbiguous = tt.userSyncIfAmbiguous
 
-		allowPI, _, _, err := perms.PersonalInfoAllowed(context.Background(), tt.bidderName, tt.publisherID, tt.gdpr, tt.consent, tt.weakVendorEnforcement)
+		_, allowID, err := perms.PersonalInfoAllowed(context.Background(), tt.bidderName, tt.publisherID, tt.gdpr, tt.consent, tt.weakVendorEnforcement)
```
SyntaxNode, line 289: Why does this not test allowGeo also?

gdpr/impl_test.go
```diff
@@ -286,10 +286,11 @@ func TestAllowPersonalInfo(t *testing.T) {
...
 
 		assert.Nil(t, err, tt.description)
-		assert.Equal(t, tt.allowPI, allowPI, tt.description)
+		// assert.Equal(t, tt.allowPI, allowPI, tt.description)
```
SyntaxNode, line 292: Oops?

gdpr/impl_test.go
bsardo: This is a known test issue that will be addressed in another PR.

gdpr/impl.go
```diff
@@ -181,17 +180,7 @@ func (p *permissionsImpl) allowPITCF2(parsedConsent api.VendorConsents, vendor a
...
-	allowPI = true
-	if p.cfg.TCF2.Purpose1.Enabled {
-		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.InfoStorageAccess, weakVendorEnforcement)
-	}
-	if p.cfg.TCF2.Purpose2.Enabled {
-		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.BasicAdserving, weakVendorEnforcement)
-	}
-	if p.cfg.TCF2.Purpose7.Enabled {
-		allowPI = allowPI && p.checkPurpose(consent, vendor, vendorID, consentconstants.AdPerformance, weakVendorEnforcement)
-	}
```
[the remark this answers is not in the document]
bsardo: I think it is outdated as I don't see this logic anywhere in the TCF 2.0 spec. Either that or we have a bug because the `allowPI` value isn't being used anywhere 🙂.  The purpose enabled config flags will be added in the near future in separate PRs to appropriately set the allowID and allowGeo flags.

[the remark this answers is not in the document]
bsardo: This logic was added as part of TCF 2.0 about a year ago. I'm wondering if the requirements were misunderstood or have changed since this logic was added. `allowPI` was used to determine whether user IDs should be stripped when TCF 2.0 was first implemented and this logic is saying to strip it if at least one of the enabled purposes does not have consent while the requirements say that they should be stripped if at least one of the enabled purposes has consent.

From the spec:
> 19) User IDs must be stripped from a given vendor’s auction in these scenarios:
- User has not provided a legal basis for any Purpose 2,3,4,5,6,7,8,9, or 10 and the publisher has not defined an exception for the vendor for any of these purposes.

Also from the spec:
**Activity** - Passing User IDs into an auction
**Legal Basis Required** - Any Purpose 2-10. User IDs are important for more than personalizing ads - they can be used in frequency capping, building profiles, counting unique users, etc. So Prebid Server should pass User IDs through the auction if any of Purposes 2-10 pass the legal basis test.

I was planning to add this logic under a separate PR.

@hhhjort thoughts?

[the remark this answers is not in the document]
hhhjort: We have both AllowID and AllowPI. So lines 177-182 take care of AllowID for point 19 from the spec quoted above. AllowPI is to flag other non-geo personal information, like age or gender, for screening or not. The valid uses cases for these other pieces of info were considered more limited, and thus only available for a limited selection of purposes.

hhhjort (approved):
LGTM

SyntaxNode (approved):

PR #1840, opened 2021-05-07
Conversant bid floor handling

johnwier:
conversant adapter was setting the bid floor to 0 when the custom floor was not set.

dropped the trailing space on the firstpartydata-imp-ext-multiple-prebid-bidders.json filename as it was causing git to fail on windows

/src/github.com/prebid/prebid-server/exchange/exchangetest# ls -1 firstpartydata-im
p-ext-multiple-*
firstpartydata-imp-ext-multiple-bidders.json
'firstpartydata-imp-ext-multiple-prebid-bidders.json '


adapters/conversant/conversant.go
```diff
@@ -74,8 +74,14 @@ func (c ConversantAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *a
 func parseCnvrParams(imp *openrtb2.Imp, cnvrExt openrtb_ext.ExtImpConversant) {
 	imp.DisplayManager = "prebid-s2s"
 	imp.DisplayManagerVer = "2.0.0"
-	imp.BidFloor = cnvrExt.BidFloor
-	imp.TagID = cnvrExt.TagID
+
+	if imp.BidFloor <= 0 && cnvrExt.BidFloor > 0 {
+		imp.BidFloor = cnvrExt.BidFloor
+	}
```
guscarreon, line 80: In case the `cnvrExt.BidFloor` is zero, and `imp.BidFloor` is negative, `imp.BidFloor` will continue being negative. Do we mind this? In your context, does it make sense to default `imp.BidFloor` to a non-negative value?
```
74   func parseCnvrParams(imp *openrtb2.Imp, cnvrExt openrtb_ext.ExtImpConversant) {
75       imp.DisplayManager = "prebid-s2s"
76       imp.DisplayManagerVer = "2.0.0"
77  
78       if imp.BidFloor <= 0 && cnvrExt.BidFloor > 0 {
79           imp.BidFloor = cnvrExt.BidFloor
80       }
   +      
   +     if imp.BidFloor < 0 {
   +         // set to default value
   +         imp.BidFloor = defaultBidFloorValue
   +     }
81  
82       if len(cnvrExt.TagID) > 0 {
83           imp.TagID = cnvrExt.TagID
84       }
85  
86       // Take care not to override the global secure flag
87       if (imp.Secure == nil || *imp.Secure == 0) && cnvrExt.Secure != nil {
88           imp.Secure = cnvrExt.Secure
89       }
adapters/conversant/conversant.go
```

adapters/conversant/conversanttest/supplemental/test_params.json
```diff
@@ -0,0 +1,258 @@
...
+          "format": [{"w": 300, "h": 250}],
+          "pos": 1,
+          "mimes": ["video/mp4"],
+          "api": [1,2],
+          "protocols": [2],
+          "maxduration": 60
+        },
+        "ext": {
+          "bidder": {
+            "site_id": "108060"
```
guscarreon, line 20: Can we please add a test case were ext-defined `bidFloor` is zero and the `bidFloor` defined at the `imp` level is negative?
```
{
  "mockBidRequest": {
    "id": "testauction",
    "imp": [
      {
        "id": "1",
        "banner": {
          "format": [{"w": 300, "h": 250}]
        },
        "bidfloor": -1,
        "ext": {
          "bidder": {
            "site_id": "108060",
          }
        }
      }
    ],
```

guscarreon (changes requested):

adapters/conversant/conversant.go
johnwier: We treat zero and negative bid floors the same so it won't make a difference.

PR #1842, opened 2021-05-08
Pangle: add appid & placementid to bidder param

chshpra:
doc PR: https://github.com/prebid/prebid.github.io/pull/2942

changes:
- add the optional appid and placementid (which when used, must be used in conjunction) to bidder param, then use them to form networkids in imp.ext
- add a true is_prebid flag for the auto recognition of prebid requests on our side

### DIFF
--- a/adapters/pangle/pangle.go
+++ b/adapters/pangle/pangle.go
@@ -16,9 +16,16 @@ type adapter struct {
 	Endpoint string
 }
 
+type NetworkIDs struct {
+	AppID       string `json:"appid,omitempty"`
+	PlacementID string `json:"placementid,omitempty"`
+}
+
 type wrappedExtImpBidder struct {
 	*adapters.ExtImpBidder
-	AdType int `json:"adtype,omitempty"`
+	AdType     int         `json:"adtype,omitempty"`
+	IsPrebid   bool        `json:"is_prebid,omitempty"`
+	NetworkIDs *NetworkIDs `json:"networkids,omitempty"`
 }
 
 type pangleBidExt struct {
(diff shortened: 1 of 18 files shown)


adapters/pangle/pangle.go
```diff
@@ -16,9 +16,16 @@ type adapter struct {
...
+type NetworkIDs struct {
+	AppID       string `json:"appid,omitempty"`
+	PlacementID string `json:"placementid,omitempty"`
+}
+
 type wrappedExtImpBidder struct {
 	*adapters.ExtImpBidder
-	AdType int `json:"adtype,omitempty"`
+	AdType     int         `json:"adtype,omitempty"`
+	IsPrebid   bool        `json:"is_prebid,omitempty"`
```
SyntaxNode, line 27: I'm curious, why does Prebid need to be flagged specially?

static/bidder-params/pangle.json
```diff
@@ -8,9 +8,27 @@
       "type": "string",
       "description": "Access Token",
       "pattern": ".+"
+    },
+    "appid": {
+      "type": "string",
+      "description": "App ID",
+      "pattern": "[0-9]+"
```
SyntaxNode, line 15: Would this be the same effect as an integer type? Do you expect appid and placementid to have leading zeros or exceed an int32's length?

adapters/pangle/pangle.go
```diff
@@ -78,23 +85,31 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
...
-			if newImpExt, err := json.Marshal(impExt); err == nil {
-				imp.Ext = newImpExt
-			} else {
-				errs = append(errs, fmt.Errorf("failed re-marshalling imp ext with adtype"))
-				continue
+		}
+		// remarshal imp.ext
+		impExt.AdType = adType
+		impExt.IsPrebid = true
+		if len(bidderImpExt.AppID) > 0 || len(bidderImpExt.PlacementID) > 0 {
```
VeronikaSolovei9, line 103: Question: do you expect to have one of AppId and PlacementId or both of them? In description you mentioned `add the optional appid and placementid (which when used, must be used in conjunction)`

adapters/pangle/pangle.go
chshpra: Our service handles Prebid requests and other ordinary OpenRTB requests differently. I'm hoping that we could automatically distinguish that by the request itself, not by some other manual configurations.

static/bidder-params/pangle.json
chshpra: Placementid could exceeds an int32's length, appid don't. they're both set to be of string type because of a historical problem on our side. We prefer to make this compatible with the existing codes.

adapters/pangle/pangle.go
chshpra: yeah this is a negligence. I have modified this part. Please take a look at it again. btw, isn't there a bidder param check before going into this adapter? can I assume that it's always both or none here at this part of codes?

SyntaxNode: >  btw, isn't there a bidder param check before going into this adapter? can I assume that it's always both or none here at this part of codes?

Yes. The bidder param check isn't part of the json tests, but is guaranteed to run by PBS-Core. You may assume the check passed in your bidder code.

bsardo (approved):
LGTM

chshpra:
hi @bsardo, if there is no more problems, could we merge this pr now? our clients are waiting for this version of new adapter. thanks!

bsardo:
Hi @chshpra. I'm just waiting for another reviewer to sign off on it; we require two approvals before merging. I'll try to get someone else to look at this today.

chshpra:
@bsardo I see. thanks!

VeronikaSolovei9:
LGTM

VeronikaSolovei9 (approved):

PR #1853, opened 2021-05-18
Debug override header [Unit tests tbd]

VeronikaSolovei9:


exchange/exchange.go
```diff
@@ -163,8 +163,8 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 	requestDebugInfo := getDebugInfo(r.BidRequest, requestExt)
 
-	debugInfo := requestDebugInfo && r.Account.DebugAllow
-	debugLog.Enabled = debugLog.Enabled && r.Account.DebugAllow
+	debugInfo := debugLog.DebugOverride || requestDebugInfo && r.Account.DebugAllow
```
hhhjort, line 166: Why do you have () in 167, but not 166? If I am parsing it properly, I think the logic works out the same either way, but it is confusing as it looks like the logic should be different in the two cases.

VeronikaSolovei9: Yes, good catch, fixed.

hhhjort (dismissed):
LGTM

hhhjort (dismissed):

bsardo (commented):
Looks good. Just one stylistic suggestion.

exchange/auction.go
```diff
@@ -17,14 +17,21 @@ import (
...
+	Enabled       bool
+	CacheType     prebid_cache_client.PayloadType
+	Data          DebugData
+	TTL           int64
+	CacheKey      string
+	CacheString   string
+	Regexp        *regexp.Regexp
+	DebugOverride bool
+	//little optimization, it stores value of debugLog.Enabled || debugLog.DebugOverride
+	DebugConsolidated bool
```
SyntaxNode, line 34: Consider renaming to avoid the comment, perhaps `DebugEnabledOrOverriden`. There's also a chance for deviation if `debugLog.Enabled` or `debugLog.DebugOverride` are changed. Do you think this would be better served as a method on the `debugLog` struct?

I'm not concerned with the runtime performance of or-ing two booleans.

exchange/auction_test.go
```diff
@@ -118,6 +118,55 @@ func TestCacheJSON(t *testing.T) {
...
 }
 
+func TestIsDebugOverrideEnabled(t *testing.T) {
+
+	type inTest struct {
+		debugHeader string
+		configToken string
+	}
+	type outTest struct {
+		result bool
```
SyntaxNode, line 128: IMHO there's no need to define a structure with just a single item. Since this is a test though, my opinion on this is weak. Do whatever you think looks best.

exchange/auction_test.go
```diff
@@ -118,6 +118,55 @@ func TestCacheJSON(t *testing.T) {
...
+		{
+			desc: "test debug header is present, config token is present, not equal",
+			in:   inTest{debugHeader: "TestToken123", configToken: "TestToken"},
+			out:  outTest{result: false},
+		},
+		{
+			desc: "test debug header is present, config token is present, equal",
+			in:   inTest{debugHeader: "TestToken", configToken: "TestToken"},
+			out:  outTest{result: true},
+		},
```
SyntaxNode, line 160: Consider adding one more test to make it explicit that the comparison is case sensitive.

exchange/exchange.go
```diff
@@ -164,8 +164,8 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 	requestDebugInfo := getDebugInfo(r.BidRequest, requestExt)
 
-	debugInfo := requestDebugInfo && r.Account.DebugAllow
-	debugLog.Enabled = debugLog.Enabled && r.Account.DebugAllow
+	debugInfo := debugLog.DebugOverride || (requestDebugInfo && r.Account.DebugAllow)
+	debugLog.Enabled = debugLog.DebugOverride || (debugLog.Enabled && r.Account.DebugAllow)
```
SyntaxNode, line 168: Is it necessary to factor in the override token for  `debugLog.Enabled` is there's also `DebugConsolidated`? Switching `DebugConsolidated` to a method might be a better approach here.

endpoints/openrtb2/video_auction.go
```diff
@@ -128,11 +128,13 @@ func (deps *endpointDeps) VideoAuctionEndpoint(w http.ResponseWriter, r *http.Re
...
 	debugLog := exchange.DebugLog{
-		Enabled:   strings.EqualFold(debugQuery, "true"),
-		CacheType: prebid_cache_client.TypeXML,
-		TTL:       cacheTTL,
-		Regexp:    deps.debugLogRegexp,
+		Enabled:       strings.EqualFold(debugQuery, "true"),
+		CacheType:     prebid_cache_client.TypeXML,
+		TTL:           cacheTTL,
+		Regexp:        deps.debugLogRegexp,
+		DebugOverride: exchange.IsDebugOverrideEnabled(r.Header.Get(exchange.DebugOverrideHeader), deps.cfg.Debug.OverrideToken),
```
SyntaxNode, line 135: To ensure I'm following correctly, we need to perform this check in the video endpoint because the video endpoint does it's own debug logging. The auction and amp endpoints don't require this since the logic in auction.go applies to all 3 openrtb2 endpoints, right?

exchange/bidder.go
```diff
@@ -176,12 +176,14 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb2.B
...
 		// - debugContextKey (url param) in true
 		// - account debug is allowed
 		// - bidder debug is allowed
-		if debugInfo := ctx.Value(DebugContextKey); debugInfo != nil && debugInfo.(bool) {
-			if accountDebugAllowed {
-				if bidder.config.DebugInfo.Allow {
+		debugInfo := ctx.Value(DebugContextKey)
+		if headerDebugAllowed || (debugInfo != nil && debugInfo.(bool)) {
+			if headerDebugAllowed || accountDebugAllowed {
+				if headerDebugAllowed || bidder.config.DebugInfo.Allow {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, modified.

config/config.go
```diff
@@ -985,6 +986,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("debug.timeout_notification.log", false)
 	v.SetDefault("debug.timeout_notification.sampling_rate", 0.0)
 	v.SetDefault("debug.timeout_notification.fail_only", false)
+	v.SetDefault("debug.override_token", "")
```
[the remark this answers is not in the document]
VeronikaSolovei9: Added

exchange/auction.go
```diff
@@ -47,6 +54,13 @@ func (d *DebugLog) BuildCacheString() {
 	d.CacheString = fmt.Sprintf("%s<Log>%s%s%s</Log>", xml.Header, d.Data.Request, d.Data.Headers, d.Data.Response)
 }
 
+func IsDebugOverrideEnabled(debugHeader, configOverrideToken string) bool {
+	if configOverrideToken != "" && debugHeader == configOverrideToken {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, done.

exchange/auction_test.go
```diff
@@ -118,6 +118,55 @@ func TestCacheJSON(t *testing.T) {
 	}
 }
 
+func TestIsDebugOverrideEnabled(t *testing.T) {
+
```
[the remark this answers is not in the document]
VeronikaSolovei9: Removed

exchange/auction.go
VeronikaSolovei9: Renamed. 
`There's also a chance for deviation if debugLog.Enabled or debugLog.DebugOverride are changed` -> DebugEnabledOrOverriden calculates per request. the only one place where we set value to this field is video_endpoint:
`debugLog.DebugEnabledOrOverridden = debugLog.Enabled || debugLog.DebugOverride`
I don't see any benefits to implement a method on the debugLog struct.

Originally there was `or` between two variables, but then it was extracted to separate field to optimize code.

exchange/auction_test.go
VeronikaSolovei9: yes, modified

endpoints/openrtb2/video_auction.go
VeronikaSolovei9: Correct. Now debug log is enabled for video endpoint only. For amp and auction we don't init debugLog and by default it's disabled. func (e *exchange) HoldAuction:
```
if debugLog == nil {
	debugLog = &DebugLog{Enabled: false}
}
```

exchange/exchange.go
VeronikaSolovei9: Yes, modified.

hhhjort (dismissed):

exchange/bidder.go
```diff
@@ -176,19 +176,25 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb2.B
...
-		if debugInfo := ctx.Value(DebugContextKey); debugInfo != nil && debugInfo.(bool) {
-			if accountDebugAllowed {
-				if bidder.config.DebugInfo.Allow {
-					seatBid.httpCalls = append(seatBid.httpCalls, makeExt(httpInfo))
-				} else {
-					debugDisabledWarning := errortypes.Warning{
-						WarningCode: errortypes.BidderLevelDebugDisabledWarningCode,
-						Message:     "debug turned off for bidder",
+		if headerDebugAllowed {
+			seatBid.httpCalls = append(seatBid.httpCalls, makeExt(httpInfo))
```
mansinahar, line 184: Please consider asserting that this information exists when `headerDebugAllowed` is set to `true` in one of the `requestBid` test cases

VeronikaSolovei9: Mansi, this is sort of done in test `TestRequestBidRemovesSensitiveHeaders`, line 183. 
It works because `accountDebugAllowed` for this test case is set to true. In this case it works similarly to the case where `headerDebugAllowed` is set to true. There are more cases to test debug options combination and results in `TestDebugBehavior` in exchange_test.go file.
Do you think I need to add additional test (or modify existing one) in `requestBid` test cases?

mansinahar: @VeronikaSolovei9 Thanks for that explanation. However, even though both `accountDebugAllowed` and `headerDebugAllowed` kind of have the same purpose in this function, they are both different flags and so it would be a good idea to assert the functionality based on explicit values of both the flags

mansinahar (approved):

mansinahar (approved):

PR #1855, opened 2021-05-18
New Adapter: BrightMountainMedia

BrightMountainMediaInc:


adapters/bmtm/brightmountainmedia.go
```diff
@@ -0,0 +1,160 @@
...
+	"strconv"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type BrightMountainMediaAdapter struct {
```
SyntaxNode, line 16: We prefer to use `adapter` as the name for this struct. There is no need to repeat the package name. There is no need to export the type using an uppercase letter.

BrightMountainMediaInc: Please have a look @SyntaxNode

adapters/bmtm/brightmountainmedia_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderBmtm, config.Adapter{
+		Endpoint: "https://one.elitebidder.com/api/pbs"})
```
VeronikaSolovei9, line 13: Nitpick: you can use test url here, not real one. Something like "https://testbmtm.com"

VeronikaSolovei9 (commented):
Just a minor comment, other than that everything lgtm

adapters/bmtm/brightmountainmedia_test.go
BrightMountainMediaInc: Please have a look @VeronikaSolovei9

VeronikaSolovei9:
Can you please also add docs PR to this repository: https://github.com/prebid/prebid.github.io ? Please include bidder information, bidder parameters and description. This information will be deployed to https://docs.prebid.org/dev-docs/pbs-bidders.html

Here is the example:
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/appnexus.md

adapters/bmtm/brightmountainmedia.go
```diff
@@ -0,0 +1,153 @@
...
+	if ortbImp.Banner == nil && ortbImp.Video == nil {
+		return nil, &errortypes.BadInput{
+			Message: fmt.Sprintf("For Imp ID %s Banner or Video is undefined", ortbImp.ID),
+		}
+	}
+
+	var bidderExt adapters.ExtImpBidder
+	if err := json.Unmarshal(ortbImp.Ext, &bidderExt); err != nil {
+		return nil, &errortypes.BadInput{
+			Message: "Error unmarshalling ExtImpBidder",
```
mansinahar, line 53: Please append the original error to the error message as well. You can do so as follows:
```
Message: fmt.Sprintf("Error unmarshalling ExtImpBidder: %s", err.Error()),
```

adapters/bmtm/brightmountainmedia.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+
+	var bmtmExt openrtb_ext.ImpExtBmtm
+	if err := json.Unmarshal(bidderExt.Bidder, &bmtmExt); err != nil {
+		return nil, &errortypes.BadInput{
+			Message: "Error unmarshalling ExtImpBmtm",
```
mansinahar, line 60: Please append the original error to the error message here. Similar as above

adapters/bmtm/brightmountainmedia_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderBmtm, config.Adapter{
+		Endpoint: "https://example.com/api/pbs"})
+
+	if buildErr != nil {
```
mansinahar, line 15: Please consider changing this to:
```
assert.NoError(t, buildErr, "Builder returned unexpected error")
```

adapters/bmtm/params_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestValidParams(t *testing.T) {
+	validator, err := openrtb_ext.NewBidderParamsValidator("../../static/bidder-params")
+	if err != nil {
```
mansinahar, line 12: Please consider changing this to using `assert.NoError` as well. Similar to above

adapters/bmtm/params_test.go
```diff
@@ -0,0 +1,44 @@
...
+	for _, validParam := range validParams {
+		if err := validator.Validate(openrtb_ext.BidderBmtm, json.RawMessage(validParam)); err != nil {
+			t.Errorf("Schema rejected brightMountainMedia params: %s", validParam)
+		}
+	}
+}
+
+func TestInvalidParams(t *testing.T) {
+	validator, err := openrtb_ext.NewBidderParamsValidator("../../static/bidder-params")
+	if err != nil {
```
mansinahar, line 25: Please consider changing this to using `assert.NoError` as well. Similar to above

adapters/bmtm/params_test.go
```diff
@@ -0,0 +1,44 @@
...
+		t.Fatalf("Failed to fetch the json-schemas. %v", err)
+	}
+
+	for _, invalidParam := range invalidParams {
+		if err := validator.Validate(openrtb_ext.BidderBmtm, json.RawMessage(invalidParam)); err == nil {
+			t.Errorf("Schema allowed unexpected params: %s", invalidParam)
```
mansinahar, line 31: Please consider changing this to using `assert.Error` here

adapters/bmtm/params_test.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func TestValidParams(t *testing.T) {
+	validator, err := openrtb_ext.NewBidderParamsValidator("../../static/bidder-params")
+	if err != nil {
+		t.Fatalf("Failed to fetch the json-schemas. %v", err)
+	}
+
+	for _, validParam := range validParams {
+		if err := validator.Validate(openrtb_ext.BidderBmtm, json.RawMessage(validParam)); err != nil {
```
mansinahar, line 17: Please consider changing this to using `assert.NoError` as well. Similar to above

mansinahar (commented):
Other than the few minor comments I left above, LGTM in general!

adapters/bmtm/brightmountainmedia_test.go
```diff
@@ -1,19 +1,21 @@
...
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestJsonSamples(t *testing.T) {
 	bidder, buildErr := Builder(openrtb_ext.BidderBmtm, config.Adapter{
 		Endpoint: "https://example.com/api/pbs"})
 
 	if buildErr != nil {
```
mansinahar, line 17: With `assert.NoError` you don't need the `if` condition here anymore. `assert.NoError` will make sure that `buildErr` is `nil` and if it isn't then fail the test with the error message you're passing to it as the last param

adapters/bmtm/params_test.go
```diff
@@ -2,33 +2,35 @@ package bmtm
...
 
 func TestValidParams(t *testing.T) {
 	validator, err := openrtb_ext.NewBidderParamsValidator("../../static/bidder-params")
 	if err != nil {
-		t.Fatalf("Failed to fetch the json-schemas. %v", err)
+		assert.NoError(t, err, fmt.Sprintf("Failed to fetch the json-schemas: %s", err.Error()))
 	}
 
 	for _, validParam := range validParams {
 		if err := validator.Validate(openrtb_ext.BidderBmtm, json.RawMessage(validParam)); err != nil {
```
mansinahar, line 19: Please remove the `err != nil` check as suggested above. It is no longer needed with `assert.NoError`. You can just have the following:
```
err := validator.Validate(openrtb_ext.BidderBmtm, json.RawMessage(validParam))
assert.NoError(t, err, fmt.Sprintf("Schema rejected brightMountainMedia params: %s", validParam))
```

adapters/bmtm/params_test.go
```diff
@@ -2,33 +2,35 @@ package bmtm
...
-			t.Errorf("Schema rejected brightMountainMedia params: %s", validParam)
+			assert.NoError(t, err, fmt.Sprintf("Schema rejected brightMountainMedia params: %s", validParam))
 		}
 	}
 }
 
 func TestInvalidParams(t *testing.T) {
 	validator, err := openrtb_ext.NewBidderParamsValidator("../../static/bidder-params")
 	if err != nil {
```
mansinahar, line 27: Same as above. You can remove the `err != nil` check

adapters/bmtm/params_test.go
```diff
@@ -2,33 +2,35 @@ package bmtm
...
-		t.Fatalf("Failed to fetch the json-schemas. %v", err)
+		assert.NoError(t, err, fmt.Sprintf("Failed to fetch the json-schemas: %s", err.Error()))
 	}
 
 	for _, invalidParam := range invalidParams {
 		if err := validator.Validate(openrtb_ext.BidderBmtm, json.RawMessage(invalidParam)); err == nil {
```
mansinahar, line 32: You can remove the `err == nil` check here but you'll have to change the assert function below to `assert.Error` because in case of invalid params you do expect an error to occur

BrightMountainMediaInc:
Please have a look @mansinahar

mansinahar (approved):

adapters/bmtm/brightmountainmedia_test.go
```diff
@@ -11,12 +11,7 @@ import (
...
 func TestJsonSamples(t *testing.T) {
-	bidder, buildErr := Builder(openrtb_ext.BidderBmtm, config.Adapter{
-		Endpoint: "https://example.com/api/pbs"})
-
-	if buildErr != nil {
-		assert.NoError(t, buildErr, fmt.Sprintf("Builder returned unexpected error: %s", buildErr.Error()))
-	}
-
+	bidder, buildErr := Builder(openrtb_ext.BidderBmtm, config.Adapter{Endpoint: "https://example.com/api/pbs"})
+	assert.NoError(t, buildErr, fmt.Sprintf("Builder returned unexpected error: %s", buildErr.Error()))
```
mansinahar, line 15: Looks like appending the actual error message here is causing the tests to panic because we expect `buildErr` to be nil and we're trying to pass `buildErr.Error()` to the `assert.NoError` function. Can you please remove that here and in other places to avoid the test to panic? Also, the `assert` library itself will print the actual error message in case `buildErr` ends up being not nil so we don't need to explicitly do that

SyntaxNode: Please use the standard test function:
```
  bidder, buildErr := Builder(openrtb_ext.Bidder{Bidder}, config.Adapter{
    Endpoint: "http://whatever.url"})
  
  if buildErr != nil {
    t.Fatalf("Builder returned unexpected error %v", buildErr)
  }
  
  adapterstest.RunJSONBidderTest(t, "{bidder}test", bidder)
}
```

mansinahar (approved):

adapters/bmtm/brightmountainmedia_test.go
mansinahar: @SyntaxNode That's how they had it before. I actually suggested them to change it to using the `assert` library because that makes the code a little more concise and that's what we use in the core code for tests. However, I understand that this is how all other adapters have it currently. This was more intended to perhaps make that change moving forward. For now, I have approved the PR since this is not critical. 

Sorry about the different directions @BrightMountainMediaInc

VeronikaSolovei9 (approved):

PR #1857, opened 2021-05-21
Bugfix for applyCategoryMapping

VeronikaSolovei9:


exchange/exchange_test.go
```diff
@@ -2547,6 +2547,108 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
 
 }
 
+func TestCategoryMappingTwoBiddersManyBidsEachNoCategorySamePrice(t *testing.T) {
```
SyntaxNode, line 2550: The large amount of test bids and iteration of 100 to workaround the random aspect makes it very hard to understand the test setup. Would this be a good time to isolate the random library and substitute it for just this test (for now) for a more targeted setup?

exchange/exchange.go
```diff
@@ -756,11 +756,31 @@ func applyCategoryMapping(ctx context.Context, requestExt *openrtb_ext.ExtReques
...
+
+							//Find index of bid to remove
+							dupeBidIndex := -1
+							for i, bid := range oldSeatBid.bids {
+								if bid.bid.ID == dupe.bidID {
+									dupeBidIndex = i
+									break
+								}
+							}
+							if dupeBidIndex != -1 {
```
SyntaxNode, line 777: What does it mean if we determine a dedupe is required but the bid index can't be found?

exchange/exchange.go
```diff
@@ -756,11 +756,31 @@ func applyCategoryMapping(ctx context.Context, requestExt *openrtb_ext.ExtReques
...
+								if dupeBidIndex < len(oldSeatBid.bids)-1 {
+									oldSeatBid.bids = append(oldSeatBid.bids[:dupeBidIndex], oldSeatBid.bids[dupeBidIndex+1:]...)
+								} else if dupeBidIndex == len(oldSeatBid.bids)-1 {
+									oldSeatBid.bids = oldSeatBid.bids[:len(oldSeatBid.bids)-1]
```
SyntaxNode, line 781: Is the else case necessary? I think append might gracefully handle nil slices or slices with no elements.

SyntaxNode (commented):
Thank you for investigating this edge case. Nice find.

exchange/exchange.go
```diff
@@ -756,11 +756,31 @@ func applyCategoryMapping(ctx context.Context, requestExt *openrtb_ext.ExtReques
...
 						if len(oldSeatBid.bids) == 1 {
 							seatBidsToRemove = append(seatBidsToRemove, dupe.bidderName)
-							rejections = updateRejections(rejections, dupe.bidID, "Bid was deduplicated")
 						} else {
-							oldSeatBid.bids = append(oldSeatBid.bids[:dupe.bidIndex], oldSeatBid.bids[dupe.bidIndex+1:]...)
+							// This is a very unique, but still possible case where bid needs to be removed from already processed bidder
+							// This happens when current processing bidder has a bid that has same deduplication key as a bid from already processed bidder
+							// and already processed bid was selected to be removed
+							// See example of input data in unit test `TestCategoryMappingTwoBiddersManyBidsEachNoCategorySamePrice`
+							// Need to remove bid by name, not index in this case
```
guscarreon, line 767: Can we make the entirety of the else branch into its own function?
```
762              } else {
763                  // This is a very unique, but still possible case where bid needs to be removed from already processed bidder
764                  // This happens when current processing bidder has a bid that has same deduplication key as a bid from already processed bidder
765                  // and already processed bid was selected to be removed
766                  // See example of input data in unit test `TestCategoryMappingTwoBiddersManyBidsEachNoCategorySamePrice`
767                  // Need to remove bid by name, not index in this case
    +                removeByName(dupe.bidID, oldSeatBid.bids)
768 -
769 -                //Find index of bid to remove
770 -                dupeBidIndex := -1
771 -                for i, bid := range oldSeatBid.bids {
772 -                    if bid.bid.ID == dupe.bidID {
773 -                        dupeBidIndex = i
774 -                        break
775 -                    }
776 -                }
777 -                if dupeBidIndex != -1 {
778 -                    if dupeBidIndex < len(oldSeatBid.bids)-1 {
779 -                        oldSeatBid.bids = append(oldSeatBid.bids[:dupeBidIndex], oldSeatBid.bids[dupeBidIndex+1:]...)
780 -                    } else if dupeBidIndex == len(oldSeatBid.bids)-1 {
781 -                        oldSeatBid.bids = oldSeatBid.bids[:len(oldSeatBid.bids)-1]
782 -                    }
783 -                }
784              }
785          }
786          delete(res, dupe.bidID)
exchange/exchange.go
```
I find the `TestCategoryMappingTwoBiddersManyBidsEachNoCategorySamePrice` pretty solid, but if you agree to move the else statement to its own function, can we also add a unit test for it? I don't think we should remove the `TestCategoryMappingTwoBiddersManyBidsEachNoCategorySamePrice` just add a test for `removeByName(string, []*pbsOrtbBid)` (if you decide to name it that).

exchange/exchange.go
VeronikaSolovei9: This should not be the case. But if this happens then there is nothing to delete from list and we just skip it.

exchange/exchange.go
VeronikaSolovei9: `else if` is the case when we need to remove last element from list. First `if` is the case when we need to remove element from the middle of array.

exchange/exchange.go
VeronikaSolovei9: Yes, good idea, refactored!

exchange/exchange_test.go
VeronikaSolovei9: Yes, refactored.

exchange/exchange.go
```diff
@@ -99,6 +99,16 @@ func (big *bidIDGenerator) New() (string, error) {
 	return rawUuid.String(), err
 }
 
+type BooleanGenerator interface {
```
SyntaxNode, line 102: IMHO - The type name is too generic to understand it's intent. Consider a more specific name such as `Chance`, `Chancer`, or `CoinFlipper`. I also recommend to start with a lower case letter since it's not used my other packages there is no need to export it.

exchange/exchange_test.go
```diff
@@ -1800,6 +1800,14 @@ func (big *mockBidIDGenerator) New() (string, error) {
 
 }
 
+type MockRandomBooleanGenerator struct {
```
SyntaxNode, line 1803: Nitpick: Consider using the [term 'Fake' instead of 'Mock here](https://stackoverflow.com/questions/346372/whats-the-difference-between-faking-mocking-and-stubbing#:~:text=Fake%3A%20a%20class%20that%20implements,fixed%20data%20and%20no%20logic.&text=Mock%3A%20a%20class%20that%20implements,have%20been%20called%2Fnot%20called.).

exchange/exchange_test.go
```diff
@@ -2547,6 +2555,122 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+		includeWinners:   true,
+	}
+
+	requestExt.Prebid.Targeting.DurationRangeSec = []int{30}
+	requestExt.Prebid.Targeting.IncludeBrandCategory.WithCategory = false
+
+	cats1 := []string{"IAB1-3"}
+	cats2 := []string{"IAB1-4"}
+
+	bidApn1_1 := openrtb2.Bid{ID: "bid_idApn1_1", ImpID: "imp_idApn1_1", Price: 10.0000, Cat: cats1, W: 1, H: 1}
```
SyntaxNode, line 2578: This is a non trivial test since it needs to activate a very specific branch. Please add a comment describing the test setup, mentioning which bids we expect to be removed due to deduplication.

exchange/exchange_test.go
```diff
@@ -2547,6 +2555,122 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+	if bidsFromFirstBidder.bids != nil {
+		totalNumberOfbids += len(bidsFromFirstBidder.bids)
+	}
+
+	if bidsFromSecondBidder.bids != nil {
+		totalNumberOfbids += len(bidsFromSecondBidder.bids)
+	}
+
+	assert.Equal(t, 2, totalNumberOfbids, "2 bids total should be returned")
+	assert.Len(t, rejections, 2, "10 bids should be deduplicated")
```
SyntaxNode, line 2630: Typo in the string description.

exchange/exchange_test.go
```diff
@@ -2547,6 +2555,122 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+
+	if bidsFromFirstBidder.bids != nil {
+		totalNumberOfbids += len(bidsFromFirstBidder.bids)
+	}
+
+	if bidsFromSecondBidder.bids != nil {
+		totalNumberOfbids += len(bidsFromSecondBidder.bids)
+	}
+
+	assert.Equal(t, 2, totalNumberOfbids, "2 bids total should be returned")
```
SyntaxNode, line 2629: How do we know the correct bids were removed? Consider validating the bid ids.

exchange/exchange_test.go
```diff
@@ -2547,6 +2555,122 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+	assert.Len(t, rejections, 2, "10 bids should be deduplicated")
+
+}
+
+func TestRemoveBidByName(t *testing.T) {
+	cats1 := []string{"IAB1-3"}
```
SyntaxNode, line 2635: Please consider using table driven tests, where one case addresses one branch of the `removeBidByName` function.

exchange/exchange_test.go
```diff
@@ -2547,6 +2555,122 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+		&bid1_Apn1_4,
+		&bid1_Apn1_5,
+		&bid1_Apn1_6,
+	}
+
+	seatBidApn1 := &pbsOrtbSeatBid{bids: innerBidsApn1, currency: "USD"}
+
+	//Please keep these calls in this order
+	removeBidByName(seatBidApn1, "bid_idApn1_3")
+	assert.Len(t, seatBidApn1.bids, 5, "Bids number is incorrect after removing element from the middle")
```
SyntaxNode, line 2664: Similar comment from the other test, how do we know the correct one was removed?

exchange/exchange.go
VeronikaSolovei9: Renamed to `deduplicateBidBooleanGenerator`, hope it makes sense

exchange/exchange_test.go

exchange/exchange_test.go
VeronikaSolovei9: Added

exchange/exchange_test.go
VeronikaSolovei9: Fixed

exchange/exchange_test.go
VeronikaSolovei9: Added bid ids validation

exchange/exchange_test.go
VeronikaSolovei9: Refactored

exchange/exchange_test.go
VeronikaSolovei9: Added checks

exchange/exchange.go
SyntaxNode: That works. Another option might be `deduplicateChanceGenerator`, but now I'm splitting hairs.

exchange/exchange.go
```diff
@@ -99,6 +99,16 @@ func (big *bidIDGenerator) New() (string, error) {
 	return rawUuid.String(), err
 }
 
+type deduplicateBidBooleanGenerator interface {
+	New() bool
```
SyntaxNode, line 103: Consider renaming `New` to `Generate` to pair with the "Generator" part of the struct name.

exchange/exchange.go
```diff
@@ -798,6 +813,24 @@ func applyCategoryMapping(ctx context.Context, requestExt *openrtb_ext.ExtReques
 	return res, seatBids, rejections, nil
 }
 
+func removeBidByName(seatBid *pbsOrtbSeatBid, bidID string) {
```
SyntaxNode, line 816: Should this be `removeBidByID` (and the associated test name)?

exchange/exchange_test.go
```diff
@@ -2540,13 +2548,164 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+		}
+	}
+}
+
+func TestCategoryMappingTwoBiddersManyBidsEachNoCategorySamePrice(t *testing.T) {
+	// This test covers a very rare de-duplication case where bid needs to be removed from already processed bidder
+	// This happens when current processing bidder has a bid that has same de-duplication key as a bid from already processed bidder
+	// and already processed bid was selected to be removed
+
+	//In this test case bids bid_idApn1_1 and bid_idApn1_2 will be removed due to hardcoded "fakeRandomDeduplicateBidBooleanGenerator{true}"
```
SyntaxNode, line 2560: Thank you. If there a reason we need `bid_idApn1_1` and `bid_idApn1_2` to model this situation? Would just `bid_idApn1_1` also be a valid test? Or do we need to the 2 placements for the bidder to already be processed? which if so, could you make a bit clearer in the description?

exchange/exchange_test.go
```diff
@@ -2540,13 +2548,164 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+
+func TestRemoveBidByName(t *testing.T) {
+	cats1 := []string{"IAB1-3"}
+
+	bidApn1_1 := openrtb2.Bid{ID: "bid_idApn1_1", ImpID: "imp_idApn1_1", Price: 10.0000, Cat: cats1, W: 1, H: 1}
+	bidApn1_2 := openrtb2.Bid{ID: "bid_idApn1_2", ImpID: "imp_idApn1_2", Price: 20.0000, Cat: cats1, W: 1, H: 1}
+	bidApn1_3 := openrtb2.Bid{ID: "bid_idApn1_3", ImpID: "imp_idApn1_3", Price: 10.0000, Cat: cats1, W: 1, H: 1}
+	bidApn1_4 := openrtb2.Bid{ID: "bid_idApn1_4", ImpID: "imp_idApn1_4", Price: 20.0000, Cat: cats1, W: 1, H: 1}
+	bidApn1_5 := openrtb2.Bid{ID: "bid_idApn1_5", ImpID: "imp_idApn1_5", Price: 10.0000, Cat: cats1, W: 1, H: 1}
+	bidApn1_6 := openrtb2.Bid{ID: "bid_idApn1_6", ImpID: "imp_idApn1_6", Price: 20.0000, Cat: cats1, W: 1, H: 1}
```
SyntaxNode, line 2653: I see why you have multiple bids to test the positions (before, middle, after) - but I think the same test could be accomplished with just 3 bids instead of 6.

exchange/exchange_test.go
```diff
@@ -2540,13 +2548,164 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+	bid1_Apn1_5 := pbsOrtbBid{&bidApn1_5, "video", nil, &openrtb_ext.ExtBidPrebidVideo{Duration: 30}, nil, 0, false, ""}
+	bid1_Apn1_6 := pbsOrtbBid{&bidApn1_6, "video", nil, &openrtb_ext.ExtBidPrebidVideo{Duration: 30}, nil, 0, false, ""}
+
+	type aTest struct {
+		desc string
+		in   string
+	}
+	testCases := []aTest{
+		{
+			desc: "incorrect result after removing element from the middle",
```
SyntaxNode, line 2668: I recommend updating the descriptions to remove the "incorrect" text, as that implies to me the test is looking for an error. ex: "remove element from middle".

exchange/exchange_test.go
```diff
@@ -2540,13 +2548,164 @@ func TestCategoryMappingTwoBiddersOneBidEachNoCategorySamePrice(t *testing.T) {
...
+			&bid1_Apn1_4,
+			&bid1_Apn1_5,
+			&bid1_Apn1_6,
 		}
 
+		seatBidApn1 := &pbsOrtbSeatBid{bids: innerBidsApn1, currency: "USD"}
+
+		removeBidByName(seatBidApn1, test.in)
+		assert.Len(t, seatBidApn1.bids, 5, test.desc)
+		assert.False(t, checkBidPresentInArray(seatBidApn1.bids, test.in), "Bid should not be present in result")
```
SyntaxNode, line 2695: I think it's important to check we have the exact bids we expect. Rather than checking for the length and absence, I recommend testing for the complete collection. Something like:

```
{
	desc: "incorrect result after removing element from the middle",
	in:   "bid_idApn1_2",
	out:  []*pbsOrtbBid{&bid1_Apn1_1, &bid1_Apn1_3},
},
```

with the assertion
```
assert.ElementsMatch(t, test.out, seatBidApn1.bids, test.desc)
```

In which case the `checkBidPresentInArray` helper is no longer needed.

exchange/exchange.go
VeronikaSolovei9: Ok, renamed

exchange/exchange.go
VeronikaSolovei9: Renamed

exchange/exchange.go
VeronikaSolovei9: Renamed

exchange/exchange_test.go
VeronikaSolovei9: We need to have more than one bid in each bidder to test how we remove element from array. If there is only one element in bidder, then we remove the entire bidder. 
I added explicit comment for this.

exchange/exchange_test.go
VeronikaSolovei9: Right, removed redundant bids

exchange/exchange_test.go
VeronikaSolovei9: refactored

exchange/exchange_test.go
VeronikaSolovei9: Good idea! Modified!

guscarreon (approved):
LGTM

SyntaxNode (approved):

PR #1859, opened 2021-05-21
GDPR: require host specify default value

bsardo:
`gdpr.default_value` is now required and may be specified through a config file or as an env var. If it is not provided, we will fail hard reporting a validation error to the host.

All other validation looks at the unmarshaled viper data but I didn't want to convert `default_value` to a pointer so I made use of the viper `IsSet` method to detect presence.

Given the need for the viper instance now to perform validation, I originally tried to keep things simple by placing this validation logic at the top of the top-level config validate function instead of passing the viper instance down into the GDPR validate method. Ultimately I ended up passing the viper instance down so that all GDPR-related config validation is kept together.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -93,7 +93,7 @@ type HTTPClient struct {
 	IdleConnTimeout     int `mapstructure:"idle_connection_timeout_seconds"`
 }
 
-func (cfg *Configuration) validate() []error {
+func (cfg *Configuration) validate(v *viper.Viper) []error {
 	var errs []error
 	errs = cfg.AuctionTimeouts.validate(errs)
 	errs = cfg.StoredRequests.validate(errs)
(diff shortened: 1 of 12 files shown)


config/config.go
```diff
@@ -943,7 +946,6 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("amp_timeout_adjustment_ms", 0)
 	v.SetDefault("gdpr.enabled", true)
 	v.SetDefault("gdpr.host_vendor_id", 0)
-	v.SetDefault("gdpr.usersync_if_ambiguous", false)
```
VeronikaSolovei9, line 946: Should it be "gdpr.default_value" instead?

endpoints/auction_test.go
```diff
@@ -346,6 +346,7 @@ func TestCacheVideoOnly(t *testing.T) {
 	ctx := context.TODO()
 	v := viper.New()
 	config.SetupViper(v, "")
+	v.Set("gdpr.default_value", false)
```
hhhjort, line 349: shouldn't it be set to "0" or "1"?

hhhjort (approved):
LGTM

VeronikaSolovei9 (approved):

PR #1861, opened 2021-05-25
New Adapter: AlgoriX

Bugxyb:
Add AlgoriX Adapter to Prebid-Server(Go), and We add Prebid-Server-Java in progress(https://github.com/prebid/prebid-server-java/pull/1275)


adapters/algorix/algorix.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/http"
+	"text/template"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type AlgoriXAdapter struct {
```
SyntaxNode, line 17: Please use the name `adapter`. There is no need to repeat the package name and no need to export with a uppercase letter.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse endpoint url template: %v", err)
+	}
+	bidder := &AlgoriXAdapter{
+		EndpointTemplate: *template,
+	}
+	return bidder, nil
+}
+
+// MakeRequests Make Requests
```
SyntaxNode, line 33: IMHO This comment doesn't provide any useful context. If you'd like to have a comment, please make it more descriptive. The `MakeRequests` method is common with the app so if you don't want to add comments, there is no pressure to do so.

SyntaxNode:
Doc PR: https://github.com/prebid/prebid.github.io/pull/2972

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,171 @@
...
+	if err == nil {
+		adapterRequests = append(adapterRequests, adapterRequest)
+	} else {
+		errs = append(errs, err)
+	}
+	return adapterRequests, errs
+}
+
+func (adapter *Adapter) makeRequest(request *openrtb2.BidRequest) (*adapters.RequestData, error) {
+	if len(request.Imp) == 0 {
```
guscarreon, line 48: We could remove this statement because prebid server core does this check upstream. Therefore, no request with an empty `Imp` array will get to this point.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,171 @@
...
+func parseAlgoriXExt(request *openrtb2.BidRequest) *openrtb_ext.ExtImpAlgorix {
+	var extImpAlgoriX openrtb_ext.ExtImpAlgorix
+	for _, imp := range request.Imp {
+		var extBidder adapters.ExtImpBidder
+		err := json.Unmarshal(imp.Ext, &extBidder)
+		if err != nil {
+			continue
+		}
+		err = json.Unmarshal(extBidder.Bidder, &extImpAlgoriX)
+		if err != nil || len(extImpAlgoriX.Sid) == 0 || len(extImpAlgoriX.Token) == 0 {
```
guscarreon, line 92: File `static/bidder-params/algorix.json` defines a `"minLength": 1` for both parameters which makes prebid server core to discard `Imp[i].Ext` entries whenever these two fields are empty. Therefore, we can discard these two checks:
```
91         err = json.Unmarshal(extBidder.Bidder, &extImpAlgoriX)
92 -       if err != nil || len(extImpAlgoriX.Sid) == 0 || len(extImpAlgoriX.Token) == 0 {
   +       if err != nil {
93             continue
94         }
adapters/algorix/algorix.go
```

adapters/algorix/algorixtest/params/race/banner.json
```diff
@@ -0,0 +1,4 @@
+{
+  "sid": "testSid",
```
guscarreon, line 2: Prebid server core is not using these files to assert for shared memory race conditions. This file can be entirely removed.

adapters/algorix/algorixtest/params/race/native.json
```diff
@@ -0,0 +1,4 @@
+{
+  "sid": "testSid",
```
guscarreon, line 2: Prebid server core is not using these files to assert for shared memory race conditions. This file can be entirely removed.

adapters/algorix/algorixtest/params/race/video.json
```diff
@@ -0,0 +1,4 @@
+{
+  "sid": "testSid",
```
guscarreon, line 2: Prebid server core is not using these files to assert for shared memory race conditions. This file can be entirely removed.

adapters/algorix/algorixtest/supplemental/no_imp_error.json
```diff
@@ -0,0 +1,13 @@
+{
+  "mockBidRequest": {
```
guscarreon, line 2: Requests with empty `Imp` arrays are discarded upstream by prebid server core, no need to include this test.

guscarreon (commented):
Looks pretty solid. Some minor comments.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,167 @@
...
+				bannerCopy.H = &firstFormat.H
+				request.Imp[i].Banner = &bannerCopy
+			}
+		}
+	}
+}
+
+func (a *adapter) MakeBids(internalRequest *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if response.StatusCode == http.StatusNoContent {
+		return nil, nil
```
VeronikaSolovei9, line 120: Should it be error here?

adapters/algorix/algorix_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderAlgorix, config.Adapter{
+		Endpoint: "https://xyz.svr-algorix.com/rtb/sa?sid={{.SourceId}}&token={{.AccountID}}"})
```
VeronikaSolovei9, line 14: This can be test url like "test.com"

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,167 @@
...
+			})
+		}
+	}
+	return bidResponse, nil
+}
+
+// getBidType get Bid Type default: BannerType
+func getBidType(impId string, imps []openrtb2.Imp) openrtb_ext.BidType {
+	for _, imp := range imps {
+		if imp.ID == impId {
```
VeronikaSolovei9, line 157: Is this check necessary here?

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,166 @@
...
+	"fmt"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/macros"
+	"net/http"
+	"text/template"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
```
SyntaxNode, line 14: Nitpick: It's common to arrange the framework imports in the grouping at the top and all other imports in the grouping at the bottom, ie:

```
import (
	"encoding/json"
	"fmt"
	"net/http"
	"text/template"

	"github.com/prebid/prebid-server/macros"
)
```

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+type adapter struct {
+	EndpointTemplate template.Template
+}
+
+// Builder builds a new instance of the Foo adapter for the given bidder with the given config.
```
SyntaxNode, line 21: Please replace "Foo" with "AlgoriX" in the comment.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,166 @@
...
+
+	return &adapters.RequestData{
+		Method:  "POST",
+		Uri:     endPoint,
+		Body:    reqBody,
+		Headers: headers,
+	}, nil
+}
+
+//parseAlgoriXExt parse AlgoriX Ext
```
SyntaxNode, line 77: Nitpick: The comment isn't particularly descriptive, and it doesn't need to be in my opinion since the name is straightforward. Consider removing the comment.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,166 @@
...
+			continue
+		}
+		err = json.Unmarshal(extBidder.Bidder, &extImpAlgoriX)
+		if err != nil {
+			continue
+		}
+		return &extImpAlgoriX
+	}
+	return nil
+}
```
SyntaxNode, line 93: This algorithm is searching for the first Imp with valid parameters and then using those for the entire request. I have two recommendations:

1 - Return an error if the unmarshal fails for any Imp. This indicates an invalid request and is a good reason to return an error. In practice, this won't happen since the core logic of Prebid Server will perform this check earlier in the request pipeline and return a fatal error. IMHO it's best to match Prebid Server core's behavior here.

2 - Please make it clear by adding a note to your separate documentation PR that only the parameters of the first Imp matter. The rest are ignored by this adapter. Consider adding a json test for this situation as well to provide an example to publishers.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,166 @@
...
+
+// getEndPoint get Endpoint Target
```
SyntaxNode, line 95: Similar nitpick as before, this comment is not necessary.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// preProcess process Request check and modify banner imp
+func preProcess(request *openrtb2.BidRequest) {
+	for i := range request.Imp {
+		if request.Imp[i].Banner != nil {
+			banner := *request.Imp[i].Banner
+			if (banner.W == nil || banner.H == nil || *banner.W == 0 || *banner.H == 0) && len(banner.Format) > 0 {
+				firstFormat := banner.Format[0]
+				bannerCopy := *request.Imp[i].Banner
```
SyntaxNode, line 108: You already made a copy with the previous line `banner := *request.Imp[i].Banner`. Two copies are not necessary. Please just use one.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,166 @@
...
+		if imp.ID == impId {
+			if imp.Native != nil {
+				return openrtb_ext.BidTypeNative
+			}
+			if imp.Video != nil {
+				return openrtb_ext.BidTypeVideo
+			}
+		}
+	}
+	return openrtb_ext.BidTypeBanner
```
SyntaxNode, line 165: This is an anti-pattern. Does your bidding server include the bid type in it's response?

static/bidder-info/algorix.yaml
```diff
@@ -0,0 +1,8 @@
+maintainer:
+  email: "xunyunbo@algorix.co"
+capabilities:
```
SyntaxNode, line 3: Just to confirm, is AlgoriX not registered with the IAB Europe for GDPR?

Bugxyb: AlgoriX not registered with the IAB Europe for GDPR.

adapters/algorix/algorix.go
Bugxyb: We change the code. only check and fetch first Imp extBidder, ignored rest imp.
also we add json test example(two imps and second is invalid, Prebid server adapter will ignored it. and we add this note for Doc PR

static/bidder-info/algorix.yaml
SyntaxNode: Fantastic. Kindly please update your adapter in the future when you have a GDPR vendor id.

adapters/algorix/algorix.go
VeronikaSolovei9: This is how this function is called:
```
mediaType := getBidType(seatBid.Bid[idx].ImpID, internalRequest.Imp)
```
where first argument is bid from response and second argument is impressions from request. 
this line `if imp.ID == impId {` determines type if bid based on imp type. 

Please advice if you think it's not ok.

Bugxyb: @SyntaxNode @VeronikaSolovei9 Hello, this func is determines BidType in the bidResponse.Bids. In my understanding, this func is similar in many adapters(base OpenRtb 2.5 protocol). can you tell me which adapter func getMediaTypeForImp(getBidType) is standard, thx. we check Inmobi, smartyads, yeahmobi, BrightMountainMedia and others adapter is the same as func getMediaTypeForImp(getBidType).

I see the new Adapters E-Volution, has the same problem? So the newest solution is add bid type in bid.ext?

So I support add the sample func getBidType in Doc, in this doc: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html, we can't find getMediaTypeForBid() how to implement.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	err = json.Unmarshal(extBidder.Bidder, &extImpAlgoriX)
+	if err != nil {
+		return nil, err
+	}
+	return &extImpAlgoriX, nil
+}
+
+func (a *adapter) getEndPoint(ext *openrtb_ext.ExtImpAlgorix) (string, error) {
+	endPointParams := macros.EndpointTemplateParams{SourceId: ext.Sid, AccountID: ext.Token}
```
guscarreon, line 93: Do you think it'd be safer to call `url.PathEscape()` on `ext.Sid` and `ext.Token` just to be sure they are correctly escaped and can safely be placed inside a URL path?
```
92   func (a *adapter) getEndPoint(ext *openrtb_ext.ExtImpAlgorix) (string, error) {
93 -     endPointParams := macros.EndpointTemplateParams{SourceId: ext.Sid, AccountID: ext.Token}
   +     endPointParams := macros.EndpointTemplateParams{
   +         SourceId: url.PathEscape(ext.Sid),
   +         AccountID: url.PathEscape(ext.Token),
   +     }
94       return macros.ResolveMacros(a.EndpointTemplate, endPointParams)
95   }
adapters/algorix/algorix.go
```
Unless you are perfectly confident on this parameter's contents and you deem it unnecessary, of course.

adapters/algorix/algorixtest/supplemental/check_first_imp.json
```diff
@@ -0,0 +1,123 @@
...
+      },
+      {
+        "id": "test-imp-id_2",
+        "banner": {
+          "format": [{"w": 300, "h": 250}]
+        },
+        "ext": {
+          "bidder": "aaa"
+        }
+      }
```
guscarreon, line 25: Prebid Server code would filter out the second imp in this test case but it doesn't hurt to include this test case.

adapters/algorix/algorix.go
```diff
@@ -0,0 +1,163 @@
...
+			}
+			if imp.Native != nil {
+				return openrtb_ext.BidTypeNative
+			}
+			if imp.Video != nil {
+				return openrtb_ext.BidTypeVideo
+			}
+		}
+	}
+	return mediaType
```
guscarreon, line 162: We could remove `mediaType`. Unless you believe it makes for a more straightforward read, I'm good either way.
```
147   func getBidType(impId string, imps []openrtb2.Imp) openrtb_ext.BidType {
148 -     mediaType := openrtb_ext.BidTypeBanner
149       for _, imp := range imps {
150           if imp.ID == impId {
151               if imp.Banner != nil {
152 -                 return openrtb_ext.BidTypeBanner
    +                 break
153               }
154               if imp.Native != nil {
155                   return openrtb_ext.BidTypeNative
156               }
157               if imp.Video != nil {
158                   return openrtb_ext.BidTypeVideo
159               }
160           }
161       }
162 -     return mediaType
    +     return openrtb_ext.BidTypeBanner
163   }
adapters/algorix/algorix.go
```

guscarreon (commented):
Looks pretty solid. A couple more questions before approving.

adapters/algorix/algorixtest/supplemental/check_first_imp.json
Bugxyb: we delete this test check file.

SyntaxNode:
@Bugxyb The build failed due to formatting. Please run gofmt on all files in the PR.

Bugxyb:
> @Bugxyb The build failed due to formatting. Please run gofmt on all files in the PR.

run gofmt done.

guscarreon (approved):
LGTM. Thank you @Bugxyb for addressing the feedback

VeronikaSolovei9 (approved):

PR #1867, opened 2021-05-26
Facebook: Drop consented providers

VeronikaSolovei9:


VeronikaSolovei9:
We had a discussion about moving a function that drops element from json array to common utils. What would be the best place for it? Now it's a function in facebook.go file.

adapters/audienceNetwork/audienceNetworktest/exemplary/video_consented_providers.json
```diff
@@ -0,0 +1,147 @@
...
+                    },
+                    "device": {
+                        "ip": "152.193.6.74"
+                    },
+                    "user": {
+                        "id": "db089de9-a62e-4861-a881-0ff15e052516",
+                        "buyeruid": "v4_bidder_token",
+                        "ext": {
+                            "consent": "TESTCONSENT",
+                            "consented_providers_settings": {
```
SyntaxNode, line 90: Should we remove the `consented_providers_settings` parent object if it's empty after removing the `consented_providers`?

adapters/audienceNetwork/facebook.go
```diff
@@ -39,6 +41,10 @@ type facebookReqExt struct {
 	AuthID     string `json:"authentication_id"`
 }
 
+const (
+	consentedProviders = "consented_providers"
```
SyntaxNode, line 45: IMHO there's not much value in this constant declaration. Directly using the string in code would be fine, depending on how you feel about "magic strings".

adapters/audienceNetwork/facebook.go
```diff
@@ -115,6 +128,50 @@ func (this *FacebookAdapter) makeAuthID(req *openrtb2.BidRequest) string {
 	return hex.EncodeToString(h.Sum(nil))
 }
 
+func dropSingleElement(extension []byte, elementName string, i interface{}) ([]byte, error) {
```
SyntaxNode, line 131: Nice. I wouldn't have thought to use the decoder for this scenario.

SyntaxNode, line 131: Is it necessary to use `i` in this code? If you're removing an element by it's name, why does the type matter?

adapters/audienceNetwork/facebook.go
```diff
@@ -115,6 +128,50 @@ func (this *FacebookAdapter) makeAuthID(req *openrtb2.BidRequest) string {
...
+				if err != nil {
+					return nil, err
+				}
+			}
+			endIndex := dec.InputOffset()
+
+			if dec.More() {
+				for {
+					//structure has more elements, need to find index of comma
+					if extension[endIndex] == []byte(",")[0] {
```
SyntaxNode, line 157: Consider hardcoding the character code for a comma rather than computing it each loop iteration.

SyntaxNode:
> We had a discussion about moving a function that drops element from json array to common utils. What would be the best place for it?

I suggest placing the helper method in a new `util/jsonutil` package.

adapters/audienceNetwork/facebook.go
VeronikaSolovei9: OMG it works without `i`! Code is so much easier now! 
P.S. I swear I tried it already, but it didn't work :\

adapters/audienceNetwork/facebook.go
VeronikaSolovei9: Done!

adapters/audienceNetwork/facebook.go
VeronikaSolovei9: Modified to "magic string"

adapters/audienceNetwork/audienceNetworktest/exemplary/video_consented_providers.json
VeronikaSolovei9: Code now supports dropping any kind of element. I modified it to "consented_providers_settings". Checking if parent element has no other sub-elements is not trivial thing to do. Should I add this functionality? If yes: is this behavior obvious?  Should it be additional flag "removeParentElementIfEmpty"?

SyntaxNode: I can't find any other elements in `consented_providers_settings`. I think we're fine to always remove it with no need to add the "is empty" logic.

util/jsonutil/jsonutil_test.go
```diff
@@ -0,0 +1,154 @@
+package jsonutil
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestDropSingleElementAfterAnotherElement(t *testing.T) {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Agree, added some test cases to cover error conditions and refactored existing tests accordingly

util/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+)
+
+var comma = []byte(",")[0]
+
+func DropElement(extension []byte, elementName string) ([]byte, error) {
```
hhhjort, line 11: Nice use of the JSON library to do the minimal decoding needed to snip out the JSON object.

util/jsonutil/jsonutil.go
```diff
@@ -0,0 +1,57 @@
...
+	buf := bytes.NewBuffer(extension)
+	dec := json.NewDecoder(buf)
+	var startIndex int64
+	var i interface{}
+	for {
+		token, err := dec.Token()
+		if err == io.EOF {
```
hhhjort, line 18: Do we want to return any signal that the element was not found?

hhhjort (approved):
One question to consider, but it otherwise looks good to me.

util/jsonutil/jsonutil.go
SyntaxNode: We don't have a need for a signal currently. We can add in the feature now and maybe use it in the future, or I also support deferring this feature until/if we need it.

adapters/audienceNetwork/audienceNetworktest/exemplary/video_consented_providers.json
```diff
@@ -0,0 +1,147 @@
...
+                    },
+                    "device": {
+                        "ip": "152.193.6.74"
+                    },
+                    "user": {
+                        "id": "db089de9-a62e-4861-a881-0ff15e052516",
+                        "buyeruid": "v4_bidder_token",
+                        "ext": {
+                            "consent": "TESTCONSENT",
+                            "test": 123
```
SyntaxNode, line 91: `"test": 123` is not a valid part of the Facebook request. I recommend excluding it from this test and relying on the unit tests to ensure other json elements aren't inappropriately affected.

util/jsonutil/jsonutil_test.go
```diff
@@ -0,0 +1,122 @@
...
+			output:          []byte(``),
+			errorExpected:   true,
+			errorContains:   "invalid character",
+		},
+	}
+
+	for _, tt := range tests {
+		res, err := DropElement(tt.input, tt.elementToRemove)
+
+		if !tt.errorExpected {
```
SyntaxNode, line 113: The `!` can be easy to miss. I recommend swapping the cases to avoid it.

adapters/audienceNetwork/audienceNetworktest/exemplary/video_consented_providers.json
VeronikaSolovei9: Removed

util/jsonutil/jsonutil_test.go
VeronikaSolovei9: Agree, modified

SyntaxNode (approved):

SyntaxNode:
@hhhjort  is not available to re-review today. The changes made since his previous approval are minor in nature, only affecting tests - I'm going to still consider his approval valid and merge this in.

PR #1869, opened 2021-05-27
Criteo: accept zoneId and networkId alternate case

leonardlabat:
Our PBS adapter has been released initially accepting both **zoneid** and **networkid** (with that specific case).
However, our public Prebid documentation states that we are accepting **zoneId** and **networkId**, causing issues for publishers using both adapters.

So, we want our PBS adapter to accept both cases.

### DIFF
--- a/static/bidder-params/criteo.json
+++ b/static/bidder-params/criteo.json
@@ -5,13 +5,23 @@
     "type": "object",
     "properties": {
         "zoneid": {
-            "type": "number",
+            "type": "integer",
             "description": "Impression's zone ID.",
             "minimum": 0
         },
+        "zoneId": {
+            "type": "integer",
+            "description": "Impression's zone ID, preferred.",
+            "minimum": 0
+        },
         "networkid": {
-            "type": "number",
+            "type": "integer",
             "description": "Impression's network ID.",
+            "minimum": 0
+        },
+        "networkId": {
+            "type": "integer",
+            "description": "Impression's network ID, preferred.",
             "minimum": 0
         }
     },
@@ -23,7 +33,17 @@
         },
         {
             "required": [
+                "zoneId"
+            ]
+        },
+        {
+            "required": [
                 "networkid"
+            ]
+        },
+        {
+            "required": [
+                "networkId"
             ]
         }
     ]
--- a/adapters/criteo/criteotest/supplemental/multislots-alt-case.json
+++ b/adapters/criteo/criteotest/supplemental/multislots-alt-case.json
@@ -1,0 +1,232 @@
+{
+    "mockBidRequest": {
+      "id": "test-request-id",
+      "site": {
+        "id": "site-id",
+        "page": "criteo.com"
+      },
+      "device": {
+        "os": "android",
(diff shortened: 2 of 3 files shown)


hhhjort (dismissed):
Looks good to me.

AlexBVolcy (approved):

SyntaxNode:
This looks god technically. I have a concern over publisher confusion before we merge this in. There are some hosts who build a user interface on top of this information and publishers might be confused on which field to choose especially if they don't notice the capitalization difference. Do you prefer one style over another?

leonardlabat:
Well the one zoneId / networkId is cleary the prefered one (that's the one authored in the Prebid website documentation).

SyntaxNode:
> Well the one zoneId / networkId is cleary the prefered one (that's the one authored in the Prebid website documentation).

Gotcha. Thank you for clarifying. I don't believe there's a way in json schema to specify which one is preferred. To let publishers know and to give a hint to automated systems, are you willing to include the word "preferred" in the description of the lower case ones? Something like:

"Impression's zone ID., preferred",
"Impression's zone ID (preferred).",

leonardlabat:
I've pushed an update @SyntaxNode thanks for the suggestion !

Do you think it'll be possible to have this included in the next release of prebid server please ? we have a couple of publishers waiting for that change

Thanks

hhhjort (approved):

static/bidder-params/criteo.json
```diff
@@ -1,30 +1,50 @@
...
+            "description": "Impression's zone ID, preferred.",
+            "minimum": 0
+        },
+        "networkid": {
+            "type": "number",
+            "description": "Impression's network ID.",
+            "minimum": 0
+        },
+        "networkId": {
+            "type": "number",
```
SyntaxNode, line 23: Last change request. Could you please change the types from "number" to "integer"? Number includes floating point values, but the underlying data structure is an int64 - so I think integer would be most appropriate. This is confusing some automated config systems.

leonardlabat: Indeed that makes sense
I'm pushing the change

SyntaxNode (approved):

static/bidder-params/criteo.json
SyntaxNode: Excellent. Thank you. We'll merge this in for today's release.

AlexBVolcy (approved):

static/bidder-params/criteo.json
```diff
@@ -1,30 +1,50 @@
...
+    "title": "Criteo adapter params",
+    "description": "The schema to validate Criteo specific params accepted by Criteo adapter",
+    "type": "object",
+    "properties": {
+        "zoneid": {
+            "type": "integer",
+            "description": "Impression's zone ID.",
+            "minimum": 0
+        },
+        "zoneId": {
```
agilfix, line 12: There should only be one zoneId and one networkId parameter, correct - rather than 2 of each?

hhhjort (approved):

PR #1877, opened 2021-06-03
New Adapter: Smile Wanted

MaxSmileWanted:
Please see the "Testing" part to have a Bid Response : https://docs.prebid.org/dev-docs/bidders/smilewanted


MaxSmileWanted:
I create a PR on the prebid.github.io repository to indicate the PBS support : https://github.com/prebid/prebid.github.io/pull/3008

adapters/smilewanted/smilewanted.go
```diff
@@ -0,0 +1,162 @@
...
+			} else {
+				return nil, []error{&errortypes.BadInput{
+					Message: "No valid banner format in the bid request",
+				}}
+			}
+		} else if request.Imp[i].Video != nil {
+			validImpExists = true
+		}
+	}
+	if !validImpExists {
```
SyntaxNode, line 70: The `validImpExists` check is not necessary. I see this is largely being used to ensure the impression is a banner and/or video, but that constraint is already present in `smilewanted.yaml` Prebid Server's code logic will check there are valid imps before calling your adapter. The banner format check still makes to keep.

adapters/smilewanted/smilewanted.go
```diff
@@ -0,0 +1,162 @@
...
+		bid := sb.Bid[i]
+		bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
+			Bid:     &bid,
+			BidType: getMediaTypeForImp(bid.ImpID, internalRequest.Imp),
+		})
+	}
+	return bidResponse, nil
+}
+
+func addHeaderIfNonEmpty(headers http.Header, headerName string, headerValue string) {
```
SyntaxNode, line 136: This function is not used. Please remove it.

adapters/smilewanted/smilewanted.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, imp := range imps {
+		if imp.ID == impId {
+			if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			return mediaType
+		}
+	}
+	return mediaType
+}
```
SyntaxNode, line 154: This is an anti-pattern. Does your bidding server return the media type of the bid in a bid response extension?

adapters/smilewanted/smilewanted_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderSmileWanted, config.Adapter{
+		Endpoint: "http://prebid-server.smilewanted.com"})
```
SyntaxNode, line 13: Nitpick: Please consider using an obviously fake domain for the tests. There is no need to use a real address here.

config/config.go
```diff
@@ -639,6 +639,7 @@ func (cfg *Configuration) setDerivedDefaults() {
...
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderSmartRTB, "https://market-global.smrtb.com/sync/all?nid=smartrtb&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&rr="+url.QueryEscape(externalURL)+"%252Fsetuid%253Fbidder%253Dsmartrtb%2526gdpr%253D{{.GDPR}}%2526gdpr_consent%253D{{.GDPRConsent}}%2526uid%253D%257BXID%257D")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderSmartyAds, "https://as.ck-ie.com/prebid.gif?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redir="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dsmartyads%26uid%3D%5BUID%5D")
+	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderSmileWanted, "https://csync.smilewanted.com/getuid?source=prebid-server&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redirect="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dsmilewanted%26uid%3D%24UID")
```
SyntaxNode, line 642: The redirect url should also include the gdpr and gdpr_consent values for Prebid Server to properly enforce privacy regulations.

static/bidder-info/smilewanted.yaml
```diff
@@ -0,0 +1,12 @@
+maintainer:
+  email: "maxime@smilewanted.com"
```
SyntaxNode, line 2: This looks like a personal email address. We require the use of a shared email group to be sure Prebid.org has a communication path if that individual leaves the company.

SyntaxNode (commented):
Thank for your contribution. It looks like you modeled your adapter after some of our existing ones. We're always trying to improve, so please don't be put off by the comments. The changes requested will make your adapter and easier to maintain.

MaxSmileWanted:
Ok thank you for your attention to detail, I have applied the requested changes.

adapters/smilewanted/smilewanted.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, []error{&errortypes.BadInput{
+			Message: "ext.bidder.zoneId not provided",
+		}}
+	}
+
+	if smilewantedExt.ZoneID == "" {
+		return nil, []error{&errortypes.BadInput{
+			Message: "ZoneId is empty",
+		}}
+	}
```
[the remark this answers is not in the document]
SyntaxNode: I see you added the min length constraint to the json schema. That now allows you to safely remove the following code from your adapter, since it is just checking validity of the bidder extension which is already performed by PBS-Core.

```
if err := json.Unmarshal(request.Imp[0].Ext, &bidderExt); err != nil {
        Message: "ext.bidder not provided",
    }}
}

var smilewantedExt openrtb_ext.ExtImpSmileWanted
if err := json.Unmarshal(bidderExt.Bidder, &smilewantedExt); err != nil {
        Message: "ext.bidder.zoneId not provided",
    }}
}
```

adapters/smilewanted/smilewanted.go
SyntaxNode: I didn't see changes for this comment. The main problem is ensuring the bid type is correct. If your bidding server sees an impression with a banner and a video object, will it always just make a video bid?

adapters/smilewanted/smilewanted.go
```diff
@@ -0,0 +1,127 @@
...
+
+	request.AT = 1 //Defaulting to first price auction for all prebid requests
+
+	reqJSON, err := json.Marshal(request)
+	if err != nil {
+		return nil, []error{&errortypes.BadInput{
+			Message: fmt.Sprintf("Json not encoded. err: %s", err),
+		}}
+	}
```
SyntaxNode, line 54: Please add tests for the errors above this comment in the `MakeRequests` method.

SyntaxNode (commented):
That was a great update. There are two comments I didn't see addressed or a response. My only other comment is an ask to increase test coverage of MakeBids errors.

adapters/smilewanted/smilewanted.go
MaxSmileWanted: We always attach a Type on the Zone that we create on our UI and the publisher implement this zone in the ZoneId params, So when we receive a new ad request, we know that it is a request for Display or Video in the zone configuration on our end.

SyntaxNode: Interesting. Would your server return an error if it receives an Imp with both banner and video for a zone that is configured for video?

MaxSmileWanted:
I Improve the test coverage to 93,5%

`ok      github.com/prebid/prebid-server/adapters/smilewanted    0.155s  coverage: 93.5% of statements`

adapters/smilewanted/smilewanted.go
MaxSmileWanted: No but we will force this on publisher side to have only one zone per type of request.

We have already done that with PrebidJS and it's works fine ;)

SyntaxNode: Gotcha. Do you think it would be worthwhile to add a check that impressions for your adapter only include banner or video, but not both?

SyntaxNode:
> I Improve the test coverage to 93,5%
> 
> `ok github.com/prebid/prebid-server/adapters/smilewanted 0.155s coverage: 93.5% of statements`

Thank you for the quick updates. Code coverage looks superb. I left a comment in the thread about the media type. Please respond to my latest question and I'll seek a second review tomorrow.

adapters/smilewanted/smilewanted.go
MaxSmileWanted: Ok so i looked at when I make a config with two types (Banner and Video) and we answer on our side with one ad (the most expensive one between Video and Display on our side), so then PrebidJS displays it ;)

It will be an optimization that we can do later to take into consideration several requests of different types and make several responses, but it requires quite a few changes on our side ;)

jeanstemp:
docs PR https://github.com/prebid/prebid.github.io/pull/3008

adapters/smilewanted/smilewanted.go
hhhjort: Since it is not a guaranteed correct identification, one possibility, until you have the infrastructure in place to properly in place to identify bid type, is to set your adapter disabled by default. That will give you a chance to open a dialogue with the PBS host and ensure that the client side has the ability to properly detect and handle the difference between banner and video bids.

MaxSmileWanted: Just to clarify : We identify on our end the type, we see in each Bid Request BUT we will answer with only one ad (corresponding to the winning ad on our internal competition between all types).

We support multi request but we do not support multi response, but it's only an optimization for us to have more opportunities to win the competition on publisher's side.

openrtb_ext/imp_smilewanted.go
```diff
@@ -0,0 +1,6 @@
+package openrtb_ext
+
+// ExtImpSmileWanted defines the contract for bidrequest.imp[i].ext.smilewanted
```
mansinahar, line 3: Since your adapter code doesn't really. unmarshal the imp ext and reads/modifies the `ZoneID`, you don't need this file and it can be safely removed

adapters/smilewanted/smilewanted.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+
+	for i := 0; i < len(request.Imp); i++ {
+		if request.Imp[i].Banner != nil {
+			bannerCopy := *request.Imp[i].Banner
+			if len(bannerCopy.Format) > 0 {
+				firstFormat := bannerCopy.Format[0]
+				bannerCopy.W = &(firstFormat.W)
```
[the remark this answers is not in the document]
MaxSmileWanted: I deleted this part, I don't need it

openrtb_ext/imp_smilewanted.go
MaxSmileWanted: Done

adapters/smilewanted/smilewanted.go
mansinahar: I think overall the PR LGTM now. Just want to warn you that the above approach of assigning the media type for the bid might lead to the final ad not being rendered properly if the media type wasn't correctly set for a multi-format request. This might happen if for an imp that has both banner and a video object, your server returns a banner bid but then here you assign the media type to be video.

hhhjort: We have had issues with other bidders in the past. If the final rendering client is not smart enough, it can get confused as to what it is trying to display. Just considering how you can get ahead of it rather than having your publisher clients have a poor experience that needs to be debugged after the fact.

mansinahar (approved):

hhhjort (approved):
LGTM

Issue #1554, opened 2020-10-23
Support both image and iframe usersyncs 


bretg:
Currently Prebid Server adapters are limited to one type of usersync - they have to choose either a image-based sync or an iframe-based sync. We've learned that iframe usersyncs don't work well in the AMP environment, but we want bidders to have the option of iframe-based syncs.

Proposed enhancement:

1) enhance the /cookie_sync endpoint to accept iframe/pixel preferences

This should align with the Prebid.js way of defining [userSync](https://docs.prebid.org/dev-docs/publisher-api-reference.html#setConfig-Configure-User-Syncing) `filterSettings`:

```
        filterSettings: {
            iframe: {
                bidders: ['def'],  // only this bidder is excluded from syncing iframe pixels, all other bidders are allowed
                filter: 'exclude'
            },
            image: {
                bidders: ['abc', 'def', 'xyz'],  //only these 3 bidders are allowed to sync image pixels
                filter: 'include'
            }
        },
```

2) The AMP load-cookie invocation of /cookie_sync would exclude iframes for all bidders
```
        filterSettings: {
            iframe: {
                bidders: '*',
                filter: 'exclude'
            }
        },
```

3) Bid adapters should be able specify both pixel and iframe usersyncs, and the preferred method. They may also specify no usersync at all. 

4) When it receives a request, the /cookie_sync endpoint should determine which usersync to respond with:

- Determine which bidders need to be synced:
    - Prioritize the bidders listed in the incoming `bidders` param. Remove them from the list if already in the `uids` cookie or if they don't have a usersync method allowed by current filterSettings
    - Randomly add other supported bidders that have a usersync allowed by current filterSettings but aren't in the `uids` cookie up to the `limit` (cooperative syncing)

- Loop through bidders to sync
    - if cookie_sync.filterSettings allows this bidder to use their preferred method, add it to the response JSON
    - else if cookie_sync.filterSettings allows this bidder to use the other method, add it to the response JSON

5) /cookie_sync needs to tell /setuid which kind of response to make.
    - For iframe syncs, it should add f=b into the /setuid URL. 
    - For image syncs, it should add f=i into the /setuid URL.
    - how exactly this is done is up to the engineering teams. 

6) See https://github.com/prebid/prebid-server/issues/1477 -- we also need to change the /setuid endpoint to look for f=i and f=b parameters to return with the appropriate response.

bretg:
Approved in Prebid Server committee

patmmccann:
[...]

Issue #1477, opened 2020-09-04
Safari trying to download setuid result


Hamper:
When setuid endpoint requested from prebid.js in safari browser it shows `Do you want to allow downloads on "site"` dialog. If we try directly open setuid in safari like `https://prebid-server/setuid?bidder=lunamedia&gdpr=0&gdpr_consent=&uid=$UID` we get `Do you want to download "setuid"` dialog. I think safari interprets `200` responses without `Content-Type` header as downloadable files.

bretg:
I'm not seeing this behavior with either prebid.adnxs.com or prebid-server.rubiconproject.com.

https://prebid-server.rubiconproject.com/setuid?bidder=33across&gdpr=&gdpr_consent=&us_privacy=&uid=210745218046930

Anyhow, it's possible that we need to tune something here, but need someone to define the solution. PBS currently responds with empty content regardless of whether that bidder's usersync type is redirect or iframe.

1) should PBS set content-type: text/html for iframe usersyncers? Is it valid to respond with content-length:0?
2) should PBS set content-type: image/png for redirect usersyncers? Is it valid to respond with content-length:0 instead of a blank pixel?
3) should we add a 'format' parameter to /setuid to let the caller define the desired response? (we did that with the /event endpoint)

bretg:
Discussed in PBS committee. We agreed to modify the PBS /setuid endpoint:

1) If we do not have permission to set a cookie, the response should be HTTP 451 with a blank body.
2) else if the named bidder has a usersync type of "redirect", PBS will respond with a blank 1x1 PNG, set the Content-Length to the appropriate number of bytes, and set Content-Type to image/png
3) else if the named bidder has a usersync type "iframe", PBS will response with empty HTML with Content-Length 0 and Content-Type to text/html

bretg:
Extended the algorithm with f=i and f=b in light of the changes being made in https://github.com/prebid/prebid-server/issues/1554

bretg:
Implemented in PBS-Java 1.62

However, I think there's a gap now with the support for bidders having both redirect and iframe syncs -- we need PBS to update the /setuid URL to add f=i when it's a redirect or f=b when it's an iframe.  Opened https://github.com/prebid/prebid-server/issues/1831 to track.

SyntaxNode:
Implemented in PBS-Go 0.173.0.

Issue #924, opened 2019-05-31
Proposal: Cooperative Cookie Syncing


bretg:
We've discussed in the Prebid Server committee meetings a feature that allows a Prebid Server host company to pre-sync cookies with more bidders than are in the page. 

1) Prebid Server host company can configure lists of bidder adapters that can be included as part of a cookie-sync response even if that bidder is not specifically requested by the page. 

The lists of bidders should be prioritizable in groups, representing the fact that not all bidders are of equal importance to the publishers of a given host company.

```
cookie-sync:
   coop-sync:
       default: true
       pri[0]: bidderA, bidderB
       pri[1]: bidderC, bidderD
```

Which bidders are chosen within a given group should be random.
If no priority groups are specified, the default is to have all bidders at the same level.

2) The publisher should be able to limit how many sync pixels can happen in a given cookie-sync, and should be able to turn off cooperative syncing.

```
{"bidders": ["appnexus", "rubicon"], coopSync: true, limit: 5}
```

For example, if appnexus and rubicon are on the page and biddersA-D are in the coop-sync config as shown above, PBS would respond with up to 5 sync pixels. e.g. if the user had no IDs in the `uids` cookie, then the response to /cookie-sync would include pixels for appnexus, rubicon, bidderA, bidderB, and (randomly) either bidderC or D.

3) Whether coop syncing is on by default should be configurable by the Prebid Server host company

bretg:
Note: working on this in PBS-Java. 
We tweaked the format of the config above to have it better conform to YAML

bretg:
This is done in PBS-Java. Assigning to @hhhjort for PBS-Go implementation.

SyntaxNode:
@bretg I'm not familiar with the `pri[0]` style of syntax for a yaml array. I expect this will more like this:
```
cookie-sync:
    coop-sync:
        default: true
        pri:
            - [bidderA, bidderB]
            - [bidderC, bidderD]
```

Is this how it was implemented in PBS-Java?

SyntaxNode:
Implemented in PBS-Go 0.173.0.

PR #1879, opened 2021-06-05
/cookie_sync Endpoint Rewrite

SyntaxNode:
I've rewritten the majority of the /cookie_sync endpoint and user syncer architecture to implement https://github.com/prebid/prebid-server/issues/924, https://github.com/prebid/prebid-server/issues/1477, https://github.com/prebid/prebid-server/issues/1554, and to further progress https://github.com/prebid/prebid-server/issues/1429 as the config system now includes user syncing settings.

## Todo
- [x] Merge from master
- [x] Add config tests
- [x] Standardize metrics and add unit tests.
- [x] Convert all existing user sync configs to the new format.

## Configuration Changes
_Technical details will be included in a separate PR to https://github.com/prebid/prebid.github.io before this is merged._
- User sync configuration is relocated to the bidder yaml files in `static/bidder-info` from being hardcoded in the `config.go` and adapter `usersync.go` files. Hosts can now override all user syncing options via either `pbs.yaml` or environment variables.
- User sync configuration now supports new macros, automatic url encoding, and a new configuration hierarchy. The external url may now be overridden at the host level, the bidder level, and the endpoint level. The default redirect url is now represented with macros such that no modification is needed for most cases. Look for more details in the docs PR.
- Existing `usersync_url` overrides are temporarily supported for backwards compatibility, but please update your configs as soon as possible after release.
- Bidders may share the same user sync key, this is ideal for hardcoded aliases or bidders with multiple adapter implementations. As a side effect, I'm delighted to announce this removes the special `adnxs` hardcoded handling for AppNexus.
- Bidders may now specify both iframe and/or redirect endpoints.
- Cooperative cookie syncing is implemented and disabled by default.

Example Current Config: Buried In `config/config.go`
```
setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderAppnexus, "https://ib.adnxs.com/getuid?"+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dadnxs%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
```

Example New Config: In `static/bidder-info/appnexus.yaml`
```
userSync:
  key: "adnxs"
  redirect:
    url: "https://ib.adnxs.com/getuid?{{.RedirectURL}}"
    userMacro: "$UID"
```

## API Changes
- The `/cookie_sync` endpoint now defaults to all bidders if the bidders field in the request is missing, null, or empty. This matches behavior in PBS-Java. We discussed and couldn't find a use case for handling each of these conditions separately. Currently, a missing bidders parameter results in an error response, a null parameter results in using all bidders, and an empty parameter results in no user syncing.
- The `/cookie_sync endpoint` will now accept filters in the request defining which user sync types (iframe, redirect) are permitted. The format is fully compatible with Prebid.js and PBS-Java.
- Cooperative cookie syncing is now supported. Requests may specify if they want this feature enabled or disabled.
- The `SSCookie` sidecar cookie is no longer required and is removed with this PR.
- PBS will no longer invalidate user syncs when it detects the browser has upgraded to a new version with support for same site cookies. This behavior isn't present in PBS-Java and our usage data shows the vast majority of end users are already using same site cookie compatible browsers.
- The `/setuid` endpoint will now properly respond with an empty image for redirects and an empty html response for iframes.

### Influx
Metrics have be reorganized around endpoint names and syncer keys. `cookie_sync_requests.*` are directly related to the `/cookie_sync` endpoint, `set_uid_requests.*` are directly related to the `/setuid` endpoint, and `syncer.*` are related to each syncer. Specifically,
 - `usersync.opt_outs` and `usersync.bad_requests` are removed and replaced with `cookie_sync_requests.[status]` where status is `ok`, `bad_request`, `opt_out`, and `gdpr_blocked_host_cookie`.
- `usersync.unknown.sets` and `usersync.unknown.gdpr_prevent` are removed and replaced with `set_uid_requests.key_unknown`.
- `cookie_syncs[bidder].gen` and `cookie_sync.[bidder].gdpr_prevent` are removed and replaced with `syncer.[key].request.[status]` where the syncer key may be shared among multiple bidders and status is `ok`, `privacy_blocked`, `already_synced`, `type_not_supported`.
-  `usersync.[bidder].sets` and `usersync.[bidder].gdpr_prevent` are removed and replaced with `syncer.[key].set.[status]` where the syncer key may be shared among multiple bidders and status is `ok` and `privacy_blocked`.
- `setuid_requests.[status]` is new where status is `ok`,` opt_out`, `gdpr_blocked_host_cookie`, `key_unknown`, and `type_not_supported`.

### Prometheus
Metrics have be reorganized around endpoint names and syncer keys.` cookie_sync_requests` is directly related to the `/cookie_sync endpoint`, `set_uid_requests` is directly related to the `/setuid` endpoint, `syncer_requests` and `syncer_sets` are related to each syncer. Specifically,
- `cookie_sync_request` now has a status label with values `ok`, `bad_request`, `opt_out`, and `gdpr_blocked_host_cookie`. Existing PromQL queries will still accurately report the total number of requests to the endpoint.
- `setuid_requests` is new and has a status label with values `ok`, `opt_out`, `gdpr_blocked_host_cookie`, `key_unknown`, and `type_not_supported`.
- `adapter_cookie_sync` has been removed and replaced with `syncer_requests` which has a syncer key label and status label with the values `ok`, `privacy_blocked`, `already_synced`, `type_not_supported`. These syncer metrics are only counted when the syncer has been selected as a user sync candidate.
- `adapter_user_sync` has been removed and replaced with `syncer_sets` which has a syncer key label and status label with the values `ok` and `privacy_blocked`.

## Future Plans
I'm working on account settings for the /cookie_sync endpoint, but this PR is already huge so that will be separate later on.


adapters/acuityads/acuityads.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 type AcuityAdsAdapter struct {
-	endpoint template.Template
+	endpoint *template.Template
```
hhhjort, line 18: Why did this become a pointer?

SyntaxNode: The method to render a template is a pointer receiver and much of the internal data structures to a Template are also pointers. We were converting the value to a pointer right before rendering, so this is a minor improvement to avoid the unnecessary copies and simplifies the template rendering code. I don't expect this to be a problem since the Go docs specifically state: 

> a template may be executed safely in parallel

config/adapter.go
```diff
@@ -9,25 +9,13 @@ import (
 )
 
 type Adapter struct {
-	Endpoint string `mapstructure:"endpoint"` // Required
-	// UserSyncURL is the URL returned by /cookie_sync for this Bidder. It is _usually_ optional.
```
hhhjort, line 13: Do we want to save the template documentation?

config/bidderinfo_test.go
```diff
@@ -184,6 +160,161 @@ func TestLoadBidderInfo(t *testing.T) {
 	}
 }
 
+func TestSyncerOverride(t *testing.T) {
```
hhhjort, line 163: It doesn't look like you have any tests where you confirm that some fields are untouched while others are overridden in one request. (Where the override syncer only provides a subset of original fields.)

config/adapter.go
SyntaxNode: This has been moved to `config/bidderinfo.go` for the `SyncerEndpoint` type.

router/router.go
```diff
@@ -315,6 +330,42 @@ func New(cfg *config.Configuration, rateConvertor *currency.RateConverter) (r *R
...
+				if endpointsCount == 0 {
+					return fmt.Errorf("failed to apply legacy usersync_url setting for bidder %s, bidder does not define user sync endpoints", bidderName)
+				}
+
+				if endpointsCount > 1 {
+					return fmt.Errorf("failed to apply legacy usersync_url setting for bidder %s, bidder defines multiple user sync endpoints", bidderName)
+				}
+			}
+
+			bidderInfos[bidderName] = bidderInfo
```
hhhjort, line 363: I had to stare at this for a few minutes. `bidderInfo` is a loop variable, so gets reused. However we are copying the `bidderInfo` struct into a map of them, so we copy normally. No issues here.

router/router.go
```diff
@@ -315,6 +330,42 @@ func New(cfg *config.Configuration, rateConvertor *currency.RateConverter) (r *R
...
+				if bidderInfo.Syncer.Redirect != nil {
+					bidderInfo.Syncer.Redirect.URL = adapterCfg.UserSyncURL
+					endpointsCount++
+				}
+
+				if endpointsCount == 0 {
+					return fmt.Errorf("failed to apply legacy usersync_url setting for bidder %s, bidder does not define user sync endpoints", bidderName)
+				}
+
+				if endpointsCount > 1 {
```
hhhjort, line 358: From the description: `Bidders may now specify both iframe and/or redirect endpoints.`

Given this, why is it an error for a bidder to specify both?

router/router.go
SyntaxNode: Sorry for the confusion. Do you recommend a different way of writing it to be less confusing? .. or perhaps a comment might be helpful here?

router/router.go
SyntaxNode: This is not a general validation error. This logic only kicks in if an older style `PBS_ADAPTERS_FOO_USERSYNC_URL` override value is used. It is now possible for the bidder to have two endpoints, so which one is the override intended for? Rather than trying to guess, it's an error to provide an older-style user sync url override if the bidder defines more than one endpoint. The replacement format would be to include the type in the override key, such as `PBS_ADAPTERS_FOO_USERSYNC_IFRAME_URL`.

SyntaxNode: Alternatively, I can omit this attempt at a compatibility layer and make it clear in the release notes this format is no longer supported.

config/bidderinfo_test.go
SyntaxNode: Good point. I intend to write this test soon.

router/router.go
hhhjort: This makes sense now. You can only use the URL override if there is only one option. I was thinking the other way around, where we would only use the old style config if the new one is missing. Perhaps instead of checking for both new options, you could check which type the old URL was for, and only overwrite that setting in the new config.

One worry with this method is these old overrides sitting forgotten in the config in a legacy install, and someone having issues getting updated URL values to take. But given all the defaults have been removed, this will only occur when a custom usersync URL was set, and then the administrator forgets where they set the custom value.

All in all, I will say this can stay here. It is just migrating config location is always a pain.

router/router.go
hhhjort: No, no changes, just I try to be alert when I see `bidderInfos[bidderName] = bidderInfo` where `bidderInfo` is a loop variable. :) Figured I would make a note here for other reviewers.

usersync/bidderchooser.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if cooperative.Enabled {
+		return c.chooseCooperative(requested, available, cooperative.PriorityGroups)
+	}
+
+	return c.shuffledCopy(requested)
+}
+
+func (c standardBidderChooser) chooseCooperative(requested, available []string, priorityGroups [][]string) []string {
+	biddersCapacity := int(float64(len(available)) * 1.5)
```
hhhjort, line 27: Are you setting a length of 1.5 because it is possible the same bidder appears more than once in the list of chosen bidders?

router/router.go
SyntaxNode: > Perhaps instead of checking for both new options, you could check which type the old URL was for, and only overwrite that setting in the new config.

The information is not represented in code. I would need to add a new yaml setting for something like "legacy sync type", which I'm not sure is worth it to just support this one case.

> One worry with this method is these old overrides sitting forgotten in the config in a legacy install

I share that concern. Perhaps it would be simpler to just throw an error if the old override format is used and ask hosts to update for this breaking config change.

usersync/bidderchooser.go
SyntaxNode: Yes, exactly. I'm not sure what to use for a default array size, so I figure there would be some overlap at least with the `available` items. I randomly chose to use a 50% overhead. This is purely a guess to try to avoid array reallocation within the slice.

SyntaxNode: I'll add a comment in the code to explain the 1.5.

usersync/bidderfilter.go
```diff
@@ -0,0 +1,53 @@
...
+// BidderFilterMode represents the inclusion mode of a BidderFilter.
+type BidderFilterMode int
+
+const (
+	BidderFilterModeInclude BidderFilterMode = iota
+	BidderFilterModeExclude
+)
+
+// BidderFilter determines if a bidder has permission to perform a user sync activity.
+type BidderFilter struct {
```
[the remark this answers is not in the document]
hhhjort: Is it possible to have a `nil` value for `biddersLookup map[string]struct{}`? That could take the place of the `biddersAll` flag if possible, as there is no need for a map if all are valid.

usersync/bidderchooser.go
```diff
@@ -0,0 +1,60 @@
...
+
+// standardBidderChooser implements the bidder choosing algorithm per official Prebid specification.
+type standardBidderChooser struct {
+	shuffler shuffler
+}
+
+func (c standardBidderChooser) choose(requested, available []string, cooperative Cooperative) []string {
+	if len(requested) == 0 {
+		return c.shuffledCopy(available)
+	}
```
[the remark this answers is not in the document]
hhhjort: It seems like it would make sense to prioritize the priority groups if cooperative is enabled, but requested was empty. If we call `chooseCooperative()` before checking for an empty `requested`, we will want to make sure `shuffledAppend()` deals well with an empty append. Which may not be a bad thing to verify in general.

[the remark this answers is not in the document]
hhhjort: The prebid.org docs don't have cooperative mode documented yet, and issue #924 doesn't seem to mention the possible case. My best guess is that `requested` is intended to be the bidders on the page, and the priority groups are prioritizing any other bidders to be synced in their groups. In that world, you would only get an empty `requested` if the page had no prebid ads, which seems like a valid, if rare, use case. And of course there is the case where the publisher doesn't follow the (nonexistent) docs correctly, and only fills in the priority groups, which makes sense as a way to go. (what if you want to ensure some bidders on the page are synced before others.)

router/router.go
```diff
@@ -315,6 +330,42 @@ func New(cfg *config.Configuration, rateConvertor *currency.RateConverter) (r *R
 	return r, nil
 }
 
+func applyBidderInfoConfigOverrides(bidderInfos config.BidderInfos, adaptersCfg map[string]config.Adapter) error {
+	for bidderName, bidderInfo := range bidderInfos {
+		if adapterCfg, exists := adaptersCfg[bidderName]; exists {
+			bidderInfo.Syncer = adapterCfg.Syncer.Override(bidderInfo.Syncer)
+
+			// comptability with legacy settings (if possible unambiguously)
```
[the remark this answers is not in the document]
SyntaxNode: Comment added.

usersync/bidderfilter.go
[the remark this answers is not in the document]
SyntaxNode: Agreed. Separated.

endpoints/setuid.go
```diff
@@ -42,37 +37,44 @@ func NewSetUIDEndpoint(cfg config.HostCookie, syncers map[openrtb_ext.BidderName
...
-			})
+			metricsEngine.RecordSetUid(metrics.SetUidSyncerUnknown)
+			so.Status = http.StatusBadRequest
+			return
+		}
+		so.Bidder = syncerKey
+
+		responseFormat, err := getResponseFormat(query, syncers[syncerKey])
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
```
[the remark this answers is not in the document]
SyntaxNode: It's not possible to have an invalid default sync type since it's validated on server startup and an error is treated as fatal. With this context, do you think it's worth returning an empty string without an error? I need to handle the default case in code, even though it's not reachable in practice.

endpoints/setuid.go
```diff
@@ -81,38 +83,67 @@ func NewSetUIDEndpoint(cfg config.HostCookie, syncers map[openrtb_ext.BidderName
 		so.UID = uid
 
 		if uid == "" {
-			pc.Unsync(familyName)
+			pc.Unsync(syncerKey)
+			metricsEngine.RecordSetUid(metrics.SetUidOK)
+			metricsEngine.RecordSyncerSet(syncerKey, metrics.SyncerSetUidCleared)
+			so.Success = true
 		} else {
```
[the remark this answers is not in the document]
SyntaxNode: Good catch. Fixed.

endpoints/auction.go
```diff
@@ -58,14 +60,14 @@ func writeAuctionError(w http.ResponseWriter, s string, err error) {
...
-	syncers       map[openrtb_ext.BidderName]usersync.Usersyncer
+	syncers       map[string]usersync.Syncer
 	gdprPerms     gdpr.Permissions
 	metricsEngine metrics.MetricsEngine
 	dataCache     cache.Cache
 	exchanges     map[string]adapters.Adapter
 }
 
-func Auction(cfg *config.Configuration, syncers map[openrtb_ext.BidderName]usersync.Usersyncer, gdprPerms gdpr.Permissions, metricsEngine metrics.MetricsEngine, dataCache cache.Cache, exchanges map[string]adapters.Adapter) httprouter.Handle {
+func Auction(cfg *config.Configuration, syncers map[string]usersync.Syncer, gdprPerms gdpr.Permissions, metricsEngine metrics.MetricsEngine, dataCache cache.Cache, exchanges map[string]adapters.Adapter) httprouter.Handle {
```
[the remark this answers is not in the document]
SyntaxNode: BidderName is just a type alias for string without any additional functionality. On one hand I like that it uses the type name to describe the data, on the other hand I dislike all of the casting needed to maintain the alias - so i haven't been using it in new code I write. Do others have a preference here?

config/bidderinfo.go
```diff
@@ -12,37 +12,192 @@ import (
...
 type DebugInfo struct {
 	Allow bool `yaml:"allow"`
 }
 
+// Syncer specifies the user sync settings for a bidder. This struct is shared by the account config,
+// so it needs to have both yaml and mapstructure mappings.
+type Syncer struct {
+	// Key is used as the record key for the user sync cookie. We recommend using the bidder name
+	// as the key for consistency, but that is not enforced as a requirement. Each bidder must
+	// have a unique key.
```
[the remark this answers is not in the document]
SyntaxNode: Removed.

usersync/bidderchooser.go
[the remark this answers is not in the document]
SyntaxNode: I agree with the proposal to consider cooperative priority groups if no bidders are specified. The end result is the same where all bidders are equally considered if priority groups aren't specified. Changed.

> we will want to make sure shuffledAppend() deals well with an empty append
There is already test coverage for this case.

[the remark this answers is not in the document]
bsardo: This change looks good. One more possible case to consider is if cooperative cookie sync is off but the limit specified is greater than the number of bidders requested. Right now we are just considering the bidders that were requested but should we also consider other available enabled bidders to reach the limit?
For example, cooperative sync is off, the request limit is five but only two bidders are requested. Should we also sync three other bidders?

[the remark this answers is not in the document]
SyntaxNode: > Right now we are just considering the bidders that were requested

Correct. This is the desired behavior.

> should we also consider other available enabled bidders to reach the limit?

We should not. That's not how the endpoint is designed to work.

endpoints/setuid.go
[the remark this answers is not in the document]
SyntaxNode: Discussed offline. I'll change `getResponseFormat` to return an empty string if it's not provided or if PBS doesn't know the default.

static/bidder-info/triplelift_native.yaml
```diff
@@ -9,7 +9,6 @@ capabilities:
     mediaTypes:
       - native
 userSync:
-  key: "triplelift"
```
hhhjort: Why was the key being dropped?

SyntaxNode: I'd like to keep the identity pools exactly as they are right now. I'll put this in a separate PR for review from TripleLift at a later date.

bsardo (commented):
This is looking good.

usersync/syncer_test.go
```diff
@@ -300,31 +309,31 @@ func TestBuildTemplate(t *testing.T) {
...
+			expectedRendered: "https://bidder.com/sync?redirect=%7B%7BxExternalURL%7D%7D",
 		},
 		{
 			description: "Invalid Macro - Syncer Key",
 			givenSyncerEndpoint: config.SyncerEndpoint{
 				URL:         "https://bidder.com/sync?redirect={{.RedirectURL}}",
 				RedirectURL: "{{xSyncerKey}}",
 			},
-			expectedError: "template: anykey_usersync_url:1: function \"xSyncerKey\" not defined",
+			expectedRendered: "https://bidder.com/sync?redirect=%7B%7BxSyncerKey%7D%7D",
```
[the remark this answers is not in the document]
SyntaxNode: Good catch. They used to be, but I've since updated the code to allow these and escape them instead of considering them macros. This was necessary since TappX uses a double curly brace in their user macro.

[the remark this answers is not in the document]
SyntaxNode: Fixed.

hhhjort (approved):
LGTM

bsardo (dismissed):
LGTM

config/bidderinfo_validate_test.go
```diff
@@ -54,7 +55,16 @@ func TestBidderInfoFiles(t *testing.T) {
...
+		err = validateSyncer(fileInfoContent)
+		assert.NoError(t, err, "Invalid syncer config in static/bidder-info/%s: %v", fileInfo.Name(), err)
+
+		fileNameWithoutExtension := fileInfo.Name()[:len(fileInfo.Name())-len(filepath.Ext(fileInfo.Name()))]
+		bidderInfos[fileNameWithoutExtension] = fileInfoContent
 	}
+
+	errs := validateSyncers(t, bidderInfos)
+	assert.Empty(t, errs, "syncer errors")
 }
```
SyntaxNode, line 68: I improved the unit tests to use the same exact loading mechanism the app uses at startup to ensure no issues / the same validation is performed.

usersync/syncersbuilder.go
```diff
@@ -80,6 +80,20 @@ func BuildSyncers(hostConfig *config.Configuration, bidderInfos config.BidderInf
...
+		return false
+	}
+
+	if cfg.Syncer == nil {
+		return false
+	}
+
+	// a syncer may provide just a Supports field to provide hints to the host. we should only try to create a syncer
+	// if there is at least one non-Supports value populated.
+	return cfg.Syncer.Key != "" || cfg.Syncer.Default != "" || cfg.Syncer.IFrame != nil || cfg.Syncer.Redirect != nil || cfg.Syncer.SupportCORS != nil
```
SyntaxNode, line 94: This is a bit hacky. Perhaps it would be better to use different data structures for yaml vs config overrides... but I think this works.

usersync/syncersbuilder_test.go
```diff
@@ -77,21 +84,27 @@ func TestBuildSyncers(t *testing.T) {
...
-			givenConfig:         hostConfig,
-			givenBidderInfos:    map[string]config.BidderInfo{"bidder1": infoKeyAEmpty, "bidder2": infoKeyAError},
-			expectedErrorHeader: "user sync (1 error)",
-			expectedErrorSegments: []string{
-				"cannot create syncer for bidder bidder2 with key a. default is set to redirect but no redirect endpoint is configured\n",
+			description:      "Many - Sync Error - Bidder Correct",
+			givenConfig:      hostConfig,
+			givenBidderInfos: map[string]config.BidderInfo{"bidder1": infoKeyAEmpty, "bidder2": infoKeyAError},
+			expectedErrors: []string{
+				"cannot create syncer for bidder bidder2 with key a: default is set to redirect but no redirect endpoint is configured",
```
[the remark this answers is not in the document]
SyntaxNode: Correct. Exactly. I wrote it to cover the case where two bidders share the same key but the primary configuration (bidder2 here) is invalid.

config/bidderinfo_validate_test.go
```diff
@@ -121,28 +127,33 @@ func validatePlatformInfo(info *config.PlatformInfo) error {
...
+	if bidderInfo.Syncer == nil {
+		return nil
 	}
 
-	// ensure final macro substitution
-	privacyPolicies := privacy.Policies{}
-	_, err = syncer.GetSync([]usersync.SyncType{usersync.SyncTypeIFrame, usersync.SyncTypeRedirect}, privacyPolicies)
-	if err != nil {
-		return fmt.Errorf("syncer has invalid macro: %s", err)
+	for _, v := range bidderInfo.Syncer.Supports {
```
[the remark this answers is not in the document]
SyntaxNode: Good point. It was my intention to emit a warning for supported user syncs which aren't configured. I've added that feature with the supports validation.

router/router_test.go
```diff
@@ -74,6 +74,178 @@ func TestExchangeMap(t *testing.T) {
...
+	}
+
+	for _, test := range testCases {
+		resultErr := applyBidderInfoConfigOverrides(test.givenBidderInfos, test.givenAdaptersCfg)
+		if test.expectedError == "" {
+			assert.NoError(t, resultErr, test.description+":err")
+			assert.Equal(t, test.expectedBidderInfos, test.givenBidderInfos, test.description+":result")
+		} else {
+			assert.EqualError(t, resultErr, test.expectedError, test.description+":err")
+		}
```
SyntaxNode, line 148: I realize that I'm not asserting the glog warnings. That's because we don't have a good way to intercept the log writes for tests. I'd like to see that fixed, but that would be too much work for this PR. I manually verified for now.

router/router.go
```diff
@@ -307,14 +327,95 @@ func New(cfg *config.Configuration, rateConvertor *currency.RateConverter) (r *R
...
+		if info.Syncer == nil {
+			continue
+		}
+
+		for _, endpoint := range info.Syncer.Supports {
+			endpointLower := strings.ToLower(endpoint)
+			switch endpointLower {
+			case "iframe":
+				if info.Syncer.IFrame == nil {
+					glog.Warningf("bidder %s supports iframe user sync, but doesn't have a default and must be configured by the host", name)
```
hhhjort, line 405: Just wanted to confirm that you don't feel a need to test for the warning. You do test for the error below.

SyntaxNode: I added a comment about that lack of coverage a little while ago.

I'd really like to test the warning, but we don't have an interface for logging and intercepting glog through output redirection isn't a good option. I think adding that logging interface is too much to do here, so I leave it for a future project. There are many other glog warnings we equally have no automated test coverage on. I ran through the test cases manually to ensure the proper glog output.

I also thought of using warnings and errors as output instead, but then I need to filter them out in the router and it seemed to be getting messy IMHO.

Happy to add the assertions if you have an idea for how to reasonable approach it.

bsardo (dismissed):
LGTM

hhhjort (dismissed):
LGTM

SyntaxNode:
Docs PR: https://github.com/prebid/prebid.github.io/pull/3220

bsardo (approved):

Issue #1876, opened 2021-06-03
Remove lifestreet adapter


bretg:
We received notification from the lifestreet team that they no longer wish to support their PBS adapter.

We have email from Yulia and Michael at lifestreet, but adding the GitHub handle to make sure everyone knows we're going to remove the adapter 

^^ @Lifestreet

SyntaxNode:
I appreciate the confirmation, but am sad to see them leave. I'm targeting removal in PBS-Go 0.163.0.

SyntaxNode:
Removed in PBS-Go 0.163.0.

bretg:
Removed with PBS-Java 1.68

PR #1883, opened 2021-06-09
Remove LifeStreet + Legacy Cleanup

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/1876.

LifeStreet was the last bidder which exclusively used the legacy model. The code to map a legacy model adapter to an OpenRTB model adapter is no longer required and is remove with this PR.


hhhjort (approved):
LGTM

exchange/adapter_util.go
```diff
@@ -127,7 +77,9 @@ func GetActiveBidders(infos config.BidderInfos) map[string]openrtb_ext.BidderNam
 
 // GetDisabledBiddersErrorMessages returns a map of error messages for disabled bidders.
 func GetDisabledBiddersErrorMessages(infos config.BidderInfos) map[string]string {
-	disabledBidders := make(map[string]string)
+	disabledBidders := map[string]string{
```
mansinahar, line 80: From what I understand, Lifestreet being an exclusive legacy adapter was only supported on the `/auction` endpoint, correct? If that is the case, then do we need to add it in the `disabledAdapters` map here? Because this map is only used by openrtb2 endpoints and they never supported `lifestreet` to start with

hhhjort: All legacy adapters had a wrapper that would allow them to be used with the openrtb2 endpoint. So yes we do.

mansinahar (approved):

PR #1885, opened 2021-06-09
Sovrn fix for issue #1787

rachelrj:
Support floors as required by prebid server. If both custom floor param and impression floor param exist, use impression floor param. Continue to support custom floor param.


rachelrj:
I closed because I am having a difficult time with the git history. I opened new PR using my personal fork.

Issue #1075, opened 2019-10-16
Echo SR video attributes in OpenRTB response


bszekely1:
## Type of issue

Intent to implement

## Description

Prebid Mobile (SDK) is adding video support that includes rendering video using the devices player. Several video specific fields can be set in the Stored Request DB that the device player would not have context to at time of render. The requester (in this case Prebid SDK) can send a signal to Prebid Server, which Stored Request object it expects PBS to respond with in an extension object. 

## Goals

- Supply Prebid Server a signal to respond with specific Stored Request attributes in it's Bid Response

## Proposed Design

Copy all the stored requests attributes for that impression into an extension object called "storedRequestAttributes".

### Prebid Server OpenRTB Changes
The requester will signal to Prebid Server which stored request object it would like to echo back in the Bid Response.

SDK will not expose any features to the publisher facing API. All requesting objects to Prebid Server will be generated from the Prebid SDK itself.

OpenRTB request to Prebid Server
```
{
   ... 
    
    "imp": [
        {
            "banner": {
                "format": [
                    {
                        "h": 250, 
                        "w": 300
                    }
                ]
            }, 
            "ext": {
                "prebid": {
                    "storedrequest": {
                        "id": "xxx"
                    },
                    "options": {
                        "echovideoattrs": true
                    } 
                }
            }, 
            "id": "PrebidMobile", 
            "secure": 1
        }
    ]
    ...
}
```

Proposed OpenRTB response
```
{
    "seatbid": [
        {
            "bid": [
                {
                   ...
                    "ext": {
                        "prebid": {
                            "type": "video",
                            ...
                            },
                            "storedrequestattributes":{
                                "maxduration": 60, 
                                "mimes": [
                                    "video/mp4"
                                ], 
                                "minduration": 15, 
                                "protocols": [
                                    1, 
                                    2
                                ], 
                                "skipafter": 0, 
                                "skipmin": 0, 
                                "startdelay": 0,
                                "playbackmethod": [1]
                            }
                        },
                        "bidder": {
                            "prebid": {
                                "type": "video"
                            }
                        }
                    }
                }
            ]
        }
    ],
..
}
```

### Prebid SDK Changes
Prebid SDK will be responsible for reading any or all attributes. 

### Prebid Server Changes

Prebid Server will be required to parse the Stored Request object to look for video attributes. If one or more attributes are present, echo in OpenRTB extension object in the response. Otherwise, do not include the "storedRequestAttributes" object.

[...]

PR #1889, opened 2021-06-15
Mirror stored imp data to bid response

VeronikaSolovei9:
Implements https://github.com/prebid/prebid-server/issues/1075

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1396,6 +1396,10 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		}
 		imps[idIndices[i]] = resolvedImp
 
+		// This is substantially faster for reading values from a json blob than the Go json package,
+		// but keep in mind that each jsonparser.GetXXX call re-parses the entire json.
+		// Based on performance measurements, the tipping point of efficiency is around 4 calls.
+		// At that point, please consider switching to EachKey to use a single pass.
 		includeVideoAttributes, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
 		if err != nil && err != jsonparser.KeyPathNotFoundError {
 			return nil, nil, []error{err}
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1378,7 +1378,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
-	impToStoredReq := make(map[string][]byte)
+	impToStoredVideoReq := make(map[string][]byte)
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
 
@@ -1392,7 +1392,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 					err = fmt.Errorf("imp.ext.prebid.storedrequest.id %s: Stored Imp has Invalid JSON: %s", impIds[i], Err)
 				}
 			}
-			return nil, impToStoredReq, []error{err}
+			return nil, impToStoredVideoReq, []error{err}
 		}
 		imps[idIndices[i]] = resolvedImp
 
@@ -1416,7 +1416,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 				//extract video attributes only
 				videoData, _, _, err := jsonparser.Get(storedImps[impIds[i]], "video")
 				if err == nil {
-					impToStoredReq[impId] = videoData
+					impToStoredVideoReq[impId] = videoData
 				}
 			}
 		}
@@ -1433,7 +1433,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		}
 	}
 
-	return resolvedRequest, impToStoredReq, nil
+	return resolvedRequest, impToStoredVideoReq, nil
 }
 
 // parseImpInfo parses the request JSON and returns several things about the Imps
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1378,7 +1378,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
-	impToStoredVideoReq := make(map[string][]byte)
+	impToStoredReq := make(map[string][]byte)
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
 
@@ -1392,10 +1392,14 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 					err = fmt.Errorf("imp.ext.prebid.storedrequest.id %s: Stored Imp has Invalid JSON: %s", impIds[i], Err)
 				}
 			}
-			return nil, impToStoredVideoReq, []error{err}
+			return nil, nil, []error{err}
 		}
 		imps[idIndices[i]] = resolvedImp
 
+		impId, err := jsonparser.GetString(resolvedImp, "id")
+		if err != nil {
+			return nil, nil, []error{err}
+		}
 		// This is substantially faster for reading values from a json blob than the Go json package,
 		// but keep in mind that each jsonparser.GetXXX call re-parses the entire json.
 		// Based on performance measurements, the tipping point of efficiency is around 4 calls.
@@ -1404,27 +1408,9 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		if err != nil && err != jsonparser.KeyPathNotFoundError {
 			return nil, nil, []error{err}
 		}
-
 		if includeVideoAttributes {
-			//determine if this is video impression
-			_, _, videoOffset, err := jsonparser.Get(resolvedImp, "video")
-			if err != nil && err != jsonparser.KeyPathNotFoundError {
-				return nil, nil, []error{err}
-			}
-
-			if videoOffset != -1 {
-				impId, err := jsonparser.GetString(resolvedImp, "id")
-				if err != nil {
-					return nil, nil, []error{err}
-				}
-				//extract video attributes only
-				videoData, _, _, err := jsonparser.Get(storedImps[impIds[i]], "video")
-				if err == nil {
-					impToStoredVideoReq[impId] = videoData
-				}
-			}
+			impToStoredReq[impId] = storedImps[impIds[i]]
 		}
-
 	}
 	if len(impIds) > 0 {
 		newImpJson, err := json.Marshal(imps)
@@ -1437,7 +1423,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		}
 	}
 
-	return resolvedRequest, impToStoredVideoReq, nil
+	return resolvedRequest, impToStoredReq, nil
 }
 
 // parseImpInfo parses the request JSON and returns several things about the Imps
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1397,10 +1397,8 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		imps[idIndices[i]] = resolvedImp
 
 		includeStoredImps, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
-		if err == jsonparser.KeyPathNotFoundError {
-			err = nil
-		}
-		if err != nil {
+
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
 			return nil, nil, []error{err}
 		}
 		if includeStoredImps {
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1379,6 +1379,8 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
+		//writing stored imp data back to imp.ext
+		resolvedImp, err = jsonparser.Set(resolvedImp, storedImps[impIds[i]], "ext", "prebid", exchange.StoredRequestAttributes)
 		if err != nil {
 			hasErr, Err := getJsonSyntaxError(imps[idIndices[i]])
 			if hasErr {
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1382,9 +1382,6 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
 
-		impId, err := jsonparser.GetString(resolvedImp, "id")
-		impToStoredReq[impId] = storedImps[impIds[i]]
-
 		if err != nil {
 			hasErr, Err := getJsonSyntaxError(imps[idIndices[i]])
 			if hasErr {
@@ -1398,6 +1395,21 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 			return nil, impToStoredReq, []error{err}
 		}
 		imps[idIndices[i]] = resolvedImp
+
+		includeStoredImps, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
+		if err == jsonparser.KeyPathNotFoundError {
+			err = nil
+		}
+		if err != nil {
+			return nil, nil, []error{err}
+		}
+		if includeStoredImps {
+			impId, err := jsonparser.GetString(resolvedImp, "id")
+			if err != nil {
+				return nil, nil, []error{err}
+			}
+			impToStoredReq[impId] = storedImps[impIds[i]]
+		}
 	}
 	if len(impIds) > 0 {
 		newImpJson, err := json.Marshal(imps)
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1378,7 +1378,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
-	impExtInfoMap := make(map[string]exchange.ImpExtInfo)
+	impExtInfoMap := make(map[string]exchange.ImpExtInfo, len(impIds))
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
 
@@ -1409,7 +1409,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 			return nil, nil, []error{err}
 		}
 		if storedImps[impIds[i]] != nil {
-			impExtInfoMap[impId] = exchange.ImpExtInfo{includeVideoAttributes, storedImps[impIds[i]]}
+			impExtInfoMap[impId] = exchange.ImpExtInfo{EchoVideoAttrs: includeVideoAttributes, StoredImp: storedImps[impIds[i]]}
 		}
 	}
 	if len(impIds) > 0 {
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -235,7 +235,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 // possible, it will return errors with messages that suggest improvements.
 //
 // If the errors list has at least one element, then no guarantees are made about the returned request.
-func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_ext.RequestWrapper, impToStoredReq map[string][]byte, errs []error) {
+func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_ext.RequestWrapper, impToStoredReq map[string]exchange.StoredImpData, errs []error) {
 	req = &openrtb_ext.RequestWrapper{}
 	req.BidRequest = &openrtb2.BidRequest{}
 	errs = nil
@@ -1317,7 +1317,7 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
 	return false, ""
 }
 
-func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, map[string][]byte, []error) {
+func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, map[string]exchange.StoredImpData, []error) {
 	// Parse the Stored Request IDs from the BidRequest and Imps.
 	storedBidRequestId, hasStoredBidRequest, err := getStoredRequestId(requestJson)
 	if err != nil {
@@ -1378,7 +1378,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
-	impToStoredReq := make(map[string][]byte)
+	impToStoredReq := make(map[string]exchange.StoredImpData)
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
 
@@ -1408,8 +1408,8 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		if err != nil && err != jsonparser.KeyPathNotFoundError {
 			return nil, nil, []error{err}
 		}
-		if includeVideoAttributes {
-			impToStoredReq[impId] = storedImps[impIds[i]]
+		if storedImps[impIds[i]] != nil {
+			impToStoredReq[impId] = exchange.StoredImpData{includeVideoAttributes, storedImps[impIds[i]]}
 		}
 	}
 	if len(impIds) > 0 {
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -135,7 +135,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		deps.analytics.LogAuctionObject(&ao)
 	}()
 
-	req, impToStoredReq, errL := deps.parseRequest(r)
+	req, impExtInfoMap, errL := deps.parseRequest(r)
 
 	if errortypes.ContainsFatalError(errL) && writeError(errL, w, &labels) {
 		return
@@ -192,7 +192,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		LegacyLabels:               labels,
 		Warnings:                   warnings,
 		GlobalPrivacyControlHeader: secGPC,
-		ImpToStoredReq:             impToStoredReq,
+		ImpToStoredReq:             impExtInfoMap,
 	}
 
 	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
@@ -235,7 +235,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 // possible, it will return errors with messages that suggest improvements.
 //
 // If the errors list has at least one element, then no guarantees are made about the returned request.
-func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_ext.RequestWrapper, impToStoredReq map[string]exchange.StoredImpData, errs []error) {
+func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_ext.RequestWrapper, impExtInfoMap map[string]exchange.ImpExtInfo, errs []error) {
 	req = &openrtb_ext.RequestWrapper{}
 	req.BidRequest = &openrtb2.BidRequest{}
 	errs = nil
@@ -263,7 +263,7 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
 	defer cancel()
 
 	// Fetch the Stored Request data and merge it into the HTTP request.
-	if requestJson, impToStoredReq, errs = deps.processStoredRequests(ctx, requestJson); len(errs) > 0 {
+	if requestJson, impExtInfoMap, errs = deps.processStoredRequests(ctx, requestJson); len(errs) > 0 {
 		return
 	}
 
@@ -1317,7 +1317,7 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
 	return false, ""
 }
 
-func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, map[string]exchange.StoredImpData, []error) {
+func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, map[string]exchange.ImpExtInfo, []error) {
 	// Parse the Stored Request IDs from the BidRequest and Imps.
 	storedBidRequestId, hasStoredBidRequest, err := getStoredRequestId(requestJson)
 	if err != nil {
@@ -1378,7 +1378,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
-	impToStoredReq := make(map[string]exchange.StoredImpData)
+	impExtInfoMap := make(map[string]exchange.ImpExtInfo)
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
 
@@ -1409,7 +1409,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 			return nil, nil, []error{err}
 		}
 		if storedImps[impIds[i]] != nil {
-			impToStoredReq[impId] = exchange.StoredImpData{includeVideoAttributes, storedImps[impIds[i]]}
+			impExtInfoMap[impId] = exchange.ImpExtInfo{includeVideoAttributes, storedImps[impIds[i]]}
 		}
 	}
 	if len(impIds) > 0 {
@@ -1423,7 +1423,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		}
 	}
 
-	return resolvedRequest, impToStoredReq, nil
+	return resolvedRequest, impExtInfoMap, nil
 }
 
 // parseImpInfo parses the request JSON and returns several things about the Imps
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1401,12 +1401,24 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		if err != nil && err != jsonparser.KeyPathNotFoundError {
 			return nil, nil, []error{err}
 		}
-		if includeStoredImps {
+
+		//determine if this is video impression
+		_, _, offset, err := jsonparser.Get(resolvedImp, "video")
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
+			return nil, nil, []error{err}
+		}
+
+		if includeStoredImps && offset != -1 {
 			impId, err := jsonparser.GetString(resolvedImp, "id")
 			if err != nil {
 				return nil, nil, []error{err}
 			}
-			impToStoredReq[impId] = storedImps[impIds[i]]
+			//extract video attributes only
+			videoData, _, _, err := jsonparser.Get(storedImps[impIds[i]], "video")
+			if err == nil {
+				impToStoredReq[impId] = videoData
+			}
+
 		}
 	}
 	if len(impIds) > 0 {
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -135,7 +135,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		deps.analytics.LogAuctionObject(&ao)
 	}()
 
-	req, errL := deps.parseRequest(r)
+	req, impToStoredReq, errL := deps.parseRequest(r)
 
 	if errortypes.ContainsFatalError(errL) && writeError(errL, w, &labels) {
 		return
@@ -192,6 +192,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		LegacyLabels:               labels,
 		Warnings:                   warnings,
 		GlobalPrivacyControlHeader: secGPC,
+		ImpToStoredReq:             impToStoredReq,
 	}
 
 	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
@@ -234,7 +235,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 // possible, it will return errors with messages that suggest improvements.
 //
 // If the errors list has at least one element, then no guarantees are made about the returned request.
-func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_ext.RequestWrapper, errs []error) {
+func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_ext.RequestWrapper, impToStoredReq map[string][]byte, errs []error) {
 	req = &openrtb_ext.RequestWrapper{}
 	req.BidRequest = &openrtb2.BidRequest{}
 	errs = nil
@@ -262,7 +263,7 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
 	defer cancel()
 
 	// Fetch the Stored Request data and merge it into the HTTP request.
-	if requestJson, errs = deps.processStoredRequests(ctx, requestJson); len(errs) > 0 {
+	if requestJson, impToStoredReq, errs = deps.processStoredRequests(ctx, requestJson); len(errs) > 0 {
 		return
 	}
 
@@ -1316,15 +1317,15 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
 	return false, ""
 }
 
-func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, []error) {
+func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, map[string][]byte, []error) {
 	// Parse the Stored Request IDs from the BidRequest and Imps.
 	storedBidRequestId, hasStoredBidRequest, err := getStoredRequestId(requestJson)
 	if err != nil {
-		return nil, []error{err}
+		return nil, nil, []error{err}
 	}
 	imps, impIds, idIndices, errs := parseImpInfo(requestJson)
 	if len(errs) > 0 {
-		return nil, errs
+		return nil, nil, errs
 	}
 
 	// Fetch the Stored Request data
@@ -1334,7 +1335,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	}
 	storedRequests, storedImps, errs := deps.storedReqFetcher.FetchRequests(ctx, storedReqIds, impIds)
 	if len(errs) != 0 {
-		return nil, errs
+		return nil, nil, errs
 	}
 
 	// Apply the Stored BidRequest, if it exists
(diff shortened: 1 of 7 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -1381,6 +1381,8 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
+		//writing stored imp data back to imp.ext
+		resolvedImp, err = jsonparser.Set(resolvedImp, storedImps[impIds[i]], "ext", "prebid", exchange.StoredRequestAttributes)
```
SyntaxNode, line 1385: I'm afraid you can't use `jsonparser.Set`. That's not a production ready API and we have encountered problems with it before.  `jsonparser` should be used for reads only.

exchange/exchange.go
```diff
@@ -447,7 +450,17 @@ func (e *exchange) getAllBids(
 			var reqInfo adapters.ExtraRequestInfo
 			reqInfo.PbsEntryPoint = bidderRequest.BidderLabels.RType
 			reqInfo.GlobalPrivacyControlHeader = globalPrivacyControlHeader
-			bids, err := e.adapterMap[bidderRequest.BidderCoreName].requestBid(ctx, bidderRequest.BidRequest, bidderRequest.BidderName, adjustmentFactor, conversions, &reqInfo, accountDebugAllowed, headerDebugAllowed)
+
+			//map imp to stored request data if exists
+			hasStoredReq, impsToStoredReq, parseErr := parseStoredImp(bidderRequests)
```
SyntaxNode, line 455: The tests are saying this line introduces a data race.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Rafactored according to discussion.

exchange/exchange.go
VeronikaSolovei9: Modified.

endpoints/openrtb2/auction.go
```diff
@@ -1371,16 +1372,21 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
 
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
+	impToStoredReq := make(map[string][]byte)
 	for i := 0; i < len(impIds); i++ {
 		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
+
+		impId, err := jsonparser.GetString(resolvedImp, "id")
+		impToStoredReq[impId] = storedImps[impIds[i]]
```
SyntaxNode, line 1388: Please check the error form the call to `GetString`.

exchange/exchange.go
```diff
@@ -474,6 +478,12 @@ func (e *exchange) getAllBids(
 					e.me.RecordAdapterBidReceived(bidderRequest.BidderLabels, bid.bidType, bid.bid.AdM != "")
 				}
 			}
+
+			//put stored req data back to bid.ext, map by bid.imp_id
+			if len(impToStoredReq) != 0 {
+				insertStoredImpData(impToStoredReq, bids)
```
SyntaxNode, line 484: I think we missed a key part of the spec. This should only be done if a new config option `imp.ext.prebid.options.echovideoattrs` is set to true.

exchange/exchange.go
```diff
@@ -497,6 +507,15 @@ func (e *exchange) getAllBids(
 	return adapterBids, adapterExtra, bidsFound
 }
 
+func insertStoredImpData(impsToStoredRequest map[string][]byte, bids *pbsOrtbSeatBid) {
+	for _, bid := range bids.bids {
+		if val, ok := impsToStoredRequest[bid.bid.ImpID]; ok {
+			result, _ := jsonparser.Set(bid.bid.Ext, val, StoredRequestAttributes)
```
SyntaxNode, line 513: Don't forget to remove the `jsonparser.Set`. :) I also think we may need to transform the data, please see this example:

```
"ext": {
    "prebid": {
        "type": "video",
        ...
    },
    "storedrequestattributes": {
        "maxduration": 60,
        "mimes": [
            "video/mp4"
        ],
        "minduration": 15,
        "protocols": [
            1,
            2
        ],
        "skipafter": 0,
        "skipmin": 0,
        "startdelay": 0,
        "playbackmethod": [1]
    }
}
```

Is that processed or is that exactly from the stored imp?

endpoints/openrtb2/auction.go
VeronikaSolovei9: Added.

exchange/exchange.go
VeronikaSolovei9: Wow, for some reason I missed this part. Good catch! Check added.

exchange/exchange.go
VeronikaSolovei9: Right, modified to marshal/unmarshal logic. Please take another look. 

With current implementation I just write the entire stored impression back to bid.ext, no processing. 
What kind of processing do you mean?

exchange/exchange.go
```diff
@@ -497,6 +508,30 @@ func (e *exchange) getAllBids(
...
+			storedImpVal := make(map[string]interface{})
+			err := json.Unmarshal(bid.bid.Ext, &bidExtData)
+			if err != nil {
+				return err
+			}
+			err = json.Unmarshal(val, &storedImpVal)
+			if err != nil {
+				return err
+			}
+			bidExtData[StoredRequestAttributes] = storedImpVal
```
mansinahar, line 524: Is there a reason why we are inserting the entire stored request object here? I see the ask in the spec is to parse and include only video attributes.
```
Prebid Server will be required to parse the Stored Request object to look for video attributes. If one or more attributes are present, echo in OpenRTB extension object in the response. Otherwise, do not include the "storedRequestAttributes" object.
```

I check Prebid Java and they parse the video attributes of the stored imp and echo just those in the bid ext as well, unless I am missing something. Check here: https://github.com/prebid/prebid-server-java/pull/537/files#diff-f0a1b91c0eb7400d0051d08a2f6b06bedd05e1462b33501abe80cd8641439334R140. 

Also, this is probably too late as this has already been implemented in Java but having the flag specific to video attributes and the bid response ext being more general `storedrequestattributes` is kinda confusing :/

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,37 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
 			}
-			return nil, []error{err}
+			return nil, impToStoredReq, []error{err}
 		}
 		imps[idIndices[i]] = resolvedImp
+
+		includeStoredImps, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
+		if err == jsonparser.KeyPathNotFoundError {
+			err = nil
+		}
```
mansinahar, line 1404: You can just merge these two if conditions as:
```
if err != nil && err != jsonparser.KeyPathNotFoundError {
	return nil, nil, []error{err}
}
```

exchange/exchange.go
```diff
@@ -497,6 +508,30 @@ func (e *exchange) getAllBids(
...
+			result, err := json.Marshal(bidExtData)
+			if err != nil {
+				return err
+			}
+			bid.bid.Ext = result
```
SyntaxNode, line 529: Looks good. With the code as it is now, you don't need to unmarshal the stored request. This could be simplified as:
```
bidExtData := make(map[string]json.RawMessage)
err := json.Unmarshal(bid.bid.Ext, &bidExtData)
if err != nil {
	return err
}
bidExtData[StoredRequestAttributes] = val
result, err := json.Marshal(bidExtData)
if err != nil {
	return err
}
bid.bid.Ext = result
```

This approach may also be faster since `interface{}` forces type discovery and `json.RawMessage` does not.

Anyway, this may be a moot point as you'll likely need to unmarshal the stored request - at least partially - to accomodate @mansinahar's comment.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Yes, modified!

exchange/exchange.go
VeronikaSolovei9: Yes, it works! Modified

exchange/exchange.go
VeronikaSolovei9: Good point! Modified to add stored request attributes for video imps only. Also added another check if stored imp had a video data itself (it may have only `ext `for example). If there is video data in stored impression - then I put it back to `bid.ext. storedrequestattributes`

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,47 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
 		imps[idIndices[i]] = resolvedImp
+
+		includeStoredImps, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
+
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
+			return nil, nil, []error{err}
+		}
+
+		//determine if this is video impression
+		_, _, offset, err := jsonparser.Get(resolvedImp, "video")
```
SyntaxNode, line 1408: Please consider a clearer name for `offset`, something along the lines of `videoOffset`.

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,47 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
+			return nil, nil, []error{err}
+		}
+
+		if includeStoredImps && offset != -1 {
+			impId, err := jsonparser.GetString(resolvedImp, "id")
```
SyntaxNode, line 1414: A Note On `jsonparser`:
This is substantially faster for reading values from a json blob than the Go json package, but you need to keep in mind that each `jsonparser.GetXXX` call re-parses the entire json. Based on performance measurements, the tipping point of efficiency is around 4 calls.  At that point, please consider switching to `EachKey` to use a single pass.

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,47 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+			if err != nil {
+				return nil, nil, []error{err}
+			}
+			//extract video attributes only
+			videoData, _, _, err := jsonparser.Get(storedImps[impIds[i]], "video")
+			if err == nil {
+				impToStoredReq[impId] = videoData
+			}
```
SyntaxNode, line 1422: To improve performance for non-video ads, please nest the other jsonparser checks under `includeStoredImps`. ie:

```
includeVideoAttributes, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
if err != nil && err != jsonparser.KeyPathNotFoundError {
	return nil, nil, []error{err}
}

if includeVideoAttributes {
	_, _, videoImpOffset, err := jsonparser.Get(resolvedImp, "video")
	if err != nil && err != jsonparser.KeyPathNotFoundError {
		return nil, nil, []error{err}
	}

	if videoImpOffset != -1 {
		impId, err := jsonparser.GetString(resolvedImp, "id")
		if err != nil {
			return nil, nil, []error{err}
		}
		//extract video attributes only
		videoData, _, _, err := jsonparser.Get(storedImps[impIds[i]], "video")
		if err == nil {
			impToStoredReq[impId] = videoData
		}

	}
}
```

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,47 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+
+		if includeStoredImps && offset != -1 {
+			impId, err := jsonparser.GetString(resolvedImp, "id")
+			if err != nil {
+				return nil, nil, []error{err}
+			}
+			//extract video attributes only
+			videoData, _, _, err := jsonparser.Get(storedImps[impIds[i]], "video")
+			if err == nil {
+				impToStoredReq[impId] = videoData
```
SyntaxNode, line 1421: The name `impToStoredReq` is now a bit misleading since it's a subset for video auctions. Perhaps it would be best to include the full stored request here and "cut out" the video section to bid response handler?

endpoints/openrtb2/auction_test.go
```diff
@@ -3209,6 +3273,23 @@ var testFinalRequests = []string{
...
+	`{
+        				"w":200,
+        				"h":300
+        			}`,
+	`{
+					"w":200,
+					"h":300
+				}`,
+	``,
+	``,
```
SyntaxNode, line 3290: Nitpick: This is harder to read with the whitespace. Please consider using just a few spaces instead?

util/jsonutil/jsonutil_test.go
```diff
@@ -17,71 +16,71 @@ func TestDropElement(t *testing.T) {
 		errorContains   string
 	}{
 		{
-			description:     "Drop Single Element After Another Element",
+			description:     "Drop single element after another element",
```
SyntaxNode, line 19: IMHO: Either casing approach is fine.

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,47 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
 			}
-			return nil, []error{err}
+			return nil, impToStoredReq, []error{err}
 		}
 		imps[idIndices[i]] = resolvedImp
+
+		includeStoredImps, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
+
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
+			return nil, nil, []error{err}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Added

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,47 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+		}
+
+		//determine if this is video impression
+		_, _, offset, err := jsonparser.Get(resolvedImp, "video")
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
+			return nil, nil, []error{err}
```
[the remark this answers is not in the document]
VeronikaSolovei9: I haven't found the way to "break" it here. Json should be valid already at this point and I can only get KeyNotFound error

endpoints/openrtb2/auction.go
VeronikaSolovei9: Renamed

endpoints/openrtb2/auction.go
VeronikaSolovei9: Yes, modified

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: Formatted

endpoints/openrtb2/auction.go
VeronikaSolovei9: I went through this again. The reason I put this here is to optimize amount of data we have in memory (have only video impressions we will use later) and to reduce complexity of code execution in `insertStoredImpData` function: if no stored video data found - do not execute any logic. Also if we leave this code here - it will fit the logic around it: prepare stored request/video data. 
I'm open to modify this and move this piece of code to bid response handler (insertStoredImpData). Please advice.

VeronikaSolovei9: Renamed `impToStoredReq` to `impToStoredVideoReq`

endpoints/openrtb2/auction.go
VeronikaSolovei9: Good point. We are about to reach 4 calls. 
The very last call takes `storedImps[impIds[i]]` as input, 3 others `resolvedImp`
I added a comment about this.

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
SyntaxNode: No worries. Thank you for trying. I understand if it's not possible.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Modified according to discussion

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,52 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 					err = fmt.Errorf("imp.ext.prebid.storedrequest.id %s: Stored Imp has Invalid JSON: %s", impIds[i], Err)
 				}
 			}
-			return nil, []error{err}
+			return nil, impToStoredVideoReq, []error{err}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Agreed, modified

endpoints/openrtb2/auction.go
```diff
@@ -1391,22 +1394,52 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+				return nil, nil, []error{err}
+			}
+
+			if videoOffset != -1 {
+				impId, err := jsonparser.GetString(resolvedImp, "id")
+				if err != nil {
+					return nil, nil, []error{err}
+				}
+				//extract video attributes only
+				videoData, _, _, err := jsonparser.Get(storedImps[impIds[i]], "video")
```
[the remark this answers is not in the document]
VeronikaSolovei9: Modified according to discussion

VeronikaSolovei9:
@mansinahar @SyntaxNode  code modified according to our code review discussion, please re-review

exchange/exchange.go
```diff
@@ -135,6 +136,11 @@ func NewExchange(adapters map[openrtb_ext.BidderName]adaptedBidder, cache prebid
 	}
 }
 
+type StoredImpData struct {
+	IncludeVideoAttributes bool
+	Data                   []byte
```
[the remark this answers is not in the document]

exchange/exchange.go
```diff
@@ -135,6 +136,11 @@ func NewExchange(adapters map[openrtb_ext.BidderName]adaptedBidder, cache prebid
 	}
 }
 
+type StoredImpData struct {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Modified

exchange/exchange_test.go
```diff
@@ -3443,6 +3443,41 @@ func TestUpdateHbPbCatDur(t *testing.T) {
 	}
 }
 
+func TestInsertStoredImpData(t *testing.T) {
+	impsToStoredRequest := make(map[string]ImpExtInfo)
+	impsToStoredRequest["imp_idApn1_1"] = ImpExtInfo{true, []byte(`{"video":{"h":480,"mimes":["video/mp4"]}}`)}
```
mansinahar, line 3448: Please consider using table tests here for ease of readability

exchange/exchange_test.go
```diff
@@ -3443,6 +3443,41 @@ func TestUpdateHbPbCatDur(t *testing.T) {
 	}
 }
 
+func TestInsertStoredImpData(t *testing.T) {
+	impsToStoredRequest := make(map[string]ImpExtInfo)
```
mansinahar, line 3447: Please rename this to `impExtInfoMap` for consistency with the new names

endpoints/openrtb2/auction.go
```diff
@@ -1369,16 +1370,18 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
-			return nil, []error{err}
+			return nil, nil, []error{err}
 		}
 		resolvedRequest = aliasedRequest
 	}
 
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
+	impExtInfoMap := make(map[string]exchange.ImpExtInfo)
```
SyntaxNode, line 1381: Should we set the initial capacity to `impIds`? .. or do we expect most impressions seen by PBS to not use a stored impression.

endpoints/openrtb2/auction.go
```diff
@@ -1389,22 +1392,38 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 					err = fmt.Errorf("imp.ext.prebid.storedrequest.id %s: Stored Imp has Invalid JSON: %s", impIds[i], Err)
 				}
 			}
-			return nil, []error{err}
+			return nil, nil, []error{err}
 		}
 		imps[idIndices[i]] = resolvedImp
+
+		impId, err := jsonparser.GetString(resolvedImp, "id")
```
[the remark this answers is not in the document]
SyntaxNode: At this point, I don't think we've parsed the `resolvedImp` yet. Could the stored impression contain the imp id?

endpoints/openrtb2/auction_test.go
```diff
@@ -1035,6 +1035,41 @@ func TestStoredRequests(t *testing.T) {
...
+		openrtb_ext.BuildBidderMap(),
+		nil,
+		nil,
+		hardcodedResponseIPValidator{response: true},
+	}
+
+	for i, requestData := range testStoredRequestsErrors {
+		_, _, errList := deps.processStoredRequests(context.Background(), json.RawMessage(requestData))
+		if len(errList) == 0 {
+			t.Error("processStoredRequests should return error")
```
SyntaxNode, line 1069: Does t.Error stop the running test - or is that t.Fatal?

endpoints/openrtb2/auction_test.go
```diff
@@ -3122,56 +3193,228 @@ var testFinalRequests = []string{
...
+}`,
+}
+
+var testStoredImpIds = []string{
+	"adUnit1", "adUnit2", "adUnit1", "some-static-imp",
+}
+
+var testStoreVideoAttr = []bool{
+	true, true, false, false,
+}
```
SyntaxNode, line 3380: Style Nitpick: Consider making the slices definitions in-line.

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
VeronikaSolovei9: `Could the stored impression contain the imp id? ` - this is possible. I take id from final impression data, this id can come from request.imp[] or stored impression.

exchange/exchange.go
```diff
@@ -464,6 +473,14 @@ func (e *exchange) getAllBids(
 				ae.HttpCalls = bids.httpCalls
 			}
 
+			//put stored req data back to bid.ext, map by bid.imp_id
+			if len(impExtInfoMap) != 0 {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Not necessarily, request.imp like this is also valid and I made sure it work properly: 
```
{
      "id": "1_0",   -> imp with stored req and echovideoattrs
      "ext": {
        "prebid": {
          "storedrequest": {
            "id": "3c5416a2-e17a-4c9a-9379-3f74fd3fc378"
          },
          "options": {
            "echovideoattrs": true
          }
        }
      }
    },
    {
      "id": "1_2",  -> imp with stored req and without echovideoattrs
      "ext": {
        "prebid": {
          "storedrequest": {
            "id": "3c5416a2-e17a-4c9a-9379-3f74fd3fc378"
          }
        }
      }
    },
    {
      "id": "1_3",    -> imp without stored req and without echovideoattrs
      "video": {
        "mimes": [
          "video/mp4"
        ],
        "maxduration": 30,
        "protocols": [
          2,
          3
        ],
        "w": 640,
        "h": 480
      },
      "ext": {
        "appnexus": {
          "placementId": 14997137
        }
      }
    }
```

endpoints/openrtb2/auction.go
VeronikaSolovei9: `do we expect most impressions seen by PBS to not use a stored impression` -> yes, this is my default assumption. We now don't have many stored imps in database. 
Please refer to my comment earlier with imp array example where we want to mirror stored data in one imp out of 3.

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: For some reason I cannot answer to your comment about `resolvedImp`.
At this moment we already have final imp in byte array, but don't have it in object yet.
`resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])` 
We have raw data here that we want to return in response.

VeronikaSolovei9: `t.Error` reports error in logs, but doesn't stop the execution

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: Modified!

endpoints/openrtb2/auction.go
```diff
@@ -192,6 +192,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		LegacyLabels:               labels,
 		Warnings:                   warnings,
 		GlobalPrivacyControlHeader: secGPC,
+		ImpToStoredReq:             impExtInfoMap,
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, renamed.

endpoints/openrtb2/auction_test.go
```diff
@@ -1021,7 +1021,7 @@ func TestStoredRequests(t *testing.T) {
 	}
 
 	for i, requestData := range testStoredRequests {
-		newRequest, errList := deps.processStoredRequests(context.Background(), json.RawMessage(requestData))
+		newRequest, impToStoredReq, errList := deps.processStoredRequests(context.Background(), json.RawMessage(requestData))
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, I apologize, I forgot about renaming it in unit tests

endpoints/openrtb2/auction_test.go
```diff
@@ -1035,6 +1035,41 @@ func TestStoredRequests(t *testing.T) {
 		if !jsonpatch.Equal(newRequest, expectJson) {
 			t.Errorf("Error in processStoredRequests, test %d failed on compare\nFound:\n%s\nExpected:\n%s", i, string(newRequest), string(expectJson))
 		}
+		expectedImp := testStoredImpIds[i]
+		expectedStoredReq := json.RawMessage(testStoredImps[i])
```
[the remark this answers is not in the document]
VeronikaSolovei9: Renamed

endpoints/openrtb2/auction_test.go
```diff
@@ -1035,6 +1035,41 @@ func TestStoredRequests(t *testing.T) {
...
+		if !jsonpatch.Equal(impToStoredReq[expectedImp].StoredImp, expectedStoredReq) {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, modified

exchange/exchange_test.go
VeronikaSolovei9: This is just one scenario here. If there are more scenarios  - I would implement it as a table, agree.

exchange/exchange_test.go
VeronikaSolovei9: yes, renamed

exchange/exchange.go
```diff
@@ -502,6 +519,50 @@ func (e *exchange) getAllBids(
 	return adapterBids, adapterExtra, bidsFound
 }
 
+func insertStoredImpData(imps []openrtb2.Imp, impExtInfoMap map[string]ImpExtInfo, bids *pbsOrtbSeatBid) error {
+	for _, bid := range bids.bids {
+		//find impression
+		var impression *openrtb2.Imp
+		for _, imp := range imps {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good catch ! Modified!

exchange/exchange.go
```diff
@@ -502,6 +519,50 @@ func (e *exchange) getAllBids(
...
+				break
+			}
+		}
+		if impression == nil {
+			continue
+		}
+
+		if val, ok := impExtInfoMap[bid.bid.ImpID]; ok {
+
+			bidExtData := make(map[string]json.RawMessage)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Agree, modified. 
It's even easier to do - `impExtInfo` is the only needed variable there.

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
mansinahar: > At this point, I don't think we've parsed the resolvedImp yet

But. I don't think we need the resolvedImp to get the impID. In fact, we use the impID to get the stored Imp and construct the resolved Imp. See L1383. We have an `impsIds` array that we're looping on in this loop so why not just get the imp Id from that array as `impIds[i]`? Am I missing something here which blocks us from doing that?

exchange/exchange_test.go
VeronikaSolovei9: This test is gone due to code refactoring

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
VeronikaSolovei9: `impId` in L1383 is not imp id, it's stored imp id. It works together with `idIndices` which specifies what imp index has stored impId. This is extra variable to support requests where not all imps have stored imps. 
Please take a look at the attached screenshot. 
`for` loop in line 1383 iterates over all found stored imp ids. In my request example there are 3 impressions and first doesn't have stored imp id, but second ant third impressions have it. 
 - on iteration 0 it takes stored imp id `3c5416a2-e17a-4c9a-9379-3f74fd3fc378` and impression with index 1 from idIndexes. And merge them together. 
 - on iteration 1 it takes stored imp id `3c5416a2-e17a-4c9a-9379-3f74fd3fc378` and impression with index 2 from idIndexes. And merge them together. 
 
If we want to use `impIds[i]` we also need additional data to map it to impression id where it was used.
Using map here and in further processing is easier to understand than this `extra connection` variable. 
Hope this makes sense.
Screen Shot 2021-08-02 at 4 52 07 PM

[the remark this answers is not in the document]
VeronikaSolovei9: In my request imp ids were 1_1, 1_2 and 1_3. At the moment of merge we don't care about imp id because it can be overwritten by stored imp.  But we need to know the final impId to map it stored data.

endpoints/openrtb2/auction_test.go
```diff
@@ -1032,9 +1033,45 @@ func TestStoredRequests(t *testing.T) {
...
+		[]byte{},
+		openrtb_ext.BuildBidderMap(),
+		nil,
+		nil,
+		hardcodedResponseIPValidator{response: true},
+	}
+
+	for i, requestData := range testStoredRequestsErrors {
+		_, _, errList := deps.processStoredRequests(context.Background(), json.RawMessage(requestData))
+		if len(errList) == 0 {
```
mansinahar, line 1070: **Nitpick:** I'd recommend using `assert.Len` here rather. I know there are parts of our code where we don't use the assert library for test assertions but if we can do that going forward to keep consistency then that'd be great. The assert library makes the test assertion code a little leaner and much easier to read

endpoints/openrtb2/auction_test.go
```diff
@@ -1032,9 +1033,45 @@ func TestStoredRequests(t *testing.T) {
...
+			t.Error("processStoredRequests should return error")
 		}
+		err := errList[0]
+		assert.Equal(t, testStoredRequestsErrorsResults[i], err.Error(), "Incorrect error")
```
mansinahar, line 1074: You can rather use `assert.Contains` here so that you don't need to hardcode always looking at the first error in the slice. If the code changes in the future which may result in that error to move further down the slice then this would break even though the slice would still contain the expected error

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
mansinahar: I see. Makes sense. Thanks for that clarification. That variable name is deceiving :D

exchange/exchange.go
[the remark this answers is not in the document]
mansinahar: It does. Thanks for the clarification!

exchange/exchange.go
```diff
@@ -935,7 +943,7 @@ func (e *exchange) makeBid(bids []*pbsOrtbBid, auc *auction, returnCreative bool
 			}
 		}
 
-		if bidExtJSON, err := makeBidExtJSON(bid.bid.Ext, bidExtPrebid); err != nil {
+		if bidExtJSON, err := makeBidExtJSON(bid.bid.Ext, bidExtPrebid, impExtInfoMap[bid.bid.ImpID]); err != nil {
```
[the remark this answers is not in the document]
VeronikaSolovei9: I was thinking about this too, but it appears it works and when value is not found -this variable has a default structure with empty fields. 
If you think we need to add explicit check and create default variable - I'm open to do this. 
But looks like Golang does it already.

exchange/exchange.go
```diff
@@ -949,19 +957,29 @@ func (e *exchange) makeBid(bids []*pbsOrtbBid, auc *auction, returnCreative bool
...
-	}
+func makeBidExtJSON(ext json.RawMessage, prebid *openrtb_ext.ExtBidPrebid, impExtInfo ImpExtInfo) (json.RawMessage, error) {
 
-	// update existing bid.ext with our prebid section. if bid.ext.prebid already exists, it will be overwritten.
 	var extMap map[string]interface{}
-	if err := json.Unmarshal(ext, &extMap); err != nil {
-		return nil, err
+
+	if len(ext) != 0 {
+		// update existing bid.ext with our prebid section. if bid.ext.prebid already exists, it will be overwritten.
```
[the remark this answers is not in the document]
VeronikaSolovei9: This comment is still valid. 
`if bid.ext.prebid already exists, it will be overwritten` - this is what it does. 
EchoVideoAttrs will be inserted in `bid.ext` not `bid.ext.prebid` in further processing. 
I'm open to modify this comment to anything better, please post your suggestions.

exchange/exchange.go
```diff
@@ -949,19 +957,29 @@ func (e *exchange) makeBid(bids []*pbsOrtbBid, auc *auction, returnCreative bool
...
+		if err := json.Unmarshal(ext, &extMap); err != nil {
+			return nil, err
+		}
+	} else {
+		extMap = make(map[string]interface{})
 	}
+
+	if impExtInfo.EchoVideoAttrs {
+		videoData, _, _, err := jsonparser.Get(impExtInfo.StoredImp, "video")
+		if err != nil {
```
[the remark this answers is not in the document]
VeronikaSolovei9: If request is valid, there shouldn't be `KeyPathNotFoundError`. But I agree this case needs to be handled. Unit test added as well. Good catch!

exchange/exchange.go
```diff
@@ -27,11 +26,13 @@ import (
 	"github.com/prebid/prebid-server/metrics"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/prebid_cache_client"
+	"github.com/prebid/prebid-server/stored_requests"
 )
 
 type ContextKey string
 
 const DebugContextKey = ContextKey("debugInfo")
+const StoredRequestAttributes = "storedrequestattributes"
```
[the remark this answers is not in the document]
VeronikaSolovei9: I was also thinking about it too. And the thing is that `PrebidExtKey` constant is located in `openrtb_ext/device.go`. There are no bid.ext constants in bid.go. Not sure where this one should go.

exchange/exchange_test.go
```diff
@@ -3443,6 +3443,72 @@ func TestUpdateHbPbCatDur(t *testing.T) {
...
+			expectedErrMessage: "invalid character",
+		},
+		{
+			description:        "Valid extension, empty extBidPrebid and invalid imp ext info",
+			ext:                json.RawMessage(`{"video":{"h":100}}`),
+			extBidPrebid:       openrtb_ext.ExtBidPrebid{},
+			impExtInfo:         ImpExtInfo{true, []byte(`{"video":{!}}`)},
+			expectedBidExt:     ``,
+			expectedErrMessage: "invalid character",
+		},
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, added.

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: Modified it to `assert.Empty`, but personally `len(errList) == 0` is easier to read for me.

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: Yes, my bad I usually use `assert.Contains`.

exchange/exchange_test.go
```diff
@@ -1400,7 +1400,7 @@ func TestBidReturnsCreative(t *testing.T) {
 
 	//Run tests
 	for _, test := range testCases {
-		resultingBids, resultingErrs := e.makeBid(sampleBids, sampleAuction, test.inReturnCreative)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Added new unit test for this: TestMakeBidsVideoAttributes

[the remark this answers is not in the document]
mansinahar: I am not sure if we need `TestMakeBidsVideoAttributes` because all those test cases are already being tested in the `TestMakeBidExtJSON` test that you added. In `makeBids` we only pass `impExtInfo` downstream, we don't really have any logic addition/deletion/modification there. In such cases I'd just add one new test case for that upstream method (in this case `makeBids`) to make sure the value is being passed downstream as expected. Exhaustive test cases about the downstream method (in this case `makeBidExtJSON`) where the actual logic is changed makes sense only for the test for that method. This is why I recommended that we just add a new test case for `buildBidResponse` or modify and existing test case for `buildBidResponse` to just make sure that the value is properly passed all the way downstream to `makeBidExtJSON`. 

I hope this makes sense!

exchange/exchange.go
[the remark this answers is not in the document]
mansinahar: Sorry, I should have been more clearer. I agree with what you said above. What I meant by my comment was that maybe we should extend that comment to also add a note about the video attributes. Something like:
```
// update existing bid.ext with our prebid section and add stored video 
// attributes for imp if echoVideoAttrs set to true.
// if bid.ext.prebid already exists, it will be overwritten.
```
Also, perhaps this comment makes more sense to be on the first line of this function

exchange/exchange.go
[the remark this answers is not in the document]
mansinahar: Yeah, I was wondering the same about `PrebidExtKey` but `PrebidExtKey` is a constant key that is valid for different extension objects such as bid and imp. So we want to have that in a common place. That common place definitely shouldn't be `device.go`.. maybe an `openrtb_ext.go` file that can hold common stuff like this but that's for another PR some other time :D. For now, I think `StoredRequestAttributes` would make more sense in `openrtb_ext/bid.go` since that is specific to bid ext.

endpoints/openrtb2/auction_test.go
```diff
@@ -1032,9 +1033,45 @@ func TestStoredRequests(t *testing.T) {
...
+		hardcodedResponseIPValidator{response: true},
+	}
+
+	for i, requestData := range testStoredRequestsErrors {
+		_, _, errList := deps.processStoredRequests(context.Background(), json.RawMessage(requestData))
+		if len(errList) == 0 {
+			t.Error("processStoredRequests should return error")
 		}
+		err := errList[0]
+		assert.Contains(t, err.Error(), testStoredRequestsErrorsResults[i], "Incorrect error")
```
mansinahar, line 1074: You can just do: 
```
assert.Contains(t, errList, testStoredRequestsErrorsResults[i], "Incorrect error")
```
No need to create the `err` variable.

endpoints/openrtb2/auction_test.go
```diff
@@ -1020,9 +1019,11 @@ func TestStoredRequests(t *testing.T) {
 		hardcodedResponseIPValidator{response: true},
 	}
 
+	testStoreVideoAttr := []bool{true, true, false, false}
+
 	for i, requestData := range testStoredRequests {
-		newRequest, errList := deps.processStoredRequests(context.Background(), json.RawMessage(requestData))
-		if len(errList) != 0 {
+		newRequest, impExtInfoMap, errList := deps.processStoredRequests(context.Background(), json.RawMessage(requestData))
+		if assert.Empty(t, errList) {
```
mansinahar, line 1026: I think you by mistake added the `assert.Empty` in the wrong place. This should still be `if len(errList) != 0 {`. You probably wanna change to `assert.NotEmpty` on L1070

exchange/exchange.go
```diff
@@ -949,19 +957,34 @@ func (e *exchange) makeBid(bids []*pbsOrtbBid, auc *auction, returnCreative bool
...
+		if err := json.Unmarshal(ext, &extMap); err != nil {
+			return nil, err
+		}
+	} else {
+		extMap = make(map[string]interface{})
 	}
+
 	extMap[openrtb_ext.PrebidExtKey] = prebid
+
+	if impExtInfo, ok := impExtInfoMap[impId]; ok {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, modified

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: This is partial string comparison, not list contains element. I reduced it to 
`assert.Contains(t, errList[0].Error(), testStoredRequestsErrorsResults[i], "Incorrect error")`

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: oh ok, modified

VeronikaSolovei9:
Comments addressed. please take another look

endpoints/openrtb2/auction.go
```diff
@@ -135,7 +135,7 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		deps.analytics.LogAuctionObject(&ao)
 	}()
 
-	req, errL := deps.parseRequest(r)
+	req, impExtInfoMap, errL := deps.parseRequest(r)
```
SyntaxNode, line 138: Nitpick: Looking at the code a bit more with the new flow, how do you feel about naming this `impExtInfoByID` to provide context of the map key? I don't feel strongly about this if you feel differently.

endpoints/openrtb2/auction.go
SyntaxNode, line 1381: We can possibly avoid memory reallocation by making the map with the expected size:
```
make(map[string]exchange.ImpExtInfo, len(impIds))
```

endpoints/openrtb2/auction_test.go
```diff
@@ -1032,9 +1033,43 @@ func TestStoredRequests(t *testing.T) {
...
+		if len(impExtInfoMap[expectedImp].StoredImp) > 0 {
+			assert.JSONEq(t, string(expectedStoredImp), string(impExtInfoMap[expectedImp].StoredImp), "Incorrect expected stored imp %d", i)
+
 		}
+		assert.Equalf(t, testStoreVideoAttr[i], impExtInfoMap[expectedImp].EchoVideoAttrs, "EchoVideoAttrs value is incorrect")
+	}
+}
+
+func TestStoredRequestsVideoErrors(t *testing.T) {
+	deps := &endpointDeps{
```
SyntaxNode, line 1050: Could you please add a comment regarding which part of this setup is important for generating the error?

exchange/exchange_test.go
```diff
@@ -1685,12 +1685,71 @@ func TestBidResponseCurrency(t *testing.T) {
...
+			expectedBidResponseExt: `{"prebid":{"type":"video"},"storedrequestattributes":{"h":480,"mimes":["video/mp4"]}}`,
+		},
+	}
+	// Run tests
+	for i := range testCases {
+		actualBidResp, err := e.buildBidResponse(context.Background(), liveAdapters, adapterBids, bidRequest, nil, nil, nil, true, testCases[i].impExtInfo, errList)
+		resBidExt := string(actualBidResp.SeatBid[0].Bid[0].Ext)
+		assert.NoError(t, err, fmt.Sprintf("[TEST_FAILED] e.buildBidResponse resturns error in test: %s Error message: %s \n", testCases[i].description, err))
+		assert.Equalf(t, testCases[i].expectedBidResponseExt, resBidExt, fmt.Sprintf("[TEST_FAILED] Objects must be equal for test: %s \n Expected: >>%s<< \n Actual: >>%s<< ", testCases[i].description, resBidExt, actualBidResp.Ext))
+	}
```
SyntaxNode, line 1750: Since there is just one test case, please simplify by removing the `aTest` struct and loop, you can hardcode the one test.

exchange/exchange.go
```diff
@@ -949,19 +956,34 @@ func (e *exchange) makeBid(bids []*pbsOrtbBid, auc *auction, returnCreative bool
...
 	var extMap map[string]interface{}
-	if err := json.Unmarshal(ext, &extMap); err != nil {
-		return nil, err
+
+	if len(ext) != 0 {
+		if err := json.Unmarshal(ext, &extMap); err != nil {
+			return nil, err
+		}
+	} else {
+		extMap = make(map[string]interface{})
```
SyntaxNode, line 970: Please add test coverage for this condition (nil or empty ext) in TestMakeBidExtJSON.

exchange/exchange.go
```diff
@@ -949,19 +956,34 @@ func (e *exchange) makeBid(bids []*pbsOrtbBid, auc *auction, returnCreative bool
...
 	}
+
 	extMap[openrtb_ext.PrebidExtKey] = prebid
+
+	if impExtInfo, ok := impExtInfoMap[impId]; ok && impExtInfo.EchoVideoAttrs {
+
+		videoData, _, _, err := jsonparser.Get(impExtInfo.StoredImp, "video")
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
+			return nil, err
```
SyntaxNode, line 979: Please add test coverage for this condition in TestMakeBidExtJSON.

endpoints/openrtb2/auction.go
```diff
@@ -1389,22 +1392,38 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+		// This is substantially faster for reading values from a json blob than the Go json package,
+		// but keep in mind that each jsonparser.GetXXX call re-parses the entire json.
+		// Based on performance measurements, the tipping point of efficiency is around 4 calls.
+		// At that point, please consider switching to EachKey to use a single pass.
+		includeVideoAttributes, err := jsonparser.GetBoolean(resolvedImp, "ext", "prebid", "options", "echovideoattrs")
+		if err != nil && err != jsonparser.KeyPathNotFoundError {
+			return nil, nil, []error{err}
+		}
+		if storedImps[impIds[i]] != nil {
+			impExtInfoMap[impId] = exchange.ImpExtInfo{includeVideoAttributes, storedImps[impIds[i]]}
```
SyntaxNode, line 1412: Please use keyed fields:
```
exchange.ImpExtInfo{EchoVideoAttrs: includeVideoAttributes, StoredImp: storedImps[impIds[i]]}
```

endpoints/openrtb2/auction.go
VeronikaSolovei9: We agreed with this name with Mansi earlier. I personally like `impExtInfoMap` more, but I'm open to rename it. @mansinahar what do you think?

exchange/exchange.go
VeronikaSolovei9: I haven't found the way to do this. There are some error cases in `TestMakeBidExtJSON`, but yes, this specific line of code is not covered.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Done

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: Comment added

exchange/exchange_test.go
VeronikaSolovei9: Done

exchange/exchange.go
VeronikaSolovei9: Added

endpoints/openrtb2/auction.go
VeronikaSolovei9: Done

endpoints/openrtb2/auction.go
mansinahar: Both of those names sound good to me. I don't feel strongly about either. Will leave it up to you Veronika.

mansinahar (approved):
LGTM! Thanks for being patient with this one and addressing all the feedback @VeronikaSolovei9. Appreciate it :)

SyntaxNode (approved):

PR #1890, opened 2021-06-16
Fix a weak vendor enforcement bug where vendor does not exist

hhhjort:


gdpr/impl.go
```diff
@@ -128,14 +128,18 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, consen
...
-	parsedConsent, vendor, err := p.parseVendor(ctx, vendorID, consent)
+	parsedConsent, vendor, err := p.parseVendor(ctx, vendorID, consent, weakVendorEnforcement)
 	if err != nil {
 		return false, false, false, err
 	}
 
 	// vendor will be nil if not a valid TCF2 consent string
 	if vendor == nil {
-		return false, false, false, nil
+		if weakVendorEnforcement {
```
bsardo, line 138: I think this will now allow a TCF1 consent string to be processed if `weakVendorEnforcement` is set to true. `parseVendor` returns with a nil vendor and a nil error when the consent string is considered a well-formed TCF1 or TCF2 string with a version != 2. That caused it to enter this `if` block which would return `false, false, false, nil`. See `TestTCF1Consent`.

Perhaps `parseVendor` should return an error when `version != 2`. Then a valid TCF1 consent string would be handled by the err check on line 132 and you could keep your logic changes here. This change would also mean a version != 2 would be handled the same way as a malformed consent string. Thoughts?

gdpr/impl.go
```diff
@@ -241,3 +246,22 @@ func (a AlwaysAllow) BidderSyncAllowed(ctx context.Context, bidder openrtb_ext.B
 func (a AlwaysAllow) AuctionActivitiesAllowed(ctx context.Context, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal Signal, consent string, weakVendorEnforcement bool) (allowBidRequest bool, passGeo bool, passID bool, err error) {
 	return true, true, true, nil
 }
+
+// Mock Vendor claims everything.
```
SyntaxNode, line 250: I wouldn't call this a mock, that terminology is usually reserved for tests. Consider something like: `// vendorTrue claims everything`

gdpr/impl.go
```diff
@@ -194,7 +198,7 @@ func (p *permissionsImpl) checkPurpose(consent tcf2.ConsentMetadata, vendor api.
 	return purposeAllowed || legitInterest
 }
 
-func (p *permissionsImpl) parseVendor(ctx context.Context, vendorID uint16, consent string) (parsedConsent api.VendorConsents, vendor api.Vendor, err error) {
+func (p *permissionsImpl) parseVendor(ctx context.Context, vendorID uint16, consent string, weakVendorEnforcement bool) (parsedConsent api.VendorConsents, vendor api.Vendor, err error) {
```
bsardo, line 201: This new parameter doesn't appear to be used anywhere.

gdpr/impl.go
hhhjort: Good idea, but moved the verification slightly.

gdpr/impl.go
```diff
@@ -135,7 +135,11 @@ func (p *permissionsImpl) allowActivities(ctx context.Context, vendorID uint16,
 
 	// vendor will be nil if not a valid TCF2 consent string
 	if vendor == nil {
-		return false, false, false, nil
+		if weakVendorEnforcement && parsedConsent.Version() == 2 {
```
SyntaxNode, line 138: Could we fix this by changing `parseVendor` to return an error when the version is incompatible? Otherwise, this is fine for now and @bsardo could keep it in mind for refactoring.

hhhjort: I went with this method to provide the least possible change in behavior. But we could start pushing back a bad TCF version when this happens. Making the change would also mean pushing `weakVendorEnforcement` back down a level again.

bsardo: I'm good with this change for now. I can make a note on the refactor story to return an error when the version is incompatible.

SyntaxNode (approved):

bsardo (approved):

PR #1894, opened 2021-06-17
Friendlier Startup Error Messages

SyntaxNode:
Small change to omit the stack trace when startup errors are found. The stack trace just points to `main.go`, and is therefore useless. `glog.Exitf` exists immediately with a status code of 1. Previously the status code was 255.

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,12 +34,12 @@ func main() {
 
 	cfg, err := loadConfig()
 	if err != nil {
-		glog.Fatalf("Configuration could not be loaded or did not pass validation: %v", err)
+		glog.Exitf("Configuration could not be loaded or did not pass validation: %v", err)
 	}
 
 	err = serve(Rev, cfg)
 	if err != nil {
-		glog.Errorf("prebid-server failed: %v", err)
+		glog.Exitf("prebid-server failed: %v", err)
 	}
 }
 


hhhjort (approved):
LGTM

main.go
```diff
@@ -34,12 +34,12 @@ func main() {
 
 	cfg, err := loadConfig()
 	if err != nil {
-		glog.Fatalf("Configuration could not be loaded or did not pass validation: %v", err)
+		glog.Exitf("Configuration could not be loaded or did not pass validation: %v", err)
```
guscarreon, line 37: I understand switching `Errorf` to `Exitf` in line 42 but I don't quite understand the reason behind the change from `Fatalf` to `Exitf` in line 37. The error code "255" usually means that it is not able to find a file the system needs to execute. Isn't it more descriptive to use `Fatalf` instead of `Exitf` in line 37?
```
1144 // Fatalf logs to the FATAL, ERROR, WARNING, and INFO logs,
1145 // including a stack trace of all running goroutines, then calls os.Exit(255).
1146 // Arguments are handled in the manner of fmt.Printf; a newline is appended if missing.
1147 func Fatalf(format string, args ...interface{}) {
1148     logging.printf(fatalLog, format, args...)
1149 }
1150 +-- 24 lines: fatalNoStacks is non-zero if we are to exit without dumping goroutine stacks.--
1174
1175 // Exitf logs to the FATAL, ERROR, WARNING, and INFO logs, then calls os.Exit(1).
1176 // Arguments are handled in the manner of fmt.Printf; a newline is appended if missing.
1177 func Exitf(format string, args ...interface{}) {
1178     atomic.StoreUint32(&fatalNoStacks, 1)
1179     logging.printf(fatalLog, format, args...)
1180 }
~/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go [RO]
```

SyntaxNode: `Exitf` will return exit code 1, or general error. I can't find a definition of 255 to mean file not found. Where did you see that correlation?

guscarreon: You are right. I got it wrong, according to [this bash shell doc](https://tldp.org/LDP/abs/html/exitcodes.html) it is regarded as `Exit status out of range`. An exit status of `1`, should do.

guscarreon (approved):
LGTM

PR #1895, opened 2021-06-18
Outbrain adapter: overwrite tagid only if it exists

rokostik:
Hey,

this PR changes the adapter so that the the imp tagid is overwritten with the ext tagid only if it exists.

### DIFF
--- a/adapters/outbrain/outbrain.go
+++ b/adapters/outbrain/outbrain.go
@@ -43,8 +43,10 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 			errs = append(errs, err)
 			continue
 		}
-		imp.TagID = outbrainExt.TagId
-		reqCopy.Imp[i] = imp
+		if outbrainExt.TagId != "" {
+			imp.TagID = outbrainExt.TagId
+			reqCopy.Imp[i] = imp
+		}
 	}
 
 	publisher := &openrtb2.Publisher{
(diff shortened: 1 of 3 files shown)


adapters/outbrain/outbrain.go
```diff
@@ -43,8 +43,10 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 			errs = append(errs, err)
 			continue
 		}
-		imp.TagID = outbrainExt.TagId
-		reqCopy.Imp[i] = imp
+		if outbrainExt.TagId != "" {
```
SyntaxNode, line 46: Please add a new json test for this condition, to be sure the tests cover both tagid = "" and tagid = "<something else>"

hhhjort (approved):
LGTM

adapters/outbrain/outbraintest/supplemental/optional_params.json
```diff
@@ -26,6 +27,8 @@
         }
       }
     ],
+    "bcat": ["should-be-overwritten"],
```
mansinahar, line 30: Not sure why are `bcat` and `badv` values are same as `tagid` in this test.. If the intention was to just assign these some string values and have nothing to do with `tagid`, I'd suggest replacing it with some other string values so that it's not confusing with the `tagid` having the same value

mansinahar (approved):

guscarreon (approved):
LGTM

Issue #1482, opened 2020-09-08
Currency conversion utility function


bszekely1:
## Type of issue

The overall objective of Prebid is provide global attributes for those that are common across bid adapters, examples being bidfloor, video attributes, etc. This ticket is specific to support for bid floor. When the currency of the bid floor is different than the currency of the bid adapter supports, there needs to be a method bid adapters can call to convert the currency.

## Goals

- Provide a method for PBS bid adapters to convert auction currency of bid floor object

## Proposed Design

If the currency conversion feature is enabled in a host PBS instance, Bid Adapters need a utility function to convert currencies of different types. Below are the technical requirements:

1. As an input, the utility function sold take the impression ID and the desired currency
    - The currency conversion function should look up the floor associated with that impression object and convert as needed
2. If the currency feature is not enabled in a PBS instance, the currency conversion utility function should not be available
    - It is the job of the Bid Adapter to check if the utility function exists 
3.If the input currency in the utility function is the same as the set currency, no conversion is necessary
4. if the input currency in the utility function is different, conversion should occur, rounding the converted currency up after a precision of 4 decimal points
5. The response object should include both the converted floor and currency if conversion was applied
    - If desired currency is not available in the currency feature, return the original floor and currency
6. Response should be a JSON object

### Interfaces

Request
```
convertBidFloor(imp.id, currency)
```

Response
```
{
 "currency": "USD",
 "bidFloor":  0.3344
}
```

bretg:
We discussed in last week's Prebid Server committee meeting and have a simplified proposal:

provide a simple utility function:

```
   pbsConvertCurrency(fromCurrency, toCurrency, value)
```

bid adapters can make use of this function as needed. e.g. if they only support floors in USD, they can convert non-USD imp[].bidfloor values.

BraslavskiyAndrey:
Hi @bretg, @bszekely1
I need one clarification. PBS has two currency sources, external currency service and currencies from bidrequest.prebid.currencies. If we want to use both, like for bid price currency coversion, utility function will look like:

`   pbsConvertCurrency(value, bidRequest, fromCurrency, toCurrency)
`
  or if BidAdapter will be responsible for retrieving bidrequest.prebid.currency, utility function will be the same as for bid:

    ` pbsConvertCurrency(value, requestCurrencies, boolean prioritizeRequestRates, fromCurrency, toCurrency)`

[...]

PR #1901, opened 2021-06-24
Currency Conversion Utility Function

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/1482

### DIFF
--- a/adapters/bidder.go
+++ b/adapters/bidder.go
@@ -150,7 +150,7 @@ func NewExtraRequestInfo(c currency.Conversions) ExtraRequestInfo {
 
 // ConvertCurrency converts a given amount from one currency to another, or returns:
 //  - Error if the `from` or `to` arguments are malformed or unknown ISO-4217 codes.
-//  - ConversionRateNotFoundError if the conversion mapping is unknown to Prebid Server
+//  - ConversionNotFoundError if the conversion mapping is unknown to Prebid Server
 //    and not provided in the bid request.
 func (r ExtraRequestInfo) ConvertCurrency(value float64, from, to string) (float64, error) {
 	if rate, err := r.currencyConversions.GetRate(from, to); err == nil {
--- a/adapters/bidder.go
+++ b/adapters/bidder.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/mxmCherry/openrtb/v15/openrtb2"
 	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/currency"
 	"github.com/prebid/prebid-server/metrics"
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
@@ -138,6 +139,25 @@ func (r *RequestData) SetBasicAuth(username string, password string) {
 type ExtraRequestInfo struct {
 	PbsEntryPoint              metrics.RequestType
 	GlobalPrivacyControlHeader string
+	currencyConversions        currency.Conversions
+}
+
+func NewExtraRequestInfo(c currency.Conversions) ExtraRequestInfo {
+	return ExtraRequestInfo{
+		currencyConversions: c,
+	}
+}
+
+// ConvertCurrency converts a given amount from one currency to another, or returns:
+//  - Error if the `from` or `to` arguments are malformed or unknown ISO-4217 codes.
+//  - ConversionRateNotFoundError if the conversion mapping is unknown to Prebid Server
+//    and not provided in the bid request.
+func (r ExtraRequestInfo) ConvertCurrency(value float64, from, to string) (float64, error) {
+	if rate, err := r.currencyConversions.GetRate(from, to); err == nil {
+		return value * rate, nil
+	} else {
+		return 0, err
+	}
 }
 
 type Builder func(openrtb_ext.BidderName, config.Adapter) (Bidder, error)
--- a/adapters/bidder_test.go
+++ b/adapters/bidder_test.go
@@ -1,0 +1,66 @@
+package adapters
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestExtraRequestInfoConvertCurrency(t *testing.T) {
+	var (
+		givenValue float64 = 2
+		givenFrom  string  = "AAA"
+		givenTo    string  = "BBB"
+	)
+
+	testCases := []struct {
+		description   string
+		setMock       func(m *mock.Mock)
+		expectedValue float64
+		expectedError string
+	}{
+		{
+			description:   "Success",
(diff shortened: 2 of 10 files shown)


adapters/bidder.go
```diff
@@ -138,6 +139,25 @@ func (r *RequestData) SetBasicAuth(username string, password string) {
...
+
+func NewExtraRequestInfo(c currency.Conversions) ExtraRequestInfo {
+	return ExtraRequestInfo{
+		currencyConversions: c,
+	}
+}
+
+// ConvertCurrency converts a given amount from one currency to another, or returns:
+//  - Error if the `from` or `to` arguments are malformed or unknown ISO-4217 codes.
+//  - ConversionRateNotFoundError if the conversion mapping is unknown to Prebid Server
```
SyntaxNode, line 153: @guscarreon What do you think is simplifying this error name to `ConversionNotFoundError`?

currency/rates.go
```diff
@@ -70,7 +70,7 @@ func (r *Rates) GetRate(from string, to string) (float64, error) {
 			// In case we have an entry TO -> FROM
 			return 1 / conversion, nil
 		}
-		return 0, ConversionRateNotFound{fromUnit.String(), toUnit.String()}
+		return 0, ConversionRateNotFoundError{FromCur: fromUnit.String(), ToCur: toUnit.String()}
```
SyntaxNode, line 73: Resolved GoLint flags.

adapters/bidder_test.go
```diff
@@ -0,0 +1,66 @@
...
+		result, err := extraRequestInfo.ConvertCurrency(givenValue, givenFrom, givenTo)
+
+		mockConversions.AssertExpectations(t)
+		if test.expectedError == "" {
+			assert.NoError(t, err, test.description+":err")
+			assert.Equal(t, test.expectedValue, result, test.description+":result")
+		} else {
+			assert.Error(t, err, test.description+":err")
+			assert.Empty(t, result, test.description+":result")
+		}
```
guscarreon, line 50: For an easier read, we could probably remove the `if-else` clause altogether for two `assert.Equal()` calls instead.
```
44       testCases := []struct {
45           description   string
46           setMock       func(m *mock.Mock)
47           expectedValue float64
48 -         expectedError string
   +         expectedError error
49       }{
50           {
44               description:   "Success",
45               setMock:       func(m *mock.Mock) { m.On("GetRate", "AAA", "BBB").Return(2.5, nil) },
46               expectedValue: 5,
47           },
48           {
49               description:   "Error",
50               setMock:       func(m *mock.Mock) { m.On("GetRate", "AAA", "BBB").Return(2.5, errors.New("some error")) },
47 -             expectedError: "some error",
   +             expectedError: errors.New("The actual error message"),
48           },
49       }
                .
                .
                .
             result, err := extraRequestInfo.ConvertCurrency(givenValue, givenFrom, givenTo)

             mockConversions.AssertExpectations(t)
44 -         if test.expectedError == "" {
45 -             assert.NoError(t, err, test.description+":err")
46 -             assert.Equal(t, test.expectedValue, result, test.description+":result")
47 -         } else {
48 -             assert.Error(t, err, test.description+":err")
49 -             assert.Empty(t, result, test.description+":result")
50 -         }
   +             assert.Equal(t, test.expectedValue, result, test.description+":result")
   +             assert.Equal(t, test.expectedError, err, test.description+":result")
51       }
52   }
adapters/bidder_test.go
```

adapters/bidder.go
```diff
@@ -138,6 +139,25 @@ func (r *RequestData) SetBasicAuth(username string, password string) {
 type ExtraRequestInfo struct {
 	PbsEntryPoint              metrics.RequestType
 	GlobalPrivacyControlHeader string
+	currencyConversions        currency.Conversions
```
guscarreon, line 142: I almost forgot about the `ExtraRequestInfo` struct. This is a great use for it.

exchange/exchange_test.go
```diff
@@ -539,7 +540,7 @@ func TestTwoBiddersDebugDisabledAndEnabled(t *testing.T) {
 
 func TestOverrideWithCustomCurrency(t *testing.T) {
 
-	mockCurrencyClient := &mockCurrencyRatesClient{
+	mockCurrencyClient := &fakeCurrencyRatesHttpClient{
```
guscarreon, line 543: I thought we favored the `mock` term over the `fake` term. I agree with `fake`

adapters/bidder.go
SyntaxNode: Agreed. I view this as a more broad "stuff from PBS-Core" object. I suggest we consider renaming it in the future, but not in this PR.

exchange/exchange_test.go
SyntaxNode: [Depends on the context.](https://stackoverflow.com/questions/346372/whats-the-difference-between-faking-mocking-and-stubbing#:~:text=Mock%20%2D%20an%20object%20on%20which,people%20simply%20call%20them%20mocks.)

I personally prefer to use mock when testify mock's framework is used with test defined behavior/expectations and fake when the object does very little and is somewhat configurable by the test, such as with the request body in this instance.

exchange/exchange_test.go
```diff
@@ -694,6 +695,72 @@ func TestOverrideWithCustomCurrency(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, "some-request-id", response.ID, "Response ID")
+	assert.Empty(t, response.SeatBid, "Response Bids")
+	assert.Contains(t, string(response.Ext), `"errors":{"foo":[{"code":5,"message":"The adapter failed to generate any bid requests, but also failed to generate an error explaining why"}]}`, "Response Ext")
+
+	// Test Currency Converter Properly Passed To Adapter
+	if assert.NotNil(t, mockBidder.lastExtraRequestInfo, "Currency Conversion Argument") {
+		converted, err := mockBidder.lastExtraRequestInfo.ConvertCurrency(2.0, "USD", "MXN")
+		assert.NoError(t, err, "Currency Conversion Error")
+		assert.Equal(t, 40.0, converted, "Currency Conversion Response")
```
guscarreon, line 760: I don't mind asserting this element but, do you believe it'd be better to mock the adapter bidder so it writes the converted price and the actual currency to the `SeatBid` instead?  Such an assertion could be done instead of the `assert.Empty` in line 753 above

SyntaxNode: That would certainly work, but it would increase the length / complexity of this test since the mockBidder would now need to build a proper seat bid response whereas now the call to MakeBids is skipped. I think we're ok since this proves the currency convert object passed into the adapter funcs is correct.

guscarreon (approved):
LGTM. I'm curious to see an adapter use this new functionality.

bsardo (approved):
LGTM

PR #1905, opened 2021-06-28
New Adapter: HuaweiAds

wwwyyy:
add new adapter: huaweiads 
document PR: https://github.com/prebid/prebid.github.io/pull/3071

### DIFF
--- a/adapters/huaweiads/huaweiads.go
+++ b/adapters/huaweiads/huaweiads.go
@@ -1,0 +1,1007 @@
+package huaweiads
+
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	nativeRequests "github.com/mxmCherry/openrtb/v15/native1/request"
+	nativeResponse "github.com/mxmCherry/openrtb/v15/native1/response"
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const HuaweiAdxApiVersion string = "3.4"
+const DefaultBidRequestArrayLength int = 1
+const DefaultCountryName = "ZA"
+const DefaultUnknownNetworkType = 0
+const TimeFormat = "2006-01-02 15:04:05.000"
+const DefaultModelName = "HUAWEI"
+
+// creative type
+const (
+	text        int32 = 1
+	bigPicture  int32 = 2
+	bigPicture2 int32 = 3
+	gif         int32 = 4
+	// used for magazine lock screen
+	noSpecificCreativeType int32 = 5
+	videoText              int32 = 6
+	smallPicture           int32 = 7
+	threeSmallPicturesText int32 = 8
+	video                  int32 = 9
+	iconText               int32 = 10
+	// RewardedVideo only
+	videoWithPicturesText int32 = 11
+	windowAdvertisement   int32 = 13
+)
+
+// ads events
+const (
+	click         string = "click"
+	imp           string = "imp"
+	userClose     string = "userclose"
+	download      string = "download"
+	install       string = "install"
+	downloadstart string = "downloadstart"
+	playStart     string = "playStart"
+	playEnd       string = "playEnd"
+	playResume    string = "playResume"
+	playPause     string = "playPause"
+	appOpen       string = "appOpen"
+)
+
+// ads type
+const (
+	banner        int32 = 8
+	native        int32 = 3
+	rewardedVideo int32 = 7
+	splash        int32 = 1
+	interstitial  int32 = 12
+	roll          int32 = 60
+)
+
+type HuaweiAdsAdapter struct {
+	endpoint string
+}
+
+func (a *HuaweiAdsAdapter) MakeRequests(openRTBRequest *openrtb2.BidRequest,
+	reqInfo *adapters.ExtraRequestInfo) (requestsToBidder []*adapters.RequestData, errs []error) {
+	if openRTBRequest == nil {
+		errs = append(errs, &errortypes.BadInput{
+			Message: "MakeRequests: openRTBRequest is nil",
+		})
+		return nil, errs
+	}
+
+	numRequests := len(openRTBRequest.Imp)
+	if numRequests == 0 || openRTBRequest.Imp == nil {
+		errs = append(errs, &errortypes.BadInput{
+			Message: "MakeRequests: No impression in the bid request",
+		})
+		return nil, errs
+	}
+
+	var request HuaweiAdsRequest
+	var header http.Header
+	var multislot = make([]Adslot30, 0, numRequests)
+
+	var huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds
+	var err1 error
+	for index := 0; index < numRequests; index++ {
+		huaweiAdsImpExt, err1 = unmarshalExtImpHuaweiAds(&openRTBRequest.Imp[index])
+
+		if err1 != nil {
+			errs = append(errs, err1)
+			return nil, errs
+		}
+
+		adslot30, err := getHuaweiAdsReqAdslot30(huaweiAdsImpExt, &openRTBRequest.Imp[index], openRTBRequest)
+		if err != nil {
+			errs = append(errs, err)
+			return nil, errs
+		}
+
+		multislot = append(multislot, adslot30)
+	}
+	request.Multislot = multislot
+
+	if err := getHuaweiAdsReqJson(&request, openRTBRequest, huaweiAdsImpExt); err != nil {
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	reqJSON, err := json.Marshal(request)
+	if err != nil {
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	//	our request header's Authorization is changing by time, cannot verify by a certain string,
+	//	use isAddAuthorization = false only when run testcase,
+	//	isAddAuthorization = true when we request for a new ad from Huawei Ads ADX.
+	var isAddAuthorization = true
+	if huaweiAdsImpExt.IsAddAuthorization == "false" {
+		isAddAuthorization = false
+	}
+	header = getHeaders(huaweiAdsImpExt, openRTBRequest, isAddAuthorization)
+	bidRequest := &adapters.RequestData{
+		Method:  http.MethodPost,
+		Uri:     a.endpoint,
+		Body:    reqJSON,
+		Headers: header,
+	}
+
+	var bidRequestArray = make([]*adapters.RequestData, 0, DefaultBidRequestArrayLength)
+	bidRequestArray = append(bidRequestArray, bidRequest)
+
+	return bidRequestArray, errs
+}
+
+func (a *HuaweiAdsAdapter) MakeBids(openRTBRequest *openrtb2.BidRequest, requestToBidder *adapters.RequestData,
+	bidderRawResponse *adapters.ResponseData) (bidderResponse *adapters.BidderResponse, errs []error) {
+	if bidderRawResponse.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	httpStatusError := checkRespStatusCode(bidderRawResponse)
+	if httpStatusError != nil {
+		return nil, []error{httpStatusError}
+	}
+
+	var huaweiAdsResponse HuaweiAdsResponse
+	if err := json.Unmarshal(bidderRawResponse.Body, &huaweiAdsResponse); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: "Bad Server Response",
+		}}
+	}
+
+	huaweiAdsResponseError := checkHuaweiAdsResponseRetcode(huaweiAdsResponse)
+	if huaweiAdsResponseError != nil {
+		return nil, []error{huaweiAdsResponseError}
+	}
+
+	var err error
+	bidderResponse, err = convertHuaweiAdsResp2BidderResp(&huaweiAdsResponse, openRTBRequest)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	return bidderResponse, nil
+}
+
+// Builder builds a new instance of the HuaweiAds adapter for the given bidder with the given config.
+func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
+	bidder := &HuaweiAdsAdapter{
+		endpoint: config.Endpoint,
+	}
+	return bidder, nil
+}
+
+// get request header
+func getHeaders(huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds, request *openrtb2.BidRequest, isAddAuthorization bool) http.Header {
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	if isAddAuthorization {
+		headers.Add("Authorization", getDigestAuthorization(huaweiAdsImpExt))
+	}
+
+	if request.Device != nil && len(request.Device.UA) > 0 {
+		headers.Add("User-Agent", request.Device.UA)
+	}
+
+	return headers
+}
+
+// get body json for HuaweiAds request
+func getHuaweiAdsReqJson(request *HuaweiAdsRequest, openRTBRequest *openrtb2.BidRequest, huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds) error {
+	request.Version = HuaweiAdxApiVersion
+	var err error
+	if err = getHuaweiAdsReqAppInfo(request, openRTBRequest); err != nil {
+		return err
+	}
+	if err = getHuaweiAdsReqDeviceInfo(request, openRTBRequest, huaweiAdsImpExt); err != nil {
+		return err
+	}
+	getHuaweiAdsReqNetWorkInfo(request, openRTBRequest)
+	getHuaweiAdsReqRegsInfo(request, openRTBRequest)
+	getHuaweiAdsReqGeoInfo(request, openRTBRequest)
+	return nil
+}
+
+// get adslot30
+func getHuaweiAdsReqAdslot30(huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds,
+	openRTBImp *openrtb2.Imp, openRTBRequest *openrtb2.BidRequest) (Adslot30, error) {
+	adtypeLower := strings.ToLower(huaweiAdsImpExt.Adtype)
+	var adslot30 = Adslot30{
+		Slotid: huaweiAdsImpExt.SlotId,
+		Adtype: convertAdtypeString2Integer(adtypeLower),
+		Test:   int32(openRTBRequest.Test),
+	}
+
+	if adtypeLower == "roll" {
+		if openRTBImp.Video != nil {
+			adslot30.TotalDuration = int32(openRTBImp.Video.MaxDuration)
+		} else {
+			return adslot30, errors.New("GetHuaweiAdsReqAdslot30: MaxDuration is empty when adtype is roll.")
+		}
+	}
+	return adslot30, nil
+}
+
+// convert adtype String -> Integer
+func convertAdtypeString2Integer(adtypeLower string) int32 {
+	if adtypeLower == "banner" {
+		return 8
+	} else if adtypeLower == "native" {
+		return 3
+	} else if adtypeLower == "rewarded" {
+		return 7
+	} else if adtypeLower == "splash" {
+		return 1
+	} else if adtypeLower == "interstitial" {
+		return 12
+	} else if adtypeLower == "roll" {
+		return 60
+	} else {
+		return 8
+	}
+}
+
+// get app information for HuaweiAds request
+func getHuaweiAdsReqAppInfo(request *HuaweiAdsRequest, openRTBRequest *openrtb2.BidRequest) error {
+	var app App
+	if openRTBRequest.App != nil {
+		if openRTBRequest.App.Ver != "" {
+			app.Version = openRTBRequest.App.Ver
+		}
+		if openRTBRequest.App.Name != "" {
+			app.Name = openRTBRequest.App.Name
+		}
+
+		// bundle cannot be empty, we need package name.
+		if openRTBRequest.App.Bundle != "" {
+			app.Pkgname = openRTBRequest.App.Bundle
+		} else {
+			return errors.New("HuaweiAdsReqApp: Pkgname is empty.")
+		}
+
+		if openRTBRequest.App.Content != nil && openRTBRequest.App.Content.Language != "" {
+			app.Lang = openRTBRequest.App.Content.Language
+		} else {
+			app.Lang = "en"
+		}
+	}
+
+	if openRTBRequest.User != nil && openRTBRequest.User.Geo != nil && openRTBRequest.User.Geo.Country != "" {
+		app.Country = openRTBRequest.User.Geo.Country
+	} else if openRTBRequest.Device != nil && openRTBRequest.Device.Geo != nil && openRTBRequest.Device.Geo.Country != "" {
+		app.Country = openRTBRequest.Device.Geo.Country
+	} else {
+		app.Country = DefaultCountryName
+	}
+
+	request.App = app
+	return nil
+}
+
+// get field clientTime, format: 2006-01-02 15:04:05.000+2000
+func getClientTime(clientTime string) (newClientTime string) {
+	zone, _ := time.Now().Local().Zone()
+	if isMatched, _ := regexp.MatchString("[+-]{1}\\d{2}", zone); isMatched {
+		zone = zone + "00"
+	} else {
+		zone = "+0200"
+	}
+
+	if clientTime == "" {
+		return time.Now().Format(TimeFormat) + zone
+	}
+
+	if isMatched1, _ := regexp.MatchString("^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}.\\d{3}[+-]{1}\\d{4}$", clientTime); isMatched1 {
+		return clientTime
+	}
+	if isMatched2, _ := regexp.MatchString("^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}.\\d{3}$", clientTime); isMatched2 {
+		return clientTime + "+0200"
+	}
+	return time.Now().Format(TimeFormat) + zone
+}
+
+// get device information for HuaweiAds request
+func getHuaweiAdsReqDeviceInfo(request *HuaweiAdsRequest, openRTBRequest *openrtb2.BidRequest, huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds) (err error) {
+	var device Device
+	if openRTBRequest.Device != nil {
+		device.Type = int32(openRTBRequest.Device.DeviceType)
+		device.Useragent = openRTBRequest.Device.UA
+		device.Os = openRTBRequest.Device.OS
+		device.Version = openRTBRequest.Device.OSV
+		device.Maker = openRTBRequest.Device.Make
+		device.Model = openRTBRequest.Device.Model
+		if device.Model == "" {
+			device.Model = DefaultModelName
+		}
+		device.Height = int32(openRTBRequest.Device.H)
+		device.Width = int32(openRTBRequest.Device.W)
+		device.Language = openRTBRequest.Device.Language
+		device.Pxratio = float32(openRTBRequest.Device.PxRatio)
+		device.ClientTime = getClientTime(huaweiAdsImpExt.ClientTime)
+
+		// oaid  IsTrackingEnabled = 1 - DNT
+		if device.Oaid != "" && openRTBRequest.Device.DNT != nil {
+			device.IsTrackingEnabled = strconv.Itoa(1 - int(*openRTBRequest.Device.DNT))
+		}
+		if device.Gaid != "" && openRTBRequest.Device.DNT != nil {
+			device.GaidTrackingEnabled = strconv.Itoa(1 - int(*openRTBRequest.Device.DNT))
+		}
+
+		if openRTBRequest.User != nil && openRTBRequest.User.Geo != nil && openRTBRequest.User.Geo.Country != "" {
+			device.BelongCountry = openRTBRequest.User.Geo.Country
+			device.LocaleCountry = openRTBRequest.User.Geo.Country
+		} else if openRTBRequest.Device != nil && openRTBRequest.Device.Geo != nil && openRTBRequest.Device.Geo.Country != "" {
+			device.BelongCountry = openRTBRequest.Device.Geo.Country
+			device.LocaleCountry = openRTBRequest.Device.Geo.Country
+		} else {
+			device.BelongCountry = DefaultCountryName
+			device.LocaleCountry = DefaultCountryName
+		}
+		device.Ip = openRTBRequest.Device.IP
+	}
+
+	// get oaid gaid imei in openRTBRequest.User.Ext.Data
+	if err = getDeviceID(&device, openRTBRequest); err != nil {
+		return err
+	}
+
+	request.Device = device
+	return nil
+}
+
+// get device id  include oaid gaid imei.
+func getDeviceID(device *Device, openRTBRequest *openrtb2.BidRequest) (err error) {
+	if openRTBRequest.User == nil {
+		return errors.New("getDeviceID: openRTBRequest.User is nil.")
+	}
+	if openRTBRequest.User.Ext == nil {
+		return errors.New("getDeviceID: openRTBRequest.User.Ext is nil.")
+	}
+	var extUserDataHuaweiAds openrtb_ext.ExtUserDataHuaweiAds
+	if err := json.Unmarshal(openRTBRequest.User.Ext, &extUserDataHuaweiAds); err != nil {
+		return errors.New("Unmarshal: openRTBRequest.User.Ext -> extUserDataHuaweiAds failed")
+	}
+	var deviceId = extUserDataHuaweiAds.Data
+	if len(deviceId.Imei) == 0 && len(deviceId.Gaid) == 0 && len(deviceId.Oaid) == 0 {
+		return errors.New("getDeviceID: Imei ,Oaid, Gaid are all empty.")
+	}
+	if len(deviceId.Oaid) > 0 {
+		device.Oaid = deviceId.Oaid[0]
+	}
+	if len(deviceId.Gaid) > 0 {
+		device.Gaid = deviceId.Gaid[0]
+	}
+	if len(deviceId.Imei) > 0 {
+		device.Imei = deviceId.Imei[0]
+	}
+	return nil
+}
+
+// get network information for HuaweiAds request
+func getHuaweiAdsReqNetWorkInfo(request *HuaweiAdsRequest, openRTBRequest *openrtb2.BidRequest) {
+	var network Network
+	if openRTBRequest.Device != nil && openRTBRequest.Device.ConnectionType != nil {
+		network.Type = int32(*openRTBRequest.Device.ConnectionType)
+	} else {
+		network.Type = DefaultUnknownNetworkType
+	}
+	request.Network = network
+}
+
+// get regs information for HuaweiAds request
+func getHuaweiAdsReqRegsInfo(request *HuaweiAdsRequest, openRTBRequest *openrtb2.BidRequest) {
+	var regs Regs
+	if openRTBRequest.Regs != nil {
+		regs.Coppa = int32(openRTBRequest.Regs.COPPA)
+	}
+	request.Regs = regs
+}
+
+// get geo information for HuaweiAds request
+func getHuaweiAdsReqGeoInfo(request *HuaweiAdsRequest, openRTBRequest *openrtb2.BidRequest) {
+	var geo Geo
+	if openRTBRequest.Device != nil && openRTBRequest.Device.Geo != nil {
+		geo.Lon = float32(openRTBRequest.Device.Geo.Lon)
+		geo.Lat = float32(openRTBRequest.Device.Geo.Lat)
+		geo.Accuracy = int32(openRTBRequest.Device.Geo.Accuracy)
+		geo.Lastfix = int32(openRTBRequest.Device.Geo.LastFix)
+	}
+	request.Geo = geo
+}
+
+func unmarshalExtImpHuaweiAds(openRTBImp *openrtb2.Imp) (*openrtb_ext.ExtImpHuaweiAds, error) {
+	var bidderExt adapters.ExtImpBidder
+	var huaweiAdsImpExt openrtb_ext.ExtImpHuaweiAds
+	if err := json.Unmarshal(openRTBImp.Ext, &bidderExt); err != nil {
+		return nil, errors.New("Unmarshal: openRTBImp.Ext -> bidderExt failed")
+	}
+	if err := json.Unmarshal(bidderExt.Bidder, &huaweiAdsImpExt); err != nil {
+		return nil, errors.New("Unmarshal: bidderExt.Bidder -> huaweiAdsImpExt failed")
+	}
+	if huaweiAdsImpExt.SlotId == "" {
+		return nil, errors.New("ExtImpHuaweiAds: slotid is empty.")
+	}
+	if huaweiAdsImpExt.Adtype == "" {
+		return nil, errors.New("ExtImpHuaweiAds: adtype is empty.")
+	}
+	if huaweiAdsImpExt.PublisherId == "" {
+		return nil, errors.New("ExtHuaweiAds: publisherid is empty.")
+	}
+	if huaweiAdsImpExt.SignKey == "" {
+		return nil, errors.New("ExtHuaweiAds: signkey is empty.")
+	}
+	if huaweiAdsImpExt.KeyId == "" {
+		return nil, errors.New("ExtImpHuaweiAds: keyid is empty.")
+	}
+	return &huaweiAdsImpExt, nil
+}
+
+// check response status code
+func checkRespStatusCode(response *adapters.ResponseData) error {
+	if response.StatusCode == http.StatusNoContent {
+		return nil
+	}
+
+	if response.StatusCode == http.StatusBadRequest {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: [ %d ]", response.StatusCode),
+		}
+	}
+
+	if response.StatusCode == http.StatusServiceUnavailable {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Something went wrong, please contact your Account Manager. Status Code: [ %d ] ", response.StatusCode),
+		}
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: [ %d ]. Run with request.debug = 1 for more info", response.StatusCode),
+		}
+	}
+
+	if response.Body == nil {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("bidderRawResponse body is empty"),
+		}
+	}
+	return nil
+}
+
+// check HuaweiAds response retcode
+func checkHuaweiAdsResponseRetcode(response HuaweiAdsResponse) error {
+	if response.Retcode == 200 || response.Retcode == 204 || response.Retcode == 206 {
+		return nil
+	}
+
+	if (response.Retcode < 600 && response.Retcode >= 400) || (response.Retcode < 300 && response.Retcode > 200) {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("HuaweiAdsResponse retcode: %d , reason: %s", response.Retcode, response.Reason),
+		}
+	}
+	return nil
+}
+
+// convert HuaweiAds' response into bidder's response
+func convertHuaweiAdsResp2BidderResp(huaweiAdsResponse *HuaweiAdsResponse, openRTBRequest *openrtb2.BidRequest) (bidderResponse *adapters.BidderResponse, err error) {
+	if len(huaweiAdsResponse.Multiad) == 0 {
+		return nil, errors.New("convertHuaweiAdsResp2BidderResp: multiad length is 0, get no ads from huawei side.")
+	}
+	bidderResponse = adapters.NewBidderResponseWithBidsCapacity(len(huaweiAdsResponse.Multiad))
+	// Default Currency: CNY
+	bidderResponse.Currency = "CNY"
+
+	// record request Imp (slotid->imp, slotid->openrtb_ext.bidtype)
+	mapSlotid2Imp := make(map[string]openrtb2.Imp, len(openRTBRequest.Imp))
+	mapSlotid2MediaType := make(map[string]openrtb_ext.BidType, len(openRTBRequest.Imp))
+	for _, imp := range openRTBRequest.Imp {
+		huaweiAdsExt, err := unmarshalExtImpHuaweiAds(&imp)
+		if err != nil {
+			continue
+		}
+		mapSlotid2Imp[huaweiAdsExt.SlotId] = imp
+
+		var mediaType = openrtb_ext.BidTypeBanner
+		if imp.Video != nil {
+			mediaType = openrtb_ext.BidTypeVideo
+		} else if imp.Native != nil {
+			mediaType = openrtb_ext.BidTypeNative
+		} else if imp.Audio != nil {
+			mediaType = openrtb_ext.BidTypeAudio
+		}
+		mapSlotid2MediaType[huaweiAdsExt.SlotId] = mediaType
+	}
+
+	if len(mapSlotid2MediaType) < 1 || len(mapSlotid2Imp) < 1 {
+		return nil, errors.New("convertHuaweiAdsResp2BidderResp: openRTBRequest.imp is nil")
+	}
+	if huaweiAdsResponse.Multiad == nil {
+		return nil, errors.New("convertHuaweiAdsResp2BidderResp: huaweiAdsResponse.Multiad is nil")
+	}
+
+	for _, ad30 := range huaweiAdsResponse.Multiad {
+		if mapSlotid2Imp[ad30.Slotid].ID == "" {
+			continue
+		}
+
+		if ad30.Retcode30 != 200 {
+			continue
+		}
+
+		if ad30.Content == nil {
+			continue
+		}
+
+		for _, content := range ad30.Content {
+			var bid openrtb2.Bid
+			bid.ID = mapSlotid2Imp[ad30.Slotid].ID
+			bid.ImpID = mapSlotid2Imp[ad30.Slotid].ID
+			// The bidder has already helped us automatically convert the currency price, here only the CNY price is filled in
+			bid.Price = content.Price
+			// Advertising creative ID, used for logging and behavior tracking
+			bid.CrID = content.Contentid
+			// All currencies should be the same
+			if content.Cur != "" {
+				bidderResponse.Currency = content.Cur
+			}
+
+			bid.AdM, bid.W, bid.H, err = getAdmFromHuaweiAdsContent(ad30.AdType, &content, mapSlotid2MediaType[ad30.Slotid], mapSlotid2Imp[ad30.Slotid])
+			if err != nil {
+				return nil, err
+			}
+
+			bid.ADomain = append(bid.ADomain, "huaweiads")
+			// TODO
+			bid.NURL = ""
+			bid.LURL = ""
+			bid.IURL = ""
+			bid.BURL = ""
+
+			bidderResponse.Bids = append(bidderResponse.Bids, &adapters.TypedBid{
+				Bid:     &bid,
+				BidType: mapSlotid2MediaType[ad30.Slotid],
+			})
+		}
+	}
+	return bidderResponse, nil
+}
+
+// get field Adm
+func getAdmFromHuaweiAdsContent(adType int32, content *Content, bidType openrtb_ext.BidType, imp openrtb2.Imp) (adm string, adWidth int64, adHeight int64, err error) {
+	var creativeType = content.Creativetype
+	if content.Creativetype > 100 {
+		creativeType = creativeType - 100
+	}
+
+	if bidType == openrtb_ext.BidTypeVideo {
+		if creativeType == videoText || creativeType == video || creativeType == videoWithPicturesText {
+			adm, adWidth, adHeight, err = extractAdmVideo(adType, content, bidType, imp)
+		} else {
+			return "", 0, 0, fmt.Errorf("huaweiads response dose not have video")
+		}
+	} else if bidType == openrtb_ext.BidTypeNative {
+		if adType == native {
+			adm, adWidth, adHeight, err = extractAdmNative(adType, content, bidType, imp)
+		} else {
+			return "", 0, 0, fmt.Errorf("huaweiads response is not a native ad")
+		}
+	} else if bidType == openrtb_ext.BidTypeBanner {
+		if creativeType == text || creativeType == bigPicture || creativeType == bigPicture2 ||
+			creativeType == smallPicture || creativeType == threeSmallPicturesText ||
+			creativeType == iconText || creativeType == gif {
+			adm, adWidth, adHeight, err = extractAdmPicture(content)
+		} else if creativeType == videoText || creativeType == video || creativeType == videoWithPicturesText {
+			adm, adWidth, adHeight, err = extractAdmVideo(adType, content, bidType, imp)
+		} else {
+			return "", 0, 0, fmt.Errorf("no support creativetype: " + strconv.Itoa(int(creativeType)))
+		}
+	} else if bidType == openrtb_ext.BidTypeAudio {
+		return "", 0, 0, fmt.Errorf("no support bidtype: audio")
+	}
+
+	if err != nil {
+		return "", 0, 0, fmt.Errorf("getAdmFromHuaweiAdsContent failed: %s", err)
+	}
+	return adm, adWidth, adHeight, nil
+}
+
+// For native ad
+func extractAdmNative(adType int32, content *Content, bidType openrtb_ext.BidType, openrtb2Imp openrtb2.Imp) (adm string,
+	adWidth int64, adHeight int64, err error) {
+	if openrtb2Imp.Native == nil {
+		return "", 0, 0, fmt.Errorf("extractAdmNative: imp.Native is nil")
+	}
+
+	if openrtb2Imp.Native.Request == "" {
+		return "", 0, 0, fmt.Errorf("extractAdmNative: imp.Native.Request is empty")
+	}
+
+	var nativePayload nativeRequests.Request
+	if err := json.Unmarshal(json.RawMessage(openrtb2Imp.Native.Request), &nativePayload); err != nil {
+		return "", 0, 0, err
+	}
+
+	var nativeResult nativeResponse.Response
+	var linkObject nativeResponse.Link
+	linkObject.URL = content.MetaData.ClickUrl
+	if content.MetaData.ClickUrl == "" {
+		return "", 0, 0, fmt.Errorf("extractAdmNative: content.MetaData.ClickUrl is empty")
+	}
+
+	nativeResult.Assets = make([]nativeResponse.Asset, 0, len(nativePayload.Assets))
+	var ImgIndex = 0
+	for _, asset := range nativePayload.Assets {
+		var responseAsset nativeResponse.Asset
+		if asset.Title != nil {
+			var titleObject nativeResponse.Title
+			titleObject.Text = ""
+			if content.MetaData.Title != "" {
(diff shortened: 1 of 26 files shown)


SyntaxNode:
Thank you for your submission. We'll review this PR soon. This is a larger adapter than usual, so we ask for your patience with the review process.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+const (
+	banner        int32 = 8
+	native        int32 = 3
+	rewardedVideo int32 = 7
+	splash        int32 = 1
+	interstitial  int32 = 12
+	roll          int32 = 60
+)
+
+type HuaweiAdsAdapter struct {
```
hhhjort, line 77: This can just be `type adapter struct {`. The package identifies it as `huaweiads` adapter, and all adapters are referenced as interfaces. The `Builder()` function gives access to an instance of your adapter specifically.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+	}
+
+	numRequests := len(openRTBRequest.Imp)
+	if numRequests == 0 || openRTBRequest.Imp == nil {
+		errs = append(errs, &errortypes.BadInput{
+			Message: "MakeRequests: No impression in the bid request",
+		})
+		return nil, errs
+	}
+
```
hhhjort, line 97: Just a note, the upstream code already confirms that there is a non-zero number of impressions, so these two checks are unnecessary.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+			adslot30.TotalDuration = int32(openRTBImp.Video.MaxDuration)
+		} else {
+			return adslot30, errors.New("GetHuaweiAdsReqAdslot30: MaxDuration is empty when adtype is roll.")
+		}
+	}
+	return adslot30, nil
+}
+
+// convert adtype String -> Integer
+func convertAdtypeString2Integer(adtypeLower string) int32 {
```
hhhjort, line 247: You have defined these numbers above as constants with corresponding names. Why are you not using those constants here?

adapters/huaweiads/huaweiads_req_resp.go
```diff
@@ -0,0 +1,199 @@
+package huaweiads
+
+type HuaweiAdsRequest struct {
```
hhhjort, line 3: Why are these exported? other files in the same package can reference unexported objects.

hhhjort:
The distinction of the tests in directory `huaweiadstest/exemplary` is that they should provide a good example for publishers to follow. Additional success tests can be placed in `huaweiadstest/supplemental` to test valid variations that are not meant as a preferred example. Are all the tests you have in `exemplary` intended to be used by publishers as your preferred request examples?

wwwyyy:
I will check and correct it, thank you for correcting my mistake

adapters/huaweiads/huaweiads.go
wwwyyy: Corrected thanks

adapters/huaweiads/huaweiads.go
wwwyyy: adtype is integer, but the user parameter adtype is string, I can only transfer it

adapters/huaweiads/huaweiads_req_resp.go
wwwyyy: Deleted, sorry, I often write like this when writing  java

adapters/huaweiads/huaweiads.go
wwwyyy: Deleted

adapters/huaweiads/huaweiads.go
hhhjort: I don't think you understood my intent. Just below, you could do:
```
	if adtypeLower == "banner" {
		return banner
	} else if adtypeLower == "native" {
		return native
...
```
This makes the code a little cleaner I think, and insures you are at least consistent with the numerical value for "banner" and all.

adapters/huaweiads/huaweiads_req_resp.go
hhhjort: I see you moved these all over to the main `huaweiads.go` file, but you are continuing to export them. I think it was a bit more organized to have them over in this file, I was suggesting that you move them. `huaweiads.go` is long enough as it is. My point was you could have gone with `type huaweiAdsRequest struct {` so that Go would know that the struct is only intended to be referenced within this package.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1395 @@
...
+	var request HuaweiAdsRequest
+	var header http.Header
+	var multislot = make([]Adslot30, 0, numRequests)
+
+	var huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds
+	var err1 error
+	for index := 0; index < numRequests && numRequests > 0; index++ {
+		huaweiAdsImpExt, err1 = unmarshalExtImpHuaweiAds(&openRTBRequest.Imp[index])
+		if err1 != nil || huaweiAdsImpExt == nil {
+			errs = append(errs, err1)
```
hhhjort, line 293: You probably should set `err1` to be non nil in the case that `huaweiAdsImpExt` is nil.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1395 @@
...
+	if err = getHuaweiAdsReqDeviceInfo(request, openRTBRequest, huaweiAdsImpExt); err != nil {
+		return err
+	}
+	getHuaweiAdsReqNetWorkInfo(request, openRTBRequest)
+	getHuaweiAdsReqRegsInfo(request, openRTBRequest)
+	getHuaweiAdsReqGeoInfo(request, openRTBRequest)
+	return nil
+}
+
+// get adslot30
```
hhhjort, line 437: You should be able to drop this comment, or in the alternative describe what this function is supposed to do.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1395 @@
...
+	} else if openRTBImp.Native != nil {
+		if err := getNativeFormat(&adslot30, openRTBImp); err != nil {
+			return adslot30, err
+		}
+	}
+
+	if adtypeLower == "roll" {
+		if openRTBImp.Video != nil {
+			adslot30.TotalDuration = int32(openRTBImp.Video.MaxDuration)
+		} else {
```
hhhjort, line 470: Do you want to verify that `MaxDuration` is a positive number?

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1395 @@
...
+func unmarshalExtImpHuaweiAds(openRTBImp *openrtb2.Imp) (*openrtb_ext.ExtImpHuaweiAds, error) {
+	var bidderExt adapters.ExtImpBidder
+	var huaweiAdsImpExt openrtb_ext.ExtImpHuaweiAds
+	if err := json.Unmarshal(openRTBImp.Ext, &bidderExt); err != nil {
+		return nil, errors.New("Unmarshal: openRTBImp.Ext -> bidderExt failed")
+	}
+	if err := json.Unmarshal(bidderExt.Bidder, &huaweiAdsImpExt); err != nil {
+		return nil, errors.New("Unmarshal: bidderExt.Bidder -> huaweiAdsImpExt failed")
+	}
+	if huaweiAdsImpExt.SlotId == "" {
```
hhhjort, line 746: Since these are required in your JSON schema, the entire prebid request will fail with an error (no bidders will bid) if any of these parameters are missing. I have no problem with verifying this again here, but thought you should know.

adapters/huaweiads/huaweiads.go
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
wwwyyy: In huaweiads document, roll ad need TotalDuration , but now only support banner native,  will support the roll ad later .

adapters/huaweiads/huaweiads.go
wwwyyy: yes, I know, all of them are required parameters ，and I confirm.

adapters/huaweiads/huaweiads.go
wwwyyy: corrected

adapters/huaweiads/huaweiads_req_resp.go
wwwyyy: already corrected ， thanks

hhhjort (approved):
LGTM

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	} else {
+		if len(str) == 1 {
+			result = "00" + str
+		} else if len(str) == 2 {
+			result = "0" + str
+		} else {
+			result = str
+		}
+	}
+	return result
```
SyntaxNode, line 1148: Since the if/else structure is exhaustive, you can make return from each line, ex:
```
if isMmm == false {
	if len(str) < 2 {
		return "0" + str + dot
	} else {
		return str + dot
	}
} else {
	if len(str) == 1 {
		return "00" + str
	} else if len(str) == 2 {
		return "0" + str
	} else {
		return str
	}
}
```

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	}
+
+	var adTitle, _ = getDecodeValue(content.MetaData.Title)
+	var adId = content.Contentid
+	var creativeId = content.Contentid
+	var trackingEvents strings.Builder
+	var dspImpTracking2Str = ""
+	var dspClickTracking2Str = ""
+	var errorTracking2Str = ""
+	if content.Monitor != nil {
```
SyntaxNode, line 1208: I don't believe the nil check is necessary. The range loop gracefully handles nil slices.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		"height=\"" + strconv.Itoa(int(adHeight)) + "\" scalable=\"true\" maintainAspectRatio=\"true\"> " +
+		"<![CDATA[" + resourceUrl + "]]>" +
+		"</MediaFile>" +
+		"</MediaFiles>" +
+		"</Linear>" +
+		"</Creative>" +
+		"</Creatives>" +
+		"</InLine>" +
+		"</Ad>" +
+		"</VAST>"
```
SyntaxNode, line 1273: Consider using back tick literal strings here to avoid escaping the double quote.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+func getVastImpClickErrorTrackingUrls(urls []string, eventType string) (result string) {
+	var trackingUrls strings.Builder
+	for _, url := range urls {
+		if eventType == "click" {
+			trackingUrls.WriteString("<ClickTracking><![CDATA[" + url + "]]></ClickTracking>")
+		} else if eventType == "imp" {
+			trackingUrls.WriteString("<Impression><![CDATA[" + url + "]]></Impression>")
+		} else if eventType == "vastError" {
+			trackingUrls.WriteString("<Error><![CDATA[" + url + "&et=[ERRORCODE]]]></Error>")
+		}
```
SyntaxNode, line 1286: To make the most use of the string builder, please break up the string concatenation into separate WriteString calls. Same for getVastEventTrackingUrls.

adapters/huaweiads/huaweiads_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"testing"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderHuaweiAds, config.Adapter{
+		Endpoint: "https://acd.op.hicloud.com/ppsadx/getResult"})
```
SyntaxNode, line 12: Consider using an obviously fake endpoint for the purposes of tests. This doesn't need to be the real one here.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	} else if openRTBImp.Native != nil {
+		if err := getNativeFormat(&adslot30, openRTBImp); err != nil {
+			return adslot30, err
+		}
+	}
+
+	// Currently does not support roll type ads, roll ad need TotalDuration
+	if adtypeLower == "roll" {
+		if openRTBImp.Video != nil && openRTBImp.Video.MaxDuration >= 0 {
```
SyntaxNode, line 404: If you accept video ads, please add that to `huaweiads.yaml `.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		if openRTBRequest.App.Bundle != "" {
+			app.Pkgname = openRTBRequest.App.Bundle
+		} else {
+			return errors.New("HuaweiAdsReqApp: Pkgname is empty.")
+		}
+
+		if openRTBRequest.App.Content != nil && openRTBRequest.App.Content.Language != "" {
+			app.Lang = openRTBRequest.App.Content.Language
+		} else {
+			app.Lang = "en"
```
SyntaxNode, line 500: Please add a json test for this case.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		return nil, errors.New("ExtHuaweiAds: signkey is empty.")
+	}
+	if huaweiAdsImpExt.KeyId == "" {
+		return nil, errors.New("ExtImpHuaweiAds: keyid is empty.")
+	}
+	return &huaweiAdsImpExt, nil
+}
+
+// check response status code
+func checkRespStatusCode(response *adapters.ResponseData) error {
```
SyntaxNode, line 711: Please add a json test file for each condition in the `checkRespStatusCode` method.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+			var formats = make([]format, 0, len(openRTBImp.Banner.Format))
+			for _, f := range openRTBImp.Banner.Format {
+				if f.H != 0 && f.W != 0 {
+					formats = append(formats, format{f.W, f.H})
+				}
+			}
+			adslot30.Format = formats
+		}
+	} else if openRTBImp.Native != nil {
+		if err := getNativeFormat(&adslot30, openRTBImp); err != nil {
```
SyntaxNode, line 397: Please add a native test case which doesn't return an error. There doesn't seem to be one currently.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		} else {
+			return adslot30, errors.New("GetHuaweiAdsReqAdslot30: MaxDuration is empty when adtype is roll.")
+		}
+	}
+	return adslot30, nil
+}
+
+func getNativeFormat(adslot30 *adslot30, openRTBImp *openrtb2.Imp) error {
+	if openRTBImp.Native.Request == "" {
+		return errors.New("extractAdmNative: imp.Native.Request is empty")
```
SyntaxNode, line 415: Please add a json test file for this case.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if index > 0 && len(t)-index == 5 {
+		zone = t[index:]
+	}
+	if clientTime == "" {
+		return time.Now().Format(TimeFormat) + zone
+	}
+	if isMatched, _ := regexp.MatchString("^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}.\\d{3}[+-]{1}\\d{4}$", clientTime); isMatched {
+		return clientTime
+	}
+	if isMatched, _ := regexp.MatchString("^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}.\\d{3}$", clientTime); isMatched {
```
SyntaxNode, line 522: Please add a json test case for this second regex.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		request.Network = network
+	}
+}
+
+// get regs information for HuaweiAds request
+func getHuaweiAdsReqRegsInfo(request *huaweiAdsRequest, openRTBRequest *openrtb2.BidRequest) {
+	if openRTBRequest.Regs != nil && openRTBRequest.Regs.COPPA >= 0 {
+		var regs regs
+		regs.Coppa = int32(openRTBRequest.Regs.COPPA)
+		request.Regs = regs
```
SyntaxNode, line 667: Please add test coverage for when coppa is specified.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	return nil
+}
+
+// check HuaweiAds response retcode
+func checkHuaweiAdsResponseRetcode(response huaweiAdsResponse) error {
+	if response.Retcode == 200 || response.Retcode == 204 || response.Retcode == 206 {
+		return nil
+	}
+
+	if (response.Retcode < 600 && response.Retcode >= 400) || (response.Retcode < 300 && response.Retcode > 200) {
```
SyntaxNode, line 748: Please add test coverage for this condition.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	}
+
+	for _, ad30 := range huaweiAdsResponse.Multiad {
+		if mapSlotid2Imp[ad30.Slotid].ID == "" {
+			continue
+		}
+
+		if ad30.Retcode30 != 200 {
+			continue
+		}
```
SyntaxNode, line 800: Please add tests for the `mapSlotid2Imp[ad30.Slotid].ID == ""` and `ad30.Retcode30 != 200` conditions.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+func (a *adapter) handleHuaweiAdsContent(adType int32, content *content, bidType openrtb_ext.BidType, imp openrtb2.Imp) (
+	adm string, adWidth int64, adHeight int64, err error) {
+	// v1: only support banner, native
+	switch bidType {
+	case openrtb_ext.BidTypeBanner:
+		adm, adWidth, adHeight, err = a.extractAdmBanner(adType, content, bidType, imp)
+	case openrtb_ext.BidTypeNative:
+		adm, adWidth, adHeight, err = a.extractAdmNative(adType, content, bidType, imp)
+	case openrtb_ext.BidTypeVideo:
+		adm, adWidth, adHeight, err = a.extractAdmVideo(adType, content, bidType, imp)
```
SyntaxNode, line 856: Please add a video json test file.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	adWidth int64, adHeight int64, err error) {
+	if adType != native {
+		return "", 0, 0, fmt.Errorf("extractAdmNative: response is not a native ad")
+	}
+	if openrtb2Imp.Native == nil {
+		return "", 0, 0, fmt.Errorf("extractAdmNative: imp.Native is nil")
+	}
+	if openrtb2Imp.Native.Request == "" {
+		return "", 0, 0, fmt.Errorf("extractAdmNative: imp.Native.Request is empty")
+	}
```
SyntaxNode, line 899: Please add test coverage for these three if statements.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+			if videoObject.VASTTag, adWidth, adHeight, err = a.extractAdmVideo(adType, content, bidType, openrtb2Imp); err != nil {
+				return "", 0, 0, err
+			}
+			responseAsset.Video = &videoObject
+		} else if asset.Img != nil {
+			var imgObject nativeResponse.Image
+			imgObject.URL = ""
+			imgObject.Type = asset.Img.Type
+			if asset.Img.Type == native1.ImageAssetTypeIcon {
+				if len(content.MetaData.Icon) > iconIndex {
```
SyntaxNode, line 936: Please add test coverage for this condition.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if content.MetaData.ClickUrl != "" {
+		clickUrl = content.MetaData.ClickUrl
+	} else if content.MetaData.Intent != "" {
+		clickUrl, _ = getDecodeValue(content.MetaData.Intent)
+	}
+
+	var mime = "video/mp4"
+	var resourceUrl = ""
+	var duration = ""
+	if adType == roll {
```
SyntaxNode, line 1168: Please add test coverage for `adType == roll`.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		if content.MetaData.VideoInfo.VideoDownloadUrl != "" {
+			resourceUrl = content.MetaData.VideoInfo.VideoDownloadUrl
+		} else {
+			return "", 0, 0, errors.New("extractAdmVideo: content.MetaData.VideoInfo.VideoDownloadUrl is empty")
+		}
+		if content.MetaData.VideoInfo.Width != 0 && content.MetaData.VideoInfo.Height != 0 {
+			adWidth = int64(content.MetaData.VideoInfo.Width)
+			adHeight = int64(content.MetaData.VideoInfo.Height)
+		} else if bidType == openrtb_ext.BidTypeVideo {
+			if opentrb2Imp.Video != nil && opentrb2Imp.Video.W != 0 && opentrb2Imp.Video.H != 0 {
```
SyntaxNode, line 1191: Please add test coverage for `bidType == openrtb_ext.BidTypeVideo`.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+
+// compute HmacSha256
+func computeHmacSha256(message string, signKey string) string {
+	h := hmac.New(sha256.New, []byte(signKey))
+	h.Write([]byte(message))
+	return hex.EncodeToString(h.Sum(nil))
+}
+
+// get digest authorization for request header, huawei ads ppsadx API
+func getDigestAuthorization(huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds) string {
```
SyntaxNode, line 1312: Please add test coverage for computeHmacSha256 and getDigestAuthorization to ensure proper operation.

SyntaxNode (commented):
Thank you for patience through the review process. This is a very large PR due to the custom Huawei bidding format . That is perfectly fine, but is a more involved review due to the amount of code.

mansinahar:
@wwwyyy Just a friend reminder on the above comments

wwwyyy:
Don't worry, I will upload the code which has corrected the issues soon when my team test completed. Just this two days, thank you.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+	"strings"
+	"time"
+)
+
+const HuaweiAdxApiVersion string = "3.4"
+const DefaultBidRequestArrayLength int = 1
+const DefaultCountryName = "ZA"
+const DefaultUnknownNetworkType = 0
+const TimeFormat = "2006-01-02 15:04:05.000"
+const DefaultModelName = "HUAWEI"
```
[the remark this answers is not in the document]
wwwyyy: corrected， thank you

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+		})
+		return nil, errs
+	}
+
+	var request HuaweiAdsRequest
+	var header http.Header
+	var multislot = make([]Adslot30, 0, numRequests)
+
+	var huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds
+	var err1 error
```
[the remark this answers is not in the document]
wwwyyy: yeah，err1 is unused at this scope， but huaweiAdsImpExt will be used at this scope，so I only move err1 into this scope

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+	for index := 0; index < numRequests; index++ {
+		huaweiAdsImpExt, err1 = unmarshalExtImpHuaweiAds(&openRTBRequest.Imp[index])
+
+		if err1 != nil {
+			errs = append(errs, err1)
+			return nil, errs
```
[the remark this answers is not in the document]
wwwyyy: all corrected，thanks

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+	reqJSON, err := json.Marshal(request)
+	if err != nil {
+		errs = append(errs, err)
+		return nil, errs
+	}
+
+	//	our request header's Authorization is changing by time, cannot verify by a certain string,
+	//	use isAddAuthorization = false only when run testcase,
+	//	isAddAuthorization = true when we request for a new ad from Huawei Ads ADX.
+	var isAddAuthorization = true
```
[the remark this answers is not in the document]
wwwyyy: The function  getDigestAuthorization  here has been tested by our testers, and there is no problem with the access. Since the output value changes according to time, I have to exclude it from the unit test.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	var header http.Header
+	var multislot = make([]adslot30, 0, numRequests)
+
+	var huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds
+	var err1 error
+	for index := 0; index < numRequests && numRequests > 0; index++ {
+		huaweiAdsImpExt, err1 = unmarshalExtImpHuaweiAds(&openRTBRequest.Imp[index])
+		if err1 != nil {
+			errs = append(errs, err1)
+			return nil, errs
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+		Method:  http.MethodPost,
+		Uri:     a.endpoint,
+		Body:    reqJSON,
+		Headers: header,
+	}
+
+	var bidRequestArray = make([]*adapters.RequestData, 0, DefaultBidRequestArrayLength)
+	bidRequestArray = append(bidRequestArray, bidRequest)
+
+	return bidRequestArray, errs
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1028 @@
...
+}
+
+// get request header
+func getHeaders(huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds, request *openrtb2.BidRequest, isAddAuthorization bool) http.Header {
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+	if isAddAuthorization {
+		headers.Add("Authorization", getDigestAuthorization(huaweiAdsImpExt))
+	}
```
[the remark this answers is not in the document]
wwwyyy: yes, isAddAuthorization = false only when run testcase, because our request header's Authorization is changing by time, cannot verify by a certain string; It is always true in actual use

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	"net/url"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const HuaweiAdxApiVersion string = "3.4"
+const DefaultBidRequestArrayLength int = 1
+const DefaultCountryName = "ZA"
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	}
+	if huaweiAdsImpExt.KeyId == "" {
+		return nil, errors.New("ExtImpHuaweiAds: keyid is empty.")
+	}
+	return &huaweiAdsImpExt, nil
+}
+
+// check response status code
+func checkRespStatusCode(response *adapters.ResponseData) error {
+	if response.StatusCode == http.StatusNoContent {
```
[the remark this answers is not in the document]
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		return nil
+	}
+
+	if response.StatusCode == http.StatusBadRequest {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: [ %d ]", response.StatusCode),
+		}
+	}
```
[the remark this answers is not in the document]
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+
+	if response.StatusCode != http.StatusOK {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: [ %d ]. Run with request.debug = 1 for more info", response.StatusCode),
+		}
+	}
+
+	if response.Body == nil {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("bidderRawResponse body is empty"),
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+
+	httpStatusError := checkRespStatusCode(bidderRawResponse)
+	if httpStatusError != nil {
+		return nil, []error{httpStatusError}
+	}
+
+	var huaweiAdsResponse huaweiAdsResponse
+	if err := json.Unmarshal(bidderRawResponse.Body, &huaweiAdsResponse); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: "Bad Server Response",
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		}}
+	}
+
+	huaweiAdsResponseError := checkHuaweiAdsResponseRetcode(huaweiAdsResponse)
+	if huaweiAdsResponseError != nil {
+		return nil, []error{huaweiAdsResponseError}
+	}
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+
+	var err error
+	bidderResponse, err = a.convertHuaweiAdsResp2BidderResp(&huaweiAdsResponse, openRTBRequest)
+	if err != nil {
+		return nil, []error{err}
+	}
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		return rewarded
+	} else if adtypeLower == "splash" {
+		return splash
+	} else if adtypeLower == "interstitial" {
+		return interstitial
+	} else if adtypeLower == "roll" {
+		return roll
+	} else {
+		return banner
+	}
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	app.Country = getCountryCode(openRTBRequest)
+	request.App = app
+	return nil
+}
+
+// get field clientTime, format: 2006-01-02 15:04:05.000+0200
+func getClientTime(clientTime string) (newClientTime string) {
+	var zone = DefaultTimeZone
+	t := time.Now().Local().Format(time.RFC822Z)
```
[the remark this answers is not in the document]
wwwyyy: "ext": {
        "data": {
          "gaid": [
            "gaid"
          ],
          "oaid": [
            "oaid"
          ],
          "clientTime": [
            "2018-11-02 16:34:07.981+1300"
          ]
        }
      }
already tested in the test case，It’s best to use the time in this format from the media itself. If not, consider using the server time.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	index := strings.IndexAny(t, "-+")
+	if index > 0 && len(t)-index == 5 {
+		zone = t[index:]
+	}
+	if clientTime == "" {
+		return time.Now().Format(TimeFormat) + zone
+	}
+	if isMatched, _ := regexp.MatchString("^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}.\\d{3}[+-]{1}\\d{4}$", clientTime); isMatched {
```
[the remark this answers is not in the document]
wwwyyy: corrected， thank you

adapters/huaweiads/huaweiads.go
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		device.Useragent = openRTBRequest.Device.UA
+		device.Os = openRTBRequest.Device.OS
+		device.Version = openRTBRequest.Device.OSV
+		device.Maker = openRTBRequest.Device.Make
+		device.Model = openRTBRequest.Device.Model
+		if device.Model == "" {
+			device.Model = DefaultModelName
+		}
+		device.Height = int32(openRTBRequest.Device.H)
+		device.Width = int32(openRTBRequest.Device.W)
```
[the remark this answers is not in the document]
wwwyyy: yeah，thank you for reminding, there should be no too large numbers

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		return DefaultCountryName
+	}
+}
+
+//ISO 3166-1 Alpha3 -> Alpha2, Some countries may use
+func convertCountryCode(country string) (out string) {
+	if country == "" {
+		return DefaultCountryName
+	}
+	var MapCountryCodeAlpha3ToAlpha2 = map[string]string{
```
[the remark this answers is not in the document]
wwwyyy: yeah，I tried it, but GO seems to be unable to create a static map

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		"CHL": "CL", "CHN": "CN", "ARE": "AE"}
+	if MapCountryCodeAlpha3ToAlpha2[country] != "" {
+		return MapCountryCodeAlpha3ToAlpha2[country]
+	} else if len(country) >= 3 {
+		return country[0:2]
+	} else {
+		return DefaultCountryName
+	}
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if decodeValue, err = url.QueryUnescape(str); err == nil {
+		return decodeValue, nil
+	} else {
+		return "", err
+	}
+}
+
+// JSON custom
+func JSON(nativeResult nativeResponse.Response) ([]byte, error) {
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if MapCountryCodeAlpha3ToAlpha2[country] != "" {
+		return MapCountryCodeAlpha3ToAlpha2[country]
+	} else if len(country) >= 3 {
+		return country[0:2]
+	} else {
+		return DefaultCountryName
+	}
+}
+
+// get device id  include oaid gaid imei.
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("bidderRawResponse body is empty"),
+		}
+	}
+	return nil
+}
+
+// check HuaweiAds response retcode
+func checkHuaweiAdsResponseRetcode(response huaweiAdsResponse) error {
+	if response.Retcode == 200 || response.Retcode == 204 || response.Retcode == 206 {
```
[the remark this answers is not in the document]
wwwyyy: no，this is a field name Retcode ，206 is its value.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		mapSlotid2Imp[huaweiAdsExt.SlotId] = imp
+
+		var mediaType = openrtb_ext.BidTypeBanner
+		if imp.Video != nil {
+			mediaType = openrtb_ext.BidTypeVideo
+		} else if imp.Native != nil {
+			mediaType = openrtb_ext.BidTypeNative
+		} else if imp.Audio != nil {
+			mediaType = openrtb_ext.BidTypeAudio
+		}
```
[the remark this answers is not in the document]
wwwyyy: bid response don't have,

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+			continue
+		}
+
+		if ad30.Retcode30 != 200 {
+			continue
+		}
+
+		if ad30.Content == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+
+		for _, content := range ad30.Content {
+			var bid openrtb2.Bid
+			bid.ID = mapSlotid2Imp[ad30.Slotid].ID
+			bid.ImpID = mapSlotid2Imp[ad30.Slotid].ID
+			// The bidder has already helped us automatically convert the currency price, here only the CNY price is filled in
+			bid.Price = content.Price
+			bid.CrID = content.Contentid
+			// All currencies should be the same
```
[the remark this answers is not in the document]
wwwyyy: we make sure they are not empty

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+				Bid:     &bid,
+				BidType: mapSlotid2MediaType[ad30.Slotid],
+			})
+		}
+	}
+	return bidderResponse, nil
+}
+
+// get nurl when win auction.
+func getNurl(content *content) string {
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		return "", 0, 0, fmt.Errorf("getAdmFromHuaweiAdsContent failed: %s", err)
+	}
+	return adm, adWidth, adHeight, nil
+}
+
+// banner ad
+func (a *adapter) extractAdmBanner(adType int32, content *content, bidType openrtb_ext.BidType, imp openrtb2.Imp) (adm string,
+	adWidth int64, adHeight int64, err error) {
+	if adType != banner {
+		return "", 0, 0, fmt.Errorf("extractAdmBanner: huaweiads response is not a banner ad")
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if err := json.Unmarshal(json.RawMessage(openrtb2Imp.Native.Request), &nativePayload); err != nil {
+		return "", 0, 0, err
+	}
+
+	var nativeResult nativeResponse.Response
+	var linkObject nativeResponse.Link
+	if content.MetaData.ClickUrl != "" {
+		linkObject.URL = content.MetaData.ClickUrl
+	} else if content.MetaData.Intent != "" {
+		linkObject.URL, _ = getDecodeValue(content.MetaData.Intent)
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	}
+
+	nativeResult.Assets = make([]nativeResponse.Asset, 0, len(nativePayload.Assets))
+	var imgIndex = 0
+	var iconIndex = 0
+	for _, asset := range nativePayload.Assets {
+		var responseAsset nativeResponse.Asset
+		if asset.Title != nil {
+			var titleObject nativeResponse.Title
+			titleObject.Text, _ = getDecodeValue(content.MetaData.Title)
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+				adWidth = imgObject.W
+			}
+			responseAsset.Img = &imgObject
+		} else if asset.Data != nil {
+			var dataObject nativeResponse.Data
+			dataObject.Label = ""
+			dataObject.Value = ""
+			if asset.Data.Type == native1.DataAssetTypeDesc || asset.Data.Type == native1.DataAssetTypeDesc2 {
+				dataObject.Label = "desc"
+				dataObject.Value, _ = getDecodeValue(content.MetaData.Description)
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+func (a *adapter) extractAdmPicture(content *content) (adm string, adWidth int64, adHeight int64, err error) {
+	if content == nil {
+		return "", 0, 0, fmt.Errorf("extractAdmPicture: content is empty")
+	}
+
+	var clickUrl = ""
+	if content.MetaData.ClickUrl != "" {
+		clickUrl = content.MetaData.ClickUrl
+	} else if content.MetaData.Intent != "" {
+		clickUrl, _ = getDecodeValue(content.MetaData.Intent)
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if content.MetaData.ImageInfo != nil {
+		imageInfoUrl = content.MetaData.ImageInfo[0].Url
+		adHeight = content.MetaData.ImageInfo[0].Height
+		adWidth = content.MetaData.ImageInfo[0].Width
+	} else {
+		return "", 0, 0, fmt.Errorf("content.MetaData.ImageInfo is empty")
+	}
+
+	var imageTitle = ""
+	imageTitle, _ = getDecodeValue(content.MetaData.Title)
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+			}
+			if monitor.EventType == "click" {
+				for _, trackerUrl := range monitor.Url {
+					linkObject.ClickTrackers = append(linkObject.ClickTrackers, trackerUrl)
+				}
+			}
+			if monitor.EventType == "imp" {
+				for _, trackerUrl := range monitor.Url {
+					nativeResult.ImpTrackers = append(nativeResult.ImpTrackers, trackerUrl)
+				}
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	// dsp imp click tracking + imp click tracking
+	if content.Monitor != nil {
+		for _, monitor := range content.Monitor {
+			if len(monitor.Url) == 0 {
+				continue
+			}
+			if monitor.EventType == "click" {
+				for _, trackerUrl := range monitor.Url {
+					linkObject.ClickTrackers = append(linkObject.ClickTrackers, trackerUrl)
+				}
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+		"var req = new XMLHttpRequest();" +
+		"req.open('GET', url, true);" +
+		"req.send(null);" +
+		"}" +
+		"function sendSomeGetReq(urls) {" +
+		"for (var i = 0; i < urls.length; i++) {" +
+		"sendOneGetReq(urls[i]);" +
+		"}" +
+		"}" +
+		"</script>"
```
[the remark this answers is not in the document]
wwwyyy: done

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	return adm, adWidth, adHeight, nil
+}
+
+func getDspImpClickTrackings(content *content) (dspImpTrackings []string, dspClickTrackings string) {
+	if content.Monitor != nil {
```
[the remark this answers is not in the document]
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if len(eles) == 0 {
+		return ""
+	}
+	var strs strings.Builder
+	for i := 0; i < len(eles); i++ {
+		strs.WriteString("\"" + eles[i] + "\"")
+		if i < len(eles)-1 {
+			strs.WriteString(",")
+		}
+	}
```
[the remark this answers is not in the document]
wwwyyy: thank you

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	if duration == 0 {
+		return "00:00:00.000"
+	}
+	var totalSec = int(duration) / 1000
+	var mmm = int(duration) % 1000
+	var hour = totalSec / 3600
+	var tmp = totalSec % 3600
+	var min = tmp / 60
+	var sec = tmp % 60
+	var result string
```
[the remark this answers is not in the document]
wwwyyy: done

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1320 @@
...
+	result += addZero(strconv.Itoa(hour), ":", false)
+	result += addZero(strconv.Itoa(min), ":", false)
+	result += addZero(strconv.Itoa(sec), ".", false)
+	result += addZero(strconv.Itoa(mmm), "", true)
+	return result
+}
+
+func addZero(str string, dot string, isMmm bool) (result string) {
```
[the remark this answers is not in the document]
wwwyyy: done

adapters/huaweiads/huaweiads.go
wwwyyy: done

adapters/huaweiads/huaweiads.go
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
wwwyyy: corrected

adapters/huaweiads/huaweiads_test.go
wwwyyy: corrected

adapters/huaweiads/huaweiads.go
wwwyyy: The current demand only supports banner ，native ， and will support videos, etc. later；

adapters/huaweiads/huaweiads.go
wwwyyy: added in banner1.json

adapters/huaweiads/huaweiads.go
wwwyyy: done

adapters/huaweiads/huaweiads.go
wwwyyy: already added nativeIncludeVide nativeSingleImage nativeThreeImage

adapters/huaweiads/huaweiads.go
wwwyyy: add missing_native_request.json

adapters/huaweiads/huaweiads.go
wwwyyy: added in banner1.json

adapters/huaweiads/huaweiads.go
wwwyyy: added bad_response_retcode_210.json bad_response_retcode_408.json bad_response_retcode_500.json

adapters/huaweiads/huaweiads.go
wwwyyy: added bad_response_retcode30_204.json bad_response_dont_find_impid.json

adapters/huaweiads/huaweiads.go
wwwyyy: added missing_native_request.json

adapters/huaweiads/huaweiads.go
wwwyyy: add video.json

adapters/huaweiads/huaweiads.go
wwwyyy: add nativeThreeImageIncludeIcon.json

adapters/huaweiads/huaweiads.go
wwwyyy: add video.json

adapters/huaweiads/huaweiads.go
wwwyyy: add video.json

adapters/huaweiads/huaweiads.go
wwwyyy: This tester has already tested it, I don’t know how to add it, because the logic is based on the current time to obtain the results, it is dynamically generated

adapters/huaweiads/huaweiads.go
wwwyyy: sorry，the zone contained in the result is dynamic, how to add a fixed value to verify it?

adapters/huaweiads/huaweiads.go
wwwyyy: add bad_response_not_native.json

adapters/huaweiads/huaweiads.go
hhhjort: What you can do is add  a field to your adapter that takes an interface with one method. The method would normally just call `time.Now()`, and return that, or your entire string value for `nonce` below. The `builder` function would populate the adapter with an empty struct with the intended method. In testing, you would either replace the interface with one that returns a constant value, or just build the adapter by hand with a constant value providing interface. That way in tests you will always be dealing with a known value for the time, rather than the `time.Now()` that you will be using in production.

This does slightly complicate your adapter, but makes it much more testable.

wwwyyy: Done, thank you @hhhjort

adapters/huaweiads/huaweiads.go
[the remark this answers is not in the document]
hhhjort: Is the purpose of `t` now just to get a default timezone? If so then perhaps better to do it once on building your adapter, then save it in the struct and use adapter methods to access it rather than independent functions.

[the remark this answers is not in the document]
wwwyyy: In fact, when users use it, if they pass the time zone and the time under this time zone according to our requirements, our default time zone will not be used. This is for the user to be normal when the time zone is not passed.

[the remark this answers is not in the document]
wwwyyy: This default time zone will not be used under normal circumstances

wwwyyy:
@hhhjort @SyntaxNode Sorry，may I ask when you can merge the adapter code? the client needs the code to debug. Thanks a lot

hhhjort (dismissed):
LGTM

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1293 @@
...
+	var tmp = totalSec % 3600
+	var min = tmp / 60
+	var sec = tmp % 60
+	var result strings.Builder
+	result.WriteString(addZero(strconv.Itoa(hour), ":", false))
+	result.WriteString(addZero(strconv.Itoa(min), ":", false))
+	result.WriteString(addZero(strconv.Itoa(sec), ".", false))
+	result.WriteString(addZero(strconv.Itoa(mmm), "", true))
+	return result.String()
+}
```
guscarreon, line 1090: I don't have a problem with `getDuration(duration int64)`, it looks pretty nice and I don't mind leaving it as is. However, golang's `"time"` package seems to be able to do something similar. An alternative to `getDuration` solely with tools provided in the `"time"` package could probably be:
```
func getDurationWithGolangTools(dur int64) string {
    var d time.Duration = time.Duration(dur) * time.Millisecond
    t := time.Time{}.Add(d)

    return t.Format("15:04:05.000")
}
```
The following test case seems to match the return values of both functions:
```
func TestFunc(t *testing.T) {
	alternativeFunc := func(dur int64) string {
		var d time.Duration = time.Duration(dur) * time.Millisecond
		t := time.Time{}.Add(d)

		return t.Format("15:04:05.000")
	}

	testCases := []struct {
		desc     string
		duration int64
	}{
		{
			"01:03:20.001",
			3800001,
		},
		{
			"13:03:20.001",
			3600*13 + 3*60 + 20001,
		},
		{
			"twenty five hours",
			25 * 1000,
		},
		{
			"23:59:59.000",
			23*3600*1000 + 3599*1000,
		},
	}
	for _, tc := range testCases {
		// current duration function
		curr := getDuration(tc.duration)

		// duration function using golang libraries
		proposed := alternativeFunc(tc.duration)

		assert.Equal(t, curr, proposed)
	}
}
```
This is just a suggestion to maintain less code. Feel free to leave `getDuration(duration int64)` as is If you think it's more convenient.

guscarreon:
Also, our testing framework to detect race conditions was modified some months ago to no longer require the `params/race/*.json` files. We are sorry for the delay in our documentation update.  Can you please remove the following, no longer needed files?
- `adapters/huaweiads/huaweiadstest/params/race/banner.json`
- `adapters/huaweiads/huaweiadstest/params/race/native.json`

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1293 @@
...
+
+func getDspImpClickTrackings(content *content) (dspImpTrackings []string, dspClickTrackings string) {
+	for _, monitor := range content.Monitor {
+		if monitor.EventType == "imp" && len(monitor.Url) != 0 {
+			dspImpTrackings = monitor.Url
+		}
+		if monitor.EventType == "click" && len(monitor.Url) != 0 {
+			dspClickTrackings = getStrings(monitor.Url)
+		}
```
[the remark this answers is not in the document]
wwwyyy: corrected

adapters/huaweiads/params_test.go
```diff
@@ -0,0 +1,54 @@
...
+		if err := validator.Validate(openrtb_ext.BidderGumGum, json.RawMessage(invalidParam)); err == nil {
+			t.Errorf("Schema allowed unexpected params: %s", invalidParam)
+		}
+	}
+}
+
+var validParams = []string{
+	`{"slotid": "m8x9x3rzff","adtype": "banner","publisherid": "123","signkey": "2f910deac52ff34f0d80585d8664c55e3422ff3c6aeb5e1cf2ff94f1ac6a9642","keyid": "41","clienttime": "2018-11-02 16:34:07.981+1300"}`,
+	`{"slotid": "m8x9x3rzff","adtype": "banner","publisherid": "123","signkey": "2f910deac52ff34f0d80585d8664c55e3422ff3c6aeb5e1cf2ff94f1ac6a9642","keyid": "41"}`,
+	`{"slotid": "o68q7hg4i4","adtype": "native","publisherid": "123","signkey": "2f910deac52ff34f0d80585d8664c55e3422ff3c6aeb5e1cf2ff94f1ac6a9642","keyid": "35"}`,
```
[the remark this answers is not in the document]
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1293 @@
...
+func (a *adapter) MakeRequests(openRTBRequest *openrtb2.BidRequest,
+	reqInfo *adapters.ExtraRequestInfo) (requestsToBidder []*adapters.RequestData, errs []error) {
+	// the upstream code already confirms that there is a non-zero number of impressions
+	numRequests := len(openRTBRequest.Imp)
+	var request huaweiAdsRequest
+	var header http.Header
+	var multislot = make([]adslot30, 0, numRequests)
+
+	var huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds
+	for index := 0; index < numRequests && numRequests > 0; index++ {
```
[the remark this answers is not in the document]
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1293 @@
...
+		} else if imp.Audio != nil {
+			mediaType = openrtb_ext.BidTypeAudio
+		}
+		mapSlotid2MediaType[huaweiAdsExt.SlotId] = mediaType
+	}
+
+	if len(mapSlotid2MediaType) < 1 || len(mapSlotid2Imp) < 1 {
+		return nil, errors.New("convertHuaweiAdsResp2BidderResp: openRTBRequest.imp is nil")
+	}
+	if huaweiAdsResponse.Multiad == nil {
```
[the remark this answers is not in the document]
wwwyyy: deleted

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1293 @@
...
+		if monitor.EventType == "click" && len(monitor.Url) != 0 {
+			dspClickTrackings = getStrings(monitor.Url)
+		}
+	}
+	return dspImpTrackings, dspClickTrackings
+}
+
+func getStrings(eles []string) string {
+	if len(eles) == 0 {
+		return ""
```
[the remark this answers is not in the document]
wwwyyy: thank you, I will leaving it

wwwyyy:
@guscarreon thank you, I will deleted those race files.

adapters/huaweiads/huaweiads.go
```diff
@@ -0,0 +1,1293 @@
...
+func unmarshalExtImpHuaweiAds(openRTBImp *openrtb2.Imp) (*openrtb_ext.ExtImpHuaweiAds, error) {
+	var bidderExt adapters.ExtImpBidder
+	var huaweiAdsImpExt openrtb_ext.ExtImpHuaweiAds
+	if err := json.Unmarshal(openRTBImp.Ext, &bidderExt); err != nil {
+		return nil, errors.New("Unmarshal: openRTBImp.Ext -> bidderExt failed")
+	}
+	if err := json.Unmarshal(bidderExt.Bidder, &huaweiAdsImpExt); err != nil {
+		return nil, errors.New("Unmarshal: bidderExt.Bidder -> huaweiAdsImpExt failed")
+	}
+	if huaweiAdsImpExt.SlotId == "" {
```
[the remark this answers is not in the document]
wwwyyy: thank you, deleted

adapters/huaweiads/huaweiads.go
guscarreon: Awesome. Just in case I missed and edge test case, make sure the new function is 100% equivalent to yours.

wwwyyy: yeah, I have tested.

guscarreon:
> @guscarreon thank you, I will deleted those race files.

@wwwyyy I added two tests to assert both functions behave the same. Do we expect negative values or very big values (greater than a day in milliseconds)?
```
func TestFunc(t *testing.T) {
	alternativeFunc := func(dur int64) string {
		var d time.Duration = time.Duration(dur) * time.Millisecond
		t := time.Time{}.Add(d)

		return t.Format("15:04:05.000")
	}

	testCases := []struct {
		desc     string
		duration int64
	}{
		{
			"00:00:00.000",
			0,
		},
		{ //fails
			"negative",
			-1,
		},
		{ //fails
			"a week minus one millisecond",
			3600*24*7*1000 - 1,
		},
		{
			"01:03:20.001",
			3800001,
		},
		{
			"13:03:20.001",
			3600*13 + 3*60 + 20001,
		},
		{
			"twenty five hours",
			25 * 1000,
		},
		{
			"23:59:59.000",
			23*3600*1000 + 3599*1000,
		},
	}
	for _, tc := range testCases {
		// current duration function
		curr := getDuration(tc.duration)

		// duration function using golang libraries
		proposed := alternativeFunc(tc.duration)

		assert.Equal(t, curr, proposed, tc.desc)
	}
}
```

wwwyyy:
@guscarreon there is no big number, and I think it's enough, thank you.

guscarreon (approved):
@wwwyyy thank you for addressing all the feedback. Your adapter looks fantasic.

wwwyyy:
@guscarreon Thanks for your suggestion 👍

Issue #1308, opened 2020-05-19
Support bids with 0.00$ price.


moteus:
We set up a private deal with 0.00$ CPM. But when prebid-server gets this bid it treats it as invalid and responses with the error message like `Bid "***" does not contain a positive 'price'`.

bretg:
Please help us understand why 0.00 should be considered a valid bid. Could there be negative 
 value bids? (oil prices have opened my mind to general strangeness. :-)

I'm generally not in favor of globally allowing 0-value open market bids. But am more open to a couple of compromise options:
1) Allow 0-value bids for deals only
2) Add a flag to the request where PBS skips that check
```
ext: {
  prebid: {
    options: {
      allowzerovaluebids: true
    }
  }
}
```
ext.prebid can be specified on Prebid.js s2sConfig or in the Stored Requests.

moteus:
> Please help us understand why 0.00 should be considered a valid bid.
My adops team told me `donation for social campaigns`.

A flag is ok for us especially if it will be possible to set it per imp object.

bretg:
We discussed in the committee meeting and propose allowing 0-value bids for deals. The use case will be discussed in other Prebid meetings next week.

bretg:
The Publisher Taskforce has agreed that zero-value bids should be allowed for deals.

bretg:
Done in PBS-Java 1.46

AlexBVolcy:
Hey @bretg I'm addressing this in PBS in this [PR] (https://github.com/prebid/prebid-server/pull/1910) and was wondering about your comment about adding a flag to skip this check (i.e. allowzerovaluebids).

In my implementation where I allow $0.00 price bids, I don't include this flag, but I'm wondering if this flag is something you would still want to see even though I've gone about this issue in a different way? Thanks!

bretg:
There's no flag. The list above was a set of options. The comment on May 29th says "propose allowing 0-value bids for deals" -- that's option 1. 

I've updated the May 19th comment to hopefully make it clear that we settled on option 1.

AlexBVolcy:
Got it, thanks!

PR #1910, opened 2021-06-30
Allowed $0.00 price bids if there are deals

AlexBVolcy:
Addressing this issue https://github.com/prebid/prebid-server/issues/1308

### DIFF
--- a/exchange/bidder_validate_bids.go
+++ b/exchange/bidder_validate_bids.go
@@ -110,8 +110,11 @@ func validateBid(bid *pbsOrtbBid) (bool, error) {
 	if bid.bid.ImpID == "" {
 		return false, fmt.Errorf("Bid \"%s\" missing required field 'impid'", bid.bid.ID)
 	}
-	if bid.bid.Price <= 0.0 {
-		return false, fmt.Errorf("Bid \"%s\" does not contain a positive 'price'", bid.bid.ID)
+	if bid.bid.Price < 0.0 {
+		return false, fmt.Errorf("Bid \"%s\" does not contain a positive (or zero if there is a deal) 'price'", bid.bid.ID)
+	}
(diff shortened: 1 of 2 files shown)


exchange/bidder_validate_bids.go
```diff
@@ -110,8 +110,11 @@ func validateBid(bid *pbsOrtbBid) (bool, error) {
 	if bid.bid.ImpID == "" {
 		return false, fmt.Errorf("Bid \"%s\" missing required field 'impid'", bid.bid.ID)
 	}
-	if bid.bid.Price <= 0.0 {
-		return false, fmt.Errorf("Bid \"%s\" does not contain a positive 'price'", bid.bid.ID)
+	if bid.bid.Price < 0.0 {
+		return false, fmt.Errorf("Bid \"%s\" does not contain a positive (or zero if there is a deal) 'price'", bid.bid.ID)
+	}
+	if bid.bid.Price <= 0.0 && bid.bid.DealID == "" {
```
hhhjort, line 116: You should be able to have 'bid.bid.Price == 0.0 && bid.bid.DealID', I don't think it is possible for both 'x < 0.0' and 'x == 0.0' to be false yet 'x <= 0.0' be true.

mansinahar:
The issue suggests adding a `allowzerovaluebids` flag that skips this check for $0.00 bids for deals. However, I don't see this implemented as part of the Java PR for this either but can you please check with @bretg to confirm that's not needed?

exchange/bidder_validate_bids.go
AlexBVolcy: Ah yeah that makes sense. Will update!

AlexBVolcy: Just updated!

hhhjort (approved):
LGTM assuming Mansi's concern is cleared.

mansinahar (approved):

PR #1912, opened 2021-07-01
New Adapter: Axonix

cesarfd:
Add a new Prebid-Server (Go) adapter for Axonix (we already have a prebid.js adapter)


guscarreon:
@cesarfd thank you for submitting your adapter pull request. We will review soon. Did you also submit a documentation PR for this newly created adapter in https://github.com/prebid/prebid.github.io/pulls?

jeanstemp:
Docs PR: https://github.com/prebid/prebid.github.io/pull/3085

adapters/axonix/axonix.go
```diff
@@ -0,0 +1,135 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type AxonixAdapter struct {
```
hhhjort, line 15: This can simply be `type adapter struct {`, as if you were to reference it from outside this package, it would be `axonix.AxonixAdapter`, which is redundant. Also the adapters are only used as interfaces, so it need not be exported. The `Builder()` method provides the hook to access your adapter implementation.

adapters/axonix/axonix.go
```diff
@@ -0,0 +1,135 @@
...
+				BidType: getMediaType(bid.ImpID, request.Imp),
+			}
+			bidResponse.Bids = append(bidResponse.Bids, b)
+		}
+	}
+
+	return bidResponse, nil
+}
+
+func getMediaType(impId string, imps []openrtb2.Imp) openrtb_ext.BidType {
```
hhhjort, line 123: Do you always strictly only respond with bids of these types when given multiformat requests, or are these just highly likely types? It looks likely that OpenRTB 2.6 will natively support reporting bid type, which will help here when it comes out and we move to supporting it.

Knowing the bid type is useful for display clients, so they don't have to guess at the payload type and try to display properly based on that.

cesarfd: This function follows the same order of preference that our backend follows to decide the imp type so it should match.

adapters/axonix/axonix_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamplesWithConfiguredURI(t *testing.T) {
```
hhhjort, line 11: The /exemplary directory is intended to hold just a few reference examples a publisher can look at for a reference as to how to use your adapter. Additional tests to catch corner cases where there are no errors go into /supplemental. Are you sure all your positive tests belong in /exemplary? too many examples may confuse publishers as to what they should try to follow.

config/config.go
```diff
@@ -837,6 +837,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.audiencenetwork.disabled", true)
 	v.SetDefault("adapters.audiencenetwork.endpoint", "https://an.facebook.com/placementbid.ortb")
 	v.SetDefault("adapters.avocet.disabled", true)
+	v.SetDefault("adapters.axonix.disabled", true)
```
hhhjort, line 840: I am curious if there is a specific reason to default disabled. Is it just to have more control over who is using your adapter?

cesarfd: Yeah, we prefer to enable our supply partners on a one by one basis.

hhhjort (approved):
LGTM

adapters/axonix/axonix.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		errors = append(errors, err)
+		return nil, errors
+	}
+
+	var axonixExt openrtb_ext.ExtImpAxonix
+	err = json.Unmarshal(bidderExt.Bidder, &axonixExt)
+	if err != nil {
+		err = &errortypes.BadInput{
+			Message: "ext.bidder.supplyId not provided",
```
guscarreon, line 58: This message is a little misleading for the following reason: given that the `"supplyId"` was listed as `"required"` in `static/bidder-params/axonix.json`, there's no way to reach this line with a missing `ext.bidder.supplyId` field.  I believe this message would be more useful if `ext.bidder.supplyId` was described as "malformed". Thoughts?

guscarreon (commented):
Some comments before approving

adapters/axonix/axonix.go
```diff
@@ -0,0 +1,135 @@
...
+		err = &errortypes.BadInput{
+			Message: "supplyId is empty",
+		}
+		errors = append(errors, err)
+		return nil, errors
+	}
+
+	thisURI := a.URI
+	if len(thisURI) == 0 {
+		thisURI = "https://openrtb-us-east-1.axonix.com/supply/prebid-server/" + axonixExt.SupplyId
```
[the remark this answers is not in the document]
cesarfd: `SupplyId` is a UUID that will be assigned to each supply partner so highly unlikely if not impossible that it will come with any other spurious character.

guscarreon (approved):
@cesarfd thank you for addressing the feedback comments. LGTM

hhhjort (approved):

PR #1913, opened 2021-07-01
TheMediaGrid: Add params.keywords processing

TheMediaGrid:
Update s2s TheMediaGrid adapter to read imp.ext.bidder.keywords and copy it in ext.keywords


adapters/grid/grid.go
```diff
@@ -69,16 +74,41 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
...
+		return err
+	}
+	if gridExt.Keywords != nil {
+		if request.Ext != nil {
+			if err := json.Unmarshal(request.Ext, &reqExt); err != nil {
+				return err
+			}
+		}
+
+		reqExt.Keywords = gridExt.Keywords
```
guscarreon, line 94: I realize the `root.ext` json usually doesn't come with a `keywords` field but, should we check whether or not it is empty before rewriting? Is there any possibility that someone might slip some keywords in there that we might be interested in?

guscarreon (commented):
Looks good to me, just one quick question before approving

adapters/grid/grid.go
TheMediaGrid: Hello! We have fixed this case + added keywords format validation

SyntaxNode:
Looks like a recent push introduced new validation errors. We'll re-review once those are addressed.

adapters/grid/grid.go
```diff
@@ -33,6 +34,26 @@ type ExtImp struct {
...
+
+type KeywordsPublisherItem struct {
+	Name     string           `json:"name"`
+	Segments []KeywordSegment `json:"segments"`
+}
+
+type ExtKeywords struct {
+	Site json.RawMessage `json:"site,omitempty"`
+	User json.RawMessage `json:"user,omitempty"`
+}
```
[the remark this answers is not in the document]
TheMediaGrid: Fiexd.
+ there is fix by this discuss https://github.com/prebid/prebid-server/issues/1922 to support correct usage of bid[].ext.prebid.meta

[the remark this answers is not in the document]
TheMediaGrid: @guscarreon @SyntaxNode can you please review this PR ASAP?

adapters/grid/grid.go
```diff
@@ -69,16 +97,224 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
...
+					} else {
+						extKeywords[key] = keyword
+					}
+				}
+			}
+		}
+	}
+}
+
+func reformatExtKeywords(extKeywords map[string]interface{}) {
```
guscarreon, line 158: Can we refactor the `reformatExtKeywords(extKeywords map[string]interface{})` function or subdivide it into subfunctions to make it easier to read?

adapters/grid/grid.go
```diff
@@ -69,16 +97,224 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
...
+		switch pubData.(type) {
+		default:
+			delete(extKeywords, name)
+		case []interface{}:
+			formatedPubArr := make([]KeywordsPublisherItem, 0) // make([]interface{}, 0)
+			pubArr := pubData.([]interface{})
+			for _, item := range pubArr {
+				switch item.(type) {
+				// default:
+				//	  formatedPubArr = append(formatedPubArr, item)
```
guscarreon, line 169: if these couple of lines are of no use, can we remove?

adapters/grid/grid.go
```diff
@@ -69,16 +97,224 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
...
+					}
+				}
+			}
+		}
+	}
+}
+
+func reformatExtKeywords(extKeywords map[string]interface{}) {
+	for name, pubData := range extKeywords {
+		switch pubData.(type) {
```
guscarreon, line 160: Given that the `type switch` statements in lines 160 and 167 only seem to have one `case` statement, can we use a type assertion instead?
```
if v, ok = x.(T); ok {
    // ...
}
```

Or even the reflect library:
```
x := 42

xtype := reflect.TypeOf(x).Kind()

if xtype == reflect.Int {
    println("x is int")
}
```

guscarreon (commented):
Left some comments

adapters/bidder.go
```diff
@@ -92,11 +92,13 @@ func NewBidderResponse() *BidderResponse {
 // TypedBid.BidType will become "response.seatbid[i].bid.ext.prebid.type" in the final OpenRTB response.
 // TypedBid.BidVideo will become "response.seatbid[i].bid.ext.prebid.video" in the final OpenRTB response.
 // TypedBid.DealPriority is optionally provided by adapters and used internally by the exchange to support deal targeted campaigns.
+// TypedBid.BidMeta will become "response.seatbid[i].bid.ext.prebid.meta" in the final OpenRTB response.
 type TypedBid struct {
 	Bid          *openrtb2.Bid
 	BidType      openrtb_ext.BidType
 	BidVideo     *openrtb_ext.ExtBidPrebidVideo
 	DealPriority int
+	BidMeta      *openrtb_ext.ExtBidPrebidMeta
```
SyntaxNode, line 101: Please include `BidMeta` alphabetically in the comments and `TypedBid` object.

exchange/bidder.go
```diff
@@ -71,6 +72,7 @@ type pbsOrtbBid struct {
 	dealPriority      int
 	dealTierSatisfied bool
 	generatedBidID    string
+	bidMeta           *openrtb_ext.ExtBidPrebidMeta
```
SyntaxNode, line 75: Please include bidMeta alphabetically in the comments and pbsOrtbBid object.

SyntaxNode (commented):
I completely agree with the BidMeta changes to PBS-Core to solve the overwriting problem.

adapters/grid/grid.go
```diff
@@ -69,16 +97,222 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
...
+		}
+	}
+}
+
+func updateExtKeywords(keywords json.RawMessage, request *openrtb2.BidRequest) json.RawMessage {
+	var extKeywords map[string]interface{}
+	var extKWSite map[string]interface{}
+	var extKWUser map[string]interface{}
+
+	json.Unmarshal(keywords, &extKeywords)
```
guscarreon, line 225: We realize an error scenario here is unlikely but we believe it's better to try to check for it and handle it

adapters/grid/grid.go
```diff
@@ -69,16 +97,222 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
...
+
+func mergeWithReqExtKeywords(extKeywords map[string]interface{}, request *openrtb2.BidRequest) error {
+	var reqExt ReqExt
+	var reqExtKeywords map[string]interface{}
+
+	if err := json.Unmarshal(request.Ext, &reqExt); err != nil {
+		return err
+	}
+	if reqExt.Keywords != nil {
+		json.Unmarshal(reqExt.Keywords, &reqExtKeywords)
```
guscarreon, line 132: We realize an error scenario here is unlikely but we believe it's better to try to check for it and handle it

static/bidder-params/grid.json
```diff
@@ -7,7 +7,19 @@
...
+          "type": "object"
+        },
+        "user": {
+          "type": "object"
+        }
+      }
     }
   },
-  "required": []
+  "required": ["uid"]
```
hhhjort, line 24: Do you have any statistics on how many requests come in currently without a `"uid"` parameter? There was a design choice that any request that does not pass validation on any one of the bidder JSON validation schemas fails entirely, not just for the one bidder. This was done as when a publisher sets up a new configuration with multiple bidders, a bidder parameter failure on one bidder may be masked by all the successful bids from other bidders. We would rather the requests were correct from the start.

This means any current bidder configs that do not include `uid` will fail to fetch ads from any bidder. This will break publisher configs across all bidders, not just yours. The preferred way to add a new required field is to initially list it as optional in the schema, and have the adapter return an error if it is missing. An adapter returning an error outside of JSON schema validation will not invalidate the whole request. Then once we all have some confidence that all the publishers have updated their configs, put in a followup PR to change the schema to required.

TheMediaGrid:
@guscarreon @SyntaxNode @hhhjort any chance you could take a look at the most recent changes that we've made as per Scott's comments? Any insight into why the auto tests are failing would be very useful as well.

SyntaxNode:
> @guscarreon @SyntaxNode @hhhjort any chance you could take a look at the most recent changes that we've made as per Scott's comments? Any insight into why the auto tests are failing would be very useful as well.

I'm looking into it. I didn't see this locally either. I suspect it's a case where the segments array order is indeterminate, but will confirm.

SyntaxNode:
> > @guscarreon @SyntaxNode @hhhjort any chance you could take a look at the most recent changes that we've made as per Scott's comments? Any insight into why the auto tests are failing would be very useful as well.
> 
> I'm looking into it. I didn't see this locally either. I suspect it's a case where the segments array order is indeterminate, but will confirm.

Found it.

The issue is caused by the indeterminate ordering of the segments when constructed from an array of strings, referred to as the "compatible alternate segment format" in the code. 

The ordering relies on the iteration order of keys within a map, which is purposefully inconsistent in Go. I don't believe the ordering matters to your bidding server, but it does matter to the json tests since arrays are always tested as order dependent. I'm not sure how easily or safely we can modify the test layer, but it easy to sort the keys to ensure deterministic order.

Change:
```
// compose compatible alternate segment format
for potentialSegmentName, potentialSegmentValues := range publisherItem {
	// values must be an array
	if valuesSlice, ok := potentialSegmentValues.([]interface{}); ok {
		for _, value := range valuesSlice {
			if valueAsString, ok := value.(string); ok {
				segments = append(segments, KeywordSegment{Name: potentialSegmentName, Value: valueAsString})
			}
		}
	}
}
```

To:
```
// ensure consistent ordering for publisher item map
publisherItemKeys := make([]string, 0, len(publisherItem))
for v := range publisherItem {
	publisherItemKeys = append(publisherItemKeys, v)
}
sort.Strings(publisherItemKeys)

// compose compatible alternate segment format
for _, potentialSegmentName := range publisherItemKeys {
	potentialSegmentValues := publisherItem[potentialSegmentName]
	// values must be an array
	if valuesSlice, ok := potentialSegmentValues.([]interface{}); ok {
		for _, value := range valuesSlice {
			if valueAsString, ok := value.(string); ok {
				segments = append(segments, KeywordSegment{Name: potentialSegmentName, Value: valueAsString})
			}
		}
	}
}
```

This will then consistently fail until the mixed test is updated to the following order:
```
"segments": [
    { "name": "segName1", "value": "segVal1" },
    { "name": "segName2", "value": "segVal2" },
    { "name": "bottom", "value": "bottomKey1" },
    { "name": "bottom", "value": "bottomKey2" },
    { "name": "segments", "value": "stringKey" }
]
```

After which the tests will consistently passes. I don't believe this is a new issue, but was missed earlier just by chance.

TheMediaGrid:
@SyntaxNode @guscarreon @hhhjort @bsardo can anybody please review this PR? it's been a while

util/maputil/maputil.go
```diff
@@ -19,3 +19,12 @@ func ReadEmbeddedSlice(m map[string]interface{}, k string) ([]interface{}, bool)
 
 	return nil, false
 }
+
+// ReadEmbeddedString reads element k from the map m as a string
+func ReadEmbeddedString(m map[string]interface{}, k string) (string, bool) {
```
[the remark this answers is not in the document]
TheMediaGrid: @bsardo we added the test as requested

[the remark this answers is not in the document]
bsardo: 👍 looks good

guscarreon (commented):
@TheMediaGrid : In order to double-check the expected behavior or your keywords algorithm could you please write 3 more tests where there are keywords:
- In the root.Ext only 
- In the root.site only 
- In the root.user only

I think your code looks pretty solid and with these three extra tests I'd be good to approve.

guscarreon (approved):
The code looks pretty solid. It'd be nice to add 3 more tests in a follow-up PR where there are keywords:

- In the root.Ext only
- In the root.site only
- In the root.user only

bsardo (approved):

PR #1917, opened 2021-07-07
Criteo: Fix fields mapping error when building bid from bidder response

leonardlabat:
We've detected an error in our adapter as the field mapped when handling bid responses from our bidder are not consistent with the actual naming of the fields used in our production bidder bid responses.


AlexBVolcy (approved):
LGTM

leonardlabat:
Thanks for the review @AlexBVolcy 
Could it be part of the next PBS release ? 

Thanks

bsardo (approved):
LGTM

PR #1920, opened 2021-07-12
Cookie refresh expiration date on/off

matthieularere-msq:
In order to comply with the french interpretation of GDPR, and more specificaly about TCF v2.1, I need to be able to ensure that user sync won't refresh the cookie expiration date. In order to do that I've added a config field host_cookie.refresh which is set to true by default so the current behavior is not broke.

Then, when cookie is set, in host_cookie.refresh is set to false, the ttl value is updated to the duration that left based on the cookie birth date + normal duration (birthday + ttl - today).

### DIFF
--- a/usersync/cookie.go
+++ b/usersync/cookie.go
@@ -176,6 +176,9 @@ func (cookie *PBSCookie) GetId(bidderName openrtb_ext.BidderName) (id string, ex
 
 // SetCookieOnResponse is a shortcut for "ToHTTPCookie(); cookie.setDomain(domain); setCookie(w, cookie)"
 func (cookie *PBSCookie) SetCookieOnResponse(w http.ResponseWriter, setSiteCookie bool, cfg *config.HostCookie, ttl time.Duration) {
+	if !cfg.Refresh {
+		ttl = cookie.birthday.Add(ttl).Sub(time.Now())
+	}
 	httpCookie := cookie.ToHTTPCookie(ttl)
 	var domain string = cfg.Domain
 
--- a/config/config.go
+++ b/config/config.go
@@ -321,7 +321,7 @@ type HostCookie struct {
 	OptInURL           string `mapstructure:"opt_in_url"`
 	MaxCookieSizeBytes int    `mapstructure:"max_cookie_size_bytes"`
 	OptOutCookie       Cookie `mapstructure:"optout_cookie"`
-	Refresh            Bool   `mapstructure:"refresh"`
+	Refresh            bool   `mapstructure:"refresh"`
 	// Cookie timeout in days
 	TTL int64 `mapstructure:"ttl_days"`
 }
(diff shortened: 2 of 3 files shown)


usersync/cookie.go
```diff
@@ -176,6 +176,9 @@ func (cookie *PBSCookie) GetId(bidderName openrtb_ext.BidderName) (id string, ex
 
 // SetCookieOnResponse is a shortcut for "ToHTTPCookie(); cookie.setDomain(domain); setCookie(w, cookie)"
 func (cookie *PBSCookie) SetCookieOnResponse(w http.ResponseWriter, setSiteCookie bool, cfg *config.HostCookie, ttl time.Duration) {
+	if !cfg.Refresh {
+		ttl = cookie.birthday.Add(ttl).Sub(time.Now())
+	}
```
SyntaxNode, line 181: Please add a test case for this new condition.

SyntaxNode (commented):
I'm not familiar with the French interpretation of GDPR. Exactly what requirement does have France have which this PR addresses?

usersync/cookie.go
matthieularere-msq: They decided that cookie refresh is forbiden whatever the user consent might have been and regardless how the vendor declared himself in the TCF.

matthieularere-msq: I've added a test case, hope it's what was expected.

usersync/cookie.go
```diff
@@ -176,6 +176,9 @@ func (cookie *PBSCookie) GetId(bidderName openrtb_ext.BidderName) (id string, ex
 
 // SetCookieOnResponse is a shortcut for "ToHTTPCookie(); cookie.setDomain(domain); setCookie(w, cookie)"
 func (cookie *PBSCookie) SetCookieOnResponse(w http.ResponseWriter, setSiteCookie bool, cfg *config.HostCookie, ttl time.Duration) {
+	if !cfg.Refresh {
+		ttl = cookie.birthday.Add(ttl).Sub(time.Now())
```
SyntaxNode, line 180: There is danger in using `time.Now()` to compute the ttl and another `time.Now()` call to compute the new expiration because the `time.Now()` values may be different and still trigger a refresh. I recommend a small refactor to use the exact "birthday" rather than converting to and from a ttl. Along the lines of....

```
func (cookie *PBSCookie) SetCookieOnResponse(w http.ResponseWriter, setSiteCookie bool, cfg *config.HostCookie, ttl time.Duration) {
	cookieExpiration := *cookie.birthday
	if cfg.Refresh {
		cookieExpiration = time.Now().Add(ttl)
	}
	httpCookie := cookie.ToHTTPCookie(cookieExpiration)
	...
	Expires: cookieExpiration,
	...
}
```
This would also require updating `ToHTTPCookie` to accept an expiration instead of a ttl. There are side effects, so perhaps it might be better to add a new `ToHTTPCookieWithExpiration` to differentiate it.

SyntaxNode (commented):
> They decided that cookie refresh is forbiden whatever the user consent might have been and regardless how the vendor declared himself in the TCF.

Could you please point us to docs on this requirement?

patmmccann:
Documneting this:
"Trackers may be used for a maximum period of thirteen (13) months. Once this period has expired, the user's consent must be renewed if the cookies continue to be actively used. A user's frequent use of a website does not constitute a valid renewal of the user's consent."
Source https://www.fieldfisher.com/en/services/privacy-security-and-information/privacy-security-and-information-law-blog/cnil-publishes-revised-guidelines-on-cookies-and-other-tracking-technology

mansinahar:
Closing this out due to inactivity. You're welcome to open it up again in the future with our comments addressed.

PR #1921, opened 2021-07-13
GDPR: host-level per-purpose enforce vendor signals config

bsardo:
- Updated GDPR TCF2 config defaults test to assert using object compare
- Added host-level per-purpose bool enforce vendors config flags that indicate whether to skip consent string vendor checks
- Refactored `impl.go` consent and legit interest conditional statements into methods to improve readability and to ensure test code coverage. These methods also check whether to bypass the consent string vendor check based on the enforce vendor config flag
- Created helper function `allPurposesEnabledPermissions` in `impl_test.go` that returns a ready-made permissions object for use across tests
- Updated `impl_test.go` `TestAllowActivitiesBidRequests` to achieve code coverage of enforce vendor checks

### DIFF
--- a/gdpr/impl.go
+++ b/gdpr/impl.go
@@ -210,8 +210,8 @@ func (p *permissionsImpl) checkPurpose(consent tcf2.ConsentMetadata, vendor api.
 		return true
 	}
 
-	purposeAllowed := consent.PurposeAllowed(purpose) && (weakVendorEnforcement || (vendor.Purpose(purpose) && consent.VendorConsent(vendorID)))
-	legitInterest := consent.PurposeLITransparency(purpose) && (weakVendorEnforcement || (vendor.LegitimateInterest(purpose) && consent.VendorLegitInterest(vendorID)))
+	purposeAllowed := p.consentEstablished(consent, vendor, vendorID, purpose, weakVendorEnforcement)
+	legitInterest := p.legitInterestEstablished(consent, vendor, vendorID, purpose, weakVendorEnforcement)
 
 	if consent.CheckPubRestriction(uint8(purpose), pubRestrictRequireConsent, vendorID) {
 		return purposeAllowed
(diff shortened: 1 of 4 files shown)


gdpr/impl_test.go
```diff
@@ -79,6 +80,9 @@ func TestAllowedSyncs(t *testing.T) {
 			}),
 		},
 	}
+	perms.purposeConfigs = map[consentconstants.Purpose]config.TCF2Purpose{
+		consentconstants.Purpose(1): perms.cfg.TCF2.Purpose1,
+	}
```
guscarreon, line 85: `vendor2AndPurpose1Consent` was defined above, should we add `consentconstants.Purpose(2)` to the map instead of `consentconstants.Purpose(1)`?

`TestProhibitedVendors` defines `purpose1NoVendorConsent` which is passed as third parameter to `perms.HostCookiesAllowed` and `perms.BidderSyncAllowed` and includes `consentconstants.Purpose(1)` in the `perms.purposeConfigs` map, should we expect the same in this test?

guscarreon (commented):
Looks good to me. Left:
- A question
-  A small nitpick that by no means should stop this PR from merging if not implemented.

gdpr/impl_test.go
bsardo: This is ok as is. In this test and `TestProhibitedVendors`, we are using vendor 2 and purpose 1 to exercise different code paths.

In this test, we are verifying that PBS determines host cookies and syncs are allowed when the config says to enforce vendors, which means that the consent string must indicate the user consents to the purpose and to the vendor (which is what the `vendor2AndPurpose1Consent` variable name indicates).

In the `TestProhibitedVendors` test, everything is the same except that the user consents to the purpose but not the vendor (which is what the `purpose1NoVendorConsent` variable name is meant to indicate) so the host cookies allowed and sync allowed checks return false instead of true.

gdpr/impl.go
```diff
@@ -210,8 +210,8 @@ func (p *permissionsImpl) checkPurpose(consent tcf2.ConsentMetadata, vendor api.
 		return true
 	}
 
-	purposeAllowed := consent.PurposeAllowed(purpose) && (weakVendorEnforcement || (vendor.Purpose(purpose) && consent.VendorConsent(vendorID)))
-	legitInterest := consent.PurposeLITransparency(purpose) && (weakVendorEnforcement || (vendor.LegitimateInterest(purpose) && consent.VendorLegitInterest(vendorID)))
+	purposeAllowed := p.consentEstablished(consent, vendor, vendorID, purpose, weakVendorEnforcement)
+	legitInterest := p.legitInterestEstablished(consent, vendor, vendorID, purpose, weakVendorEnforcement)
```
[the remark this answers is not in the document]
bsardo: I'm on the fence; this is a very good point. On the flip side, I like having the logic separate for consent established and legit interest established because it helps us verify we have sufficient test coverage (we want to make sure we exercise each of the conditions for both consent and legit interest). I’d like to punt on this decision and reevaluate your suggestion in a future PR that involves overhauling impl.go and impl_test.go . I’ll note it in that refactor story so I don’t forget.

guscarreon (approved):
LGTM

hhhjort (commented):
LGTM

hhhjort (approved):

PR #1807, opened 2021-04-14
TheMediaGrid: Added processing of imp[].ext.data

TheMediaGrid:
Added processing of imp[].ext.data to add imp[].ext.gpid = imp[].ext.data.adserver.adslot

PR #1925, opened 2021-07-14
Sharethrough: Add support for GPID

epechuzal:
## Type of change
- [x] Feature

## Description of change

Add GPID from bid request.

- [x] official adapter submission

### DIFF
--- a/openrtb_ext/imp_sharethrough.go
+++ b/openrtb_ext/imp_sharethrough.go
@@ -1,13 +1,18 @@
 package openrtb_ext
 
-type ExtImpSharethrough struct {
-	Pkey       string  `json:"pkey"`
-	Iframe     bool    `json:"iframe"`
-	IframeSize []int   `json:"iframeSize"`
-	BidFloor   float64 `json:"bidfloor"`
+type ExtData struct {
+	PBAdSlot string `json:"pbadslot"`
 }
 
 // ExtImpSharethrough defines the contract for bidrequest.imp[i].ext.sharethrough
+type ExtImpSharethrough struct {
+	Pkey       string   `json:"pkey"`
+	Iframe     bool     `json:"iframe"`
+	IframeSize []int    `json:"iframeSize"`
+	BidFloor   float64  `json:"bidfloor"`
+	Data       *ExtData `json:"data,omitempty"`
+}
+
 type ExtImpSharethroughResponse struct {
 	AdServerRequestID string                       `json:"adserverRequestId"`
 	BidID             string                       `json:"bidId"`
--- a/adapters/sharethrough/butler.go
+++ b/adapters/sharethrough/butler.go
@@ -21,6 +21,7 @@ const defaultTmax = 10000 // 10 sec
 type StrAdSeverParams struct {
 	Pkey               string
 	BidID              string
+	GPID               string
 	ConsentRequired    bool
 	ConsentString      string
 	USPrivacySignal    string
@@ -97,6 +98,11 @@ func (s StrOpenRTBTranslator) requestFromOpenRTB(imp openrtb2.Imp, request *open
 		return nil, err
 	}
 
+	var gpid string
+	if strImpParams.Data != nil && strImpParams.Data.PBAdSlot != "" {
+		gpid = strImpParams.Data.PBAdSlot
+	}
+
 	usPolicySignal := ""
 	if usPolicy, err := ccpa.ReadFromRequest(request); err == nil {
 		usPolicySignal = usPolicy.Consent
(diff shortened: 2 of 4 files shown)


openrtb_ext/imp_sharethrough.go
```diff
@@ -1,13 +1,18 @@
...
+	PBAdSlot string `json:"pbadslot"`
 }
 
 // ExtImpSharethrough defines the contract for bidrequest.imp[i].ext.sharethrough
+type ExtImpSharethrough struct {
+	Pkey       string   `json:"pkey"`
+	Iframe     bool     `json:"iframe"`
+	IframeSize []int    `json:"iframeSize"`
+	BidFloor   float64  `json:"bidfloor"`
+	Data       *ExtData `json:"data,omitempty"`
```
VeronikaSolovei9, line 13: You defined a new bidder parameter. Can you please update `static/bidder-params/sharethrough.json` file as well?

epechuzal: @VeronikaSolovei9 do we have to? None of the other adapters which implement some sort of `ExtData` on their bidder parameters have declared it there. What is the correct way to do this?

openrtb_ext/imp_sharethrough.go
```diff
@@ -1,13 +1,18 @@
 package openrtb_ext
 
-type ExtImpSharethrough struct {
-	Pkey       string  `json:"pkey"`
-	Iframe     bool    `json:"iframe"`
-	IframeSize []int   `json:"iframeSize"`
-	BidFloor   float64 `json:"bidfloor"`
+type ExtData struct {
```
bsardo, line 3: Just checking, do you need `ExtData` or could you remove it and just have `PBAdSlot` in `ExtImpSharethrough` with an `omitempty`?

openrtb_ext/imp_sharethrough.go
bsardo: You don't have to but it is definitely preferred/recommended. You'll get parameter validation by PBS core upstream so you can be sure you have valid data in your ext when your adapter code executes.

Bidder param documentation can be found here: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#bidder-parameters

openrtb_ext/imp_sharethrough.go
epechuzal: @bsardo I am not sure. I am trying to get the GPID found in `imp[].ext.data`, similar to #1807.

In pbjs the method to implement was to extract the `ortb2Imp.ext.data.pbadslot` value. I am looking for the equivalent for prebid-server. As far as I can tell from checking PRs from other adapters they all implement it this way, but I was unable to find any official documentation that says exactly how this should be done. 

Can you help? I see that you approved the PR from `TheMediaGrid`, is there something that our implementation is missing?

bsardo: Sure I can help. I did some digging and I now believe your implementation is correct assuming the GPID, which I'm guessing stands for global placement ID, is considered ad-specific first party data.

The documentation here indicates that all PBS request ad-specific first party data belongs at `imp[].ext.data`:
https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#prebid-server-ortb2-extension-summary

bsardo: @epechuzal, sorry I got confused above, hence the strikethrough. I still believe your implementation is correct but I wasn't clear why. Let me back up.

You're expecting `PbAdslot` and `AdServer` at `imp[].ext.data` in the PBS inbound request and you're putting `GPID` in the query string of the PBS outbound request to your server.
For inbound request data, if it is ad-specific first party data, it belongs at `imp[]ext.data` according to the referenced documentation above. It looks like `PbAdslot` and `AdServer` are indeed ad-specific and in the right place, similar to how The Media Grid does it in #1807.

As for `GPID`, you and The Media Grid are both setting it to `PbAdSlot` but in different places in your bid requests. You're setting it in each query string and they are setting it in each `imp.ext` which is fine since that is where your servers are expecting to find that data.

I hope that helps.

epechuzal: @bsardo thank you for looking into it! let me know if there is anything that I should change for this PR to go in

openrtb_ext/imp_sharethrough.go
bsardo: @epechuzal I discussed this with the team and this is actually required because other systems and front-ends may use the defined schema.

bsardo:
This is looking good from my perspective; it just needs `data.pbadslot` added to the json schema as described in this thread: https://github.com/prebid/prebid-server/pull/1925/files#r672619506

epechuzal:
> This is looking good from my perspective; it just needs `data.pbadslot` added to the json schema as described in this thread: https://github.com/prebid/prebid-server/pull/1925/files#r672619506

Excellent, I have added the `data` object to the schema.

bsardo (approved):
LGTM

bsardo:
@epechuzal before we can merge this in, please open a docs PR to update https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/sharethrough.md.

epechuzal:
@bsardo does this work? prebid/prebid.github.io#3141

VeronikaSolovei9 (approved):
LGTM

PR #1927, opened 2021-07-15
New Adapter: OpenWeb

GeneGenie:
Doc: https://github.com/prebid/prebid.github.io/pull/3101


adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type OpenWebAdapter struct {
```
SyntaxNode, line 15: This type name should be `adapter`. There is no need to repeat the package name and no reason to export it with an upper case letter. Please review our new adapter docs: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+
+type openwebImpExt struct {
+	OpenWeb openrtb_ext.ExtImpOpenWeb `json:"openweb"`
+}
+
+func (a *OpenWebAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+
+	totalImps := len(request.Imp)
+	errors := make([]error, 0, totalImps)
+	imp2source := make(map[int][]int)
```
SyntaxNode, line 27: Please consider a more descriptive name, such as `impBySourceID`.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+
+		reqs = append(reqs, &adapters.RequestData{
+			Method:  "POST",
+			Uri:     a.endpoint + fmt.Sprintf("?aid=%d", sourceId),
+			Body:    body,
+			Headers: headers,
+		})
+	}
+
+	if 0 == len(reqs) {
```
SyntaxNode, line 80: I don't think this needs to be s special case. `return reqs, errors` should work fine for an empty `reqs` slice.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (a *OpenWebAdapter) MakeBids(bidReq *openrtb2.BidRequest, unused *adapters.RequestData, httpRes *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+
+	if httpRes.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	var bidResp openrtb2.BidResponse
+	if err := json.Unmarshal(httpRes.Body, &bidResp); err != nil {
```
SyntaxNode, line 95: This will return an error if there is no body, as may be the case for other error types. Should this behavior be scoped to just 4xx errors or 5xx errors? What errors codes do you expect your server to return a body for?

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+	return bidResponse, errors
+}
+
+func validateImpression(imp *openrtb2.Imp) (int, error) {
+
+	if imp.Banner == nil && imp.Video == nil {
+		return 0, &errortypes.BadInput{
+			Message: fmt.Sprintf("ignoring imp id=%s, OpenWeb supports only Video and Banner", imp.ID),
+		}
+	}
```
SyntaxNode, line 147: Please remove this check. It is already performed by the core logic per your bidder info yaml file.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+
+	if 0 == len(imp.Ext) {
+		return 0, &errortypes.BadInput{
+			Message: fmt.Sprintf("ignoring imp id=%s, extImpBidder is empty", imp.ID),
+		}
+	}
```
SyntaxNode, line 153: Please remove this check. It is already performed by the core logic per your bidder json schema.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+
+	impExt := openrtb_ext.ExtImpOpenWeb{}
+	err := json.Unmarshal(bidderExt.Bidder, &impExt)
+	if err != nil {
+		return 0, &errortypes.BadInput{
+			Message: fmt.Sprintf("ignoring imp id=%s, error while decoding impExt, err: %s", imp.ID, err),
+		}
+	}
+
+	// common extension for all impressions
```
SyntaxNode, line 171: What does this comment mean? This code is running for specific impressions and is not common logic.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+	var impExtBuffer []byte
+
+	impExtBuffer, err = json.Marshal(&openwebImpExt{
+		OpenWeb: impExt,
+	})
```
SyntaxNode, line 176: Please do not ignore errors. Add handling logic for the err generated by `json.Marshal`.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+	if err := json.Unmarshal(httpRes.Body, &bidResp); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("error while decoding response, err: %s", err),
+		}}
+	}
+
+	bidResponse := adapters.NewBidderResponse()
+	var errors []error
+
+	var impOK bool
```
SyntaxNode, line 104: IMHO it's confusing to have the impOK defined at this level when it's scoped to the `for i := 0; i < len(sb.Bid); i++ {` loop. Please consider moving down to that level.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,193 @@
...
+				if imp.ID == bid.ImpID {
+
+					impOK = true
+
+					if imp.Video != nil {
+						mediaType = openrtb_ext.BidTypeVideo
+						break
+					}
+				}
+			}
```
SyntaxNode, line 122: This is an anti-pattern as its often error prone and can cause problems with imp rendering. Does your bidding server respond with extension values indicating the bid type?

config/config.go
```diff
@@ -661,6 +661,7 @@ func (cfg *Configuration) setDerivedDefaults() {
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderNanoInteractive, "https://ad.audiencemanager.de/hbs/cookie_sync?gdpr={{.GDPR}}&consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redirectUri="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dnanointeractive%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderNinthDecimal, "https://rtb.ninthdecimal.com/xp/user-sync?acctid={aid}&&redirect="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dninthdecimal%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID")
 	setDefaultUsersync(cfg.Adapters, openrtb_ext.BidderNoBid, "https://ads.servenobid.com/getsync?tek=pbs&ver=1&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redirect="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dnobid%26uid%3D%24UID")
+	// openrtb_ext.BidderOpenWeb doesn't have a good default.
```
SyntaxNode, line 664: Why is the url in `TestOpenWebSyncer` not a good default?

openrtb_ext/imp_openweb.go
```diff
@@ -0,0 +1,9 @@
+package openrtb_ext
+
+// ExtImpOpenWeb defines the contract for bidrequest.imp[i].ext.openweb
+type ExtImpOpenWeb struct {
+	SourceId    int     `json:"aid"`
+	PlacementId int     `json:"placementId,omitempty"`
+	SiteId      int     `json:"siteId,omitempty"`
+	BidFloor    float64 `json:"bidFloor,omitempty"`
+}
```
SyntaxNode, line 9: Nitpick: It's preferred in Go to use `ID` instead of `Id`:

```
type ExtImpOpenWeb struct {
	SourceID    int     `json:"aid"`
	PlacementID int     `json:"placementId,omitempty"`
	SiteID      int     `json:"siteId,omitempty"`
	BidFloor    float64 `json:"bidFloor,omitempty"`
}
```

SyntaxNode:
@GeneGenie Friendly ping on feedback, and now there's also a merge conflict.

adapters/openweb/openweb.go
GeneGenie: I guess it's ok to skip it, since its value got from Unmarshall

adapters/openweb/openweb.go
GeneGenie: In most cases we do not have and adType extension, but out rtb server guarantees that response type will be matched with request type

config/config.go
GeneGenie: Each PBS   publisher in our agreements has its own sync id. So we provide it during contract signing. 
Maybe we can consider   specifying publisherId in bidRequest params **IF** we can specify such macro in default syncUrl

e.g: 
```
https://sync.spotim.market/?customerId={{.params.customerId}}&gdpr={{.GDPR}}&euconsent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&url="+url.QueryEscape(externalURL)+"%2Fsetuid%3Fbidder%3Dbrightroll%26gdpr%3D{{.GDPR}}%26gdpr_consent%3D{{.GDPRConsent}}%26uid%3D%24UID
```

adapters/openweb/openweb.go
GeneGenie: I guess it may happen if body is not a JSON string. Just additional safe check here

config/config.go
SyntaxNode: Gotcha. PBS does not support parameter macros in the user sync url and would most likely not know the publisher id. I'm fine with leaving it empty and having the host configure it. Please be aware, a change is going in soon which moves the user sync url config to a host parameter via either pbs.yaml or an environment variable.

adapters/openweb/openweb.go
SyntaxNode: What happens if an impression has multiple request types? Such as banner and video?

GeneGenie: Currently adserver works in such way that if video is present in Imp, then request is handled as video supply. So video is sort of  "prioritized" and banner is skipped completely. 
I see what you mean, but I can't change the response format right now =\
Though adserver changes are coming, and Adapter will be updated to conform adserver changes

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,179 @@
...
+		if _, ok := sourceIdToImpIds[sourceId]; !ok {
+			sourceIdToImpIds[sourceId] = make([]int, 0, totalImps-i)
+		}
+
+		sourceIdToImpIds[sourceId] = append(sourceIdToImpIds[sourceId], i)
+
+	}
+
+	totalReqs := len(sourceIdToImpIds)
+	if 0 == totalReqs {
```
mansinahar, line 47: **Nitpick:** Please consider writing this as `if totalReqs == 0` for ease of readability

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,179 @@
...
+			impOK := false
+			mediaType := openrtb_ext.BidTypeBanner
+			for _, imp := range bidReq.Imp {
+				if imp.ID == bid.ImpID {
+
+					impOK = true
+
+					if imp.Video != nil {
+						mediaType = openrtb_ext.BidTypeVideo
+						break
```
mansinahar, line 119: We should `break` out of this for loop even if `imp.Video != nil` is false because if you've gotten here that means you've found the corresponding imp for the bid. With the current logic, in case of banner imps you'd still be looping over all the imps even after you've found the imp you were looking for.

Also, I'd suggest. extracting this code for getting the bid type in its own function named something along the lines of `getBidType`

adapters/openweb/openwebtest/exemplary/simple-banner.json
```diff
@@ -0,0 +1,63 @@
...
+                  "aid": 1000,
+                  "siteId": 1234,
+                  "bidFloor": 20
+                }
+              }
+            }
+          ]
+        }
+      },
+      "mockResponse": {
```
mansinahar, line 57: Please add a valid mock response for this test and the `sample-video.json` test so that the `MakeBids` part of the code can be covered for the tests as well and if you've added the `media-type-mapping.json` to do that then you can remove that test after adding the mock response here and in `sample-video.json`.

adapters/openweb/openwebtest/supplemental/explicit-dimensions.json
```diff
@@ -0,0 +1,59 @@
+{
+  "mockBidRequest": {
+    "id": "test-request-id",
+    "imp": [
+      {
```
mansinahar, line 5: I am not sure I understand what the name of this test `explicit-dimensions` mean.. Can you please elaborate on what exact functionality are you trying to test with this test case?

adapters/openweb/openwebtest/supplemental/wrong-impression-ext.json
```diff
@@ -0,0 +1,26 @@
...
+    "imp": [
+      {
+        "id": "unsupported-native-imp",
+        "video": {
+         "w": 100,
+          "h": 200
+        },
+        "ext": {
+          "bidder": {
+            "aid": "some string instead of int"
```
mansinahar, line 13: You don't need a JSON test for invalid param values. That should  be covered in `params_test.go`

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	var impExtBuffer []byte
+
+	impExtBuffer, _ = json.Marshal(&openwebImpExt{
+		OpenWeb: impExt,
+	})
+
+	if impExt.BidFloor > 0 {
+		imp.BidFloor = impExt.BidFloor
```
mansinahar, line 165: Please add or modify existing test case where both imp and imp ext has a bidfloor set and the imp bid floor gets overwritten by the one in imp ext

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,179 @@
...
+
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json;charset=utf-8")
+	headers.Add("Accept", "application/json")
+
+	reqs := make([]*adapters.RequestData, 0, totalReqs)
+
+	imps := request.Imp
+	request.Imp = make([]openrtb2.Imp, 0, len(imps))
+	for sourceId, impIds := range sourceIdToImpIds {
```
mansinahar, line 59: Please add a JSON test for this logic where a request has multiple imps with multiple source IDs and they are properly grouped together as requests for your bidder service

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (a *adapter) MakeBids(bidReq *openrtb2.BidRequest, unused *adapters.RequestData, httpRes *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+
+	if httpRes.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if httpRes.StatusCode != http.StatusOK {
+		return nil, []error{&errortypes.BadServerResponse{
```
mansinahar, line 91: Please also add JSON test cases for 4xx/5xx responses

adapters/openweb/openweb.go
SyntaxNode: Ok. So long as this code is preferring video over banner to match the server behavior, then we're fine for now. The media type will be included in responses for OpenRTB 2.6 and will be the long term solution to this problem.

mansinahar:
@GeneGenie Just a friendly reminder about the above comments

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,186 @@
...
+			sourceIds = append(sourceIds, sourceId)
+		}
+
+		sourceIdToImpIds[sourceId] = append(sourceIdToImpIds[sourceId], i)
+
+	}
+
+	totalReqs := len(sourceIdToImpIds)
+	if totalReqs == 0 {
+		return nil, errors
```
SyntaxNode, line 49: Please add a json test case for this situation.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	return reqs, errors
+
+}
+
+func (a *adapter) MakeBids(bidReq *openrtb2.BidRequest, unused *adapters.RequestData, httpRes *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+
+	if httpRes.StatusCode == http.StatusNoContent {
+		return nil, nil
```
SyntaxNode, line 90: Please add a test case for a 204 response.

adapters/openweb/openweb.go
SyntaxNode: While I agree it's unlikely to produce error, we've seen strange things happen before - sometimes due to memory corruption. If the error here is handled it aids with debugging.

adapters/openweb/openweb.go
```diff
@@ -0,0 +1,186 @@
...
+
+	imps := request.Imp
+	reqCopy := *request
+	reqCopy.Imp = make([]openrtb2.Imp, totalImps)
+	for _, sourceId := range sourceIds {
+		impIds := sourceIdToImpIds[sourceId]
+		reqCopy.Imp = reqCopy.Imp[:0]
+
+		for i := 0; i < len(impIds); i++ {
+			reqCopy.Imp = append(reqCopy.Imp, imps[impIds[i]])
```
SyntaxNode, line 66: Could you please add a json test where there are two impressions with the same aid to ensure this loop works correctly for n>1?

static/bidder-info/openweb.yaml
```diff
@@ -0,0 +1,12 @@
...
+gvlVendorID: 280
+capabilities:
+  app:
+    mediaTypes:
+      - banner
+      - video
+  site:
+    mediaTypes:
+      - banner
+      - video
```
SyntaxNode, line 12: We've made changes to how user syncs are configured since you first opened this PR. If OpenWeb supports user syncing and doesn't have a good default, please add the following to the end of this yaml file. Replace 'redirect' with 'iframe' if appropriate.

```
userSync:
  # openweb supports user syncing, but requires configuration by the host. contact this
  # bidder directly at the email address in this file to ask about enabling user sync.
  supports:
    - redirect
```

SyntaxNode (approved):

guscarreon (approved):
LGTM

PR #1928, opened 2021-07-15
InMobi: adding native support

cormorinth:
Accompanying docs PR: https://github.com/prebid/prebid.github.io/pull/3109


adapters/inmobi/inmobitest/exemplary/simple-app-native.json
```diff
@@ -0,0 +1,105 @@
...
+        "nurl": "https://some.event.url/params",
+        "ext": {
+          "prebid": {
+            "meta": {
+              "networkName": "inmobi"
+            }
+          }
+        }
+      },
+      "type": "banner"
```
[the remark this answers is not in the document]
cormorinth: Should be fixed. Apologies, was on vacation.

bsardo (approved):

adapters/inmobi/inmobi.go
```diff
@@ -124,6 +124,9 @@ func getMediaTypeForImp(impId string, imps []openrtb2.Imp) openrtb_ext.BidType {
 			if imp.Video != nil {
 				mediaType = openrtb_ext.BidTypeVideo
 			}
+			if imp.Native != nil {
```
mansinahar, line 127: Just a call out since this is logic was already present in your adapter code: Since your adapter is determining the bid type based on what object of the imp is set i.e `banner`, `video` or `native`, this function might return the wrong media type if the imp was a multi-format imp i.e an imp that has both `banner` and `video` set, for example.

static/bidder-info/inmobi.yaml
```diff
@@ -6,6 +6,7 @@ capabilities:
     mediaTypes:
       - banner
       - video
+      - native
```
mansinahar, line 9: Just to confirm, you want the native support only for `app` and not for `site`, right?

adapters/inmobi/inmobi.go
cormorinth: This should be fine. Will likely look into this in the future though (and stop relying on the bid request passthrough.)

static/bidder-info/inmobi.yaml
cormorinth: Correct.

mansinahar (approved):

PR #1930, opened 2021-07-16
Smaato: Split multiple media types

el-chuck:
Split multiple media types into individual requests


AlexBVolcy (approved):
Looks good to me!

adapters/smaato/smaato.go
```diff
@@ -436,7 +465,7 @@ func setImpForAdspace(imp *openrtb2.Imp) error {
 		return nil
 	}
 
-	return &errortypes.BadInput{Message: "Invalid MediaType. Smaato only supports Banner and Video."}
+	return nil
```
guscarreon, line 468: Newly created `splitImpressionsByMediaType(...)` function and the `return nil` statements in lines 459 and 465 make this line unreachable. It doesn't affect anything and I don't have a problem with it so, if you like to fix it in this PR, ok, but I will approve your PR either way. Your code looks pretty solid.

guscarreon (approved):
I left a comment in regards a small nitpick but it's so small that it's up to you @el-chuck if you think it's worth addressing or not. These modifications look good to me.

Approved

adapters/smaato/smaato.go
el-chuck: ok, we will address this with next PR

guscarreon: Awesome @el-chuck . Thank you for addressing our feedback

PR #1931, opened 2021-07-16
Tappx: new bidder params

prebidtappx:
In this pull request we are going to add the next new params:

- mktag
- bcid
- bcrid
- rewarded
(At the same time we are going to send the tappxkey in the ext/bidder to unify all the requests json received in our system)


static/bidder-params/tappx.json
```diff
@@ -12,13 +12,29 @@
...
 		"endpoint": {
 			"type": "string",
 			"description": "Endpoint provided to publisher"
 		},
 		"bidfloor": {
 			"type": "number",
 			"description": "Minimum bid for this impression expressed in CPM (USD)"
+		},
+		"rewarded": {
+			"type": "number",
```
SyntaxNode, line 28: Please consider changing the type to an integer, as a number can also be floating point or scientific notation. I also recommend adding a min and max constraint in the schema:
```
  "minimum": 0,
  "maximum": 1,
```

adapters/tappx/params_test.go
```diff
@@ -60,4 +60,7 @@ var invalidParams = []string{
 	`{"tappxkey": 1, "endpoint":"ZZ1INTERNALTEST149147915", "host":""}`,
 	`{"tappxkey":"pub-12345-android-9876", "endpoint": 1, "host":""}`,
 	`{"tappxkey": 1, "endpoint": 1, "host": 123}`,
+	`{"tappxkey": 1, "endpoint": 1, "rewarded": []}`,
+	`{"tappxkey": 1, "endpoint": 1, "rewarded": 1, "bcid":""}`,
+	`{"tappxkey": 1, "endpoint": 1, "bcid":["123"], bcrid: 123}`,
 }
```
SyntaxNode, line 66: Please add valid test cases too.

adapters/tappx/tappx.go
```diff
@@ -51,14 +51,58 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
 	if err := json.Unmarshal(bidderExt.Bidder, &tappxExt); err != nil {
 		return nil, []error{&errortypes.BadInput{
 			Message: "Error parsing tappxExt parameters",
 		}}
 	}
 
+	//In this section we create all EXT parameters
+	type VideoExt struct {
+		Rewarded int `json:"rewarded,omitempty"`
+	}
```
SyntaxNode, line 64: This structure is unused.

adapters/tappx/tappx.go
```diff
@@ -51,14 +51,58 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+	if tappxExt.Rewarded == 1 {
+		request.Imp[0].Video.Ext = json.RawMessage(`{"rewarded": 1}`)
+	}
```
SyntaxNode, line 67: I'm personally fine with this approach, but please understand this will overwrite all publisher provided ext values in the video object.

SyntaxNode, line 67: This new conditional requires json test coverage. You'll then see that the Video object is shared memory and must be copied first before making a change.

adapters/tappx/tappx.go
```diff
@@ -51,14 +51,58 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+	if tappxExt.Mktag != "" {
+		ext.Bidder.Mktag = tappxExt.Mktag
+	}
+	if tappxExt.Bcid != nil {
+		ext.Bidder.Bcid = tappxExt.Bcid
+	}
+	if tappxExt.Bcrid != nil {
+		ext.Bidder.Bcrid = tappxExt.Bcrid
+	}
```
SyntaxNode, line 91: The if statements have no effect, since the tested values are all the default values. Consider changing to:
```
ext := Ext{
		Bidder: Bidder{
			Tappxkey: tappxExt.TappxKey,
			Mktag: tappxExt.Mktag,
			Bcid: tappxExt.Bcid,
			Bcrid: tappxExt.Bcrid,
		},
	}
```

adapters/tappx/tappx.go
```diff
@@ -51,14 +51,58 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+
+	request.Ext = nil
+
+	jsonext, err := json.Marshal(ext)
+	if err == nil {
+		request.Ext = jsonext
+	}
```
SyntaxNode, line 98: What if there is an error during marshaling? Would that still generate a valid Tappx request?

adapters/tappx/tappx.go
```diff
@@ -51,14 +51,58 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+
+	if err := json.Unmarshal(bidderExt.Bidder, &tappxExt); err != nil {
+		return nil, []error{&errortypes.BadInput{
+			Message: "Error parsing tappxExt parameters",
+		}}
+	}
```
SyntaxNode, line 104: Why are you unmarshalling the tappxExt again?

adapters/tappx/tappxtest/exemplary/single-banner-impression-extra.json
```diff
@@ -0,0 +1,124 @@
...
+  "expectedBidResponses": [
+    {
+      "currency": "USD",
+      "bids": [
+        {
+          "bid": {
+            "id": "wehM-93KGr0_0_0",
+            "impid": "adunit-1",
+            "price": 0.5,
+            "adm": "<!-- admarkup -->",
```
SyntaxNode, line 110: Thank you for using a fake ad markup value for tests.

adapters/tappx/tappx.go
prebidtappx: Hi! A lot of thanks of this point. We now find useless to use the rewarded tappx param. We think that is better to tell to our publishers to send in video ext param directly like in all our oRTB connections. 😅

We clean all this in **a2ddc1f** commit: https://github.com/prebid/prebid-server/pull/1931/commits/a2ddc1f91aca8451a19aa4e651095b91338881dc

prebidtappx: Hi! A lot of thanks of this point. We now find useless to use the rewarded tappx param. We think that is better to tell to our publishers to send in video ext param directly like in all our oRTB connections. 😅

We clean all this in a2ddc1f commit.

adapters/tappx/params_test.go
prebidtappx: Hi! Any problem! 😉

Done in 4cf8eb0 commit.

static/bidder-params/tappx.json
prebidtappx: Hi! We resolve this in other comment. We now find useless to use the rewarded tappx param. We think that is better to tell to our publishers to send in video ext param directly like in all our oRTB connections. If the data is correct or incorrect we check in our side! 

We clean all this in a2ddc1f commit.

adapters/tappx/tappx.go
prebidtappx: Hi! We resolve this in other comment. We now find useless to use the rewarded tappx param. 

We clean all this in a2ddc1f commit.

adapters/tappx/tappx.go
prebidtappx: We consider to change... And make the change! 😉
A lot of thanks for the help! 😊

Change in commit 89cfb84.

adapters/tappx/tappx.go
prebidtappx: Sorry!!!! It was a code that I forgot to delete when move the unmashalling up. Sorry!!! 😥

I delete this code in the 1208cdf commit.

adapters/tappx/tappx.go
prebidtappx: Now yes, but in the future not.
Also, it is a good point to prevent now to the future.
Thank you!!!

We repair this in the 06a898d commit.

adapters/tappx/tappxtest/exemplary/single-banner-impression-extra.json
prebidtappx: Thank you for all the help and detail attention! 🤗

adapters/tappx/tappxtest/exemplary/single-banner-impression-extra.json
```diff
@@ -0,0 +1,124 @@
...
+        "banner": {
+          "w": 300,
+          "h": 250
+        },
+        "ext": {
+         "bidder": {
+            "tappxkey": "pub-12345-android-9876",
+            "endpoint": "ZZ123456PS",
+            "host": "testing.ssp.tappx.com/rtb/v2/",
+            "mktag": "mktag-12345-android-9876"
```
bsardo, line 17: Can you please either update this test with `bcid` and `bcrid` or add a new test with them? I don't see these fields in any of the tests.

adapters/tappx/params_test.go
```diff
@@ -60,4 +65,11 @@ var invalidParams = []string{
 	`{"tappxkey": 1, "endpoint":"ZZ1INTERNALTEST149147915", "host":""}`,
 	`{"tappxkey":"pub-12345-android-9876", "endpoint": 1, "host":""}`,
 	`{"tappxkey": 1, "endpoint": 1, "host": 123}`,
+	`{"tappxkey": "1", "endpoint": 1}`,
+	`{"tappxkey": "1", "endpoint": 1}`,
+	`{"tappxkey": "1", "endpoint": "1", "mktag":1}`,
+	`{"tappxkey": "1", "endpoint": "1", "mktag":[1,2]}`,
+	`{"tappxkey": "1", "endpoint": "1", "host":"host", "bcid":""}`,
+	`{"tappxkey": "1", "endpoint": "1", "host":"host", "bcid":["123"], bcrid: 123}`,
+	`{"tappxkey": "1", "endpoint": "1", "host":"host", "bcid":["123"], bcrid: 123}`,
```
bsardo, line 74: Please delete. This line appears twice.

adapters/tappx/tappx.go
```diff
@@ -51,14 +51,40 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
 
+	type Bidder struct {
+		Tappxkey string   `json:"tappxkey"`
+		Mktag    string   `json:"mktag,omitempty"`
+		Bcid     []string `json:"bcid,omitempty"`
+		Bcrid    []string `json:"bcrid,omitempty"`
+	}
+	type Ext struct {
+		Bidder `json:"bidder"`
+	}
```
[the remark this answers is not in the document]
prebidtappx: Ok.
Done in 3c1a308 commit.

static/bidder-params/tappx.json
```diff
@@ -12,13 +12,29 @@
...
+			"description": "Indicates with a 0 or 1 if is rewarded or no. (Only with video)"
+		},
+		"bcid": {
+			"type": "array",
+			"description": "Block list of CID"
+		},
+		"bcrid": {
+			"type": "array",
+			"description": "Block list of CRID"
```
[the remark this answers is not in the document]
prebidtappx: Ok, any problem.
Done in b1a6fc6 commit.

(Extra: We remove rewarded property that was discarted in past commits)

adapters/tappx/tappx.go
```diff
@@ -51,14 +51,40 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+	ext := Ext{
+		Bidder: Bidder{
+			Tappxkey: tappxExt.TappxKey,
```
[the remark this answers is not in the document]
prebidtappx: Yes we know. We want in the first place (1) because is the standard place for this parameter in our side and we are going to don't accept the request that send in other place. We still sending the (2) place because if we need to know or analyze what are sending our partners in the bidder of prebid.
If is not important, we prefer now to have the tappxkey in both places.

adapters/tappx/params_test.go
prebidtappx: Yeep, sorry, solve error and improve the params_test a little in commit 2b45efd.

adapters/tappx/tappxtest/exemplary/single-banner-impression-extra.json
prebidtappx: Ok, we can add the params to the new test! 👍 
We add in the commit 09ad06d.

bsardo (dismissed):
LGTM

adapters/tappx/tappx.go
```diff
@@ -51,14 +62,31 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+		},
+	}
+
+	jsonext, err := json.Marshal(ext)
+	if err == nil {
+		request.Ext = jsonext
+	} else {
+		return nil, []error{&errortypes.BadInput{
+			Message: "Error with tappxExt parameters",
+		}}
```
SyntaxNode, line 87: This is an error which should never occur, but might sometimes if there is a memory corruption issue. I wouldn't classify this as BadInput. I recommend to just use a normal error which PBS-Core will classify as a fatal error and warn the host of a problem.

Something like:
```
return nil, []error{fmt.Errorf("Error writing TappX request.ext.")
```

SyntaxNode (commented):
Looks good. I left one minor comment. Before we can merge this is in, please open a docs PR to update https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/tappx.md.

adapters/tappx/tappx.go
prebidtappx: About change de docs. I open the pull request e88cf1c.
About the BadInput error I make the change in the e88cf1c commit.

bsardo (dismissed):

adapters/tappx/params_test.go
```diff
@@ -60,4 +65,11 @@ var invalidParams = []string{
 	`{"tappxkey": 1, "endpoint":"ZZ1INTERNALTEST149147915", "host":""}`,
 	`{"tappxkey":"pub-12345-android-9876", "endpoint": 1, "host":""}`,
 	`{"tappxkey": 1, "endpoint": 1, "host": 123}`,
+	`{"tappxkey": "1", "endpoint": 1}`,
+	`{"tappxkey": "1", "endpoint": 1}`,
+	`{"tappxkey": "1", "endpoint": "1", "mktag":1}`,
+	`{"tappxkey": "1", "endpoint": "1", "mktag":[1,2]}`,
+	`{"tappxkey": "1", "endpoint": "1", "host":"host", "bcid":""}`,
+	`{"tappxkey": "1", "endpoint": "1", "host":"host", "bcid":"123", bcrid: ["123"]}`,
+	`{"tappxkey": "1", "endpoint": "1", "host":"host", "bcid":["123"], bcrid: 123}`,
```
SyntaxNode, line 74: Could you please add invalid test cases for `bcid` and `bcrid` where the array item type is invalid? Something like: .`[123]`. This will ensure json schema is property enforcing the array type constraint.

adapters/tappx/tappx.go
```diff
@@ -51,14 +62,31 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+			Bcrid:    tappxExt.Bcrid,
+		},
+	}
+
+	jsonext, err := json.Marshal(ext)
+	if err == nil {
+		request.Ext = jsonext
+	} else {
+		return nil, []error{
+			fmt.Errorf("Error writing TappX request.ext."),
```
SyntaxNode, line 86: This error doesn't contain a parameter, so please use `errors.New` rather than `fmt.Errorf`.

adapters/tappx/tappx.go
```diff
@@ -51,14 +62,31 @@ func (a *TappxAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+		}
```
SyntaxNode, line 87: Nitpick: This may be written slightly more compact with something like:
```
if jsonext, err := json.Marshal(ext); err == nil {
  request.Ext = jsonext
} else {
  return nil, []error{errors.New("Error writing TappX request.ext.")}
}
```

.. and thank you for including full error checking with this statement. I love to see the marshal result only used if there is no error.

adapters/tappx/params_test.go
prebidtappx: Done in 9e3aff2 and 9636101 commits.

adapters/tappx/tappx.go
prebidtappx: Done in c140289, solving error in 56c585f.

adapters/tappx/tappx.go
prebidtappx: Done in c140289, solving error in 56c585f.

adapters/tappx/params_test.go
```diff
@@ -60,4 +65,13 @@ var invalidParams = []string{
 	`{"tappxkey": 1, "endpoint":"ZZ1INTERNALTEST149147915", "host":""}`,
 	`{"tappxkey":"pub-12345-android-9876", "endpoint": 1, "host":""}`,
 	`{"tappxkey": 1, "endpoint": 1, "host": 123}`,
+	`{"tappxkey": "1", "endpoint": 1}`,
+	`{"tappxkey": "1", "endpoint": 1}`,
```
bsardo, line 69: Please delete. This is a duplicate of the case above.

adapters/tappx/params_test.go
```diff
@@ -60,4 +65,13 @@ var invalidParams = []string{
...
+	`{"tappxkey": "1", "endpoint": "1", "mktag":1}`,
+	`{"tappxkey": "1", "endpoint": "1", "mktag":[1,2]}`,
```
bsardo, line 71: Can you add `"host":"host"` to each of these two test cases where `mktag` is an invalid type to isolate that it is `mktag` that is causing the error and not a missing host?

adapters/tappx/params_test.go
prebidtappx: I modified to add more cases in 8d09494 and now is not duplicated!
Thanks for all.

adapters/tappx/params_test.go
prebidtappx: Yeep, thanks!
Change in 8d09494.

SyntaxNode (approved):

bsardo (approved):
LGTM

PR #1934, opened 2021-07-21
IX: update required site id field to be more flexible

grossjo:


guscarreon (changes requested):
Can you please add test cases to assert for the correct validation of the json fields? Other adapters include these test in a `params_test.go` file but your adapter is missing its corresponding `adapters/ix/params_test.go`.

grossjo:
@guscarreon tests were added.

adapters/ix/params_test.go
```diff
@@ -0,0 +1,54 @@
...
+	`{"siteid":"1234"}`,
+	`{"siteID":"12345"}`,
+	`{"siteId":"123456"}`,
+	`{"siteid":"1234567", "size": [640,480]}`,
+}
+
+var invalidParams = []string{
+	`{"siteid":123}`,
+	`{"siteids":"123"}`,
+	`{"notaparam":"123"}`,
```
bsardo, line 46: Can you add the following test cases here to verify the min length and oneOf required configuration:
```
`{"siteid":""}`,
`{"siteID":""}`,
`{"siteId":""}`,
`{"siteid":"1234", "siteID":"12345"}`,
`{"siteid":"1234", "siteId":"123456"}`,
```

static/bidder-params/ix.json
```diff
@@ -4,11 +4,21 @@
...
     "siteId": {
       "type": "string",
       "minLength": 1,
       "description": "An ID which identifies the site selling the impression"
     },
+    "siteID": {
+      "type": "string",
+      "minLength": 1,
+      "description": "An ID which identifies the site selling the impression"
+    },
```
SyntaxNode, line 21: Some hosting companies, such as Xandr, provide a user interface to publishers based on this json schema. To help the UI understand which field to ask the publisher to provide in a case such as this (multiple fields differing only by case) can you please add the word "preferred" to the description of the format you'd like to be the default?

Please see [Criteo's json schema](https://github.com/prebid/prebid-server/blob/master/static/bidder-params/criteo.json) as an example.

Thank you.

adapters/ix/params_test.go
grossjo: done.

static/bidder-params/ix.json
grossjo: done.

bsardo (approved):
LGTM

guscarreon (approved):
Looks good to me. @grossjo , thank you for addressing our comments

PR #1936, opened 2021-07-23
Axonix: enabled by default

cesarfd:
Axonix adapter is now live.

- Adapter is now enabled by default.
- Endpoint is built using macros.

### DIFF
--- a/adapters/axonix/axonix.go
+++ b/adapters/axonix/axonix.go
@@ -32,7 +32,7 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
 
 func (a *adapter) getEndPoint(ext *openrtb_ext.ExtImpAxonix) (string, error) {
 	endpointParams := macros.EndpointTemplateParams{
-		ZoneID:    url.PathEscape(ext.Region),
+		Host:      url.PathEscape(ext.Host),
 		AccountID: url.PathEscape(ext.SupplyId),
 	}
 	return macros.ResolveMacros(a.EndpointTemplate, endpointParams)
--- a/adapters/axonix/axonix.go
+++ b/adapters/axonix/axonix.go
@@ -30,7 +30,7 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
 	return bidder, nil
 }
 
-func (a *adapter) getEndPoint(ext *openrtb_ext.ExtImpAxonix) (string, error) {
+func (a *adapter) getEndpoint(ext *openrtb_ext.ExtImpAxonix) (string, error) {
 	endpointParams := macros.EndpointTemplateParams{
 		AccountID: url.PathEscape(ext.SupplyId),
 	}
@@ -58,7 +58,7 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 		return nil, errors
 	}
 
-	endpoint, err := a.getEndPoint(&axonixExt)
+	endpoint, err := a.getEndpoint(&axonixExt)
 	if err != nil {
 		return nil, errors
 	}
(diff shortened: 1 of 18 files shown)


static/bidder-params/axonix.json
```diff
@@ -4,11 +4,16 @@
   "description": "A schema which validates params accepted by the Axonix adapter",
   "type": "object",
   "properties": {
+    "region": {
+      "type": "string",
+      "minLength": 1,
+      "description": "Datacenter zone"
+    },
```
SyntaxNode, line 11: I'm curios why you are adding region to the bidder parameters. The common solution on the server side is to instruct Prebid Server hosts via [your bidder docs](https://github.com/prebid/prebid.github.io/blob/f3823a982fce777b964c279414a461e315462100/dev-docs/bidders/axonix.md) to set different endpoints in their regions to ensure the shorted routing path. This is different than the approach from Prebid.js.

cesarfd: Yeah... I am kind of lost with this tbh. 

Looking at the `EndpointTemplateParams` struct and how some other adapters use it I figured out the usual way is to set a template endpoint and fill the macros with whatever variable parts the endpoint might need. Kind of like [algorix](https://github.com/prebid/prebid-server/blob/master/config/config.go#L871).

But the values of these macros come from the bidder parameters, right? Should I use a totally different approach?

cesarfd: I've simplified how the endpoint is resolved using `Host` instead of region as it is more flexible.

SyntaxNode: Sorry the delayed response. I've been busy with the large amount of PRs open and with summer holidays.

You have the correct approach for using macros, however macros should not be used in the host for geo regions and the entire host may not be supplied as a bidder parameter. That's an anti-pattern which we as a Prebid Server committee strongly oppose. It's odd to ask the publisher to specify the server address of the bidder, it greatly harms Prebid Server's performance which is based on keep alive connections to bidders, and allowing the publisher to pass the host through opens up PBS to a variety of attacks.

I am aware of some bidders doing this today. We're in talks with them to correct their behavior and we're focused on preventing this design from propagating further.

I recommend using a default endpoint pointing to your company's global load balancer. If you don't have a global load balancer, consider marking the adapter as disabled by default and add instructions in your bidder docs to let hosts know which endpoints to use for their geo regions, or ask them to contact you directly to setup if you don't want to publicly post those endpoints. This is the preferred approach for geo region configurations.

cesarfd: Thank you for the explanation, I've changed it so that the only the path uses macros.

adapters/axonix/axonix.go
```diff
@@ -4,25 +4,39 @@ import (
...
+		return nil, fmt.Errorf("unable to parse endpoint url template: %v", err)
+	}
 	bidder := &adapter{
-		URI: config.Endpoint,
+		EndpointTemplate: *endpoint,
 	}
 	return bidder, nil
 }
 
+func (a *adapter) getEndPoint(ext *openrtb_ext.ExtImpAxonix) (string, error) {
```
mansinahar, line 33: **Nitpick:** `P` in `getEndPoint` doesn't need to be capital. Please consider changing it to a lower-case `p`

mansinahar (commented):
LGTM other than the nitpick I pointed out.

adapters/axonix/axonix.go
```diff
@@ -44,9 +58,9 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 		return nil, errors
 	}
 
-	thisURI := a.URI
-	if len(thisURI) == 0 {
-		thisURI = "https://openrtb-us-east-1.axonix.com/supply/prebid-server/" + axonixExt.SupplyId
+	endpoint, err := a.getEndpoint(&axonixExt)
+	if err != nil {
+		return nil, errors
```
SyntaxNode, line 63: Please add the `err` from this call in the `errors` slice before returning it:

```
endpoint, err := a.getEndpoint(&axonixExt)
if err != nil {
	errors = append(errors, err)
	return nil, errors
}
```

SyntaxNode (approved):

mansinahar (approved):

Issue #1507, opened 2020-09-21
bidrequest.id is not always unique in PBS requests


epouradier:
A publisher partner who is using Prebid Server for their AMP inventory is sending us bid requests with a static value of BidRequest.id for all ad opportunities, while this field should be populated with a unique ID as per openRTB specifications.

Key takeaway from the initial discussion, from @bretg:

1. Add a configuration to overwrite any bidRequest.id set in a storedrequest or add one.

```
settings:
  generate-storedrequest-bidrequest-id: true
```

2. if the config is on and the stored request is for AMP or called from ext.prebid.storedrequest
2a. If the storedrequest contains an "id" field, generate a UUID and replace the value
2b. If the storedrequest doesn't contain an "id" field, generate a UUID and add it
3. Configuration should be at the host-company level.
4. Default should be to not overwrite ids

bretg:
- Refined the proposed algorithm
- Removed the per-account config. I don't see this as something a publisher is going to care to control.
- The default is still for the feature to be turned off so we don't risk a breaking change. But overall this seems like something not likely to break PBS functionality.

hhhjort:
So to be clear, this is only for requests coming into the AMP endpoint. Do we want to extend this functionality to `request.imp[n].id`, since these also won't be unique? Also if we ever want to extend this to non-AMP, I think we should never overwrite an ID that appears in the incoming request.

epouradier:
Per openRTB specs, for request.imp.id is
> A unique identifier for this impression within the context of
> the bid request (typically, starts with 1 and increments).
So as soon as it is unique **in the context of the bid request**, it should be ok.

bretg:
Discussed in committee.

Another option is to support a macro in the ID field. e.g. { id: "{{UUID}}" }. If Prebid Server sees {{UUID}} then resolve to a UUID.

We generally agreed that this is a cleaner approach, but it could require host companies to update thousands of database entries and the code that manages those entries.

So we agreed to move forward with the "override" setting noted above as an initial solution but to also support {{UUID}} in the storedrequest top level ID so we can someday deprecate the override.

bretg:
> So to be clear, this is only for requests coming into the AMP endpoint

Not quite true -- this is for AMP and App. Point number 2 says "if the config is on and the stored request is for AMP or called from ext.prebid.storedrequest" -- the latter clause covers mobile app.

bretg:
[...]

PR #1938, opened 2021-07-26
Generate Unique Bidrequest IDs for Amp Auctions

AlexBVolcy:
Addressing this issue here https://github.com/prebid/prebid-server/issues/1507

I've added a configuration boolean flag, `GenerateRequestId`, which when set to true, will generate a uuid for auction bid requests, which can be used to solve the amp auction issue found here https://github.com/prebid/prebid-server/issues/1507

This generation can occur in two different functions:
1. In amp_auction.go/loadRequestJSONForAmp( ) -- where we load the stored openRTB request for an incoming AMP request. Here we can check our boolean flag, and generate an ID if necessary.
2. In auction.go/processStoredRequests() -- we check here if `GenerateRequestId` is true, and if so, we generate a uuid and patch it into the request

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -31,6 +31,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
 	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/util/uuidutil"
 )
 
 const defaultAmpRequestTimeoutMillis = 900
@@ -45,6 +46,7 @@ type AmpResponse struct {
 // NewAmpEndpoint modifies the OpenRTB endpoint to handle AMP requests. This will basically modify the parsing
 // of the request, and the return value, using the OpenRTB machinery to handle everything in between.
 func NewAmpEndpoint(
+	uuidGenerator uuidutil.UUIDGenerator,
 	ex exchange.Exchange,
 	validator openrtb_ext.BidderParamValidator,
 	requestsById stored_requests.Fetcher,
@@ -69,6 +71,7 @@ func NewAmpEndpoint(
 	}
 
 	return httprouter.Handle((&endpointDeps{
+		uuidGenerator,
 		ex,
 		validator,
 		requestsById,
@@ -343,6 +346,15 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req 
 	if err := json.Unmarshal(requestJSON, req); err != nil {
 		errs = []error{err}
 		return
+	}
+
+	if deps.cfg.GenerateRequestID {
+		newBidRequestId, err := deps.uuidGenerator.Generate()
+		if err != nil {
+			errs = []error{err}
+			return
+		}
+		req.ID = newBidRequestId
 	}
 
 	if ampParams.Debug {
--- a/endpoints/openrtb2/amp_auction_test.go
+++ b/endpoints/openrtb2/amp_auction_test.go
@@ -11,6 +11,8 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/julienschmidt/httprouter"
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
 	"github.com/prebid/prebid-server/analytics"
 	analyticsConf "github.com/prebid/prebid-server/analytics/config"
 	"github.com/prebid/prebid-server/config"
@@ -19,7 +21,6 @@ import (
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/stored_requests/backends/empty_fetcher"
 
-	"github.com/mxmCherry/openrtb/v15/openrtb2"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -38,6 +39,7 @@ func TestGoodAmpRequests(t *testing.T) {
 	}
 
 	endpoint, _ := NewAmpEndpoint(
+		fakeUUIDGenerator{},
 		&mockAmpExchange{},
 		newParamsValidator(t),
 		&mockAmpStoredReqFetcher{goodRequests},
@@ -91,6 +93,7 @@ func TestAMPPageInfo(t *testing.T) {
 	exchange := &mockAmpExchange{}
 
 	endpoint, _ := NewAmpEndpoint(
+		fakeUUIDGenerator{},
 		exchange,
 		newParamsValidator(t),
 		&mockAmpStoredReqFetcher{stored},
@@ -188,6 +191,7 @@ func TestGDPRConsent(t *testing.T) {
 		// Build Exchange Endpoint
 		mockExchange := &mockAmpExchange{}
 		endpoint, _ := NewAmpEndpoint(
+			fakeUUIDGenerator{},
 			mockExchange,
 			newParamsValidator(t),
 			&mockAmpStoredReqFetcher{stored},
@@ -340,6 +344,7 @@ func TestCCPAConsent(t *testing.T) {
 		// Build Exchange Endpoint
 		mockExchange := &mockAmpExchange{}
 		endpoint, _ := NewAmpEndpoint(
+			fakeUUIDGenerator{},
 			mockExchange,
 			newParamsValidator(t),
 			&mockAmpStoredReqFetcher{stored},
@@ -450,6 +455,7 @@ func TestConsentWarnings(t *testing.T) {
 			mockExchange = &mockAmpExchange{}
 		}
 		endpoint, _ := NewAmpEndpoint(
+			fakeUUIDGenerator{},
 			mockExchange,
 			newParamsValidator(t),
 			&mockAmpStoredReqFetcher{stored},
@@ -542,6 +548,7 @@ func TestNewAndLegacyConsentBothProvided(t *testing.T) {
 		// Build Exchange Endpoint
 		mockExchange := &mockAmpExchange{}
 		endpoint, _ := NewAmpEndpoint(
+			fakeUUIDGenerator{},
 			mockExchange,
 			newParamsValidator(t),
 			&mockAmpStoredReqFetcher{stored},
@@ -593,6 +600,7 @@ func TestAMPSiteExt(t *testing.T) {
 	}
 	exchange := &mockAmpExchange{}
 	endpoint, _ := NewAmpEndpoint(
+		fakeUUIDGenerator{},
 		exchange,
 		newParamsValidator(t),
 		&mockAmpStoredReqFetcher{stored},
@@ -629,6 +637,7 @@ func TestAmpBadRequests(t *testing.T) {
 	}
 
 	endpoint, _ := NewAmpEndpoint(
+		fakeUUIDGenerator{},
 		&mockAmpExchange{},
 		newParamsValidator(t),
 		&mockAmpStoredReqFetcher{badRequests},
@@ -659,6 +668,7 @@ func TestAmpDebug(t *testing.T) {
 	}
 
 	endpoint, _ := NewAmpEndpoint(
+		fakeUUIDGenerator{},
 		&mockAmpExchange{},
 		newParamsValidator(t),
 		&mockAmpStoredReqFetcher{requests},
@@ -731,6 +741,7 @@ func TestQueryParamOverrides(t *testing.T) {
 	}
 
 	endpoint, _ := NewAmpEndpoint(
+		fakeUUIDGenerator{},
 		&mockAmpExchange{},
 		newParamsValidator(t),
 		&mockAmpStoredReqFetcher{requests},
@@ -883,6 +894,7 @@ func (s formatOverrideSpec) execute(t *testing.T) {
 	}
 
 	endpoint, _ := NewAmpEndpoint(
+		fakeUUIDGenerator{},
 		&mockAmpExchange{},
 		newParamsValidator(t),
 		&mockAmpStoredReqFetcher{requests},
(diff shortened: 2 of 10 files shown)


endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1297,6 +1275,81 @@ func TestBuildAmpObject(t *testing.T) {
 	}
 }
 
+func TestIdGeneration(t *testing.T) {
+	testCases := []struct {
+		description                  string
+		givenInTagId                 string
+		givenInStoredRequest         json.RawMessage
+		givenGenerateStoredRequestId bool
+		expected                     int
```
hhhjort, line 1284: Might want to call this `expectedIDLength` as at first I was wondering how you arranged it to have generated ID always be "36".

endpoints/openrtb2/auction.go
```diff
@@ -1448,7 +1449,13 @@ func getStoredRequestId(data []byte) (string, bool, error) {
 		return "", true, errors.New("ext.prebid.storedrequest.id must be a string")
 	}
 
-	return string(value), true, nil
+	// Generates Request ID if flag is set to true -- used to solve issue with amp auctions
```
hhhjort, line 1452: This is supposed to return the ID of the stored request, not the openrtb request's ID. Also by generating a random stored request ID you are virtually guaranteeing that the stored request lookup will fail, and potentially hammering the database with failing lookups.

hhhjort: To avoid having to unmarshal/remarshal the stored request json, you probably want to patch the retrieved stored request with `{  "id": "<generated uuid>" }` before applying the patch to the incoming request. Unless we want the spec that the flag should always overwrite any incoming request IDs, in which case we will have to do something else. But I don't see a use case for overwriting the request ID in the incoming request other than the case where a PBS instance is only supporting a client that always sends a non-random ID, and no other clients. But in this case it really should be fixed on the client side rather than hacking PBS to enforce unique IDs in all cases.

AlexBVolcy: Just to keep this PR updated, Hans and I talked about these comments and what I need to do to properly address them! A commit will be pushed today or tomorrow with the proper fixes / tests.

endpoints/openrtb2/auction.go
```diff
@@ -1340,21 +1340,38 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+		if err != nil {
+			requestErrorJson, errL := storedRequestErrorChecker(err, requestJson, storedRequests, storedBidRequestId)
+			return requestErrorJson, nil, errL
+		}
+		resolvedRequest, err = jsonpatch.MergePatch(requestJson, storedReqForPatch)
+		if err != nil {
+			requestErrorJson, errL := storedRequestErrorChecker(err, requestJson, storedRequests, storedBidRequestId)
+			return requestErrorJson, nil, errL
+		}
+	} else if !deps.cfg.GenerateRequestId && hasStoredBidRequest {
```
hhhjort, line 1369: If you reach this line, and `hasStoredBidRequest` is true, then you know `!deps.cfg.GenerateRequestId` also has to be true, you don't need to test both.

Another potential logic flow is:
```suggestion
        storedReqForPatch = ''
        if hasStoredBidRequest {
          storedReqForPatch = storedRequests[storedBidRequestId]
        }
	if deps.cfg.GenerateRequestId {
	  uuidPatch, err := generateUuidForBidRequest()
	  if hasStoredBidRequest {
	    storedReqForPatch, err := jsonpatch.MergePatch(storedRequests[storedBidRequestId], uuidPatch)
	  } else {
	    storedReqForPatch := uuidPatch
	    hasStoredBidRequest = true
	  }
	}
	
	if hasStoredBidRequest {
	  resolvedRequest, err = jsonpatch.MergePatch(requestJson, storedReqForPatch)
	}
```
with the error handling and other supporting code included.

endpoints/openrtb2/auction_test.go
```diff
@@ -1089,7 +1161,7 @@ func TestOversizedRequest(t *testing.T) {
 		&mockStoredReqFetcher{},
 		empty_fetcher.EmptyFetcher{},
 		empty_fetcher.EmptyFetcher{},
-		&config.Configuration{MaxRequestSize: int64(len(reqBody) - 1)},
+		&config.Configuration{MaxRequestSize: int64(len(reqBody) - 1), GenerateRequestId: false},
```
hhhjort, line 1164: The nil value of a boolean is `false`, so you shouldn't have to explicitly set these values.

endpoints/openrtb2/auction.go
```diff
@@ -1567,3 +1584,30 @@ func getAccountID(pub *openrtb2.Publisher) string {
...
+
+func storedRequestErrorChecker(err error, requestJson []byte, storedRequests map[string]json.RawMessage, storedBidRequestId string) ([]byte, []error) {
+	if err != nil {
+		hasErr, Err := getJsonSyntaxError(requestJson)
+		if hasErr {
+			err = fmt.Errorf("Invalid JSON in Incoming Request: %s", Err)
+		} else {
+			hasErr, Err = getJsonSyntaxError(storedRequests[storedBidRequestId])
+			if hasErr {
+				fmt.Println(err)
```
SyntaxNode, line 1596: Please remove `fmt.Println(err)`.

endpoints/openrtb2/auction.go
```diff
@@ -1567,3 +1584,30 @@ func getAccountID(pub *openrtb2.Publisher) string {
...
+				err = fmt.Errorf("Invalid JSON in Stored Request with ID %s: %s", storedBidRequestId, Err)
+				err = fmt.Errorf("ext.prebid.storedrequest.id refers to Stored Request %s which contains Invalid JSON: %s", storedBidRequestId, Err)
```
SyntaxNode, line 1598: Which error message do you want to use here? The first one is currently overwritten by the second. The formatting of the second error is more Go-like, as it doesn't start with an upper case letter.

endpoints/openrtb2/auction.go
```diff
@@ -1567,3 +1584,30 @@ func getAccountID(pub *openrtb2.Publisher) string {
 	}
 	return metrics.PublisherUnknown
 }
+
+func storedRequestErrorChecker(err error, requestJson []byte, storedRequests map[string]json.RawMessage, storedBidRequestId string) ([]byte, []error) {
+	if err != nil {
```
SyntaxNode, line 1589: The original `err` passed in isn't used other than for this check, which is already covered on both code paths to this method. Consider removing the argument.

endpoints/openrtb2/auction.go
```diff
@@ -1340,21 +1340,38 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+		if err != nil {
+			requestErrorJson, errL := storedRequestErrorChecker(err, requestJson, storedRequests, storedBidRequestId)
+			return requestErrorJson, nil, errL
+		}
+	} else if !deps.cfg.GenerateRequestId && hasStoredBidRequest {
+		resolvedRequest, err = jsonpatch.MergePatch(storedRequests[storedBidRequestId], requestJson)
+		if err != nil {
+			requestErrorJson, errL := storedRequestErrorChecker(err, requestJson, storedRequests, storedBidRequestId)
+			return requestErrorJson, nil, errL
+		}
```
SyntaxNode, line 1374: Is there a way to condense these three conditions?

endpoints/openrtb2/auction.go
```diff
@@ -1340,21 +1340,38 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 
 	// Apply the Stored BidRequest, if it exists
 	resolvedRequest := requestJson
-	if hasStoredBidRequest {
-		resolvedRequest, err = jsonpatch.MergePatch(storedRequests[storedBidRequestId], requestJson)
+
+	// If GenerateRequestId is true, we'll generate a uuid for the bid request
```
SyntaxNode, line 1344: IMHO this comment doesn't provide much value. Please consider removing it.

endpoints/openrtb2/auction.go
```diff
@@ -1567,3 +1584,30 @@ func getAccountID(pub *openrtb2.Publisher) string {
...
+}
+
+func generateUuidForBidRequest() ([]byte, error) {
+	newReqId, err := uuid.NewV4()
+	if err != nil {
+		return nil, err
+	}
+	newReqString := newReqId.String()
+	return []byte(`{"id":"` + newReqString + `"}`), nil
+}
```
SyntaxNode, line 1613: Similar comment as with the AMP endpoint, could you please extract the random generator to a dependency?

endpoints/openrtb2/auction.go
```diff
@@ -1567,3 +1584,30 @@ func getAccountID(pub *openrtb2.Publisher) string {
...
+	return requestJson, nil
+}
+
+func generateUuidForBidRequest() ([]byte, error) {
+	newReqId, err := uuid.NewV4()
+	if err != nil {
+		return nil, err
+	}
+	newReqString := newReqId.String()
+	return []byte(`{"id":"` + newReqString + `"}`), nil
```
SyntaxNode, line 1612: It might be safer to use json.Marshal here to set the id instead of generating a json blob, but I wonder if jsonpatch is faster?

hhhjort (dismissed):
LGTM

endpoints/openrtb2/auction.go
```diff
@@ -1340,21 +1344,30 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
-			} else {
-				hasErr, Err = getJsonSyntaxError(storedRequests[storedBidRequestId])
-				if hasErr {
-					err = fmt.Errorf("Invalid JSON in Stored Request with ID %s: %s", storedBidRequestId, Err)
-					err = fmt.Errorf("ext.prebid.storedrequest.id refers to Stored Request %s which contains Invalid JSON: %s", storedBidRequestId, Err)
-				}
-			}
 			return nil, nil, []error{err}
 		}
+		if hasStoredBidRequest {
```
SyntaxNode, line 1353: Please add test coverage for this conditional.

SyntaxNode (commented):
This is missing the {{UUID}} macro, but otherwise looks solid. I'm good if you'd like to add the macro in a follow-up PR. The Java version only performs this behavior for app requests, is that a necessary restriction?

endpoints/openrtb2/auction.go
```diff
@@ -1567,3 +1585,35 @@ func getAccountID(pub *openrtb2.Publisher) string {
 	}
 	return metrics.PublisherUnknown
 }
+
+func storedRequestErrorChecker(err error, requestJson []byte, storedRequests map[string]json.RawMessage, storedBidRequestId string) []error {
```
SyntaxNode, line 1589: The `err` argument is not used. Please consider removing it.

endpoints/openrtb2/auction_test.go
```diff
@@ -1081,10 +1091,80 @@ func TestStoredRequestsVideoErrors(t *testing.T) {
...
+			expectedID:             uuid,
+		},
+	}
+
+	for _, test := range testCases {
+		deps.cfg.GenerateRequestID = test.givenGenerateRequestID
+		newRequest, _, errList := deps.processStoredRequests(context.Background(), json.RawMessage(test.givenRawData))
+		for _, err := range errList {
+			assert.Empty(t, err, test.description)
+		}
```
SyntaxNode, line 1155: You don't need to use a loop here. You can simply write:
```
newRequest, _, errList := deps.processStoredRequests(context.Background(), json.RawMessage(test.givenRawData))
assert.Empty(t, errList, test.description)
```

endpoints/openrtb2/amp_auction.go
```diff
@@ -344,6 +347,15 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
 		errs = []error{err}
 		return
 	}
+	// Generates Request ID if flag is set to true
```
SyntaxNode, line 350: Nitpick: Comment does not seem valuable.

endpoints/openrtb2/auction.go
```diff
@@ -1338,23 +1342,37 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		return nil, nil, errs
 	}
 
+	isAppRequest, err := checkIfAppRequest(requestJson)
+	if err != nil {
+		return nil, nil, []error{err}
+	}
```
SyntaxNode, line 1348: Looks good. I like that you've extracted the check to a method. To increase performance, I recommend moving this check within the `deps.cfg.GenerateRequestID ` check, such that we only bother to check for an app request if the feature is turned on. I usually don't recommend nesting, but in this case it's a performance improvement in certain cases.

hhhjort (approved):
LGTM

endpoints/openrtb2/auction.go
```diff
@@ -1354,20 +1358,35 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
-				err = fmt.Errorf("Invalid JSON in Incoming Request: %s", Err)
-			} else {
-				hasErr, Err = getJsonSyntaxError(storedRequests[storedBidRequestId])
-				if hasErr {
-					err = fmt.Errorf("Invalid JSON in Stored Request with ID %s: %s", storedBidRequestId, Err)
-					err = fmt.Errorf("ext.prebid.storedrequest.id refers to Stored Request %s which contains Invalid JSON: %s", storedBidRequestId, Err)
+			return nil, nil, []error{err}
+		}
+		if isAppRequest {
+			uuidPatch, err := generateUuidForBidRequest(deps.uuidGenerator)
```
SyntaxNode, line 1368: The `uuidPatch` is being generated for app requests, but it should only be generated when also `hasStoredBidRequest` is true.

SyntaxNode (changes requested):
The AMP flow checks out with local end to end tests. Please see me comment for a bug in the APP flow.

endpoints/openrtb2/auction.go
hhhjort: Why do we want to disable generating request IDs when there is no stored request for an App request?

hhhjort: Talked to SyntaxNode and this is desired behavior. Should be documented somewhere, in the end (not in the code, but the docs elsewhere) since it is not intuitively obvious when this feature should or should not kick in.

hhhjort (approved):
LGTM

SyntaxNode (approved):

PR #1940, opened 2021-07-27
IX: avoid htmlescaping in json encoding for native ad markup

manigandham:
The last adapter change required merging `eventtrackers` with `imptrackers` in Native ad markup in the bid responses. Unfortunately go's standard `json.Marshal` function runs through HtmlEscape and replaces characters when encoding strings, and since native markup is sent as JSON encoded in JSON, this is causing issues in upstream servers when they receive the response.

This change uses a separate JSON encoder with HTML escaping disabled. Tests have been updated to include more complex URLs.


manigandham:
Adding more context (from in code comments):

**json.Marshal**
> String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune. So that the JSON will be safe to embed inside HTML <script> tags, the string is encoded using HTMLEscape, which replaces "<", ">", "&", U+2028, and U+2029 are escaped to "\u003c","\u003e", "\u0026", "\u2028", and "\u2029". This replacement can be disabled when using an Encoder, by calling SetEscapeHTML(false).
https://pkg.go.dev/encoding/json#Marshal

**json.SetEscapeHTML**
> SetEscapeHTML specifies whether problematic HTML characters should be escaped inside JSON quoted strings. The default behavior is to escape &, <, and > to \u0026, \u003c, and \u003e to avoid certain safety problems that can arise when embedding JSON in HTML.
https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML

**json.Encode** (for removing the newline when encoding the new ad markup)
> Encode writes the JSON encoding of v to the stream, followed by a newline character.
https://pkg.go.dev/encoding/json#Encoder.Encode

grossjo (approved):

manigandham:
Hi @guscarreon @bsardo - Would you be able to look at this soon? It's a small change and we would like to get it into the next release if possible...

mansinahar:
@manigandham Currently we have a lot of open PRs that need our attention so we might not be able to attend to this immediately but we will definitely try to prioritize it and take a look as soon as possible. Thanks for your contribution!

PR #1944, opened 2021-08-02
IX: avoid htmlescaping in json encoding for native ad markup

mlb7687:
Copy of PR 1940 (https://github.com/prebid/prebid-server/pull/1940) which was closed due to employee departure from IX.

The last adapter change required merging eventtrackers with imptrackers in Native ad markup in the bid responses. Unfortunately go's standard json.Marshal function runs through HtmlEscape and replaces characters when encoding strings, and since native markup is sent as JSON encoded in JSON, this is causing issues in upstream servers when they receive the response.

This change uses a separate JSON encoder with HTML escaping disabled. Tests have been updated to include more complex URLs.


robhazan (approved):

mansinahar (approved):

bsardo (approved):
LGTM

PR #1946, opened 2021-08-03
New Adapter: YSSP (First Step In Rebranding From VerizonMedia)

oath-jac:
As first step in the VerizonMedia rebranding we cloned the VerizonMedia adapter into YSSP adapter. After all the publishers are migrated we'll remove the VerizonMedia code.


SyntaxNode:
What are the next steps / end goal of the rebranding? We may be able to offer better ways if you don't plan to change the adapter and are only looking to add a new bidder id.

oath-jac:
> What are the next steps / end goal of the rebranding? We may be able to offer better ways if you don't plan to change the adapter and are only looking to add a new bidder id.

Hi, the end goal is to rebrand the original verizonmedia adapter to a new yssp adapter. We have to remove the verizonmedia brand from our ads related products. We gonna use the yssp name (yahoo ssp).
https://www.sportspromedia.com/news/verizon-media-apollo-global-management-sale-yahoo-sports-ryot-nfl-nba
the steps are:
1) add a new yssp adapter, it works identical as verizonmedia adapter
2) migrate current onborded publishers to use the new yssp adapter
3) when publisher team informs us that everybody was migrated and they use the new adapter, we'll remove the 'verizonmedia` adapter

SyntaxNode:
Thank you for the context. In this case, I recommend deleting the verizonmedia in this PR and using the YSSP `Builder` for both:

```
openrtb_ext.BidderVerizonMedia:      yssp.Builder,
openrtb_ext.BidderYSSP:              yssp.Builder,
```

You'll need to keep the verizonmedia usersync.go and usersync_test.go - but the other files in that folder may be deleted now.

Later, when you delete the verizonmedia adapter, please extend the `GetDisabledBiddersErrorMessages` method to add an error message:

```
func GetDisabledBiddersErrorMessages(infos config.BidderInfos) map[string]string {
	disabledBidders := map[string]string{
		"lifestreet": `Bidder "lifestreet" is no longer available in Prebid Server. Please update your configuration.`,
		"verizonmedia": `Bidder "verizonmedia" is replaced wth bidder 'yssp'. Please update your configuration.`,
	}
	...
```

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type YSSPAdapter struct {
```
SyntaxNode, line 16: Please name this `adapter`. There is no need to repeat the package name. There is no need to export this type with an uppercase letter.

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+func (a *YSSPAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	errors := make([]error, 0, 1)
+
+	if len(request.Imp) == 0 {
+		err := &errortypes.BadInput{
+			Message: "No impression in the bid request",
+		}
+		errors = append(errors, err)
+		return nil, errors
+	}
```
SyntaxNode, line 29: Please remove this duplicate check. The core logic of PBS ensures adapters are not called without at least one impression.

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type YSSPAdapter struct {
+	URI string
+}
+
+func (a *YSSPAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	errors := make([]error, 0, 1)
```
SyntaxNode, line 21: Let's be more optimistic and assume there will be no errors, instead of assuming 1 error. Please replace with:
```
var errors []error
```

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+			continue
+		}
+
+		if ysspExt.Pos == "" {
+			err = &errortypes.BadInput{
+				Message: fmt.Sprintf("imp #%d: missing param pos", idx),
+			}
+			errors = append(errors, err)
+			continue
+		}
```
SyntaxNode, line 77: If you add a minimum length in the json schema file, you can remove the Dcn and Pos empty string checks.

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Bad server response: %d.", err),
+		}}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(internalRequest.Imp))
+
+	for _, sb := range bidResp.SeatBid {
+		for _, bid := range sb.Bid {
```
SyntaxNode, line 136: There's actually a subtle bug here. Please update with either:

```
for _, bid := range sb.Bid {
  bid := bid
```

or

```
for i := range sb.Bid {
  bid := sb.Bid[i]
```

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+	if request.Imp[0].Banner == nil {
+		return nil
+	}
+
+	banner := *request.Imp[0].Banner
+	request.Imp[0].Banner = &banner
+
+	if banner.W != nil && banner.H != nil {
+		if *banner.W == 0 || *banner.H == 0 {
+			return errors.New(fmt.Sprintf("Invalid sizes provided for Banner %dx%d", *banner.W, *banner.H))
```
SyntaxNode, line 192: Please use `fmt.Errorf` here instead:
```
return fmt.Errorf("Invalid sizes provided for Banner %dx%d", *banner.W, *banner.H)
```

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		return nil
+	}
+
+	if len(banner.Format) == 0 {
+		return errors.New(fmt.Sprintf("No sizes provided for Banner %v", banner.Format))
```
SyntaxNode, line 198: Same here:
```
return fmt.Errorf("No sizes provided for Banner %v", banner.Format)

```

adapters/yssp/yssp.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if request.Imp[0].Banner == nil {
+		return nil
+	}
+
+	banner := *request.Imp[0].Banner
+	request.Imp[0].Banner = &banner
+
+	if banner.W != nil && banner.H != nil {
+		if *banner.W == 0 || *banner.H == 0 {
```
SyntaxNode, line 191: Please add a test case for this branch.

adapters/yssp/yssp.go
SyntaxNode, line 198: Please add a test case for this branch.

SyntaxNode (commented):
I understand this is a copy/paste of the existing verizonmedia adapter, but we've updated our code recommendations and this is a good time to make a few small updates.

mansinahar:
@oath-jac Friendly ping about the comments above

adapters/yssp/yssp.go
oath-jac: I addressed this, what is the benefit here? semantically we say we have a formatting error ?

VeronikaSolovei9 (dismissed):
LGTM

adapters/yssp/yssp.go
SyntaxNode: Correct. It's a better choice to use the most specific method. Since using `errors.New` and `fmt.Sprintf` together is a very common use case, Go has `fmt.Errorf` to make this more explicit.

exchange/adapter_util.go
```diff
@@ -78,7 +78,8 @@ func GetActiveBidders(infos config.BidderInfos) map[string]openrtb_ext.BidderNam
 // GetDisabledBiddersErrorMessages returns a map of error messages for disabled bidders.
 func GetDisabledBiddersErrorMessages(infos config.BidderInfos) map[string]string {
 	disabledBidders := map[string]string{
-		"lifestreet": `Bidder "lifestreet" is no longer available in Prebid Server. Please update your configuration.`,
+		"lifestreet":   `Bidder "lifestreet" is no longer available in Prebid Server. Please update your configuration.`,
+		"verizonmedia": `Bidder "verizonmedia" is replaced wth bidder 'yssp'. Please update your configuration.`,
```
SyntaxNode, line 82: I think there is some confusion with my comment for the `GetDisabledBiddersErrorMessages` method. This change is only necessary after the `verizonmedia` adapter has been removed - so that publishers would receive this upgrade notice rather than an unknown bidder error message. This has no affect on the current code as it will either be overwritten in the loop with a standard disabled message for entirely ignored.

SyntaxNode (commented):
Thank you for making the requested changes. Looks great except for the disabled message. If you could please remove the change to `GetDisabledBiddersErrorMessages` we'll get tis merged in.

Sorry for the delay in re-reviewing. We've had a higher than normal amount of submitted PRs and much of our team has been taking vacation over the summer.

SyntaxNode (approved):

VeronikaSolovei9 (approved):

PR #1948, opened 2021-08-03
Rubicon: Implement FPD resolving functionality

snahornyi:


adapters/rubicon/rubicon.go
```diff
@@ -941,6 +909,104 @@ func resolveBidFloor(bidFloor float64, bidFloorCur string, reqInfo *adapters.Ext
...
+func addStringArrayAttribute(attribute []string, target json.RawMessage, attributeName string) json.RawMessage {
+	targetAsMap := messageToMap(target)
+	targetAsMap[attributeName] = attribute
+
+	updatedTarget, _ := json.Marshal(targetAsMap)
+	return updatedTarget
+}
+
+func getAdSlot(imp openrtb2.Imp) string {
+	adServerContextName, _ := jsonparser.GetString(imp.Ext, "context", "data", "adserver", "name")
```
mansinahar, line 980: A Note On `jsonparser`:
This is substantially faster for reading values from a json blob than the Go json package, but you need to keep in mind that each jsonparser.GetXXX call re-parses the entire json. Based on performance measurements, the tipping point of efficiency is around 4 calls and you're using it on `imp.Ext` more than that. Here and then above at L947 and L948. At this point, please consider switching to EachKey to use a single pass.

adapters/rubicon/rubicon.go
```diff
@@ -941,6 +909,104 @@ func resolveBidFloor(bidFloor float64, bidFloorCur string, reqInfo *adapters.Ext
...
+	if adServerDataName == "gam" {
+		dataAdSlot, _ := jsonparser.GetString(imp.Ext, "data", "adserver", "adslot")
+		return dataAdSlot
+	}
+
+	return ""
+}
+
+func updateUserRpTargetWithFpdAttributes(visitor json.RawMessage, user openrtb2.User) (json.RawMessage, error) {
+	existingTarget, _, _, _ := jsonparser.Get(user.Ext, "rp", "target")
```
mansinahar, line 996: Please do not ignore errors and add appropriate error handling logic here

adapters/rubicon/rubicon.go
```diff
@@ -941,6 +909,104 @@ func resolveBidFloor(bidFloor float64, bidFloorCur string, reqInfo *adapters.Ext
...
+
+	target := populateFirstPartyDataAttributes(visitor, existingTarget)
+
+	userExtData, _, _, _ := jsonparser.Get(user.Ext, "data")
```
mansinahar, line 1000: Please do not ignore errors and add appropriate error handling logic here

adapters/rubicon/rubicon.go
```diff
@@ -941,6 +909,104 @@ func resolveBidFloor(bidFloor float64, bidFloorCur string, reqInfo *adapters.Ext
...
+
+	updatedTarget, _ := json.Marshal(targetAsMap)
+	return updatedTarget
+}
+
+func getAdSlot(imp openrtb2.Imp) string {
+	adServerContextName, _ := jsonparser.GetString(imp.Ext, "context", "data", "adserver", "name")
+
+	if adServerContextName == "gam" {
+		contextAdSlot, _ := jsonparser.GetString(imp.Ext, "context", "data", "adserver", "adslot")
```
mansinahar, line 983: Please do not ignore errors and add appropriate error handling logic here

adapters/rubicon/rubicon.go
```diff
@@ -941,6 +909,104 @@ func resolveBidFloor(bidFloor float64, bidFloorCur string, reqInfo *adapters.Ext
...
+		return contextAdSlot
+	}
+	adServerDataName, _ := jsonparser.GetString(imp.Ext, "data", "adserver", "name")
```
mansinahar, line 986: Please do not ignore errors and add appropriate error handling logic here

adapters/rubicon/rubicon.go
```diff
@@ -941,6 +909,104 @@ func resolveBidFloor(bidFloor float64, bidFloorCur string, reqInfo *adapters.Ext
...
+	if adServerDataName == "gam" {
+		dataAdSlot, _ := jsonparser.GetString(imp.Ext, "data", "adserver", "adslot")
```
mansinahar, line 988: Please do not ignore errors and add appropriate error handling logic here

mansinahar (changes requested):
I took an initial pass at changes in `rubicon.go` and have left some comments. I also noticed that for most of the `jsonparser.GetXXX` calls and the `json.Marshal` and `json.Unmarshal` calls, errors are being ignored. Please _do not_ ignore errors and add appropriate error handling logic. Once updated, we will take a look at this again

adapters/rubicon/rubicon.go
```diff
@@ -964,6 +1030,73 @@ func updateExtWithIabAttribute(target json.RawMessage, data []openrtb2.Data, seg
...
+	var stringArray []string
+	for _, val := range arr {
+		stringArray = append(stringArray, strconv.FormatBool(val.(bool)))
+	}
+	return stringArray
+}
+
+func messageToMap(message json.RawMessage) map[string]interface{} {
+	targetAsMap := make(map[string]interface{})
+	_ = json.Unmarshal(message, &targetAsMap)
```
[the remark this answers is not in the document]
snahornyi: This is was done, because `message ` can be nil and this will cause an error, and by ignoring this err I will just return empty map

adapters/rubicon/rubicon.go
```diff
@@ -964,6 +1030,73 @@ func updateExtWithIabAttribute(target json.RawMessage, data []openrtb2.Data, seg
...
 
+func populateFirstPartyDataAttributes(source json.RawMessage, target json.RawMessage) json.RawMessage {
+	targetAsMap := messageToMap(target)
+	sourceAsMap := messageToMap(source)
+
+	for key, val := range sourceAsMap {
+		switch typedValue := val.(type) {
+		case string:
+			targetAsMap[key] = [1]string{typedValue}
+		case float64:
```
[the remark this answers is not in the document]
snahornyi: val.(type) accepts int as float64

adapters/rubicon/rubicon.go
snahornyi: If I will not have this entry I will just have existingTarget as nil, and error `Key path not found`, which can be skipped by existing logic

adapters/rubicon/rubicon.go
snahornyi: Same as above

adapters/rubicon/rubicon.go
snahornyi: Same as above

adapters/rubicon/rubicon.go
snahornyi: Same as above

adapters/rubicon/rubicon.go
snahornyi: Same as above

adapters/rubicon/rubicon.go
snahornyi: Nice catch, thx)

adapters/rubicon/rubicon.go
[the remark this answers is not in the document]
mansinahar: > because message can be nil and this will cause an error

This is exactly why we need to add proper error handling. I understand that you just want to return an empty map when that happens but you can achieve that with proper error handling as well. We require that these errors aren't ignored because otherwise it becomes really difficult to debug (for the Prebid Server host and for the publisher) if an issue arises. Clear error messages on the other hand help properly communicate what exactly went wrong and eliminates guess work. Plus, it also saves a lot of time while debugging.

adapters/rubicon/rubicon.go
mansinahar: Please add error handling logic for errors other than key path not found for the same reason as above.

adapters/rubicon/rubicon_test.go
```diff
@@ -1306,14 +1320,410 @@ func TestOpenRTBRequestWithImpAndAdSlotIncluded(t *testing.T) {
...
+			Gender:   "female",
+			Keywords: "someWords",
+		},
+		App: &openrtb2.App{
+			ID:   "com.test",
+			Name: "testApp",
+		},
+	}
+
+	reqs, _ := bidder.MakeRequests(request, &adapters.ExtraRequestInfo{})
```
bsardo, line 1366: I'm curious why you opted to create several UTs that call `MakeRequests` when you could have created additional JSON tests. At a quick glance, I see that some of the fields in the original request that are incorporated into FPD like `keywords` and `inventory` don't appear in any of the JSON tests but are validated in some capacity in these UTs.

snahornyi: To my mind, in future this test will be enriched with some scenarios, so decided to have it as UT

adapters/rubicon/rubicon.go
```diff
@@ -941,6 +909,144 @@ func resolveBidFloor(bidFloor float64, bidFloorCur string, reqInfo *adapters.Ext
...
+
+	if len(extImp.Keywords) > 0 {
+		target = addStringArrayAttribute(extImp.Keywords, target, "keywords")
+	}
+
+	return target, nil
+}
+
+func logParsingError(err error) {
+	glog.Infof(fmt.Sprintf("Error while parsing. Error: %s", err.Error()))
```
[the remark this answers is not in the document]
snahornyi: As you wrote `We require that these errors aren't ignored because otherwise it becomes really difficult to debug`. For me this logs is not needed, and errors like `unexpected end of JSON input` should be ignored. What type of errors can be produced here, that should be returned to user? Maybe, I haven't understand you correctly from the beginning, but will appreciate your answers)
(Updated PR, with new logic for some errors)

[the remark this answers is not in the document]
mansinahar: Well, I'd say errors like `unexpected end of JSON input` shouldn't be ignored. They should rather be communicated back to the publisher so that the publisher knows that there's possibly a syntax error in their raw JSON string within the request and they can fix it accordingly.  If you just ignore the errors, publishers won't know about them and these errors might cause lower/no bids from your adapter, for example. And if a publisher then wants to debug why they are getting lower/no bids from Rubicon, it will be very difficult because these errors are just silently being ignored. I. hope that makes sense

adapters/rubicon/rubicon_test.go
bsardo: Gotcha, the team chatted and would like to see the JSON framework utilized. In that case, could you please replace the tests you added that call `MakeRequests` with corresponding supplemental JSON tests?
That would be:
`TestOpenRTBRequestUserParametersShouldBeBlocked`
`TestOpenRTBRequestImpFpdFromAppResolving`
`TestOpenRTBRequestImpFpdFromSiteResolving`
`TestOpenRTBRequestUserFpdResolving`
`TestOpenRTBSiteSegTaxes` - I believe the test cases here can all be in a single JSON test

I understand that there are other tests in `rubicon_test.go` written a couple of years ago that take the UT approach with respect to `MakeRequests` but our docs now indicate that we require use of the JSON framework wherever possible as there are some advantages to using it.
https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#adapter-code-tests

bsardo (commented):
This is looking good; just one minor suggestion.

bsardo (dismissed):

AlexBVolcy (approved):
Looks good to me!

bsardo (approved):

VeronikaSolovei9 (approved):
LGTM

AlexBVolcy (approved):
LGTM!

SyntaxNode:
Using administrator privileges to override @mansinahar's outstanding request for changes. She is currently out of the office and we have three other approvers from the core team.

PR #1955, opened 2021-08-11
InteractiveOffers - Parameters changed & dynamic endpoint

IOTiagoFaria:


VeronikaSolovei9 (commented):
Just a minor comment, other than that it looks good

config/config.go
```diff
@@ -911,7 +911,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.gumgum.endpoint", "https://g2.gumgum.com/providers/prbds2s/bid")
 	v.SetDefault("adapters.improvedigital.endpoint", "http://ad.360yield.com/pbs")
 	v.SetDefault("adapters.inmobi.endpoint", "https://api.w.inmobi.com/showad/openrtb/bidder/prebid")
-	v.SetDefault("adapters.interactiveoffers.endpoint", "https://prebid-server.ioadx.com/bidRequest/?partnerId=d9e56d418c4825d466ee96c7a31bf1da6b62fa04")
+	v.SetDefault("adapters.interactiveoffers.endpoint", "https://prebid-server.ioadx.com/bidRequest/?partnerId=")
```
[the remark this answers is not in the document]
IOTiagoFaria: I wasn't able to use the {{.PartnerId}} so I used the {{.AccountID}} ;)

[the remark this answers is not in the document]
VeronikaSolovei9: I believe you can use ExtImpInteractiveoffers
```
func (a *adapter) buildEndpointURL(params *openrtb_ext.ExtImpInteractiveoffers) (string, error) {
	endpointParams := openrtb_ext.ExtImpInteractiveoffers{PartnerId: params.PartnerId}
	return macros.ResolveMacros(a.endpoint, endpointParams)
}
```

config/config.go
```diff
@@ -911,7 +911,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.gumgum.endpoint", "https://g2.gumgum.com/providers/prbds2s/bid")
 	v.SetDefault("adapters.improvedigital.endpoint", "http://ad.360yield.com/pbs")
 	v.SetDefault("adapters.inmobi.endpoint", "https://api.w.inmobi.com/showad/openrtb/bidder/prebid")
-	v.SetDefault("adapters.interactiveoffers.endpoint", "https://prebid-server.ioadx.com/bidRequest/?partnerId=d9e56d418c4825d466ee96c7a31bf1da6b62fa04")
+	v.SetDefault("adapters.interactiveoffers.endpoint", "http://localhost:8020/bidRequest/?partnerId={{.AccountID}}")
```
VeronikaSolovei9, line 914: Please do not forget to modify it here too

VeronikaSolovei9: And localhost :)

IOTiagoFaria: I keep getting this error

F0817 08:41:39.043591   11199 main.go:37] Configuration could not be loaded or did not pass validation: validation errors (1 error):
  1: Unable to resolve endpoint: https://prebid-server.ioadx.com/bidRequest/?partnerId={{.PartnerId}} for adapter: interactiveoffers. template: endpointTemplate:1:56: executing "endpointTemplate" at <.PartnerId>: can't evaluate field PartnerId in type macros.EndpointTemplateParams


And checking the file macros/macros.go 

// EndpointTemplateParams specifies params for an endpoint template
type EndpointTemplateParams struct {
	Host        string
	PublisherID string
	ZoneID      string
	SourceId    string
	AccountID   string
	AdUnit      string
}

I'm missing something...

hhhjort: I believe "Please do not forget to modify AccountId to PartnerId" is referring to `interactiveoffers.go` line 116, which is currently correct. In the URL template above it should stay `AccountID`, as the macro function doesn't allow for `PartnerId` as a parameter.  It is a little confusing, as the macros package doesn't allow for arbitrary endpoint macro substitutions, and `AccountId` already has another meaning in the code. The reason we don't allow arbitrary macros is so at run time PBS can validate that the URL templates supplied resolve into valid URLs, which requires us to supply dummy values for all the parameters. This can only be done with a known set of macros. So as long as your `PartnerId` resembles an `AccountId` then reusing `AccountId` will work. 

A better way of accomplishing your goal would be to use the html library functions to add query string parameters to the URL instead of trying to resolve it as a macro. Or even a simple string substitution. An empty/missing query string will still resolve as a valid URL (it checks that the URL is parseable, not that it will actually connect). The provided template routines are only needed for parameters that appear in the domain/path part of the URL, so we can properly verify that the URL is a valid URL.

static/bidder-params/interactiveoffers.json
```diff
@@ -4,10 +4,10 @@
...
   "properties": {
-    "pubid": {
-      "type": "integer",
-      "description": "The publisher id"
+    "partnerId": {
+      "type": "string",
+      "description": "The partners id"
     }
   },
-  "required": ["pubid"]
```
hhhjort, line 12: This is a difficult thing to pull off. We made a design choice that any request that has any bidder that fails the JSON schema will invalidate the entire request. The idea was that this would only happen when a publisher creates/updates their prebid config. When they do this, they should be looking to see if prebid is returning any bids, but probably aren't sensitive to if all bidders are bidding. They would otherwise not notice that a bidder is misconfigured until at some point they look at their analytics and notice the bidder is not bidding on a particular page or configuration setup. We think the current design choice is the better option to ensure everyone has a chance to bid.

The recommended path here is to drop `pubid` as a required parameter, but leave `partnerId` as optional. The adapter can then return an error explaining the issue with the request. If a request passes the JSON schema, but a bidder returns an error explaining the issue when `MakeRequests` is called. This will allow the other configured bidders to bid rather than invalidating the entire request. At a later point in time, when the publishers have all had a chance to update to the new schema, we can make it required here. We just don't want all the publisher who are currently using your adapter to suddenly have all their prebid configs invalidated by surprise.

config/config.go
VeronikaSolovei9: Thank you for the answer @hhhjort! 
@IOTiagoFaria I apologize for my previous comment, I didn't check if it works all together. You were right about `AccountId`

IOTiagoFaria: Thank you both for the explanation, if its working the way it is and theres no code issues I'll keep it as it is.

I've also changed the java version and on that I've made a simple string replace and the code didn't pass some checks, I had to use some kind of macro...

Is it ok to leave the code as it is?

mansinahar:
@IOTiagoFaria Can you please rebase this on top of master and resolve the conflicts? Thanks!

IOTiagoFaria:
If this pull still has issues regarding the merge let me know, I'll just create a new fork with the most up to date repository, that way it should have no merge issues....

IOTiagoFaria:
Actually, I'll just go ahead and do that, sorry for all of this.....

Issue #1923, opened 2021-07-14
Inconsistent PBS behaviour and docs for Interstitial ads.


YuriyVelichkoOpenX:
## Background

Investigating the [issue for mobile SDK](https://github.com/prebid/prebid-mobile-ios/issues/391), we faced incorrect behavior of SDK in the first place and with the inconsistency of PBS in the second.

SDK sends the size of device's display in the `banner.format.[0]`. It leads to issues for some demand partners who expect the actual size of demanded creative in `banner.format`. 

So we are going to fix it and send an empty `banner.format` for interstitial ads. This fix won't change the current behavior and will give a better experience since publishers can set the needed formats in the stored request. It is impossible right now because of the hardcoded device size in the client request. 

According to the docs [#1](https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html), [#2](https://docs.prebid.org/prebid-server/features/pbs-interstitials.html) we can do it. And if `banner.format` in the bid request is empty, the PBS should: 

1. get the `banner.format` from the stored impression
2. if the `banner.format` is still empty - use the device size for `banner.format.[0]`. 

**We think that documented behavior is totally correct.** 

## Current Behavior

However, the **implementation is different**. 

If  `banner.format` is empty in the both - bid request and stored impression, the initial request fails here: 

https://github.com/prebid/prebid-server/blob/aff5f70d8fee97d4641f8686c2a34d91a56385b4/endpoints/openrtb2/auction.go#L586

## Expected Behavior

If the `banner.format` is empty but `imp.instl` is `1` - don't fail the bid request. Let it pass further to [this point](https://github.com/prebid/prebid-server/blob/d384e91bf99f886927a2bd412fd439502a1de088/endpoints/openrtb2/interstitial.go#L41) where PBS will place the `device.w` and `device.h` into banner.format.[0] respectively to the docs.

## Known Workaround

We know that if we set the minSizePercentage properties, the bid request will work as expected. However, we believe that it is not a proper way because::

1.  The display size as a default requested size for interstitial ad looks inappropriate
2. Publishers unable to customize the ad sizes via stored requests, which can be critical for mobile integration. 

## Steeps to reproduce 

An example of a bid request:

```
[...]

PR #1957, opened 2021-08-11
Interstitial behavior fix for issue 1923

ChrisHuie:
@SyntaxNode I added the logic to the main validate function but seem to have a bug in the above validateImp function from referencing the struct. Do I need to make a type conversion on imp.instl paramter? Sorry new to structs and that argument seems to be the issue. Adding the test function too. Fix for issue -> #1923


SyntaxNode:
> @SyntaxNode I added the logic to the main validate function but seem to have a bug in the above validateImp function from referencing the struct. Do I need to make a type conversion on imp.instl paramter? 

Looks like you figured it out. Indeed, you are using the imp.Instl parameter correctly. No type conversion is required.

SyntaxNode:
Looks good now, please add new json tests for this new use case. I recommend looking at which tests failed when you had the logic inverted and use one of those as a template.

endpoints/openrtb2/auction_test.go
```diff
@@ -3491,3 +3491,32 @@ type hardcodedResponseIPValidator struct {
 func (v hardcodedResponseIPValidator) IsValid(net.IP, iputil.IPVersion) bool {
 	return v.response
 }
+
+func TestValidateBanner(t *testing.T) {
+	impIndex := 4
```
SyntaxNode, line 3496: Looks like the impIndex is just used for the error messages in `validateBanner`. Why not use 0?

ChrisHuie: no reason. I can change to 0

SyntaxNode (dismissed):

endpoints/openrtb2/auction.go
```diff
@@ -567,7 +567,7 @@ func (deps *endpointDeps) validateImp(imp *openrtb2.Imp, aliases map[string]stri
 	return nil
 }
 
-func validateBanner(banner *openrtb2.Banner, impIndex int) error {
+func validateBanner(banner *openrtb2.Banner, imp *openrtb2.Imp, impIndex int) error {
```
mansinahar, line 570: I'd suggest passing in a boolean flag like `isInterstitial` that is set if `imp.Instl == 1` rather than passing the entire `imp`

mansinahar:
Looks pretty straightforward to me other than the above comment. Thanks for helping with this @ChrisHuie

ChrisHuie:
fixing this today.

endpoints/openrtb2/auction.go
ChrisHuie: I tried to work on this and understand the idea but not exactly sure how to go about accessing the imp.Instl property without the imp object to set the bool flag. Would I be setting the flag outside this function using the imp object there or am I off base?

mansinahar: Yup. You're right. You'd have to set the boolean flag outside this function and then pass it to this function when it is called.
You can do this by changing the signature of this function to be:
```
func validateBanner(banner *openrtb2.Banner, impIndex int, isInterstitial bool) 
```

ChrisHuie: @mansinahar I know there probably is a better place for the interstitial function but wanted to make sure I was on the right track?

endpoints/openrtb2/auction.go
```diff
@@ -548,7 +548,7 @@ func (deps *endpointDeps) validateImp(imp *openrtb2.Imp, aliases map[string]stri
 		return []error{fmt.Errorf("request.imp[%d] must contain at least one of \"banner\", \"video\", \"audio\", or \"native\"", index)}
 	}
 
-	if err := validateBanner(imp.Banner, index); err != nil {
+	if err := validateBanner(imp.Banner, imp, index); err != nil {
```
mansinahar, line 551: `ValidateBanner` expects imp index to be the second argument and the third argument needs to be a boolean flag that you'll have to create before calling this function. You can set the value of that boolean flag based on the value of `imp.Instl` and then pass that boolean flag to this function

ChrisHuie: thanks :)

ChrisHuie: @mansinahar think I finally got it. Let me know though if there are more changes needed!

endpoints/openrtb2/auction.go
```diff
@@ -576,7 +576,15 @@ func (deps *endpointDeps) validateImp(imp *openrtb2.Imp, aliases map[string]stri
...
 }
 
-func validateBanner(banner *openrtb2.Banner, impIndex int) error {
+func isInterstitial(imp *openrtb2.Imp) bool {
+	if imp.Instl == 1 {
+		return true
+	}
+
+	return false
+}
```
SyntaxNode, line 585: Nitpick: This method can be simplified to:
```
func isInterstitial(imp *openrtb2.Imp) bool {
	return imp.Instl == 1
}
```

SyntaxNode (commented):
Looks good. It looks like the unit tests are no longer included in the PR. Could you please bring back the tests?

endpoints/openrtb2/auction.go
ChrisHuie: Changed. I must have had a typo somewhere because that was probably the first way I tried because that would work in Javascript :) Think I got this working though and hopefully cleaned!! Thanks @mansinahar and @SyntaxNode for the help on this!

hhhjort (dismissed):
LGTM

SyntaxNode:
Looks good. The `isInterstitial` method is never called in a test though. This is tough to do from unit tests because the design of that part of code is... less than ideal. I think this would be a good use of json tests. I recommend adding the following two cases for an end-to-end verification of full functionality in addition to the unit tests you already have:

/endpoints/openrtb2/sample-requests/valid-whole/exemplary/interstitial-no-size.json
```
{
    "description": "Interstitial Banner Ad Without Sizes Defined",
    "mockBidRequest": {
        "id": "some-request-id",
        "site": {
            "page": "test.somepage.com"
        },
        "imp": [{
            "id": "some-imp-id",
            "instl": 1,
            "banner": {},
            "ext": {
                "prebid": {
                    "bidder": {
                        "appnexus": {
                            "placementId": 12883451
                        }
                    }
                }
            }
        }]
    },
    "expectedBidResponse": {
        "id": "some-request-id",
        "seatbid": [{
            "bid": [{
                "id": "appnexus-bid",
                "impid": "",
                "price": 0
            }],
            "seat": "appnexus-bids"
        }],
        "bidid": "test bid id",
        "cur": "USD",
        "nbr": 0
    },
    "expectedReturnCode": 200
}
```

/endpoints/openrtb2/sample-requests/valid-whole/exemplary/banner-no-size.json
```
{
    "description": "Banner Ad Without Sizes Defined",
    "mockBidRequest": {
        "id": "some-request-id",
        "site": {
            "page": "test.somepage.com"
        },
        "imp": [{
            "id": "some-imp-id",
            "banner": {},
            "ext": {
                "prebid": {
                    "bidder": {
                        "appnexus": {
                            "placementId": 12883451
                        }
                    }
                }
            }
        }]
    },
    "expectedErrorMessage": "Invalid request: request.imp[0].banner has no sizes. Define \"w\" and \"h\", or include \"format\" elements.",
    "expectedReturnCode": 400
}
```

SyntaxNode (approved):

hhhjort (approved):

PR #1961, opened 2021-08-12
processStoredRequests refactoring

VeronikaSolovei9:
Due to upcoming stored response integration, amount of `jsonparser.Get` usages will be dramatically increased. It makes more sense to unmarshal `imp.ext.prebid`to access stored requests and stored responses data without using `jsonparser.Get`more than 3 times in a row.

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1393,19 +1393,19 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
 	impExtInfoMap := make(map[string]exchange.ImpExtInfo, len(impInfo))
-	resolvedImps := make([]json.RawMessage, 0)
+	resolvedImps := make([]json.RawMessage, 0, len(impInfo))
 	for i, impData := range impInfo {
 		if impData.ImpExtPrebid.StoredRequest != nil && len(impData.ImpExtPrebid.StoredRequest.ID) > 0 {
 			resolvedImp, err := jsonpatch.MergePatch(storedImps[impData.ImpExtPrebid.StoredRequest.ID], impData.Imp)
 
 			if err != nil {
-				hasErr, Err := getJsonSyntaxError(impData.Imp)
+				hasErr, errMessage := getJsonSyntaxError(impData.Imp)
 				if hasErr {
-					err = fmt.Errorf("Invalid JSON in Imp[%d] of Incoming Request: %s", i, Err)
+					err = fmt.Errorf("Invalid JSON in Imp[%d] of Incoming Request: %s", i, errMessage)
 				} else {
-					hasErr, Err = getJsonSyntaxError(storedImps[impData.ImpExtPrebid.StoredRequest.ID])
+					hasErr, errMessage = getJsonSyntaxError(storedImps[impData.ImpExtPrebid.StoredRequest.ID])
 					if hasErr {
-						err = fmt.Errorf("imp.ext.prebid.storedrequest.id %s: Stored Imp has Invalid JSON: %s", impData.ImpExtPrebid.StoredRequest.ID, Err)
+						err = fmt.Errorf("imp.ext.prebid.storedrequest.id %s: Stored Imp has Invalid JSON: %s", impData.ImpExtPrebid.StoredRequest.ID, errMessage)
 					}
 				}
 				return nil, nil, []error{err}
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1322,7 +1322,7 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
 	return false, ""
 }
 
-func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte, impInfo []ImpData) ([]byte, map[string]exchange.ImpExtInfo, []error) {
+func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte, impInfo []ImpExtPrebidData) ([]byte, map[string]exchange.ImpExtInfo, []error) {
 	// Parse the Stored Request IDs from the BidRequest and Imps.
 	storedBidRequestId, hasStoredBidRequest, err := getStoredRequestId(requestJson)
 	if err != nil {
@@ -1335,16 +1335,14 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		storedReqIds = []string{storedBidRequestId}
 	}
 
-	//impStoredReqIds list of all stored impression ids
 	impStoredReqIds := make([]string, 0)
-	impStoredReqIdsUnique := make(map[string]interface{})
+	impStoredReqIdsUniqueTracker := make(map[string]struct{})
 	for _, impData := range impInfo {
 		if impData.ImpExtPrebid.StoredRequest != nil && len(impData.ImpExtPrebid.StoredRequest.ID) > 0 {
 			storedImpId := impData.ImpExtPrebid.StoredRequest.ID
-			_, present := impStoredReqIdsUnique[storedImpId]
-			if !present {
+			if _, present := impStoredReqIdsUniqueTracker[storedImpId]; !present {
 				impStoredReqIds = append(impStoredReqIds, storedImpId)
-				impStoredReqIdsUnique[storedImpId] = true
+				impStoredReqIdsUniqueTracker[storedImpId] = struct{}{}
 			}
 		}
 	}
@@ -1394,7 +1392,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
-	impExtInfoMap := make(map[string]exchange.ImpExtInfo, 0)
+	impExtInfoMap := make(map[string]exchange.ImpExtInfo, len(impInfo))
 	resolvedImps := make([]json.RawMessage, 0)
 	for i, impData := range impInfo {
 		if impData.ImpExtPrebid.StoredRequest != nil && len(impData.ImpExtPrebid.StoredRequest.ID) > 0 {
@@ -1421,7 +1419,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 
 			echovideoAttributes := false
 			if impData.ImpExtPrebid.Options != nil {
-				echovideoAttributes = impData.ImpExtPrebid.Options.Echovideoattrs
+				echovideoAttributes = impData.ImpExtPrebid.Options.EchoVideoAttrs
 			}
 			impExtInfoMap[impId] = exchange.ImpExtInfo{EchoVideoAttrs: echovideoAttributes, StoredImp: storedImps[impData.ImpExtPrebid.StoredRequest.ID]}
 
@@ -1445,7 +1443,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 }
 
 // parseImpInfo parses the request JSON and returns impression and unmarshalled imp.ext.prebid
-func parseImpInfo(requestJson []byte) (impData []ImpData, errs []error) {
+func parseImpInfo(requestJson []byte) (impData []ImpExtPrebidData, errs []error) {
 
 	if impArray, dataType, _, err := jsonparser.Get(requestJson, "imp"); err == nil && dataType == jsonparser.Array {
 		_, err = jsonparser.ArrayEach(impArray, func(imp []byte, _ jsonparser.ValueType, _ int, err error) {
@@ -1456,14 +1454,14 @@ func parseImpInfo(requestJson []byte) (impData []ImpData, errs []error) {
 					errs = append(errs, err)
 				}
 			}
-			newImpData := ImpData{imp, impExtPrebid}
+			newImpData := ImpExtPrebidData{imp, impExtPrebid}
 			impData = append(impData, newImpData)
 		})
 	}
 	return
 }
 
-type ImpData struct {
+type ImpExtPrebidData struct {
 	Imp          json.RawMessage
 	ImpExtPrebid openrtb_ext.ExtImpPrebid
 }
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1335,8 +1335,8 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 		storedReqIds = []string{storedBidRequestId}
 	}
 
-	impStoredReqIds := make([]string, 0)
-	impStoredReqIdsUniqueTracker := make(map[string]struct{})
+	impStoredReqIds := make([]string, 0, len(impInfo))
+	impStoredReqIdsUniqueTracker := make(map[string]struct{}, len(impInfo))
 	for _, impData := range impInfo {
 		if impData.ImpExtPrebid.StoredRequest != nil && len(impData.ImpExtPrebid.StoredRequest.ID) > 0 {
 			storedImpId := impData.ImpExtPrebid.StoredRequest.ID
@@ -1417,11 +1417,11 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 				return nil, nil, []error{err}
 			}
 
-			echovideoAttributes := false
+			echoVideoAttributes := false
 			if impData.ImpExtPrebid.Options != nil {
-				echovideoAttributes = impData.ImpExtPrebid.Options.EchoVideoAttrs
+				echoVideoAttributes = impData.ImpExtPrebid.Options.EchoVideoAttrs
 			}
-			impExtInfoMap[impId] = exchange.ImpExtInfo{EchoVideoAttrs: echovideoAttributes, StoredImp: storedImps[impData.ImpExtPrebid.StoredRequest.ID]}
+			impExtInfoMap[impId] = exchange.ImpExtInfo{EchoVideoAttrs: echoVideoAttributes, StoredImp: storedImps[impData.ImpExtPrebid.StoredRequest.ID]}
 
 		} else {
 			resolvedImps = append(resolvedImps, impData.Imp)
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -14,7 +14,7 @@ import (
 	"time"
 
 	"github.com/buger/jsonparser"
-	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/evanphx/json-patch"
 	"github.com/gofrs/uuid"
 	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
@@ -262,8 +262,13 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	impInfo, errs := parseImpInfo(requestJson)
+	if len(errs) > 0 {
+		return nil, nil, errs
+	}
+
 	// Fetch the Stored Request data and merge it into the HTTP request.
-	if requestJson, impExtInfoMap, errs = deps.processStoredRequests(ctx, requestJson); len(errs) > 0 {
+	if requestJson, impExtInfoMap, errs = deps.processStoredRequests(ctx, requestJson, impInfo); len(errs) > 0 {
 		return
 	}
 
@@ -1317,15 +1322,11 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
 	return false, ""
 }
 
-func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte) ([]byte, map[string]exchange.ImpExtInfo, []error) {
+func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson []byte, impInfo []ImpData) ([]byte, map[string]exchange.ImpExtInfo, []error) {
 	// Parse the Stored Request IDs from the BidRequest and Imps.
 	storedBidRequestId, hasStoredBidRequest, err := getStoredRequestId(requestJson)
 	if err != nil {
 		return nil, nil, []error{err}
-	}
-	imps, impIds, idIndices, errs := parseImpInfo(requestJson)
-	if len(errs) > 0 {
-		return nil, nil, errs
 	}
 
 	// Fetch the Stored Request data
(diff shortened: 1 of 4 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -1426,29 +1443,31 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
-// 1. A list of the JSON for every Imp.
-// 2. A list of all IDs which appear at `imp[i].ext.prebid.storedrequest.id`.
-// 3. A list intended to parallel "ids". Each element tells which index of "imp[index]" the corresponding element of "ids" should modify.
-// 4. Any errors which occur due to bad requests. These should warrant an HTTP 4xx response.
-func parseImpInfo(requestJson []byte) (imps []json.RawMessage, ids []string, impIdIndices []int, errs []error) {
+// parseImpInfo parses the request JSON and returns impression and unmarshalled imp.ext.prebid
+func parseImpInfo(requestJson []byte) (impData []ImpData, errs []error) {
+
+	//impData := make([]ImpData, 0)
 	if impArray, dataType, _, err := jsonparser.Get(requestJson, "imp"); err == nil && dataType == jsonparser.Array {
```
guscarreon, line 1450: Nitpick: Do you think it'd be more useful and make for a simpler code if we use `json.Unmarshal` for the entire thing instead?
```
1446   // parseImpInfo parses the request JSON and returns impression and unmarshalled imp.ext.prebid
1447   func parseImpInfo(requestJson []byte) (impData []ImpData, errs []error) {
1448  
     +     // EXTRACT IMP EXTS     
     +     type ExtOnlyImp struct {
     +         Ext openrtb_ext.ExtImpPrebid `json:"ext"`
     +     }
     +     type requestWithExtOnlyImps struct {
     +         Imp []ExtOnlyImp `json:"imp"`
     +     }
     +     var reqOfImpsWithExts requestWithExtOnlyImps
     +     if err := json.Unmarshal(requestJson, &reqOfImpsWithExts); err != nil {
     +         return nil, append(errs, err)
     +     }
     +
     +     // EXTRACT JSON RAW IMPS
     +     type requestWithRawJsonImps struct {
     +         Imp []json.RawMessage `json:"imp"`
     +     }
     +     var reqOfRawImps requestWithRawJsonImps
     +     if err := json.Unmarshal(requestJson, &reqOfRawImps); err != nil {
     +         return nil, append(errs, err)
     +     }
     +
     +     // PUT THEM TOGETHER
     +     for i:=0;i< len(reqOfImpsWithExts); i++ {
     +         newImpData := ImpData{reqOfRawImps[i], reqOfImpsWithExts[i]}
     +         impData = append(impData, newImpData)
     +     }
1449 -     //impData := make([]ImpData, 0)
1450 -     if impArray, dataType, _, err := jsonparser.Get(requestJson, "imp"); err == nil && dataType == jsonparser.Array {
1451 -         _, err = jsonparser.ArrayEach(impArray, func(imp []byte, _ jsonparser.ValueType, _ int, err error) {
1452 -             impExtData, _, _, err := jsonparser.Get(imp, "ext", "prebid")
1453 -             var impExtPrebid openrtb_ext.ExtImpPrebid
1454 -             if impExtData != nil {
1455 -                 if err := json.Unmarshal(impExtData, &impExtPrebid); err != nil {
1456 -                     errs = append(errs, err)
1457 -                 }
1458 -             }
1459 -             newImpData := ImpData{imp, impExtPrebid}
1460 -             impData = append(impData, newImpData)
1461 -         })
1462 -     }
1463       return
1464   }
1465
endpoints/openrtb2/auction.go
```

guscarreon (commented):
Looks good. Have some nitpicks but if you deem them unnecessary, ready to approve either way. Let me know.

endpoints/openrtb2/auction.go
```diff
@@ -1426,29 +1443,31 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
-//
-// 1. A list of the JSON for every Imp.
-// 2. A list of all IDs which appear at `imp[i].ext.prebid.storedrequest.id`.
-// 3. A list intended to parallel "ids". Each element tells which index of "imp[index]" the corresponding element of "ids" should modify.
-// 4. Any errors which occur due to bad requests. These should warrant an HTTP 4xx response.
-func parseImpInfo(requestJson []byte) (imps []json.RawMessage, ids []string, impIdIndices []int, errs []error) {
+// parseImpInfo parses the request JSON and returns impression and unmarshalled imp.ext.prebid
+func parseImpInfo(requestJson []byte) (impData []ImpData, errs []error) {
+
+	//impData := make([]ImpData, 0)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Sure, removed

endpoints/openrtb2/auction.go
```diff
@@ -1317,23 +1322,33 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
...
 	// Fetch the Stored Request data
 	var storedReqIds []string
 	if hasStoredBidRequest {
 		storedReqIds = []string{storedBidRequestId}
 	}
-	storedRequests, storedImps, errs := deps.storedReqFetcher.FetchRequests(ctx, storedReqIds, impIds)
+
+	//impStoredReqIds list of all stored impression ids
+	impStoredReqIds := make([]string, 0)
+	impStoredReqIdsUnique := make(map[string]bool)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Modified, hope usage of it makes sense, please take another look

endpoints/openrtb2/auction.go
VeronikaSolovei9: This is a good idea from the first glance. I started to implement it and realize this is similar to full request unmarshalling we do later in code. 
Idea here is to find very specific part of imp.ext, which I think will be present in minority of cases. 
I honestly don't have a strong opinion about this. Need to discuss it with a team.

endpoints/openrtb2/auction_test.go
```diff
@@ -1047,40 +1139,6 @@ func TestStoredRequests(t *testing.T) {
 	}
 }
 
-func TestStoredRequestsVideoErrors(t *testing.T) {
-	deps := &endpointDeps{
```
guscarreon, line 1051: Did we move the error tests elsewhere? Or do these scenarios no longer apply?

endpoints/openrtb2/auction.go
```diff
@@ -262,8 +262,13 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	impInfo, errs := parseImpInfo(requestJson)
+	if len(errs) > 0 {
+		return nil, nil, errs
+	}
```
guscarreon, line 268: The error scenarios seem to not be covered. Can we add test scenarios that throw error?
image

endpoints/openrtb2/auction_test.go
```diff
@@ -1000,6 +1000,96 @@ func TestRefererParsing(t *testing.T) {
...
+			Imp:          json.RawMessage(`{"id": "adUnit2","ext": {"prebid": {"storedrequest": {"id": "1"},"options": {"echovideoattrs": true}},"appnexus": {"placementId": "def","trafficSourceCode": "mysite.com","reserve": null},"rubicon": null}}`),
+			ImpExtPrebid: openrtb_ext.ExtImpPrebid{StoredRequest: &openrtb_ext.ExtStoredRequest{ID: "1"}, Options: &openrtb_ext.Options{Echovideoattrs: true}},
+		},
+		{
+			Imp:          json.RawMessage(`{"ext": {"prebid": {"storedrequest": {"id": "2"},"options": {"echovideoattrs": false}}}}`),
+			ImpExtPrebid: openrtb_ext.ExtImpPrebid{StoredRequest: &openrtb_ext.ExtStoredRequest{ID: "2"}, Options: &openrtb_ext.Options{Echovideoattrs: false}},
+		},
+		{
+			Imp:          json.RawMessage(`{"id": "some-static-imp","video":{"mimes":["video/mp4"]},"ext": {"appnexus": {"placementId": "abc","position": "below"}}}`),
+			ImpExtPrebid: openrtb_ext.ExtImpPrebid{},
```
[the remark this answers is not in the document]
VeronikaSolovei9: Oh because there is no `storedrequest` in imp. We only extract ExtStoredRequest and Echovideoattrs. I added comment

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: This test is not relevant anymore, now only unmarshal error can be returned. I added test for your another comment and it covers this case.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Added

guscarreon (dismissed):
LGTM

endpoints/openrtb2/auction.go
```diff
@@ -1317,23 +1322,34 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
...
-	}
 
 	// Fetch the Stored Request data
 	var storedReqIds []string
 	if hasStoredBidRequest {
 		storedReqIds = []string{storedBidRequestId}
 	}
-	storedRequests, storedImps, errs := deps.storedReqFetcher.FetchRequests(ctx, storedReqIds, impIds)
+
+	//impStoredReqIds list of all stored impression ids
```
SyntaxNode, line 1338: IMHO this comment is not necessary as it's just describing the impStoredReqIds variable. If you believe it's necessary, please find a better name for the impStoredReqIds variable.

endpoints/openrtb2/auction.go
```diff
@@ -1317,23 +1322,34 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
...
+	impStoredReqIds := make([]string, 0)
+	impStoredReqIdsUnique := make(map[string]interface{})
```
SyntaxNode, line 1340: Consider renaming `impStoredReqIdsUnique` to `impStoredReqIdsUniqueTracker` so it's more clear the map is temporary and only used for deduplication.

SyntaxNode, line 1340: Hash Sets in Go are best represented using empty structs: `make(map[string]struct{})`. They use no memory versus some memory using booleans.

endpoints/openrtb2/auction.go
```diff
@@ -1317,23 +1322,34 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
...
+	for _, impData := range impInfo {
+		if impData.ImpExtPrebid.StoredRequest != nil && len(impData.ImpExtPrebid.StoredRequest.ID) > 0 {
+			storedImpId := impData.ImpExtPrebid.StoredRequest.ID
+			_, present := impStoredReqIdsUnique[storedImpId]
+			if !present {
+				impStoredReqIds = append(impStoredReqIds, storedImpId)
+				impStoredReqIdsUnique[storedImpId] = true
+			}
```
SyntaxNode, line 1348: Nitpick: You can compile lines 1344 and 1345:
```
if _, present := impStoredReqIdsUnique[storedImpId]; !present {
  impStoredReqIds = append(impStoredReqIds, storedImpId)
  impStoredReqIdsUnique[storedImpId] = true
}
```

endpoints/openrtb2/auction.go
```diff
@@ -1378,42 +1394,44 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	// Apply any Stored Imps, if they exist. Since the JSON Merge Patch overrides arrays,
 	// and Prebid Server defers to the HTTP Request to resolve conflicts, it's safe to
 	// assume that the request.imp data did not change when applying the Stored BidRequest.
-	impExtInfoMap := make(map[string]exchange.ImpExtInfo, len(impIds))
-	for i := 0; i < len(impIds); i++ {
-		resolvedImp, err := jsonpatch.MergePatch(storedImps[impIds[i]], imps[idIndices[i]])
+	impExtInfoMap := make(map[string]exchange.ImpExtInfo, 0)
+	resolvedImps := make([]json.RawMessage, 0)
```
SyntaxNode, line 1398: Should `impExtInfoMap` and `resolvedImps` start with a capacity of `len(impInfo)`?

endpoints/openrtb2/auction.go
```diff
@@ -1378,42 +1394,44 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+			resolvedImp, err := jsonpatch.MergePatch(storedImps[impData.ImpExtPrebid.StoredRequest.ID], impData.Imp)
 
-		if err != nil {
-			hasErr, Err := getJsonSyntaxError(imps[idIndices[i]])
-			if hasErr {
-				err = fmt.Errorf("Invalid JSON in Imp[%d] of Incoming Request: %s", i, Err)
-			} else {
-				hasErr, Err = getJsonSyntaxError(storedImps[impIds[i]])
+			if err != nil {
+				hasErr, Err := getJsonSyntaxError(impData.Imp)
```
SyntaxNode, line 1404: I think this is a good time to fix the improper use of `Err` as a sneaky way of avoiding shadowing with the `err` variable.

endpoints/openrtb2/auction.go
```diff
@@ -1378,42 +1394,44 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+
+		} else {
+			resolvedImps = append(resolvedImps, impData.Imp)
 		}
+
 	}
-	if len(impIds) > 0 {
-		newImpJson, err := json.Marshal(imps)
+	if len(resolvedImps) > 0 {
+		newImpJson, err := json.Marshal(resolvedImps)
```
SyntaxNode, line 1434: I'll be honest that's it a bit hard for me to follow the changes in this function. Why is this now using `resolvedImps` when before it was using `imps`?

endpoints/openrtb2/auction.go
```diff
@@ -1426,29 +1444,30 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
-			imps = append(imps, imp)
-			i++
+			newImpData := ImpData{imp, impExtPrebid}
+			impData = append(impData, newImpData)
 		})
 	}
 	return
 }
 
+type ImpData struct {
```
SyntaxNode, line 1466: Let's consider a new name for `ImpData`. It's too generic IMHO.

openrtb_ext/imp.go
```diff
@@ -14,9 +14,15 @@ type ExtImpPrebid struct {
...
+	Options *Options `json:"options,omitempty"`
 }
 
 // ExtStoredRequest defines the contract for bidrequest.imp[i].ext.prebid.storedrequest
 type ExtStoredRequest struct {
 	ID string `json:"id"`
 }
+
+type Options struct {
+	Echovideoattrs bool `json:"echovideoattrs"`
```
SyntaxNode, line 27: Nitpick: The variable name can have proper capitalization and naming, ie: `EchoVideoAttributes `

openrtb_ext/imp.go
```diff
@@ -14,9 +14,15 @@ type ExtImpPrebid struct {
 
 	// Bidder is the preferred approach for providing paramters to be interepreted by the bidder's adapter.
 	Bidder map[string]json.RawMessage `json:"bidder"`
+
+	Options *Options `json:"options,omitempty"`
```
SyntaxNode, line 18: Did we not have this defined before?

SyntaxNode (commented):
Did you consult with @hhhjort for considering if the RequestWrapper would be a good fit in this case?

endpoints/openrtb2/auction.go
VeronikaSolovei9: `Err` is a string description of error if it's present which I believe should be a nicer representation of `err.Error()`. 
It's handled further in code like this:
```
hasErr, Err := getJsonSyntaxError(impData.Imp)
				if hasErr {
					err = fmt.Errorf("Invalid JSON in Imp[%d] of Incoming Request: %s", i, Err)
...
```
Should it be modified to something like this:
```
err = fmt.Errorf("Invalid JSON in Imp[%d] of Incoming Request: %s, original error: %s", i, Err, err.Error())
```

endpoints/openrtb2/auction.go
VeronikaSolovei9: Renamed to `ImpExtPrebidData`.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Cannot init  `resolvedImps` array with `len(impInfo)` because it fails if some or all impressions in request don't have stored request.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Comment removed

endpoints/openrtb2/auction.go
VeronikaSolovei9: Renamed

VeronikaSolovei9: Yes, modified

endpoints/openrtb2/auction.go
VeronikaSolovei9: Done!

openrtb_ext/imp.go
VeronikaSolovei9: Sure, renamed.

openrtb_ext/imp.go
VeronikaSolovei9: Nothing I can think about.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Idea here is to simplify extraction of stored impression id. Before there were 3 structures in code to do this:
- `impIds` - is not imp ids, it's stored imp ids
- `imps` - impressions itself
- `idIndices` - specifies what imp index has stored impId, this is extra variable to support requests where not all imps have stored imps.

So we had to do this `imps[idIndices[i]] = resolvedImp` in order to find what impression had what id. 
Please refer to this comment: https://github.com/prebid/prebid-server/pull/1889#discussion_r681340201

Also part of this PR is a base for future stored responses implementation. 
Here we only have 

```
type ExtImpPrebid struct {
	// StoredRequest specifies which stored impression to use, if any.
	StoredRequest *ExtStoredRequest `json:"storedrequest"`

	// IsRewardedInventory is a signal intended for video impressions. Must be 0 or 1.
	IsRewardedInventory int8 `json:"is_rewarded_inventory"`

	// Bidder is the preferred approach for providing paramters to be interepreted by the bidder's adapter.
	Bidder map[string]json.RawMessage `json:"bidder"`

	Options *Options `json:"options,omitempty"`
}
``` 
which is used as a part of impression.ext parsing to this struct:
```
type ImpExtPrebidData struct {
	Imp          json.RawMessage
	ImpExtPrebid openrtb_ext.ExtImpPrebid
}
```
For Stored responses implementation ExtImpPrebid will be extended with 

```
// ExtStoredAuctionResponse specifies which stored auction response to use, if any.
ExtStoredAuctionResponse *ExtStoredAuctionResponse `json:"storedauctionresponse,omitempty"`

// ExtStoredBidResponse specifies which stored bid responseto use, if any.
ExtStoredBidResponse []*ExtStoredBidResponse `json:"storedbidresponse,omitempty"`
```

Hope this makes sense

VeronikaSolovei9:
> Did you consult with @hhhjort for considering if the RequestWrapper would be a good fit in this case?

No, I didn't. 
Data we extract here is not a part. of openRTB request. It's a part of imp[].ext.prebid.  Not sure how to consolidate it to existing request wrapper and if this data needs to be there. 

@hhhjort please advice

endpoints/openrtb2/auction.go
hhhjort: Why does it fail? Did you create the `resolvedImps` with a length of 0 and a capacity of `len(impInfo)`? If you just use `make([]json.RawMessage, len(impInfo))` it will create an array with a length `len(impInfo)` with nil jsons for all the entries. You want `make([]json.RawMessage, 0, len(impInfo))` which will create a zero length array, with room preallocated for all the possible entries, so you will never have to reallocate memory as you append to the array.

hhhjort:
This would be good for the wrapper when it supports the `Imp.ext`s. It is probably a bit much to cram that support into this PR.

endpoints/openrtb2/auction.go
SyntaxNode: @hhhjort explains it well. Setting the length as the third parameter will preallocate enough memory for the worst case scenario such that Go doesn't need to grow the underlying array. This uses a bit more memory, but saves on cpu and memory by avoiding potential growth logic which would be automatically triggered by using `append`.

endpoints/openrtb2/auction.go
SyntaxNode: Thank you. I think there's still room to replace "Data" with something more specific, but we can revisit that in a future refactor pass. This works for me.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Renamed to `errMessage`

endpoints/openrtb2/auction.go
VeronikaSolovei9: Oh right, my bad. Fixed.

guscarreon (approved):
LGTM

endpoints/openrtb2/auction.go
```diff
@@ -1317,23 +1322,32 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
...
 
 	// Fetch the Stored Request data
 	var storedReqIds []string
 	if hasStoredBidRequest {
 		storedReqIds = []string{storedBidRequestId}
 	}
-	storedRequests, storedImps, errs := deps.storedReqFetcher.FetchRequests(ctx, storedReqIds, impIds)
+
+	impStoredReqIds := make([]string, 0)
+	impStoredReqIdsUniqueTracker := make(map[string]struct{})
```
SyntaxNode, line 1339: We can specify the capacities of these data structures to avoid potential memory reallocation:
```
impStoredReqIds := make([]string, 0, len(impInfo))
impStoredReqIdsUniqueTracker := make(map[string]struct{}, len(impInfo))
```

endpoints/openrtb2/auction.go
```diff
@@ -1378,42 +1392,44 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
-			return nil, nil, []error{err}
-		}
-		if storedImps[impIds[i]] != nil {
-			impExtInfoMap[impId] = exchange.ImpExtInfo{EchoVideoAttrs: includeVideoAttributes, StoredImp: storedImps[impIds[i]]}
+			impId, err := jsonparser.GetString(resolvedImp, "id")
+			if err != nil {
+				return nil, nil, []error{err}
+			}
+
+			echovideoAttributes := false
```
SyntaxNode, line 1420: Nitpick: You can also use an upper case "v" -> "echoVideoAttributes"

endpoints/openrtb2/auction.go
VeronikaSolovei9: Done!

endpoints/openrtb2/auction.go
VeronikaSolovei9: Renamed

SyntaxNode (approved):

guscarreon (approved):
LGTM

PR #1962, opened 2021-08-13
Yieldlab: Fix ad unit handling

BertiBauer:
Fixes a bug where when requesting bids for more than one ad unit, the resulting bids could be associated with the wrong ad unit.

This bug was due to the returned bids being dealt with in a for-loop whose index was being used to step through the configured ad units (see lines 215+232). If the returned bids for the ad units were not in the same order as the configured ad units, they would be matched wrongly.  

The fix uses the respective Yieldlab adslot ID to find the correct impression and ad unit for a given bid.

### DIFF
--- a/adapters/yieldlab/yieldlab.go
+++ b/adapters/yieldlab/yieldlab.go
@@ -212,7 +212,7 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 		Bids:     []*adapters.TypedBid{},
 	}
 
-	for i, bid := range bids {
+	for _, bid := range bids {
 		width, height, err := splitSize(bid.Adsize)
 		if err != nil {
 			return nil, []error{err}
@@ -225,22 +225,24 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 			}
 		}
 
+		imp := a.findImp(bid.ID, internalRequest.Imp)
+
 		var bidType openrtb_ext.BidType
 		responseBid := &openrtb2.Bid{
 			ID:     strconv.FormatUint(bid.ID, 10),
 			Price:  float64(bid.Price) / 100,
-			ImpID:  internalRequest.Imp[i].ID,
+			ImpID:  imp.ID,
 			CrID:   a.makeCreativeID(req, bid),
 			DealID: strconv.FormatUint(bid.Pid, 10),
 			W:      int64(width),
 			H:      int64(height),
 		}
 
-		if internalRequest.Imp[i].Video != nil {
+		if imp.Video != nil {
 			bidType = openrtb_ext.BidTypeVideo
-			responseBid.NURL = a.makeAdSourceURL(internalRequest, req, bid)
+			responseBid.AdM = a.makeAdSourceURL(internalRequest, req, bid)
 
-		} else if internalRequest.Imp[i].Banner != nil {
+		} else if imp.Banner != nil {
 			bidType = openrtb_ext.BidTypeBanner
 			responseBid.AdM = a.makeBannerAdSource(internalRequest, req, bid)
 		} else {
@@ -259,12 +261,34 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 
 func (a *YieldlabAdapter) findBidReq(adslotID uint64, params []*openrtb_ext.ExtImpYieldlab) *openrtb_ext.ExtImpYieldlab {
 	slotIdStr := strconv.FormatUint(adslotID, 10)
+
 	for _, p := range params {
 		if p.AdslotID == slotIdStr {
 			return p
 		}
 	}
 
+	return nil
+}
+
+func (a *YieldlabAdapter) findImp(adslotID uint64, imps []openrtb2.Imp) *openrtb2.Imp {
+	slotIdStr := strconv.FormatUint(adslotID, 10)
+
+	for _, imp := range imps {
+		bidderExt := new(adapters.ExtImpBidder)
+		if err := json.Unmarshal(imp.Ext, bidderExt); err != nil {
+			continue
+		}
+
+		yieldlabExt := new(openrtb_ext.ExtImpYieldlab)
+		if err := json.Unmarshal(bidderExt.Bidder, yieldlabExt); err != nil {
+			continue
+		}
+
+		if yieldlabExt.AdslotID == slotIdStr {
+			return &imp
+		}
+	}
 	return nil
 }
 
--- a/adapters/yieldlab/yieldlab.go
+++ b/adapters/yieldlab/yieldlab.go
@@ -212,6 +212,14 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 		Bids:     []*adapters.TypedBid{},
 	}
 
+	adslotToImpMap := make(map[string]*openrtb2.Imp)
+	for i := 0; i < len(internalRequest.Imp); i++ {
+		adslotID := a.extractAdslotID(internalRequest.Imp[i])
+		if internalRequest.Imp[i].Video != nil || internalRequest.Imp[i].Banner != nil {
+			adslotToImpMap[adslotID] = &internalRequest.Imp[i]
+		}
+	}
+
 	for _, bid := range bids {
 		width, height, err := splitSize(bid.Adsize)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mansinahar:
@BertiBauer Can you please add JSON tests for these changes?

BertiBauer:
@mansinahar Thank you for your input and apologies it took so long - I just added multiple_impressions.json to the exemplary tests which tests for the changes made in this branch.

adapters/yieldlab/yieldlab.go
```diff
@@ -225,22 +225,24 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 			}
 		}
 
+		imp := a.findImp(bid.ID, internalRequest.Imp)
```
guscarreon, line 228: The fact that the `a.findImp(bid.ID, internalRequest.Imp)` call traverses all of the imps in `internalRequest.Imp` at every loop makes `MakeBids` to have a time complexity of `O(nm)` where `n` is the number of bids and `m` the number of imps inside `internalRequest.Imp`. If we expect bids to come with unique `bid.ID`'s maybe we could simplify to `O(n+m)` at the expense of some memory by creating a map. (Prebid server has some memory to spare compared to CPU time). Do you think something like this would be a godd idea?
```
210     bidderResponse := &adapters.BidderResponse{
211         Currency: currency.EUR.String(),
212         Bids:     []*adapters.TypedBid{},
213     }
214
    +   // At a expense of O(m) memory
    +   var adslotToImpMap map[string]*openrtb2.Imp
    +   for i:=0; i<len(internalRequest.Imp); i++ {
    +       // extract adslotID from Ext
    +       adslotID := someFunctionToExtractAdslotID(internalRequest.Imp[i])
    +       if internalRequest.Imp[i].Video != nil || internalRequest.Imp[i].Banner != nil {
    +           // we could only allow Video or Banners into the map 
    +           adslotToImpMap[adslotID] = &internalRequest.Imp[i]
    +       }
    +   }
    + 
215     for _, bid := range bids {
216         width, height, err := splitSize(bid.Adsize)
217         if err != nil {
218             return nil, []error{err}
219         }
220
221         req := a.findBidReq(bid.ID, params)
222         if req == nil {
223             return nil, []error{
224                 fmt.Errorf("failed to find yieldlab request for adslotID %v. This is most likely a programming issue", bid.ID),
225             }
226         }
227
228  -      imp := a.findImp(bid.ID, internalRequest.Imp)
229
     +      // Does `bid.ID` match with any element in our map?
     +      if imp, exists := goMap[bid.ID]; !exists {
     +          continue
     +      } else {
     +          // Imp matches, do all the rest
                 . 
                 . 
                 . 
 adapters/yieldlab/yieldlab.go
```
Unless of course none of this makes sense in your context. Maybe you know you expect responses to only come with one bid, or imp arrays with only one imp. You decide I'm good either way.

guscarreon (commented):
Code looks pretty good. Have some minor nitpicks and a comment.

SyntaxNode:
@BertiBauer Friendly reminder to review the PR feedback.

adapters/yieldlab/yieldlab.go
```diff
@@ -225,22 +225,24 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
...
-		if internalRequest.Imp[i].Video != nil {
+		if imp.Video != nil {
 			bidType = openrtb_ext.BidTypeVideo
 			responseBid.NURL = a.makeAdSourceURL(internalRequest, req, bid)
 
-		} else if internalRequest.Imp[i].Banner != nil {
+		} else if imp.Banner != nil {
 			bidType = openrtb_ext.BidTypeBanner
 			responseBid.AdM = a.makeBannerAdSource(internalRequest, req, bid)
 		} else {
```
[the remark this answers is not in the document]
BertiBauer: Hi guscarreon, thanks for your nitpicking - highly appreciated (seriously!)! Re this nitpick, fair point. However the bid should always be either ```imp.Video``` or ```imp.Banner```, as the publisher needs to configure this beforehand in our system (see the next point as well). In case the publisher configures an adslot as a type that is not supported yet by our PBS adapter and uses it in PBS, it should become obvious very quickly and then be corrected.

adapters/yieldlab/yieldlab.go
```diff
@@ -225,22 +225,24 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
...
-			ImpID:  internalRequest.Imp[i].ID,
+			ImpID:  imp.ID,
 			CrID:   a.makeCreativeID(req, bid),
 			DealID: strconv.FormatUint(bid.Pid, 10),
 			W:      int64(width),
 			H:      int64(height),
 		}
 
-		if internalRequest.Imp[i].Video != nil {
+		if imp.Video != nil {
```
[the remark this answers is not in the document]
BertiBauer: This scenario isn't possible - a Yieldlab adslot always has a unique ID and a statically pre-defined type (```Banner```, ```Video```etc.) in our system, so the above situation can never arise.

adapters/yieldlab/yieldlab.go
BertiBauer: This is a very good one - thanks again not only for bringing this up, but in particular for giving very helpful pointers in how this could be implemented! I wasn't particularly happy with our original solution but as I have very little experience with Go so far, it was all I could come up with. I have just updated this PR with another commit which implements the suggested solution from above. Let me know if that looks good to you!

guscarreon (approved):
Looks good to me. @BertiBauer ghank you for addressing our feedback

adapters/yieldlab/yieldlab.go
```diff
@@ -212,7 +212,15 @@ func (a *YieldlabAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 		Bids:     []*adapters.TypedBid{},
 	}
 
-	for i, bid := range bids {
+	adslotToImpMap := make(map[string]*openrtb2.Imp)
+	for i := 0; i < len(internalRequest.Imp); i++ {
+		adslotID := a.extractAdslotID(internalRequest.Imp[i])
```
hhhjort, line 217: It is possible that `extractAdslotID()` returns a nil value if there is an issue unmarshalling the bidder parameters. It also seems that your bidding depends on `adslotID` being unique to each imp in the request. But I don't see anywhere you enforce this in `MakeRequests()`. You probably want to throw out any imps with duplicate adslots in the `MakeRequests()` stage, and return either a warning or an error about the issue. A nil adslot will likely mean that the imp will get no. bids anyway, and if it does will just be skipped below so again not a fatal issue.

The possibility of duplicate adslots in Imps does seem like something that should be addressed, as it can otherwise lead to bids being associated with the wrong imp.

adapters/yieldlab/yieldlab.go
```diff
@@ -268,6 +281,14 @@ func (a *YieldlabAdapter) findBidReq(adslotID uint64, params []*openrtb_ext.ExtI
 	return nil
 }
 
+func (a *YieldlabAdapter) extractAdslotID(internalRequestImp openrtb2.Imp) string {
+	bidderExt := new(adapters.ExtImpBidder)
+	json.Unmarshal(internalRequestImp.Ext, bidderExt)
```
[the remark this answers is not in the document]
BertiBauer: See comment below please

adapters/yieldlab/yieldlab.go
BertiBauer: Hi @hhhjort, thanks for your input!

You are correct that our bidding depends on adslotID being unique to each imp in the request and that we don't enforce this within Prebid. This is enforced from our servers though: When we get a bid request for adslot IDs 1, 1, 1 and 2, this will actualy be flattened to a request for adslot IDs 1 and 2 in our servers, and so only one bid per adslotID will be returned.

This is not specific to how we handle Prebid by the way, but also applies to any header bidding integration with Yieldlab and is something that is being pointed out to every publisher once they start working with us.

Does this ease your concerns?

hhhjort: There is the issue that you map the bids back into the imps via the adslot ID So either when you flatten on the servers, you need to pick an ad that should be valid for all the relevant imps, or you run the danger of delivering a bid for an imp that that bid is not valid for. (perhaps a wrong size banner if the adslots are not coded for a specific size.) If the bid will be valid for all imps with the same adslot, then I don't have an issue.

BertiBauer: `If the bid will be valid for all imps with the same adslot, then I don't have an issue.`

Yes, that will be the case! Thanks for all your help!

hhhjort (approved):
LGTM

PR #1964, opened 2021-08-16
New Adapter: IQZone

IQZoneAdx:
Doc - https://github.com/prebid/prebid.github.io/pull/3210

### DIFF
--- a/adapters/iqzone/iqzone.go
+++ b/adapters/iqzone/iqzone.go
@@ -43,13 +43,6 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
 		return nil, nil
 	}
 
-	if responseData.StatusCode == http.StatusBadRequest {
-		err := &errortypes.BadInput{
-			Message: "Unexpected status code: 400. Bad request from publisher. Run with request.debug = 1 for more info.",
-		}
-		return nil, []error{err}
-	}
-
 	if responseData.StatusCode != http.StatusOK {
 		err := &errortypes.BadServerResponse{
 			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info.", responseData.StatusCode),
@@ -65,11 +58,15 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
 	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(request.Imp))
 	bidResponse.Currency = response.Cur
 	for _, seatBid := range response.SeatBid {
-		for _, bid := range seatBid.Bid {
-			bid := bid // pin https://github.com/kyoh86/scopelint#whats-this
+		for i := range seatBid.Bid {
+			bidType, err := getMediaTypeForImp(seatBid.Bid[i].ImpID, request.Imp)
+			if err != nil {
+				return nil, []error{err}
+			}
+
 			b := &adapters.TypedBid{
-				Bid:     &bid,
-				BidType: getMediaTypeForImp(bid.ImpID, request.Imp),
+				Bid:     &seatBid.Bid[i],
+				BidType: bidType,
 			}
 			bidResponse.Bids = append(bidResponse.Bids, b)
 		}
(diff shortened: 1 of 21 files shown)


openrtb_ext/imp_iqzone.go
```diff
@@ -0,0 +1,5 @@
+package openrtb_ext
+
+type ImpExtFoo struct {
```
guscarreon, line 3: Even if it's not being used, please rename to something less generic, `ImpExtIqzone` maybe

static/bidder-params/iqzone.json
```diff
@@ -0,0 +1,15 @@
...
+    "type": "object",
+  
+    "properties": {
+      "placementId": {
+        "type": "string",
+        "description": "Placement ID"
+      }
+    },
+  
+    "required": ["placementId"]
```
guscarreon, line 14: Just to double-check: Are we sure every request will come with a `"placementId"` parameter? If it's set as `"required"`, an incoming request with a missing `"placementId"` will get rejected and Prebid Server won't send a bid request to any bidder. So adding a required field can break the entire setup of a publisher.

adapters/iqzone/params_test.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+	}
+}
+
+var validParams = []string{
+	`{"placementId": "test"}`,
+}
+
+var invalidParams = []string{
+	`{"placementId": 42}`,
```
guscarreon, line 41: If you decide to keep the `"placementId"` listed in the `"required"` array inside `static/bidder-params/iqzone.json`, can we please add a couple more test cases in order to assert the request will be invalid when `"placementId"` isn't found in the `Ext`?
```
40   var invalidParams = []string{
41       `{"placementId": 42}`,
   +     `{}`,
   +     `{"someOtherParam": "value"}`,
42   }
adapters/iqzone/params_test.go
```

adapters/iqzone/iqzone.go
```diff
@@ -0,0 +1,94 @@
...
+	var response openrtb2.BidResponse
+	if err := json.Unmarshal(responseData.Body, &response); err != nil {
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(request.Imp))
+	bidResponse.Currency = response.Cur
+	for _, seatBid := range response.SeatBid {
+		for _, bid := range seatBid.Bid {
+			bid := bid // pin https://github.com/kyoh86/scopelint#whats-this
```
guscarreon, line 69: Good catch, we can remove the comment if you want

adapters/iqzone/iqzone.go
```diff
@@ -0,0 +1,94 @@
...
+func getMediaTypeForImp(impID string, imps []openrtb2.Imp) openrtb_ext.BidType {
+	mediaType := openrtb_ext.BidTypeBanner
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner == nil && imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			if imp.Banner == nil && imp.Video == nil && imp.Native != nil {
+				mediaType = openrtb_ext.BidTypeNative
+			}
```
guscarreon, line 89: Is the `imp.Banner == nil && imp.Video == nil && imp.Native == nil` scenario possible? If so, should `getMediaTypeForImp` return an error instead?

guscarreon (changes requested):
Left a few comments. Also, there is already a pull request to remove the old race condition test framework and an additional PR is under development to update the documents. We apologize for not having the documentation up to date yet. In the meantime, can we please remove the following files?
```
iqzone/iqzonetest/params/race/banner.json
iqzone/iqzonetest/params/race/native.json
iqzone/iqzonetest/params/race/video.json
```
These were required under the old data race tests and are no longer necessary. In regards of test case coverage your adapter looks file. It only seems to be missing the following scenarios inside `MakeBids`:
image
Can we please add a couple more JSON tests under `iqzone/iqzonetest/supplemental/` to cover them?

openrtb_ext/imp_iqzone.go
SyntaxNode: This just looks like a copy/paste error from the docs. Easy fix. Thank you @IQZoneAdx for faithfully following our new bid adapter guide. :)

adapters/iqzone/iqzone.go
SyntaxNode: This is a copy/paste from the docs. A clearer way of writing this loop could be:
```
for i := range seatBid.Bid {
  b := &adapters.TypedBid{
    Bid:     &seatBid.Bid[i],
    BidType: getMediaTypeForImp(bid.ImpID, request.Imp),
  }
}
```
This does look better IMHO and doesn't require extra explanation via the pin link. I'll update the docs soon with this approach.

IQZoneAdx:
@guscarreon @SyntaxNode thanks for the review. We fixed all the problems

guscarreon (approved):
Looks good to me. @IQZoneAdx thank you for addressing the feedback and following the documentation steps so thoroughly.

SyntaxNode (approved):

PR #1965, opened 2021-08-17
Vrtcal: Added Video(VAST) Support

vrtcal-dev:
Docs PR: https://github.com/prebid/prebid.github.io/pull/3214

### DIFF
--- a/adapters/vrtcal/vrtcal.go
+++ b/adapters/vrtcal/vrtcal.go
@@ -68,6 +68,7 @@ func (a *VrtcalAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalR
 
 	bidResponse := adapters.NewBidderResponseWithBidsCapacity(1)
 
+	var errs []error
 	for _, sb := range bidResp.SeatBid {
 		for i := range sb.Bid {
 			bidType, err := getReturnTypeForImp(sb.Bid[i].ImpID, internalRequest.Imp)
@@ -76,10 +77,12 @@ func (a *VrtcalAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalR
 					Bid:     &sb.Bid[i],
 					BidType: bidType,
 				})
+			} else {
+				errs = append(errs, err)
 			}
 		}
 	}
-	return bidResponse, nil
+	return bidResponse, errs
 
 }
 
--- a/adapters/vrtcal/vrtcal.go
+++ b/adapters/vrtcal/vrtcal.go
@@ -70,9 +70,14 @@ func (a *VrtcalAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalR
 
 	for _, sb := range bidResp.SeatBid {
 		for i := range sb.Bid {
+			bidType, err := getReturnTypeForImp(sb.Bid[i].ImpID, internalRequest.Imp)
+			if err != nil {
+				return nil, []error{err}
+			}
+
 			bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
 				Bid:     &sb.Bid[i],
-				BidType: "banner",
+				BidType: bidType,
 			})
 		}
 	}
@@ -86,4 +91,27 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
 		endpoint: config.Endpoint,
 	}
 	return bidder, nil
+}
+
+func getReturnTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType, error) {
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner != nil {
+				return openrtb_ext.BidTypeBanner, nil
+			}
+
+			if imp.Video != nil {
+				return openrtb_ext.BidTypeVideo, nil
+			}
+
+			return "", &errortypes.BadServerResponse{
+				Message: fmt.Sprintf("Unsupported return type for ID: \"%s\"", impID),
+			}
+		}
+	}
+
+	//Failsafe default in case impression ID is not found
+	return "", &errortypes.BadServerResponse{
+		Message: fmt.Sprintf("Failed to find impression for ID: \"%s\"", impID),
+	}
 }
--- a/adapters/vrtcal/vrtcal.go
+++ b/adapters/vrtcal/vrtcal.go
@@ -71,14 +71,12 @@ func (a *VrtcalAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalR
 	for _, sb := range bidResp.SeatBid {
 		for i := range sb.Bid {
 			bidType, err := getReturnTypeForImp(sb.Bid[i].ImpID, internalRequest.Imp)
-			if err != nil {
-				return nil, []error{err}
+			if err == nil {
+				bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
+					Bid:     &sb.Bid[i],
+					BidType: bidType,
+				})
 			}
-
-			bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
-				Bid:     &sb.Bid[i],
-				BidType: bidType,
-			})
 		}
 	}
 	return bidResponse, nil
--- a/adapters/vrtcal/vrtcaltest/exemplary/simple-video.json
+++ b/adapters/vrtcal/vrtcaltest/exemplary/simple-video.json
@@ -1,0 +1,96 @@
+{
+  "mockBidRequest": {
+    "id": "test-request-id",
+    "imp": [
+      {
+        "id": "test-imp-id",
+        "video": {
+                "mimes": ["video/mp4"],
+                "protocols": [2, 5],
+                "w": 300,
+                "h": 250
+	},
+        "ext": {
+          "bidder": {
+            "Just_an_unused_vrtcal_param": "unused_data_for_this"
+          }
+        }
+      }
+    ],
+    "app": {
+      "id": "fake-app-id"
+    }
+  },
+  "httpCalls": [
+    {
+      "expectedRequest": {
+        "uri": "http://rtb.vrtcal.com/bidder_prebid.vap?ssp=1804",
+        "body": {
+          "id": "test-request-id",
+          "imp": [
+            {
+              "id": "test-imp-id",
+	      "video": {
+		"mimes": ["video/mp4"],
+		"protocols": [2, 5],
+		"w": 300,
+		"h": 250
+	      },
+
+	      "ext": {
+		  "bidder": {
+		    "Just_an_unused_vrtcal_param": "unused_data_for_this"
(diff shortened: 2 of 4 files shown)


adapters/vrtcal/vrtcal.go
```diff
@@ -87,3 +92,26 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
 	}
 	return bidder, nil
 }
+
+func getReturnTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType, error) {
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner != nil {
```
mansinahar, line 99: This is actually considered as an anti-pattern. Consider a request with multiple formats, that is `video` and `banner` both set. Is your bidder service confirmed to always return a `banner` bid? If not necessarily then you might be assigning the wrong bid type here and that will cause the ad to not be properly rendered. Does your bidder service perhaps communicate the bid type in the bid extension? If so, we highly recommend you to use that to set the bid type.

mansinahar:
Can you please put a doc PR to include the new supported `vide` type: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/vrtcal.md ?

vrtcal-dev:
> Can you please put a doc PR to include the new supported `video` type: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/vrtcal.md ?

vrtcal-dev:
-----Mansinihar: This is actually considered as an anti-pattern. Consider a request with multiple formats, that is video and banner both set. Is your bidder service confirmed to always return a banner bid? If not necessarily then you might be assigning the wrong bid type here and that will cause the ad to not be properly rendered. Does your bidder service perhaps communicate the bid type in the bid extension? If so, we highly recommend you to use that to set the bid type.--------------
=========================================


Thanks for the feedback mansinihar. Our backend RTB platform only accepts single impression requests and responds with a single impression response 100% of the time so there is no possible ambiguity of bid type returned. So, this won't be an issue with our platform. Also, I've updated https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/vrtcal.md to include video in a pull request. Thanks again.

mansinahar:
@vrtcal-dev Thanks for that clarification. However, a single impression could have both video and banner set, making it a multi-format request. This is different from multi-impression requests. Also, just looking at your adapter, code I don't see the requests to your backend being broken down for each impression which means if a publisher sends Prebid Server a request with multiple impressions, that gets forwarded as is to your backend.. Would your backend consider this as an invalid request and drop it or just bid for the first impression in the list and ignore others? In either case, if your backend only accepts single impression requests, you might wanna add logic in your Prebid Server adapter code to make multiple requests to your backend server for every impression in a request or else you might be losing on publisher inventory. You can do this as a separate PR if you want to but I just wanted to bring this to your attention.

vrtcal-dev:
> @vrtcal-dev Thanks for that clarification. However, a single impression could have both video and banner set, making it a multi-format request. This is different from multi-impression requests. Also, just looking at your adapter, code I don't see the requests to your backend being broken down for each impression which means if a publisher sends Prebid Server a request with multiple impressions, that gets forwarded as is to your backend.. Would your backend consider this as an invalid request and drop it or just bid for the first impression in the list and ignore others? In either case, if your backend only accepts single impression requests, you might wanna add logic in your Prebid Server adapter code to make multiple requests to your backend server for every impression in a request or else you might be losing on publisher inventory. You can do this as a separate PR if you want to but I just wanted to bring this to your attention.

-------------------------------------------------------------------

Yes, our platform will only bid on the first impressions as we've always put this controlled spec on our backend servers...Yes possibly in the future we will support multiple impressions in a single request, but this is not in our plans at the moment. If, we do decide on such an enhancement later, I'll definitely add in another PR...Thanks again...

guscarreon (dismissed):
There is already a pull request to remove the old race condition test framework and an additional PR is under development to update the documents. We apologize for not having the documentation up to date yet. In the meantime can we please remove the following files?
```
iqzone/iqzonetest/params/race/banner.json
iqzone/iqzonetest/params/race/native.json
iqzone/iqzonetest/params/race/video.json
```

mansinahar:
@vrtcal-dev Sounds good! Just want to bring this to your attention that Prebid Server has the capability to make multiple requests to a bidder for multiple imps in a single incoming request. So say, if a request to Prebid Server comes in with multiple imps, your adapter's `MakeRequests` function can split that request into multiple requests to your backend service. This way you won't lose out on publisher inventory. You can refer to the `Impression Splitting` example here: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#makerequests if you want to add that support in the future.

vrtcal-dev:
> @vrtcal-dev Sounds good! Just want to bring this to your attention that Prebid Server has the capability to make multiple requests to a bidder for multiple imps in a single incoming request. So say, if a request to Prebid Server comes in with multiple imps, your adapter's `MakeRequests` function can split that request into multiple requests to your backend service. This way you won't lose out on publisher inventory. You can refer to the `Impression Splitting` example here: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#makerequests if you want to add that support in the future.

------------------------
Thank you. I've noted for future consideration.

adapters/vrtcal/vrtcal.go
```diff
@@ -87,3 +92,26 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
+			}
+
+			return "", &errortypes.BadServerResponse{
+				Message: fmt.Sprintf("Unsupported return type for ID: \"%s\"", impID),
+			}
+		}
+	}
+
+	//Failsafe default in case impression ID is not found
+	return "", &errortypes.BadServerResponse{
```
mansinahar, line 114: Can you please add test coverage for this case here?

mansinahar (commented):
Looks pretty straightforward to me. Just have one comment to add more test coverage

adapters/vrtcal/vrtcal.go
vrtcal-dev: > Can you please add test coverage for this case here?

I've added supplemental test coverage the failsafe unmatched impression ID. Thanks.

hhhjort:
> > @vrtcal-dev Thanks for that clarification. However, a single impression could have both video and banner set, making it a multi-format request. This is different from multi-impression requests. Also, just looking at your adapter, code I don't see the requests to your backend being broken down for each impression which means if a publisher sends Prebid Server a request with multiple impressions, that gets forwarded as is to your backend.. Would your backend consider this as an invalid request and drop it or just bid for the first impression in the list and ignore others? In either case, if your backend only accepts single impression requests, you might wanna add logic in your Prebid Server adapter code to make multiple requests to your backend server for every impression in a request or else you might be losing on publisher inventory. You can do this as a separate PR if you want to but I just wanted to bring this to your attention.
> 
> Yes, our platform will only bid on the first impressions as we've always put this controlled spec on our backend servers...Yes possibly in the future we will support multiple impressions in a single request, but this is not in our plans at the moment. If, we do decide on such an enhancement later, I'll definitely add in another PR...Thanks again...

This does not address the comment about multiformat impressions. If you receive a request for a single imp, with both the `banner` and `video` objects filled in, this is a request that allows for both banner and video bids. According to your logic in choosing the bid type, you will always bid with a banner bid even if both banners and videos are valid. Some other adapters specify they will always choose video as video typically involves higher CPMs. We just want to confirm that, given a choice between the two, your bidder will always choose banner.

vrtcal-dev:
> given a choice between the two, your bidder will always choose banner.

Correct, our bidder will always choose banner if a video object is also specified within an impression request.

adapters/vrtcal/vrtcal.go
```diff
@@ -70,9 +70,14 @@ func (a *VrtcalAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalR
 
 	for _, sb := range bidResp.SeatBid {
 		for i := range sb.Bid {
+			bidType, err := getReturnTypeForImp(sb.Bid[i].ImpID, internalRequest.Imp)
+			if err != nil {
+				return nil, []error{err}
```
mansinahar, line 75: Sorry, I missed this in my earlier review. In case there's an error getting the type of one of the bids, are you sure you want to return an error and discard the rest of the bids? It might be better to rather append the error in an error slice and move on to the next bid by adding a `continue` here. 

Thinking more about this, I realize that you mentioned your backend service only bids on a single impression for a single request so the response likely is gonna have a single bid as well but I'd still recommend adding a `continue` here and appending the error to an error slice as it is easy to do and this way you won't have to update this logic in the future even when your backend service does start supporting multiple impressions and returning multiple bids

vrtcal-dev: > Sorry, I missed this in my earlier review. In case there's an error getting the type of one of the bids, are you sure you want to return an error and discard the rest of the bids? It might be better to rather append the error in an error slice and move on to the next bid by adding a `continue` here.
> 
> Thinking more about this, I realize that you mentioned your backend service only bids on a single impression for a single request so the response likely is gonna have a single bid as well but I'd still recommend adding a `continue` here and appending the error to an error slice as it is easy to do and this way you won't have to update this logic in the future even when your backend service does start supporting multiple impressions and returning multiple bids

Thanks Mansinahar. I've added a patch to allow the continuance of processing of further bids after ones with an unmatched Impression ID...

mansinahar:
@vrtcal-dev Looks like there are some conflicts with the recent stuff that got merged in. Can you please rebase your PR on top of master and resolve the conflicts?

adapters/vrtcal/vrtcal.go
```diff
@@ -70,10 +70,13 @@ func (a *VrtcalAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalR
 
 	for _, sb := range bidResp.SeatBid {
 		for i := range sb.Bid {
-			bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
-				Bid:     &sb.Bid[i],
-				BidType: "banner",
-			})
+			bidType, err := getReturnTypeForImp(sb.Bid[i].ImpID, internalRequest.Imp)
+			if err == nil {
```
mansinahar, line 74: @vrtcal-dev This check would just ignore the error. I'd recommend changing this to the following so that the errors are properly communicated to the user.
```
if (err != nil) {
	errs = append(errs, err)
	continue
}
bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
	Bid:     &sb.Bid[i],
	BidType: bidType,
})
```
You'll just. have to instantiate and `errs` slice either at the start of this function or before the for loop

vrtcal-dev: > This check would just ignore the error. I'd recommend changing this to the following so that the errors are properly communicated to the user.

Thanks mansinahar, I've udpated to properly report errored bids while continuing to proces further bids.

PR #1969, opened 2021-08-20
Update Viper

SyntaxNode:
The [/cookie_sync Endpoint Rewrite ](https://github.com/prebid/prebid-server/pull/1879) exposed a performance bug in the version of Viper we were using when using BindEnv. This PR updates Viper to the latest version to solve the performance bug, cleans up the config tests related to user syncing, and adds a user syncing environment variable override test.


config/config_test.go
```diff
@@ -757,7 +748,11 @@ func TestInvalidAdapterEndpointConfig(t *testing.T) {
 	v.SetConfigType("yaml")
 	v.ReadConfig(bytes.NewBuffer(invalidAdapterEndpointConfig))
 	_, err := New(v)
-	assert.Error(t, err, "invalid endpoint in config should return an error")
+
+	if assert.IsType(t, errortypes.AggregateError{}, err) {
+		aggErr := err.(errortypes.AggregateError)
+		assert.ElementsMatch(t, []error{errors.New("The endpoint: ib.adnxs.com/some/endpoint for appnexus is not a valid URL")}, aggErr.Errors)
+	}
```
SyntaxNode, line 755: Verified the exact error to be sure it's catching the error we expect.

config/config_test.go
```diff
@@ -936,6 +931,53 @@ func TestValidateAccountsConfigRestrictions(t *testing.T) {
...
+	assert.Equal(t, cfg.Adapters["appnexus"].Syncer.Redirect.UserMacro, "[UID]")
+	assert.Nil(t, cfg.Adapters["appnexus"].Syncer.IFrame)
+	assert.Equal(t, cfg.Adapters["appnexus"].Syncer.SupportCORS, &truePtr)
+
+	assert.Equal(t, cfg.Adapters["rubicon"].Syncer.IFrame.URL, "http://somedifferent.url/sync?redirect={{.RedirectURL}}")
+	assert.Nil(t, cfg.Adapters["rubicon"].Syncer.Redirect)
+	assert.Nil(t, cfg.Adapters["rubicon"].Syncer.SupportCORS)
+
+	assert.Nil(t, cfg.Adapters["brightroll"].Syncer)
+}
```
SyntaxNode, line 979: Ensures the proper handling of pointers with nil defaults via BindEnv.

go.mod
```diff
@@ -53,7 +45,7 @@ require (
 	github.com/yudai/gojsondiff v0.0.0-20170107030110-7b1b7adf999d
 	github.com/yudai/golcs v0.0.0-20170316035057-ecda9a501e82 // indirect
 	github.com/yudai/pp v2.0.1+incompatible // indirect
-	golang.org/x/net v0.0.0-20201202161906-c7110b5ffcbb
+	golang.org/x/net v0.0.0-20210405180319-a5a99cb37ef4
 	golang.org/x/text v0.3.6
 	gopkg.in/yaml.v2 v2.4.0
```
SyntaxNode, line 50: I only updated viper. The rest of the updates are transitive.

go.sum
```diff
@@ -1,135 +1,316 @@
-github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
hhhjort, line 1: why did this trigger so many changes in go.sum? go.mod seems more reasonable since there are roughly the same number of new lines as old lines.

SyntaxNode: Good question. I'm not sure. They are from the Viper go.sum, but I don't understand why the hashes need to exist. go.sum is kind of a mystery to me.

hhhjort (approved):

guscarreon (approved):
LGTM

Issue #1818, opened 2021-04-20
Prebid version in an HTTP header


bretg:
## Type of issue
Proposed Enhancement

## Description
The Prebid Server side of https://github.com/prebid/Prebid.js/issues/6627

There's interest from some bidders to be able to cleanly distinguish the source of traffic. Currently it's up to each bidder to read the version and pass it to their endpoint. There aren't any conventions about where bidders should place this version, so they've probably done it in various ways. 

The proposal is to update PBS to add/modify a new HTTP header `x-prebid` which carries the PBS version:

The Prebid.js team attempted to implement this header, but ran afoul of a browser CORS check -- it causes a pre-flight OPTIONS check, which is not acceptable. So instead, the proposal is to have the PrebidServer BidAdapter put the PBJS version in ext.prebid.channel: {name: "pbjs", version: "4.39"}. 

This value should be made available to bid adapters who want to know the version of PBJS that initiated the request.

It's assumed that PBS can still add an HTTP header, but we should find a way to test this.

### Prebid Server creates the header

Prebid Server creates the header, prepending the name and version from ext.prebid.channel.{name/version} if it exists:
```
x-prebid: pbjs/4.39,pbs-go/0.155
x-prebid: prebid-mobile/1.2.3,pbs-java/1.64
```

### SDK

The SDK already sends the version in app.ext.prebid::
```
app.ext.prebid: {
   source:  "prebid-mobile"
   version: "1.2.3"
}
```

PBS should read app.ext.prebid and if present, add the source/version to outgoing HTTP headers.

bszekely1:
Prebid SDK will be passing environment as well:
`x-prebid: pbsdk-ios/1.12`

bretg:
Please hold off on this update -- it caused major problems in Prebid.js.

TheMediaGrid:
Hi team,

I'd really love to make sure MediaGrid could support reading the PBS version. Any insights on the right param for us to be reading this from?

bretg:
Updated the proposal to drop the header supplied by PBJS. Since it was a browser CORS thing that prevents the header, we believe it's still acceptable in server-to-server communication. Will discuss in the next committee meeting.

bretg:
Discussed in committee. We need to consider the difference between the "integration" vs this ext.prebid.version proposal.

Also discussed whether we might need to make an opt-out on the new header. Looking for community input on this. Suggestion is to make it a global behavior flag initially until we learn whether we need a per-bidder flag.

[...]

PR #1981, opened 2021-08-30
Send x-prebid header in requests

rokostik:
Hey!

This PR implements https://github.com/prebid/prebid-server/issues/1818. I looked at https://github.com/prebid/prebid-server-java/pull/1301 for reference but changed a few things when constructing the header. If the binary is not built with the revision flag the version is `pbs-go/unknown`. Similarly for the channel and appExt, if the version is not given only the name or source is taken and the version is `unknown`. My reasoning is that it's still better to know the client type if the version is missing/unavailable. Any thoughts on this?

### DIFF
--- a/endpoints/version_test.go
+++ b/endpoints/version_test.go
@@ -14,9 +14,9 @@ func TestVersion(t *testing.T) {
 		input    string
 		expected string
 	}{
-		{"", `{"revision": "not-set"}`},
-		{"abc", `{"revision": "abc"}`},
-		{"d6cd1e2bd19e03a81132a23b2025920577f84e37", `{"revision": "d6cd1e2bd19e03a81132a23b2025920577f84e37"}`},
+		{"", `{"version": "not-set"}`},
+		{"abc", `{"version": "abc"}`},
+		{"d6cd1e2bd19e03a81132a23b2025920577f84e37", `{"version": "d6cd1e2bd19e03a81132a23b2025920577f84e37"}`},
 	}
 
 	for _, tc := range testCases {
@@ -29,6 +29,45 @@ func TestVersion(t *testing.T) {
 
 		// Verify:
 		var result, expected versionModel
+		err := json.NewDecoder(w.Body).Decode(&result)
+		if err != nil {
+			t.Errorf("Bad response body. Expected: %s, got an error %s", tc.expected, err)
+		}
+
+		err = json.Unmarshal([]byte(tc.expected), &expected)
+		if err != nil {
+			t.Errorf("Error while trying to unmarshal expected result JSON")
+		}
+
+		if !reflect.DeepEqual(expected, result) {
+			responseBodyBytes, _ := ioutil.ReadAll(w.Body)
+			responseBodyString := string(responseBodyBytes)
+			t.Errorf("Bad response body. Expected: %s, got %s", tc.expected, responseBodyString)
+		}
+	}
+}
+
+func TestRevision(t *testing.T) {
+	// Setup:
+	var testCases = []struct {
+		input    string
+		expected string
+	}{
+		{"", `{"revision": "not-set"}`},
+		{"abc", `{"revision": "abc"}`},
+		{"d6cd1e2bd19e03a81132a23b2025920577f84e37", `{"revision": "d6cd1e2bd19e03a81132a23b2025920577f84e37"}`},
+	}
+
+	for _, tc := range testCases {
+
+		handler := NewRevisionEndpoint(tc.input)
+		w := httptest.NewRecorder()
+
+		// Execute:
+		handler(w, nil)
+
+		// Verify:
+		var result, expected revisionModel
 		err := json.NewDecoder(w.Body).Decode(&result)
 		if err != nil {
 			t.Errorf("Bad response body. Expected: %s, got an error %s", tc.expected, err)
--- a/endpoints/version_test.go
+++ b/endpoints/version_test.go
@@ -11,17 +11,18 @@ import (
 func TestVersion(t *testing.T) {
 	// Setup:
 	var testCases = []struct {
-		input    string
+		version  string
+		revision string
 		expected string
 	}{
-		{"", `{"version": "not-set"}`},
-		{"abc", `{"version": "abc"}`},
-		{"d6cd1e2bd19e03a81132a23b2025920577f84e37", `{"version": "d6cd1e2bd19e03a81132a23b2025920577f84e37"}`},
+		{"", "", `{"version":"not-set","revision":"not-set"}`},
+		{"abc", "def", `{"version":"abc","revision"	:"def"}`},
+		{"1.2.3", "d6cd1e2bd19e03a81132a23b2025920577f84e37", `{"version":"1.2.3","revision":"d6cd1e2bd19e03a81132a23b2025920577f84e37"}`},
 	}
 
 	for _, tc := range testCases {
 
-		handler := NewVersionEndpoint(tc.input)
+		handler := NewVersionEndpoint(tc.version, tc.revision)
 		w := httptest.NewRecorder()
 
 		// Execute:
@@ -29,7 +30,8 @@ func TestVersion(t *testing.T) {
 
 		// Verify:
 		var result, expected versionModel
-		err := json.NewDecoder(w.Body).Decode(&result)
+		responseBodyBytes, _ := ioutil.ReadAll(w.Body)
+		err := json.Unmarshal(responseBodyBytes, &result)
 		if err != nil {
 			t.Errorf("Bad response body. Expected: %s, got an error %s", tc.expected, err)
 		}
@@ -40,46 +42,6 @@ func TestVersion(t *testing.T) {
 		}
 
 		if !reflect.DeepEqual(expected, result) {
-			responseBodyBytes, _ := ioutil.ReadAll(w.Body)
-			responseBodyString := string(responseBodyBytes)
-			t.Errorf("Bad response body. Expected: %s, got %s", tc.expected, responseBodyString)
-		}
-	}
-}
-
-func TestRevision(t *testing.T) {
-	// Setup:
-	var testCases = []struct {
-		input    string
-		expected string
-	}{
-		{"", `{"revision": "not-set"}`},
-		{"abc", `{"revision": "abc"}`},
-		{"d6cd1e2bd19e03a81132a23b2025920577f84e37", `{"revision": "d6cd1e2bd19e03a81132a23b2025920577f84e37"}`},
-	}
-
-	for _, tc := range testCases {
-
-		handler := NewRevisionEndpoint(tc.input)
-		w := httptest.NewRecorder()
-
-		// Execute:
-		handler(w, nil)
-
-		// Verify:
-		var result, expected revisionModel
-		err := json.NewDecoder(w.Body).Decode(&result)
-		if err != nil {
-			t.Errorf("Bad response body. Expected: %s, got an error %s", tc.expected, err)
-		}
-
-		err = json.Unmarshal([]byte(tc.expected), &expected)
-		if err != nil {
-			t.Errorf("Error while trying to unmarshal expected result JSON")
-		}
-
-		if !reflect.DeepEqual(expected, result) {
-			responseBodyBytes, _ := ioutil.ReadAll(w.Body)
 			responseBodyString := string(responseBodyBytes)
 			t.Errorf("Bad response body. Expected: %s, got %s", tc.expected, responseBodyString)
 		}
--- a/endpoints/version_test.go
+++ b/endpoints/version_test.go
@@ -30,8 +30,11 @@ func TestVersion(t *testing.T) {
 
 		// Verify:
 		var result, expected versionModel
-		responseBodyBytes, _ := ioutil.ReadAll(w.Body)
-		err := json.Unmarshal(responseBodyBytes, &result)
+		responseBodyBytes, err := ioutil.ReadAll(w.Body)
+		if err != nil {
+			t.Errorf("Error reading response body bytes: %s", err)
+		}
+		err = json.Unmarshal(responseBodyBytes, &result)
 		if err != nil {
 			t.Errorf("Bad response body. Expected: %s, got an error %s", tc.expected, err)
 		}
--- a/exchange/bidder.go
+++ b/exchange/bidder.go
@@ -49,7 +49,7 @@ type adaptedBidder interface {
 	//
 	// Any errors will be user-facing in the API.
 	// Error messages should help publishers understand what might account for "bad" bids.
-	requestBid(ctx context.Context, request *openrtb2.BidRequest, name openrtb_ext.BidderName, bidAdjustment float64, conversions currency.Conversions, reqInfo *adapters.ExtraRequestInfo, revision string, accountDebugAllowed, headerDebugAllowed bool) (*pbsOrtbSeatBid, []error)
+	requestBid(ctx context.Context, request *openrtb2.BidRequest, name openrtb_ext.BidderName, bidAdjustment float64, conversions currency.Conversions, reqInfo *adapters.ExtraRequestInfo, version string, accountDebugAllowed, headerDebugAllowed bool) (*pbsOrtbSeatBid, []error)
 }
 
 // pbsOrtbBid is a Bid returned by an adaptedBidder.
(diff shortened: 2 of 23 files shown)


exchange/utils.go
```diff
@@ -700,3 +701,42 @@ func getExtBidAdjustmentFactors(requestExt *openrtb_ext.ExtRequest) map[string]f
...
 	return bidAdjustmentFactors
 }
+
+func buildXPrebidHeader(bidRequest *openrtb2.BidRequest, revision string) string {
+	req := &openrtb_ext.RequestWrapper{BidRequest: bidRequest}
+
+	pbsRecord := createNameVersionRecord("pbs-go", revision)
+
+	var channelVersionRecord string
+	if reqExt, _ := req.GetRequestExt(); reqExt != nil {
```
SyntaxNode, line 711: Please do not ignore errors. Capture the error to a variable and include a check for it in the if clause. Same for the sdk version.

exchange/utils.go
```diff
@@ -700,3 +701,42 @@ func getExtBidAdjustmentFactors(requestExt *openrtb_ext.ExtRequest) map[string]f
...
+			sdkVersionRecord = createNameVersionRecord(prebidExt.Source, prebidExt.Version)
+		}
+	}
+
+	var nonEmptyRecords []string
+	for _, record := range []string{pbsRecord, channelVersionRecord, sdkVersionRecord} {
+		if record != "" {
+			nonEmptyRecords = append(nonEmptyRecords, record)
+		}
+	}
```
SyntaxNode, line 730: I like how this code reads, but the creation of a new slices causes two heap memory allocation. We can avoid this using direct checks for each record with a string builder.

exchange/bidder.go
```diff
@@ -139,16 +139,19 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, request *openrtb2.B
...
-			}
+	xPrebidHeader := buildXPrebidHeader(request, revision)
+
+	for i := 0; i < len(reqData); i++ {
+		if reqData[i].Headers != nil {
+			reqData[i].Headers = reqData[i].Headers.Clone()
+		} else {
+			reqData[i].Headers = http.Header{}
+		}
+		if revision != "" {
```
SyntaxNode, line 150: Is it possible for revision to be empty by this point?

router/router.go
```diff
@@ -171,7 +171,7 @@ type Router struct {
 	Shutdown        func()
 }
 
-func New(cfg *config.Configuration, rateConvertor *currency.RateConverter) (r *Router, err error) {
+func New(cfg *config.Configuration, rateConvertor *currency.RateConverter, revision string) (r *Router, err error) {
```
SyntaxNode, line 174: Passing down the revision begins a long chain of refactors. I'm personally fine referring back to the static variable, such as `main.Version`.

SyntaxNode (commented):
Thank you for tackling this issue.

> If the binary is not built with the revision flag the version is `pbs-go/unknown`.

I agree this is reasonable behavior for when the binary doesn't have access to the version. Please see my other comment regarding use of the version rather than the revision.

> Similarly for the channel and appExt, if the version is not given only the name or source is taken and the version is `unknown`.

I'm less certain about this case. Please mention this in issue https://github.com/prebid/prebid-server/issues/1818 for a larger discussion with the community.

main.go
```diff
@@ -59,7 +59,7 @@ func serve(revision string, cfg *config.Configuration) error {
 	currencyConverterTickerTask := task.NewTickerTask(fetchingInterval, currencyConverter)
 	currencyConverterTickerTask.Start()
 
-	r, err := router.New(cfg, currencyConverter)
+	r, err := router.New(cfg, currencyConverter, revision)
```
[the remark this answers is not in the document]
rokostik: Cool, added the new variable but would need help modifying the Dockerfile because I can't find where the Rev string is actually populated... if it's actually not set I can add both Version and Rev to the build script. 
I would also need help describing how it's populated in the comment above the variable since i'm not sure if the Rev part is still true.

exchange/utils.go
rokostik: Oh yeah, of course... kinda forgot the returned object can be non-nil if there's an error. Thanks for noticing.

exchange/utils.go
rokostik: Agreed, fixed.

exchange/bidder.go
rokostik: It can't be, you're correct. This is from the first implementation I did where it could be empty... fixed.

router/router.go
rokostik: Yeah, wasn't sure here whether to pass it around or access the populated variable... but since main is not importable we would need to move it somewhere else. Should I create a package named version that would contain both these variables?

bretg:
yes, the 'unknown' edge cases mentioned are fine.

pbs-go/unknown
pbjs/unknown
prebid-mobile/unknown

router/router.go
guscarreon: Hi @rokostik we discussed this a bit offline and we support creating a new package put the `version` value in. Maybe inside the `util/version/version.go` directory? We believe it'd greater simplify the code.
```
138             errs = append(errs, &errortypes.FailedToRequestBids{Message: "The adapter failed to generate any bid requests, but also failed to generate an error explaining why"})
139         }
140         return nil, errs
141     }
142
143 -   xPrebidHeader := buildXPrebidHeader(request, version)
    +   xPrebidHeader := buildXPrebidHeader(request, versionValuePackage.GetVersion())
144
145     for i := 0; i < len(reqData); i++ {
146         if reqData[i].Headers != nil {
147             reqData[i].Headers = reqData[i].Headers.Clone()
exchange/bidder.go
```
What do you think? (Feel free to name the package to something better than `versionValuePackage`)

guscarreon (commented):
Hi @rokostik, we agree on creating a package, see comment:
https://github.com/prebid/prebid-server/pull/1981#discussion_r700522619

main.go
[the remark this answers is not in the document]
SyntaxNode: It appears we are not actually setting the the Rev during the build process for the Docker image. We also used to build a binary and did specify it there, but we removed that artifact as no one was using it. 

> if it's actually not set I can add both Version and Rev to the build script.

Yes, please do. Bonus points if you test the Docker build process locally to ensure the variables are properly set. You would make the change in Dockerfile on line 23 `RUN go build -mod=vendor .`

exchange/bidder.go
SyntaxNode: Thank you. This makes me a lot more comfortable with always creating or cloning the headers, since we know we'll always set something.

router/router.go
SyntaxNode: > but since main is not importable

Doh! I didn't think of that.

> Maybe inside the `util/version`

Maybe a version package at the top level? rather than within util? Doesn't really matter. So long as it's easily discoverable.

exchange/utils.go
rokostik: That's a great suggestion, will do, thanks!

router/router.go
rokostik: I think a top level version directory would work best in this case. I'll also move the Rev variable to there.

main.go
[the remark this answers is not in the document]
rokostik: Added both flags to the Dockerfile. Tested locally and in Docker that both are correctly set.

router/router.go
rokostik: Extracted both Rev and Ver (newly named Version) to their own package. I also renamed the /version endpoint to /revision and added a new /version endpoint. So that these two endpoints actually return the things they're named after.

exchange/utils.go
```diff
@@ -700,3 +701,39 @@ func getExtBidAdjustmentFactors(requestExt *openrtb_ext.ExtRequest) map[string]f
 	}
 	return bidAdjustmentFactors
 }
+
+func buildXPrebidHeader(bidRequest *openrtb2.BidRequest, version string) string {
+	req := &openrtb_ext.RequestWrapper{BidRequest: bidRequest}
+
+	sb := new(strings.Builder)
```
SyntaxNode, line 708: Syntax Nitpick: It's not common in Go to use the new keyword. Please consider using `sb := &strings.Builder{}`.

main.go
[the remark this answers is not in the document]
SyntaxNode: > Tested locally and in Docker that both are correctly set.

Yay! That's fantastic. Thank you.

router/router.go
SyntaxNode: > Extracted both Rev and Ver (newly named Version) to their own package.

Perfect.

>  I also renamed the /version endpoint to /revision and added a new /version endpoint. So that these two endpoints actually return the things they're named after.

This is technically a breaking change, but it makes sense and should be fine since it's currently broken. I don't believe it's necessary to add an new revision endpoint, we could keep the version endpoint and change the response to include both:

```
type versionModel struct {
	Version string `json:"version"`
	Revision string `json:"revision"`
}
```

I also don't know why my predecessors on this project choose to use a short hash instead of the full hash. What do you think of changing the build script to use the full hash for Rev?

rokostik: Good idea, I now merged both under the /version endpoint. 
As for why it's the short sha, I couldn't find the reason in the original issue and since you say that it's not used anywhere I think it's ok if I change it to the full hash so I went ahead and did that.

SyntaxNode (dismissed):

exchange/utils.go
```diff
@@ -700,3 +701,39 @@ func getExtBidAdjustmentFactors(requestExt *openrtb_ext.ExtRequest) map[string]f
...
+			if channel := prebidExt.Channel; channel != nil {
+				writeNameVersionRecord(sb, channel.Name, channel.Version)
+			}
+		}
+	}
+	if appExt, err := req.GetAppExt(); err == nil && appExt != nil {
+		if prebidExt := appExt.GetPrebid(); prebidExt != nil {
+			writeNameVersionRecord(sb, prebidExt.Source, prebidExt.Version)
+		}
+	}
```
guscarreon, line 722: What if we have valid values in both `reqExt` and `appExt`? In other words, what if all conditions of `if` statements found in lines 711, 712, and 713 as well as 718 and 719 are all true? Do we want `buildXPrebidHeader` to have the ability to return something like:
```
"pbs-go/test-version,pbjs/test-pbjs-version,prebid-mobile/test-prebid-mobile-version"
                               ^                              ^---- mobile version
                               |----------channel version
```
If we don't want this scenario, should we maybe use a `return` statement if valid `channel.Name` and `channel.Version` are parsed?
```
705   func buildXPrebidHeader(bidRequest *openrtb2.BidRequest, version string) string {
706       req := &openrtb_ext.RequestWrapper{BidRequest: bidRequest}
707  
708       sb := &strings.Builder{}
709       writeNameVersionRecord(sb, "pbs-go", version)
710  
711       if reqExt, err := req.GetRequestExt(); err == nil && reqExt != nil {
712           if prebidExt := reqExt.GetPrebid(); prebidExt != nil {
713               if channel := prebidExt.Channel; channel != nil {
714                   writeNameVersionRecord(sb, channel.Name, channel.Version)
    +                 return sb.String()
715               }
716           }
717       }
718       if appExt, err := req.GetAppExt(); err == nil && appExt != nil {
719           if prebidExt := appExt.GetPrebid(); prebidExt != nil {
720               writeNameVersionRecord(sb, prebidExt.Source, prebidExt.Version)
721           }
722       }
723       return sb.String()
724   }
exchange/utils.go
```

exchange/utils_test.go
```diff
@@ -2228,3 +2228,98 @@ func TestRemoveUnpermissionedEidsEmptyValidations(t *testing.T) {
...
+		{
+			description: "app ext without a source",
+			version:     "test-version",
+			requestAppExt: &openrtb_ext.ExtApp{
+				Prebid: openrtb_ext.ExtAppPrebid{
+					Version: "test-version",
+				},
+			},
+			result: "pbs-go/test-version",
+		},
```
guscarreon, line 2307: Can we add a test where both `requestExt` and `requestAppExt` hold valid values?
```
2232 func TestBuildXPrebidHeader(t *testing.T) {
2233     testCases := []struct {
2234         description   string
2235         version       string
2236         requestExt    *openrtb_ext.ExtRequest
2237         requestAppExt *openrtb_ext.ExtApp
2238         result        string
2239     }{
2240         {
2241             description: "No versions",
2242             version:     "",
2243             result:      "pbs-go/unknown",
2244         },
2245 -- 53 lines: {------------------------------------------------------------
2298         {
2299             description: "app ext without a source",
2300             version:     "test-version",
2301             requestAppExt: &openrtb_ext.ExtApp{
2302                 Prebid: openrtb_ext.ExtAppPrebid{
2303                     Version: "test-version",
2304                 },
2305             },
2306             result: "pbs-go/test-version",
2307         },
     +       {
     +           description: "Version found in both req.Ext and req.App.Ext",
     +           version:     "test-version",
     +           requestExt: &openrtb_ext.ExtRequest{  // Valid
     +               Prebid: openrtb_ext.ExtRequestPrebid{
     +                   Channel: &openrtb_ext.ExtRequestPrebidChannel{
     +                       Name:    "pbjs",
     +                       Version: "test-pbjs-version",
     +                   },
     +               },
     +           },
     +           requestAppExt: &openrtb_ext.ExtApp{  // Valid
     +               Prebid: openrtb_ext.ExtAppPrebid{
     +                   Source:  "prebid-mobile",
     +                   Version: "test-prebid-mobile-version",
     +               },
     +           },
     +           result: "PUT BOTH VERSIONS IN THE RESULT OR ONLY ONE OF THEM, WHICHEVER TAKES PRIORITY"
     +       },
2308     }
2309
2310     for _, test := range testCases {
2311         req := &openrtb2.BidRequest{}
2312 -- 11 lines: if test.requestExt != nil {--------------------------------
2323         assert.Equal(t, test.result, result, test.description+":result")
2324     }
2325 }
exchange/utils_test.go
```

guscarreon (commented):
Code-wise looks pretty good: lean, readable and well tested. I was mainly just left with a question in regards to the (probably unlikely) scenario where both `request.Ext` and `request.App.Ext` hold valid values

exchange/utils.go
guscarreon: Please disregard my comment above I've been told it's fine to have both `reqExt` and `appExt` values.

SyntaxNode: I think it's fine. I mean, it doesn't really make sense for both to be provided. This may never actually happen in the wild, but if it does - I don't personally see an issue with a comma separated list of 3 items. The bidder can parse out what they like.

guscarreon (changes requested):
@rokostik we discussed the output of the `/version` endpoint this pull request implements and we decided that only the version number should be printed as opposed to the extra infomation after the `-` character that could be displayed if not found on the master branch: `"0.174.0-17-g9aaa543b"`

In other words, we'd like the `/version` endpoint to always respond:
```
╰─➤  curl http://localhost:9000/version
{"version":"0.174.0","revision":"9aaa543b0c66cc6696ef14f9be906f492eb72e51"}%
```
And according to the `git describe` man pages the command to do this trick is:
```
git describe --abbrev=0 --tags
```
Could you please modify?

exchange/utils_test.go
rokostik: Done.

rokostik:
> @rokostik we discussed the output of the `/version` endpoint this pull request implements and we decided that only the version number should be printed as opposed to the extra infomation after the `-` character that could be displayed if not found on the master branch: `"0.174.0-17-g9aaa543b"`

Sure thing but in my opinion it makes sense for the version that is ahead of the tag to be reported differently since it's not actually running the tagged version. Maybe we should do something similar as to what java projects do: `$(version+1)-SNAPSHOT`. But it would be hard to determine what version+1 means here... what do you think?

endpoints/version_test.go
```diff
@@ -11,25 +11,27 @@ import (
...
+		handler := NewVersionEndpoint(tc.version, tc.revision)
 		w := httptest.NewRecorder()
 
 		// Execute:
 		handler(w, nil)
 
 		// Verify:
 		var result, expected versionModel
-		err := json.NewDecoder(w.Body).Decode(&result)
+		responseBodyBytes, _ := ioutil.ReadAll(w.Body)
```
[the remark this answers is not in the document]
rokostik: Done.

SyntaxNode:
> what do you think?

That seems complicated to me, but then again I'm not a wiz at the command line.

I suggest we optimize for the use case where the HEAD is tagged with the version we want to use in the build. This is how the build process for the public Docker image works today.

rokostik:
Well if the build is run on HEAD and HEAD and the latest tag both point to the same commit only the tag will be printed as you requested. So for proper release builds it already work this way. This is why I made it so that "untagged" builds don't forward just the tag but also some info that it's ahead of the tag by some commits.

SyntaxNode:
> Well if the build is run on HEAD and HEAD and the latest tag both point to the same commit only the tag will be printed as you requested. So for proper release builds it already work this way. This is why I made it so that "untagged" builds don't forward just the tag but also some info that it's ahead of the tag by some commits.

@guscarreon Does this work for you?

guscarreon:
> > Well if the build is run on HEAD and HEAD and the latest tag both point to the same commit only the tag will be printed as you requested. So for proper release builds it already work this way. This is why I made it so that "untagged" builds don't forward just the tag but also some info that it's ahead of the tag by some commits.
> 
> @guscarreon Does this work for you?

It does. It makes sense

SyntaxNode (approved):

guscarreon (approved):
LGTM

Issue #1922, opened 2021-07-14
Transmit ext info from ad response to publisher


Niksok:
Hello!
Is there a way to convey some ext info from the ad response to the publisher?
Maybe it is possible to set additional targeting keys?

Something like in this client side PR https://github.com/prebid/Prebid.js/pull/7179

SyntaxNode:
Prebid Server does not have the ability for an adapter to specify a custom targeting key. PBS-Java allows publishers to specify custom targeting keys to be resolved from the OpenRTB data structures but this feature is not yet present in PBS-Go, please see https://github.com/prebid/prebid-server/issues/1167 for further details.

Adapters may transmit any ext info they would like in their bid.ext response which will be sent to the publisher as `seatbid.bid[].ext` in the Prebid Server response. Please note, any values in `ext.prebid` from the bidder will be overwritten with Prebid Server specific information.

Niksok:
Ok, thank you for response. May i make pull request with modified prebidServerBidAdapter module? To support transmit bid[].ext.bidder[<bidderName>] data in this module?

patmmccann:
https://github.com/prebid/Prebid.js/pulls/6939 allows anything in the meta field to flow through. To quote Bret in Slack:  "the server-side adapter needs to set .Bids[].Bid.Ext.Prebid.Meta.MediaType . If it does, and PBJS is recent enough (4-5 months, could pin the version if needed), then the data will pass through to the PBJS client-side request object"

bretg:
If the field you're wanting to pass through doesn't belong in Bids[].Bid.Ext.Prebid.Meta, you could submit a PBJS PR to read Bids[].Bid.Ext and put it somewhere (BidResponse.pbsBidExt?)

Niksok:
But if my endpoint response has field seatbid[].bid[].ext.prebid.meta, this value is overrided in this line But if my endpoint response has field seatbid[].bid[].ext.prebid.meta, this value is overridden in this line 1. Or i am false and it will work fine? I see that on test files it work, but when i start prebid-server local  and test it - i see that bid[].ext.prebid is realy overridden and my prebid.meta is lose.

bretg:
@SyntaxNode or @hhhjort - please take a look? The idea here is that we want the "Setting Metadata" example to work as documented in https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html

hhhjort:
[...]

PR #1983, opened 2021-09-02
Persist bid.ext.prebid.meta From Bidding Server Response

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/1922

Persists the `bid.ext.prebid.meta` object from the bidding server response with no changes required by the bid adapter.
- If a bid adapter explicitly specifies metadata, that takes precedence.
- If metadata from the bidding server is invalid / fails validation, PBS returns an error.
- If unknown metadata is specified from the bidding server, it is ignored.


openrtb_ext/bid.go
```diff
@@ -38,18 +39,20 @@ type ExtBidPrebidCacheBids struct {
...
-	PrimaryCategoryID    string   `json:"primaryCatId,omitempty"`
-	SecondaryCategoryIDs []string `json:"secondaryCatIds,omitempty"`
+	AdvertiserDomains    []string        `json:"advertiserDomains,omitempty"` // or advertiserDomain?
+	AdvertiserID         int             `json:"advertiserId,omitempty"`
+	AdvertiserName       string          `json:"advertiserName,omitempty"`
+	AgencyID             int             `json:"agencyId,omitempty"`
+	AgencyName           string          `json:"agencyName,omitempty"`
+	BrandID              int             `json:"brandId,omitempty"`
+	BrandName            string          `json:"brandName,omitempty"`
+	DemandSource         string          `json:"demandSource,omitempty"`
```
mansinahar, line 49: I am okay with this change but just curious to know where are these new fields `DemandSource` and `DChain` coming from? I don't see them mentioned in the issue linked

exchange/exchange_test.go
```diff
@@ -3571,6 +3619,14 @@ func TestMakeBidExtJSON(t *testing.T) {
 			expectedBidExt:     ``,
 			expectedErrMessage: "invalid character",
 		},
+		{
```
mansinahar, line 3622: The added test cases look great! Can we please add just one more test case where `prebid.meta` doesn't exist in both `ext` and the bid?

openrtb_ext/bid.go
```diff
@@ -38,18 +39,20 @@ type ExtBidPrebidCacheBids struct {
...
-	AgencyID             int      `json:"agencyId,omitempty"`
-	AgencyName           string   `json:"agencyName,omitempty"`
-	BrandID              int      `json:"brandId,omitempty"`
-	BrandName            string   `json:"brandName,omitempty"`
-	MediaType            string   `json:"mediaType,omitempty"`
-	NetworkID            int      `json:"networkId,omitempty"`
-	NetworkName          string   `json:"networkName,omitempty"`
-	PrimaryCategoryID    string   `json:"primaryCatId,omitempty"`
-	SecondaryCategoryIDs []string `json:"secondaryCatIds,omitempty"`
+	AdvertiserDomains    []string        `json:"advertiserDomains,omitempty"` // or advertiserDomain?
```
mansinahar, line 42: **Nitpick:** I know this isn't your change but could you please remove this leftover comment: `// or advertiserDomain?` ?

util/maputil/maputil_test.go
```diff
@@ -164,3 +164,84 @@ func TestReadEmbeddedString(t *testing.T) {
...
+		{
+			description: "Level 2 - Does Not Exist At All",
+			value:       map[string]interface{}{"foo": map[string]interface{}{"bar": "exists"}},
+			keys:        []string{"doesnotexist", "doesnotexist"},
+			expected:    false,
+		},
+		{
+			description: "Keys Nil",
+			value:       map[string]interface{}{"foo": "exists"},
+			keys:        []string{},
```
mansinahar, line 214: Looks like from the test description, you meant to pass `nil` for `keys` here

mansinahar (commented):
LGTM! Just some minor comments

openrtb_ext/bid.go
SyntaxNode: The authoritative source for those metadata fields is the Prebid.js docs: https://docs.prebid.org/dev-docs/bidder-adaptor.html

exchange/exchange_test.go
SyntaxNode: Sure. I added a test case where there are no `ext`s. There are also many other tests cases which do not define any `meta` fields.

openrtb_ext/bid.go
SyntaxNode: Removed. Still my fault though, just from a previous PR.

openrtb_ext/bid.go
mansinahar: I see. Thanks!

mansinahar (approved):

exchange/exchange.go
```diff
@@ -979,10 +977,22 @@ func makeBidExtJSON(ext json.RawMessage, prebid *openrtb_ext.ExtBidPrebid, impEx
 		extMap = make(map[string]interface{})
 	}
 
+	// ext.prebid
+	if prebid.Meta == nil && maputil.HasElement(extMap, "prebid", "meta") {
```
VeronikaSolovei9, line 981: My idea was to reuse existing `jsonutil`, but I'm ok with this too :)

VeronikaSolovei9 (approved):

exchange/exchange.go
VeronikaSolovei9: In `fpd` branch I modified `jsonutil` to support exactly what you wrote but with byte array:
`func FindElement(extension []byte, elementNames ...string)` - to check element. with any nexting level exists
and 
`func SetElement(originDataInput []byte, setValue []byte, keys ...string)` - to set element with any nested level.
Using this functions will replace logic related to `extMap` variable and Unmarshal and Marshal will not be necessary anymore

SyntaxNode: Thank you for sharing. The `jsontuil` methods operate on a json byte array and new `maputil` is operating on an already unmashalled map. There are other functions in this method which require full blown json marshal. So, I don't think the `jsontuil` would be a good use in this case.

PR #1988, opened 2021-09-07
Fix external cache host validation

rokostik:
Hey!

This PR fixes the external cache host validation to support having a host with a port (useful for local development)


SyntaxNode (approved):

guscarreon (approved):
Pretty good catch @rokostik . Thanks for fixing this

Issue #1788, opened 2021-04-02
Make /info/bidders endpoint consistent between Go and Java


bretg:
As discussed in PBS committee, there's currently a difference between PBS-Go and PBS-Java in the /info/bidders endpoint. 

- PBS-Go lists all bid adapters whether or not they're enabled
- PBS-Java lists only enabled adapters

As both use cases are useful, the proposal is to add a query string parameter (?enabledonly=true/false). This value has no default and will return HTTP 400 if not provided with body "missing enabledonly flag"

The results of the /info/bidders/BIDDER endpoint should include all the fields currently in the PBS-Go output:

```
{
    "status": "ACTIVE",
    "usesHttps": true,
    "maintainer": {
        "email": "scope.sspp@adform.com"
    },
    "capabilities": {
        "app": {
            "mediaTypes": [
                "banner",
                "video"
            ]
        },
        "site": {
            "mediaTypes": [
                "banner",
                "video"
            ]
        }
    }
}
```

SyntaxNode:
Sounds good to me. What should the behavior be if enabledonly is not to either true or false?  Perhaps we should use the default, or another option is to return an error.

SyntaxNode:
Discussed with a few other folks. We currently prefer to validate `enabledonly` and return an error if it's not true or false, case insensitive.

bretg:
this was released with PBS-Java 1.71

SyntaxNode:
I implemented this for PBS-Go, but noticed the specifications had since changed to return a 400 if the new `enabledonly` is not provided. This is a breaking change which will affect folks at Xandr.  To avoid problems for existing usages, I will have `enabledonly` default to `false` for PBS-Go.

bretg:
This is ok @SyntaxNode , but the spec changing was due to your request from April 7th. :-)

SyntaxNode:
Ah! I see. In that case I believe this is a misunderstanding. I didn't intend to make it required, but to validate the parameter only when its provided, such that:

/info/bidders -> ok (backwards compability)
/info/bidders?enabledonly=false -> all bidders (current default in PBS-Go)
/info/bidders?enabledonly=true -> enabled only
/info/bidders?enabledonly= -> 400 error
/info/bidders?enabledonly=foo -> 400 error

SyntaxNode:
Implemented in PBS-Go 0.177.0.

PR #1990, opened 2021-09-08
/info/bidders endpoint consistency

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/1788

### DIFF
--- a/endpoints/info/bidders.go
+++ b/endpoints/info/bidders.go
@@ -4,28 +4,69 @@ import (
 	"encoding/json"
 	"net/http"
 	"sort"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
 	"github.com/prebid/prebid-server/config"
 )
 
+var invalidEnabledOnly = []byte(`Invalid value for 'enabledonly' query param, must be of boolean type`)
+
 // NewBiddersEndpoint builds a handler for the /info/bidders endpoint.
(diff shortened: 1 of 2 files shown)


endpoints/info/bidders.go
```diff
@@ -4,28 +4,69 @@ import (
...
+		var writeErr error
+
+		enabledOnly, exists := readEnabledOnly(r)
+
+		if !exists {
+			// Backwards compatibility. PBS would previously respond with all enabled bidders.
+			w.Header().Set("Content-Type", "application/json")
+			_, writeErr = w.Write(responseAll)
+		} else {
+			switch enabledOnly {
```
mansinahar, line 38: Thoughts about extracting this logic in a diff function? Perhaps we can call it something like `checkEnabledOnly` which takes in a request and returns a bool value `enabledOnly` and an error. It will call the `readEnabledOnly` function and return `true` if the `enabledOnly` query parameter exists and is set to `true`, `false` if it either doesn't exist or is explicitly set to `false`, error otherwise. With that the logic in this handler function will become much more sleek and easier to follow. It would look something along the lines of:
```
enabledOnly, err := checkEnabledOnly(r)
if(err != nil) {
	_, writeErr = w.Write(invalidEnabledOnly)
}

w.Header().Set("Content-Type", "application/json")
if(enabledOnly) {
	_, writeErr = w.Write(responseEnabledOnly)
} else {
	_, writeErr = w.Write(responseAll)
}
```

mansinahar (approved):

bsardo (approved):
LGTM

PR #1991, opened 2021-09-08
PubMatic + OpenWrap: Support request level Bidder Params #1042

sachin-pubmatic:
Changes for https://github.com/prebid/prebid-server/issues/1042


exchange/utils_test.go
```diff
@@ -868,6 +868,90 @@ func TestCleanOpenRTBRequestsSChain(t *testing.T) {
...
 
+func TestCleanOpenRTBRequestsBidderParams(t *testing.T) {
+	testCases := []struct {
+		description string
+		inExt       json.RawMessage
+		expectedExt map[string]json.RawMessage
+		hasError    bool
+	}{
+		{
+			description: "Bidder params for single partner",
```
hhhjort, line 879: These descriptions should be unique to each test, so you can identify the failing test if a test fails. Please modify the descriptions so the differences between the two tests are clear.

exchange/utils.go
```diff
@@ -226,21 +233,33 @@ func getAuctionBidderRequests(req AuctionRequest,
...
 		coreBidder := resolveBidder(bidder, aliases)
 
 		reqCopy := *req.BidRequest
 		reqCopy.Imp = imps
-		reqCopy.Ext = reqExt
 
 		prepareSource(&reqCopy, bidder, sChainsByBidder)
 
+		if len(bidderExt) != 0 {
+			bidderName := openrtb_ext.BidderName(bidder)
```
hhhjort, line 246: Why do you cast the string `bidder` into a `bidderName`, just to cast it back into a string?

hhhjort:
Seems to e some confusion. This seems more in line with the initially proposed spec in #1042 , however in the discussion that followed, it was proposed that PBS core take the parameters in ext.prebid.bidderparameters and merge them into imp.ext.bidder, so each imp would have a unified set of parameters that could easily be validated as conforming to the bidder parameter schema. It would also allow for some flexibility in that any bidder parameter could be moved into the request level if it had the same value in all the imps, or even have a default request level value, with one or more imps overriding it at the imp level.

What you have here seems to be an independent set of request level parameters that gets no schema validation, so always optional and unvalidated. Also all adapters must be aware of the existence of the request level parameters to use them. Should probably return to the issue and hash out what we want to support here and how, as it seems like this PR hasn't taken note of the later discussions in the issue.

hhhjort:
Wanted to be clear, what you are doing here is only a small part of #1042, a concern that was independent of the code level issues that you have addressed. Is this the extent of your interest in #1042, or is this just the first of a series of PRs?

hhhjort:
To the summarize the feature set as I understand it:

- map `request.ext.prebid.bidderparams.<bidder>` into `bidderparams` for each bidder, hiding the the adapter configs from each other.
- merge the request level parameters into all the imp level parameters, giving priority to the imp level version of the parameter
- Validate the merged set of imp level bidder parameters against the bidder's schema to ensure that a valid set of parameters have been applied.

You have done the first item. The other two would help drive a more powerful and flexible usage of the feature. With the merger, publishers would be free to move any bidder parameters to the request level if they are consistent across imps. Bidder adapters would not need to be updated to support request level parameters.

bretg:
@sachin-pubmatic - please ask your team to give you the time to complete the original feature as specified in #1042 . Thanks.

exchange/utils_test.go
sachin-pubmatic: Fixed this error.

exchange/utils.go
sachin-pubmatic: Fixed it.

sachin-pubmatic:
Thanks @hhhjort for highlighting and summarising the pending changes.

I have incorporated changes for second item i.e merging the request level parameters into all the imp level parameters.  
Third item i.e validate the merged set of imp level bidder parameters, gets automatically taken care off with the existing flow which happens after merging parameters.

Please review and let me know if I am missing anything.

endpoints/openrtb2/auction.go
```diff
@@ -314,6 +320,68 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
+			impCpy = append(impCpy, updatedImp)
+			continue
+		}
+
+		err := json.Unmarshal(imp.Ext, &impExt)
+		if err != nil {
+			return fmt.Errorf("error decoding imp[].ext : %s", err.Error())
+		}
+
+		for bidder, impParams := range impExt {
```
hhhjort, line 348: Currently the preferred location for bidder parameters is `imp[].ext.prebid. bidder.BIDDER`, rather than `imp[].ext.BIDDER`. We have code further down the chain to preserve backwards compatibility, which makes merging at this point in the code more difficult.

 However since this is a new feature, I would be happy with merging only working with bidder parameters in the newer preferred location, rather than the legacy location as you have it now. This should just be simple change digging a little deeper to find the map of bidders to bidder parameters in `imp[].ext.prebid.bidder` rather than `imp[].ext`.

adapters/openrtb_util.go
```diff
@@ -172,3 +174,29 @@ func copyFormats(sizes []openrtb2.Format) []openrtb2.Format {
 	}
 	return sizesCopy
 }
+
+func ExtractBidderParams(bidRequest *openrtb2.BidRequest) (map[string]json.RawMessage, error) {
+	if bidRequest == nil {
+		return nil, fmt.Errorf("error bidRequest should not be nil")
```
SyntaxNode, line 180: Nitpick: There is no need to use `fmt.Errorf` here for a fixed string. Please use `errors.New` instead. Not that it matters much.. this condition will never be hit in practice since the adapter is never called with a nil request. I'm also fine if you just want to remove this check.

adapters/pubmatic/pubmatic.go
```diff
@@ -337,11 +342,17 @@ func (a *PubmaticAdapter) Call(ctx context.Context, req *pbs.PBSRequest, bidder
 func (a *PubmaticAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
 	errs := make([]error, 0, len(request.Imp))
 
-	wrapExt := ""
+	wrapperExt := new(pubmaticWrapperExt)
```
SyntaxNode, line 345: Nitpick: It's idiomatic in Go to use `&pubmaticWrapperExt{}`.

adapters/pubmatic/pubmatic.go
```diff
@@ -355,8 +366,25 @@ func (a *PubmaticAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *ad
...
+			}
+			if wrpExt.VersionID != 0 {
+				wrapperExt.VersionID = wrpExt.VersionID
+				wrapperExtSet = true
+			}
+		}
+	}
+
+	if wrapperExtSet {
+		jsonData, _ := json.Marshal(wrapperExt)
```
SyntaxNode, line 386: Please do not ignore errors from `json.Marshal`.

adapters/pubmatic/pubmatic.go
```diff
@@ -355,8 +366,25 @@ func (a *PubmaticAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *ad
...
+		rawExt := fmt.Sprintf("{\"wrapper\": %s}", string(jsonData))
```
SyntaxNode, line 387: There's always a risk of introducing a json injection attack when manually composing a json blob. Since you're already performing a marshal, consider introducing a wrapper container struct.

endpoints/openrtb2/auction.go
```diff
@@ -314,6 +320,68 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
+
+	reqBidderParams, err := adapters.ExtractBidderParams(req.BidRequest)
+	if err != nil {
+		return err
+	}
+
+	impCpy := make([]openrtb2.Imp, 0)
+	for _, imp := range req.BidRequest.Imp {
+		updatedImp := imp
+		var impExt map[string]map[string]interface{}
```
SyntaxNode, line 337: Nitpick: Please move the variable definition closer to it's variable usage by the Unmarshal call.

endpoints/openrtb2/auction.go
SyntaxNode: +1. There be dragons here with the two locations available. We should also be careful as this is the 3rd implementation of code which searches for bidder exts. The code in `splitImps` is the newest iteration. Perhaps parts can be extracted to a method and used here? Or as @hhhjort said, just make this available for the newer `imp[].ext.prebid.bidder` location. I'm not sure if that would be too confusing for publishers.

endpoints/openrtb2/auction_test.go
```diff
@@ -3715,3 +3715,90 @@ type fakeUUIDGenerator struct {
...
+				hardcodedResponseIPValidator{response: true},
+			}
+
+			req := httptest.NewRequest("POST", "/openrtb2/auction", strings.NewReader(tt.args.reqBody))
+
+			resReq, _, errL := deps.parseRequest(req)
+
+			var expIExt, iExt map[string]interface{}
+			_ = json.Unmarshal(tt.expectedImpExt, &expIExt)
+			_ = json.Unmarshal(resReq.BidRequest.Imp[0].Ext, &iExt)
```
SyntaxNode, line 3777: Please don't ignore errors. You can assert that the function call has no error instead. This will be easier to find if the test framework reports on a specific unmarshal issue.

exchange/utils.go
```diff
@@ -217,6 +218,12 @@ func getAuctionBidderRequests(req AuctionRequest,
 		return nil, []error{err}
 	}
 
+	var bidderExt map[string]json.RawMessage
+	bidderExt, err = adapters.ExtractBidderParams(req.BidRequest)
```
SyntaxNode, line 222: Does `ExtractBidderParams` return the bidder ext? Looks like it returns the bidder params, and if so - consider renaming the variable here please.

exchange/utils.go
```diff
@@ -227,21 +234,32 @@ func getAuctionBidderRequests(req AuctionRequest,
...
+		if len(bidderExt) != 0 {
+			if bidderParams, ok := bidderExt[bidder]; ok {
+				if requestExt == nil {
+					requestExt = &openrtb_ext.ExtRequest{}
+				}
+				requestExt.Prebid.BidderParams = bidderParams
+			} else if requestExt != nil {
+				requestExt.Prebid.BidderParams = nil
+			}
+		}
```
SyntaxNode, line 255: It's difficult for me to understand what this block of code is doing. Consider extracting to a method and use it's name to convey purpose / meaning.

SyntaxNode (commented):
Thanks for picking up this older spec.

endpoints/openrtb2/auction.go
sachin-pubmatic: Thanks @hhhjort & @SyntaxNode for review. I have updated pull request with code review comments. 
Currently I am merging request level bidder params at both new (imp[].ext.prebid.bidder) level & old (imp[].ext) level.
Let me know if I should remove support for old level from pull request.

adapters/openrtb_util.go
sachin-pubmatic: Fixed.

adapters/pubmatic/pubmatic.go
sachin-pubmatic: Fixed.

adapters/pubmatic/pubmatic.go
sachin-pubmatic: Fixed.

adapters/pubmatic/pubmatic.go
sachin-pubmatic: Fixed.

endpoints/openrtb2/auction.go
sachin-pubmatic: Fixed.

endpoints/openrtb2/auction_test.go
sachin-pubmatic: Fixed.

exchange/utils.go
sachin-pubmatic: Fixed.

exchange/utils.go
sachin-pubmatic: Currently I am merging request level bidder params at both new (imp[].ext.prebid.bidder) level & old (imp[].ext) level.
Let me know if I should remove support for old level from pull request.

adapters/openrtb_util.go
hhhjort: Github is still showing `fmt.Errorf()`?

mansinahar:
@sachin-pubmatic Can you please rebase this on top of master and resolve the merge conflicts when you have a chance? Thanks.

SyntaxNode:
> rebase

.. or just merge in the latest master branch.

endpoints/openrtb2/auction.go
```diff
@@ -314,6 +320,113 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
+// mergeBidderParams merges bidder parameters passed at req.ext level with imp[].ext level.
+// Preference is given to parameters at imp[].ext level over req.ext level.
+// Parameters at req.ext level are propagated to adapters as is without any validation.
+func mergeBidderParams(req *openrtb_ext.RequestWrapper) error {
+	reqBidderParams, err := adapters.ExtractReqExtBidderParams(req.BidRequest)
+	if err != nil {
+		return err
+	}
+
+	impCpy := make([]openrtb2.Imp, 0)
```
hhhjort, line 332: Probably better to make this:
```suggestion
	impCpy := make([]openrtb2.Imp, len(req.BidRequest.Imp), 0)
```
This would preallocate enough space in the array under the slice for all the imps, avoiding and memory reallocations as we append to the slice.

exchange/utils.go
hhhjort: It must be confusing as I am reading this block of code as doing something completely different. It is modifying the `request.ext.prebid.bidderparams` of the bidder specific copy of a request to only contain the bidder parameters of the bidder in questions, hiding info about all other bidders from this particular bidder. Some comments or removal to its own function may make this all clearer.

exchange/utils.go
```diff
@@ -227,21 +233,35 @@ func getAuctionBidderRequests(req AuctionRequest,
...
+
+			params, err := getBidderParamsForBidder(bidderParamsInReqExt, bidder)
+			if err != nil {
+				return nil, []error{err}
+			}
+
+			if params != nil && requestExt == nil {
+				requestExt = &openrtb_ext.ExtRequest{}
+			}
+			requestExt.Prebid.BidderParams = params
```
hhhjort, line 255: This will cause an error if `params` is nil and `requestExt` is nil. If both are nil, the above if fails, but we should not have to do anything.

endpoints/openrtb2/auction.go
mansinahar: I believe the order here should be `len, capacity` and so it should look like:
```
impCpy := make([]openrtb2.Imp, 0, len(req.BidRequest.Imp))
```

hhhjort: My mistake, thanks for the correction Mansi.

sachin-pubmatic: Fixed

exchange/utils.go
sachin-pubmatic: Removed confusing check for if params != nil && requestExt == nil. It is not required as we already have check if len(bidderParamsInReqExt) != 0 at line number 245. Code will not enter if block at line 245 if requestExt is nil.

adapters/openrtb_util.go
sachin-pubmatic: Fixed

sachin-pubmatic:
Updated the PR with suggested code changes. Please verify.

exchange/utils.go
sachin-pubmatic: Added comment in code block to explain the logic.

endpoints/openrtb2/auction.go
```diff
@@ -314,6 +320,113 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
+			return err
+		}
+		impExtBidder["prebid"]["bidder"] = bidderParamsJson
+	}
+
+	return nil
+}
+
+// mergeBidderParamsInImpExt merges bidder parameters passed at req.ext level with imp[].ext level.
+func mergeBidderParamsInImpExt(impExtBidder map[string]map[string]json.RawMessage, reqExtParams map[string]map[string]json.RawMessage) error {
```
guscarreon, line 414: Just to double-check, `mergeBidderParamsInImpExt` as written wouldn't add the `"complementaryData"` into `req.imp[0].ext`. Was this intended? I realize that  `req.imp[0].ext` takes precedence over the `req.ext` values but, given the "merge" name of this function, it could get implicitly understood that `"complementaryData"` would get "merged" into `req.imp[0].ext`. If we don't intend to add this functionality, should we think of renaming this function to `addMissingRequestExtParams`? (Or something of that sorts)
```
{
  "id": "some-request-id",
  "imp": [
    {
      "id": "some-impression-id",
      "banner": {
        "format": [ { "w": 300, "h": 250 } ]
      },
      "ext": {
        "myBidder": {
          "myField": {
            "data":"value",
            "data2":1234
          }
        }
      }
    }
  ],
  "ext": {
    "prebid": {
      "bidderparams": { 
        "myBidder": {   # <-- myBidder also found here
          "myField": {  # <-- Same field as in req.imp[0].ext
              "complementaryData":1000  # <-- different set of data that adds up to that of req.imp[0].ext
          }
        }
      }
    }
  }
} 
```

exchange/utils.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/prebid/prebid-server/adapters"
```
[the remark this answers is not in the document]
sachin-pubmatic: Fixed

endpoints/openrtb2/auction.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/prebid/prebid-server/adapters"
```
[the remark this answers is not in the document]
sachin-pubmatic: Fixed

endpoints/openrtb2/auction.go
```diff
@@ -314,6 +320,113 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
+		if !isBidderToValidate(bidder) {
+			continue
+		}
+
+		for key, value := range reqExtParams[bidder] {
+			if _, present := bidderExt[key]; !present {
+				bidderExt[key] = value
+			}
+		}
+		impExtBidder[bidder] = bidderExt
```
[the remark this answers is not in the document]
sachin-pubmatic: Incorporated the suggestion. Thanks @guscarreon.

sachin-pubmatic:
Updated the PR with suggested code changes. Please verify.

endpoints/openrtb2/auction.go
```diff
@@ -314,6 +320,117 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
+		for key, value := range reqExtParams[bidder] {
+			if _, present := bidderExt[key]; !present {
+				bidderExt[key] = value
+				wasModified = true
+			}
+		}
+		if wasModified {
+			impExtBidder[bidder] = bidderExt
+		}
+	}
```
guscarreon, line 430: Nitpick: Can we add test scenarios for the uncovered lines in this newly created function?
image

endpoints/openrtb2/auction.go
```diff
@@ -1068,6 +1185,8 @@ func isBidderToValidate(bidder string) bool {
 		return false
 	case openrtb_ext.BidderReservedSKAdN:
 		return false
+	case openrtb_ext.BidderReservedBidder:
+		return false
```
guscarreon, line 1189: Nitpick: Can we add a unit test for the added `case` statement?
image

openrtb_ext/bidders.go
```diff
@@ -36,6 +36,7 @@ const (
 	BidderReservedGeneral BidderName = "general" // Reserved for non-bidder specific messages when using a map keyed on the bidder name.
 	BidderReservedPrebid  BidderName = "prebid"  // Reserved for Prebid Server configuration.
 	BidderReservedSKAdN   BidderName = "skadn"   // Reserved for Apple's SKAdNetwork OpenRTB extension.
+	BidderReservedBidder  BidderName = "bidder"  // Reserved for Apple's SKAdNetwork OpenRTB extension.
```
guscarreon, line 39: Nitpick: I believe the comment was simply copy pasted from the entry above. Can we elaborate a bit more?

endpoints/openrtb2/auction.go
sachin-pubmatic: Added a test case to cover this scenario.

openrtb_ext/bidders.go
sachin-pubmatic: Updated the comment description.

guscarreon (dismissed):
LGTM

hhhjort (dismissed):
LGTM

guscarreon (commented):
A couple more comments before approving. Thank you for addressing our feedback.

adapters/pubmatic/pubmatic.go
```diff
@@ -191,36 +239,34 @@ func parseImpressionObject(imp *openrtb2.Imp, wrapExt *string, pubID *string) er
...
 	var pubmaticExt openrtb_ext.ExtImpPubmatic
 	if err := json.Unmarshal(bidderExt.Bidder, &pubmaticExt); err != nil {
-		return err
+		return wrapExt, pubID, err
 	}
 
-	if *pubID == "" {
-		*pubID = strings.TrimSpace(pubmaticExt.PublisherId)
+	if extractPubIDFromImp {
+		pubID = strings.TrimSpace(pubmaticExt.PublisherId)
```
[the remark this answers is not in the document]
sachin-pubmatic: We are passing boolean to indicate whether to extract publisher ID value from impression or not. This is just to avoid calling strings.TrimSpace function unnecessarily as parseImpressionObject function gets called for all impressions in the request and we need to extract publisher ID just once from any impression.
Earlier code was passing string pointer and setting value in the called function itself. @mansinahar pointed that it should be good to just extract value in the function and return it instead of setting value. So refactored the code to return publisher ID value.
Passing pubId pointer and updating value of it in parseImpressionObject function was misleading so refactored the code to extract and return publisher ID value on if necessary.

adapters/pubmatic/pubmatic.go
```diff
@@ -253,7 +299,25 @@ func parseImpressionObject(imp *openrtb2.Imp, wrapExt *string, pubID *string) er
...
+	}
+
+	//get request ext bidder params
+	if wrapperObj, present := reqExtBidderParams["wrapper"]; present && len(wrapperObj) != 0 {
+		err = json.Unmarshal(wrapperObj, &wrpExt)
+		if err == nil {
+			return &wrpExt, nil
+		}
+	}
+	return nil, nil
```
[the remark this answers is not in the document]
sachin-pubmatic: Fixed

guscarreon (approved):
LGTM. @sachin-pubmatic thank you for addressing our feedback

sachin-pubmatic:
We have completed QA verification. We are good to merge the PR.

hhhjort (approved):

PR #1995, opened 2021-09-13
Adot: Add Usersync

savarino-o:


static/bidder-info/adot.yaml
```diff
@@ -12,3 +12,8 @@ capabilities:
       - banner
       - video
       - native
+userSync:
+  redirect:
+    url: "https://sync.adotmob.com/cookie/pbs?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&r={{.RedirectURL}}"
+    userMacro: "{amob_user_id}"
+    redirectUrl: "{{.ExternalURL}}/setuid?bidder={{.SyncerKey}}&f={{.SyncType}}&uid={{.UserMacro}}"
```
SyntaxNode, line 19: You should not define the `redirectUrl`. That's reserved for hosts to configure since it's a redirect back to the host's server.

static/bidder-info/adot.yaml
```diff
@@ -12,3 +12,8 @@ capabilities:
       - banner
       - video
       - native
+userSync:
+  redirect:
+    url: "https://sync.adotmob.com/cookie/pbs?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&r={{.RedirectURL}}"
```
SyntaxNode, line 17: I tested this syncer and found that a `gdpr_consent` is always added to the end of the redirect url. The adapter should not change any part of the redirect url except for the user macro replacement.

savarino-o: I updated the syncer and it should now always use the redirect url in the query param with no modification except for the uid macro. Can you tell me if it looks good to you ?

SyntaxNode (approved):

mansinahar (approved):

PR #1996, opened 2021-09-13
Fix UserSync Config Case Sensitivity Issue

SyntaxNode:
Fixes an issue where the host configured user sync override is ignored due to a case sensitivity issue between the adapter config (always lower case) and bidder info config (potentially mixed case).


VeronikaSolovei9 (approved):
LGTM

guscarreon (approved):
LGTM

PR #2000, opened 2021-09-15
Adding Support for Macro {{UUID}} to Support Feature: Generate Unique Bidrequest ID

AlexBVolcy:
In the original issue https://github.com/prebid/prebid-server/issues/1507, there was a request to add support for a macro ID, where if Prebid Server sees {{UUID}} then resolve to a UUID.

I didn't include this feature in the already merged issue https://github.com/prebid/prebid-server/pull/1938

So this pull request is to add support for that feature.

### DIFF
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -348,7 +348,7 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req 
 		return
 	}
 
-	if deps.cfg.GenerateRequestID {
+	if deps.cfg.GenerateRequestID || req.ID == "{{UUID}}" {
 		newBidRequestId, err := deps.uuidGenerator.Generate()
 		if err != nil {
 			errs = []error{err}
--- a/endpoints/openrtb2/amp_auction_test.go
+++ b/endpoints/openrtb2/amp_auction_test.go
@@ -1304,9 +1304,15 @@ func TestIdGeneration(t *testing.T) {
 			expectedID:             "ThisID",
 		},
 		{
-			description:            "The givenGenerateRequestID flag is true, and if the id field isn't included in the stored request",
+			description:            "The givenGenerateRequestID flag is true, and the id field isn't included in the stored request, we should still generate a uuid",
 			givenInStoredRequest:   json.RawMessage(`{"site":{"page":"prebid.org"},"imp":[{"id":"some-imp-id","banner":{"format":[{"w":300,"h":250}]},"ext":{"appnexus":{"placementId":1}}}],"tmax":1}`),
 			givenGenerateRequestID: true,
+			expectedID:             uuid,
+		},
+		{
+			description:            "The givenGenerateRequestID flag is false, but id field is the macro option {{UUID}}, we should generate a uuid",
+			givenInStoredRequest:   json.RawMessage(`{"id":"{{UUID}}","site":{"page":"prebid.org"},"imp":[{"id":"some-imp-id","banner":{"format":[{"w":300,"h":250}]},"ext":{"appnexus":{"placementId":1}}}],"tmax":1}`),
+			givenGenerateRequestID: false,
 			expectedID:             uuid,
 		},
 	}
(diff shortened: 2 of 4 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -1356,10 +1356,15 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+	bidRequestID, err := getBidRequestID(requestJson)
+	if err != nil {
+		return nil, nil, []error{err}
+	}
+
 	// Apply the Stored BidRequest, if it exists
 	resolvedRequest := requestJson
 
-	if deps.cfg.GenerateRequestID {
+	if deps.cfg.GenerateRequestID || bidRequestID == "{{UUID}}" {
```
SyntaxNode, line 1367: Does this work for AMP requests too?

endpoints/openrtb2/auction_test.go
```diff
@@ -3616,7 +3622,7 @@ var testStoredRequestsUuid = []string{
 		}
 	}`,
 	`{
-		"id": "ThisID",
+		"id": "{{UUID}}",
```
hhhjort, line 3625: With this change, there is no longer a test where we generate a uuid where we don't have the `{{UUID}}` macro. Shouldn't we still have a test that shows this case still works?

endpoints/openrtb2/auction.go
AlexBVolcy: This doesn't work for AMP, so I'll update this PR so that it does!

hhhjort (dismissed):
LGTM

endpoints/openrtb2/amp_auction.go
```diff
@@ -348,7 +348,7 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
 		return
 	}
 
-	if deps.cfg.GenerateRequestID {
+	if deps.cfg.GenerateRequestID || req.ID == "{{UUID}}" {
```
SyntaxNode, line 351: Is `req.ID` referencing the stored request, or the original request? The token should only apply to when it's used in the stored request.

endpoints/openrtb2/auction.go
SyntaxNode, line 1367: This looks to be checking for the token in the original bid request, but the spec says to check for the token in just the stored request.

endpoints/openrtb2/amp_auction.go
hhhjort: In AMP, there is no original request. So it would have to be the stored request. (Or possibly the query string)

hhhjort (approved):
LGTM

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1304,11 +1304,17 @@ func TestIdGeneration(t *testing.T) {
...
 			givenInStoredRequest:   json.RawMessage(`{"site":{"page":"prebid.org"},"imp":[{"id":"some-imp-id","banner":{"format":[{"w":300,"h":250}]},"ext":{"appnexus":{"placementId":1}}}],"tmax":1}`),
 			givenGenerateRequestID: true,
 			expectedID:             uuid,
 		},
+		{
+			description:            "The givenGenerateRequestID flag is false, but id field is the macro option {{UUID}}, we should generate a uuid",
+			givenInStoredRequest:   json.RawMessage(`{"id":"{{UUID}}","site":{"page":"prebid.org"},"imp":[{"id":"some-imp-id","banner":{"format":[{"w":300,"h":250}]},"ext":{"appnexus":{"placementId":1}}}],"tmax":1}`),
+			givenGenerateRequestID: false,
+			expectedID:             uuid,
+		},
```
SyntaxNode, line 1317: Please add a test case for "{{uuid}}" to demonstrate the macro is case sensitive.

endpoints/openrtb2/auction_test.go
```diff
@@ -1180,26 +1180,38 @@ func TestStoredRequestGenerateUuid(t *testing.T) {
...
 			givenGenerateRequestID: false,
 			expectedID:             "ThisID",
 		},
 		{
 			description:            "GenerateRequestID is true, but rawData is a site request, we should not generate uuid",
-			givenRawData:           testStoredRequestsUuid[1],
+			givenRawData:           testBidRequests[1],
 			givenGenerateRequestID: true,
 			expectedID:             "ThisID",
 		},
```
SyntaxNode, line 1217: Please add a test case for "{{uuid}}" to demonstrate the macro is case sensitive.

SyntaxNode (approved):

hhhjort (approved):

PR #2007, opened 2021-09-22
Remove Dummy Terminology From Core Code

SyntaxNode:
Replaces the term "Dummy" with the more appropriate and professional "Nil" or "Test" in PBS-Core code. Except for the legacy cache, as that is tied to a host configuration variable and will be removed in the near future anyway.


cache/dummycache/dummycache.go
```diff
@@ -38,11 +38,6 @@ func (s *accountService) Get(id string) (*cache.Account, error) {
 	}, nil
 }
 
-// Set will always return nil since this is a dummy service
-func (s *accountService) Set(account *cache.Account) error {
-	return nil
-}
```
SyntaxNode, line 44: I was considering renaming this concept, when I noticed all implements just return `nil` and is never called.

exchange/utils_test.go
```diff
@@ -1936,22 +1936,39 @@ func newBidRequest(t *testing.T) *openrtb2.BidRequest {
...
+		copy(biddersWorkingCopy, test.bidders)
+
+		randomizeList(biddersWorkingCopy)
 
-	if len(adapters) != 1 {
-		t.Errorf("RandomizeList, expected a list of 1, found %d", len(adapters))
+		// test all bidders are still present, ignoring order. we are testing the algorithm doesn't loose
+		// elements. we are not testing the random number generator itself.
+		assert.ElementsMatch(t, test.bidders, biddersWorkingCopy)
 	}
```
SyntaxNode, line 1972: Updated the test code while I was here removing "dummy" references.

metrics/config/metrics.go
```diff
@@ -258,117 +258,118 @@ func (me *MultiMetricsEngine) RecordAdapterGDPRRequestBlocked(adapter openrtb_ex
 	}
 }
 
-// DummyMetricsEngine is a Noop metrics engine in case no metrics are configured. (may also be useful for tests)
-type DummyMetricsEngine struct{}
+// NilMetricsEngine implements the MetricsEngine interface where no metrics are actually captured. This is
+// used if no metric backend is configured and also for tests.
+type NilMetricsEngine struct{}
```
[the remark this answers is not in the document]
SyntaxNode: That object already exists. It implements Testify's Mock library for some tests. This object throws away / ignores all metric calls.

guscarreon (commented):
Some adapters still use `Dummy`. I guess we could either their authors to remove it or remove it ourselves in a future PR. I believe we could remove `Dummy` from `adapters/appnexus/appnexus_test.go` ourselves without much hassle though. Do you think we should do it as part of this PR? (should be quick)

SyntaxNode:
> Some adapters still use `Dummy`. I guess we could either their authors to remove it or remove it ourselves in a future PR. I believe we could remove `Dummy` from `adapters/appnexus/appnexus_test.go` ourselves without much hassle though. Do you think we should do it as part of this PR? (should be quick)

I didn't bother with changing references in code for the legacy auction endpoint since it will be removed in the very near future.

guscarreon (approved):
LGMT

hhhjort (approved):

Issue #1172, opened 2020-01-22
Prebid Server Version Info


SyntaxNode:
We have received several requests to add the ability to retrieve the version number of Prebid Server. We discussed extending the existing `/status` endpoint, but the response is not json and thus would require a breaking change to extend. I propose we add a new '/version' endpoint which will respond with the following json including both the friendly version number and the git hash:

```
{
  "version": "0.98.0",
  "hash": "bd43afbca80114e76d1b14a2843b71f8a5f62894"
}
```
The default values would be "0.0.0" for the version and an empty string for the hash for local builds.

Questions:
1) Should we include the name/path of the repository as well to account for potential forked builds?

It may also be worthwhile to include the version information in the http response headers, so in the case of a mixed deploy scenario you would know exactly which version handled the request, such as:

```
Server: prebid/0.98.0
```
.. although I'm not certain on the name of the header to avoid it getting overwritten by networking proxies like nginx. So perhaps something like `X-Server`.

benjaminch:
Hey @SyntaxNode 

I think there is already a `/version` endpoint for PBS.
[Endpoint implementation](https://github.com/prebid/prebid-server/blob/master/endpoints/version.go) and [tests](https://github.com/prebid/prebid-server/blob/master/endpoints/version_test.go)

[Original issue](https://github.com/prebid/prebid-server/issues/559) and [PR](https://github.com/prebid/prebid-server/pull/577).

This can be extended if needed though :)

Cheers

SyntaxNode:
Thank you for posting the original issue.

The existing endpoint is exclusive to the admin port which the users of the service cannot access. I don't personally see the need to keep this information restricted. The original argument against making it publicly available is two fold:

> Since anyone can easily compile any version with any -ldflags, it's quite fragile. I'd rather that publishers didn't learn about this endpoint and believe they could trust it.

That is certainly true, but that logic applies to any open source project. I don't view this as a deal breaker.

> This keeps our API's surface area smaller, which is good for project maintenance. Since your use-case is a dev one, we don't need to expose it to publishers.

I'm not worried about the level of effort needed to support a public version endpoint.

I'll update the proposal to consider the existing admin only endpoint. For the sake of compatibility, we may wish to leave it alone. I don't want to break anyone and I have no idea if any other hosts are making use of that endpoint for reporting.

benjaminch:
[...]

PR #2009, opened 2021-09-24
Version Endpoint

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/1172

### DIFF
--- a/endpoints/version.go
+++ b/endpoints/version.go
@@ -7,11 +7,11 @@ import (
 	"github.com/golang/glog"
 )
 
-const versionNotSet = "not-set"
+const versionEndpointValueNotSet = "not-set"
 
 // NewVersionEndpoint returns the latest git tag as the version and commit hash as the revision from which the binary was built
 func NewVersionEndpoint(version, revision string) http.HandlerFunc {
-	response, err := prepareVersionResponse(version, revision)
+	response, err := prepareVersionEndpointResponse(version, revision)
 	if err != nil {
 		glog.Fatalf("error creating /version endpoint response: %v", err)
 	}
@@ -21,12 +21,12 @@ func NewVersionEndpoint(version, revision string) http.HandlerFunc {
 	}
 }
 
-func prepareVersionResponse(version, revision string) (json.RawMessage, error) {
+func prepareVersionEndpointResponse(version, revision string) (json.RawMessage, error) {
 	if version == "" {
-		version = versionNotSet
+		version = versionEndpointValueNotSet
 	}
 	if revision == "" {
-		revision = versionNotSet
+		revision = versionEndpointValueNotSet
 	}
 
 	return json.Marshal(struct {
--- a/endpoints/version.go
+++ b/endpoints/version.go
@@ -7,31 +7,33 @@ import (
 	"github.com/golang/glog"
 )
 
-type versionModel struct {
-	Version  string `json:"version"`
-	Revision string `json:"revision"`
-}
+const versionNotSet = "not-set"
 
 // NewVersionEndpoint returns the latest git tag as the version and commit hash as the revision from which the binary was built
-func NewVersionEndpoint(version string, revision string) http.HandlerFunc {
-	if version == "" {
-		version = "not-set"
-	}
-	if revision == "" {
-		revision = "not-set"
+func NewVersionEndpoint(version, revision string) http.HandlerFunc {
+	response, err := prepareVersionResponse(version, revision)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


endpoints/version.go
```diff
@@ -7,31 +7,33 @@ import (
...
+	}
+
+	return json.Marshal(struct {
+		Revision string `json:"revision"`
+		Version  string `json:"version"`
+	}{
+		Revision: revision,
+		Version:  version,
+	})
 }
```
SyntaxNode, line 39: Updated the version endpoint to create a static response once at startup, mimicking the pattern used for the bidder info handler. Changed response structure to list the revision first to match PBS-Java. This is not a breaking change for the admin endpoint and we might as well make it match for the public debut.

router/router.go
```diff
@@ -310,6 +311,7 @@ func New(cfg *config.Configuration, rateConvertor *currency.RateConverter) (r *R
 	r.POST("/cookie_sync", endpoints.NewCookieSyncEndpoint(syncersByBidder, cfg, gdprPerms, r.MetricsEngine, pbsAnalytics, activeBidders).Handle)
 	r.GET("/status", endpoints.NewStatusEndpoint(cfg.StatusResponse))
 	r.GET("/", serveIndex)
+	r.Handler("GET", "/version", endpoints.NewVersionEndpoint(version.Ver, version.Rev))
```
SyntaxNode, line 314: The endpoint registration looks different because it's adapting to the http.Handle interface already in use by the admin endpoint. The `r.GET` call is just syntactical sugar around this anyway.

endpoints/version.go
```diff
@@ -7,31 +7,33 @@ import (
...
+	}
+}
 
-		w.Write(jsonOutput)
+func prepareVersionResponse(version, revision string) (json.RawMessage, error) {
+	if version == "" {
+		version = versionNotSet
 	}
+	if revision == "" {
+		revision = versionNotSet
```
guscarreon, line 29: Nitpick: This could be a bit confusing. Ideally we should assign `revisionNotSet` and not `versionNotSet` to `revision`. Do you think renaming `versionNotSet` to either `notSet` or `valueNotSet` could be more straightforward?

endpoints/version.go
```diff
@@ -7,31 +7,33 @@ import (
...
 	}
 
 	return func(w http.ResponseWriter, _ *http.Request) {
-		jsonOutput, err := json.Marshal(versionModel{
-			Version:  version,
-			Revision: revision,
-		})
-		if err != nil {
-			glog.Errorf("/version Critical error when trying to marshal versionModel: %v", err)
-			w.WriteHeader(http.StatusInternalServerError)
```
guscarreon, line 31: I see we are not using `w.WriteHeader(http.StatusInternalServerError)` in the new version. Why are we not writing this error anymore?

endpoints/version_test.go
```diff
@@ -1,52 +1,43 @@
...
+			version:     "",
+			revision:    "",
+			expected:    `{"revision":"not-set","version":"not-set"}`,
+		},
+		{
+			description: "Populated",
+			version:     "1.2.3",
+			revision:    "d6cd1e2bd19e03a81132a23b2025920577f84e37",
+			expected:    `{"revision":"d6cd1e2bd19e03a81132a23b2025920577f84e37","version":"1.2.3"}`,
+		},
```
guscarreon, line 29: Can we include these additional two tests?
```
    {
        description: "Empty",
        version:     "1.2.3",
        revision:    "",
        expected:    `{"revision":"not-set","version":"1.2.3"}`,
    },
    {
    	description: "Populated",
    	version:     "",
    	revision:    "d6cd1e2bd19e03a81132a23b2025920577f84e37",
    	expected:    `{"revision":"d6cd1e2bd19e03a81132a23b2025920577f84e37","version":"not-set"}`,
    },
```

endpoints/version.go
SyntaxNode: I moved the json marshal outside of the endpoint, so that it happens only once at application startup. If it fails, PBS will panic. Thus, it's no longer possible for the error to occur within the endpoint.

endpoints/version_test.go
SyntaxNode: Sure.

endpoints/version.go
SyntaxNode: Renamed for clarity. I cannot rename it `notSet` or `valueNotSet` because it's in the same package (naming scope) as other endpoints.

mansinahar (approved):

bsardo (approved):
LGTM

PR #2014, opened 2021-09-24
New Adapter: Adnuntius

mikael-lundin:
Hi, this is a pull request for a new adapter that will be for adnuntius adserver. This is my first time using go, please be gentle and if there's anything you need just tell me and I'll provide. :)


bsardo:
Hi @mikael-lundin. Thanks for your submission. We'll get to this shortly but could you open up a docs PR [here](https://github.com/prebid/prebid.github.io) in the meantime?

mikael-lundin:
I've added information about server side here. :)
https://github.com/prebid/prebid.github.io/pull/3312

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+		consent = extUser.Consent
+	}
+
+	return gdpr, consent, nil
+}
+
+func generateBidResponse(adnResponse *AdnResponse, request *openrtb2.BidRequest) (*adapters.BidderResponse, []error) {
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(adnResponse.AdUnits))
+	currency := bidResponse.Currency
```
bsardo, line 251: It looks like you might just want to declare currency here: `var currency string`.

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+
+	for i, adunit := range adnResponse.AdUnits {
+
+		var bid openrtb2.Bid
+		if len(adunit.Ads) > 0 {
+			ad := adunit.Ads[0]
+
+			currency = ad.Bid.Currency
```
bsardo, line 259: I'm curious about what you're doing with currency. Currency is set on the bidResponse. It looks like you iterate over all ad units and only process the first ad for an adunit. You're then setting the currency to that ad's bid currency. That means that the currency you select will be the currency of the last ad you process since you're assigning `currency` to `bidResponse.currency` on line 298 after you're done iterating over all adunits. This should be fine if it is guaranteed all ad bids will have currency set.

adapters/adnuntius/adnuntiustest/exemplary/simple-banner.json
```diff
@@ -0,0 +1,149 @@
...
+                                    "cost": {
+                                        "amount": 20.0,
+                                        "currency": "NOK"
+                                    },
+                                    "impressionTrackingUrls": [],
+                                    "impressionTrackingUrlsEsc": [],
+                                    "adId": "adn-id-1559784094",
+                                    "selectedColumn": "0",
+                                    "selectedColumnPosition": "0",
+                                    "renderedPixel": "https://ads.adnuntius.delivery/b/n6xZq-VhchUD8PbfezltPoCESuAfspN-w0wjkkwzHXsAAAAQCtjQz9kbGWD4nuZy3q6HaHGLVNv__f6bBEfVP76Mfd6iokhXOQf7XpiOka60HChCnQjfT3AAqAO1swCIxQR73xnw2Okk_huqOKANzHxOy9gOZB8wEKOW8bs-Po5kTzB8U8UaB4BF9d48j6zHV7ST6-6cRBhpWFyKZUaUDEgmwKXM8xlQC-OJY7F9UJ0rN8OZQis1rnmeXrbik9VvxuV8Owsd_I9ZMS4G.html",
```
bsardo, line 92: Just a suggestion: You could probably simplify your tests quite a bit by paring down what you're returning in your `mockResponse` as most of this isn't used by your adapter.

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+			creativeHeight, Herr := strconv.ParseInt(ad.CreativeHeight, 10, 64)
+			if Herr != nil {
+				return nil, []error{Herr}
+			}
+
+			adDomain := []string{}
+			for _, url := range ad.DestinationUrls {
+				domainArray := strings.Split(url, "/")
+				domain := strings.Replace(domainArray[2], "www.", "", -1)
+				adDomain = append(adDomain, domain)
```
bsardo, line 275: Are the destination URLs always going to start with `http(s)://`?
Nitpick: maybe split on `//` and grab `domainArray[1]` instead? IMO that seems a little more straightforward.

bsardo (commented):
@mikael-lundin Nice work; I left you some comments. Let me know if you have any questions.

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+			Headers: headers,
+		})
+
+	}
+
+	return requestData, nil
+}
+
+func (a *adapter) MakeBids(request *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if response.StatusCode == http.StatusNoContent {
```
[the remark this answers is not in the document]
mikael-lundin: Sorry, we don't have a 204 status in our adserver.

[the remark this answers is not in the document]
mikael-lundin: Ah, but then again I don't need the noContent - status :) will clear this up.

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+		return nil, nil
+	}
+
+	if response.StatusCode == http.StatusBadRequest {
+		return nil, []error{&errortypes.BadInput{
+			Message: fmt.Sprintf("Unknown status code: %d.", response.StatusCode),
+		}}
+	}
+
+	if response.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
mikael-lundin: we only deal in 400 or 200 for this endpoint. :)

adapters/adnuntius/adnuntius.go
mikael-lundin: Yeah, that seems like a good idea but I got the impression that i was only allowed to send the top level domain in that field? :D If not I can just stop what that stuff is doing and just pass the entire url. And yes, we force the url to have http(s)://

adapters/adnuntius/adnuntius.go
mikael-lundin: yeah, this is just me getting around how our ad server works. there will always be a currency set in the ad unit response.

mikael-lundin:
As a general note, I think I've managed to fix the issues mentioned in this pull request, please let me know if I'm missing anything. :)

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+				Message: fmt.Sprintf("Error unmarshalling ExtImpBidder: %s", err.Error()),
+			})
+		}
+
+		var adnuntiusExt openrtb_ext.ImpExtAdnunitus
+		if err := json.Unmarshal(bidderExt.Bidder, &adnuntiusExt); err != nil {
+			errors = append(errors, &errortypes.BadInput{
+				Message: fmt.Sprintf("Error unmarshalling ExtImpBmtm: %s", err.Error()),
+			})
+		}
```
VeronikaSolovei9, line 144: In case of error occurs, lines 128, 134 and 141 - should you just return an error and abort the execution?

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+
+		var bid openrtb2.Bid
+		if len(adunit.Ads) > 0 {
+			var adsErr []error
+			ad := adunit.Ads[0]
+
+			currency = ad.Bid.Currency
+
+			creativeWidth, Werr := strconv.ParseInt(ad.CreativeWidth, 10, 64)
+			if Werr != nil {
```
VeronikaSolovei9, line 258: Nitpick: could you please rename `Werr` to `wErr`?

adapters/adnuntius/adnuntius.go
```diff
@@ -193,19 +197,10 @@ func (a *adapter) generateRequests(ortbRequest openrtb2.BidRequest) ([]*adapters
...
-		return nil, nil
-	}
 
 	if response.StatusCode == http.StatusBadRequest {
 		return nil, []error{&errortypes.BadInput{
-			Message: fmt.Sprintf("Unknown status code: %d.", response.StatusCode),
-		}}
-	}
-
-	if response.StatusCode != http.StatusOK {
```
bsardo: I suggest keeping an error code catch-all in case something goes wrong on your server so PBS can handle it gracefully.

adapters/adnuntius/adnuntiustest/supplemental/width-error.json
```diff
@@ -0,0 +1,84 @@
...
+                                {
+                                    "destinationUrls": {
+                                        "url": "http://www.google.com"
+                                    },
+                                    "bid": {
+                                        "amount": 20.0,
+                                        "currency": "NOK"
+                                    },
+                                    "adId": "adn-id-1559784094",
+                                    "creativeWidth": 980,
```
bsardo, line 65: This is a valid scenario but you're not getting test coverage of lines `if Werr != nil {` and `if len(adsErr) > 0 {`. If you change this to a string that doesn't parse to an int (e.g. "ABC"), the call to `ParseInt` will return the error as `Werr` giving you the code coverage.

adapters/adnuntius/adnuntiustest/supplemental/height-error.json
```diff
@@ -0,0 +1,84 @@
...
+                                    "destinationUrls": {
+                                        "url": "http://www.google.com"
+                                    },
+                                    "bid": {
+                                        "amount": 20.0,
+                                        "currency": "NOK"
+                                    },
+                                    "adId": "adn-id-1559784094",
+                                    "creativeWidth": "980",
+                                    "creativeHeight": 240,
```
bsardo, line 66: This is a valid scenario but you're not getting test coverage of lines `if Herr != nil {` and `if len(adsErr) > 0 {`. If you change this to a string that doesn't parse to an int (e.g. "ABC"), the call to `ParseInt` will return the error as `Herr` giving you the code coverage.

adapters/adnuntius/adnuntius.go
```diff
@@ -102,47 +105,47 @@ func makeEndpointUrl(ortbRequest openrtb2.BidRequest, endpoint string) (string,
...
 		if err := json.Unmarshal(bidderExt.Bidder, &adnuntiusExt); err != nil {
-			return nil, &errortypes.BadInput{
+			errors = append(errors, &errortypes.BadInput{
 				Message: fmt.Sprintf("Error unmarshalling ExtImpBmtm: %s", err.Error()),
-			}
+			})
 		}
 
-		var network string
+		network := "default"
```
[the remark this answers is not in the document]
mikael-lundin: This is because it might change in the lines below, so making it a constant is not correct, I think? :) but then again my first go at golang, (pun intended) so i might be mistaking.

adapters/adnuntius/adnuntius.go
mikael-lundin: Hopefully I've done it as you meant it in my coming commit.

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,300 @@
...
+	if err != nil {
+		return "", []error{fmt.Errorf("failed to parse Adnuntius endpoint: %v", err)}
+	}
+
+	gdpr, consent, err := getGDPR(&ortbRequest)
+	if err != nil {
+		return "", []error{fmt.Errorf("failed to parse Adnuntius endpoint: %v", err)}
+	}
+
+	_, offset := a.tzo.Now().UTC().Local().Zone()
```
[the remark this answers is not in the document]
mikael-lundin: Thanks, have changed this and really hope it works for you now! :D

bsardo (commented):
This is looking good; just two more comments.
Also how did you merge with master? The commit history here looks a little funny. I see several commits from master and I also see each of your original commits twice. Can you clean this up so this PR only has your commits in it? Thanks in advance!

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,311 @@
...
+func (a *adapter) MakeBids(request *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+
+	if response.StatusCode == http.StatusBadRequest {
+		return nil, []error{&errortypes.BadInput{
+			Message: fmt.Sprintf("Status code: %d, Request malformed", response.StatusCode),
+		}}
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return nil, []error{&errortypes.BadInput{
```
bsardo, line 213: Can you change this one to `errortypes.BadServerResponse` please? `BadInput` is typically reserved for invalid input but this catch all should typically handle other types of unexpected errors.

adapters/adnuntius/adnuntius.go
VeronikaSolovei9: Thank you for the update.
Please check it one more time. You probably don't want to collect all errors. Instead return after first error occurrence. If so please modify return `[]error` to just `error` and return it immediately.

mikael-lundin:
yeah, I think I changed them earlier to do something more in the lines of
this:

if response.StatusCode == http.StatusBadRequest {
Message: fmt.Sprintf("Status code: %d, Request malformed", response.
StatusCode),
}}
}

Message: fmt.Sprintf("Status code: %d, Something went wrong with your
request", response.StatusCode),
}}
}

bsardo (approved):
LGTM

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,311 @@
...
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var extRequests []*adapters.RequestData
+	var errs []error
+
+	extRequests, err := a.generateRequests(*request)
+	if err != nil {
+		errs = append(errs, err...)
+	}
+	return extRequests, errs
```
VeronikaSolovei9, line 81: This entire function can look like:

`return a.generateRequests(*request)`

adapters/adnuntius/adnuntius.go
```diff
@@ -0,0 +1,311 @@
...
+*/
+func (a *adapter) generateRequests(ortbRequest openrtb2.BidRequest) ([]*adapters.RequestData, []error) {
+	var requestData []*adapters.RequestData
+	var errors []error
+	networkAdunitMap := make(map[string][]adnAdunit)
+	headers := setHeaders()
+
+	endpoint, err := makeEndpointUrl(ortbRequest, a)
+	if err != nil {
+		errors = append(errors, fmt.Errorf("failed to parse URL: %v", err))
```
VeronikaSolovei9, line 128: Should it be `return` here instead?

VeronikaSolovei9 (approved):
LGTM

PR #2015, opened 2021-09-27
Yieldlab: Video adm and nurl

BertiBauer:
Adds that video ads are returned as VAST wrappers via the 'adm' field in addition to being returned as plain URLs via the 'nurl' field.


adapters/yieldlab/const.go
```diff
@@ -4,4 +4,5 @@ const adSlotIdSeparator = ","
 const adsizeSeparator = "x"
 const adSourceBanner = "<script src=\"%v\"></script>"
 const adSourceURL = "https://ad.yieldlab.net/d/%v/%v/%v?%v"
+const vastMarkup = "<VAST version=\"2.0\"><Ad id=\"1234567890\"><Wrapper><AdSystem>Yieldlab</AdSystem><VASTAdTagURI><![CDATA[ %v ]]></VASTAdTagURI><Impression></Impression><Creatives></Creatives></Wrapper></Ad></VAST>"
```
AlexBVolcy, line 7: I was wondering if you could explain why is the Ad id hard coded here? `<Ad id=\"1234567890\">`

I see this in a couple places, so I'll leave the same comments there as well.

adapters/yieldlab/yieldlabtest/exemplary/video.json
```diff
@@ -124,6 +124,7 @@
             "id": "12345",
             "impid": "test-imp-id",
             "nurl": "https://ad.yieldlab.net/d/12345/123456789/728x90?id=abc&ids=ylid%3A34a53e82-0dc3-4815-8b7e-b725ede0361c&pvid=40cb3251-1e1e-4cfd-8edc-7d32dc1a21e5&ts=testing",
+            "adm": "<VAST version=\"2.0\"><Ad id=\"1234567890\"><Wrapper><AdSystem>Yieldlab</AdSystem><VASTAdTagURI><![CDATA[ https://ad.yieldlab.net/d/12345/123456789/728x90?id=abc&ids=ylid%3A34a53e82-0dc3-4815-8b7e-b725ede0361c&pvid=40cb3251-1e1e-4cfd-8edc-7d32dc1a21e5&ts=testing ]]></VASTAdTagURI><Impression></Impression><Creatives></Creatives></Wrapper></Ad></VAST>",
```
AlexBVolcy, line 127: I was wondering if you could explain why is the Ad id hard coded here? <Ad id=\"1234567890\">

I see this in a couple places, so I'll leave the same comments there as well.

adapters/yieldlab/yieldlabtest/exemplary/video_app.json
```diff
@@ -124,6 +124,7 @@
             "id": "12345",
             "impid": "test-imp-id",
             "nurl": "https://ad.yieldlab.net/d/12345/123456789/728x90?id=abc&ids=ylid%3A34a53e82-0dc3-4815-8b7e-b725ede0361c&pvid=40cb3251-1e1e-4cfd-8edc-7d32dc1a21e5&ts=testing",
+            "adm": "<VAST version=\"2.0\"><Ad id=\"1234567890\"><Wrapper><AdSystem>Yieldlab</AdSystem><VASTAdTagURI><![CDATA[ https://ad.yieldlab.net/d/12345/123456789/728x90?id=abc&ids=ylid%3A34a53e82-0dc3-4815-8b7e-b725ede0361c&pvid=40cb3251-1e1e-4cfd-8edc-7d32dc1a21e5&ts=testing ]]></VASTAdTagURI><Impression></Impression><Creatives></Creatives></Wrapper></Ad></VAST>",
```
AlexBVolcy, line 127: I was wondering if you could explain why is the Ad id hard coded here? <Ad id=\"1234567890\">

I see this in a couple places, so I'll leave the same comments there as well.

adapters/yieldlab/const.go
BertiBauer: The ID was hardcoded because the ORTB specs require an ID (otherwise we would simply have omitted it). Our system doesn't generate a unique ID outside of Prebid either though  - when we receive just a VAST-URL, we put a VAST wrapper around it (similar to what we do now for the adm field) that uses the adslot ID as the ad ID. I have now updated the code here accordingly, so our behaviour is identical in both cases.

adapters/yieldlab/yieldlabtest/exemplary/video.json
BertiBauer: See above.

adapters/yieldlab/yieldlabtest/exemplary/video_app.json
BertiBauer: See above.

mansinahar (approved):

AlexBVolcy (approved):

PR #2028, opened 2021-09-29
Update invalid-response-no-bids.json

ddantuonobeintoo:


mansinahar:
@ddantuonobeintoo Looks like you're trying to update instances of `beintoo` to `beintoo4`. Is that correct? If so, can you please open up a single PR with all the file changes rather than a separate PR for each file change?

mansinahar:
I will leave this one open and close the others out. Please add the rest of the file changes in this PR

Issue #2011, opened 2021-09-24
Remove The Legacy /auction Endpoint


SyntaxNode:
## Type Of Issue
Feature Removal

## Description
PBS-Go still implements the original /auction endpoint, utilizing a custom data model instead of OpenRTB. This endpoint has long been deprecated and removed from public facing documentation. It is difficult to maintain and is barely used. Hosts report less than 0.1% of all requests are using the endpoint. It's time has come.

## Proposal
Remove the original /auction endpoint in a phased approached as follows:
- 0.179.0: Include a config to enable the /auction endpoint, defaulting to true. Hosts should experiment with setting this option to false in staging/testing environments to ensure there are no surprises when it's removed.
- 0.180.0: Change the default to false. Hosts may set the flag to true if they encounter an unexpected problem.
- 0.181.0: Remove the endpoint entirely and all related code, including all adapter `Call` methods.

SyntaxNode:
Discussed in the Preber Server Commitee. There is no opposition to this plan.

SyntaxNode:
PBS-Go 0.179.0 was released as planned with a new config option `enable_legacy_auction` to disable the legacy /auction endpoint. It defaults to true, but will default to false in the upcoming 0.180.0 release.

SyntaxNode:
Removal of the /auction endpoint code has been delayed to 0.182.0 to give bidders a bit more time to review their adapters.

PR #2032, opened 2021-10-05
Disable Legacy /auction Endpoint By Default

SyntaxNode:
Implements phase 2 of https://github.com/prebid/prebid-server/issues/2011.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -144,7 +144,7 @@ func TestDefaults(t *testing.T) {
 	cmpStrings(t, "stored_requests.filesystem.directorypath", "./stored_requests/data/by_id", cfg.StoredRequests.Files.Path)
 	cmpBools(t, "auto_gen_source_tid", cfg.AutoGenSourceTID, true)
 	cmpBools(t, "generate_bid_id", cfg.GenerateBidID, false)
-	cmpBools(t, "enable_legacy_auction", cfg.EnableLegacyAuction, true)
+	cmpBools(t, "enable_legacy_auction", cfg.EnableLegacyAuction, false)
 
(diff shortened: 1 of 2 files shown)


AlexBVolcy (approved):
LGTM!

config/config_test.go
```diff
@@ -354,7 +354,7 @@ request_validation:
     ipv4_private_networks: ["1.1.1.0/24"]
     ipv6_private_networks: ["1111::/16", "2222::/16"]
 generate_bid_id: true
-enable_legacy_auction: false
+enable_legacy_auction: true
```
guscarreon, line 357: For the `TestFullConfig` test, can we leave its value as `false` just for the sake of it being different to the default value of `true`

config/config_test.go
```diff
@@ -579,7 +579,7 @@ func TestFullConfig(t *testing.T) {
 	cmpStrings(t, "request_validation.ipv6_private_networks", cfg.RequestValidation.IPv6PrivateNetworks[1], "2222::/16")
 	cmpBools(t, "generate_bid_id", cfg.GenerateBidID, true)
 	cmpStrings(t, "debug.override_token", cfg.Debug.OverrideToken, "")
-	cmpBools(t, "enable_legacy_auction", cfg.EnableLegacyAuction, false)
+	cmpBools(t, "enable_legacy_auction", cfg.EnableLegacyAuction, true)
```
guscarreon, line 582: For the `TestFullConfig` test, can we leave its value as `false` just for the sake of it being different to the default value of `true`?

guscarreon, line 582: For the `TestFullConfig` test, can we leave its value as `false` just for the sake of it being different to the default value of `true`?

PR #2033, opened 2021-10-05
Remove /auction Endpoint

SyntaxNode:
Implements phase 3 of https://github.com/prebid/prebid-server/issues/2011. Do not merge until after https://github.com/prebid/prebid-server/pull/2032.


stored_requests/config/config.go
```diff
@@ -107,7 +107,7 @@ func CreateStoredRequests(cfg *config.StoredRequests, metricsEngine metrics.Metr
 //
 // As a side-effect, it will add some endpoints to the router if the config calls for it.
 // In the future we should look for ways to simplify this so that it's not doing two things.
-func NewStoredRequests(cfg *config.Configuration, metricsEngine metrics.MetricsEngine, client *http.Client, router *httprouter.Router) (db *sql.DB, shutdown func(), fetcher stored_requests.Fetcher, ampFetcher stored_requests.Fetcher, accountsFetcher stored_requests.AccountFetcher, categoriesFetcher stored_requests.CategoryFetcher, videoFetcher stored_requests.Fetcher) {
+func NewStoredRequests(cfg *config.Configuration, metricsEngine metrics.MetricsEngine, client *http.Client, router *httprouter.Router) (shutdown func(), fetcher stored_requests.Fetcher, ampFetcher stored_requests.Fetcher, accountsFetcher stored_requests.AccountFetcher, categoriesFetcher stored_requests.CategoryFetcher, videoFetcher stored_requests.Fetcher) {
```
guscarreon, line 110: Can we remove DB connection entry that is no longer returned by this function from the comment above?
```
 96   // NewStoredRequests returns five things:
 97   //
 98 - // 1. A DB connection, if one was created. This may be nil.
 99   // 2. A function which should be called on shutdown for graceful cleanups.
100   // 3. A Fetcher which can be used to get Stored Requests for /openrtb2/auction
101   // 4. A Fetcher which can be used to get Stored Requests for /openrtb2/amp
102   // 5. A Fetcher which can be used to get Category Mapping data
103   // 6. A Fetcher which can be used to get Stored Requests for /openrtb2/video
104   //
105   // If any errors occur, the program will exit with an error message.
106   // It probably means you have a bad config or networking issue.
107   //
108   // As a side-effect, it will add some endpoints to the router if the config calls for it.
109   // In the future we should look for ways to simplify this so that it's not doing two things.
110   func NewStoredRequests(cfg *config.Configuration, metricsEngine metrics.MetricsEngine, client *http.Client, router *httprouter.Router) (shutdown func(), fetcher stored_requests.Fetcher, ampFetcher store
```

guscarreon (commented):
Left a couple of nitpicks. Also, the `pbs/usersync.go` file that contains in the `"/optout"` endpoint functionality was not removed, can it be relocated to, say `endpoints/output.go` in a future PR? Or maybe even renamed to `captcha/usersync.go` or something.

router/router.go
```diff
@@ -49,16 +31,15 @@ import (
 	storedRequestsConf "github.com/prebid/prebid-server/stored_requests/config"
 	"github.com/prebid/prebid-server/usersync"
 	"github.com/prebid/prebid-server/util/sliceutil"
+	"github.com/prebid/prebid-server/util/uuidutil"
+	"github.com/prebid/prebid-server/version"
 
 	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
 	_ "github.com/lib/pq"
```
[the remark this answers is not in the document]
SyntaxNode: I'm not sure. This is initializing the postgres package which we still use for stored requests. I feel more comfortable experimenting with this in a future PR.

stored_requests/config/config.go
SyntaxNode: Yes. I'll update the comment. Good catch.

guscarreon (approved):
LGTM

adapters/pubmatic/pubmatic_test.go
```diff
@@ -32,699 +18,3 @@ func TestJsonSamples(t *testing.T) {
...
-}
-
-func TestGetBidTypeForUnsupportedCode(t *testing.T) {
-	pubmaticExt := new(pubmaticBidExt)
-	pubmaticExt.BidType = new(int)
-	*pubmaticExt.BidType = 99
-	actualBidTypeValue := getBidType(pubmaticExt)
-	if actualBidTypeValue != openrtb_ext.BidTypeBanner {
-		t.Errorf("Expected Bid Type value was: %v, actual value is: %v", openrtb_ext.BidTypeBanner, actualBidTypeValue)
-	}
```
pm-viral-vala, line 729: Can you please add these required test cases?

VeronikaSolovei9 (approved):

SyntaxNode:
This was accidentally merged earlier than intended. Please see: https://github.com/prebid/prebid-server/pull/2049

PR #2035, opened 2021-10-08
add algorix placementId Support

Bugxyb:
Add Algorix PlacementId Support
* add placementId and appId
* replace imp.tagid to placementId


adapters/algorix/algorix.go
```diff
@@ -109,9 +109,6 @@ func preProcess(request *openrtb2.BidRequest, ext *openrtb_ext.ExtImpAlgorix) {
 				request.Imp[i].Banner = &banner
 			}
 		}
-		if len(ext.PlacementId) > 0 {
-			request.Imp[i].TagID = ext.PlacementId
-		}
 	}
```
AlexBVolcy: Hello! Just for my understanding, could you explain why you decided to reverse the change that replaces the `imp.TagID` with the `PlacementID`? I noticed that this was one of your goals in this issues description! 

> Add Algorix PlacementId Support
    add placementId and appId
    replace imp.tagid to placementId

static/bidder-params/algorix.json
```diff
@@ -13,6 +13,14 @@
       "type": "string",
       "description": "Your Token",
       "minLength": 1
+    },
+    "placementId": {
+      "type": "string",
+      "description": "An ID which identifies this placement of the impression"
```
mansinahar, line 19: Do we need a min/max length validation on these fields like we have for token and sid? Without the `minLength` the JSON schema will allow empty values for these fields as well

adapters/algorix/params_test.go
```diff
@@ -34,7 +34,7 @@ func TestInvalidParams(t *testing.T) {
 }
 
 var validParams = []string{
-	`{"sid": "11233", "token": "sin"}`,
+	`{"sid": "11233", "token": "sin", "placementId": "123", "appId": "abc"}`,
```
mansinahar, line 37: Please add the following test cases for `invalidParams`:
```
`{"sid": "11233", "token": "sin", "placementId": 123, "appId": "abc"}`,
`{"sid": "11233", "token": "sin", "placementId": "123", "appId": 123}`,
`{"sid": "11233", "token": "sin", "placementId": 123, "appId": 456}`,
```

static/bidder-params/algorix.json
Bugxyb: Yes, we don't need min/max length validation, placementId and appId is Optional params. it can be empty value.

PR #2036, opened 2021-10-08
GDPR: deprecate per-purpose enabled config & add enforce purpose

bsardo:
This PR does the following:
- deprecates the per-purpose bool enabled config flags (`gdpr.tcf2.purposeX.enabled`)
- introduces new per-purpose string enforce-purpose config flags (`gdpr.tcf2.purposeX.enforce_purpose`) as the new way to turn on/off a purpose. By moving from bool to string the flag can act as a tri-state in the future.
- right now there are only two valid values for `gdpr.tcf2.purposeX.enforce_purpose`: "no" and "full". Full means full enforcement IAW the prebid.org GDPR spec. Eventually basic enforcement will be available, in which case "basic" will become the third state.

The enabled config flags are still supported for now with PBS logging a warning if they are set. The enabled config flags will be removed in the future.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1081,6 +1081,14 @@ func migrateConfigTCF2PurposeEnabledFlags(v *viper.Viper) {
 				}
 			}
 		}
+
+		if v.IsSet(newField) {
+			if v.GetString(newField) == TCF2FullEnforcement {
+				v.Set(oldField, "true")
+			} else {
+				v.Set(oldField, "false")
+			}
+		}
 	}
 }
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -777,6 +777,9 @@ func TestMigrateConfigPurposeOneTreatment(t *testing.T) {
 }
 
 func TestMigrateConfigTCF2PurposeEnabledFlags(t *testing.T) {
+	trueStr := "true"
+	falseStr := "false"
+
 	tests := []struct {
 		description                 string
 		config                      []byte
@@ -790,6 +793,16 @@ func TestMigrateConfigTCF2PurposeEnabledFlags(t *testing.T) {
 		wantPurpose8EnforcePurpose  string
 		wantPurpose9EnforcePurpose  string
 		wantPurpose10EnforcePurpose string
+		wantPurpose1Enabled         string
+		wantPurpose2Enabled         string
+		wantPurpose3Enabled         string
+		wantPurpose4Enabled         string
+		wantPurpose5Enabled         string
+		wantPurpose6Enabled         string
+		wantPurpose7Enabled         string
+		wantPurpose8Enabled         string
+		wantPurpose9Enabled         string
+		wantPurpose10Enabled        string
 	}{
 		{
 			description: "New config and old config flags not set",
@@ -831,6 +844,16 @@ func TestMigrateConfigTCF2PurposeEnabledFlags(t *testing.T) {
 			wantPurpose8EnforcePurpose:  TCF2FullEnforcement,
 			wantPurpose9EnforcePurpose:  TCF2NoEnforcement,
 			wantPurpose10EnforcePurpose: TCF2FullEnforcement,
+			wantPurpose1Enabled:         falseStr,
+			wantPurpose2Enabled:         trueStr,
+			wantPurpose3Enabled:         falseStr,
+			wantPurpose4Enabled:         trueStr,
+			wantPurpose5Enabled:         falseStr,
+			wantPurpose6Enabled:         trueStr,
+			wantPurpose7Enabled:         falseStr,
+			wantPurpose8Enabled:         trueStr,
+			wantPurpose9Enabled:         falseStr,
+			wantPurpose10Enabled:        trueStr,
 		},
 		{
 			description: "New config flags set, old config flags not set - use new flags",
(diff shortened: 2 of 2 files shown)


config/config.go
```diff
@@ -229,6 +229,32 @@ func (cfg *GDPR) validate(v *viper.Viper, errs []error) []error {
...
+		cfg.TCF2.Purpose8,
+		cfg.TCF2.Purpose9,
+		cfg.TCF2.Purpose10,
+	}
+
+	for i := 0; i < len(purposeConfigs); i++ {
+		enforcePurposeValue := purposeConfigs[i].EnforcePurpose
+		enforcePurposeField := fmt.Sprintf("gdpr.tcf2.purpose%d.enforce_purpose", (i + 1))
+
+		if enforcePurposeValue != "no" && enforcePurposeValue != "full" {
```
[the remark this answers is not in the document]
SyntaxNode: > Should I just define those constants in the config package and reference them from the gdpr package to get rid of all of the string occurrences here?

Sure.

config/config.go
```diff
@@ -229,6 +229,32 @@ func (cfg *GDPR) validate(v *viper.Viper, errs []error) []error {
...
+			errs = append(errs, fmt.Errorf("%s must be no or full. Got %s", enforcePurposeField, enforcePurposeValue))
```
SyntaxNode, line 255: Nitpick: I recommend using single or double quotes around "no" and "full" to amplify those explicit values.

config/config.go
```diff
@@ -988,7 +1005,28 @@ func SetupViper(v *viper.Viper, filename string) {
...
+	v.SetDefault("gdpr.tcf2.purpose1.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose2.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose3.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose4.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose5.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose6.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose7.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose8.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose9.enforce_purpose", "full")
+	v.SetDefault("gdpr.tcf2.purpose10.enforce_purpose", "full")
```
SyntaxNode, line 1029: Why are these values not grouped with the other tcf2 values?

config/config.go
```diff
@@ -1018,6 +1056,27 @@ func migrateConfigPurposeOneTreatment(v *viper.Viper) {
...
+		oldField := fmt.Sprintf("gdpr.tcf2.purpose%d.enabled", i)
+		newField := fmt.Sprintf("gdpr.tcf2.purpose%d.enforce_purpose", i)
+
+		if v.IsSet(oldField) {
+			oldConfig := v.GetBool(oldField)
+			if v.IsSet(newField) {
+				glog.Warningf("using %s and ignoring deprecated %s", newField, oldField)
+			} else {
+				glog.Warningf("%s is deprecated and should be changed to %s", oldField, newField)
+				if oldConfig == true {
```
SyntaxNode, line 1070: Nitpick: There is no need to add `== true` to a conditional. This will generate a linter warning.

config/config_test.go
```diff
@@ -743,6 +776,184 @@ func TestMigrateConfigPurposeOneTreatment(t *testing.T) {
...
+            enabled: false
+            enforce_purpose: "full"
+          purpose9:
+            enabled: false
+            enforce_purpose: "full"
+          purpose10:
+            enabled: false
+            enforce_purpose: "full"
+    `)
+	tests := []struct {
```
SyntaxNode, line 862: I'm not sure if the tests will be easier to read if the yaml is added to each test case. These variables are only used once each, so this looks like an organizational choice.

config/config_test.go
```diff
@@ -817,6 +1028,27 @@ func TestMissingGDPRDefaultValue(t *testing.T) {
...
+	cfg.GDPR.TCF2.Purpose8.EnforcePurpose = "full"
+	cfg.GDPR.TCF2.Purpose9.EnforcePurpose = "full"
+	cfg.GDPR.TCF2.Purpose10.EnforcePurpose = "invalid3"
+
+	errs := cfg.validate(v)
+	assert.Len(t, errs, 4)
+	assert.EqualError(t, errs[0], "gdpr.tcf2.purpose1.enforce_purpose must be no or full. Got ")
+	assert.EqualError(t, errs[1], "gdpr.tcf2.purpose5.enforce_purpose must be no or full. Got invalid1")
+	assert.EqualError(t, errs[2], "gdpr.tcf2.purpose6.enforce_purpose must be no or full. Got invalid2")
+	assert.EqualError(t, errs[3], "gdpr.tcf2.purpose10.enforce_purpose must be no or full. Got invalid3")
```
SyntaxNode, line 1049: I recommend using `assert.ElementsMatch` when comparing slices instead of this approach, which has the added benefit of being order insensitive.

SyntaxNode (commented):
Test coverage looks good. Implementation looks straight forward. I left a few nitpicks.

config/config.go
SyntaxNode: Gotcha. I did not understand the significance of the placement. Could you please add a comment explaining this situation?

config/config_test.go
```diff
@@ -1042,11 +1040,14 @@ func TestInvalidEnforcePurpose(t *testing.T) {
...
 
 	errs := cfg.validate(v)
-	assert.Len(t, errs, 4)
-	assert.EqualError(t, errs[0], "gdpr.tcf2.purpose1.enforce_purpose must be no or full. Got ")
-	assert.EqualError(t, errs[1], "gdpr.tcf2.purpose5.enforce_purpose must be no or full. Got invalid1")
-	assert.EqualError(t, errs[2], "gdpr.tcf2.purpose6.enforce_purpose must be no or full. Got invalid2")
-	assert.EqualError(t, errs[3], "gdpr.tcf2.purpose10.enforce_purpose must be no or full. Got invalid3")
+
+	expectedErrs := []error{
+		fmt.Errorf("gdpr.tcf2.purpose1.enforce_purpose must be \"no\" or \"full\". Got "),
```
SyntaxNode, line 1045: Nitpick: Should just be `errors.New` since there is no formatting being performed.

SyntaxNode (approved):

mansinahar (dismissed):

mansinahar (approved):

PR #2037, opened 2021-10-09
Added channel support

AlexBVolcy:
First PR in a series of PRs that will address the Integration and Channel issue https://github.com/prebid/prebid-server/issues/1428

This PR handles the **Channel** portion of the issue, which allows us to formalize and capture where a request came from, here are some bullet points for how this is accomplished: 
- Requests can have a channel object at this path: `ext.prebid.channel`
- Channel object will have name and version, where the name defines where the request came from (app, amp, video, etc.)
- Updated code will read this channel information into a channel object of type `openrtb_ext.ExtRequestPrebidChannel` utilizing the request wrapper
- Channel object is then checked: If the channel object exists, but the name is blank, we error out because the name is required. If the channel object exists and has a name, then we don't need to do anything. If the channel object doesn't exist, we update the request ext channel info based off of some defaults. If the request has an app object, we set the channel name in the request to be `app`. If it came from an Amp URL Path, we set the channel name to be `amp`


endpoints/openrtb2/amp_auction.go
```diff
@@ -348,6 +348,14 @@ func (deps *endpointDeps) loadRequestJSONForAmp(httpRequest *http.Request) (req
 		return
 	}
 
+	// Fetch Channel Information and Update Channel Object in config with this information
+	deps.cfg.Channel = openrtb_ext.ExtRequestPrebidChannel{}
+	deps.cfg.Channel, err = getChannelObjectAmpFromRequest(requestJSON)
```
SyntaxNode, line 353: Line 352 is inconsequential due to line 353.

endpoints/openrtb2/amp_auction.go
```diff
@@ -575,3 +583,27 @@ func setEffectiveAmpPubID(req *openrtb2.BidRequest, account string) {
 		}
 	}
 }
+
+// Update config channel object with information found from request. If no info found, we set channel name to "amp"
+func getChannelObjectAmpFromRequest(request []byte) (openrtb_ext.ExtRequestPrebidChannel, error) {
+	channelInfo, dataType, _, err := jsonparser.Get(request, "ext", "prebid", "channel")
```
SyntaxNode, line 589: Each call to `jsonparser.Get` has to reparse the entire request. It's best to avoid this approach whenever possible. I think it would be more appropriate to use the `RequestWrapper` here, since it already has support for accessing the prebid request extension.

endpoints/openrtb2/auction.go
```diff
@@ -276,6 +276,30 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
 		return
 	}
 
+	// Fetch Channel Information and Update Channel Object in config with this information
+	deps.cfg.Channel = openrtb_ext.ExtRequestPrebidChannel{}
```
SyntaxNode, line 280: This looks like a typo. The configuration object in `deps` must be read-only.

endpoints/openrtb2/auction.go
```diff
@@ -276,6 +276,30 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
...
+	channelInfo, err := getChannelInfoFromRequest(requestJson)
```
SyntaxNode, line 281: You should use the `RequestWrapper` here instead. It's already in scope in this method.

endpoints/openrtb2/auction_test.go
```diff
@@ -3627,9 +3891,6 @@ var testBidRequests = []string{
 		],
 		"ext": {
 			"prebid": {
-				"cache": {
-					"markup": 1
-				},
```
SyntaxNode, line 3632: Curious, why is this removed?

endpoints/openrtb2/amp_auction.go
AlexBVolcy: Request wrapper is now utilized in latest commit!

endpoints/openrtb2/auction_test.go
```diff
@@ -3565,6 +3621,204 @@ var testStoredImps = []string{
 	``,
 }
 
+var testBidRequestForChannelTest = []string{
+	`{
```
VeronikaSolovei9, line 3625: It looks like you only use elements with indexed 0, 1 and 3. If other data is not needed, can you please delete it? 
It's superhard to read this kind of inputs.

AlexBVolcy: Yes will do!

VeronikaSolovei9 (commented):
Alex, should you have a unite vtest where you pass `isAmp = true`
```
errL = deps.validateRequest(&openrtb_ext.RequestWrapper{BidRequest: bidReq}, true)
```

AlexBVolcy:
@VeronikaSolovei9 based off my current unit tests for channel support, I cover the case where `isAmp = true` with having tests that make a call from the amp_auction endpoint. And I have a test case for each scenario related to adding channel support, and my code coverage shows this. So I'm unsure if there's another test case needed.

VeronikaSolovei9 (dismissed):
Ok, makes sense. LGTM.

endpoints/openrtb2/auction.go
```diff
@@ -1230,6 +1234,28 @@ func validateDevice(device *openrtb2.Device) error {
 	return nil
 }
 
+func validateOrFillChannel(requestPrebid *openrtb_ext.ExtRequestPrebid, requestExt *openrtb_ext.RequestExt, reqWrapper *openrtb_ext.RequestWrapper, isAmp bool) error {
+	if requestPrebid != nil {
+		if requestPrebid.Channel != nil {
+			if requestPrebid.Channel.Name == "" {
+				return errors.New("ext.prebid.channel.name can't be empty")
+			}
+		} else {
```
mansinahar, line 1243: This looks like it will add the channel object only if `ext.Prebid.Channel` is nil but what about when `ext.Prebid` itself is nil? We'd still want to add the channel info in the request for app and amp, no?

hhhjort: We don't need the `requestPrebid` test, as the caller already verifies that it is not nil before calling this function. On making this test more aggressive, we also want to make sure we don't invalidate all non-video requests do to a missing channel name. Where is the logic to determine if a channel is needed if we are not basing it on the existence of a `Channel` object?

AlexBVolcy: Hans, for my own understanding, are you saying I don't need to have the check of if ```requestPrebid != nil```, because that's already done prior to this function call? 

And I'm confused about your second point about making the test more aggressive, are you referring to the channel support tests in ```auction_test.go```, or referring to this functions conditionals?

hhhjort: Yes, it is done prior. And by making it more aggressive, I mean this validation code. Video calls may need a channel object, but I am pretty sure non-video requests don't require a channel entry. We don't want to fail valid non-video requests for the lack of a video channel, so we will want to confirm that a simple banner request will still pass as valid. With the current logic, the assumption was that if the `Channel` object exists, then a channel name is required. What will be the new "if" part that leads us to saying a channel name is required?

mansinahar: About `ext.Prebid`, my point wasn't about testing its existence to avoid a nil panic. It was about adding the channel object in the request when it doesn't exist. With the current logic we're adding the channel object only if `ext.Prebid.Channel` is not present but we should be adding the channel object also if the request doesn't have `ext.Prebid` to start with. I hope that makes it a little clearer.

About your second point Hans, the channel object in not required for any type of request but if one exists, we expect the channel name to be present. We're not invalidating requests based on the presence of the channel object but rather on the presence of the name in the channel object _if_ the channel object is present. I am pasting the original proposal below. Please see the 1st and 4th point. Hopefully that provides some clarification.

> Proposal:

> Channel is an object that contains a name (required) and a version (optional)
Update the PBJS pbsBidAdapter to pass ext.prebid.channel: {name: "pbjs", version: "4.1"} unless overridden by s2sConfig.extPrebid
PBS gets the channel name out of the request if it's there
If the channel isn't present:
4a. set channel name to 'amp' if the request came in on /openrtb2/amp
4b. set channel name to 'app' if the request contains an 'app' object
4c. otherwise leave channel blank. Analytics adapters can default if they'd like.

I hope that helps!

endpoints/openrtb2/auction.go
```diff
@@ -1334,6 +1339,36 @@ func validateDevice(device *openrtb2.Device) error {
 	return nil
 }
 
+func validateOrFillChannel(requestPrebid *openrtb_ext.ExtRequestPrebid, requestExt *openrtb_ext.RequestExt, reqWrapper *openrtb_ext.RequestWrapper, isAmp bool) error {
```
mansinahar, line 1342: Do we need to pass in `requestPrebid` and `requestExt` to this function? We can just access them via the request wrapper's `GetRequestExt()` and `GetPrebid()` methods in this function. This will reduce the number of parameters in this function

endpoints/openrtb2/auction_test.go
```diff
@@ -1145,6 +1145,135 @@ func TestStoredRequests(t *testing.T) {
...
+			expectedChannelObject: nil,
+		},
+	}
+
+	for _, test := range testCases {
+		req := &openrtb_ext.RequestWrapper{}
+		req.BidRequest = &openrtb2.BidRequest{}
+
+		err := json.Unmarshal([]byte(test.givenBidRequestData), req.BidRequest)
+		assert.Empty(t, err, test.description)
```
mansinahar, line 1257: The `validateOrFillChannel` expects the request being passed in as an object. Is there a reason why we're specifying the requests in JSON and then unmarshalling them before passing them to the request? Also for unit tests, I'd suggest keep the requests as small as possible by specifying only fields that are needed to test the function under consideration. For example, you can modify these test cases to look something like:

```
{
	description: "App request; No request.ext present; expect app channel info to be generated",
	givenReqWrapper: &openrtb_ext.RequestWrapper{
		BidRequest: &openrtb2.BidRequest{App: &openrtb2.App{}},
	},
	givenIsAmp:         false,
	expectedError:      nil,
	expectedRequestExt: []byte(`{"prebid":{"channel": {"name": "app", "version": ""}}}`),
},
```

endpoints/openrtb2/auction.go
```diff
@@ -1334,6 +1339,36 @@ func validateDevice(device *openrtb2.Device) error {
...
+	if requestPrebid != nil {
+		if requestPrebid.Channel != nil {
```
mansinahar, line 1344: I think the conditionals here can be simplified to something like:
```
if requestPrebid == nil || requestExt.GetPrebid().Channel == nil {
	fillChannel(requestPrebid, requestExt, reqWrapper, isAmp)
} else if requestPrebid.Channel.Name == "" {
	return errors.New("ext.prebid.channel.name can't be empty")
}
```

And then you can just create and empty `requestPrebid` in the `fillChannel` function if it is `nil`

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1360,3 +1359,131 @@ func AmpObjectTestSetup(t *testing.T, inTagId string, inStoredRequest json.RawMe
...
+			expectedExtObject:    json.RawMessage(testExtObjectInfo[1]),
+		},
+		{
+			description:          "No channel object in request, we expect amp name",
+			givenInStoredRequest: json.RawMessage(testStoredRequestsForChannelTest[0]),
+			expectedExtObject:    json.RawMessage(testExtObjectInfo[0]),
+		},
+	}
+
+	request := httptest.NewRequest("GET", "/openrtb2/auction/amp?tag_id=test", nil)
```
mansinahar, line 1381: Can we move this test to the existing JSON test framework?

endpoints/openrtb2/auction_test.go
```diff
@@ -1145,6 +1145,135 @@ func TestStoredRequests(t *testing.T) {
...
+		expectedChannelObject *openrtb_ext.ExtRequestPrebidChannel
+	}{
+		{
+			description:           "No channel object in app request, so we expect channel name to be set to app",
+			givenBidRequestData:   testBidRequestForChannelTest[0],
+			expectedChannelObject: &openrtb_ext.ExtRequestPrebidChannel{Name: "app", Version: ""},
+		},
+		{
+			description:           "Channel object in request with populated name/version, we expect same name/version in object that's created",
+			givenBidRequestData:   testBidRequestForChannelTest[1],
```
mansinahar, line 1180: Can we move this test to the existing JSON test framework?

endpoints/openrtb2/auction.go
AlexBVolcy: Thanks this is great!

endpoints/openrtb2/auction.go
```diff
@@ -1334,6 +1339,36 @@ func validateDevice(device *openrtb2.Device) error {
...
+		requestPrebid = &openrtb_ext.ExtRequestPrebid{}
+		fillChannel(requestPrebid, requestExt, reqWrapper, isAmp)
+	}
+	return nil
+}
+
+func fillChannel(requestPrebid *openrtb_ext.ExtRequestPrebid, requestExt *openrtb_ext.RequestExt, reqWrapper *openrtb_ext.RequestWrapper, isAmp bool) error {
+	if isAmp {
+		var err error
+		reqWrapper.Ext, err = jsonpatch.MergePatch(reqWrapper.Ext, []byte(`{"prebid":{"channel": {"name": "amp", "version": ""}}}`))
```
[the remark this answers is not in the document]
AlexBVolcy: This will be updated to use the same method that is used for app requests!

endpoints/openrtb2/auction_test.go
AlexBVolcy: That makes a lot of sense, I'll use this approach instead!

endpoints/openrtb2/auction.go
AlexBVolcy: This will be updated to reduce the number of params!

endpoints/openrtb2/auction.go
```diff
@@ -1334,6 +1339,44 @@ func validateDevice(device *openrtb2.Device) error {
...
 }
 
+func validateOrFillChannel(reqWrapper *openrtb_ext.RequestWrapper, isAmp bool) error {
+	requestExt, err := reqWrapper.GetRequestExt()
+	if err != nil {
+		return err
+	}
+	requestPrebid := requestExt.GetPrebid()
+
+	if requestPrebid == nil || requestExt.GetPrebid().Channel == nil {
```
mansinahar, line 1349: **Nitpick:** You can replace `requestExt.GetPrebid()` in the second part of this if condition to just `requestPrebid` as you've already saved that in the `requestPrebid` variable.

endpoints/openrtb2/auction.go
```diff
@@ -1334,6 +1339,44 @@ func validateDevice(device *openrtb2.Device) error {
...
+}
+
+func fillChannel(reqWrapper *openrtb_ext.RequestWrapper, isAmp bool) error {
+	requestExt, err := reqWrapper.GetRequestExt()
+	if err != nil {
+		return err
+	}
+	requestPrebid := requestExt.GetPrebid()
+
+	if requestPrebid == nil {
```
mansinahar, line 1364: What are your thoughts about rewriting this as:
```
var channelName string
if isAmp {
	channelName = ampChannelName
}
if reqWrapper.App != nil {
	channelName = appChannelName
}

if channelName != "" {
	if requestPrebid == nil {
		requestPrebid = &openrtb_ext.ExtRequestPrebid{}
	}
	requestPrebid.Channel = &openrtb_ext.ExtRequestPrebidChannel{Name: channelName}
	requestExt.SetPrebid(requestPrebid)
	reqWrapper.RebuildRequest()
}
return nil
```

endpoints/openrtb2/auction_test.go
```diff
@@ -1145,6 +1146,75 @@ func TestStoredRequests(t *testing.T) {
...
+		givenIsAmp            bool
+		givenRequestWrapper   *openrtb_ext.RequestWrapper
+		expectedError         error
+		expectedChannelObject *openrtb_ext.ExtRequestPrebidChannel
+	}{
+		{
+			description: "No request.ext info in app request, so we expect channel name to be set to app",
+			givenRequestWrapper: &openrtb_ext.RequestWrapper{
+				BidRequest: &openrtb2.BidRequest{App: &openrtb2.App{}},
+			},
```
mansinahar, line 1161: Please also add `app` and `amp` test cases where `request.ext` exists but `request.ext.prebid` doesn't exist and where `request.ext.prebid` exists but the channel object doesn't exist

endpoints/openrtb2/sample-requests/valid-whole/exemplary/all-ext.json
```diff
@@ -65,6 +65,10 @@
         "cache": {
           "bids": {}
         },
+        "channel" : {
```
mansinahar, line 68: Please correct the indentation here

endpoints/openrtb2/auction.go
AlexBVolcy: I like this rewrite! I'll implement it :)

endpoints/openrtb2/sample-requests/valid-whole/exemplary/all-ext.json
AlexBVolcy: Interesting, in my VS Code, the indentation is correct and not really spaced out as it is here. If you pull down this code, is the indentation also incorrect locally?

endpoints/openrtb2/auction.go
hhhjort: Just to confirm, we don't expect a request to come with a channel set in the AMP or APP cases? This does overwrite any channel that may be found in the original request.

mansinahar (approved):

VeronikaSolovei9 (approved):
LGTM

PR #2038, opened 2021-10-11
Fix flaky cookie trim test depending on time.Now() and the uuid order…

mansinahar:
… in the cookies

### DIFF
--- a/usersync/cookie_test.go
+++ b/usersync/cookie_test.go
@@ -338,38 +338,15 @@ func TestGetUIDsWithNilCookie(t *testing.T) {
 }
 
 func TestTrimCookiesClosestExpirationDates(t *testing.T) {
-	var expirationTime = time.Date(2021, time.January, 3, 15, 25, 0, 0, time.UTC)
-
 	cookieToSend := &Cookie{
 		uids: map[string]uidWithExpiry{
-			"k1": {
-				UID:     "12345678901234567890123456789012345678901234567890",
-				Expires: expirationTime,
-			},
-			"k2": {
-				UID:     "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwx",
-				Expires: expirationTime,
-			},
-			"k3": {
-				UID:     "ABCDEFGHIJKLMNOPQRSTUVWXYZABCDEFGHIJKLMNOPQRSTUVWX",
-				Expires: expirationTime,
-			},
-			"k4": {
-				UID:     "12345678901234567890123456789612345678901234567890",
-				Expires: expirationTime,
-			},
-			"k5": {
-				UID:     "aAbBcCdDeEfFgGhHiIjJkKlLmMnNoOpPqQrRsStTuUvVwWxXyYz",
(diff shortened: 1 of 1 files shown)


usersync/cookie_test.go
```diff
@@ -338,31 +338,54 @@ func TestGetUIDsWithNilCookie(t *testing.T) {
 }
 
 func TestTrimCookiesClosestExpirationDates(t *testing.T) {
+	var expirationTime = time.Date(2021, time.January, 3, 15, 25, 0, 0, time.UTC)
```
mansinahar, line 341: Using a random constant time with a constant time zone so that the length of the JSON encoded expiry time string is constant and the output of the cookie trim function is predictable.

SyntaxNode: A static time is a good choice. Why this date though? How about a simpler `time.UnixMilli(1)`?

usersync/cookie_test.go
```diff
@@ -338,31 +338,54 @@ func TestGetUIDsWithNilCookie(t *testing.T) {
...
+
 	cookieToSend := &Cookie{
 		uids: map[string]uidWithExpiry{
-			"k1": newTempId("12345678901234567890123456789012345678901234567890", 7),
-			"k2": newTempId("abcdefghijklmnopqrstuvwxyz", 6),
```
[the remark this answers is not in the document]
SyntaxNode: I do like the original approach of checking for specific keys. The order was not dependent upon the iteration but on the time out. In this PR, the expiration time is always the same now. Perhaps there was a problem with k2 and k3 sharing the same expiration date and causing indeterminate results?

usersync/cookie_test.go
```diff
@@ -372,7 +395,7 @@ func TestTrimCookiesClosestExpirationDates(t *testing.T) {
 			actualKeys = append(actualKeys, key)
 		}
 
-		assert.ElementsMatch(t, testCases[i].expKeys, actualKeys, "[Test %d]", i+1)
+		assert.Len(t, actualKeys, testCases[i].expNumOfKeys, "[Test %d]", i+1)
```
guscarreon, line 398: What if the lenght of `processedCookie.uids` gets directly asserted? We could reduce the code a bit further I think:
```
390       for i := range testCases {
391           processedCookie := writeThenRead(cookieToSend, testCases[i].maxCookieSize)
392  
393 -         actualKeys := make([]string, 0, 7)
394 -         for key := range processedCookie.uids {
395 -             actualKeys = append(actualKeys, key)
396 -         }
397 -  
398 -         assert.Len(t, actualKeys, testCases[i].expNumOfKeys, "[Test %d]", i+1)
    +         assert.Len(t, processedCookie.uids, testCases[i].expNumOfKeys, "[Test %d]", i+1)
399       }
400   }
usersync/cookie_test.go
```

mansinahar: Have reverted this change based on Scott's comment above

SyntaxNode (approved):

PR #2039, opened 2021-10-11
Beachfront bidder -  include primary category in response

muncha:
Beachfront's backend VAST server should always supply an array with at least one category. This is just setting the PrimaryCategory property of the response ext based on that.


muncha:
Hold up. There are some cases where the cat array can be empty. Updating now.

PR #2040, opened 2021-10-12
VIS.X: mediatype handling fix for complex requests

mk0x9:
There is an issue when a request with banner and video objects is received in case if the adapter will respond with video bid it will have its mediatype set as banner, instead of the video. The following change will fix it.

### DIFF
--- a/adapters/visx/visx.go
+++ b/adapters/visx/visx.go
@@ -16,16 +16,29 @@ type VisxAdapter struct {
 	endpoint string
 }
 
+type visxBidExtPrebidMeta struct {
+	MediaType openrtb_ext.BidType `json:"mediaType"`
+}
+
+type visxBidExtPrebid struct {
+	Meta visxBidExtPrebidMeta `json:"meta"`
+}
+
+type visxBidExt struct {
+	Prebid visxBidExtPrebid `json:"prebid"`
+}
+
 type visxBid struct {
-	ImpID   string   `json:"impid"`
-	Price   float64  `json:"price"`
-	UID     int      `json:"auid"`
-	CrID    string   `json:"crid,omitempty"`
-	AdM     string   `json:"adm,omitempty"`
-	ADomain []string `json:"adomain,omitempty"`
-	DealID  string   `json:"dealid,omitempty"`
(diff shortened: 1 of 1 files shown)


mk0x9:
@AlexBVolcy tests have been added

AlexBVolcy (approved):
LGTM

adapters/visx/visx.go
```diff
@@ -117,9 +131,19 @@ func (a *VisxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReq
...
+			var ext visxBidExt
+			if err := json.Unmarshal(bid.Ext, &ext); err == nil {
+				if ext.Prebid.Meta.MediaType == openrtb_ext.BidTypeBanner {
+					return openrtb_ext.BidTypeBanner, nil
+				}
+				if ext.Prebid.Meta.MediaType == openrtb_ext.BidTypeVideo {
+					return openrtb_ext.BidTypeVideo, nil
+				}
+			}
+
```
bsardo, line 146: From your PR description, it sounds like you want to favor setting the bid type to video if the imp has both banner and video objects. With this change I see your server will now be explicit about the bid type in its response, however, I see you're falling back to the implicit logic that checks if the `imp.Banner` is not nil in case the bid type was not specified, or is set but doesn't equal `BidTypeBanner` or `BidTypeVideo` for some reason. Do you want to prioritize setting the bid type to video in this situation by moving the block `if imp.Video != nil {` above the `if imp.Banner != nil {` block?

rrelict: Hi @bsardo, our developer is OOO, so I'll try to answer your question. Now VIS.X always returns 'banner' or 'video' in `Meta.MediaType` in bid responses to our Prebid Server adapter. Thus the code you mentioned above is going to return one of `BidTypeBanner` or `BidTypeVideo`. The code below the mentioned one is unlikely to be executed.

> Do you want to prioritize setting the bid type to video in this situation by moving the block if imp.Video != nil { above the if imp.Banner != nil { block?

The code is unlikely to be executed, I'd like to keep the current logic for now. When the developer comes back, we'll review this code and may make it clearer.

Issue #2006, opened 2021-09-21
Twin ad unit codes support on server side


muuki88:
Hi,

There are bidders that require the [twin ad unit codes workaround](https://docs.prebid.org/dev-docs/adunit-reference.html#twin-adunit-codes) in order to work properly. This can be due to

- only certain sizes are supported
- multi format support is limited or non existent

However I wasn't able to apply this workaround with prebid.js + prebid server. I also think that the twin ad unit codes configuration is more of a workaround than an actual solution.

## Background

We had an issue with the _Adform Bidder_

It's a standard OpenRTB bid adapter, but supports only one format (banner, video or native) at a time. Sending multiple formats lead to no bids at all. We implemented a hotfix by prioritzing formats (banner over native over video), but I think this is not the best way to do it.

https://github.com/prebid/prebid-server-java/blob/b23bac749010ab26dc52ee218c3408e869173584/src/main/java/org/prebid/server/bidder/adf/AdfBidder.java#L24-L27

we nulled the other formats if existent.

## Options for multi format ad units

Of course every bidder could fix this on their own. I think it would be nice to have a configuration or utility function that can split a multi format bid into separate `Imp` objects. If that makes sense. I'm still getting to know the OpenRTB spec and not sure if I got things right here :smile:

bretg:
We'll discuss this in committee, but honestly I'm not real empathetic on the behavior. Other adapters don't support multi-format either, but at least the adapter will process the video part of the request.

The spirit of OpenRTB and of supply-path optimization is that each imp is an adslot. Copying the adslot could be considered a form of bid jamming. Bidders who can support multi-format should be rewarded in the marketplace.

muuki88:
I fully agree with you and don't really have a lot of understanding for not being spec compliant.

Probably it's better to increase the pressure of being spec compliant instead of giving utilities at hand to workaround endpoint limitations.

bretg:
FWIW - this issue came up in https://github.com/prebid/Prebid.js/issues/6910 -- it was resolved there on the PBJS side. 

@braizhas - heads up about the adform adapter -- you might consider refining the multi-format behavior.

PR #2043, opened 2021-10-13
Adf adapter: new bidder params added, multiformat bids supported

braizhas:
Related issue https://github.com/prebid/prebid-server/issues/2006#issuecomment-932281103
Docs: https://github.com/prebid/prebid.github.io/pull/3343

### DIFF
--- a/static/bidder-params/adf.json
+++ b/static/bidder-params/adf.json
@@ -15,7 +15,7 @@
     },
     "mname": {
       "type": ["string"],
-      "description": "A Name which indentifies the placement selling the impression"
+      "description": "A Name which identifies the placement selling the impression"
     }
   },
   "anyOf":[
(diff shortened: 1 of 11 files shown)


static/bidder-params/adf.json
```diff
@@ -8,7 +8,21 @@
...
+      "type": ["string"],
+      "description": "A Name which identifies the placement selling the impression"
     }
   },
-  "required": ["mid"]
+  "anyOf":[
+    {
+      "required": ["mid"]
+    }, {
+      "required": ["inv", "mname"]
```
guscarreon, line 25: When changes in the `required` field of the JSON shcema occur, we encourage bidders to make sure future requests are going to comply because, if not, Prebid Server it will mark the whole request as failed and not send a bid request to any bidder. So adding or modifying a required field can break the entire prebid setup of a publisher.

braizhas:
@muuki88 Is there any chance that the community would do that for me?

PR #2062, opened 2021-10-27
Stored response fetcher (with new param)

VeronikaSolovei9:
Stored response fetcher
Fetch


hhhjort:
1. There are more than two types, there are also "video" and "category" items that can be fetched. The one thing they have in common is that they all fetch json blobs. Imp and request were initially intertwined, as when initially implemented there was no other type of fetcher contemplated. Since then, we have been trying to detangle them as we add more fetchers, and have a common codebase that all fetchers can use. So reuse existing code when it makes sense, and add response specific code as needed.
2.  In router.go, I see a separate `storedRespFetcher` object, which is the point where we need a separate object. As I said above, if we can reuse fetcher code, we should do so. We might consider renaming the module from `stored_requests` to `stored_objects`, since the fetcher framework has already expanded beyond requests. But that is probably out of scope for this PR.
3. The purpose of this code is to provide the ability to do end to end tests without external dependencies. So you should be able to set up a test with an auction object configured with a simple file based stored response fetcher, feed it a request that invokes the stored response, and verify that it returns the expected bids.

endpoints/openrtb2/amp_auction.go
```diff
@@ -86,7 +86,8 @@ func NewAmpEndpoint(
 		bidderMap,
 		nil,
 		nil,
-		ipValidator}).AmpAuction), nil
+		ipValidator,
+		empty_fetcher.EmptyFetcher{}}).AmpAuction), nil
```
hhhjort, line 90: Why are we using an empty fetcher? We should be able to use stored responses on the AMP endpoint I would imagine.

endpoints/openrtb2/auction.go
```diff
@@ -109,6 +111,7 @@ type endpointDeps struct {
 	cache                     prebid_cache_client.Client
 	debugLogRegexp            *regexp.Regexp
 	privateNetworkIPValidator iputil.IPValidator
+	storedRespFetcher         stored_requests.Fetcher
```
hhhjort, line 114: Where is the code that checks for the stored response signal, and then bypasses sending a request to the bidder and just providing the stored response to the flow?

VeronikaSolovei9: This PR is only about stored response fetcher. 
We decided to have smaller PRs that are easy to review rather than having big and complicated PRs.
I understand this fetcher is not connected to anything and has no usages. 
I'll add actual code related to stored responses in another PR.

endpoints/openrtb2/amp_auction.go
VeronikaSolovei9: I'm not sure about this yet. Do we need to support stored response functionality in amp endpoint?

mansinahar:
Closing this in favor of #2099. Will reopen if needed.

PR #2065, opened 2021-11-01
Beachfront - Currency stopgap

muncha:
The Beachfront backend only supports USD. This returns a sane error when non-USD is sent as imp.BidFloorCur, as per the email from @bretg . I'm working on a version that uses the currency conversion. It seems to work, but I'm getting a race condition in testing. So this should cover the requirement until I work that out.


guscarreon (approved):
LGTM

hhhjort (approved):
LGTM

PR #2068, opened 2021-11-02
Triplelift: adds iframe user sync to yaml

nllerandi3lift:
Adds iframe user sync option to yaml.

Any existing Prebid Server host is already allowlisted and so both of these sync urls will continue to work for them. Future hosts will need to reach out as a non-allowlisted `redir` will result in a 500 error.


hhhjort:
Your iframe response seems to drop 11 pixels. Is this mass of pixels truely necessary? Publishers may notice that dropping the triplelift "pixel" leads to more page latency than most other pixels.

`\<body\>

	
	
	
	
	
	
	
	
	
	
	
\</body\>`

nllerandi3lift:
This is the same sync endpoint we've historically used in the prebid js adapter and has not been an issue for publishers. It drops 10 'downstream' pixels (DSPs, DMPs) max - so subsequent, recognized userId's do not necessarily receive all 10. The 11th is the redirect to the prebid server host, which is the only added aspect of our /sync endoint in prebid server vs prebid js. I don't have any concerns over the quantity and existing functionality given that this has been live in prebid js for some years. Thanks for taking a look though.

hhhjort (approved):

VeronikaSolovei9 (approved):

PR #2072, opened 2021-11-05
Syncer Level ExternalURL Override

SyntaxNode:
Allows a host to override the external url at the syncer level in addition to the existing syncer endpoint level. This will make it easier for the host to configure an override since the endpoint type is no longer required, but is still available if different endpoints require different values.

### DIFF
--- a/config/bidderinfo.go
+++ b/config/bidderinfo.go
@@ -102,13 +102,9 @@ func (s *Syncer) Override(original *Syncer) *Syncer {
 
 	if original == nil {
 		copy.IFrame = s.IFrame.Override(nil)
-	} else {
-		copy.IFrame = s.IFrame.Override(original.IFrame)
-	}
-
-	if original == nil {
 		copy.Redirect = s.Redirect.Override(nil)
 	} else {
+		copy.IFrame = s.IFrame.Override(original.IFrame)
 		copy.Redirect = s.Redirect.Override(original.Redirect)
 	}
 
--- a/config/bidderinfo.go
+++ b/config/bidderinfo.go
@@ -68,6 +68,9 @@ type Syncer struct {
 	// endpoint in the Prebid.js project.
 	Redirect *SyncerEndpoint `yaml:"redirect" mapstructure:"redirect"`
 
+	// ExternalURL is available as a macro to the RedirectURL template.
+	ExternalURL string `yaml:"externalUrl" mapstructure:"external_url"`
+
 	// SupportCORS identifies if CORS is supported for the user syncing endpoints.
 	SupportCORS *bool `yaml:"supportCors" mapstructure:"support_cors"`
 }
@@ -107,6 +110,10 @@ func (s *Syncer) Override(original *Syncer) *Syncer {
 		copy.Redirect = s.Redirect.Override(nil)
 	} else {
 		copy.Redirect = s.Redirect.Override(original.Redirect)
+	}
+
+	if s.ExternalURL != "" {
+		copy.ExternalURL = s.ExternalURL
 	}
 
 	if s.SupportCORS != nil {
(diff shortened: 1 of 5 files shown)


hhhjort (approved):
LGTM

config/config.go
```diff
@@ -1102,6 +1102,7 @@ func setBidderDefaults(v *viper.Viper, bidder string) {
 	v.BindEnv(adapterCfgPrefix + ".usersync.redirect.redirect_url")
 	v.BindEnv(adapterCfgPrefix + ".usersync.redirect.external_url")
 	v.BindEnv(adapterCfgPrefix + ".usersync.redirect.user_macro")
+	v.BindEnv(adapterCfgPrefix + ".usersync.external_url")
```
guscarreon, line 1105: Can we add an assertion in `config/config_test.go`? Maybe in `TestFullConfig` or even in the `TestUnmarshalAdapterExtraInfo` test?

config/bidderinfo.go
```diff
@@ -109,6 +112,10 @@ func (s *Syncer) Override(original *Syncer) *Syncer {
 		copy.Redirect = s.Redirect.Override(original.Redirect)
 	}
```
guscarreon, line 113: I realize this isn't part of your PR but, should we use this opportunity to merge the repetitive `if` statements above into one?
```
103     if original == nil {
104         copy.IFrame = s.IFrame.Override(nil)
    +       copy.Redirect = s.Redirect.Override(nil)
105     } else {
106         copy.IFrame = s.IFrame.Override(original.IFrame)
    +       copy.Redirect = s.Redirect.Override(original.Redirect)
107     }
108
109 -   if original == nil {
110 -       copy.Redirect = s.Redirect.Override(nil)
111 -   } else {
112 -       copy.Redirect = s.Redirect.Override(original.Redirect)
113 -   }
config/bidderinfo.go
```
Or maybe we could correct it in a separate PR and properly test the change there. I'm good either way.

usersync/syncer.go
```diff
@@ -147,13 +147,16 @@ var (
...
+func buildTemplate(key, syncTypeValue string, hostConfig config.UserSync, syncerExternalURL string, syncerEndpoint config.SyncerEndpoint) (*template.Template, error) {
 	redirectTemplate := syncerEndpoint.RedirectURL
 	if redirectTemplate == "" {
 		redirectTemplate = hostConfig.RedirectURL
 	}
 
 	externalURL := syncerEndpoint.ExternalURL
+	if externalURL == "" {
+		externalURL = syncerExternalURL
+	}
```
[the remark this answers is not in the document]
SyntaxNode: Refactored.

config/config.go
SyntaxNode: Do you really think we need to? There's nothing special with this config value. Adding it to `TestFullConfig` is really just testing Viper and not so much this specific value.

config/bidderinfo.go
SyntaxNode: Sure. I didn't notice the repetition. Collapsed.

config/config.go
guscarreon: Sure, I can live with that.

config/bidderinfo.go
```diff
@@ -162,8 +165,8 @@ type SyncerEndpoint struct {
 	// `{{.ExternalURL}}/setuid?bidder={{.SyncerKey}}&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&f={{.SyncType}}&uid={{.UserMacro}}`
 	RedirectURL string `yaml:"redirectUrl" mapstructure:"redirect_url"`
 
-	// ExternalURL is available as a macro to the RedirectURL template. If not specified, the host configuration
-	// value is used.
+	// ExternalURL is available as a macro to the RedirectURL template. If not specified, either the syncer configuration
+	// value or the host configuration value is used.
```
guscarreon, line 169: Great wording

guscarreon (approved):
LGTM

VeronikaSolovei9 (approved):

PR #2071, opened 2021-11-03
Currency conversion on adapter JSON tests

guscarreon:
Pull request #1901 granted adapters the ability to convert currency inside their `MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo)` implementation. This pull request makes possible for adapters to include their own currency conversion rates in their JSON tests in order to better assert #1901's features

Also commited in this PR are JSON tests for the impactify adapter. Impactify's implementation of `MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo)` makes use of currency conversion and the additions to `adapters/impactify/impactifytest/exemplary/sample_banner.json` assert the correct conversion fo the impression bid floor, and on the other hand `adapters/impactify/impactifytest/supplemental/currency_rate_not_found.json` asserts a currency conversion error scenario.

@thomasdseao we'd like to hear your take on the tests added to your adapter. Please let us know if you approve.

PR #2078, opened 2021-11-10
Beachfront - Currency conversion

muncha:
This handles currency conversion on imp.bidfloor, corrects and expands on the logic of selecting between imp.bidfloor and ext.bidfloor, and when to apply the minimum. Testing is based on the Rubicon TestOpenRTBRequestWithDifferentBidFloorAttributes() method.


AlexBVolcy (commented):
The code updates and the test coverage from these updates looks good! I'm going to wait for the review of @guscarreon to see if I missed anything.

adapters/beachfront/beachfront.go
```diff
@@ -88,8 +88,8 @@ type beachfrontSlot struct {
 }
 
 type beachfrontSize struct {
-	W uint64 `json:"w"`
-	H uint64 `json:"h"`
+	W int `json:"w"`
```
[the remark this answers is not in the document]
muncha: I was winding up with a hex value when values were sent by the test. It was the right hex value, as in 300x250 became 0x12c x 0xFA. I have no idea why that was happening. It seemed really weird. I had no actual need for them to be uint64, and int worked, so I changed it and left it at that. It had been working for years as uint64. I don't recall why I used uint64 in the first place. I don't claim now to be a Go guru, but at the time it was quite a new language for me, and it was probably just the first hint my IDE gave. I am curious about that and will try a bit of debugging with them as uint64 and see if I can tell what is happening.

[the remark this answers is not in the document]
AlexBVolcy: Got it thanks for updating! I'll look over this again today.

AlexBVolcy (dismissed):

guscarreon (commented):
Looks good to me @muncha. Just one more ask: Now that pull request #2071 has been merged, JSON tests with currency conversion are now supported. Can you convert `TestRequestWithDifferentBidFloorAttributes` to JSON tests? 
Currency conversion rates get read from the `request.ext.prebid.currency.rates` field as shown below:
```
  "mockBidRequest": {
    "id": "test-request-id",
       .
       .
    "imp": [
       .
       .
    ],
    "ext": {
      "prebid": {
        "currency": {
          "rates": {
            "MXN": {
              "USD": 0.05,
              "EUR": 0.03
            }
            "EUR": {
              "USD": 1.12
            }
              .
              .
          },
          "usepbsrates": false
        }
      }
    }
  }
```

muncha:
@guscarreon FYI - I'm working on switching it to JSON, I'm hitting a panic on the currency conversion and trying to get to the bottom of that.

guscarreon:
> @guscarreon FYI - I'm working on switching it to JSON, I'm hitting a panic on the currency conversion and trying to get to the bottom of that.

Thank's @muncha. Currency tests will panic if the recent currency conversion test functionality recently merged into master is not in place. Maybe updating your local branch will latest master could probably solve it.

muncha:
That was the problem. I thought I had that already merged to beachfront's master. I did not.

guscarreon (commented):
Looks pretty good, just have a couple of questions:

adapters/beachfront/beachfront.go
```diff
@@ -568,29 +563,51 @@ func (a *BeachfrontAdapter) MakeBids(internalRequest *openrtb2.BidRequest, exter
...
+						err,
+					),
+				}
+			}
 		}
 	}
 
-	if floor <= minBidFloor {
-		floor = 0
+	if imp.BidFloor < ext.BidFloor {
```
guscarreon, line 599: When execution gets here, there's a possibility that `imp.BidFloor` has already been converted from a foreign currency to "USD". Are we positive that `ext.BidFloor` comes in "USD" and not the same foreign currency `imp.BidFloor` was? 

For instance: say a given request comes with both `imp.BidFloor` and `ext.BidFloor` set to "EUR". The comparisson here would be misleading because we could be comparing `imp.BidFloor` representing a price in US dollars versus `ext.BidFloor` representing Euros. Is this a feasible possibility? Can `ext.BidFloor` come in a non "USD" currency?

muncha: @guscarreon 
ext.bidfloor is always is USD. So in the case where imp.bidfloor has been converted to USD from EUR, the ext.bidfloor is not converted (it is never converted). Note the comparison at 577. This also assumes that ext.bidfloor is USD, as is minBidFloor (which beachfront may one day decide should be more than $0.01)

If ext.bidfloor is erroneously sent in non-USD, that is the value that is used. If imp.bidfloorcur = INR, imp.bidfloor= 375.23, and ext.bidfloor = 310.21, then ext.bidfloor will be used, as the converted value of imp.bidfloor = 5.00 USD is less than ext.bidfloor = 310.21 USD, which was entered in INR due to user error, and someone's not getting any ads. I verified this with beachfront, and I agree, it's a bit counter intuitive, but it is correct.

I need to update the docs as they still say the converter doesn't work, but this bit is in there as of my previous PR. I'll push that in a couple minutes.

AlexBVolcy (approved):
This looks good to me! Thanks for being so responsive to feedback!

adapters/beachfront/beachfront.go
guscarreon: Thanks for clarifying @muncha

guscarreon (approved):
LGTM. @muncha thanks for clarifying and addressing our feedback

PR #2079, opened 2021-11-10
33Across: Enable Support for SRA requests

curlyblueeagle:
In this PR we introduce SRA and group together Ad Units that have the same parameters under the same Bid Request in order to minimize http calls to our end point.


AlexBVolcy (commented):
The code to update to support SRA looks good, but I want to bring to your attention that TestJsonSamples is failing with a test case from this path, 33acrosstest/exemplary/multi-imp-banner.json. Could you address this fail case?

curlyblueeagle:
> The code to update to support SRA looks good, but I want to bring to your attention that TestJsonSamples is failing with a test case from this path, 33acrosstest/exemplary/multi-imp-banner.json. Could you address this fail case?

@AlexBVolcy tests seem to be running fine on my local and in our org's pipeline. Also it looks like the order of http reqs made of the Ad Requests matters in tests, so multiple requests would result in `expectedRequest` to not always have the same order and therefore the tests to flake. Do you have any suggestions on how I can check for the presence of requests in an agnostic manner?

AlexBVolcy:
@curlyblueeagle I'll have an answer for what to do about this tomorrow!

AlexBVolcy:
So after some investigating, the reason for this test failure is on our end, due to a shortcoming around the JSON Test Framework. 

So while we work on updating this framework, we'll need you to remove the test that is failing from this PR. We'll verify locally that this test is working properly, and we can approve this PR. 

Once we update the test framework, you can then open a new PR to add this test back in. 

Let me know if you have any other questions regarding this!

guscarreon:
@curlyblueeagle pull request #2087 is meant to correct the false negatives our testing framework throws under certain `MakeRequests()` outputs. As soon as this newly posted PR gets approved and merged, your changes will pass validation (tested your changes locally with the test framework fix). Will let you know as soon as your fix gets merged

curlyblueeagle:
@AlexBVolcy @guscarreon sorry I just removed the test that triggered the fail. Should I be waiting for the testing framework PR to be merged instead?

guscarreon:
> @AlexBVolcy @guscarreon sorry I just removed the test that triggered the fail. Should I be waiting for the testing framework PR to be merged instead?

No @curlyblueeagle, we'll add your test case to #2087 instead. Is that ok?

guscarreon:
@curlyblueeagle We've added your test case to #2087, it currently fails because this PR has not been merged yet. @AlexBVolcy and @VeronikaSolovei9 will continue to review your PR.

AlexBVolcy (approved):

VeronikaSolovei9 (approved):

mansinahar:
@curlyblueeagle Can you please review the test added to #2087 for 33across to test this behavior and make sure it looks good?

curlyblueeagle:
@mansinahar reviewed and looks good

PR #2081, opened 2021-11-11
Add a garbage collection threshold

guscarreon:
Pull request #2049 removed the legacy `/auction` endpoint. An unintended consequence of removing the legacy code was an increase in CPU time caused by garbage collection overhead. After a thorough investigation of the performance of the system, the Prebid Server Team found out that the initialization of a very large slice in the old code was preventing the garbage collection to be called repeatedly, saving CPU time. This pull request puts back a soft memory limit to save CPU performance and makes this value configurable.

Issue [#48409](https://github.com/golang/go/issues/48409) posted in the go development project better explains what happens behind the scenes, in summary, we are setting an amount of memory that is going to be held garbage before a garbage collection cycle gets triggered.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -24,9 +24,9 @@ type Configuration struct {
 	CacheClient HTTPClient `mapstructure:"http_client_cache"`
 	AdminPort   int        `mapstructure:"admin_port"`
 	EnableGzip  bool       `mapstructure:"enable_gzip"`
-	// GarbageCollectorLimit will serve as a fixed cost of memory that is going to be held garbage
-	// before a garbage collection cycle is triggered
-	GarbageCollectorLimit int `mapstructure:"garbage_collector_limit"`
+	// GarbageCollectorThreshold allocates virtual memory (in bytes) which is not used by PBS but
+	// serves as a hack to trigger the garbage collector only when the heap reaches at least this size.
+	GarbageCollectorThreshold int `mapstructure:"garbage_collector_threshold"`
 	// StatusResponse is the string which will be returned by the /status endpoint when things are OK.
 	// If empty, it will return a 204 with no content.
 	StatusResponse    string          `mapstructure:"status_response"`
@@ -637,7 +637,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("port", 8000)
 	v.SetDefault("admin_port", 6060)
 	v.SetDefault("enable_gzip", false)
-	v.SetDefault("garbage_collector_limit", 0)
+	v.SetDefault("garbage_collector_threshold", 0)
 	v.SetDefault("status_response", "")
 	v.SetDefault("auction_timeouts_ms.default", 0)
 	v.SetDefault("auction_timeouts_ms.max", 0)
--- a/config/config.go
+++ b/config/config.go
@@ -24,6 +24,9 @@ type Configuration struct {
 	CacheClient HTTPClient `mapstructure:"http_client_cache"`
 	AdminPort   int        `mapstructure:"admin_port"`
 	EnableGzip  bool       `mapstructure:"enable_gzip"`
+	// GarbageCollectorLimit will serve as a fixed cost of memory that is going to be held garbage
+	// before a garbage collection cycle is triggered
+	GarbageCollectorLimit int `mapstructure:"garbage_collector_limit"`
 	// StatusResponse is the string which will be returned by the /status endpoint when things are OK.
 	// If empty, it will return a 204 with no content.
 	StatusResponse    string          `mapstructure:"status_response"`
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -24,6 +24,9 @@ type Configuration struct {
 	CacheClient HTTPClient `mapstructure:"http_client_cache"`
 	AdminPort   int        `mapstructure:"admin_port"`
 	EnableGzip  bool       `mapstructure:"enable_gzip"`
+	// GarbageCollectorLimit will serve as a fixed cost of memory that is going to be held garbage
+	// before a garbage collection cycle is triggered
+	GarbageCollectorLimit int `mapstructure:"garbage_collector_limit"`
```
SyntaxNode, line 29: I think we can improve the wording for clarity, and a link to the Go issue might be a good idea. Perhaps something like:

// GarbageCollectorLimit allocates virtual memory (in bytes) which is not used by PBS but serves as a hack to trigger the garbage collector only when the heap reaches at least this size.

SyntaxNode, line 29: For naming, I like that GarbageCollector conveys the right intent, but I'm not sure "Limit" is a precise enough descriptor. Maybe "GarbageCollectorMinimum", "GarbageCollectorAllocationHack", "GarbageCollectorThreshold" etc...

config/config_test.go
```diff
@@ -289,6 +289,7 @@ external_url: http://prebid-server.prebid.org/
 host: prebid-server.prebid.org
 port: 1234
 admin_port: 5678
+garbage_collector_limit: 32768
```
SyntaxNode, line 292: Consider using a simpler test value, like `1`. :)

main.go
```diff
@@ -28,6 +30,13 @@ func main() {
 		glog.Exitf("Configuration could not be loaded or did not pass validation: %v", err)
 	}
 
+	// Create a soft memory limit on the total amount of memory that PBS uses to tune the behavior
+	// of the Go garbage collector. In summary, `cfg.GarbageCollectorLimit` serves as a fixed cost
+	// of memory that is going to be held garbage before a garbage collection cycle is triggered.
+	// This amount of virtual memory won’t translate into physical memory allocation unless we attempt
+	// to read or write to the slice below, which PBS will not do.
+	garbageCollectionLimit = make([]byte, cfg.GarbageCollectorLimit)
```
SyntaxNode, line 38: We'll need to test this works as intended. We saw the following recommendation from a Medium article:
```
func main() {
	ballast := make([]byte, 2<<30)
	defer runtime.KeepAlive(ballast)
}
```

SyntaxNode:
> An unintended consequence of removing the legacy code was an increase in CPU time caused by garbage collection overhead. 

To clarify, this is entirely dependent on how the Host has configured their PBS instance. This does not affect everyone.

config/config.go
VeronikaSolovei9: GarbageCollectorMemoryBallast maybe?

guscarreon: Far better worded. Done

guscarreon: `GarbageCollectorThreshold` makes the most sense to me (I even named this PR that). @VeronikaSolovei9, @SyntaxNode let me know your veredict.

config/config_test.go
guscarreon: Done

main.go
guscarreon: Going with the `defer runtime.KeepAlive(garbageCollectionLimit)` recommendation then.

config/config.go
SyntaxNode: I like GarbageCollectorThreshold.

config/config_test.go
```diff
@@ -690,6 +692,19 @@ func TestMigrateConfigFromEnv(t *testing.T) {
...
+		defer os.Setenv("PBS_GARBAGE_COLLECTOR_THRESHOLD", oldval)
+	} else {
+		defer os.Unsetenv("PBS_GARBAGE_COLLECTOR_THRESHOLD")
+	}
+	os.Setenv("PBS_GARBAGE_COLLECTOR_THRESHOLD", "1")
+	cfg, _ := newDefaultConfig(t)
+
+	// Assert config value differs from the default value of 0
+	cmpInts(t, "garbage_collector_threshold", 1, cfg.GarbageCollectorThreshold)
+}
```
SyntaxNode, line 706: I don't believe this test adds value.

SyntaxNode (approved):
Looks good. You may wish to remove the environment variable test, but not a deal breaker.

config/config_test.go
guscarreon: Removed

SyntaxNode (approved):

config/config.go
```diff
@@ -24,6 +24,10 @@ type Configuration struct {
 	CacheClient HTTPClient `mapstructure:"http_client_cache"`
 	AdminPort   int        `mapstructure:"admin_port"`
 	EnableGzip  bool       `mapstructure:"enable_gzip"`
+	// GarbageCollectorThreshold allocates virtual memory (in bytes) which is not used by PBS but
+	// serves as a hack to trigger the garbage collector only when the heap reaches at least this size.
+	// More info: https://github.com/golang/go/issues/48409
+	GarbageCollectorThreshold int `mapstructure:"garbage_collector_threshold"`
```
mansinahar, line 30: PR looks good to me in general. The only thing I'd say is if we can rename this to be something along the lines of `GCMemoryThreshold` or `GCMemoryLimit` so that it's clear that it's a memory threshold for the Go GC.

mansinahar (approved):

bsardo (approved):
LGTM

PR #2082, opened 2021-11-15
Added integration type support for host company

AlexBVolcy:
This PR partially addresses the Integration Type and Channel issue https://github.com/prebid/prebid-server/issues/1428

There will be a series of smaller PRs to address this issue, another PR is currently out for adding channel support here https://github.com/prebid/prebid-server/pull/2037

This PR is the first one dedicated towards the Integration part of the issue. What this PR does is it adds the functions that allow us to get the integration type (`getIntegrationTypeFromRequest()`) value found in a openRTB request, and to validate that the integration value in the request is allowed by the host company (`validateIntegrationType()`).

Future integration type focused PRs will add support for publishers to define a `defaultIntegration` value, and if that value can be overridden.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -84,6 +84,9 @@ type Configuration struct {
 	GenerateBidID bool `mapstructure:"generate_bid_id"`
 	// GenerateRequestID overrides the bidrequest.id in an AMP Request or an App Stored Request with a generated UUID if set to true. The default is false.
 	GenerateRequestID bool `mapstructure:"generate_request_id"`
+	// IntegrationTypes will hold the integration type values that the host will allow.
+	// Map structure isn't included here because the integration type feature is still being worked on, but once complete, this variable will be updated
+	IntegrationTypes map[string]bool
 }
 
 const MIN_COOKIE_SIZE_BYTES = 500
--- a/config/config.go
+++ b/config/config.go
@@ -84,9 +84,9 @@ type Configuration struct {
 	GenerateBidID bool `mapstructure:"generate_bid_id"`
 	// GenerateRequestID overrides the bidrequest.id in an AMP Request or an App Stored Request with a generated UUID if set to true. The default is false.
 	GenerateRequestID bool `mapstructure:"generate_request_id"`
-	// IntegrationTypes will hold the integration type values that the host will allow.
+	// IntegrationTypesMap will hold the integration type values that the host will allow.
 	// Map structure isn't included here because the integration type feature is still being worked on, but once complete, this variable will be updated
-	IntegrationTypes map[string]bool
+	IntegrationTypesMap map[string]struct{}
 }
 
 const MIN_COOKIE_SIZE_BYTES = 500
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/prebid/prebid-server/firstpartydata"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -14,8 +13,10 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/prebid/prebid-server/firstpartydata"
+
 	"github.com/buger/jsonparser"
-	"github.com/evanphx/json-patch"
+	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/gofrs/uuid"
 	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
@@ -1652,4 +1653,24 @@ func checkIfAppRequest(request []byte) (bool, error) {
 		return true, nil
 	}
 	return false, nil
+}
+
+func getIntegrationTypeFromRequest(req *openrtb_ext.RequestWrapper) (string, error) {
+	reqExt, err := req.GetRequestExt()
+	if err != nil {
+		return "", err
+	}
+	reqPrebid := reqExt.GetPrebid()
+	return reqPrebid.IntegrationType, nil
+}
+
+func validateIntegrationType(req *openrtb_ext.RequestWrapper, integrationTypes map[string]bool) error {
+	integrationType, err := getIntegrationTypeFromRequest(req)
+	if err != nil {
+		return err
+	}
+	if _, ok := integrationTypes[integrationType]; ok {
+		return nil
+	}
+	return errors.New("Integration value is invalid because it's not in allowed list from host company")
 }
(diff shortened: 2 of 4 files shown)


endpoints/openrtb2/auction.go
```diff
@@ -1653,3 +1654,23 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
+}
+
+func validateIntegrationType(req *openrtb_ext.RequestWrapper, integrationTypes map[string]bool) error {
+	integrationType, err := getIntegrationTypeFromRequest(req)
+	if err != nil {
+		return err
+	}
+	if _, ok := integrationTypes[integrationType]; ok {
+		return nil
+	}
```
bsardo, line 1674: If the integration has not been specified in the request, I believe that's valid according to the corresponding issue, is that correct? It looks like the integration would be set to some default value specified in the config in this case. If I have that right, you'll want to handle that case here as I don't imagine specifying an empty string or nil value in the slice of valid integration values in the config.

endpoints/openrtb2/auction.go
```diff
@@ -1653,3 +1654,23 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
+func getIntegrationTypeFromRequest(req *openrtb_ext.RequestWrapper) (string, error) {
+	reqExt, err := req.GetRequestExt()
+	if err != nil {
+		return "", err
+	}
+	reqPrebid := reqExt.GetPrebid()
+	return reqPrebid.IntegrationType, nil
+}
+
+func validateIntegrationType(req *openrtb_ext.RequestWrapper, integrationTypes map[string]bool) error {
```
bsardo, line 1667: Are you planning to add test coverage for these functions in this PR or a future related PR?

openrtb_ext/request.go
```diff
@@ -33,6 +33,7 @@ type ExtRequestPrebid struct {
 	Data                 *ExtRequestPrebidData     `json:"data,omitempty"`
 	Debug                bool                      `json:"debug,omitempty"`
 	Events               json.RawMessage           `json:"events,omitempty"`
+	IntegrationType      string                    `json:"integration,omitempty"`
```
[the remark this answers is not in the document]
bsardo: Actually maybe I have that wrong. Can the integration values be whatever the host wants?

endpoints/openrtb2/auction.go
AlexBVolcy: Yeah I'll add more test coverage in this PR!

endpoints/openrtb2/auction.go
AlexBVolcy: You're right in bringing up this point! However, I'll handle this Default Integration in a future PR, since we're breaking up my implementation of Integration Type and Channel into multiple smaller PRs.

endpoints/openrtb2/auction.go
```diff
@@ -1653,3 +1654,23 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
 	return false, nil
 }
+
+func getIntegrationTypeFromRequest(req *openrtb_ext.RequestWrapper) (string, error) {
+	reqExt, err := req.GetRequestExt()
+	if err != nil {
+		return "", err
+	}
+	reqPrebid := reqExt.GetPrebid()
+	return reqPrebid.IntegrationType, nil
```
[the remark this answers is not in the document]
AlexBVolcy: Will update!

openrtb_ext/request.go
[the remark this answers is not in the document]
AlexBVolcy: Yeah the integration values allowed are set by the host to whatever they want!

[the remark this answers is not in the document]
AlexBVolcy: Will update!

config/config.go
```diff
@@ -84,6 +84,9 @@ type Configuration struct {
 	GenerateBidID bool `mapstructure:"generate_bid_id"`
 	// GenerateRequestID overrides the bidrequest.id in an AMP Request or an App Stored Request with a generated UUID if set to true. The default is false.
 	GenerateRequestID bool `mapstructure:"generate_request_id"`
+	// IntegrationTypes will hold the integration type values that the host will allow.
+	// Map structure isn't included here because the integration type feature is still being worked on, but once complete, this variable will be updated
+	IntegrationTypes map[string]bool
```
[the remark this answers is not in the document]
AlexBVolcy: Thank you for this comment, I'll update this maps value accordingly!

config/config.go
```diff
@@ -84,6 +84,9 @@ type Configuration struct {
 	GenerateBidID bool `mapstructure:"generate_bid_id"`
 	// GenerateRequestID overrides the bidrequest.id in an AMP Request or an App Stored Request with a generated UUID if set to true. The default is false.
 	GenerateRequestID bool `mapstructure:"generate_request_id"`
+	// IntegrationTypesMap will hold the integration type values that the host will allow.
+	// Map structure isn't included here because the integration type feature is still being worked on, but once complete, this variable will be updated
+	IntegrationTypesMap map[string]struct{}
```
hhhjort, line 89: Two questions. First is how do you plan on populating `IntegrationTypesMap`? For the above Blacklist maps, we have two entries, one a slice of strings, and the second is the actual map. In your configuration you would provide the list of values, and then config code builds the map from the list.

The second question is how do we default to integration types not being required? We don't want to break every install of PBS that is not currently requiring integration types to be sent in requests.

AlexBVolcy: To your first question, I like the idea of copying how the Blacklist maps work, so I could have one variable `IntegrationTypes` that is just a slice of strings indicating what IntegrationTypes the host company allows, and this slice can be filled with another `SetDefault` of that `SetupViper` function. And then that slice can be apart of the `IntegrationTypesMap`. And then we can continue to utilize the fast lookup the map allows us. 

To your second question, integration types are not being required in this PR since we never actually call any of the validate/get functions set up for integration types, so the flow of current PBS instances won't be broken. But in the future, we could create a simple boolean (i.e. `AllowIntegrationTypes`), that if false, avoids any evaluation of integration type information. Or host company's could choose to not populate the IntegrationTypes slice, and then we could add a check that makes it if IntegrationTypes slice is empty, we just continue evaluating the request ignoring IntegrationType related validation.

hhhjort: Oh, yes, you don't actually call these new functions. Seems like a small enough addition, was there a reason that the actual activation was left to another PR?

hhhjort (approved):
Looks good, even if it just adds code but doesn't activate it. Will need a followup PR to activate this functionality, but will go ahead and set my approval now.

config/config.go
AlexBVolcy: Since we still have to implement the integration type features for accounts (i.e. Default Integration), we can leave the activation for that PR!

PR #2083, opened 2021-11-15
Adyoulike: Add an iframe sync endpoint

chrisg93:


AlexBVolcy (approved):
Ran through a test of this update with @VeronikaSolovei9 and it looks good!

VeronikaSolovei9 (approved):

PR #2084, opened 2021-11-15
Adform adapter lacked gross/net parameter support

bjorn-lw:
The client version of the Adform adapter supports choosing between returning gross or net via the priceType parameter.

As the new "adf" adapter defaults to net while the old "adform" adapter defaulted to gross, this can make the transition to the new adapter become quite a challenge if you are using plenty of bid conversion functions and the PBS adapter doesn't support selecting gross or net.

This PR adds the same priceType support as the client has.

### DIFF
--- a/adapters/adf/adf.go
+++ b/adapters/adf/adf.go
@@ -16,6 +16,10 @@ type adapter struct {
 	endpoint string
 }
 
+type adfRequestExt struct {
+	PriceType string `json:"pt"`
+}
+
 // Builder builds a new instance of the Adf adapter for the given bidder with the given config.
 func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
 	bidder := &adapter{
@@ -47,6 +51,15 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 
 		imp.TagID = adfImpExt.MasterTagID.String()
 		validImps = append(validImps, imp)
+
+		if adfImpExt.PriceType != "" {
+			requestExt := adfRequestExt{PriceType: adfImpExt.PriceType}
+			var err error
+			if request.Ext, err = json.Marshal(&requestExt); err != nil {
+				errors = append(errors, err)
+				continue
+			}
+		}
 	}
 
 	request.Imp = validImps
--- a/adapters/adf/adf.go
+++ b/adapters/adf/adf.go
@@ -62,18 +62,20 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 
 	if priceType != "" {
 		requestExt := adfRequestExt{}
+		var err error
 
 		if len(request.Ext) > 0 {
-			if err := json.Unmarshal(request.Ext, &requestExt); err != nil {
+			if err = json.Unmarshal(request.Ext, &requestExt); err != nil {
 				errors = append(errors, err)
 			}
 		}
 
-		requestExt.PriceType = priceType
+		if err == nil {
+			requestExt.PriceType = priceType
 
-		var err error
-		if request.Ext, err = json.Marshal(&requestExt); err != nil {
-			errors = append(errors, err)
+			if request.Ext, err = json.Marshal(&requestExt); err != nil {
+				errors = append(errors, err)
+			}
 		}
 	}
 
--- a/adapters/adf/adf.go
+++ b/adapters/adf/adf.go
@@ -17,6 +17,7 @@ type adapter struct {
 }
 
 type adfRequestExt struct {
+	openrtb_ext.ExtRequest
 	PriceType string `json:"pt"`
 }
 
@@ -53,7 +54,19 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 		validImps = append(validImps, imp)
 
 		if adfImpExt.PriceType != "" {
-			requestExt := adfRequestExt{PriceType: adfImpExt.PriceType}
+			var requestExt adfRequestExt
+
+			if len(request.Ext) > 0 {
+				if err := json.Unmarshal(request.Ext, &requestExt); err != nil {
+					errors = append(errors, err)
+					continue
+				}
+			} else {
+				requestExt = adfRequestExt{}
+			}
+
+			requestExt.PriceType = adfImpExt.PriceType
+
 			var err error
 			if request.Ext, err = json.Marshal(&requestExt); err != nil {
 				errors = append(errors, err)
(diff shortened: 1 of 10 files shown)


bjorn-lw:
Needs more work

bjorn-lw:
I'm a bit uncertain if I should preserve existing request.ext structure or if I can simply overwrite it, please advice.

bjorn-lw:
@braizhas maybe you want to have a look at this as well.

adapters/adf/adf.go
```diff
@@ -47,6 +52,27 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
...
+					continue
+				}
+			} else {
+				requestExt = adfRequestExt{}
+			}
+
+			requestExt.PriceType = adfImpExt.PriceType
+
+			var err error
+			if request.Ext, err = json.Marshal(&requestExt); err != nil {
```
VeronikaSolovei9, line 71: Line 60: you marshal req.Ext as many times as namy imps are thre in reqiest - it's located inside `for _, imp := range request.Imp {`. This should be moved ouside the for loop. 
Line 71 sets new request.ext on every iteration over imp array. It `means req.ext.PriceType` will have the value from last imp in loop. 
How will it work on your saide if there are two or more imp in request with different `PriceType`?

adapters/adf/adftest/exemplary/single-banner-pricetype-gross-extend-ext.json
```diff
@@ -0,0 +1,122 @@
...
+        }
+      }
+    },
+    "imp": [{
+      "id": "test-imp-id",
+      "ext": {
+        "bidder": {
+          "mid": "828782",
+          "priceType": "gross"
+        }
```
VeronikaSolovei9, line 17: Please add a test case where there are 2 imps in request with different `priceType`

adapters/adf/adf.go
```diff
@@ -47,6 +52,27 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
...
+			var requestExt adfRequestExt
+
+			if len(request.Ext) > 0 {
+				if err := json.Unmarshal(request.Ext, &requestExt); err != nil {
+					errors = append(errors, err)
+					continue
+				}
+			} else {
+				requestExt = adfRequestExt{}
+			}
```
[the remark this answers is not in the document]
bjorn-lw: Thanks! It will be initiated even if it may not be needed but no harm done in that I guess.

adapters/adf/adf.go
bjorn-lw: All priceTypes need to be the same, but really only the first one should be used and not the last one if they should differ.
I took a bad shortcut there. Me programming Go is also a bit like https://youtu.be/C1Sw0PDgHU4.

adapters/adf/adftest/exemplary/single-banner-pricetype-gross-extend-ext.json
bjorn-lw: Not sure if that is really needed but I can add one.

bjorn-lw: Done. Thank you @VeronikaSolovei9 for the feedback!

VeronikaSolovei9:
Please update adapter documentation with the description of new parameter

bjorn-lw:
> Please update adapter documentation with the description of new parameter

https://github.com/prebid/prebid.github.io/pull/3417

VeronikaSolovei9 (approved):
Thank you for the fixes, LGTM :)

braizhas:
> @braizhas maybe you want to have a look at this as well.

@bjorn-lw thanks for the notification.
Changes are legit and would enable gross price bids from Adform.

Please also update `adform` bidder params as now it acts as an alias to `adf` bidder. https://github.com/prebid/prebid-server/blob/063990f8cd8b43b5c8d6d25c1a4ac449350710f7/static/bidder-params/adform.json

bjorn-lw:
> Please also update `adform` bidder params as now it acts as an alias to `adf` bidder. https://github.com/prebid/prebid-server/blob/063990f8cd8b43b5c8d6d25c1a4ac449350710f7/static/bidder-params/adform.json

Fixed, thank you!

adapters/adf/adf.go
```diff
@@ -47,6 +53,28 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
...
+		if len(request.Ext) > 0 {
+			if err := json.Unmarshal(request.Ext, &requestExt); err != nil {
+				errors = append(errors, err)
+			}
+		}
+
+		requestExt.PriceType = priceType
+
+		var err error
+		if request.Ext, err = json.Marshal(&requestExt); err != nil {
```
hhhjort, line 75: At this point, you are remarshalling `requestExt`, which may not be properly populated if the above unmarshal failed. This should be very unlikely, perhaps impossible, but you may want to consider if you want the request to progress through this stage if an error happened.

bjorn-lw: @hhhjort Thanks! It should be be fixed, but I don't like the many nested if's. I would probably have done it a little different in a language I'm more familiar with. If you can give me a hint on how to make it look nicer, please do.

hhhjort: Actually this doesn't fix it. If you hit line 69, having encountered and error with unmarshalling the `.ext`, you still drop down and overwrite the original `.ext` on line 75. You would have to break out of the section, perhaps returning an error from `MakeRequests()` at that line, to avoid this issue. I am not terribly concerned with this, as it is something that should really never happen in the first place. And if it does somehow happen, it just means that the `request.ext` you are sending to your server may have missing/corrupt data in it. I really just wanted to highlight it so you could decide if it is a worry for you and change your handling if it is. If you reply that you are good with it as is, or just do nothing, I am still good with approving this PR. Just giving you some time to respond before it is merged.

hhhjort: Ack, nevermind, it does fix it. I just did not read the code properly.

hhhjort (approved):
LGTM

adapters/adf/adf.go
bjorn-lw: @hhhjort I was frantically reading trying to understand what I missed :) But I think the code is ugly. Can it be written better? You don't have to reply, I could take that online course on Go if I can just get some additional hours in a day :)

hhhjort: I don't have any specific recommendations to make this code prettier, it seems normal enough go to me. Well, one thing that might improve the view...

In the initial loop where you try to find the price type, you are only interested in the first value you find. So when you do find it, you could instead call `setRequestPriceType(request *openrtb2.BidRequest, priceType string)`, which would be the following block of code that sets the price type in `request.ext`.  Then in that function, you could return immediately if the unmarshal fails, making it more obvious that a failure aborts the rest of price type processing. Moving the code to its own function isolates it as its own step, leaving the main function cleaner in its logic. You coud also put the loop to find the price type in the function, making it easier to break out of the loop and further focusing the main function's code.

bjorn-lw: @hhhjort yes, common clean code sense.  I may leave it as is for now if you don't mind. I might get back to this at a later time if I come up with something else that needs to be done in the adapter. Is that OK?

hhhjort: That is fine, I have already approved this PR

VeronikaSolovei9 (approved):

PR #2087, opened 2021-11-16
Correct MakeRequests() output assertion in adapter JSON tests

guscarreon:
Current version of test framework function `assertMakeRequestsOutput(t *testing.T, filename string, actual []*adapters.RequestData, expected []httpCall)` throws a false negative when the order of the entries in the parameter slices `actual` and `expected` differ even when they come with matching elements. This pull request implements a correction that runs in O(n^2) time but, because tests are not run in production, we don't mind the extra processing time given that most tests come with slices of size 1 anyways.


adapters/adapterstest/test_json.go
```diff
@@ -242,19 +249,28 @@ func assertMakeBidsOutput(t *testing.T, filename string, bidderResponse *adapter
...
-		actualHeader, _ := json.Marshal(actual.Headers)
-		expectedHeader, _ := json.Marshal(expected.Headers)
-		diffJson(t, description, actualHeader, expectedHeader)
+		actualHeader, err := json.Marshal(actual.Headers)
+		if err != nil {
+			return fmt.Errorf(`%s actual.Headers could not be marshalled. Error: %s"`, description, err.Error())
+		}
+		expectedHeader, err := json.Marshal(expected.Headers)
+		if err != nil {
+			return fmt.Errorf(`%s expected.Headers could not be marshalled. Error: %s"`, description, err.Error())
```
guscarreon, line 269: Not ignoring marshalling errors anymore. (If any)

adapters/adapterstest/test_json.go
```diff
@@ -263,60 +279,61 @@ func diffBids(t *testing.T, description string, actual *adapters.TypedBid, expec
...
 // byte array is not actually JSON.
-func diffJson(t *testing.T, description string, actual []byte, expected []byte) {
+func diffJson(description string, actual []byte, expected []byte) error {
 	if len(actual) == 0 && len(expected) == 0 {
-		return
+		return nil
 	}
 	if len(actual) == 0 || len(expected) == 0 {
-		t.Fatalf("%s json diff failed. Expected %d bytes in body, but got %d.", description, len(expected), len(actual))
+		return fmt.Errorf("%s json diff failed. Expected %d bytes in body, but got %d.", description, len(expected), len(actual))
```
guscarreon, line 309: Now that `diffJson()` returns an error, no need to differentiate this error as`Fatalf` over `Errorf`. This particular JSON test won't continue to run because our test framework will simply go to the next JSON file in the directory.

adapters/adapterstest/test_json.go
```diff
@@ -263,60 +282,54 @@ func diffBids(t *testing.T, description string, actual *adapters.TypedBid, expec
 		return
 	}
 
-	diffStrings(t, fmt.Sprintf("%s.type", description), string(actual.BidType), string(expected.Type))
-	diffOrtbBids(t, fmt.Sprintf("%s.bid", description), actual.Bid, expected.Bid)
+	assert.Equal(t, string(expected.Type), string(actual.BidType), fmt.Sprintf(`%s.type "%s" does not match expected "%s."`, description, string(actual.BidType), string(expected.Type)))
```
guscarreon, line 284: No need for `diffStrings()` anymore

adapters/adapterstest/test_json.go
```diff
@@ -242,19 +249,31 @@ func assertMakeBidsOutput(t *testing.T, filename string, bidderResponse *adapter
...
-		t.Errorf("Bidders cannot return nil HTTP calls. %s was nil.", description)
-		return
+		return fmt.Errorf("Bidders cannot return nil HTTP calls. %s was nil.", description)
+	}
+
+	if expected.Uri != actual.Uri {
+		return fmt.Errorf(`%s.uri "%s" does not match expected "%s."`, description, actual.Uri, expected.Uri)
 	}
 
-	diffStrings(t, fmt.Sprintf("%s.uri", description), actual.Uri, expected.Uri)
```
guscarreon, line 251: No need for `diffStrings()` anymore

adapters/adapterstest/test_json.go
```diff
@@ -242,19 +249,30 @@ func assertMakeBidsOutput(t *testing.T, filename string, bidderResponse *adapter
...
+		expectedHeader, err := json.Marshal(expected.Headers)
+		if err != nil {
+			return fmt.Errorf(`%s expected.Headers could not be marshalled. Error: %s"`, description, err.Error())
+		}
+		if err := diffJson(description, actualHeader, expectedHeader); err != nil {
+			return err
+		}
 	}
-	diffJson(t, description, actual.Body, expected.Body)
+	return diffJson(description, actual.Body, expected.Body)
```
VeronikaSolovei9, line 275: `diffJson` can be fully rerplaced with 
`assert.JSONEq(t, string(expected), string(actualJson))`. Initial simple implementation worked on my local machine right away. It seems to be easier. It was partly replaces in FPD PR, but it looks like some tests still use it.

adapters/adapterstest/test_json.go
```diff
@@ -190,8 +190,15 @@ func assertMakeRequestsOutput(t *testing.T, filename string, actual []*adapters.
...
 	}
-	for i := 0; i < len(actual); i++ {
-		diffHttpRequests(t, fmt.Sprintf("%s: httpRequest[%d]", filename, i), actual[i], &(expected[i].Request))
+
+	for i := 0; i < len(expected); i++ {
+		var err error
+		for j := 0; j < len(actual); j++ {
+			if err = diffHttpRequests(fmt.Sprintf("%s: httpRequest[%d]", filename, i), actual[j], &(expected[i].Request)); err == nil {
+				break
+			}
```
VeronikaSolovei9, line 199: Not sure about this approach. Please correct me if I'm wrong. For example it may not work correctly in case `expected` has 2 same exact requests (IDK if this is the case) and actual has also 2 requests, one is correct and another one is empty(or different). Bacause you start iteration over actual from the beginning - such case will pass. 
I feel like you need to remove matched actual request from collection after it matched. 
I was not able to create a unit tests for this, becuse in most cases we only have 1 expected request. I can try to do this when 33across changes will be merged. 
This test needs a test :)

AlexBVolcy: I think Veronika is correct in bringing up this example that wouldn't work. I'm also not sure if this behavior of 2 same requests is possible, but the solution to remove matched request from list makes sense to me!

Another idea could be to maybe convert ```expected``` and ```actual``` lists to be maps, which could make the comparison of these objects faster in general and solve the issue Veronika brought up.

adapters/adapterstest/test_json.go
guscarreon: Did it work? I tried `assert.JSONEq(t, string(expected), string(actualJson))` and got many test failures. Where did you substitute it? In which of its calls?

adapters/adapterstest/test_json.go
guscarreon: @AlexBVolcy Maps instead of a nested for loop was my first choice but I tried `map[string]*RequestData` but couldn't get the keys to match. Maybe I need to try again because I definitely like the hash table approach better.

guscarreon: > Not sure about this approach. Please correct me if I'm wrong. For example it may not work correctly in case `expected` has 2 same exact requests (IDK if this is the case) and actual has also 2 requests, first one is correct and second one is empty(or different/incorrect). 

You are right, this approach was implemented under the assumption that the `expected` entries at least differ in the `root.imp[i].id` fields but a corner case should be considered where this is not true. I'll give the remove matched `actual` request approach a try.

adapters/adapterstest/test_json.go
VeronikaSolovei9: Here is my implementation (replace in code there 2 functions)

```

func diffBids(t *testing.T, description string, actual *adapters.TypedBid, expected *expectedBid) {
	if actual == nil {
		t.Errorf("Bidders cannot return nil TypedBids. %s was nil.", description)
		return
	}

	assert.Equal(t, string(expected.Type), string(actual.BidType), fmt.Sprintf(`%s.type "%s" does not match expected "%s."`, description, string(actual.BidType), string(expected.Type)))
	assert.NoError(t, diffOrtbBids(t, fmt.Sprintf("%s.bid", description), actual.Bid, expected.Bid))
}

// diffOrtbBids compares the actual Bid made by the adapter to the expectation from the JSON file.
func diffOrtbBids(t *testing.T, description string, actual *openrtb2.Bid, expected json.RawMessage) error {
	if actual == nil {
		return fmt.Errorf("Bidders cannot return nil Bids. %s was nil.", description)
	}

	actualJson, err := json.Marshal(actual)
	if err != nil {
		return fmt.Errorf("%s failed to marshal actual Bid into JSON. %v", description, err)
	}

	//return diffJson(description, actualJson, expected)
	assert.JSONEq(t, string(expected), string(actualJson), "message123")
	return nil
}


```

VeronikaSolovei9: Oh Gus wait, I tried to replace all of them and it fails. Let me see if I can fix it

VeronikaSolovei9: As discussed offline, this will be done in another PR. Thank you!

curlyblueeagle:
Added test looks good

VeronikaSolovei9 (approved):

AlexBVolcy (approved):

PR #2092, opened 2021-11-24
Unruly: changes to the correct user sync url

opeledtremor:
there was a mistake in the url for the user sync last time here is a fix for it to the correct one


guscarreon (approved):
LGTM

mansinahar (approved):

PR #2096, opened 2021-11-26
Algorix: add Server Region Support

Bugxyb:
Add Server Region Support:
* add region in ext.bidder: APAC for request to SG EP, USE for request to US EP, other for request to Global EP


static/bidder-params/algorix.json
```diff
@@ -21,6 +21,10 @@
     "appId": {
       "type": "string",
       "description": "An ID which identifies this app of the impression"
+    },
+    "region": {
```
SyntaxNode, line 25: This is not the intent of bidder parameters and creates a poor experience for publishers.

Instead, bidders are expected to work with Prebid Server hosts to configure optimal regional data center overrides. You can reach out to hosts directly and/or add regional overrides in your [bidder docs](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/algorix.md).

Bugxyb: Yes, I know.
We have two server region in SG and US, also we provide global endpoint to choose which server region receive, but it depends on the publisher's server.
Our publisher only have US region. when their send KR/JP traffic, it will send to US server region and spend US server budget, but it better send to our SG server region, so we need better provide server region options and make publishers choose the traffic budget region.
we add this param is optional params in [bidder docs](https://github.com/prebid/prebid.github.io/pull/3430/files). publisher also can ignored this param(use global EP), when their have this issue(US Server region spend KR/JP or Southeast Asia traffic), their can add server region optional param to choose send this traffic to our SG server region.

mansinahar: @Bugxyb I am not sure I follow why the publisher needs to choose the region the request gets sent to. Why can't they just use the global endpoint that routes the request further based on the region? Or why can't the Prebid Server host companies configure regional overrides for your adapter so that requests coming from the US region go to your US region specific endpoint and requests coming from Asia Pacific region can go to your Asia Pacific specific endpoint ?

Bugxyb: @mansinahar our publisher partner not use PBS Host companies server, use they own PBS host which only in one region. and they have global traffic send to US Region, but we want those Asia Pacific traffic to our Asia pacific region.

1. our publisher partner has their own Prebid Server host, not all publisher parter use global Prebid Server host, and the host only located in US Region. When Global request send US Prebid Server Host, and then us region send those request to our  global EP, but only our US region server can deal it(because it us host send to global EP, only us region choose). E.g if one partner build their own Prebid Server Host in Southeast Asia, but their want to spend us traffic, so their traffic send to our SG region Server(Southeast Asia is nearby SG region Server). So if they want to spend us budget with us traffic, only two ways: build US Prebid Server Host or change their host to send our us EP.

2. this region _optional params_ is support their own Prebid Server Host Publisher. if their use global Prebid Server Host, they can ignored this params. and if their have only one region Server, they can choose which server region budget spend.

if it is not recommended to modify, plz change this code to draft or failed. we will rollback PBS-Java(it is merged) and cancel bid docs PR. then if publisher partner which their use own Prebid Server Host only in one Region, we will talk with their to change request region in our server or their server.

mansinahar: @Bugxyb Thanks for the clarification. We will bring this up in this week's Prebid.org committee meeting and get back to you.

mansinahar: We discussed this in the Prebid.org meeting last Friday and we will start reviewing this. Thanks for your patience @Bugxyb.

static/bidder-params/algorix.json
```diff
@@ -21,6 +21,10 @@
...
+      "type": "string",
```
mansinahar, line 26: Would it be better to use an enum type here, like:
```
"enum": ["APAC", "USE"],
```
So that you can fail fast in case of an invalid region being passed in

mansinahar (approved):
LGTM!

bsardo (approved):
LGTM

PR #2097, opened 2021-11-26
New Adapter: apacdex and Remove Adapter: valueimpression

thuyhq:
Similar to the adapter on Prebid.js, we want to use apacdex as our primary adapter and quantumdex, valueimpression will become the alias.
So I renamed the valueimpression adapter to apacdex and updated a few things:
Update adapter endpoint.
Update usersync url and support both iframe and redirect.
Support bidder params: placementId, floorPrice


static/bidder-info/apacdex.yaml
```diff
@@ -0,0 +1,20 @@
+maintainer:
+  email: "ken@apacdex.com"
```
SyntaxNode, line 2: This must be a group email address. Please review our [adapter dev docs](https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html).

openrtb_ext/bidders.go
```diff
@@ -199,7 +200,6 @@ const (
 	BidderUcfunnel          BidderName = "ucfunnel"
 	BidderUnicorn           BidderName = "unicorn"
 	BidderUnruly            BidderName = "unruly"
-	BidderValueImpression   BidderName = "valueimpression"
```
mansinahar, line 202: You still need the valueimpression entries in this file for the alias.

You'd also need the `static/bidder-info/valueimpression.yaml` and `static/bidder-params/valueimpression.json` files for the alias.

thuyhq:
Hello @mansinahar,
Thank you for review.
I made the change with the alias Valueimpression and did the same thing so that Quantumdex became the alias of the Apacdex adapter.

adapters/apacdex/apacdex.go
```diff
@@ -0,0 +1,181 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type apacdexAdapter struct {
```
VeronikaSolovei9, line 15: Please rename this to `adapter`. There is no need to repeat package name.

adapters/apacdex/apacdex.go
```diff
@@ -0,0 +1,181 @@
...
+			result = imp
+			found = true
+			break
+		}
+	}
+	if found {
+		return result, nil
+	}
+	return result, fmt.Errorf("not found impression matched with ImpID=%s", impID)
+}
```
VeronikaSolovei9, line 181: Can be simplified to something like this:
```
func getImpressionForBid(imps []openrtb2.Imp, impID string) (openrtb2.Imp, error) {
	for _, imp := range imps {
		if imp.ID == impID {
			return imp, nil
		}
	}
	return nil, fmt.Errorf("not found impression matched with ImpID=%s", impID)
}
```

config/config.go
```diff
@@ -894,7 +896,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.ucfunnel.endpoint", "https://pbs.aralego.com/prebid")
 	v.SetDefault("adapters.unicorn.endpoint", "https://ds.uncn.jp/pb/0/bid.json")
 	v.SetDefault("adapters.unruly.endpoint", "https://targeting.unrulymedia.com/unruly_prebid_server")
-	v.SetDefault("adapters.valueimpression.endpoint", "https://rtb.valueimpression.com/endpoint")
+	v.SetDefault("adapters.valueimpression.endpoint", "http://useast.quantumdex.io/auction/pbs")
```
VeronikaSolovei9, line 899: Looks good!

VeronikaSolovei9 (commented):
Overall LGTM, can see requests to all 3 adapters in logs. Added 2 minor comments

thuyhq:
Hi @VeronikaSolovei9,
Thank you for review!

VeronikaSolovei9 (dismissed):
Thank you for addressing comments, looks good

adapters/apacdex/apacdex.go
```diff
@@ -84,74 +94,81 @@ func preprocess(request *openrtb2.BidRequest) error {
...
+				})
+				continue
 			}
 
-			if foundMatchingBid {
-				rv.Bids = append(rv.Bids, &adapters.TypedBid{
-					Bid:     &bid,
-					BidType: bidType,
-				})
+			bidType := openrtb_ext.BidTypeBanner
```
mansinahar, line 142: I know this logic is similar to what you had before but is just extended for other media types but I wanted to ask if there's any indication of the media type in the bid that gets sent back from your server and if we can use that here rather than depending on the imp media type. The issue with the imp media type is that if the imp is a multi-format imp that is say it has both video and banner objects then this logic will always set the media type as banner when the bid returned by your server could potentially be a video

config/config.go
mansinahar, line 899: Just to confirm that using `http` instead of `https` is intentional

openrtb_ext/bidders.go
```diff
@@ -103,6 +103,7 @@ const (
 	BidderAJA               BidderName = "aja"
 	BidderAlgorix           BidderName = "algorix"
 	BidderAMX               BidderName = "amx"
+	BidderApacdex           BidderName = "apacdex"
```
mansinahar, line 106: You need an entry for ValueImpression in this file as well for the alias to work

adapters/apacdex/apacdex.go
```diff
@@ -84,74 +94,81 @@ func preprocess(request *openrtb2.BidRequest) error {
...
-					if imp.Banner != nil {
-						bidType = openrtb_ext.BidTypeBanner
-					} else if imp.Video != nil {
-						bidType = openrtb_ext.BidTypeVideo
-					}
-					break
-				}
+			imp, err := getImpressionForBid(request.Imp, bid.ImpID)
+			if err != nil {
+				errors = append(errors, &errortypes.BadServerResponse{
```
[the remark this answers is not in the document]
thuyhq: Fixed it for you

config/config.go
thuyhq: Our endpoint supports both http and https. but we want to use http to save time initiating secure connection

openrtb_ext/bidders.go
thuyhq: I added this before
https://github.com/prebid/prebid-server/blob/01a1a4a58553e88e605052f476c190c090e5ab77/openrtb_ext/bidders.go#L177
https://github.com/prebid/prebid-server/blob/01a1a4a58553e88e605052f476c190c090e5ab77/openrtb_ext/bidders.go#L204

adapters/apacdex/apacdex.go
thuyhq: We currently do not support multiple formats in one impression.
We don't include media types in our bid responses, but we also identify them by the value of the impression media type.
If you want I will modify my endpoints to send them to you via bidresponse.seatbid.bid[i].ext.prebid.type

I fixed the logic to not set the default Bid type to Banner.

VeronikaSolovei9 (approved):
LGTM

openrtb_ext/bidders.go
mansinahar: Ah! You're right. My bad. Valueimpression already existed 😅

adapters/apacdex/apacdex.go
mansinahar: @thuyhq I think having your endpoint send the bid type in the bidresponse.seatbid.bid[i].ext.prebid.type is the best way to go so that in the future even if you change your endpoint to support multiple formats, this logic will still work as expected. Let us know when you've updated that on your end and can update your adapter too

mansinahar:
@thuyhq Not sure if you noticed my response on the above comment or not. Just wanna bring your attention to that. Let us know if you'll be taking care of that as part of this PR or a separate PR in the future. Thanks!

thuyhq:
Hi @mansinahar,
Sorry for the late reply.
It took me some time to update my endpoint side.
Thank you for keeping an eye on.

adapters/apacdex/apacdex.go
thuyhq: I have included the mediaType in the bid response from our endpoint and parsed the bidType from this data at the adapter.
Thank you.

mansinahar:
> Hi @mansinahar, Sorry for the late reply. It took me some time to update my endpoint side. Thank you for keeping an eye on.

No worries @thuyhq! I am looking at the changes you made and I noticed that you're trying to read the media type from `bidresponse.seatbid.bid[i].ext.pbs.mediatype`. Can I ask why can't se just used the already existing `bidresponse.seatbid.bid[i].ext.prebid.type`?

You can then just do something like: https://github.com/prebid/prebid-server/blob/master/adapters/adf/adf.go#L143

thuyhq:
Hi @mansinahar,
I'm just doing the same as the appnexus adapter
https://github.com/prebid/prebid-server/blob/a60dfe22f78ef2f702cac17a2879b77b42982521/adapters/appnexus/appnexus.go#L421
and i think the openrtb_ext.ExtBid extension is only for the prebid-serve endpoint bid response.

I edited it according to your suggestion.
Thank you.

adapters/apacdex/apacdex.go
```diff
@@ -139,36 +131,31 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
...
+	if bid.Ext != nil {
+		var bidExt openrtb_ext.ExtBid
+		err := json.Unmarshal(bid.Ext, &bidExt)
+		if err == nil && bidExt.Prebid != nil {
+			return openrtb_ext.ParseBidType(string(bidExt.Prebid.Type))
+		}
+	}
+
+	return "", &errortypes.BadServerResponse{
+		Message: fmt.Sprintf("Failed to parse impression \"%s\" mediatype", bid.ImpID),
```
mansinahar, line 159: Can you please rewrite this error message as:
```
fmt.Sprintf("Failed to parse bid mediatype for impression \"%s\"", bid.ImpID),
```

mansinahar (commented):
LGTM other than the minor comment below

adapters/apacdex/apacdex.go
thuyhq: Hi @mansinahar,
Hope you have a great weekend.
I have updated the error message.
Thank you.

mansinahar (approved):

VeronikaSolovei9 (approved):

PR #2102, opened 2021-12-03
New Adapter: Compass

CompassSSP:
doc - https://github.com/prebid/prebid.github.io/pull/3441

### DIFF
--- a/adapters/compass/compass.go
+++ b/adapters/compass/compass.go
@@ -69,21 +69,22 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 
 		reqCopy.Imp[0].Ext = finalyImpExt
 
-		adapterReq, errors := a.makeRequest(&reqCopy)
+		adapterReq, err := a.makeRequest(&reqCopy)
+		if err != nil {
+			return nil, append(errs, err)
+		}
+
 		if adapterReq != nil {
 			adapterRequests = append(adapterRequests, adapterReq)
 		}
-		errs = append(errs, errors...)
 	}
 	return adapterRequests, errs
 }
 
-func (a *adapter) makeRequest(request *openrtb2.BidRequest) (*adapters.RequestData, []error) {
-	var errs []error
-
+func (a *adapter) makeRequest(request *openrtb2.BidRequest) (*adapters.RequestData, error) {
 	reqJSON, err := json.Marshal(request)
 	if err != nil {
-		return nil, append(errs, err)
+		return nil, err
 	}
 
 	headers := http.Header{}
@@ -94,7 +95,7 @@ func (a *adapter) makeRequest(request *openrtb2.BidRequest) (*adapters.RequestDa
 		Uri:     a.endpoint,
 		Body:    reqJSON,
 		Headers: headers,
-	}, errs
+	}, err
 }
 
 func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.RequestData, responseData *adapters.ResponseData) (*adapters.BidderResponse, []error) {
@@ -139,10 +140,10 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
 			if imp.Banner != nil {
 				return openrtb_ext.BidTypeBanner, nil
 			}
-			if imp.Banner == nil && imp.Video != nil {
+			if imp.Video != nil {
 				return openrtb_ext.BidTypeVideo, nil
 			}
-			if imp.Banner == nil && imp.Video == nil && imp.Native != nil {
+			if imp.Native != nil {
 				return openrtb_ext.BidTypeNative, nil
 			}
 		}
--- a/adapters/compass/compass.go
+++ b/adapters/compass/compass.go
@@ -34,7 +34,6 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
 }
 
 func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
-	var errs []error
 	var err error
 	var adapterRequests []*adapters.RequestData
 
@@ -46,10 +45,10 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 		var compassExt openrtb_ext.ImpExtCompass
 
 		if err = json.Unmarshal(reqCopy.Imp[0].Ext, &bidderExt); err != nil {
-			return nil, append(errs, err)
+			return nil, []error{err}
 		}
 		if err = json.Unmarshal(bidderExt.Bidder, &compassExt); err != nil {
-			return nil, append(errs, err)
+			return nil, []error{err}
 		}
 
 		temp := reqBodyExt{CompassBidderExt: reqBodyExtBidder{}}
@@ -64,21 +63,21 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 
 		finalyImpExt, err := json.Marshal(temp)
 		if err != nil {
-			return nil, append(errs, err)
+			return nil, []error{err}
 		}
 
 		reqCopy.Imp[0].Ext = finalyImpExt
 
 		adapterReq, err := a.makeRequest(&reqCopy)
 		if err != nil {
-			return nil, append(errs, err)
+			return nil, []error{err}
 		}
 
 		if adapterReq != nil {
 			adapterRequests = append(adapterRequests, adapterReq)
 		}
 	}
-	return adapterRequests, errs
+	return adapterRequests, nil
 }
 
 func (a *adapter) makeRequest(request *openrtb2.BidRequest) (*adapters.RequestData, error) {
--- a/adapters/compass/compass.go
+++ b/adapters/compass/compass.go
@@ -16,6 +16,16 @@ type adapter struct {
 	endpoint string
 }
 
+type reqBodyExt struct {
+	CompassBidderExt reqBodyExtBidder `json:"bidder"`
+}
+
+type reqBodyExtBidder struct {
+	Type        string `json:"type"`
+	PlacementID string `json:"placementId,omitempty"`
+	EndpointID  string `json:"endpointId,omitempty"`
+}
+
 func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
 	bidder := &adapter{
 		endpoint: config.Endpoint,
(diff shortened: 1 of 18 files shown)


static/bidder-info/compass.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - native
+    
+  app:
+    mediaTypes:
+    - banner
+    - video
+    - native
+userSync:
+  redirect:
+    url: "https://sa-cs.deliverimp.com/pserver?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redir={{.RedirectURL}}"
```
guscarreon, line 17: I tried to test the redirect URL and got an 'Unable to connect' error:
```
An error occurred during a connection to sa-cs.deliverimp.com. 
```
Are we positive the URL is correct?

adapters/compass/compass.go
```diff
@@ -0,0 +1,147 @@
...
+
+		finalyImpExt := reqCopy.Imp[0].Ext
+		if compassExt.PlacementID != "" {
+			finalyImpExt, _ = json.Marshal(map[string]interface{}{
+				"bidder": map[string]interface{}{
+					"placementId": compassExt.PlacementID,
+					"type":        "publisher",
+				},
+			})
+		} else if compassExt.EndpointID != "" {
```
guscarreon, line 53: From the logic of this `if/else if` statement it seems that we are expecting that either `"placementId"` or `"endpointId"` are non-empty strings but the following two corner cases added to `adapters/compass/params_test.go` in my local branch seem to be able to pass JSON schema validation and would not end in either branch of the aforementioned `if/else if` statement. Test cases shown below:
```
36   var validParams = []string{
37       `{"placementId": "test"}`,
38       `{"placementId": "1"}`,
39       `{"endpointId": "test"}`,
40       `{"endpointId": "1"}`,
   +     `{"endpointId": ""}`,
   +     `{"placementId": ""}`,
41   }
adapters/compass/params_test.go
```
In case these scenarios are not intended to skip the branches of the `if/else` statment, we can prevent them from passing validation by adding a `minLength` fields in the JSON schema file `static/bidder-params/compass.json`:
```
 1   {
 2       "$schema": "http://json-schema.org/draft-04/schema#",
 3       "title": "Compass Adapter Params",
 4       "description": "A schema which validates params accepted by the Compass adapter",
 5       "type": "object",
 6  
 7       "properties": {
 8         "placementId": {
 9           "type": "string",
   +         "minLength": 1,
10           "description": "Placement ID"
11         },
12         "endpointId": {
13           "type": "string",
   +         "minLength": 1,
14           "description": "Endpoint ID"
15         }
16       },
17       "oneOf": [
18         { "required": ["placementId"] },
19         { "required": ["endpointId"] }
20       ]
21     }
static/bidder-params/compass.json
```
Do you believe these `"minLength"` fields should be added?

adapters/compass/compass.go
```diff
@@ -0,0 +1,147 @@
...
+		if err = json.Unmarshal(reqCopy.Imp[0].Ext, &bidderExt); err != nil {
+			return nil, append(errs, err)
+		}
+		if err = json.Unmarshal(bidderExt.Bidder, &compassExt); err != nil {
+			return nil, append(errs, err)
+		}
+
+		finalyImpExt := reqCopy.Imp[0].Ext
+		if compassExt.PlacementID != "" {
+			finalyImpExt, _ = json.Marshal(map[string]interface{}{
```
guscarreon, line 47: A `json.Marshal` error seems far-fetched but, can we please still add a check?
```
45         finalyImpExt := reqCopy.Imp[0].Ext
46         if compassExt.PlacementID != "" {
47 -           finalyImpExt, _ = json.Marshal(map[string]interface{}{
   +           finalyImpExt, err = json.Marshal(map[string]interface{}{
48                 "bidder": map[string]interface{}{
49                     "placementId": compassExt.PlacementID,
50                     "type":        "publisher",
51                 },
52             })
   +           if err != nil {
   +               // handle error
   +           }
53         } else if compassExt.EndpointID != "" {
54 -           finalyImpExt, _ = json.Marshal(map[string]interface{}{
   +           finalyImpExt, err = json.Marshal(map[string]interface{}{
55                 "bidder": map[string]interface{}{
56                     "endpointId": compassExt.EndpointID,
57                     "type":       "network",
58                 },
59             })
   +           if err != nil {
   +               // handle error
   +           }
60         }
61
adapters/compass/compass.go
```

adapters/compass/compass.go
```diff
@@ -0,0 +1,147 @@
...
+				"bidder": map[string]interface{}{
```
guscarreon, line 48: Any reason we decided to use a `map[string]interface{}` instead of a `map[string]string`?

CompassSSP:
@guscarreon thanks for the comments, everything has been corrected

static/bidder-info/compass.yaml
CompassSSP: Yes, we are sure the URL is correct. We are currently configuring the server.

CompassSSP:
@guscarreon Thanks! We have fixed

guscarreon (approved):
Looks good to me. @CompassSSP thanks for addressing my feedback

static/bidder-info/compass.yaml
```diff
@@ -0,0 +1,18 @@
+maintainer:
+  email: "sa-support@brightcom.com"
+capabilities:
```
SyntaxNode, line 3: Do you have a GVL ID registered with the IAB? I see vendor 943 listed as "Marfeel Solutions, SL (Compass)".

CompassSSP: Not 943, Client’s Vendor ID is 883

SyntaxNode: Please include id 883 in this configuration file. See https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#bidder-info for details.

adapters/compass/compass.go
```diff
@@ -0,0 +1,154 @@
...
+	var adapterRequests []*adapters.RequestData
+
+	reqCopy := *request
+	for _, imp := range request.Imp {
+		reqCopy.Imp = []openrtb2.Imp{imp}
+
+		var bidderExt adapters.ExtImpBidder
+		var compassExt openrtb_ext.ImpExtCompass
+
+		if err = json.Unmarshal(reqCopy.Imp[0].Ext, &bidderExt); err != nil {
```
SyntaxNode, line 48: Why is this unmarshalling the first imp's ext instead of the current imp's ext?

adapters/compass/compass.go
```diff
@@ -0,0 +1,154 @@
...
+		adapterReq, errors := a.makeRequest(&reqCopy)
+		if adapterReq != nil {
+			adapterRequests = append(adapterRequests, adapterReq)
+		}
+		errs = append(errs, errors...)
+	}
+	return adapterRequests, errs
+}
+
+func (a *adapter) makeRequest(request *openrtb2.BidRequest) (*adapters.RequestData, []error) {
```
SyntaxNode, line 81: Why does `makeRequest` return `[]error` instead of `error` since there is only one possible error?

adapters/compass/compass.go
```diff
@@ -0,0 +1,154 @@
...
+	return bidResponse, nil
+}
+
+func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType, error) {
+	for _, imp := range imps {
+		if imp.ID == impID {
+			if imp.Banner != nil {
+				return openrtb_ext.BidTypeBanner, nil
+			}
+			if imp.Banner == nil && imp.Video != nil {
```
SyntaxNode, line 142: Can be simplified to `if imp.Video != nil {` since `imp.Banner == nil` based on the previous `if imp.Banner != nil {` conditional.

adapters/compass/compass.go
```diff
@@ -0,0 +1,154 @@
...
+				return openrtb_ext.BidTypeVideo, nil
+			}
+			if imp.Banner == nil && imp.Video == nil && imp.Native != nil {
```
SyntaxNode, line 145: Similarly, can be simplified to `if imp.Native != nil {`.

adapters/compass/compass.go
```diff
@@ -0,0 +1,154 @@
...
+				return openrtb_ext.BidTypeNative, nil
+			}
+		}
+	}
```
SyntaxNode, line 149: The approach used here to identify the media type for the bid is an anti-pattern. Is there a better way for you to make this determination? If not, please be sure the order matches your bidding server, such that if a banner format is specified  your bidding server will only ever respond with banner bids.

adapters/compass/compass.go
CompassSSP: We are not unmarshalling the bidRequest first imp ext, but the reqCopy of the first imp. Please check line 43

adapters/compass/compass.go
CompassSSP: Yes, I be sure the order matches bidding server.

CompassSSP:
@SyntaxNode thanks for the review. I fixed some issues and included 883 in the config file

adapters/compass/compass.go
SyntaxNode: Ah. You're right. Subtle. Ok.

adapters/compass/compass.go
```diff
@@ -0,0 +1,155 @@
...
+
+func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
+	bidder := &adapter{
+		endpoint: config.Endpoint,
+	}
+	return bidder, nil
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errs []error
```
SyntaxNode, line 37: Nitpick: This is no longer required with the last fix refactor you pushed. You can replace all `return nil, append(errs, err)` statements with `return nil, []error{err}` like you do in `MakeBids`. The last return of this method will become `return adapterRequests, nil` which is a bit more straightforward IMHO.

SyntaxNode (commented):
Thank you for quickly addressing the feedback. The last commit you pushed opened up a small refactor which I've commented on. Otherwise looks good and I'm glad to see good test coverage using the json test framework.

AlexBVolcy (approved):
Thank you to responding to all the feedback, I'll approve, but there's one nitpick left that Scott commented on that I think should be fixed before merge!

adapters/compass/compass.go
CompassSSP: thanks, refactored

SyntaxNode (approved):

AlexBVolcy (approved):

Issue #2088, opened 2021-11-17
PBS needs to support imp.ext.gpid


bretg:
Related to https://github.com/prebid/Prebid.js/issues/7367

The OpenRTB location for GPID is imp.ext.gpid. Prebid Server filters imp.ext for valid bidders except for a list of "reserved" strings: all, context, data, general, prebid, or skadn

We need to add "gpid" to the list of reserved words ASAP.

SyntaxNode:
Shouldn't be a problem. Could you please point me to the spec for the imp.ext.gpid extension? Is this a standard IAB thing?

bretg:
It's not yet a formal IAB extension, but several of our publishers have received notice from TradeDesk that this parameter is soon to be required.

bretg:
Here's the prebid.org doc -- https://docs.prebid.org/features/pbAdSlot.html#the-gpid

bretg:
Released with PBS-Java 1.80

SyntaxNode:
Implemented in PBS-Go 0.188.0.

PR #2107, opened 2021-12-09
Add GPID Reserved Bidder Name

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/2088

I needed to refactor existing code for handling bidderparams due to a parsing problem it introduced. That code failed if any field in `imp.ext` is not an embedded object type, which is a problem since `imp.ext.gpid` is a string. Since this is code is upstream from validation (which makes sense as it could cause an invalid request to become valid) it should not "step on the toes" of the validator. If there is a problem with the request, we should let the validate method continue to handle validation issues.


adapters/openrtb_util.go
```diff
@@ -4,11 +4,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+
 	"github.com/mxmCherry/openrtb/v15/openrtb2"
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
-func ExtractAdapterReqBidderParams(bidRequest *openrtb2.BidRequest) (map[string]json.RawMessage, error) {
+func ExtractReqExtBidderParamsMap(bidRequest *openrtb2.BidRequest) (map[string]json.RawMessage, error) {
```
SyntaxNode, line 12: Provided a more precise name to better align with the other method in this file.

adapters/openrtb_util_test.go
```diff
@@ -2,116 +2,100 @@ package adapters
...
 	"testing"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/stretchr/testify/assert"
 )
 
-func TestExtractAdapterReqBidderParams(t *testing.T) {
-	type args struct {
-		bidRequest *openrtb2.BidRequest
-	}
```
SyntaxNode, line 13: Refactors to remove the args object, since it only contains a single argument I don't see the need for a separate structure.

adapters/openrtb_util_test.go
```diff
@@ -2,116 +2,100 @@ package adapters
...
-			name: "extract bidder params from nil req",
-			args: args{
-				request: nil,
-			},
-			want:    nil,
-			wantErr: true,
+			name:            "nil req",
+			givenBidRequest: nil,
+			want:            nil,
+			wantErr:         errors.New("error bidRequest should not be nil"),
```
SyntaxNode, line 64: It's better to compare the actual error message rather than verify the count of errors.

endpoints/openrtb2/auction.go
```diff
@@ -336,111 +336,136 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
-	reqBidderParams, err := adapters.ExtractReqExtBidderParams(req.BidRequest)
+	reqBidderParams, err := adapters.ExtractReqExtBidderParamsEmbeddedMap(req.BidRequest)
 	if err != nil {
-		return err
+		return nil
+	}
+
+	if len(reqBidderParams) == 0 {
+		return nil
 	}
```
SyntaxNode, line 346: Optimization to avoid the expensive operations of the loop if there are no bidderparams to apply.

endpoints/openrtb2/auction.go
```diff
@@ -336,111 +336,136 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
+		impExtPrebidModified, err := addMissingReqExtParamsInImpExtPrebid(impExt, reqBidderParams)
 		if err != nil {
-			return err
+			return fmt.Errorf("error processing bidder parameters for imp[%d]: %s", impIndex, err.Error())
 		}
 
-		iExt, err := json.Marshal(impExt)
-		if err != nil {
-			return fmt.Errorf("error marshalling imp[].ext : %s", err.Error())
+		if impExtModified || impExtPrebidModified {
```
SyntaxNode, line 377: Slight optimization to avoid marshaling if no changes were made, such as the edge case when all values are more specifically defined in imp.ext.

endpoints/openrtb2/auction.go
```diff
@@ -336,111 +336,136 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
-			return err
+			return anyModified, err
 		}
-		impExtBidder["prebid"]["bidder"] = bidderParamsJson
+		impExtByKey["prebid"] = prebidJSON
 	}
 
-	return nil
+	return anyModified, err
 }
```
SyntaxNode, line 434: Refactored to be more lenient on parsing errors and refactored to re-use addMissingReqExtParamsInImpExt instead of reimplementing a similar flow.

endpoints/openrtb2/auction.go
```diff
@@ -1173,12 +1198,12 @@ func isBidderToValidate(bidder string) bool {
...
 	case openrtb_ext.BidderReservedData:
 		return false
+	case openrtb_ext.BidderReservedGPID:
+		return false
 	case openrtb_ext.BidderReservedPrebid:
 		return false
 	case openrtb_ext.BidderReservedSKAdN:
 		return false
-	case openrtb_ext.BidderReservedBidder:
-		return false
```
SyntaxNode, line 1181: The bidderparams feature does not make special use of "bidder", so I don't see why it needed to add this reserved bidder name. Perhaps it did in an earlier version? Removed.

endpoints/openrtb2/auction_test.go
```diff
@@ -1146,6 +1146,301 @@ func TestStoredRequests(t *testing.T) {
 	}
 }
 
+func TestMergeBidderParams(t *testing.T) {
```
SyntaxNode, line 1149: Added extensive unit tests for the merge methods. I left the json based integration tests alone.

endpoints/openrtb2/sample-requests/invalid-whole/cache-nothing.json
```diff
@@ -27,5 +27,5 @@
     }
   },
   "expectedReturnCode": 400,
-  "expectedErrorMessage": "Invalid request: error decoding Request.ext : request.ext.prebid.cache requires one of the \"bids\" or \"vastxml\" properties\n"
+  "expectedErrorMessage": "Invalid request: request.ext is invalid: request.ext.prebid.cache requires one of the \"bids\" or \"vastxml\" properties\n"
```
SyntaxNode, line 30: Reverted. The bidderparams feature should not alter the validation flow.

endpoints/openrtb2/sample-requests/invalid-whole/imp-ext-invalid-params.json
```diff
@@ -20,5 +20,5 @@
     }
   },
   "expectedReturnCode": 400,
-  "expectedErrorMessage": "Invalid request: json: cannot unmarshal string into Go value of type map[string]json.RawMessage"
+  "expectedErrorMessage": "Invalid request: request.imp[0].ext.appnexus failed validation.\n(root): Invalid type. Expected: object, given: string\n"
```
SyntaxNode, line 23: Reverted. The bidderparams feature should not alter the validation flow.

openrtb_ext/bidders_test.go
```diff
@@ -104,15 +104,15 @@ func TestIsBidderNameReserved(t *testing.T) {
 		{"general", true},
 		{"gEnErAl", true},
 		{"GENERAL", true},
-		{"skadn", true},
-		{"skADN", true},
-		{"SKADN", true},
```
SyntaxNode, line 109: Ordered alphabetically.

endpoints/openrtb2/auction_test.go
```diff
@@ -4048,51 +4343,33 @@ func TestValidateBanner(t *testing.T) {
...
-			errL:           0,
-		},
-		{
-			name: "add missing bidder-params from req.ext.prebid.bidderparams to imp[].ext.prebid.bidder for reserved keyword",
-			args: args{
-				reqBody: validRequest(t, "req-ext-bidder-params-merge-reserved-keyword.json"),
-			},
-			expectedImpExt: getObject(t, "req-ext-bidder-params-merge-reserved-keyword.json", "expectedImpExt"),
-			expectedReqExt: getObject(t, "req-ext-bidder-params-merge-reserved-keyword.json", "expectedReqExt"),
-			errL:           0,
```
SyntaxNode, line 4095: Removed. I don't understand the use case of a reserved bidder keyword and I didn't see mention of special handling in the issue. Please correct me if I missed something.

endpoints/openrtb2/auction.go
```diff
@@ -336,111 +336,136 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
-		err = addMissingReqExtParamsInImpExt(impExt, reqBidderParams)
+		impExtModified, err := addMissingReqExtParamsInImpExt(impExt, reqBidderParams)
 		if err != nil {
-			return err
+			return fmt.Errorf("error processing bidder parameters for imp[%d]: %s", impIndex, err.Error())
 		}
 
 		//merges bidder parameters passed at req.ext level with imp[].ext.prebid.bidder level.
-		err = addMissingReqExtParamsInImpExtPrebid(impExt, reqBidderParams)
+		impExtPrebidModified, err := addMissingReqExtParamsInImpExtPrebid(impExt, reqBidderParams)
```
guscarreon, line 372: Given that `request.imp.ext.prebid.bidder.BIDDER` takes precedence over `request.imp.ext.BIDDER`, does this become an "either-or" situation? If a given `imp.ext` comes with a `prebid` field, should we only call `addMissingReqExtParamsInImpExtPrebid` and not both `addMissingReqExtParamsInImpExtPrebid` and `addMissingReqExtParamsInImpExt`? If the `imp.ext`s that come with a `prebid` field will ignore `request.imp.ext.BIDDER` values anyways, maybe it'd be a desirable optimization to not copy them to `request.imp.ext.BIDDER` at all.
```
1124 func (deps *endpointDeps) validateImpExt(imp *openrtb2.Imp, aliases map[string]string, impIndex int) []error {
1125     errL := []error{}
            .
            .
1134
1135     // Prefer bidder params from request.imp.ext.prebid.bidder.BIDDER over request.imp.ext.BIDDER
1136     // to avoid confusion beteween prebid specific adapter config and other ext protocols.
1137     if extPrebidJSON, ok := bidderExts[openrtb_ext.PrebidExtKey]; ok {
1138         var extPrebid openrtb_ext.ExtImpPrebid
1139         if err := json.Unmarshal(extPrebidJSON, &extPrebid); err == nil && extPrebid.Bidder != nil {
1140             for bidder, ext := range extPrebid.Bidder {
1141                 if ext == nil {
1142                     continue
1143                 }
1144                 bidderExts[bidder] = ext
1145             }
1146         }
1147     }
1148
1149     /* Process all the bidder exts in the request */
endpoints/openrtb2/auction.go
```
If implemented, the change would make `mergeBidderParams` behave differently, and a test case like the one shown below would expect a different outcome:
```
1149   func TestMergeBidderParams(t *testing.T) {
1150       testCases := []struct {
1151           description         string
1152           givenRequest        openrtb2.BidRequest
1153           expectedRequestImps []openrtb2.Imp
1154       }{
1155   *- 55 lines: {-------------------------------------------------------------------------------
1210           {
1211 -             description: "One Imp - imp.ext + imp.ext.prebid.bidder Modified - Same Bidder",
     +             description: "One Imp - modify imp.ext.prebid.bidder ONLY - Same Bidder",
1212               givenRequest: openrtb2.BidRequest{
1213                   Imp: []openrtb2.Imp{{ID: "1", Ext: json.RawMessage(`{"bidder1":{"a":1},"prebid":{"bidder":{"bidder1":{"a":"one"}}}}`)}},
1214                   Ext: json.RawMessage(`{"prebid":{"bidderparams":{"bidder1":{"b":2}}}}`),
1215               },
1216 -             expectedRequestImps: []openrtb2.Imp{{ID: "1", Ext: json.RawMessage(`{"bidder1":{"a":1,"b":2},"prebid":{"bidder":{"bidder1":{"a":"one","b":2}}}}`)}},
     +             expectedRequestImps: []openrtb2.Imp{{ID: "1", Ext: json.RawMessage(`{"bidder1":{"a":1},"prebid":{"bidder":{"bidder1":{"a":"one","b":2}}}}`)}},
1217           },
endpoints/openrtb2/auction_test.go
```
Unless I'm missing something and those values are still needed in `request.imp.ext.BIDDER`. Thoughts?

guscarreon (commented):
Left some questions below

endpoints/openrtb2/auction.go
SyntaxNode: I have no preference. I can combine them if you prefer.

endpoints/openrtb2/auction.go
```diff
@@ -336,111 +336,136 @@ func parseTimeout(requestJson []byte, defaultTimeout time.Duration) time.Duratio
...
 		if !isBidderToValidate(bidder) {
 			continue
 		}
 
-		wasModified := false
+		var params map[string]json.RawMessage
+		if err := json.Unmarshal(bidderExt, &params); err != nil {
+			// bidder should be parsable, but we haven't validated yet so there may be another issue. ignore it here
+			// and let the validation later on catch it.
+			continue
```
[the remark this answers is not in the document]
SyntaxNode: > Are we doing this to filter out the malformed imp[].ext fields and salvage the good ones?

No. There is no validation intended or performed by the merge method. The `map[string]map[string]json.RawMessage` is not a valid data structure for `imp.ext` since it forces all fields to be an object or fails to parse. This PR adds support for `imp.ext.gpid` which is a string. That's the reason for the refactor.

> should go back to unmarshaling the entire thing in mergeBidderParams(req *openrtb_ext.RequestWrapper)?

We cannot. If we do, then `imp.ext.gpid` will always cause a validation error in `mergeBidderParams`.

endpoints/openrtb2/auction.go
SyntaxNode: > Given that request.imp.ext.prebid.bidder.BIDDER takes precedence over request.imp.ext.BIDDER, does this become an "either-or" situation?

Yes. Later code will pick one or the other, but not both. Referencing the precedence logic here would increase it's complexity for IMHO minimal processing savings.

I'm not sure why it was designed this way, but separating the merging pf params from the decision on which params to use seems like a good isolation of responsibilities.

This is pre-existing behavior. The reason I didn't refactor that in this PR is that it's not required for `imp.ext.gpid` to function, which is the goal of this PR.

endpoints/openrtb2/auction.go
guscarreon: that's fine. No need to hold this PR over this.

endpoints/openrtb2/auction.go
guscarreon: makes sense. It can be refactored separately

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
guscarreon: I see

guscarreon (approved):
LGTM

mansinahar (approved):

Issue #2080, opened 2021-11-10
SChain OpenRTB Backwards Compatibility


SyntaxNode:
# Problem

The location of the SChain object was changed in OpenRTB 2.5 from previous versions. PBS (at least PBS-Go) only supports the OpenRTB 2.5 location since that's the version of the spec PBS implements. Here's the relevant quote from the [SChain Specification](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/supplychainobject.md):

> The SupplyChain object should be included in the BidRequest.Source.ext.schain attribute in OpenRTB 2.5 or later. For OpenRTB 2.4 or prior, the BidRequest.ext.schain attribute should be used.

The lifecycle of extensions was recently discussed in the IAB Programmatic Working Group and it was called out that while the newer locations are preferred, software should also fall back to previous locations. 

# Proposal

Enhance PBS to look for SChain objects at both the current OpenRTB 2.5 `BidRequest.Source.ext.schain` location and fallback to the older `BidRequest.ext.schain` location if not found. If both are defined, The OpenRTB 2.5 location will be used.

bretg:
It's fine to conform to this recommendation, but for the record, it unclear to me that PBS will ever see $.ext.schain -- the PBJS module sets it in $.source.ext.schain, and PBS host companies can control where it's set in stored requests.

bretg:
Discussed in PBS committee. Xandr has a publisher calling PBS directly with the ORTB2.4. 

The proposal is:
if $.source.ext.schain doesn't exist but $.ext.schain does, PBS would copy the latter to the former so that all bidders will see it in the ORTB2.5 location.

bretg:
Done in PBS-Java 1.84

PR #2108, opened 2021-12-09
Add ORTB 2.4 schain support

bsardo:
This PR adds ORTB 2.4 schain support as described in issue #2080.
- If an schain is defined in one of the ORTB 2.5 locations (`.source.ext.schain` or `.ext.prebid.schains`) and in the ORTB 2.4 location (`.ext.schain`), the 2.5 schains take precedence so the ORTB 2.4 schain is ignored.
- If an schain is defined in the ORTB 2.4 location (`.ext.schain`) but not in one of the ORTB 2.5 schain locations, the ORTB 2.4 schain is copied to the ORTB 2.5 location (`.source.ext.schain`) for each bid request.

The PR refactors the schain logic as part of introducing ORTB 2.4 schain support. An `SChainWriter` interface is introduced with concrete ORTB 2.4 and 2.5 implementations. Each concrete type's `Write` method modifies the request `.source.ext.schain` accordingly ensuring that any fields defined in `.source` are copied (this was a recently fixed bug).

It should also be noted that the function `BidderToPrebidSChains` in `schain/schain.go` and its corresponding tests were copied verbatim from `exchange/utils.go`.


exchange/utils_test.go
```diff
@@ -2734,47 +2623,6 @@ func TestBuildXPrebidHeader(t *testing.T) {
 	}
 }
 
-func TestSourceExtSChainCopied(t *testing.T) {
```
bsardo, line 2737: This test was deleted because the corresponding logic was moved into the schain writer implementation `Write` methods which have tests that cover this logic.

exchange/utils.go
```diff
@@ -315,43 +315,10 @@ func getExtJson(req *openrtb2.BidRequest, unpackedExt *openrtb_ext.ExtRequest) (
 
 	extCopy := *unpackedExt
 	extCopy.Prebid.SChains = nil
+	extCopy.SChain = nil
 	return json.Marshal(extCopy)
 }
 
-func prepareSource(req *openrtb2.BidRequest, bidder string, sChainsByBidder map[string]*openrtb_ext.ExtRequestPrebidSChainSChain) {
```
bsardo, line 321: This logic has been moved to the schain writer implementation `Write` methods.

exchange/utils.go
```diff
@@ -31,23 +32,6 @@ var integrationTypeMap = map[metrics.RequestType]config.IntegrationType{
 
 const unknownBidder string = ""
 
-func BidderToPrebidSChains(sChains []*openrtb_ext.ExtRequestPrebidSChain) (map[string]*openrtb_ext.ExtRequestPrebidSChainSChain, error) {
```
bsardo, line 34: This was copied verbatim to `schain/schain.go`.

schain/schainwriter.go
```diff
@@ -0,0 +1,151 @@
...
+// extPrebidSChainExists checks if an schain exists in the ORTB 2.5 req.ext.prebid.schain location
+func extPrebidSChainExists(reqExt *openrtb_ext.ExtRequest) bool {
+	if reqExt == nil {
+		return false
+	}
+	if reqExt.Prebid.SChains == nil {
+		return false
+	}
+	return true
+}
```
bsardo, line 129: This function was written this way instead of a simple one liner of `return reqExt != nil && reqExt.Prebid.SChains != nil` because I think it helps us ensure code coverage of each condition when running the tests with code coverage enabled. The same applies to the two helper functions right after it.
If anyone prefers I use a one liner LMK. I thought the benefits here outweighed the slight increase in LOC.

VeronikaSolovei9: I was going to write a comment about this :)
I only think it's better to check `reqExt.Prebid.SChains != nil`, but I'm also ok with current logic.

VeronikaSolovei9:
While reviewing this PR and reading description at the 100th time, I started to think there is no need for ORTBTwoFourSChainWriter. `If an schain is defined in the ORTB 2.4 location (.ext.schain) but not in one of the ORTB 2.5 schain locations, the ORTB 2.4 schain is copied to the ORTB 2.5 location (.source.ext.schain) for each bid request.` -> sounds there will always be ORTB 2.5. 
Also code in `func (w ORTBTwoFourSChainWriter) Write(req *openrtb2.BidRequest, bidder string) ` duplicates code in  `func (w ORTBTwoFiveSChainWriter) Write(req *openrtb2.BidRequest, bidder string)`. 
Would it be reasonable to have just one SChainWriter?

schain/schainwriter.go
bsardo: I think we also need to check if `reqExt` is nil otherwise we will get an error when attempting to access the `Prebid` field on a nil `reqExt`.

bsardo:
An schain can be defined in the ORTB 2.4 or the ORTB 2.5 location of the original request. Each of the writers is used to set the schain in the ORTB 2.5 location of each bid request. If no ORTB 2.5 schain is defined but an ORTB 2.4 schain is defined, the ORTB 2.4 schain writer is what is instantiated and then used to write the 2.4 schain to the 2.5 location. So yes, there will always be an schain in the 2.5 location of each bid request assuming an schain was defined somewhere in the original request.
I created two separate writers, each of which writes to the 2.5 location, because IMO it is better design (smaller functions, easier to test, more extensible) to separate the 2.4 and 2.5 logic than to mix that logic together. I agree there is a little code duplication in the `Write` function. That could be extracted into a common function.
Also, yes, the bidder isn't used in the 2.4 writer `Write` function; it is there to satisfy the interface contract at this time. If we have some concerns about the fact that the 2.4 writer is called in a loop and just does the same thing for each bidder, the marshalling could happen on instantiation to address any performance concerns.
Thoughts @VeronikaSolovei9 @mansinahar?

openrtb_ext/request_wrapper.go
```diff
@@ -456,6 +511,19 @@ func (re *RequestExt) SetPrebid(prebid *ExtRequestPrebid) {
 	re.prebidDirty = true
 }
 
+func (re *RequestExt) GetSChain() *ExtRequestPrebidSChainSChain {
```
mansinahar, line 514: Can we please add a comment here saying that these Schain functions on the request ext are only for backwards compatibility for openrtb 2.4 and shouldn't really be used for anything else. Only source.ext.schain or reques.ext.prebid.schain should be used?

endpoints/openrtb2/auction_test.go
```diff
@@ -2134,6 +2134,142 @@ func TestSChainInvalid(t *testing.T) {
...
+
+			reqExt, err := reqWrapper.GetRequestExt()
+			if err != nil {
+				assert.Fail(t, "Error getting request ext from wrapper", tt.description)
+			}
+			reqExtSChain := reqExt.GetSChain()
+			if tt.wantReqExtSChain == nil {
+				assert.Nil(t, reqExtSChain, tt.description)
+			} else {
+				assert.Equal(t, tt.wantReqExtSChain, reqExtSChain, tt.description)
```
mansinahar, line 2256: Wouldn't this assert.Equal call take care of the `nil` case as well? Do we need the if-else conditional here? Similar for the source ext below

VeronikaSolovei9 (approved):
Thank you for changing the code, LGTM

VeronikaSolovei9 (approved):

PR #2111, opened 2021-12-14
Integration Type Implementation

AlexBVolcy:
This PR addresses the Integration Type portion of the Integration Type and Channel issue #1428

This PR does the following:

- Reads integration type value from request from the path `request.ext.prebid.integration` 
- Allows account/publisher to have a `DefaultIntegration` value, that will be utilized in the case where an integration type isn't found in request 
- Once integration value is captured/set, it is then set as a query parameter for event URL and vtrack URL

(Note: Integration Type value is not required, nor does there need to be validation of the value per Bret)


endpoints/openrtb2/auction.go
```diff
@@ -309,6 +309,13 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
 
 	lmt.ModifyForIOS(req.BidRequest)
 
+	// Integration Type Handling
+	err = deps.setIntegrationType(req)
+	if err != nil {
+		errs = []error{err}
+		return
+	}
```
bsardo, line 317: Nitpick: I think it's preferable to call the function and check if for the error on the same line here
`if err = deps.setIntegrationType(req); err != nil {`

openrtb_ext/request.go
```diff
@@ -32,6 +32,7 @@ type ExtRequestPrebid struct {
 	Channel              *ExtRequestPrebidChannel  `json:"channel,omitempty"`
 	Data                 *ExtRequestPrebidData     `json:"data,omitempty"`
 	Debug                bool                      `json:"debug,omitempty"`
+	Integration          string                    `json:"integration,omitempty"`
```
bsardo, line 35: If a request contains `.ext.prebid`, `integration` will always be set given that this is a string and not a pointer to a string. In that case, if `integration` is not specified it will be an empty string. Is the plan to just check for an empty string when determining whether to apply the default integration value? Is there a use case where a publisher does not want to set an integration type, in which case they would explicitly set it to empty string? If so, then a pointer to a string would be needed as we would need to detect presence to differentiate between not set and set to empty string.

endpoints/openrtb2/auction.go
```diff
@@ -1795,3 +1802,18 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
 	return false, nil
 }
+
+func (deps *endpointDeps) setIntegrationType(req *openrtb_ext.RequestWrapper) error {
+	reqExt, err := req.GetRequestExt()
+	if err != nil {
+		return err
+	}
+	reqPrebid := reqExt.GetPrebid()
+	if reqPrebid == nil {
```
bsardo, line 1812: What if `reqPrebid` is defined but `integration` was not specified? Shouldn't we apply the default integration type in that case?

endpoints/openrtb2/auction.go
```diff
@@ -1795,3 +1802,18 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
+		reqPrebid = &openrtb_ext.ExtRequestPrebid{Integration: deps.cfg.AccountDefaults.DefaultIntegration}
+		reqExt.SetPrebid(reqPrebid)
+		req.RebuildRequest()
```
bsardo, line 1815: Does it make sense to rebuild the request here? This function is called from `parseRequest` which is called from `Auction` and at the end of `Auction` we rebuild the request.

endpoints/openrtb2/auction_test.go
```diff
@@ -1386,6 +1386,55 @@ func TestValidateOrFillChannel(t *testing.T) {
...
+	}{
+		{
+			description: "Request has integration type defined, expect that same integration type",
+			givenRequestWrapper: &openrtb_ext.RequestWrapper{
+				BidRequest: &openrtb2.BidRequest{Ext: []byte(`{"prebid":{"integration": "TestIntegrationType"}}`)},
+			},
+			expectedIntegrationType: "TestIntegrationType",
+		},
+		{
+			description: "Request doesn't have request.ext.prebid path, expect default integration value",
```
bsardo, line 1423: Can we add a test case for when `prebid` is defined but `integration` is not?

endpoints/events/event.go
```diff
@@ -26,10 +26,11 @@ const (
...
 	// Optional
-	BidderParameter    = "bidder"
-	TimestampParameter = "ts"
-	FormatParameter    = "f"
-	AnalyticsParameter = "x"
+	BidderParameter          = "bidder"
+	TimestampParameter       = "ts"
+	FormatParameter          = "f"
+	AnalyticsParameter       = "x"
+	IntegrationTypeParameter = "int"
```
bsardo, line 33: Just checking, is this a standardized query parameter agreed upon with the PBS committee?

analytics/event.go
```diff
@@ -28,11 +28,12 @@ const (
...
-	Bidder    string         `json:"bidder,omitempty"`
-	Timestamp int64          `json:"timestamp,omitempty"`
+	Type            EventType      `json:"type,omitempty"`
+	Format          ResponseFormat `json:"format,omitempty"`
+	Analytics       Analytics      `json:"analytics,omitempty"`
+	BidID           string         `json:"bidid,omitempty"`
+	AccountID       string         `json:"account_id,omitempty"`
+	Bidder          string         `json:"bidder,omitempty"`
+	Timestamp       int64          `json:"timestamp,omitempty"`
+	IntegrationType string         `json:"integration_type,omitempty"`
```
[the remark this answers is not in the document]
AlexBVolcy: Came up with the name myself, but I'll check the PBS-Java code to see what they do, but like you commented somewhere else, I might update it to `integration`

config/config.go
```diff
@@ -88,6 +88,8 @@ type Configuration struct {
 	GenerateBidID bool `mapstructure:"generate_bid_id"`
 	// GenerateRequestID overrides the bidrequest.id in an AMP Request or an App Stored Request with a generated UUID if set to true. The default is false.
 	GenerateRequestID bool `mapstructure:"generate_request_id"`
+	// Integration Type value lives in config so that it can be updated from the request, and then passed for the event URL
+	IntegrationType string
```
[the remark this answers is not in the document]
AlexBVolcy: It's because I was having trouble getting the integration information to update the vTrack URL integration parameter to the cachePutObjects function.  The cachePutObjects function, which eventually calls EventRequestToUrl, has the config object as a member of its function, so by adding the IntegrationType value to the config object, it allowed me to then access it to update the vtrack URL.

This was confusing for me to type, so it might be confusing for you to read, so we can also talk about it tomorrow on a call if you'd like to see if there's an easier way for me to get the integration type value to the vtrack URL.

endpoints/openrtb2/auction_test.go
AlexBVolcy: Yes will do!

endpoints/openrtb2/auction.go
AlexBVolcy: Yeah this is a good catch, I'll update this

endpoints/openrtb2/auction.go
AlexBVolcy: I'll look into this, but I think you're right!

endpoints/events/event.go
AlexBVolcy: Yes I believe so, I'll double check with Mansi, but I believe this is correct

openrtb_ext/request.go
AlexBVolcy: This is a good question to bring up in tomorrow's standup, but the current implementation I have is if integration is blank then we apply the default integration value. 

However, the answer for how a publisher could omit an integration type in general is unclear. They could just set their default integration value to be blank, but I don't know if that approaches is what we want.

analytics/event.go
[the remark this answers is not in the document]
bsardo: 👍 It looks like they do call it `integration` in PBS-Java.

config/accounts.go
```diff
@@ -13,13 +13,14 @@ const (
...
-	GDPR          AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
-	DebugAllow    bool        `mapstructure:"debug_allow" json:"debug_allow"`
+	ID                 string      `mapstructure:"id" json:"id"`
+	Disabled           bool        `mapstructure:"disabled" json:"disabled"`
+	CacheTTL           DefaultTTLs `mapstructure:"cache_ttl" json:"cache_ttl"`
+	EventsEnabled      bool        `mapstructure:"events_enabled" json:"events_enabled"`
+	CCPA               AccountCCPA `mapstructure:"ccpa" json:"ccpa"`
+	GDPR               AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
+	DebugAllow         bool        `mapstructure:"debug_allow" json:"debug_allow"`
+	DefaultIntegration string      `mapstructure:"default_integration"`
```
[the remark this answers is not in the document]
bsardo: 👍 I see PBS-Java is using `default-integration`.

openrtb_ext/request.go
bsardo: @mansinahar thoughts on this? Do we need to confirm this behavior with Bret? I did a little digging and it appears this implementation matches what's in PBS-Java and the edge case I described above is unaccounted for and perhaps not valid. In `Ortb2RequestFactory.java`, the `StringUtils.isBlank(integration)` call checks that the requeset integration is null or empty string before applying the default integration.

endpoints/events/event.go
bsardo: I see that this is the name of the parameter in PBS-Java.

openrtb_ext/request.go
mansinahar: @bsardo Yeah I think this should be fine. I don't see a specific use case for a publisher to be able to not specify any integration at all being mentioned in the issue description #1428. And as you mentioned it matches with the Java implementation as well.

mansinahar:
Can we please add some JSON tests for these changes similar to the event tests in: https://github.com/prebid/prebid-server/tree/master/exchange/exchangetest. Also, shouldn't the event URLs in these existing JSON tests should now have the  integration query param in them too? Like here for example: https://github.com/prebid/prebid-server/blob/master/exchange/exchangetest/events-vast-account-off-request-on.json#L124 or here: https://github.com/prebid/prebid-server/blob/master/exchange/exchangetest/events-bid-account-on-request-off.json#L76

endpoints/openrtb2/auction.go
```diff
@@ -1821,3 +1829,28 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
 	return false, nil
 }
+
+func (deps *endpointDeps) setIntegrationType(req *openrtb_ext.RequestWrapper, account *config.Account) error {
+	reqExt, err := req.GetRequestExt()
+	if err != nil {
+		return err
+	}
+	reqPrebid := reqExt.GetPrebid()
+	if reqPrebid == nil {
```
bsardo, line 1839: Do we want to verify `account` isn't nil here to ensure this method doesn't fail when attempting to access fields on the account? Technically the account shouldn't be nil given the current state of the code because the calling code makes sure the account exists before calling this function but adding this defensive programming check would make this more robust. @mansinahar thoughts?

mansinahar: Yeah, I agree with @bsardo. We should add an account not nil check as well before we access the Default Integration

endpoints/openrtb2/auction.go
```diff
@@ -1821,3 +1829,28 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
+		reqPrebid = &openrtb_ext.ExtRequestPrebid{Integration: account.DefaultIntegration}
+		reqExt.SetPrebid(reqPrebid)
+	} else if reqPrebid.Integration == "" {
+		reqPrebid.Integration = account.DefaultIntegration
+		reqExt.SetPrebid(reqPrebid)
+	}
+	return nil
+}
+
+func getIntegrationFromRequest(req *openrtb_ext.RequestWrapper) (string, error) {
```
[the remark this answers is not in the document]
AlexBVolcy: Yeah I wrote the function specifically for tests, I'll move the function to the test file!

endpoints/openrtb2/auction.go
```diff
@@ -1836,21 +1838,37 @@ func (deps *endpointDeps) setIntegrationType(req *openrtb_ext.RequestWrapper, ac
 		return err
 	}
 	reqPrebid := reqExt.GetPrebid()
-	if reqPrebid == nil {
+	if reqPrebid == nil && account != nil {
+		err := validateIntegrationType(account.DefaultIntegration)
```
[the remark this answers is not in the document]
bsardo: Discussed offline with the team and we will only validate on the event and vtrack endpoints to be consistent with Java so this can be disregarded.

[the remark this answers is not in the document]
bsardo: N/A since integration type validation will be removed from the auction endpoint.

endpoints/openrtb2/auction.go
```diff
@@ -1836,21 +1838,37 @@ func (deps *endpointDeps) setIntegrationType(req *openrtb_ext.RequestWrapper, ac
...
-	} else if reqPrebid.Integration == "" {
+	} else if reqPrebid.Integration == "" && account.DefaultIntegration != "" {
+		err := validateIntegrationType(account.DefaultIntegration)
+		if err != nil {
+			return err
+		}
 		reqPrebid.Integration = account.DefaultIntegration
 		reqExt.SetPrebid(reqPrebid)
+	} else {
+		err := validateIntegrationType(reqPrebid.Integration)
```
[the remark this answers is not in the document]
bsardo: Discussed offline with the team and we will only validate on the event and vtrack endpoints to be consistent with Java so this can be disregarded.

endpoints/events/event_test.go
```diff
@@ -677,3 +678,70 @@ func TestEventRequestToUrl(t *testing.T) {
...
+			description:      "Integration type in http request is valid, expect same integration time and no errors",
+			givenHttpRequest: httptest.NewRequest("GET", "/event?t=win&b=bidId&f=b&ts=1000&x=1&a=accountId&bidder=bidder&int=TestIntegrationType", strings.NewReader("")),
+			givenEventRequest: &analytics.EventRequest{
+				Type:        analytics.Imp,
+				BidID:       "bidid",
+				AccountID:   "accountId",
+				Bidder:      "bidder",
+				Timestamp:   1234567,
+				Format:      analytics.Blank,
+				Analytics:   analytics.Enabled,
```
bsardo, line 700: You can simplify all of the `givenEventRequest` objects defined in these test cases by just defining an empty object since none of these values are relevant to the tests and `Integration` can just take on the default empty string value.

endpoints/events/vtrack.go
```diff
@@ -273,8 +277,13 @@ func getAccountId(httpRequest *http.Request) string {
...
 }
 
-func getIntegrationType(httpRequest *http.Request) string {
-	return httpRequest.URL.Query().Get(IntegrationParameter)
+func getIntegrationType(httpRequest *http.Request) (string, error) {
+	integrationType := httpRequest.URL.Query().Get(IntegrationParameter)
+	err := validateIntegrationType(integrationType)
+	if err != nil {
+		return "", err
+	}
```
bsardo, line 285: Can you add test coverage for when the validate function returns an error here?

endpoints/events/event.go
```diff
@@ -162,6 +166,10 @@ func ParseEventRequest(r *http.Request) (*analytics.EventRequest, []error) {
 		errs = append(errs, err)
 	}
 
+	if err := readIntegrationType(event, r); err != nil {
+		errs = append(errs, err)
```
bsardo, line 170: Can you add coverage for this line? You should be able to just add another test which is basically a copy of `TestShouldReturnBadRequestWhenFormatValueIsInvalid`.

bsardo (dismissed):
LGTM

endpoints/events/vtrack.go
```diff
@@ -69,6 +70,13 @@ func (v *vtrackEndpoint) Handle(w http.ResponseWriter, r *http.Request, _ httpro
 		return
 	}
 
+	// get integration value from request parameter
+	integrationType, err := getIntegrationType(r)
+	if err != nil {
+		w.Write([]byte(fmt.Sprintf("Invalid integration type: %s\n", err.Error())))
```
mansinahar, line 76: Please add the `http.StatusBadRequest` response code when returning an error here. Look at L85 for reference

endpoints/events/vtrack_test.go
```diff
@@ -690,3 +692,39 @@ func getVTrackRequestData(wi bool, wic bool) (db []byte, e error) {
...
+		{
+			description:      "Integration type in http request contains invalid character, expect invalid character error",
+			givenHttpRequest: httptest.NewRequest("GET", "/event?t=win&b=bidId&f=b&ts=1000&x=1&a=accountId&bidder=bidder&int=Te$tIntegrationType", strings.NewReader("")),
+			expectedError:    errors.New("integration type can only contain numbers, letters and these characters '-', '_'"),
+		},
+	}
+
+	for _, test := range testCases {
+		integrationType, err := getIntegrationType(test.givenHttpRequest)
+		if err != nil {
```
mansinahar, line 723: Same comment as above for the `readIntegrationType` function test

endpoints/openrtb2/auction.go
```diff
@@ -1872,3 +1880,23 @@ func checkIfAppRequest(request []byte) (bool, error) {
...
+	reqExt, err := req.GetRequestExt()
+	if err != nil {
+		return err
+	}
+	reqPrebid := reqExt.GetPrebid()
+
+	if account == nil {
+		return nil
+	}
+	if reqPrebid == nil {
```
mansinahar, line 1894: You need the `account.DefaultIntegration != ""` check as part of this if condition too as it would be pointless to apply the default integration if it is empty.

The other option would be to change the if condition on L1891 to be `if account == nil || account.DefaultIntegration == ""`. This will allow you to return earlier if default integration isn't provided and then you can get rid of the `account.DefaultIntegration != ""` checks in the subsequent if conditions.

mansinahar:
LGTM in general other than the minor comments above.

PR #2114, opened 2021-12-19
TheMediaGrid: Fix bug with wrong type in response

TheMediaGrid:
In case of multiple types, TheMediaGrid was sending the wrong type if the endpoint responded with a video type


adapters/grid/grid.go
```diff
@@ -333,19 +345,23 @@ func (a *GridAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReq
 	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
 		return nil, []error{err}
 	}
+	var bidRespWithType GridResponseType
+	if err := json.Unmarshal(response.Body, &bidRespWithType); err != nil {
+		return nil, []error{err}
+	}
```
hhhjort, line 351: You are unmarshaling `response.Body` twice here, which is rather inefficient. You may be able to get around it with embedding.

`type GridBid struct {
         *openrtb.Bid
	ContentType openrtb_ext.BidType `json:"content_type"`
}

type GridSeatBid struct {
        *openrtb.SeatBid
	Bid []GridBid `json:"bid"`
}

type GridResponse struct {
         *openrtb.BidResponse
	SeatBid []GridSeatBid `json:"seatbid,omitempty"`
}
`
This way `GridResponse` has all the openrtb fields, but `Bid` is overridden to include the `content_type` field. Thus you would only need to unmarshal the response once and use the same struct for everything,

mansinahar:
@TheMediaGrid Can you please address @hhhjort's comment above when you get a chance? Thanks!

adapters/grid/grid.go
hhhjort: Oh, yes, I did not consider that you need to get back to the original `openrtb2.Bid`. As long as it is just the `Bid` you need to recast, I believe `openrtb2Bid := sb.Bid[i].Bid` will do it, assuming that `sb` is your `GridSeatBid` type.

TheMediaGrid: Thank you, your tip was very helpful!

hhhjort (approved):
LGTM

bsardo (approved):
LGTM

PR #2115, opened 2021-12-20
Huaweiads - 1. add the corresponding site endpoint selection 2. add package name replacement

wwwyyy:
update huaweiads adapter
1. add the corresponding site endpoint selection 
2. add package name replacement


hhhjort:
I just want to confirm that you have a reason for routing bid requests to a server based on the location of the of the person viewing the ad rather than the location of the PBS server. The common case we see is putting the endpoint an instance of PBS should use in the PBS config rather than calculating it on the fly as you are doing. Host based config has the advantage of, if the PBS instance is located a long distance from the actual user, you don't then make an equally long call from PBS to the bidder when there is a local bidder available.

For example, if a PBS host has instances set up in both Germany and Singapore, but due to some outage a user in France is routed to the Singapore PBS instance, it may be more efficient to have that PBS instance talk to your asian endpoint rather than your european endpoint.

However if you have your endpoints optimized for local demand, it does make sense to route the requests based on user location rather than PBS host location.

adapters/huaweiads/huaweiads.go
```diff
@@ -278,14 +296,41 @@ func (a *adapter) MakeRequests(openRTBRequest *openrtb2.BidRequest,
...
 	return []*adapters.RequestData{bidRequest}, nil
 }
 
+// countryCode is alpha2, choose the corresponding site end point
+func getFinalEndPoint(countryCode string, defaultEndpoint string) string {
+	if countryCode == "" || len(countryCode) > 2 {
+		return defaultEndpoint
+	}
+
+	var europeanSiteCountryCodeGroup = map[string]string{"AX": "1", "AL": "1", "AD": "1", "AU": "1", "AT": "1", "BE": "1",
```
hhhjort, line 313: It is slightly more efficient to define these as `map[string]interface{}`, as empty interfaces can be the most efficient to store. Also makes it a little more obvious that the value stored in the map has no actual use, it just exists to take advantage of the efficiency of the map's key lookup mechanics.

SyntaxNode: Is this attempting to simulate a hash set? If so, it's most efficient to use `map[string]struct{}` as the empty struct consumes literally no memory. All the fun idiosyncrasies of Go. See: https://medium.com/@l.peppoloni/how-to-improve-your-go-code-with-empty-structs-3bd0c66bc531

adapters/huaweiads/huaweiads.go
```diff
@@ -337,19 +400,21 @@ func getHeaders(huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds, request *openrtb2.
...
 // getHuaweiAdsReqJson: get body json for HuaweiAds request
-func getHuaweiAdsReqJson(request *huaweiAdsRequest, openRTBRequest *openrtb2.BidRequest, huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds) error {
+func getHuaweiAdsReqJson(request *huaweiAdsRequest, openRTBRequest *openrtb2.BidRequest,
+	huaweiAdsImpExt *openrtb_ext.ExtImpHuaweiAds, extraInfo ExtraInfo) (string, error) {
 	request.Version = huaweiAdxApiVersion
 	var err error
-	if err = getHuaweiAdsReqAppInfo(request, openRTBRequest); err != nil {
-		return err
+	var countryCode string
+	if countryCode, err = getHuaweiAdsReqAppInfo(request, openRTBRequest, extraInfo); err != nil {
```
hhhjort, line 408: you can avoid having to declare `err` and `countryCode` on separate lines by using
`if countryCode, err := getHuaweiAdsReqAppInfo(...`

This is just a stylistic quibble, if you prefer your current construction, that is valid as well.

adapters/huaweiads/huaweiads.go
```diff
@@ -482,9 +547,48 @@ func getHuaweiAdsReqAppInfo(request *huaweiAdsRequest, openRTBRequest *openrtb2.
...
+}
+
+// when it has pkgNameConvert (include different rules)
+// 1. when bundleName in ExceptionPkgNames, finalPkgname = bundleName
+// 2. when bundleName conform UnconvertedPkgNames, finalPkgname = ConvertedPkgName
+// 3. when bundleName conform keyword, finalPkgname = ConvertedPkgName
+// 4. when bundleName conform prefix, finalPkgname = ConvertedPkgName
+func getFinalPkgName(bundleName string, extraInfo ExtraInfo) string {
+	for _, convert := range extraInfo.PkgNameConvert {
+		if convert.ConvertedPkgName == "" {
```
hhhjort, line 563: Is there a use case for an `PkgNameConvert` entry to not have a value in `ConvertedPkgName`? If not you may want to check for that when building your adapter and throw an error if an entry is missing this field.

adapters/huaweiads/huaweiads.go
hhhjort: Thanks Scott, I was thinking empty structs but somehow landed on interfaces.

mansinahar:
@wwwyyy Friendly reminder about the above comments whenever you get a chance. Thanks!

wwwyyy:
> I just want to confirm that you have a reason for routing bid requests to a server based on the location of the of the person viewing the ad rather than the location of the PBS server. The common case we see is putting the endpoint an instance of PBS should use in the PBS config rather than calculating it on the fly as you are doing. Host based config has the advantage of, if the PBS instance is located a long distance from the actual user, you don't then make an equally long call from PBS to the bidder when there is a local bidder available.
> 
> For example, if a PBS host has instances set up in both Germany and Singapore, but due to some outage a user in France is routed to the Singapore PBS instance, it may be more efficient to have that PBS instance talk to your asian endpoint rather than your european endpoint.
> 
> However if you have your endpoints optimized for local demand, it does make sense to route the requests based on user location rather than PBS host location.

wwwyyy:
> I just want to confirm that you have a reason for routing bid requests to a server based on the location of the of the person viewing the ad rather than the location of the PBS server. The common case we see is putting the endpoint an instance of PBS should use in the PBS config rather than calculating it on the fly as you are doing. Host based config has the advantage of, if the PBS instance is located a long distance from the actual user, you don't then make an equally long call from PBS to the bidder when there is a local bidder available.
> 
> For example, if a PBS host has instances set up in both Germany and Singapore, but due to some outage a user in France is routed to the Singapore PBS instance, it may be more efficient to have that PBS instance talk to your asian endpoint rather than your european endpoint.
> 
> However if you have your endpoints optimized for local demand, it does make sense to route the requests based on user location rather than PBS host location.

 we are testing with clients, considering privacy concerns, and at the same time，
the data cannot be transferred across sites, and the country will correspond to the determined site，so make changes like this

adapters/huaweiads/huaweiads.go
wwwyyy: Thank you all, it has been revised

adapters/huaweiads/huaweiads.go
wwwyyy: Thank you，done

adapters/huaweiads/huaweiads.go
wwwyyy: thank you, parameter validation has been added, as well as test code

wwwyyy:
> @wwwyyy Friendly reminder about the above comments whenever you get a chance. Thanks!

Thank you, the issues in the code has been corrected, please check, thank you very much for your suggestions

adapters/huaweiads/huaweiads.go
wwwyyy: yes, I know, the check code was added in the Builder() -> getExtraInfo(), getExtraInfo will return the error, Builder()  can catch, you can check.

hhhjort (approved):
LGTM

VeronikaSolovei9 (approved):
LGTM

PR #2117, opened 2021-12-22
Remove Adagio adapter

osazos:
This PR is just about to remove the Adagio adapter.

Thanks


SyntaxNode:
The Adagio adapter was added to the Prebid Server project just 6 months ago. I'm sorry to see you leave. Could you please fill us in on your decision to remove your adapter?

osazos:
Hi @SyntaxNode, 

We made this choice due the lack of consistency of Prebid Server integration by each owners. Actually we have to spent too much time to try to get a correct implementation of our adapter, which is even not always possible.

mansinahar:
@osazos You will also have to add an entry for Adagio in this `disabledBidders` map here: https://github.com/prebid/prebid-server/blob/master/exchange/adapter_util.go#L81

osazos:
Hi @mansinahar, it's done.

AlexBVolcy (dismissed):

mansinahar:
@osazos Thanks for adding that. Btw, I noticed that this line needs to be removed too: https://github.com/prebid/prebid-server/blob/master/config/config.go#L796. I think once you remove this, this PR should be good. Sorry, I missed that earlier.

osazos:
hi @mansinahar, no prob, thanks for your review!.
It's done.

mansinahar (approved):

VeronikaSolovei9 (approved):
LGTM

PR #2119, opened 2021-12-27
New Adapter: Vidoomy

abdollah-aghayan:


adapters/vidoomy/params_test.go
```diff
@@ -0,0 +1,56 @@
...
+var invalidParams = []string{
+	``,
+	`null`,
+	`true`,
+	`5`,
+	`4.2`,
+	`[]`,
+	`{}`,
+	`{"notmid":"123"}`,
+	`{"zone_id":123}`,
```
SyntaxNode, line 55: Could you please add a test with `zoneId` set to an empty string to test the min length constraint and also a test with `zoneId` set to a different type to test the string constraint. I see you have a test with `zone_id` set to a number, did you intend for this to test the `zoneId` property? As of now, it's testing an unspecified property which is already covered by the `notmid` case.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,192 @@
...
+	var errors []error
+
+	reqs := make([]*adapters.RequestData, 0, len(request.Imp))
+
+	for idx, imp := range request.Imp {
+		var bidderExt adapters.ExtImpBidder
+		err := json.Unmarshal(imp.Ext, &bidderExt)
+		if err != nil {
+			err = &errortypes.BadInput{
+				Message: fmt.Sprintf("imp #%d: ext.bidder not provided", idx),
```
SyntaxNode, line 28: Please consider revising the wording of "ext.bidder not provided" since that's just one possible reason why unmarshal would fail and perhaps include the error message itself in the response to help the publisher debug. I appreciate the good approach to error handling here.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,192 @@
...
+
+		var vidoomyExt openrtb_ext.ImpExtVidoomy
+		err = json.Unmarshal(bidderExt.Bidder, &vidoomyExt)
+		if err != nil {
+			err = &errortypes.BadInput{
+				Message: fmt.Sprintf("imp #%d: %s", idx, err.Error()),
+			}
+			errors = append(errors, err)
+			continue
+		}
```
SyntaxNode, line 42: You may wish to consider using a different struct such that only one call to unmarshal is required. This will simplify the code here a bit. The data structure, which could be anonymous, would look something like:

```
type extBidder struct {
   vidoomyExt openrtb_ext.ImpExtVidoomy `json:"bidder"`
}
```

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,192 @@
...
+	var exists bool
+	for _, imp := range imps {
+		if imp.ID == impId {
+			exists = true
+			if imp.Banner != nil {
+				mediaType = openrtb_ext.BidTypeBanner
+			} else if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
+			break
```
SyntaxNode, line 180: Consider using a return statement instead of break here. I think it's a bit easier to read / reason about.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,192 @@
...
+	var mediaType openrtb_ext.BidType
+	var exists bool
+	for _, imp := range imps {
+		if imp.ID == impId {
+			exists = true
+			if imp.Banner != nil {
+				mediaType = openrtb_ext.BidTypeBanner
+			} else if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			}
```
SyntaxNode, line 179: Does this match the evaluation order of your bidding server? If an impression has both a banner and video format, is banner always bid for first?

abdollah-aghayan: We support one impression per request so we have no order, just trying to get the impression type.

adapters/vidoomy/params_test.go
abdollah-aghayan: fixed the typo in zone id, added empty `zoneId`

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Changed for better wording and including errors in messages.

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Moved ext unmarshaling to a new function to simplify.

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Done!

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,201 @@
...
+	if request.Device == nil {
+		return headers
+	}
+
+	if request.Device.UA != "" {
+		headers.Set("User-Agent", request.Device.UA)
+	}
+
+	if len(request.Device.IPv6) > 0 {
+		headers.Add("X-Forwarded-For", request.Device.IPv6)
```
SyntaxNode, line 75: Please add a json test case for IPv6.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	banner := *request.Imp[0].Banner
+	request.Imp[0].Banner = &banner
+
+	if banner.W != nil && banner.H != nil {
+		if *banner.W == 0 || *banner.H == 0 {
+			return fmt.Errorf("invalid sizes provided for Banner %d x %d", *banner.W, *banner.H)
+		}
+		return nil
```
SyntaxNode, line 97: Please add a json test case for this return.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,201 @@
...
+
+	banner.W = openrtb2.Int64Ptr(banner.Format[0].W)
+	banner.H = openrtb2.Int64Ptr(banner.Format[0].H)
+
+	return nil
+}
+
+func (a adapter) MakeBids(internalRequest *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if response.StatusCode == http.StatusNoContent {
+		return nil, nil
```
SyntaxNode, line 112: Please add a json test case for this response code.

adapters/vidoomy/vidoomy.go
SyntaxNode: Is there a way to determine from your bid response if the bid is for banner or video? If the one impression defines both options, what gives you confidence here that the correct impression type is selected?

SyntaxNode (commented):
Thank you for quickly addressing the feedback. I took another pass focusing on the test coverage and noticed a few small gaps. I'm overall impressed with the test coverage - thank you for giving that proper attention.

SyntaxNode:
Docs PR: https://github.com/prebid/prebid.github.io/pull/3477

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Done!

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Done!

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Done!

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Good point!, Based on Openrtb 2.5 which says

`The presence of Banner (Section 3.2.6), Video (Section 3.2.7), and/or Native (Section 3.2.9) objects
subordinate to the Imp object indicates the type of impression being offered. The publisher can choose
one such type which is the typical case or mix them at their discretion. However, any given bid for the
impression must conform to one of the offered types.`

So my understanding is it is not possible to have an impression with multiple types!

SyntaxNode: The `publisher can choose ... [a] mix ... at their discretion`, so the publisher may define an impression with both banner and video objects if they like. A bid is only for one., either banner or video in this case. The intent of this method is for the adapter to tell Prebid Server which format the bid is for. Some bidders include the bid type in their server response, some re-implement the same bidding logic in their adapter that their server follows, and some effectively choose one at random which could result in rejected bids.

I personally view this as an issue with the OpenRTB 2.5 spec itself. Luckily, OpenRTB 2.6 will fix this issue by including the media type (or "mtype") in the standard bid response. Until then, each adapter needs to do the best they can to match up the bid with the format. I just want to be sure that's the case here - that this is the best logic possible for Vidoomy to give the adapter the best shot at success.

abdollah-aghayan: Thanks for your explanation and your effort, I can assure you that this is the best solution for us, we only support video and banner, if we receive an impression with both types available we chose Video over Banner so only one type will be available in impression.

SyntaxNode: > if we receive an impression with both types available we chose Video over Banner

Cool. Could you please update the adapter code to match? Right now, it will prefer banner over video. If you put the video check first, then you'll have matching behavior:

```
if imp.Video != nil {
} else if imp.Banner != nil {
	mediaType = openrtb_ext.BidTypeBanner
}
```

adapters/vidoomy/vidoomytest/exemplary/simple-app-banner.json
```diff
@@ -0,0 +1,116 @@
...
+            "zoneId": "zone-id-123"
+          }
+        }
+      }
+    ],
+    "device": {
+      "ua": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.121 Safari/537.36",
+      "ip": "127.0.0.1"
+    },
+    "site": {
```
mansinahar, line 21: The title for this test says `simple-app-banner` request but I don't see an app block in here. It's a site request rather. Also, how is this one different from the `simple-banner.json` test?

adapters/vidoomy/vidoomytest/exemplary/simple-banner-no-response.json
```diff
@@ -0,0 +1,85 @@
+{
+  "mockBidRequest": {
+    "id": "test-request-id",
+    "imp": [
```
mansinahar, line 4: Can you please move this test to the supplemental directory as well?

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Since it is not possible to have multiple types based on my explanation, yet I changed the order.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	if len(request.Device.IP) > 0 {
+		headers.Add("X-Forwarded-For", request.Device.IP)
+	}
+
+	return headers
+}
+
+func changeRequestForBidService(request *openrtb2.BidRequest, extension *openrtb_ext.ImpExtVidoomy) error {
```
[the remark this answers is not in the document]
abdollah-aghayan: I removed the extension input, but I in `validateBanner` does not validate if `banner.W` or `banner.H` is set to zero.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,175 @@
...
+			continue
+		}
+
+		reqJSON, err := json.Marshal(&reqCopy)
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+
+		header := getHeaders(request)
```
hhhjort, line 41: You may want to define `header` in front of the loop, so you will only have to execute `getHeaders()` once rather than once per imp.

adapters/vidoomy/vidoomy.go
```diff
@@ -0,0 +1,175 @@
...
+			if imp.Video != nil {
+				mediaType = openrtb_ext.BidTypeVideo
+			} else if imp.Banner != nil {
+				mediaType = openrtb_ext.BidTypeBanner
+			}
+
+			return exists, mediaType
+		}
+	}
+	return exists, mediaType
```
hhhjort, line 166: You may want to just `return false, nil` since these should be the only values these variables can have at this point in the function.

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Done!

adapters/vidoomy/vidoomy.go
abdollah-aghayan: Done!

hhhjort (approved):
LGTM

PR #2122, opened 2022-01-05
GDPR: pass geo based on special feature 1 opt-in

bsardo:
This PR fixes a GDPR bug and depends on https://github.com/prebid/go-gdpr/pull/34.
We currently determine if geo information should be passed in a bid request based on whether special purpose 1 is set in the GVL for the vendor which is incorrect. Instead we should be looking at whether special feature 1 is set for the vendor. The main change is here in `impl.go` on line 137:
`passGeo = vendorException || (consentMeta.SpecialFeatureOptIn(1) && (vendor.SpecialFeature(1) || weakVendorEnforcement))`

Special purposes are defined in the TCF2 spec but they are not used by PBS. As a result, all references to special purpose 1 have been renamed to eliminate confusion.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -287,7 +287,7 @@ type TCF2 struct {
 	Purpose8            TCF2Purpose             `mapstructure:"purpose8"`
 	Purpose9            TCF2Purpose             `mapstructure:"purpose9"`
 	Purpose10           TCF2Purpose             `mapstructure:"purpose10"`
-	SpecialPurpose1     TCF2Purpose             `mapstructure:"special_purpose1"`
+	SpecialFeature1     TCF2SpecialFeature      `mapstructure:"special_feature1"`
 	PurposeOneTreatment TCF2PurposeOneTreatment `mapstructure:"purpose_one_treatment"`
 }
 
@@ -296,6 +296,13 @@ type TCF2Purpose struct {
 	Enabled        bool   `mapstructure:"enabled"` // Deprecated: Use enforce_purpose instead
 	EnforcePurpose string `mapstructure:"enforce_purpose"`
 	EnforceVendors bool   `mapstructure:"enforce_vendors"`
+	// Array of vendor exceptions that is used to create the hash table VendorExceptionMap so vendor names can be instantly accessed
+	VendorExceptions   []openrtb_ext.BidderName `mapstructure:"vendor_exceptions"`
+	VendorExceptionMap map[openrtb_ext.BidderName]struct{}
+}
+
+type TCF2SpecialFeature struct {
+	Enforce bool `mapstructure:"enforce"`
 	// Array of vendor exceptions that is used to create the hash table VendorExceptionMap so vendor names can be instantly accessed
 	VendorExceptions   []openrtb_ext.BidderName `mapstructure:"vendor_exceptions"`
 	VendorExceptionMap map[openrtb_ext.BidderName]struct{}
@@ -556,7 +563,6 @@ func New(v *viper.Viper) (*Configuration, error) {
 		&c.GDPR.TCF2.Purpose8,
 		&c.GDPR.TCF2.Purpose9,
 		&c.GDPR.TCF2.Purpose10,
-		&c.GDPR.TCF2.SpecialPurpose1,
 	}
 	for c := 0; c < len(purposeConfigs); c++ {
 		purposeConfigs[c].VendorExceptionMap = make(map[openrtb_ext.BidderName]struct{})
@@ -565,6 +571,14 @@ func New(v *viper.Viper) (*Configuration, error) {
 			bidderName := purposeConfigs[c].VendorExceptions[v]
 			purposeConfigs[c].VendorExceptionMap[bidderName] = struct{}{}
 		}
+	}
+
+	// To look for a special feature's vendor exceptions in O(1) time, we fill this hash table with bidders located in the
+	// VendorExceptions field of the GDPR.TCF2.SpecialFeature1 struct defined in this file
+	c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap = make(map[openrtb_ext.BidderName]struct{})
+	for v := 0; v < len(c.GDPR.TCF2.SpecialFeature1.VendorExceptions); v++ {
+		bidderName := c.GDPR.TCF2.SpecialFeature1.VendorExceptions[v]
+		c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap[bidderName] = struct{}{}
 	}
 
 	// To look for a request's app_id in O(1) time, we fill this hash table located in the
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -287,7 +287,7 @@ type TCF2 struct {
 	Purpose8            TCF2Purpose             `mapstructure:"purpose8"`
 	Purpose9            TCF2Purpose             `mapstructure:"purpose9"`
 	Purpose10           TCF2Purpose             `mapstructure:"purpose10"`
-	SpecialPurpose1     TCF2Purpose             `mapstructure:"special_purpose1"`
+	SpecialFeature1     TCF2SpecialFeature      `mapstructure:"special_feature1"`
```
SyntaxNode, line 290: Interesting. We were using the wrong terminology?

config/config.go
```diff
@@ -567,6 +573,14 @@ func New(v *viper.Viper) (*Configuration, error) {
 		}
 	}
 
+	// To look for a special feature's vendor exceptions in O(1) time, we fill this hash table with bidders located in the
+	// VendorExceptions field of the GDPR.TCF2.SpecialFeature1 struct defined in this file
+	c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap = make(map[openrtb_ext.BidderName]struct{})
+	for v := 0; v < len(c.GDPR.TCF2.SpecialFeature1.VendorExceptions); v++ {
+		bidderName := c.GDPR.TCF2.SpecialFeature1.VendorExceptions[v]
+		c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap[bidderName] = struct{}{}
+	}
```
SyntaxNode, line 582: Please verify this code has full test coverage for cases of none, one, and many (2). It's been a problem for us in the past.

config/config.go
bsardo: It's not just a matter of the wrong terminology. We are looking at the wrong field in the GVL. The go-gdpr package provides a `vendor.SpecialPurpose` method that checks if a special purpose is defined in the GVL `specialPurposes` field for a vendor. We are currently calling this method but that isn't correct. Instead we should be looking at GVL `specialFeatures` for a vendor. I have a go-gdpr PR out that is currently under review that will add a new method that allows us to access the correct field for a given vendor.

SyntaxNode: That'a concerning. What's the difference between a special purpose and a special feature? The TCF folks could've done better in the naming there.

config/config.go
bsardo: 👍 Done

config/config_test.go
```diff
@@ -765,6 +765,79 @@ func TestMigrateConfigPurposeOneTreatment(t *testing.T) {
...
+
+		migrateConfigSpecialFeature1(v)
+
+		if len(tt.config) > 0 {
+			assert.Equal(t, tt.wantSpecialFeature1Enforce, v.Get("gdpr.tcf2.special_feature1.enforce").(bool), tt.description)
+			assert.Equal(t, tt.wantSpecialFeature1VendorExceptions, v.GetStringSlice("gdpr.tcf2.special_feature1.vendor_exceptions"), tt.description)
+		} else {
+			assert.Nil(t, v.Get("gdpr.tcf2.special_feature1.enforce"), tt.description)
+			assert.Nil(t, v.Get("gdpr.tcf2.special_feature1.vendor_exceptions"), tt.description)
+		}
```
guscarreon, line 837: I find this to be a very comprehensive test case but: can we make one last assertion in the resulting `Configuration` object itself? I realize `TestSpecialFeature1VendorExceptionMap` does this to an extent but not when `oldSpecialFeature1Config` is set. Maybe we could add an `v.Unmarshal()` call at the end of this test case so we assert that `cfg.GDPR.TCF2.SpecialFeature1` was correctly set even when `gdpr.tcf2.special_purpose1` is set as in `oldSpecialFeature1Config`.
```
824     for _, tt := range tests {
825         v := viper.New()
826         v.SetConfigType("yaml")
827         v.ReadConfig(bytes.NewBuffer(tt.config))
828
829         migrateConfigSpecialFeature1(v)
830
831         if len(tt.config) > 0 {
832             assert.Equal(t, tt.wantSpecialFeature1Enforce, v.Get("gdpr.tcf2.special_feature1.enforce").(bool), tt.description)
833             assert.Equal(t, tt.wantSpecialFeature1VendorExceptions, v.GetStringSlice("gdpr.tcf2.special_feature1.vendor_exceptions"), tt.description)
834         } else {
835             assert.Nil(t, v.Get("gdpr.tcf2.special_feature1.enforce"), tt.description)
836             assert.Nil(t, v.Get("gdpr.tcf2.special_feature1.vendor_exceptions"), tt.description)
837         }
    +
    +	    var c Configuration
    +       err := v.Unmarshal(&c)
    +       assert.NoError(t, err, tt.desc)
    +
    +       // Assert expected values on the `cfg.GDPR.TCF2.SpecialFeature1` object itself
    +       assert.Equal(t, tt.wantSpecialFeature1Enforce, cfg.GDPR.TCF2.SpecialFeature1.Enforce, tt.desc)
    +       assert.ElementsMatch(t, tt.wantSpecialFeature1VendorExceptions, cfg.GDPR.TCF2.SpecialFeature1.VendorExceptions, tt.desc)
838     }
config/config_test.go
```
Or even better, a `New(v)` call so we can also assert the correctness of the `cfg.GDPR.TCF2.SpecialFeature1.VendorExceptionMap`
```
824     for _, tt := range tests {
825         v := viper.New()
826         v.SetConfigType("yaml")
827         v.ReadConfig(bytes.NewBuffer(tt.config))
828
829         migrateConfigSpecialFeature1(v)
830
831         if len(tt.config) > 0 {
832             assert.Equal(t, tt.wantSpecialFeature1Enforce, v.Get("gdpr.tcf2.special_feature1.enforce").(bool), tt.description)
833             assert.Equal(t, tt.wantSpecialFeature1VendorExceptions, v.GetStringSlice("gdpr.tcf2.special_feature1.vendor_exceptions"), tt.description)
834         } else {
835             assert.Nil(t, v.Get("gdpr.tcf2.special_feature1.enforce"), tt.description)
836             assert.Nil(t, v.Get("gdpr.tcf2.special_feature1.vendor_exceptions"), tt.description)
837         }
    +
    +	    cfg := New(v)
    +
    +       // Assert expected values on the `cfg.GDPR.TCF2.SpecialFeature1` object itself
    +       assert.Equal(t, tt.wantSpecialFeature1Enforce, cfg.GDPR.TCF2.SpecialFeature1.Enforce, tt.desc)
    +       assert.ElementsMatch(t, tt.wantSpecialFeature1VendorExceptions, cfg.GDPR.TCF2.SpecialFeature1.VendorExceptions, tt.desc)
    +       assert.Equal(t, tt.wantSpecialFeature1VendorExceptionsMap, cfg.GDPR.TCF2.SpecialFeature1.VendorExceptionMap, tt.desc)
838     }
config/config_test.go
```
Another option is to probably add an `oldSpecialFeature1Config` scenatio to `TestSpecialFeature1VendorExceptionMap`? What do you think?

guscarreon (commented):
Left a nitpick

mansinahar:
Sounds good! Thanks for looking into it @bsardo and providing clarification.

config/config_test.go
bsardo: @guscarreon and I spoke offline about this.

My line of thinking was that the coverage was there across three unit tests but it was hard to see:
`TestMigrateConfigSpecialFeature1` —> verifies the deprecated special purpose config values are properly mapped to the new special feature config values
`TestFullConfig` —> verifies all of the special purpose struct fields, including the vendor exception map, are correctly populated based on what’s set in the special feature config values
`TestSpecialFeature1VendorExceptionMap` —> verifies the vendor exception map is created correctly no matter how many elements are in the vendor exception slice

I'm hesitant to make use of another function in the package to perform assertions in a pure unit test so instead of calling `New()` we agreed to call `unmarshal` as a happy medium to ensure what's written to the viper config in the migrate function is able to be unmarshaled into the struct.

guscarreon (approved):
Looks great. Thank you for addressing my feedback

VeronikaSolovei9 (approved):
Looks good, very straight forward.

Issue #2010, opened 2021-09-24
PBS Version In Response Header


SyntaxNode:
## Type Of Issue
Proposed Enhancement

## Description
A complimentary feature to https://github.com/prebid/prebid-server/issues/1818, allows publishers to inspect from the response headers for which version of PBS handled their bid request. Useful for monitoring/logging, especially during a deployment where multiple versions running simultaneously and the response from the version endpoint may not be accurate. This is useful for hosts to monitor deployment progress and issues.

## Proposal
PBS creates a header in the response to a bid request with the PBS version, ex:
```
x-prebid: pbs-go/0.178.0
```

bretg:
Discussed in PBS committee. No objections.

bretg:
Released with PBS-Java 1.79

SyntaxNode:
implemented in PBS-Go 0.190.0

PR #2129, opened 2022-01-10
PBS Version In Response Header

SyntaxNode:
Implements https://github.com/prebid/prebid-server/issues/2010


endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1328,13 +1328,13 @@ func TestIdGeneration(t *testing.T) {
...
 		// Set up and run test
-		actualAmpObject, endpoint := AmpObjectTestSetup(t, "test", test.givenInStoredRequest, test.givenGenerateRequestID)
+		actualAmpObject, endpoint := ampObjectTestSetup(t, "test", test.givenInStoredRequest, test.givenGenerateRequestID)
 		endpoint(recorder, request, nil)
 		assert.Equalf(t, test.expectedID, actualAmpObject.Request.ID, "Bid Request ID is incorrect: %s\n", test.description)
 	}
 }
 
-func AmpObjectTestSetup(t *testing.T, inTagId string, inStoredRequest json.RawMessage, generateRequestID bool) (*analytics.AmpObject, httprouter.Handle) {
+func ampObjectTestSetup(t *testing.T, inTagId string, inStoredRequest json.RawMessage, generateRequestID bool) (*analytics.AmpObject, httprouter.Handle) {
```
SyntaxNode, line 1337: There was no need to export this method.

endpoints/openrtb2/video_auction_test.go
```diff
@@ -29,11 +28,7 @@ import (
 
 func TestVideoEndpointImpressionsNumber(t *testing.T) {
 	ex := &mockExchangeVideo{}
-	reqData, err := ioutil.ReadFile("sample-requests/video/video_valid_sample.json")
-	if err != nil {
-		t.Fatalf("Failed to fetch a valid request: %v", err)
-	}
-	reqBody := string(getRequestPayload(t, reqData))
+	reqBody := readVideoTestFile(t, "sample-requests/video/video_valid_sample.json")
```
SyntaxNode, line 31: Added a test loading helper method like the auction and amp endpoint have.

version/xprebidheader.go
```diff
@@ -0,0 +1,53 @@
...
+	if version == "" {
+		version = VerUnknown
+	}
+	if sb.Len() != 0 {
+		sb.WriteString(",")
+	}
+	sb.WriteString(name)
+	sb.WriteString("/")
+	sb.WriteString(version)
+}
```
SyntaxNode, line 53: Copied/pasted code from another file. No change to BuildXPrebidHeaderForRequest or writeNameVersionRecord.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1359,3 +1359,65 @@ func AmpObjectTestSetup(t *testing.T, inTagId string, inStoredRequest json.RawMe
 	)
 	return &actualAmpObject, endpoint
 }
+
+func TestAmpAuctionResponseHeaders(t *testing.T) {
+	testCases := []struct {
+		description         string
+		requestURLArguments string
+		expectedStatus      int
+		expectedHeaders     func(http.Header)
```
AlexBVolcy, line 1368: For my own understanding, what is happening here with `func(http.Header)`? I haven't seen this syntax with `func(something)` and see it in a couple of places here in this PR.

AlexBVolcy (commented):
This looks good to me! I checked code coverage against the new additions you made and that looks good too. I'm going to wait until Gus reviews to see if I may have missed something.

guscarreon (commented):
Looks pretty good. I'm ready to approve after merge conflicts get resolved.

endpoints/openrtb2/amp_auction_test.go
SyntaxNode: Good question. I am defining a function pointer variable. This is the same format as a normal function header, but I'm omitting the variable name since it doesn't matter in this context. In the test cases, I'm defining anonymous methods and need to provide a variable name for use in the body.

I'm doing this here so that I can use the `.Set()` method for adding headers to the map. Since it is a really just a map, I could get away with a map literal.. but there is something formatting performed by the headers class and I want to be sure I'm using the exact same code paths for the test. I could also use a map and then loop through in making calls to `.Set()`, but I thought this was easier and more direct. It's an implementation choice and I wouldn't comment on anyone using an alternate approach. It's just that this is what looks good / makes sense to me.

bsardo (approved):
LGTM

guscarreon (approved):
LGTM

Issue #2127, opened 2022-01-06
Bidders that need to change whole-hostname parameter


bretg:
An email was sent to the pbs-adapters mailing list in October outlining an issue that the Prebid Server committee is now going to take up.

Some adapters are using a dynamic endpoint domain as a bidder parameter. Prebid Server is not an open proxy. If necessary, you may specify a _portion_ of the domain as a parameter to support geo regions or account specific servers. However, this is discouraged and may degrade the performance of your adapter since the server needs to maintain more outgoing connections. Host companies may choose to disable your adapter if it uses a dynamically configured domain.

For example, this auction endpoint URL is no longer allowed:

```
https://{{host}}/getbids
```
Why? Because the domain name is entirely dependent on an incoming parameter that could be spoofed, causing Prebid Server to make requests anywhere on the internet unless the host company takes pains to block that behavior.

This one is better:
```
https://{{host}}.example.com/getbids
```
Here, spoofed parameters can't do as much damage.

However, even this is only necessary if a bidder insists on publisher-specific domains. So let's talk about tracking. It's understandable that bidder endpoints will want to track publishers and the source Prebid Server. But publishers should be tracked within the ORTB2 body, e.g. as $.site.publisher.id. Our recommendation is that bid adapters design their auction endpoints so that publisher is an internal value rather than being on the URL.

PBS host config can be different for each host company 
https://auction.example.com/getbids?pbs-source=magnite

Note that PBS host companies have the ability to support region-specific endpoints. For instance, if you have bidder endpoints in EU and US, you can work with the various PBS host companies running your adapter to deploy different configuration pointing properly to these endpoints.

Here are the adapters that currently utilize parameters that define the entire hostname

- [x] adkernel / adkernelAdn uses http://{{.Host}}/hb?zone={{.ZoneID}} - heads up @ckbo3hrk
- [ ] adocean uses https://{{.Host}} - heads up @mmuras 
- [ ] datablocks/zeroclickfraud uses http://{{.Host}}/openrtb2?sid={{.SourceId}} - heads up @htang555 
- [x] invibes uses https://{{.Host}}/bid/ServerBidAdContent - heads up @invibes 
- [x] tappx uses http://{{.Host}} - heads up @prebidtappx , @ah-tappx 
- [ ] yeahmobi uses https://{{.Host}}/prebid/bid - heads up @zhaojp and @junping.zhao

[...]

PR #2130, opened 2022-01-10
Tappx: Remove Host Parameter

prebidtappx:
In this pull request we solve the hostname problem (https://github.com/prebid/prebid-server/issues/2127). The solution has been to remove the host parameter and try to adapt everything to prebid's needs. However, we have made some of the code backward compatible.
In conclusion, we have made the changes to prevent prebid from being used as a proxy.


bretg:
Thanks for the action on this @prebidtappx

prebidtappx:
Hi! All the changes are done! We need to do something else or is in review progress?
Thanks for everything!

mansinahar:
@prebidtappx This is on our list to take a look at today. Thanks!

adapters/tappx/tappx.go
```diff
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
-	"strings"
+	//"strings"
```
mansinahar, line 10: Can you please remove this import if it is no longer needed?

mansinahar (commented):
@prebidtappx LGTM generally except the very minor comment above and also, can you please open up a docs PR for this as well over here: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/tappx.md ? Thanks!

adapters/tappx/tappx.go
prebidtappx: Removed the import in commit 793b2b4
Updated the docs in pull request: https://github.com/prebid/prebid.github.io/pull/3520

adapters/tappx/tappx.go
```diff
@@ -137,7 +131,15 @@ func (a *TappxAdapter) buildEndpointURL(params *openrtb_ext.ExtImpTappx, test in
 		}
 	}
 
-	endpointParams := macros.EndpointTemplateParams{Host: params.Host}
+	tappxHost := "tappx.com"
+	isNewEndpoint, err := regexp.Match(`^(zz|vz)[0-9]{3,}([a-z]{2}|test)$`, []byte(params.Endpoint))
+	if isNewEndpoint {
+		tappxHost = params.Endpoint + ".pub." + tappxHost + "/rtb/"
```
mansinahar, line 137: @prebidtappx Quick question: Is the `params.Endpoint + ".pub." + tappxHost + "/rtb/"` going to be the new Tappx endpoint and is the plan to deprecate `tappxHost = "ssp.api." + tappxHost + "/rtb/v2/"` in the future?

prebidtappx: Yes, it is the plan. However, a lot of people don't like the changes. For this reason, all the changes are backward compatible.

bretg (commented):
(sorry, meant to put this in the docs PR)
appreciate the doc update @prebidtappx , but since you support both Prebid Server and Prebid.js, please confirm that the host param is optional for both. I don't see any PRs for PBJS.

hhhjort (approved):
LGTM

mansinahar (approved):

Issue #1739, opened 2021-03-05
Pass host company config info to bid adapters


bretg:
We discussed in Prebid Server committee a desire from some adapters to know which cluster of Prebid Servers is sending requests to their endpoints. 

The proposal is that PBS should start sending identifying information that will be sent to each bid adapter, where they can determine what's relevant for their endpoint
- `external-url` on ext.prebid.server.externalurl
- `gdpr.host-vendor-id` on ext.prebid.server.gvlid
- `datacenter-region` on ext.prebid.server.datacenter

Since datacenter location is not a standard and this info is for debugging and reporting, we're not proposing any syntax or valid values here.

The committee agreed that it makes sense to place the host company info on both the JSON body and in a configuration data structure available to bid adapters.

grossjo:
Should there be consistency on data centre regions?  Maybe something like ISO country codes, or is there need for more granularity in places like the US?

bretg:
I don't think standardizing on datacenter values is practical. There are often multiple datacenters in the US (east/west), and the particular country isn't really that important. For instance, Magnite currently uses 6 datacenters, all based out of Amazon regions. But we may shift to 'bare metal'  hosting at some point, and the "Europe-Central" might move from Germany to the Netherlands for instance. Or US-East from Virginia to wherever.

However, if there's a standard I'm not aware of, we'll discuss in committee.

bretg:
released with PBS-Java 1.88

SyntaxNode:
Implemented in PBS-Go 0.229.0.

PR #2133, opened 2022-01-13
New Adapter: Medianet

eshaanmangal:
Adds Media.Net Bidder Adapter.
Vendor ID: 142
Maintainer Email: prebid-support@media.net


guscarreon (commented):
Looks good. Left a couple of comments

adapters/medianet/medianet.go
```diff
@@ -0,0 +1,139 @@
...
+
+type MedianetAdapter struct {
+	endpoint string
+}
+
+func (a *MedianetAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errs []error
+
+	if len(request.Imp) == 0 {
+		return nil, []error{&errortypes.BadInput{
```
guscarreon, line 23: Prebid server core filters out requests with empty []Imp arrays. In production line 23 would be unreachable.

adapters/medianet/medianet.go
```diff
@@ -0,0 +1,139 @@
...
+	var medianetExt openrtb_ext.ExtImpMedianet
+	if err := json.Unmarshal(bidderExt.Bidder, &medianetExt); err != nil {
+		return &errortypes.BadInput{Message: "bad Medianet bidder ext"}
+	}
+
+	if len(medianetExt.Cid) == 0 && len(medianetExt.Crid) == 0 {
+		return &errortypes.BadInput{Message: "'cid' and 'crid' are required attribute for Medianet's bidder ext"}
+	} else if len(medianetExt.Cid) == 0 {
+		return &errortypes.BadInput{Message: "'cid' is required attribute for Medianet's bidder ext"}
+	} else if len(medianetExt.Crid) == 0 {
```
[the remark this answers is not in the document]
eshaanmangal: Yes, Thanks for pointing this out, I will add it in bidder-params but for custom error message (specific to medianet) this code will still be required.
Elaboration:
For test case `medianettest/supplemental/invaid-req-missing-cid-crid-error.json`
In the above test case I will get error `"'cid' and 'crid' are required attribute for Medianet's bidder ext"` but if I remove those checks from code and add it in bidder-param json file I will get `MakeRequests had wrong error count. Expected 1, got 0 ([])` which is not quite specific

adapters/medianet/medianet.go
eshaanmangal: Resolved !!

static/bidder-info/medianet.yaml
```diff
@@ -0,0 +1,22 @@
...
+    mediaTypes:
+      - banner
+      - video
+      - native
+userSync:
+  key: medianet
+  redirect:
+    url: https://hbx.media.net/cksync.php?cs=1&type=pbs&ovsid=setstatuscode&redirect=
+    redirect_url: /setuid?bidder=medianet&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&uid={{.UserMacro}}
+    userMacro: "${UUID}"
```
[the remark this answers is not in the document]
eshaanmangal: Thanks for pointing it out. I have resolved earlier it was not working

static/bidder-params/medianet.json
```diff
@@ -0,0 +1,20 @@
...
+    "description": "A schema which validates params accepted by the Medianet adapter",
+    "type": "object",
+    "properties": {
+      "cid": {
+        "type": "string",
+        "description": "The customer id provided by Media.net."
+      },
+      "crid": {
+        "type": "string",
+        "description": "The placement id provided by Media.net."
```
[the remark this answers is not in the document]
eshaanmangal: I think we still need to add `len(crid) == 0` since we need to throw medianet custom error message.

static/bidder-params/medianet.json
```diff
@@ -0,0 +1,20 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Medianet Adapter Params",
+    "description": "A schema which validates params accepted by the Medianet adapter",
+    "type": "object",
+    "properties": {
+      "cid": {
+        "type": "string",
+        "description": "The customer id provided by Media.net."
```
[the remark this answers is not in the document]
eshaanmangal: I think we still need to add len(cid) == 0 since we need to throw medianet custom error message.

static/bidder-info/medianet.yaml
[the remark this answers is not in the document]
VeronikaSolovei9: UserMacro should be hardcoded. PBS doesn't support this value as a placeholder. With current URL I have an error at server startup: `cannot create syncer for bidder medianet with key medianet: redirect template: medianet_usersync_url:1:160: executing "medianet_usersync_url" at <.UserMacro>: can't evaluate field UserMacro in type macros.UserSyncTemplateParams`

Please refer to this paragraph for detailed info: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#user-sync-configuration

adapters/medianet/medianet.go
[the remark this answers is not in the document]
guscarreon: Ok. If the custom error messages are required, lets leave the checks in lines 110 to 116 and not add the `"minLength"` entries to `static/bidder-params/medianet.json` because those entries would make the true banches of said `if` statements unreachable.

Thank you for clarifying this point.

static/bidder-params/medianet.json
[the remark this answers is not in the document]
guscarreon: Agree. As discussed above, lets not add the `"minLength"` entries and leave lines 110 to 166 of `adapters/medianet/medianet.go` as they are.

static/bidder-params/medianet.json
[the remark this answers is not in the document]
guscarreon: Agree. As discussed above, lets not add the `"minLength"` entries and leave lines 110 to 166 of `adapters/medianet/medianet.go` as they are.

adapters/medianet/medianet.go
guscarreon: Thank you

adapters/medianet/medianet.go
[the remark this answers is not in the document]
VeronikaSolovei9: No error is expected in this case. Request.imp.ext validation with bidder param schema happens in Action code: action.go -> func validateImpExt -> deps.paramsValidator.Validate. In unit tests, bidder code (bidder.MakeRequests) executes directly from unit test itself. This is why json schema validation will not work. 
There is a params_test.go to do this. 

For real request this code will never be executed and any of those messages will ever be returned because of json schema validation happens first.

adapters/medianet/medianet.go
```diff
@@ -0,0 +1,135 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type MedianetAdapter struct {
```
VeronikaSolovei9, line 15: please rename it to `adapter`

config/config.go
```diff
@@ -892,6 +892,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.madvertise.endpoint", "https://mobile.mng-ads.com/bidrequest{{.ZoneID}}")
 	v.SetDefault("adapters.marsmedia.endpoint", "https://bid306.rtbsrv.com/bidder/?bid=f3xtet")
 	v.SetDefault("adapters.mediafuse.endpoint", "http://ghb.hbmp.mediafuse.com/pbs/ortb")
+	v.SetDefault("adapters.medianet.endpoint", "https://prebid-adapter.media.net/rtb/pb/prebids2s?src={{.Host}}")
```
VeronikaSolovei9, line 895: I don't see any logic in adapter code to handle `{{.Host}}` placeholder. 
Please refer to this example: adkernel.go -> func `buildEndpointURL`

static/bidder-info/medianet.yaml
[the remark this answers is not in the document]
VeronikaSolovei9: This is also a reason why validation job fails

adapters/medianet/medianet.go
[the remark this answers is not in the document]
eshaanmangal: Thanks

static/bidder-info/medianet.yaml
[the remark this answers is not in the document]
eshaanmangal: Resolved !!

static/bidder-params/medianet.json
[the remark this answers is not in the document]
eshaanmangal: Thanks @guscarreon, I have fixed medianet.go and is using minLength from *.json only

adapters/medianet/medianet.go
eshaanmangal: Done !!

config/config.go
eshaanmangal: Removed, no need for it

static/bidder-info/medianet.yaml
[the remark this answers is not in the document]
SyntaxNode: Please remove the `key: medianet` and `supportCors: false` values. They are defaults and should not be provided.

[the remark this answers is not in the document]
eshaanmangal: Resolved !!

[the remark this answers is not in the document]
eshaanmangal: Resolved !!

guscarreon (approved):
LGTM. Thank you for addressing my feedback.

adapters/medianet/medianet.go
```diff
@@ -0,0 +1,127 @@
...
+
+type adapter struct {
+	endpoint string
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errs []error
+
+	for _, imp := range request.Imp {
+		if err := preprocess(&imp); err != nil {
```
VeronikaSolovei9, line 23: Thank you for addressing comments. Code looks good to me. The only comment - do you really need this `preprocess` function at all? All it does - it unmarshalls imp ext and checks there are no errors. This part is covered by imp validator already. Please correct me if I miss something.

VeronikaSolovei9 (commented):
Overall LGTM, just one minor comment about `preprocess` function in MakeRequests.  
May I ask you review current Medianet documentation and ensure it's up to date:
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/medianet.md
Looks like you only support `cid` and `crid` as bidder params.

adapters/medianet/medianet.go
eshaanmangal: Yes, you are right, earlier I added this function for unit test cases (for better coverage) in which we directly can Bidder. In future iterations, we have some plans to use that function in a much better way.

eshaanmangal:
Yes @VeronikaSolovei9, `https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/medianet.md` is up to date.

adapters/medianet/medianet.go
eshaanmangal: UPDATE: I have removed `preprocess`

VeronikaSolovei9 (approved):
Thank you, LGTM

config/config.go
```diff
@@ -892,6 +892,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("adapters.madvertise.endpoint", "https://mobile.mng-ads.com/bidrequest{{.ZoneID}}")
 	v.SetDefault("adapters.marsmedia.endpoint", "https://bid306.rtbsrv.com/bidder/?bid=f3xtet")
 	v.SetDefault("adapters.mediafuse.endpoint", "http://ghb.hbmp.mediafuse.com/pbs/ortb")
+	v.SetDefault("adapters.medianet.endpoint", "https://prebid-adapter.media.net/rtb/pb/prebids2s")
+	v.SetDefault("adapters.medianet.extra_info", v.GetString("external_url"))
```
guscarreon, line 896: What's the purpose of this function call? Are we trying to set the Medianet adapter's `config.ExtraAdapterInfo` to `"http://localhost:8000"`? Is this a local test leftover code? Assuming these default values are used, your adapter's endpoint would be: `"https://prebid-adapter.media.net/rtb/pb/prebids2s?src=http%3A%2F%2Flocalhost%3A8000"`.

adapters/medianet/medianet.go
```diff
@@ -0,0 +1,118 @@
...
+func buildEndpoint(mnetUrl, hostUrl string) string {
+
+	urlObject, err := url.Parse(mnetUrl)
+	if err != nil || len(hostUrl) == 0 {
+		return mnetUrl
+	}
+	values := urlObject.Query()
+	values.Add("src", hostUrl)
+	urlObject.RawQuery = values.Encode()
+	return urlObject.String()
```
guscarreon, line 117: Could we consider additional scenarios inside `adapters/medianet/medianet_test.go` in order to test this function further? 

Current coverage:
image

config/config.go
eshaanmangal: We are passing  external_url as a param to request in MediaNet endpoint for a feature that MediaNet has implemented at their end. This is not a local test leftover code. default values are for local env but when we will use any config file (pbs.yaml or pbs.json) then in that case external_url will be overridden in viper.

adapters/medianet/medianet.go
eshaanmangal: sure let me add those.

eshaanmangal: Done, Thanks @guscarreon

config/config.go
VeronikaSolovei9: At the runtime it will be PBS prod url, not localhost.

guscarreon: Prebid Server will officially support this feature very soon, as discussed in issue #1739 which is still open and pending to be coded. As of right now, can you please either:
 1. Hardcode the Host URL as the default value of `external_url` field
```
894     v.SetDefault("adapters.mediafuse.endpoint", "http://ghb.hbmp.mediafuse.com/pbs/ortb")
895     v.SetDefault("adapters.medianet.endpoint", "https://prebid-adapter.media.net/rtb/pb/prebids2s")
896 -   v.SetDefault("adapters.medianet.extra_info", v.GetString("external_url"))
    +   v.SetDefault("adapters.medianet.extra_info", "https://medianets.default.host.com")
897     v.SetDefault("adapters.mgid.endpoint", "https://prebid.mgid.com/prebid/")
898     v.SetDefault("adapters.mobilefuse.endpoint", "http://mfx.mobilefuse.com/openrtb?pub_id={{.PublisherID}}")
899     v.SetDefault("adapters.mobfoxpb.endpoint", "http://bes.mobfox.com/?c=__route__&m=__method__&key=__key__")
config/config.go
```
 2. Or delete this default value entry entirely and just make sure implementations of prebid server are configured with this field. If so, make your adapter disabled by default because check `if err != nil || len(hostUrl) == 0` in line 111 of `adapters/medianet/medianet.go` discards empty host url strings and errors out upon Prebid Server start if this value is not found.
```
894     v.SetDefault("adapters.mediafuse.endpoint", "http://ghb.hbmp.mediafuse.com/pbs/ortb")
895     v.SetDefault("adapters.medianet.endpoint", "https://prebid-adapter.media.net/rtb/pb/prebids2s")
896 -   v.SetDefault("adapters.medianet.extra_info", v.GetString("external_url"))
    +   v.SetDefault("adapters.medianet.disabled", true)
897     v.SetDefault("adapters.mgid.endpoint", "https://prebid.mgid.com/prebid/")
898     v.SetDefault("adapters.mobilefuse.endpoint", "http://mfx.mobilefuse.com/openrtb?pub_id={{.PublisherID}}")
899     v.SetDefault("adapters.mobfoxpb.endpoint", "http://bes.mobfox.com/?c=__route__&m=__method__&key=__key__")
config/config.go
```

adapters/medianet/medianet.go
guscarreon: Thanks for adding coverage @eshaanmangal. Another ask in regards of this function: If `hostUrl` comes empty there seems to not be need to call `url.Parse(mnetUrl)` at all. Can we slightly optimize?
```
108   func buildEndpoint(mnetUrl, hostUrl string) string {
109  
    +     if len(hostUrl) == 0 {
    +         return mnetUrl
    +     }
110       urlObject, err := url.Parse(mnetUrl)
111 -     if err != nil || len(hostUrl) == 0 {
    +     if err != nil {
112           return mnetUrl
113       }
114       values := urlObject.Query()
115       values.Add("src", hostUrl)
116       urlObject.RawQuery = values.Encode()
117       return urlObject.String()
118   }
adapters/medianet/medianet.go
```

adapters/medianet/medianet.go
```diff
@@ -0,0 +1,118 @@
...
+	endpoint string
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errs []error
+
+	reqJson, err := json.Marshal(request)
+	if err != nil {
+		errs = append(errs, err)
+		return nil, errs
```
guscarreon, line 26: There seems to not be need for `errs` at all if we manage this error scenario just like `MakeBids` does in line 62. We could just:
```
20   func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
21 -     var errs []error
22   
23       reqJson, err := json.Marshal(request)
24       if err != nil {
25 -         errs = append(errs, err)
26 -         return nil, errs
   +         return nil, []error{err}
27       }
28   
adapters/medianet/medianet.go
```

guscarreon (commented):
Thanks for addressing the feedback. The code looks pretty good just left a couple last observations.

adapters/medianet/medianet.go
eshaanmangal: yes that's valid point but the reason I changed it assuming (might be over-thinking), what if some custom implementation of prebid server (manage prebid) set spammed medianet url in pbs.yaml (config) file

adapters/medianet/medianet.go
eshaanmangal: Lets keep it in place for direct Bidder request unit test cases, thru that one can test multiple request formats which medianet will accept. If I remove it, then there is not need to test cases and I wont be able to test medianet adapter as an individual module while making tampered requests from unit test cases

eshaanmangal: @guscarreon , https://github.com/prebid/prebid-server/pull/2133#discussion_r792881094 issue addressed, in which I have added `if len(hostUrl) == 0 {
    +         return mnetUrl
    +     }`
    before URL parser

eshaanmangal:
@guscarreon @VeronikaSolovei9 Can we merge it ?

guscarreon:
> @guscarreon @VeronikaSolovei9 Can we merge it ?

@eshaanmangal this comment https://github.com/prebid/prebid-server/pull/2133#discussion_r792877560 still has not been addressed or at least its update has not been pushed

eshaanmangal:
@guscarreon in comment [#2133 (comment)](https://github.com/prebid/prebid-server/pull/2133#discussion_r792877560) 
Point 1.  Hardcode the Host URL as the default value of external_url field ---> Why we will do it ? I mean we need server's Host URL in externa_url. Why will we hardcode it to `https://medianets.default.host.com` ?  
Point 2.  Or delete this default value entry entirely and just make sure implementations of prebid server are configured with this field. If so, make your adapter disabled by default because check if err != nil || len(hostUrl) == 0 in line 111 of adapters/medianet/medianet.go discards empty host url strings and errors out upon Prebid Server start if this value is not found. ------> Why there is need to disable medianet adapter in case when I am not able to fetch external_url, we are not throwing any error.

hhhjort:
> @guscarreon in comment [#2133 (comment)](https://github.com/prebid/prebid-server/pull/2133#discussion_r792877560) Point 1. Hardcode the Host URL as the default value of external_url field ---> Why we will do it ? I mean we need server's Host URL in externa_url. Why will we hardcode it to `https://medianets.default.host.com` ? In cofig file at line 651 we got `external_url` as `v.SetDefault("external_url", "http://localhost:8000")` I need this viper env variable if it is present. In case of manage prebid server this external_url will contain the url of host which we need. Point 2. Or delete this default value entry entirely and just make sure implementations of prebid server are configured with this field. If so, make your adapter disabled by default because check if err != nil || len(hostUrl) == 0 in line 111 of adapters/medianet/medianet.go discards empty host url strings and errors out upon Prebid Server start if this value is not found. ------> Why there is need to disable medianet adapter in case when I am not able to fetch external_url, we are not throwing any error.
> 
> In local testing things are working as expected.

The above will work in certain circumstances, but it is very fragile. Setting a default this way depends strongly on how viper reads its parameters and handles defaults. The way you have written it, it depends on the final value of "external_url" being loaded into viper before the `SetDefault()` call for your extra info. It might be true today that every code path that loads config data does so before we get to this line. I have not dug into the code to confirm or deny that this is true. But the naive assumption, at least in my eyes, is that defaults are set before loading the "real" values. So unless someone knows your config code depends on defaults being set last, it is quite possible a future PR around setting config parameters will have code that reads some config parameters after this code executes.

Having an incorrect extra info because of relying on the default value for extra info which relies on having the final value for external url will be very tricky to resolve. One would not normally think that the default value of a parameter depends on the final value of another config parameter. Now this may not be a fatal error in your case, but I would also like to discourage that an example of such a construction existing in the code, so someone else will not copy it into a case where the failure is much more problematic for operations. The existence of one example implies that the example will work, not that the example may fail, but the failure is minor in that particular case.

If you need to work the PBS hosting entity as some sort of source in the requests being sent to your bidder, I would suggest either adding the external_url as a parameter being passed to bidders in MakeRequests(), or just adding it in a header that all outgoing bid requests would have. Or adding another more explicit config parameter such as PBS_HOST to be used to identify the PBS hosting entity in a bit more straightforward manner. I would lean towards the header idea, the utility of external_url doesn't seem to be worth adding it to the MakeRequests() signature.

eshaanmangal:
@hhhjort I have added static url for now, will wait for https://github.com/prebid/prebid-server/issues/1739

bretg:
@eshaanmangal - I tried emailing prebid@media.net and it bounced. Please either fix that email address or provide a different one. Thanks.

eshaanmangal:
> @eshaanmangal - I tried emailing [prebid@media.net](mailto:prebid@media.net) and it bounced. Please either fix that email address or provide a different one. Thanks.

@bretg Updated

bretg:
Thanks @eshaanmangal -- the error message is different now

```
We're writing to let you know that the group you tried to contact (prebid-support) may not exist, or you may not have permission to post messages to the group. 
```

eshaanmangal:
> Thanks @eshaanmangal -- the error message is different now
> 
> ```
> We're writing to let you know that the group you tried to contact (prebid-support) may not exist, or you may not have permission to post messages to the group. 
> ```

Sorry for the inconvenience @bretg , there was issue at our end, can you try again.

VeronikaSolovei9:
Hi @eshaanmangal, I don't have any more comments related to code. I try to understand why do you need to pass this static value from config. It made sense to me at the beginning when you passed PBS host URL, you may need it to track where traffic comes from. But now you pass this URL: `https://medianet.golang.pbs.com` which looks like URL of your own PrebidServer. Is this what you are trying to do?

eshaanmangal:
> #1739

Hi @VeronikaSolovei9 , tbh we still need PBS host URL to track where traffic is coming from and set desired traffic is coming from so that we can use DC which is closer to that PBS host (for fast response), so now by adding this static address (dummy address) we will first check how much actually traffic we will get from golang pbs and will try to come up with sol (we are looking forward for https://github.com/prebid/prebid-server/issues/1739)

guscarreon (approved):
LGTM

VeronikaSolovei9 (approved):
LGTM

PR #2135, opened 2022-01-18
Richaudience: Add Video and Multibid

richaudience:


adapters/richaudience/richaudience.go
```diff
@@ -26,36 +26,14 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
 }
 
 func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
-	raiHeaders := http.Header{}
-
-	setHeaders(&raiHeaders)
-
 	isUrlSecure := getIsUrlSecure(request)
```
hhhjort, line 29: You don't seem to use this except to pass to `setImp()`, and `setImp()` is already being passed `request`. Any reason to not move it into this function?

adapters/richaudience/richaudience.go
```diff
@@ -26,36 +26,14 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
 
-	resImps, err := setImp(request, isUrlSecure)
+	requestDataRequest, err := setImp(request, isUrlSecure, a)
```
hhhjort, line 31: Is there a reason you are passing the adapter into this function rather than making it an adapter method?

adapters/richaudience/richaudience.go
```diff
@@ -200,3 +199,15 @@ func parseImpExt(imp *openrtb2.Imp) (*openrtb_ext.ExtImpRichaudience, error) {
 
 	return &richaudienceExt, nil
 }
+
+func getMediaType(impId string, imps []openrtb2.Imp) openrtb_ext.BidType {
```
hhhjort, line 203: This implies that, for a multiformat request that includes video, you will always bid a video ad or not bid at all. You may want to consider passing an explicit media type to give your bidder more flexibility in bidding. Also in OpenRTB 2.6, there will be an `mType` parameter to explicitly provide this.

hhhjort:
You should also add a couple of JSON tests, one for video and one for multiformat. Beyond the actual testing they enable, tests in /exemplary are useful for publishers in showing how they should configure the requests for your adapters.

adapters/richaudience/richaudience.go
richaudience: I have resolved this in next commit, thanks for your comments

adapters/richaudience/richaudience.go
richaudience: I have resolved this in next commit, thanks for your comments

richaudience:
I have added one test for video bid. Please check It

adapters/richaudience/richaudience.go
hhhjort: In an openrtb2 request, you can have a multiformat impression. For example the `banner` object and the `video` object may both be populated in an imp. This means that you could bid with an ad that is a banner, or an ad that is a video. The bid has to be one or the other. It is an issue with openrtb 2.5 that this ambiguity in the bid exists, prebid server is trying to resolve this issue by asking bidders to be explicit. Openrtb 2.6 will fix this by including an explicit `mType` field for advertisers to declare what sort of bid they have supplied. But for now, we are working with what we have.

This function you have here assumes that in a banner/video multiformat impression, you will bid a video or not at all. If this is true, you are limiting your options on bidding. If this is false, and you might bid a banner when you have a high value banner ad to deliver, then you are compounding the ambiguity problem by reporting video when you are actually bidding with a banner ad. In the first case, this code is fine, but you are limiting yourself. In the second case, we ask that you find a way to be more accurate in your bid type reporting.

richaudience: Thanks for your explain, I think understood.
I have resolved this in next commit, thanks for your comments

hhhjort: This is worse. Now it will only check the last imp (It loops through all the imps, but the last imp will overwrite the other results) and applies that media type to all imps. Please at least revert to the previous code.

My point above has nothing to do with multiple imps. A single imp that defines both banner and video gives you the option of bidding with either a banner ad or a video ad. Your code assumes if the video option is given, you will always bid with video and not consider a banner ad instead. This either limits your options, or results you in reporting the wrong media type. Knowing the correct video type may be important in properly displaying an ad, which is why Prebid Server reports this field. So flagging this incorrectly could result in some of your ads not being displayed properly.

richaudience:
I have discarded the changes, because I think best to leave a default bid. I have compared with others adapters and I see same configuration. ¿It's correct? 
Kind Regards

adapters/richaudience/richaudience.go
richaudience: I have discarded the changes, because I think best to leave a default bid. I have compared with others adapters and I see same configuration. ¿It's correct?
Kind Regards

hhhjort: Yes. My point is that there may be a better way, not that what you had is wrong.

hhhjort (dismissed):
LGTM

adapters/richaudience/richaudience.go
```diff
@@ -26,36 +26,85 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
+		if raiExt != nil {
+			if raiExt.Pid != "" {
+				imp.TagID = raiExt.Pid
+			}
 
-	if err = validateDevice(request); err != nil {
-		return nil, []error{err}
-	}
+			if raiExt.Test {
+				request.Device.IP = "11.222.33.44"
```
guscarreon, line 54: It seems that we should move the `validateDevice(request)` call before this line in order to avoid a potential `nil` pointer de-reference. Maybe even before the `parseImpExt(&imp)` call given that `validateDevice` is less resorce-consuming than the unmarshal operations that `parseImpExt(&imp)` performs and equally makes the function return with an error.
```
28   func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
29       var requestDataRequest []*adapters.RequestData
30   *--  7 lines: errs := make([]error, 0, len(request.Imp))---------------------------
37  
38       for _, imp := range request.Imp {
39           var secure = int8(0)
   +         if err = validateDevice(request); err != nil {
   +             errs = append(errs, &errortypes.BadInput{
   +                 Message: err.Error(),
   +             })
   +             return nil, errs
   +         }
40           raiExt, err := parseImpExt(&imp)
41           if err != nil {
42               errs = append(errs, &errortypes.BadInput{
43                   Message: err.Error(),
44               })
45               return nil, errs
46           }
47  
48           if raiExt != nil {
49               if raiExt.Pid != "" {
50                   imp.TagID = raiExt.Pid
51               }
52  
53               if raiExt.Test {
54                   request.Device.IP = "11.222.33.44"
55                   request.Test = int8(1)
56               }
57  
58               if raiExt.BidFloorCur != "" {
59                   imp.BidFloorCur = raiExt.BidFloorCur
60               } else if imp.BidFloorCur == "" {
61                   imp.BidFloorCur = "USD"
62               }
63           }
64   *-- 17 lines: if isUrlSecure {-----------------------------------------------------
81           request.Imp = []openrtb2.Imp{imp}
82  
83 -         if err = validateDevice(request); err != nil {
84 -             errs = append(errs, &errortypes.BadInput{
85 -                 Message: err.Error(),
86 -             })
87 -             return nil, errs
88 -         }
89  
90           req, err := json.Marshal(request)
91           if err != nil {
92               errs = append(errs, &errortypes.BadInput{
adapters/richaudience/richaudience.go
```
Thougts?

guscarreon: Also, can we add a JSON test where `raiExt.Test` is set to `true` and `request.Device` is `nil`? Just to assert we got this scenario segfault free

guscarreon (commented):
Looks good, just have a question

adapters/richaudience/richaudience.go
richaudience: I have changed the position of function, and add the testCase for the scenario

hhhjort (approved):

adapters/richaudience/richaudience.go
guscarreon: @richaudience I'm very sorry for this confusion, I got the location of the `validateDevice(request)` wrong in my previous comment. I meant to make the `validateDevice(request)` call from outside the for loop because there's no need to execute this function more than once since it doesn't perform any checks on the `request.Imp` elements. What do you think?
```
 28   func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
 29       var requestDataRequest []*adapters.RequestData
 30  
 31       errs := make([]error, 0, len(request.Imp))
 32  
 33       raiHeaders := http.Header{}
 34       setHeaders(&raiHeaders)
 35  
 36       isUrlSecure := getIsUrlSecure(request)
 37  
    +     if err := validateDevice(request); err != nil {
    +         errs = append(errs, &errortypes.BadInput{
    +             Message: err.Error(),
    +         })
    +         return nil, errs
    +     }
    +        
 38       for _, imp := range request.Imp {
 39           var secure = int8(0)
 40  
 41 -         if err := validateDevice(request); err != nil {
 42 -             errs = append(errs, &errortypes.BadInput{
 43 -                 Message: err.Error(),
 44 -             })
 45 -             return nil, errs
 46 -         }
 47  
 48           raiExt, err := parseImpExt(&imp)
 49           if err != nil {
 50               errs = append(errs, &errortypes.BadInput{
 51                   Message: err.Error(),
 52               })
 53               return nil, errs
 54           }
 55  
 56   *-- 43 lines: if raiExt != nil {-------------------------------------------------------------------------------------------
 99           requestDataRequest = append(requestDataRequest, &adapters.RequestData{
100               Method:  "POST",
101               Uri:     a.endpoint,
102               Body:    req,
103               Headers: raiHeaders,
104           })
105  
106       }
107  
108       return requestDataRequest, nil
109   }
adapters/richaudience/richaudience.go
```

guscarreon (commented):
@richaudience I'm very sorry for this confusion, I got the location of the `validateDevice(request)` wrong in my previous comment. I meant to make the `validateDevice(request)` call from outside the for loop because there's no need to execute this function more than once since it doesn't perform any checks on the `request.Imp` elements. What do you think?
```
 28   func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
 29       var requestDataRequest []*adapters.RequestData
 30  
 31       errs := make([]error, 0, len(request.Imp))
 32  
 33       raiHeaders := http.Header{}
 34       setHeaders(&raiHeaders)
 35  
 36       isUrlSecure := getIsUrlSecure(request)
 37  
    +     if err := validateDevice(request); err != nil {
    +         errs = append(errs, &errortypes.BadInput{
    +             Message: err.Error(),
    +         })
    +         return nil, errs
    +     }
    +        
 38       for _, imp := range request.Imp {
 39           var secure = int8(0)
 40  
 41 -         if err := validateDevice(request); err != nil {
 42 -             errs = append(errs, &errortypes.BadInput{
 43 -                 Message: err.Error(),
 44 -             })
 45 -             return nil, errs
 46 -         }
 47  
 48           raiExt, err := parseImpExt(&imp)
 49           if err != nil {
 50               errs = append(errs, &errortypes.BadInput{
 51                   Message: err.Error(),
 52               })
 53               return nil, errs
 54           }
 55  
 56   *-- 43 lines: if raiExt != nil {-------------------------------------------------------------------------------------------
 99           requestDataRequest = append(requestDataRequest, &adapters.RequestData{
100               Method:  "POST",
101               Uri:     a.endpoint,
102               Body:    req,
103               Headers: raiHeaders,
104           })
105  
106       }
107  
108       return requestDataRequest, nil
109   }
adapters/richaudience/richaudience.go
```

richaudience:
> @richaudience I'm very sorry for this confusion, I got the location of the `validateDevice(request)` wrong in my previous comment. I meant to make the `validateDevice(request)` call from outside the for loop because there's no need to execute this function more than once since it doesn't perform any checks on the `request.Imp` elements. What do you think?
> 
> ```
>  28   func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
>  29       var requestDataRequest []*adapters.RequestData
>  30  
>  31       errs := make([]error, 0, len(request.Imp))
>  32  
>  33       raiHeaders := http.Header{}
>  34       setHeaders(&raiHeaders)
>  35  
>  36       isUrlSecure := getIsUrlSecure(request)
>  37  
>     +     if err := validateDevice(request); err != nil {
>     +         errs = append(errs, &errortypes.BadInput{
>     +             Message: err.Error(),
>     +         })
>     +         return nil, errs
>     +     }
>     +        
>  38       for _, imp := range request.Imp {
>  39           var secure = int8(0)
>  40  
>  41 -         if err := validateDevice(request); err != nil {
>  42 -             errs = append(errs, &errortypes.BadInput{
>  43 -                 Message: err.Error(),
>  44 -             })
>  45 -             return nil, errs
>  46 -         }
>  47  
>  48           raiExt, err := parseImpExt(&imp)
>  49           if err != nil {
>  50               errs = append(errs, &errortypes.BadInput{
>  51                   Message: err.Error(),
>  52               })
>  53               return nil, errs
>  54           }
>  55  
>  56   *-- 43 lines: if raiExt != nil {-------------------------------------------------------------------------------------------
>  99           requestDataRequest = append(requestDataRequest, &adapters.RequestData{
> 100               Method:  "POST",
> 101               Uri:     a.endpoint,
> 102               Body:    req,
> 103               Headers: raiHeaders,
> 104           })
> 105  
> 106       }
> 107  
> 108       return requestDataRequest, nil
> 109   }
> adapters/richaudience/richaudience.go
> ```

Yes, I am sorry for this confusion. I think you are right, we have to change this allocation function.
Please check it

guscarreon (approved):
LGTM. @richaudience thank you for addressing my feedback

hhhjort (approved):
LGTM

PR #2141, opened 2022-01-25
Invibes Bid Adapter - replaced Host macro with ZoneID

rcheptanariu:
Invibes Bid Adapter update for prebid-server

Removed {Host} macro and replaced with {ZoneID}
No documentation update necessary since this is internal use only.
Maintainer: system_operations@invibes.com

### DIFF
--- a/adapters/invibes/invibes.go
+++ b/adapters/invibes/invibes.go
@@ -279,19 +279,17 @@ func (a *InvibesAdapter) makeParameter(invibesParams InvibesInternalParams, requ
 }
 
 func (a *InvibesAdapter) makeURL(request *openrtb2.BidRequest, domainID int) (string, error) {
-	host := "bid.videostep.com"
-	if domainID == 1 {
-		host = "adweb.videostepstage.com"
-	} else if domainID == 2 {
-		host = "adweb.invibesstage.com"
-	} else if domainID == 1001 {
-		host = "bid.videostep.com"
-	} else if domainID >= 1002 {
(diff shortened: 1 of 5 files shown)


bsardo (approved):

adapters/invibes/invibes.go
```diff
@@ -279,19 +279,17 @@ func (a *InvibesAdapter) makeParameter(invibesParams InvibesInternalParams, requ
...
-	host := "bid.videostep.com"
-	if domainID == 1 {
-		host = "adweb.videostepstage.com"
-	} else if domainID == 2 {
-		host = "adweb.invibesstage.com"
-	} else if domainID == 1001 {
-		host = "bid.videostep.com"
-	} else if domainID >= 1002 {
-		host = "bid" + strconv.Itoa(domainID-1000) + ".videostep.com"
+	subdomain := "bid"
```
mansinahar, line 282: This default value `bid` for the variable `subdomain` here doesn't make sense as you're setting it to a different value based on different domain IDs. You can rather just do `var subdomain string` to declare the variable

mansinahar:
@rcheptanariu Look like when you merged this on top of master, it inherited a bunch of commits and code changes from master as part of this PR as well. Can you please rather rebase on top of master and do a force push so that only your changes are included in the PR? We usually discourage force pushing but in this case as it's a small PR with not a lot of changes, this might be the easiest way.

mansinahar:
@rcheptanariu Just a friendly reminder to address the comment above

mansinahar (approved):

bsardo (approved):
LGTM

PR #2146, opened 2022-02-03
GDPR: Add account-level purpose and feature overrides

bsardo:
This PR adds account-level purpose and special feature config options that override host-level configurations when present when processing requests on the ORTB2 endpoints.
At the account-level, we can now define whether a purpose should be enforced, whether we are enforcing vendors for a given purpose and any vendor exceptions for that purpose (see `config/accounts.go`).

GDPR also applies to the cookie sync and setuid endpoints but only the host-level GDPR config info is relevant. As a result, a GDPR object (instantiated through `NewPermissions()`) is still created at startup and passed into those endpoints with the host-level config. However, in each of the ORTB2 endpoint handlers, a GDPR object is now created for each request with an aggregate config consisting of the host-level and account-level configs.


Jmp24 (dismissed):

Jmp24:
Jm

config/accounts.go
```diff
@@ -39,20 +44,147 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
 	return a.Enabled
 }
+// PurposeEnforced checks if full enforcement is turned on for a given purpose at the account level. It returns the 
+// enforcement strategy type and whether or not it is set on the account. If not set, a default value of true is 
+// returned matching host default behavior.
+func (a *AccountGDPR) PurposeEnforced(purpose consentconstants.Purpose) (value, exists bool) {
+	if a.PurposeConfigs[purpose] == nil {
+		return true, false
+	}
+	if a.PurposeConfigs[purpose].EnforcePurpose == TCF2FullEnforcement {
```
hhhjort, line 83: TCF2FullEnforcement and TCF2NoEnforcement suggest in the future that there will be more options in the future. If so, then `bool` will not work when we have more values.

gdpr/gdpr.go
```diff
@@ -24,45 +21,22 @@ type Permissions interface {
...
-		4:  cfg.TCF2.Purpose4,
-		5:  cfg.TCF2.Purpose5,
-		6:  cfg.TCF2.Purpose6,
-		7:  cfg.TCF2.Purpose7,
-		8:  cfg.TCF2.Purpose8,
-		9:  cfg.TCF2.Purpose9,
-		10: cfg.TCF2.Purpose10,
-	}
-
 	permissionsImpl := &permissionsImpl{
```
bsardo, line 33: Before we were passing the top-level GDPR config `config.GDPR` into this function and setting it on the GDPR permissions object. Now we pass in `config.GDPR` as well as `tcf2Config`, which implements the `TCF2ConfigReader` interface, and only set the `tcf2Config` on the object while extracting a couple of values from `config.GDPR` to set directly on the object. IMO, having both configs on the object would be confusing. I also figured it would be best to do it this way instead of just setting `config.GDPR` on the object where `config.GDPR` implements the `TCF2ConfigReader` interface as I thought the interface methods made more send on the `config.GDPR.TCF2` object. Thoughts?

gdpr/gdpr.go
```diff
@@ -24,45 +21,22 @@ type Permissions interface {
...
-		cfg:              cfg,
-		gdprDefaultValue: gdprDefaultValue,
-		purposeConfigs:   purposeConfigs,
-		vendorIDs:        vendorIDs,
-		fetchVendorList: map[uint8]func(ctx context.Context, id uint16) (vendorlist.VendorList, error){
```
bsardo, line 64: The vendor list fetcher was previously created when the single GDPR permissions object was created at startup in the router and was tightly coupled. Now, since we need to instantiate a permissions object at startup for cookie_sync and setuid endpoints and also on a per request basis for the ORTB2 endpoints, I had to decouple the fetcher from the permissions object as I figured it made sense to continue having just a single instance of the vendor list fetcher. As a result, the fetcher is created at startup using the same functional programming style we've always used to create it and then passed into the GDPR permissions object created at startup and those created during the auction.

gdpr/impl.go
```diff
@@ -10,51 +10,51 @@ import (
...
+	gdprDefaultValue      string
+	hostVendorID          int
+	nonStandardPublishers map[string]struct{}
+	cfg                   TCF2ConfigReader
+	vendorIDs             map[openrtb_ext.BidderName]uint16
 }
 
 func (p *permissionsImpl) HostCookiesAllowed(ctx context.Context, gdprSignal Signal, consent string) (bool, error) {
-	gdprSignal = SignalNormalize(gdprSignal, p.cfg)
+	gdprSignal = SignalNormalize(gdprSignal, config.GDPR{DefaultValue: p.gdprDefaultValue})
```
bsardo, line 27: Instead of creating my own `SignalNormalize` that takes either a GDPR signal value or its string equivalent, I figured I would just wrap the default value specified when the permissions object was created in an instance of `config.GDPR` and pass that in. With this new implementation, `config.GDPR` is no longer available on the object.

gdpr/impl.go
```diff
@@ -160,28 +170,11 @@ func (p *permissionsImpl) allowActivities(ctx context.Context, vendorID uint16,
...
-	return
-}
-
-func (p *permissionsImpl) isSpecialFeatureVendorException(bidder openrtb_ext.BidderName) (vendorException bool) {
-	if _, ok := p.cfg.TCF2.SpecialFeature1.VendorExceptionMap[bidder]; ok {
-		vendorException = true
-	}
-	return
-}
-
```
bsardo, line 176: These functions have been moved onto the `TCF2ConfigReader` implementations `config.GDPR.TCF2` and the aggregated config.

gdpr/impl.go
```diff
@@ -251,7 +244,7 @@ func (p *permissionsImpl) parseVendor(ctx context.Context, vendorID uint16, cons
 		return
 	}
 
-	vendorList, err := p.fetchVendorList[version](ctx, parsedConsent.VendorListVersion())
```
bsardo, line 254: `fetchVendorList` no longer needs to be a map because the only version we support now is TCF2 given TCF1 was removed a few months ago.

gdpr/gdpr.go
```diff
@@ -24,45 +21,22 @@ type Permissions interface {
...
-func NewPermissions(ctx context.Context, cfg config.GDPR, vendorIDs map[openrtb_ext.BidderName]uint16, client *http.Client) Permissions {
+func NewPermissions(cfg config.GDPR, tcf2Config TCF2ConfigReader, vendorIDs map[openrtb_ext.BidderName]uint16, fetcher VendorListFetcher) Permissions {
 	if !cfg.Enabled {
 		return &AlwaysAllow{}
 	}
 
-	gdprDefaultValue := SignalYes
-	if cfg.DefaultValue == "0" {
-		gdprDefaultValue = SignalNo
-	}
```
bsardo, line 44: I decided to just use the string value rather than convert it to a `Signal` constant so we can continue making use of `SignalNormalize` in `impl.go` even though I removed the `config.GDPR` instance from the permissions object.

gdpr/gdpr.go
```diff
@@ -24,45 +21,22 @@ type Permissions interface {
...
-
-	purposeConfigs := map[consentconstants.Purpose]config.TCF2Purpose{
```
bsardo, line 46: This now exists on the `TCF2ConfigReader` host implementation `config.GDPR.TCF2` since we now access all purpose config through it.

exchange/utils.go
```diff
@@ -40,11 +40,11 @@ func cleanOpenRTBRequests(ctx context.Context,
 	req AuctionRequest,
 	requestExt *openrtb_ext.ExtRequest,
 	bidderToSyncerKey map[string]string,
-	gDPR gdpr.Permissions,
+	vendorListFetcher gdpr.VendorListFetcher,
 	metricsEngine metrics.MetricsEngine,
 	gdprDefaultValue gdpr.Signal,
 	privacyConfig config.Privacy,
-	account *config.Account) (allowedBidderRequests []BidderRequest, privacyLabels metrics.PrivacyLabels, errs []error) {
```
bsardo, line 47: `account` was redundant here; it's already available in `req AuctionRequest` above.

exchange/utils.go
```diff
@@ -118,17 +123,9 @@ func cleanOpenRTBRequests(ctx context.Context,
...
 		if gdprEnforced {
-			weakVendorEnforcement := false
-			if account != nil {
-				for _, vendor := range account.GDPR.BasicEnforcementVendors {
-					if vendor == string(bidderRequest.BidderCoreName) {
-						weakVendorEnforcement = true
-						break
-					}
-				}
-			}
```
bsardo, line 129: This logic has been moved. The basic enforcement vendor check has been moved onto the `config.AccountsGDPR` object and is accessed downstream to determine weak vendor enforcement when `AuctionActivitiesAllowed` is called a few lines down from here.

config/accounts.go
```diff
@@ -39,20 +44,147 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+type AccountGDPRSpecialFeature struct {
+	Enforce *bool `mapstructure:"enforce" json:"enforce"`
+	// Array of vendor exceptions that is used to create the hash table VendorExceptionMap so vendor names can be instantly accessed
+	VendorExceptions   []openrtb_ext.BidderName `mapstructure:"vendor_exceptions" json:"vendor_exceptions"`
+	VendorExceptionMap map[openrtb_ext.BidderName]struct{}
+}
+
+// AccountGDPRPurposeOneTreatment represents account-specific GDPR purpose one treatment configuration
+type AccountGDPRPurposeOneTreatment struct {
+	Enabled       *bool `mapstructure:"enabled"`
```
hhhjort, line 185: I am a little worried that Purpose one treatment would seem to default to off unless explicitly enabled in the account config. Is this really what we want to happen?

hhhjort: Nevermind, I see they are bool pointers now.

config/accounts.go
```diff
@@ -39,20 +44,147 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+// enforcement strategy type and whether or not it is set on the account. If not set, a default value of true is 
+// returned matching host default behavior.
+func (a *AccountGDPR) PurposeEnforced(purpose consentconstants.Purpose) (value, exists bool) {
+	if a.PurposeConfigs[purpose] == nil {
+		return true, false
+	}
+	if a.PurposeConfigs[purpose].EnforcePurpose == TCF2FullEnforcement {
+		return true, true
+	}
+	if a.PurposeConfigs[purpose].EnforcePurpose == TCF2NoEnforcement {
```
guscarreon, line 86: I realize this is a matter of personal taste but, given that the `if` statements in lines 84 and 87 are almost identical, and the fact that `TCF2FullEnforcement` and `TCF2NoEnforcement` would be the only two values `a.PurposeConfigs[purpose].EnforcePurpose` is expected to hold, maybe we could reduce a little bit. This is just a nitpick, what do you think? 
```
77   // PurposeEnforced checks if full enforcement is turned on for a given purpose at the account level. It returns the
78   // enforcement strategy type and whether or not it is set on the account. If not set, a default value of true is
79   // returned matching host default behavior.
80   func (a *AccountGDPR) PurposeEnforced(purpose consentconstants.Purpose) (value, exists bool) {
81 -     if a.PurposeConfigs[purpose] == nil {
82 -         return true, false
83 -     }
84 -     if a.PurposeConfigs[purpose].EnforcePurpose == TCF2FullEnforcement {
85 -         return true, true
86 -     }
87 -     if a.PurposeConfigs[purpose].EnforcePurpose == TCF2NoEnforcement {
88 -         return false, true
   +     if a.PurposeConfigs[purpose] != nil {
   +         // Either a.PurposeConfigs[purpose].EnforcePurpose comes with TCF2FullEnforcement or not
   +         return a.PurposeConfigs[purpose].EnforcePurpose == TCF2FullEnforcement, true
89       }
90       return true, false
91   }
92   
config/accounts.go
```
Another nitpick: we are not using the names `value` and `exists`.

config/accounts.go
```diff
@@ -39,20 +44,147 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+
+// PurposeEnforcingVendors gets the account level enforce vendors setting for a given purpose returning the value and
+// whether or not it is set. If not set, a default value of true is returned matching host default behavior.
+func (a *AccountGDPR) PurposeEnforcingVendors(purpose consentconstants.Purpose) (value, exists bool) {
+	if a.PurposeConfigs[purpose] == nil {
+		return true, false
+	}
+	if a.PurposeConfigs[purpose].EnforceVendors == nil {
+		return true, false
+	}
```
guscarreon, line 100: Another nitpick:
```
87   // PurposeEnforcingVendors gets the account level enforce vendors setting for a given purpose returning the value and
88   // whether or not it is set. If not set, a default value of true is returned matching host default behavior.
89   func (a *AccountGDPR) PurposeEnforcingVendors(purpose consentconstants.Purpose) (value, exists bool) {
90 -     if a.PurposeConfigs[purpose] == nil {
91 -         return true, false
92 -     }
93 -     if a.PurposeConfigs[purpose].EnforceVendors == nil {
   +     if a.PurposeConfigs[purpose] == nil || a.PurposeConfigs[purpose].EnforceVendors == nil {
94           return true, false
95       }
96       return *a.PurposeConfigs[purpose].EnforceVendors, true
97   }
98   
config/accounts.go
```

config/accounts.go
```diff
@@ -39,20 +44,147 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+func (a *AccountGDPR) PurposeVendorException(purpose consentconstants.Purpose, bidder openrtb_ext.BidderName) (value, exists bool) {
+	if a.PurposeConfigs[purpose] == nil {
+		return false, false
+	}
+	if a.PurposeConfigs[purpose].VendorExceptionMap == nil {
+		return false, false
+	}
+	_, exists = a.PurposeConfigs[purpose].VendorExceptionMap[bidder]
+
+	return exists, true
```
guscarreon, line 114: Is the order of `value` and `exist` in the function signature wrong? If so, can we correct? If not, can we rename `exists` to `value`?

guscarreon: Another nitpick in regards to the consecutive `if` statements:
```
 99   // PurposeVendorException checks if the given bidder is a vendor exception for a given purpose.
100   func (a *AccountGDPR) PurposeVendorException(purpose consentconstants.Purpose, bidder openrtb_ext.BidderName) (value, exists bool) {
101 -     if a.PurposeConfigs[purpose] == nil {
102 -         return false, false
103 -     }
104 -     if a.PurposeConfigs[purpose].VendorExceptionMap == nil {
    +     if a.PurposeConfigs[purpose] == nil || a.PurposeConfigs[purpose].VendorExceptionMap == nil {
105           return false, false
106       }
107       _, exists = a.PurposeConfigs[purpose].VendorExceptionMap[bidder]
108
109       return exists, true
110   }
config/accounts.go
```

guscarreon (commented):
The first part of my review. Also a conflict should be resolved in `account/account.go`

config/config.go
```diff
@@ -277,21 +278,92 @@ const (
 
 // TCF2 defines the TCF2 specific configurations for GDPR
 type TCF2 struct {
```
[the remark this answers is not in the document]
guscarreon: | TCF2ConfigReader interface method                | *tcf2Config |      *AccountGDPR          |  *TCF2    |
|--------------------------------------------------|-------------|----------------------------|-----------|
|IntegrationEnabled(config.IntegrationType) bool   |    yes      | EnabledForIntegrationType defined instead |   no   |
|IsEnabled()                                       |    yes      |            no              |   yes  |

config/accounts.go
bsardo: Fair points. I deliberately write code in this style for a few reasons (these aren't hard rules or anything):
- favoring early returns
- avoiding nested conditionals wherever possible
- avoiding complex conditionals where possible, which in some cases results in a little extra code but doesn't sacrifice clarity, so that the test coverage tooling can more easily identify key cases that were missed. If we were to embed that conditional in the return statement, the test coverage tool would show that we have full coverage as long as execute `return a.PurposeConfigs[purpose].EnforcePurpose == TCF2FullEnforcement, true` when we probably want to make sure we are testing the case where that conditional is both true and false.

This is up for debate of course.

bsardo: Re: `nitpick: we are not using the names value and exists`:
My general thought here is that given this function is rather concise (~10 lines), such that we can view the whole function on the screen, and there are just a few return values of simple types (bool), specifying named return values without explicitly assigning them keeps the function concise without introducing confusion. This is just my opinion of course.
If we were to say something like:
```
if .... {
    value = true
    exists = false
    return value, exists
}
```
that would be a bit unnecessary IMO. I think that the reader can just glance at the function signature viewable without scrolling to see what the returned bools represent.
If this function was much larger with more complex logic or types I would probably shy away from this approach.
Curious to hear what your thoughts are.

config/accounts.go
bsardo: > Is the order of `value` and `exist` in the function signature wrong? If so, can we correct? If not, can we rename `exists` to `value`?

The order is correct but I see how setting `exists` and then assigning it to `value` is confusing. How about changing `exists` to `found` when looking up `bidder` in the `VendorExceptionMap`:
```
_, found = a.PurposeConfigs[purpose].VendorExceptionMap[bidder]

return found, true
```
The second return value `exists` defined as a named return value represents whether `VendorExceptionMap` was defined. The `value` is whether it was found in the exception map.
to:

config/accounts.go
bsardo: See my thoughts here: https://github.com/prebid/prebid-server/pull/2146#discussion_r819808602. We can discuss offline.

hhhjort (approved):
LGTM

config/config.go
```diff
@@ -584,6 +667,14 @@ func New(v *viper.Viper) (*Configuration, error) {
...
 	}
 
+	// To look for a special feature's vendor exceptions in O(1) time, we fill this hash table with bidders located in the
+	// VendorExceptions field of the GDPR.TCF2.SpecialFeature1 struct defined in this file
+	c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap = make(map[openrtb_ext.BidderName]struct{})
+	for v := 0; v < len(c.GDPR.TCF2.SpecialFeature1.VendorExceptions); v++ {
+		bidderName := c.GDPR.TCF2.SpecialFeature1.VendorExceptions[v]
+		c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap[bidderName] = struct{}{}
+	}
+
```
guscarreon, line 677: This newly added block seems to simply be a repetition of the one above, comments and all
```
              665     // To look for a special feature's vendor exceptions in O(1) time, we fill this hash table with bidders located in the
              666     // VendorExceptions field of the GDPR.TCF2.SpecialFeature1 struct defined in this file
              667     c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap = make(map[openrtb_ext.BidderName]struct{})
              668     for v := 0; v < len(c.GDPR.TCF2.SpecialFeature1.VendorExceptions); v++ {
              669         bidderName := c.GDPR.TCF2.SpecialFeature1.VendorExceptions[v]
              670         c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap[bidderName] = struct{}{}
              671     }
              672
same as above 673     // To look for a special feature's vendor exceptions in O(1) time, we fill this hash table with bidders located in the
same as above 674     // VendorExceptions field of the GDPR.TCF2.SpecialFeature1 struct defined in this file
same as above 675     c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap = make(map[openrtb_ext.BidderName]struct{})
same as above 676     for v := 0; v < len(c.GDPR.TCF2.SpecialFeature1.VendorExceptions); v++ {
same as above 677         bidderName := c.GDPR.TCF2.SpecialFeature1.VendorExceptions[v]
same as above 678         c.GDPR.TCF2.SpecialFeature1.VendorExceptionMap[bidderName] = struct{}{}
same as above 679     }
              680
config/config.go
```

guscarreon (commented):
second part of my review

config/config.go
bsardo: Ah I think the duplicate code block here was introduced when merging with master. Removed.

hhhjort (approved):

guscarreon (commented):
Code looks pretty solid. Mostly nitpicks left below.

config/accounts.go
```diff
@@ -39,20 +44,147 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
 
 // AccountGDPR represents account-specific GDPR configuration
 type AccountGDPR struct {
```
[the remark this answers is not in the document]
bsardo: Sorry, I mistakenly said that `AccountGDPR` is supposed to implement the `TCF2ConfigReader` interface in a previous conversation but that isn't the case.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+	Purpose10                  AccountGDPRPurpose `mapstructure:"purpose10" json:"purpose10"`
+	// Hash table of purpose configs for convenient purpose config lookup
+	PurposeConfigs      map[consentconstants.Purpose]*AccountGDPRPurpose
+	PurposeOneTreatment AccountGDPRPurposeOneTreatment `mapstructure:"purpose_one_treatment" json:"purpose_one_treatment"`
+	SpecialFeature1     AccountGDPRSpecialFeature      `mapstructure:"special_feature1" json:"special_feature1"`
+}
+
+// BasicEnforcementVendor checks if the given bidder is considered a basic enforcement vendor which indicates whether
+// weak vendor enforcement applies to that bidder.
+func (a *AccountGDPR) BasicEnforcementVendor(bidder openrtb_ext.BidderName) (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: As mentioned in a previous comment, `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
-
 	if integrationEnabled := a.IntegrationEnabled.GetByIntegrationType(integrationType); integrationEnabled != nil {
 		return integrationEnabled
 	}
 	return a.Enabled
 }
 
+// FeatureOneEnforced gets the account level feature one enforced setting returning the value and whether or not it
+// was set. If not set, a default value of true is returned matching host default behavior.
+func (a *AccountGDPR) FeatureOneEnforced() (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+	if a.SpecialFeature1.Enforce == nil {
+		return true, false
+	}
+	return *a.SpecialFeature1.Enforce, true
+}
+
+// FeatureOneVendorException checks if the given bidder is a vendor exception.
+func (a *AccountGDPR) FeatureOneVendorException(bidder openrtb_ext.BidderName) (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+	}
+	_, found := a.SpecialFeature1.VendorExceptionMap[bidder]
+
+	return found, true
+}
+
+// PurposeEnforced checks if full enforcement is turned on for a given purpose at the account level. It returns the
+// enforcement strategy type and whether or not it is set on the account. If not set, a default value of true is
+// returned matching host default behavior.
+func (a *AccountGDPR) PurposeEnforced(purpose consentconstants.Purpose) (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+	}
+	if a.PurposeConfigs[purpose].EnforcePurpose == TCF2NoEnforcement {
+		return false, true
+	}
+	return true, false
+}
+
+// PurposeEnforcingVendors gets the account level enforce vendors setting for a given purpose returning the value and
+// whether or not it is set. If not set, a default value of true is returned matching host default behavior.
+func (a *AccountGDPR) PurposeEnforcingVendors(purpose consentconstants.Purpose) (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+		return true, false
+	}
+	if a.PurposeConfigs[purpose].EnforceVendors == nil {
+		return true, false
+	}
+	return *a.PurposeConfigs[purpose].EnforceVendors, true
+}
+
+// PurposeVendorException checks if the given bidder is a vendor exception for a given purpose.
+func (a *AccountGDPR) PurposeVendorException(purpose consentconstants.Purpose, bidder openrtb_ext.BidderName) (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+		return false, false
+	}
+	_, found := a.PurposeConfigs[purpose].VendorExceptionMap[bidder]
+
+	return found, true
+}
+
+// PurposeOneTreatmentEnabled gets the account level purpose one treatment enabled setting returning the value and
+// whether or not it is set. If not set, a default value of true is returned matching host default behavior.
+func (a *AccountGDPR) PurposeOneTreatmentEnabled() (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

config/accounts.go
```diff
@@ -38,23 +43,150 @@ func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType)
...
+	if a.PurposeOneTreatment.Enabled == nil {
+		return true, false
+	}
+	return *a.PurposeOneTreatment.Enabled, true
+}
+
+// PurposeOneTreatmentAccessAllowed gets the account level purpose one treatment access allowed setting returning the
+// value and whether or not it is set. If not set, a default value of true is returned matching host default behavior.
+func (a *AccountGDPR) PurposeOneTreatmentAccessAllowed() (value, exists bool) {
```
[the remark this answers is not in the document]
bsardo: `AccountGDPR` isn't supposed to implement the `TCF2ConfigReader` interface so this is ok.

router/router.go
```diff
@@ -197,7 +197,8 @@ func New(cfg *config.Configuration, rateConvertor *currency.RateConverter) (r *R
 	}
 
 	gvlVendorIDs := bidderInfos.ToGVLVendorIDMap()
-	gdprPerms := gdpr.NewPermissions(context.Background(), cfg.GDPR, gvlVendorIDs, generalHttpClient)
+	vendorListFetcher := gdpr.NewVendorListFetcher(context.Background(), cfg.GDPR, generalHttpClient, gdpr.VendorListURLMaker)
+	gdprPerms := gdpr.NewPermissions(cfg.GDPR, &cfg.GDPR.TCF2, gvlVendorIDs, vendorListFetcher)
```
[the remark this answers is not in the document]
bsardo: I started to make this change and then realized it is probably not worth it right now because I have another PR coming that will move the router GDPR object instantiation into the cookie sync and setuid endpoints, in which case the line `gvlVendorIDs := bidderInfos.ToGVLVendorIDMap()` in the router will no longer be executed twice (once in the router and once when the exchange is created) for the auction/amp/video endpoints.
As you mentioned, this shouldn't add any overhead so I'd rather not make that short-term change here if that's ok with you. This will be addressed though shortly :)

guscarreon (approved):
LGTM. Thanks for addressing the feedback.

hhhjort (approved):
LGTM

PR #2148, opened 2022-02-03
synacormedia - Update YAML with new contact email and usersync endpoint

ecammit:
This updates the synacormedia adapter with:
- contact email uses the new email domain after Synacor Media was acquired by iMedia Brands Inc. In September 2021
- updates the iframe usersync URL

### DIFF
--- a/static/bidder-info/synacormedia.yaml
+++ b/static/bidder-info/synacormedia.yaml
@@ -1,5 +1,5 @@
 maintainer:
-  email: "eng-demand@synacor.com"
+  email: "eng-demand@imds.tv"
 capabilities:
   app:
     mediaTypes:
@@ -11,7 +11,7 @@ capabilities:
       - video
 userSync:
   iframe:
-    url: "https://sync.technoratimedia.com/services?srv=cs&pid=70&cb={{.RedirectURL}}"
+    url: "https://ad-cdn.technoratimedia.com/html/usersync.html?cb={{.RedirectURL}}"
     userMacro: "[USER_ID]"
 userSync:
   # synacormedia supports user syncing, but requires configuration by the host. contact this


static/bidder-info/synacormedia.yaml
```diff
@@ -11,7 +11,7 @@ capabilities:
       - video
 userSync:
   iframe:
-    url: "https://sync.technoratimedia.com/services?srv=cs&pid=70&cb={{.RedirectURL}}"
+    url: "https://ad-cdn.technoratimedia.com/html/usersync.html?cb={{.RedirectURL}}"
```
hhhjort, line 14: I tried testing this redirect out, and I could not find a call to the redirect URL supplied in `cb` ... this isn't qorking properly?

ecammit: It doesn't actually redirect to the cb= URL. Rather, it drops a new call inside the iframe to the cb= URL. We could probably make it redirect the iframe to the cb= URL if that's required, but it should work as-is.

hhhjort: Sorry, I did not mean the whole call should redirect, the calls inside should be fine. What I meant was that I passed `cb=https%3A%2F%2Fib.adnxs.com%2Fprebid%2Fsetuid%3Fbidder%3Dtriplelift%26gdpr%3D%26gdpr_consent%3D%26uid%3D%24UID` in, but I only see references to `secure.adnxs.com`, and none of them seem to hit `/prebid/setuid`.

ecammit: I just loaded this:

> https://ad-cdn.technoratimedia.com/html/usersync.html?cb=https%3A%2F%2Fib.adnxs.com%2Fprebid%2Fsetuid%3Fbidder%3Dtriplelift%26gdpr%3D%26gdpr_consent%3D%26uid%3D%24UID

And received a call to:

> https://ib.adnxs.com/prebid/setuid?bidder=triplelift&gdpr=&gdpr_consent=&uid=$UID

Screen Shot 2022-02-04 at 9 55 55 AM

I'll note two things:

1. Your test URL uses "$UID" as the macro, but our userMacro is set to "[USER_ID]" in the YAML file. There is a "tads_uid" cookie set for the "technoratimedia.com" domain, which contains the user ID that will be replaced where [USER_ID] is located. I hope this assists with testing that part.
2. Our advertising system is only available to users that are geolocated by IP inside the United States or Canada. Anyone outside those two countries is provided with a tads_uid cookie value of "GDPR" instead of a user ID. That cookie persists even if your IP moves to be located within the US/Canada. If you have a "GDPR" cookie, the IFRAME will not call the provided "cb=" callback. This same functionality can also be verified through the old usersync endpoint of https://sync.technoratimedia.com/services?srv=cs&pid=70&cb={{.RedirectURL}}. It will not redirect if you have a GDPR value in your tads_uid cookie. Instead you'll receive a 204 No Content response.

hhhjort: I was using Postman to do my test, so there should be no cookies, and I was conducting my test from within the united states, so GDPR concerns should not be in play here. I missed the userMacro was specified while reading your PR here. Trying https://ad-cdn.technoratimedia.com/html/usersync.html?cb=https%3A%2F%2Fib.adnxs.com%2Fprebid%2Fsetuid%3Fbidder%3Dtriplelift%26gdpr%3D%26gdpr_consent%3D%26uid%3D%5BUSER_ID%5D, I still fail to see ib.adnxs.com, just secure.adnxs.com From your response, I thought perhaps the ib domain was missing as the redirect URL did not have a valid user macro, but this last attempt should be correct now? If not, can you give me a working redirect so I can see what I did wrong and we don't have to go back and forth several times to narrow down where the fault is.

hhhjort (approved):
LGTM

static/bidder-info/synacormedia.yaml
hhhjort: Odd, looks like I don't understand Postman as well as I thought.

AlexBVolcy (approved):

Issue #2142, opened 2022-01-25
PBS-Go: Dependency Updates


SyntaxNode:
Several dependencies used by PBS-Go (see `go.mod`) are not up to date. We will be updating all dependencies over the next several releases. There are no known issues with any of our dependencies, but we want to keep current.

We are testing each updated dependency before release and are reviewing their change logs to ensure there are no negative consequences expected. Out of an abundance of caution, we will be isolating all dependency updates in their own releases (even numbered releases) and updating only a handful at a time.

If you notice an issue, the expectation is you can safely roll back to the previous version and not lose features. Please report any issues here and we will respond quickly.

ie: 0.193.0 will be a feature release and 0.194.0 will contain only dependency updates.

We'll continue with this approach and keep this issue open until all updates have been made.

SyntaxNode:
We're done here, folks. All dependencies have been updated to the most recent version except for `gopkg.in/evanphx/json-patch` which is kept on version 4 (still supported) due to a performance bug introduced in 5.1.0. See https://github.com/evanphx/json-patch/issues/159 for further information.

PR #2151, opened 2022-02-04
Update Some Dependencies - Part 3

SyntaxNode:
**github.com/yudai/gojsondiff**
There doesn't appear to be any changes. We're just switching to Go Modules versioning. This dependency is only used for tests and the tests still pass. This library also looks abandoned and we might want to consider replacing it in the future.

**golang.org/x/net**
This package doesn't have formal releases. We were 21 commits behind. From the features we use, there is an updated public domain suffix list and no change to the context object.

**golang.org/x/text**
We were 1 release behind. Source code analysis of changes reveals a slight behavior difference where panics with text parsing are now caught and transferred to standard Go errors. Looks safe to me.

**github.com/spf13/viper** (Highest Risk In This PR)
We were 3 releases behind. Release notes indicate tfvars config file format was added and maintenance updates.

 **github.com/OneOfOne/xxhash**
We were 3 releases behind. Release notes indicate a few new features which we are not using.

See https://github.com/prebid/prebid-server/issues/2142 for more information.


VeronikaSolovei9 (approved):
LGTM

PR #2153, opened 2022-02-07
Address Dockerfile Security Recommendations

mansinahar:


Dockerfile
```diff
@@ -25,12 +25,14 @@ RUN go build -mod=vendor -ldflags "-X github.com/prebid/prebid-server/version.Ve
...
-COPY static static/
-COPY stored_requests/data stored_requests/data
+COPY --chmod=a+xr --from=build /app/prebid-server .
+COPY --chmod=a+r static static/
+COPY --chmod=a+r stored_requests/data stored_requests/data
 RUN apt-get update && \
     apt-get install -y ca-certificates mtr && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN adduser prebid_user
+USER prebid_user
```
SyntaxNode, line 35: Does `prebid_user` need any explicit permissions?

SyntaxNode (commented):
Thank you for hardening the Docker images, @mansinahar. Did you test this by creating and running the image locally?

Dockerfile
```diff
@@ -25,12 +25,14 @@ RUN go build -mod=vendor -ldflags "-X github.com/prebid/prebid-server/version.Ve
...
-COPY --from=build /app/prebid-server/prebid-server .
-COPY static static/
-COPY stored_requests/data stored_requests/data
+COPY --chmod=a+xr --from=build /app/prebid-server .
+COPY --chmod=a+r static static/
+COPY --chmod=a+r stored_requests/data stored_requests/data
 RUN apt-get update && \
     apt-get install -y ca-certificates mtr && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN adduser prebid_user
```
mansinahar, line 34: Executing prebid-server with a non-root user

Dockerfile
```diff
@@ -25,12 +25,14 @@ RUN go build -mod=vendor -ldflags "-X github.com/prebid/prebid-server/version.Ve
 FROM ubuntu:18.04 AS release
 LABEL maintainer="hans.hjort@xandr.com" 
 WORKDIR /usr/local/bin/
-COPY --from=build /app/prebid-server/prebid-server .
-COPY static static/
-COPY stored_requests/data stored_requests/data
+COPY --chmod=a+xr --from=build /app/prebid-server .
```
mansinahar, line 28: Making Prebid Server binary and other static resources to be non-writable

SyntaxNode:
Looks good. Please make the same changes to `.devcontainer/Dockerfile`.

mansinahar:
> Looks good. Please make the same changes to `.devcontainer/Dockerfile`.

I tried this but I am getting an issue when executing the apt-get commands with a non-root user as it requires permission to sybdirectories under /var/lib. This is the exact error:
```
> [4/4] RUN apt-get update && export DEBIAN_FRONTEND=noninteractive     && apt-get -y install --no-install-re
[2022-02-09T15:44:30.767Z] commends vim:
#7 0.253 Reading package lists...
#7 0.265 E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)
```
This also got me thinking that since this is a dev container, meant only for development purposes, executing with a root user should be okay, no? I am not sure what other issues will be uncovered by moving it to be executed by a non-root user. I think it's good to have the flexibility for development purposes. Thoughts @SyntaxNode ?

SyntaxNode:
> Thoughts @SyntaxNode ?

Ah, ok. Makes sense.

SyntaxNode (approved):

bsardo (approved):
LGTM

Dockerfile
mansinahar: Looks like this was released as part of 20.10 but was somehow missed from release notes (See: https://github.com/moby/moby/issues/34819#issuecomment-749570216) and requires `DOCKER_BUILDKIT=1` for it to work correctly (See: https://github.com/moby/moby/issues/34819#issuecomment-765637751).

We will change this to use the `RUN` command for now as this is a relatively new feature and might break for a lot of other people. Thanks for calling it out!

Issue #2152, opened 2022-02-04
Remove SomoAudience bid adapter


bretg:
SomoAudience no longer supports their server-side adapter, so would like to drop it. Email below.

We discussed in PBS committee and agreed that the best way to remove an adapter is:

- make sure there's a way to put the biddercode on a "soft-error" list so publishers don't start getting 400 errors from Prebid Server if they're utilizing the code in their AdUnits
- then remove the adapter code from the repo

```
---------- Forwarded message ---------
From: Todd H
Date: Wed, Jan 26, 2022 at 11:34 AM
Subject: Re: SomoAudience Prebid Server adapter floors issue
To: Bret G
Cc: Todd H


Hi Bret.,

We will not be maintaining the Somo Prebid adapter.  Let me know what steps I need to take to remove the Somo from prebid.  

TY,

Todd
```

SyntaxNode:
@bretg Will Somo open a PR themselves to remove their adapter? That would be preferred. The "soft-error" list already exists in PBS-Go, Please see https://github.com/prebid/prebid-server/blob/145499d2a1cd716cbfe1c8c1d893d9c1e03716b9/exchange/adapter_util.go#L79. A removed adapter is considered a disabled adapter with a custom error message indicating its removal from PBS. At some point in the future when enough time has passed we may decide to clean up these entries.

bretg:
I doubt we'll get a PR from SomoAudience. I updated the docs side to get them off that list.

bretg:
Done with PBS-Java 1.84

PR #2154, opened 2022-02-07
SomoAudience: Remove Adapter

SyntaxNode:
Per Somo's request: https://github.com/prebid/prebid-server/issues/2152

We asked Somo to author this PR themselves, but we haven't heard back. Going to assume they will not remove it themselves.


exchange/adapter_builders.go
```diff
@@ -2,7 +2,7 @@ package exchange
 
 import (
 	"github.com/prebid/prebid-server/adapters"
-	"github.com/prebid/prebid-server/adapters/33across"
+	ttx "github.com/prebid/prebid-server/adapters/33across"
```
AlexBVolcy, line 5: Interested in what this ttx is supposed to represent, and why removing the SomoAudience adapter includes this update related to the 33across adapter?

SyntaxNode: This is automatically inserted by Go's formatter since the package name doesn't match the directory name. Go has a convention of making the package name match the directory name, and when it doesn't this sort of stuff happens.

guscarreon (approved):
LGTM

AlexBVolcy (approved):
Ran an end to end test and all looks good. LGTM

PR #2160, opened 2022-02-16
Stored bid response

VeronikaSolovei9:
Stored bid responses implementation. 
-Depends on stored auction response branch, PR https://github.com/prebid/prebid-server/pull/2132
-No unit tests

Code flow: 
auction.go -> `func parseRequest`: added functionality to extract all stored bid responses and loads them from data store. 
returns stored bid responses: imp id to bidder to stored response body
`impBidderToStoredBidResponse := make(map[string]map[string]json.RawMessage)`
if request is valid, data successfully loaded - sets `StoredBidResponses:   storedBidResponses` to AuctionRequest and passes it to `HoldAuction`

utils.go -> `func cleanOpenRTBRequests`: builds list of Bidder Requests. In order to support stored bid responses, where in the same request impressions may have real bidder data and others may have stored bid responses I added `BidderStoredResponses map[string]json.RawMessage` to `BidderRequest`. One BidderRequest may have both real request and/or `BidderStoredResponses` initialized. 

exchange/bidder.go -> `func requestBid`: executes real and fake requests to adapters
case `if len(bidderRequest.BidRequest.Imp) > 0 {` handles real request fetching, executes `MakeRequests`, sends real http requests
case `if len(bidderRequest.BidderStoredResponses) > 0 {` handles stored bid responses, doesn't execute `MakeRequests`
both of these cases sets response data to channel `responseChannel`.  Stored bid response in this case looks like real bidder response, so it goes through usual processing including exercising bidder related code -  `func MakeBids`


VeronikaSolovei9:
Merged dependent branch for single stored response and rebased on top of latest upstream

exchange/exchange.go
```diff
@@ -218,7 +221,7 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 	var liveAdapters []openrtb_ext.BidderName
 
 	if len(r.StoredAuctionResponses) > 0 {
-		adapterBids, liveAdapters, err = buildStoredAuctionResponse(r.StoredAuctionResponses)
+		adapterBids, liveAdapters, err = buildStoreAuctionResponse(r.StoredAuctionResponses)
```
AlexBVolcy, line 224: This might be a spelling error, do we want this to say `buildStoredAuctionResponse` instead of `buildStoreAuctionResponse`

exchange/exchange.go
```diff
@@ -1094,11 +1097,11 @@ func listBiddersWithRequests(bidderRequests []BidderRequest) []openrtb_ext.Bidde
 	return liveAdapters
 }
 
-func buildStoredAuctionResponse(storedAuctionResponses map[string]json.RawMessage) (map[openrtb_ext.BidderName]*pbsOrtbSeatBid, []openrtb_ext.BidderName, error) {
+func buildStoreAuctionResponse(storedActionResponses map[string]json.RawMessage) (map[openrtb_ext.BidderName]*pbsOrtbSeatBid, []openrtb_ext.BidderName, error) {
 
 	adapterBids := make(map[openrtb_ext.BidderName]*pbsOrtbSeatBid, 0)
 	liveAdapters := make([]openrtb_ext.BidderName, 0)
-	for impId, storedResp := range storedAuctionResponses {
+	for impId, storedResp := range storedActionResponses {
```
AlexBVolcy, line 1104: Also wondering if this is a spelling error where you have the word `Action` instead of `Auction`

exchange/exchange_test.go
```diff
@@ -3885,7 +3885,7 @@ func TestBuildStoredAuctionResponses(t *testing.T) {
 	}
 	for _, test := range testCases {
 
-		bids, adapters, err := buildStoredAuctionResponse(test.in.StoredAuctionResponses)
+		bids, adapters, err := buildStoreAuctionResponse(test.in.StoredAuctionResponses)
```
AlexBVolcy, line 3888: Same question about potential spelling error `Store` vs. `Stored`

exchange/utils.go
```diff
@@ -741,3 +765,41 @@ func applyFPD(fpd *firstpartydata.ResolvedFirstPartyData, bidReq *openrtb2.BidRe
 		bidReq.User = fpd.User
 	}
 }
+
+func buildBidderRequestsWithBidResponses(req AuctionRequest, aliases map[string]string) map[openrtb_ext.BidderName]BidderRequest {
+	var bidderToBidderResponse map[openrtb_ext.BidderName]BidderRequest
+	if len(req.StoredBidResponses) > 0 {
+		// delete imps with stored bid resp
+		imps := req.BidRequest.Imp
+		req.BidRequest.Imp = nil //or new slice?
```
AlexBVolcy, line 774: Is this comment `//or new slice?` still needed?

exchange/utils.go
```diff
@@ -741,3 +765,41 @@ func applyFPD(fpd *firstpartydata.ResolvedFirstPartyData, bidReq *openrtb2.BidRe
...
+				req.BidRequest.Imp = append(req.BidRequest.Imp, imp)
+			}
+		}
+		// bidder -> imp id -> stored bid resp
+		bidderToBidderResponse = make(map[openrtb_ext.BidderName]BidderRequest)
+		for impID, storedData := range req.StoredBidResponses {
+			for bidderName, storedResp := range storedData {
+				if _, ok := bidderToBidderResponse[openrtb_ext.BidderName(bidderName)]; !ok {
+					//new bidder with stored bid responses
+					//!!! check if bidder is valid/exists
```
AlexBVolcy, line 787: Is this comment still needed, or if it is, could those `!!!` be removed?

exchange/exchange.go
VeronikaSolovei9: Renamed. It was fixed in Stored Single Response PR, but I rebased it incorrect. Good catch!

exchange/exchange.go

exchange/exchange_test.go

exchange/utils.go
VeronikaSolovei9: Removed, no need of this comment

exchange/utils.go
VeronikaSolovei9: Removed and added this validation to 'processStoredAuctionResponses'

VeronikaSolovei9:
Good comments, addressed all of them, added description of changes to PR

endpoints/openrtb2/auction.go
```diff
@@ -1582,44 +1583,76 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
...
+			bidderStoredRespId := make(map[string]string)
+			for _, bidderResp := range impData.ImpExtPrebid.StoredBidResponse {
+				if len(bidderResp.ID) == 0 || len(bidderResp.Bidder) == 0 {
+					return nil, nil, []error{fmt.Errorf("request.imp[%d] has ext.prebid.storedbidresponse specified, but \"id\" or/and \"bidder\" fields are missing ", index)}
+				}
+				//check if bidder is valid/exists
+				if _, isValid := deps.bidderMap[bidderResp.Bidder]; !isValid {
+					return nil, nil, []error{fmt.Errorf("request.imp[impId: %s].ext contains unknown bidder: %s. Did you forget an alias in request.ext.prebid.aliases?", impId, bidderResp.Bidder)}
+				}
+				//assuming bidder is unique per one bid stored response
```
AlexBVolcy, line 1629: Is this comment necessary? Just checking :)

endpoints/openrtb2/auction.go
```diff
@@ -1582,44 +1583,76 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
...
+				//storedAuctionResponseIds are not unique, but fetch will return single data for repeated ids
+				storedAuctionResponseIds = append(storedAuctionResponseIds, bidderResp.ID)
+			}
+		}
 	}
 	if len(storedAuctionResponseIds) > 0 {
 		storedAuctionResponses, errs := deps.storedRespFetcher.FetchResponses(ctx, storedAuctionResponseIds)
 		if len(errs) > 0 {
-			return nil, errs
+			return nil, nil, errs
```
AlexBVolcy, line 1640: Could you add test coverage for this error case

AlexBVolcy (commented):
So I tested code coverage for this PR, and I left comments for which new code additions aren't yet covered by tests. Could you update this?

endpoints/openrtb2/auction.go
VeronikaSolovei9: I'm 90% sure this should be unique, but I need to check with Bret. I'll remove the comment when I get the answer.

VeronikaSolovei9:
When unit tests are added I'll make sure it's all covered.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Per discussion with Bret we don't need to support this. I'll remove a comment on my next commit.

exchange/bidder.go
```diff
@@ -129,54 +129,87 @@ type bidderAdapterConfig struct {
...
+				request: &reqDataForStoredResp,
+				response: &adapters.ResponseData{
+					StatusCode: 200,
+					Body:       bidresp,
+				},
+				err: nil,
+			}
+			go func() { responseChannel <- respData }()
+		}
+
```
bsardo, line 200: This seems like another possible function that can be extracted into a stored responses package to reduce some of the complexity here.

exchange/utils.go
```diff
@@ -46,14 +46,18 @@ func cleanOpenRTBRequests(ctx context.Context,
 	privacyConfig config.Privacy,
 	account *config.Account) (allowedBidderRequests []BidderRequest, privacyLabels metrics.PrivacyLabels, errs []error) {
 
-	impsByBidder, err := splitImps(req.BidRequest.Imp)
-	if err != nil {
-		errs = []error{err}
+	aliases, errs := parseAliases(req.BidRequest)
+	if len(errs) > 0 {
 		return
 	}
```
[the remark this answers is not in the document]
VeronikaSolovei9: I just moved it to the first statement of the  `cleanOpenRTBRequests` function, because aliases are needed in `buildBidderRequestsWithBidResponses`.  `buildBidderRequestsWithBidResponses` should be executed before everything else, because it modifies request - removes all imps with stored bid responses from imp array. 

```
        aliases, errs := parseAliases(req.BidRequest)
	if len(errs) > 0 {
		return
	}

	allowedBidderRequests = make([]BidderRequest, 0, 0)

	bidderToBidderResponse := buildBidderRequestsWithBidResponses(req, aliases)

       impsByBidder, err := splitImps(req.BidRequest.Imp)
...
```

endpoints/openrtb2/auction.go
```diff
@@ -1582,44 +1583,76 @@ func getJsonSyntaxError(testJSON []byte) (bool, string) {
 // Note that processStoredAuctionResponses must be called after processStoredRequests
 // because stored imps and stored requests can contain stored auction responses
 // so the stored requests/imps have to be merged into the incoming request prior to processing stored auction responses.
-func (deps *endpointDeps) processStoredAuctionResponses(ctx context.Context, requestJson []byte) (map[string]json.RawMessage, []error) {
+func (deps *endpointDeps) processStoredAuctionResponses(ctx context.Context, requestJson []byte) (map[string]json.RawMessage, map[string]map[string]json.RawMessage, []error) {
```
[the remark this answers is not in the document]
VeronikaSolovei9: This function extracts storedResponses and storedBidResponses from raw request. We use the same fetcher for both of them. Because of parsing a raw request we extract all ids and return them `return impIdToStoredResp, impBidderToStoredBidResponse, nil`. What do you think would be a good split?

exchange/utils.go
```diff
@@ -46,14 +46,18 @@ func cleanOpenRTBRequests(ctx context.Context,
...
 
-	aliases, errs := parseAliases(req.BidRequest)
-	if len(errs) > 0 {
+	allowedBidderRequests = make([]BidderRequest, 0, 0)
+
+	bidderToBidderResponse := buildBidderRequestsWithBidResponses(req, aliases)
+
+	impsByBidder, err := splitImps(req.BidRequest.Imp)
```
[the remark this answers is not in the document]
VeronikaSolovei9: `buildBidderRequestsWithBidResponses` function removes "fake" imps from request. Split imps works with "real" impressions only. Please my previous comment - it has more context on this.

Issue #1641, opened 2021-01-05
Account-level defaults for /cookie_sync


bretg:
As a followup to https://github.com/prebid/prebid-server/issues/924, there's a request to add server-side per-account control for the coopSync and limit flags -- being able to specify the default values.

The proposal is to add (yet another) column to the DB. `endpoint_config` column with JSON type.

We'd support 3 values initially:
```
{
 "cookie_sync": {
    "defaultLimit": 4,
    "maxLimit": 8,
    "defaultCoopSync": false
   }
}
```
Meanings:
- cookie_sync.defaultLimit: if the "limit" isn't specified in the /cookie_sync request, this is what to use
- cookie_sync.maxLimit: if the "limit" is specified, it can't be greater than this value
- cookie_sync.defaultCoopSync: if the "coopSync" value isn't specified in the /cookie_sync request, use this

SyntaxNode:
Why do publishers need PBS to store default values instead of passing it with the cookie sync request as we have today?

bretg:
This could be used in the future for preference of iframe-vs-redirect syncs when that feature is built.

SyntaxNode:
Implemented in PBS-Go 0.204.0.

PR #2161, opened 2022-02-17
Account-Level Defaults for Cookie Sync Endpoint

AlexBVolcy:
Addressing this issue: https://github.com/prebid/prebid-server/issues/1641

This PR adds Account-Level Defaults to the cookie-sync endpoint

The defaults are `DefaultLimit`, `MaxLimit`, and `DefaultCoopSync`, and were added to the `Account` struct in `config/accounts.go`

The defaults are utilized in this manner:
- If a cookie sync request doesn't contain a `limit` field, we then set the field to equal the `DefaultLimit`
- If a cookie sync request contains a `limit` field, but it's larger than the `MaxLimit`, then we set the field equal to `MaxLimit`
- If a cookie sync request doesn't contain a `coopSync` field, we set the field equal to `DefaultCoopSync`

The `TestCookieSyncParseRequest` has been updated with more cases, along with tweaks to existing cases, to support the changes.


config/accounts.go
```diff
@@ -21,6 +21,9 @@ type Account struct {
 	GDPR               AccountGDPR `mapstructure:"gdpr" json:"gdpr"`
 	DebugAllow         bool        `mapstructure:"debug_allow" json:"debug_allow"`
 	DefaultIntegration string      `mapstructure:"default_integration" json:"default_integration"`
+	DefaultLimit       int         `mapstructure:"default_limit" json:"default_limit"`
+	MaxLimit           int         `mapstructure:"max_limit" json:"max_limit"`
+	DefaultCoopSync    bool        `mapstructure:"default_coop_sync" json:"default_coop_sync"`
```
SyntaxNode, line 26: I recommend creating an object structure to provide context and further extension points in the future, such as embedding these three within a "UserSync" object.

endpoints/cookie_sync.go
```diff
@@ -63,6 +63,7 @@ func NewCookieSyncEndpoint(
 type cookieSyncEndpoint struct {
 	chooser          usersync.Chooser
 	config           config.UserSync
+	accountConfig    config.Account
```
SyntaxNode, line 66: This data structure is for the life of the application, but  account config should be for the life of the request. You'll want to add the account as a string in the `cookieSyncRequest` struct and use that with the `stored_requests.AccountFetcher` to lookup the account config. So it's really the AccountFetcher you want to include at this level.

There is also no automatic dependency injection in this project. You'll need to wire that up manually in router.go#New. Look for the line:

```
r.POST("/cookie_sync", endpoints.NewCookieSyncEndpoint(syncersByBidder, cfg, gdprPerms, r.MetricsEngine, pbsAnalytics, activeBidders).Handle)
```

where you'll have the `accounts` variable for the AccountFetcher in scope.

config/accounts.go
mansinahar: I agree with @SyntaxNode. I'd just say that we should probably use `cookie_sync` as the wrapper object for these fields so as to stay consistent with naming in Java.
Ref: https://github.com/prebid/prebid-server-java/blob/master/src/main/java/org/prebid/server/settings/model/Account.java#L23

endpoints/cookie_sync_test.go
```diff
@@ -647,18 +657,109 @@ func TestCookieSyncParseRequest(t *testing.T) {
...
+		{
+			description: "Request where account level defaults are applied: - MaxLimit and Default Coop",
+			givenBody: strings.NewReader(`{` +
+				`"bidders":["a", "b"],` +
+				`"gdpr":1,` +
+				`"gdpr_consent":"anyGDPRConsent",` +
+				`"us_privacy":"1NYN",` +
+				`"limit":42,` +
+				`"account":"TestAccount"` +
+				`}`),
```
SyntaxNode, line 669: I recommend removing gdpr, gdpr_consent, and us_privacy from the request. It's ideal for the test setup to be as simple as possible and these values don't seem to have an impact on the area of code you are testing. Look to a test case like "Cooperative True - Default False" as an example.

endpoints/cookie_sync.go
```diff
@@ -125,6 +132,21 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
...
+		}
+		if request.Limit == 0 {
+			request.Limit = accountInfo.Defaults.DefaultLimit
+		} else if request.Limit > accountInfo.Defaults.MaxLimit {
+			request.Limit = accountInfo.Defaults.MaxLimit
+		}
+		if request.CooperativeSync == nil {
+			request.CooperativeSync = &accountInfo.Defaults.DefaultCoopSync
+		}
+	}
```
SyntaxNode, line 148: I like the tests you have in `TestNewCookieSyncEndpoint` for end to end functionality. Please also add tests to `TestCookieSyncParseRequest` which can include additional permutations, such as handling a potential error from the account fetcher on line 136.

VeronikaSolovei9 (commented):
Overall LGTM, please address SyntaxNode's comments

endpoints/cookie_sync.go
```diff
@@ -47,26 +50,30 @@ func NewCookieSyncEndpoint(
 
 	return &cookieSyncEndpoint{
 		chooser:          usersync.NewChooser(syncersByBidder),
-		config:           config.UserSync,
+		config:           config,
+		configUserSync:   config.UserSync,
```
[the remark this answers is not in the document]
AlexBVolcy: Do you think we should alter the cookie sync endpoint struct to just include a config field now as well?

endpoints/cookie_sync.go
```diff
@@ -125,6 +132,21 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
 		}
 	}
 
+	if request.Account != "" {
+		accountInfo, errs := accountService.GetAccount(context.Background(), c.config, c.accountsFetcher, request.Account)
+		if len(errs) > 0 {
+			return usersync.Request{}, privacy.Policies{}, errs[0]
```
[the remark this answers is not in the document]
AlexBVolcy: Yeah I was also confused about this, because account service returns an array of errors, but the parse request function only returns a single error. Veronika, are you suggesting we refactor the GetAccount function?

endpoints/cookie_sync.go
```diff
@@ -257,6 +258,15 @@ func (c *cookieSyncEndpoint) handleError(w http.ResponseWriter, err error, httpS
 	})
 }
 
+func combineErrors(errs []error) error {
+	var errorStringList []string
```
SyntaxNode, line 262: The `List` at the end of the variable name looks like Hungarian notation. Please consider removing it.

endpoints/cookie_sync.go
```diff
@@ -125,6 +131,21 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
...
+			return usersync.Request{}, privacy.Policies{}, combineErrors(errs)
+		}
+		if request.Limit == 0 {
+			request.Limit = accountInfo.CookieSync.DefaultLimit
+		} else if request.Limit > accountInfo.CookieSync.MaxLimit {
+			request.Limit = accountInfo.CookieSync.MaxLimit
+		}
+		if request.CooperativeSync == nil {
+			request.CooperativeSync = &accountInfo.CookieSync.DefaultCoopSync
+		}
```
SyntaxNode, line 146: There is a logic gap here. As a publisher, I expect my max limit to always be honored. However, if you have a 0 limit and a 0 default limit, the limit will remain at 0 which is interpreted as unlimited. I had luck with the following:

```
	if request.Account != "" {
		accountInfo, errs := accountService.GetAccount(context.Background(), c.config, c.accountsFetcher, request.Account)
		if len(errs) > 0 {
			return usersync.Request{}, privacy.Policies{}, combineErrors(errs)
		}

		if request.Limit == 0 {
			request.Limit = accountInfo.CookieSync.DefaultLimit
		}

		if request.Limit == 0 || request.Limit > accountInfo.CookieSync.MaxLimit {
			request.Limit = accountInfo.CookieSync.MaxLimit
		}

		if request.CooperativeSync == nil {
			request.CooperativeSync = &accountInfo.CookieSync.DefaultCoopSync
		}
	}
```

VeronikaSolovei9 (approved):
LGTM

guscarreon (commented):
Looks good to me. Left a couple of nitpicks

VeronikaSolovei9 (dismissed):
Looks good

endpoints/cookie_sync.go
```diff
@@ -125,6 +132,22 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
...
+		accountInfo, errs := accountService.GetAccount(context.Background(), c.config, c.accountsFetcher, request.Account)
+		if len(errs) > 0 {
+			return usersync.Request{}, privacy.Policies{}, combineErrors(errs)
+		}
+		if request.Limit == 0 {
+			request.Limit = accountInfo.CookieSync.DefaultLimit
+		}
+		if request.Limit == 0 || request.Limit > accountInfo.CookieSync.MaxLimit {
+			request.Limit = accountInfo.CookieSync.MaxLimit
+		}
```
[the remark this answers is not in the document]
guscarreon: Also, can we move logic between lines 135 and 149 into its own function? It seems that the `parseRequest(r *http.Request)` function is already big enough. Let me know your thoughts.

guscarreon (commented):
Looks great. Last couple of comments

AlexBVolcy:
I put the main part of the implementation into its own function, `setLimit`, which gets called during `parseRequest`, to respond to Gus' comment! I've also added a new `TestSetLimit` function that covers all code from `setLimit`. Some of the limit related edge case tests have been moved from `TestCookieSyncParseRequest`to `TestSetLimit`, since it makes more sense for them to live there.

guscarreon (approved):
LGTM. Thanks for addressing the feedback.

VeronikaSolovei9 (approved):

PR #2162, opened 2022-02-18
Update Some Dependencies - Part 4

SyntaxNode:
This release focuses on updating GoMetrics and InfuxDB hooks, including some **major changes**. If you use InfluxDB, please weigh in.

**Influx DB Changes:**
- New required config value `metrics.influxdb.measurement`. This is now required by the updated library we use for InfluxDB and is described as "Add a measurement to the output to better align with the InfluxDB idiom." 
- New optional config value `metrics.align_send_timestamp` which will "align all the points to the same time when metrics are sent". It is `false` be default to match our existing behavior.
- Aligns the configuration of histograms with the behavior of telegraf, where there is a bucket tag rather than an individual metric.
- Adds a boolean to align time at the beginning of the interval, so a 10s interval would evenly align the time stamp at 00,10,20, etc. seconds

Please see https://github.com/vrischmann/go-metrics-influxdb/pull/16 for further details.

**github.com/influxdata/influxdb** - Removed
We were only using this package directly for a few tests, likely inadvertently. I changed these tests to use testify.

**github.com/rcrowley/go-metrics**
This package doesn't have formal releases. We were 28 commits behind master. From the commits, most of the changes seem to be bug fixes and speed improvements.

**github.com/vrischmann/go-metrics-influxdb**
This package only recently added formal releases. We were 22 commits behind master. The InfluxDB changes listed at the top of this PR are from upgrading this package.

**gopkg.in/yaml**
We were 1 major release behind. This package is only used for loading bidder yaml files once at startup, so the risk seems minimal. The only change I've noticed so far is duplicate entries now cause an error instead of being merged behind the scenes, and it caught I mistake I previously made which is corrected in this PR.

**github.com/mitchellh/copystructure**
We were 1 release behind. This package is only used for tests and the tests still pass.

**github.com/evanphx/json-patch**
I downgraded from v5 to v4 due to performance issues. The latest version of v5 uses 3x more memory and is 3x slower. This is impactful since we use this library for merging in stored requests, including the optional host-level stored request. Version 4 doesn't have this issue and is still supported.

See https://github.com/prebid/prebid-server/issues/2142 for more information.


VeronikaSolovei9 (commented):
Local testing looks good. 

I found [this article](https://docs.influxdata.com/influxdb/v1.8/concepts/key_concepts/) says ```The measurement acts as a container for tags, fields, and the time column, and the measurement name is the description of the data that are stored in the associated fields. Measurement names are strings, and, for any SQL users out there, a measurement is conceptually similar to a table```. 
Maybe we can leave it empty by default if other configs (host, username etc) are not specified?

config/config_test.go
```diff
@@ -557,8 +559,10 @@ func TestFullConfig(t *testing.T) {
 	cmpStrings(t, "recaptcha_secret", cfg.RecaptchaSecret, "asdfasdfasdfasdf")
 	cmpStrings(t, "metrics.influxdb.host", cfg.Metrics.Influxdb.Host, "upstream:8232")
 	cmpStrings(t, "metrics.influxdb.database", cfg.Metrics.Influxdb.Database, "metricsdb")
+	cmpStrings(t, "metrics.influxdb.measurement", cfg.Metrics.Influxdb.Measurement, "anyMeasurement")
 	cmpStrings(t, "metrics.influxdb.username", cfg.Metrics.Influxdb.Username, "admin")
 	cmpStrings(t, "metrics.influxdb.password", cfg.Metrics.Influxdb.Password, "admin1324")
+	cmpBools(t, "metrics.align_send_timestamp", cfg.Metrics.Influxdb.AlignSendTimestamp, true)
```
VeronikaSolovei9, line 565: Should we have these properties in config/config.go (L695-699)?

SyntaxNode:
> I found [this article](https://docs.influxdata.com/influxdb/v1.8/concepts/key_concepts/) says `The measurement acts as a container for tags, fields, and the time column, and the measurement name is the description of the data that are stored in the associated fields. Measurement names are strings, and, for any SQL users out there, a measurement is conceptually similar to a table`. Maybe we can leave it empty by default if other configs (host, username etc) are not specified?

Fantastic. Thank you for the guidance. I'm fine leaving it empty by default, which is the behavior currently in this PR.

config/config_test.go
SyntaxNode: Yes. Yes they should. Good catch. This doesn't change the default values, but will enable binding to the environment variables.

VeronikaSolovei9:
Thank you for the updates. LGTM.

VeronikaSolovei9 (approved):

bsardo (approved):

PR #2164, opened 2022-02-22
Adot: add new bid param publisherPath

ramyferjaniadot:
Hello,

This PR is about adding a new bid param `placementPath` to our adapter.
We also updated our doc in this [PR](https://github.com/prebid/prebid.github.io/pull/3590) 

Best regards,


adapters/adot/adot.go
```diff
@@ -46,9 +47,17 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 	headers := http.Header{}
 	headers.Add("Content-Type", "application/json;charset=utf-8")
 
+	if adotExt := getImpAdotExt(&request.Imp[0]); adotExt != nil {
```
AlexBVolcy, line 50: For my understanding, could there be a situation where `request.Imp` has multiple entries, and you'd want the publisher path from an entry other than `request.Imp[0]`?

hhhjort (approved):
LGTM

adapters/adot/adot.go
ramyferjaniadot: The publisher path will be used with some of our partners for custom integrations, so the value should be the same across all `request.Imp` entries

AlexBVolcy (approved):

PR #2166, opened 2022-02-23
[Sharethrough] Bid Adapter Revamp connecting to new Endpoint

maphe:
We have a new endpoint to connect to, leveraging Open RTB format, so we're refreshing our old adapter and all its custom logic.

Changes are mostly cleaning up old complex logic, and keeping it simple from open RTB to open RTB. We also removed all the custom unit tests to leverage the json-test framework.


adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,103 @@
...
+	}
+	var sourceExt map[string]interface{}
+	if err := json.Unmarshal(modifiableSource.Ext, &sourceExt); err == nil {
+		sourceExt["str"] = strVersion
+		sourceExt["version"] = version.Ver
+	} else {
+		sourceExt = map[string]interface{}{"str": strVersion, "version": version.Ver}
 	}
-	var domain = Util{}.parseDomain(request.Site.Page)
+	modifiableSource.Ext, _ = json.Marshal(sourceExt)
```
SyntaxNode, line 48: Eeek! Please do not ignore errors from `json.Marshal`.

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,103 @@
...
+	if request.Source != nil {
+		modifiableSource = *request.Source
+	}
+	var sourceExt map[string]interface{}
+	if err := json.Unmarshal(modifiableSource.Ext, &sourceExt); err == nil {
+		sourceExt["str"] = strVersion
+		sourceExt["version"] = version.Ver
+	} else {
+		sourceExt = map[string]interface{}{"str": strVersion, "version": version.Ver}
 	}
```
SyntaxNode, line 47: It is inefficient to use a `map[string]interface{}` to marshal an ext, both in terms of cpu and memory usage. It's always better in Go to use an explicit `struct`. In this specific case, using an explicit struct is ~3x more efficient.

I'm cool if you want to use an anonymous struct to keep the code clean.

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,103 @@
...
+		var strImpParams openrtb_ext.ExtImpSharethrough
+		if err := json.Unmarshal(strImpExt.Bidder, &strImpParams); err != nil {
+			errors = append(errors, err)
+			continue
+		}
 
-	for i := 0; i < len(request.Imp); i++ {
-		req, err := a.AdServer.requestFromOpenRTB(request.Imp[i], request, domain)
+		// Convert Floor into USD
+		if imp.BidFloor > 0 && imp.BidFloorCur != "" && strings.ToUpper(imp.BidFloorCur) != "USD" {
```
SyntaxNode, line 66: It's more efficient to use the case insensitive `EqualFold` instead of converting the bid floor currency to an uppercase string.

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,103 @@
...
+			convertedValue, err := reqInfo.ConvertCurrency(imp.BidFloor, imp.BidFloorCur, "USD")
+			if err != nil {
+				return nil, []error{err}
```
SyntaxNode, line 69: Why is an error with currency conversion treated more severely than an error with json unmarshaling?

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,103 @@
...
+			}
+			imp.BidFloorCur = "USD"
+			imp.BidFloor = convertedValue
+		}
+
+		// Relocate Custom Params
+		imp.TagID = strImpParams.Pkey
+		requestCopy.BCat = strImpParams.BCat
+		requestCopy.BAdv = strImpParams.BAdv
```
SyntaxNode, line 78: Why do you replace the request bcat and badv with the params from this bidder? What if the params are empty but the publisher has information on the request? We've had complaints about similar behavior with some bidders for bid floors. In those cases, the adapter only replaced the value in the request if the adapter specific param is non-empty.

adapters/sharethrough/sharethrough.go
```diff
@@ -71,5 +112,32 @@ func (a SharethroughAdapter) MakeBids(internalRequest *openrtb2.BidRequest, exte
...
+	for _, seatBid := range bidResp.SeatBid {
+		for _, bid := range seatBid.Bid {
+			bidType := openrtb_ext.BidTypeBanner
+			if bidReq.Imp[0].Video != nil {
+				bidType = openrtb_ext.BidTypeVideo
+			}
+
+			bidderResponse.Bids = append(bidderResponse.Bids, &adapters.TypedBid{
+				BidType: bidType,
+				Bid:     &bid,
```
SyntaxNode, line 137: There's a subtle bug here with exporting a reference to a for loop iterator variable. You should use the index instead,

Please review our updated guidance from the adapter dev guide: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#adapter-code

SyntaxNode (commented):
> Changes are mostly cleaning up old complex logic, and keeping it simple from open RTB to open RTB. 

Sweet! I love code simplification.

> We also removed all the custom unit tests to leverage the json-test framework.

Thank you. We've since enhanced the json test framework to check for shared memory changes, so this is much appreciated. I'll review test coverage in my next review pass.

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,103 @@
...
 	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/version"
+	"net/http"
+	"strings"
 )
 
-const supplyId = "FGMrCMMc"
-const strVersion = 8
+var strVersion = "10.0"
```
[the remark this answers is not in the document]
maphe: I used a `var` to overwrite in tests, so it can be hardcoded there and we don't have to change all json files when the version is incremented, if I can do that with a const I'm cool with it.

I tried renaming to `version` but it conflicts with `github.com/prebid/prebid-server/version`, I'm going with `adapterVersion`

adapters/sharethrough/sharethrough.go
maphe: I assumed we'll never be sure of what's coming in the `ext` and we just wanna add a couple attributes to what's already there. I'm not sure whether you mean I should only change it on line 46 (which looks more doable), or line 42 as well (which is the tricky one).

maphe:
Thanks @SyntaxNode for the in depth review! I haven't had the time to dive deep into it today but I responded inline with a couple of questions. I'll address the rest tomorrow. Go is not my primary language, as you probably figured 😆 but I'll do my best.

adapters/sharethrough/sharethrough.go
maphe: I copy pasted from the example here: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#currency

You might wanna update the example in the doc, or future adapters might have the same inefficiency ;)

adapters/sharethrough/sharethrough.go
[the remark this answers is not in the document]
SyntaxNode: > I used a `var` to overwrite in tests, so it can be hardcoded there and we don't have to change all json files when the version is incremented, if I can do that with a const I'm cool with it.

Ah, ok. That's fine.

> I tried renaming to `version` but it conflicts with `github.com/prebid/prebid-server/version`, I'm going with `adapterVersion`

Sounds good. Thank you.

adapters/sharethrough/sharethrough.go
SyntaxNode: Gotcha. I forgot it was also used with the unmarshal call when I made the comment. You'll start to get into type issues if you make the change on just line 46, so.. carry on. :)

adapters/sharethrough/sharethrough.go
SyntaxNode: Fair point. I'll update the example tomorrow.

SyntaxNode:
> Go is not my primary language, as you probably figured 😆 but I'll do my best.

Not a problem. Your code looks good to me. Happy to provide guidance.

adapters/sharethrough/sharethrough.go
maphe: I'm assuming I don't even need `imp.BidFloorCur != ""` in the test either, the `EqualFold` should handle it juste fine, so I'm removing it, let me know if that's a mistake

adapters/sharethrough/sharethrough.go
maphe: tbh the primary reason is that I took this piece from the example in the doc, but thinking about it I think it's ok if it fails here, having an erroneous floor-currency pair seems bad enough to reject the request. I'm open to suggestions tho

maphe:
@SyntaxNode I think I addressed everything, feel free to take another look

hhhjort (dismissed):
LGTM

adapters/sharethrough/sharethrough.go
SyntaxNode: I'm good with that.

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,109 @@
...
+			continue
+		}
+		var strImpParams openrtb_ext.ExtImpSharethrough
+		if err := json.Unmarshal(strImpExt.Bidder, &strImpParams); err != nil {
+			errors = append(errors, err)
+			continue
+		}
+
+		// Convert Floor into USD
+		if imp.BidFloor > 0 && strings.EqualFold(imp.BidFloorCur, "USD") {
```
SyntaxNode, line 72: You need to invert the `EqualFold`, as you want to convert if the bid floor currency is **not** USD. You may also want to keep `imp.BidFloorCur != ""` so that an attempted conversion to USD isn't performed if no currency is specified. Your bidding server can handle this case as it sees fit.

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,109 @@
...
+	} else {
+		sourceExt = map[string]interface{}{"str": adapterVersion, "version": version.Ver}
+	}
 
-	if request.Site == nil {
-		return nil, []error{fmt.Errorf("request must include a site; in-app placements are not supported")}
+	var err error
+	if modifiableSource.Ext, err = json.Marshal(sourceExt); err != nil {
+		errors = append(errors, err)
 	}
```
SyntaxNode, line 54: Should this be a fatal error? If for some reason the str and version can't be sent to your bidding server, would it still be considered a valid bid?

adapters/sharethrough/sharethrough_test.go
```diff
@@ -1,278 +1,24 @@
...
-		},
-	}
-
-	mockAdServer := MockStrAdServer{
-		mockResponseToOpenRTB: func() (*adapters.BidderResponse, []error) {
-			return &stubBidderResponse, []error{}
-		},
-	}
+func TestJsonSamples(t *testing.T) {
+	version.Ver = "42.42"
```
SyntaxNode, line 12: This is a bit scary since you're changing the shared definition for all tests running after this. I recommend you either write your tests against the default empty string or capture the version's state in your adapter struct and modify it there. I don't care as much about the state of your adapterVersion as it's already isolated to just this test.

adapters/sharethrough/sharethrough.go
SyntaxNode, line 72: There is no test coverage within this condition. Is that possible to add? .. or is our json test framework not up for the task?

adapters/sharethrough/sharethrough.go
```diff
@@ -1,62 +1,109 @@
...
 
-	// We never add to the errs slice (early return), so we just create an empty one to return
-	return reqs, []error{}
+	return requests, errors
 }
 
-func (a SharethroughAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+func (a adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
 	if response.StatusCode == http.StatusNoContent {
```
SyntaxNode, line 107: Please add json tests for each `response.StatusCode` condition.

adapters/sharethrough/sharethrough.go
maphe: it's fine if we miss those, they're mostly here for logging/analytics purpose 👍

adapters/sharethrough/sharethrough.go
maphe: I haven't found a way to mock the bid floor conversion, also I was ok with not testing it since it's a copy paste from the Prebid doc. If there's a way and some examples of testing this via the json framework, I'm happy to add a test.

maphe:
@SyntaxNode thanks, good catch on the currency conversion condition. I fixed everything you called out except for adding a test for currency conversion as I'm not sure it's even doable.

hhhjort (approved):

adapters/sharethrough/sharethrough.go
SyntaxNode: I don't think we have an easy way to do that. I'm fine with skipping test coverage for now. We may add in tests for this adapter, and the others, when we add per-test definable currency conversion rates.

SyntaxNode (approved):
Looks good. Thank you for the quick follow-ups.

PR #2169, opened 2022-03-01
Adkernel: static endpoint host

ckbo3hrk:
Switched to static adkernel host instead of publisher-side configured domains to address https://github.com/prebid/prebid-server/issues/2127


hhhjort:
You should drop `host` as a required bidder parameter, and probably drop it completely, unless your bidder is making some use of the host parameter being passed, even though your adapter is ignoring it.

AlexBVolcy (commented):
This looks good, but I won't approve until other comments are addressed (i.e. removing host from bidder params)

hhhjort (approved):
LGTM

AlexBVolcy (approved):

PR #2185, opened 2022-03-17
AdkernelAdn: static endpoint host

ckbo3hrk:
Switched to static adkernel host instead of publisher-side configured domains to address https://github.com/prebid/prebid-server/issues/2127


hhhjort (approved):
LGTM

bsardo (approved):
LGTM

Issue #1727, opened 2021-02-26
AMP first party data v2


bretg:
The original implementation for AMP first party data was pretty complicated. Upon revisiting it, I'm thinking it was also incorrect in that it mixed GAM targeting with PBS FPD. So I'd like to significantly simplify. However, would like community members with AMP experience to weigh in.

## Assumptions
- The AMP "TGT" macro is resolved from json.targeting. Because these values are also utilized by GAM, there cannot be any additional structure imposed on this object.
- Dynamic values (e.g. user IDs or segments) are not supported in AMP. When they are, we will revise this spec.
- We could add additional fields to pass static data through the AMP call, but it's easier to just store them in the storedrequest database.
- Adding global first party data fields to the AMP storedrequests keyed to the stored request ID is feasible
- No one started making use of the original FPD implementation. No need for backwards compatibility.

## Requirements
1. The AMP openrtb in the stored request must be able to carry global site data.
2. The stored request must also be able to define bidder permissions either with ext.prebid.data.bidders or ext.prebid.bidderconfig.
    - It's ok if both kinds of permission are specified: i.e. every entry in ext.prebid.bidderconfig[].bidders would also need to be in ext.prebid.data.bidders or it will be ignored.
3. Some fields defined in the stored request will be overridden by values coming in on the query string:
    - curl overwrites site.page
    - account overwrites site.publisher.id if it's valid
4. Read the targeting attribute on the AMP query string. Add all key-value-pairs found there to imp[].ext.data.
5. PBS must adhere to privacy rules as for other interfaces:
    - Follow COPPA anonymization rules
    - Follow usp_privacy anonymization rules
    - Follow TCF1 anonymization rules
    - Follow TCF2 rules

## Examples

Here's an amp-ad tag that passes targeting data to GAM:
```
<amp-ad width="300" height="50"
    type="doubleclick"
    data-slot="/1111/amp_test"
    data-multi-size-validation="false"
    rtc-config='{"vendors": {"prebidrubicon": {"REQUEST_ID": "blah"}}}'
    json='{ "targeting": {"gam-key1": "val1","gam-key2": "val2"}}' >
  </amp-ad>
```
This ends up calling the PBS AMP endpoint with something like:

[...]

PR #2196, opened 2022-03-22
First party data fix for resolved request

VeronikaSolovei9:
Added original request copy before First party data extraction to save it for resolved request in debug mode.


endpoints/openrtb2/auction.go
```diff
@@ -214,8 +205,27 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 
 	secGPC := r.Header.Get("Sec-GPC")
 
+	//ExtractFPDForBidders removes FPD related structures from request.
+	//In order to preserve original request and return it properly in response.ext.debug.resolvedrequest
+	//original request should be copied before FPD execution
+	resolvedBidReq, err := exchange.DeepCopy(req.BidRequest)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good points. I'll fix point 1, didn't think about this optimization. Right, in majority of cases this functionality will not be needed.
I was thinking about point 2 as well. I'll see how much changes it will need

exchange/utils.go
```diff
@@ -4,10 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"math/rand"
-
 	"github.com/mxmCherry/openrtb/v15/openrtb2"
 	"github.com/prebid/go-gdpr/vendorconsent"
+	"math/rand"
```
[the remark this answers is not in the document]
VeronikaSolovei9: I think Intellij rearranged imports automatically. It was on the line 7 before. I moved all `github.com` import down and separated them from other imports

endpoints/openrtb2/auction.go
```diff
@@ -214,8 +205,31 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 
 	secGPC := r.Header.Get("Sec-GPC")
 
+	var resolvedBidReq json.RawMessage
+	requestExt, _ := req.GetRequestExt()
+	if req.Test == 1 || (requestExt != nil && requestExt.GetPrebid() != nil && requestExt.GetPrebid().Debug) {
```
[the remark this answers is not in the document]
VeronikaSolovei9: I tried two approaches that we discussed and didn't like any of them.
1) If I move FPD extraction down to `HoldAuction` - it will be executed for amp and video endpoints too, and we don't want this. 
2) If I move `getDebugInfo` up to `action` endpoint handler - I will need to duplicate it for amp and video as well. 

Line 210 is a duplicate of `parseRequestDebugValues` function. For resolved request we only need to check `req.Test` is 1 or `req.ext.Prebid.debug` is enabled. For action endpoint DebugLog is always nil, so this simple check does the thing.
This approach is some sort of cheating, I understand. I didn't find another good way to implement this. 
We discussed it with team and they are ok with this approach. 
Please let me know what you think.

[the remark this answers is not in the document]
SyntaxNode: > it will be executed for amp and video endpoints too, and we don't want this.

We actually do want to run FPD logic for AMP, per https://github.com/prebid/prebid-server/issues/1727. The HoldAuction endpoint also has knowledge of which endpoint it is serving through AuctionRequest.RequestType.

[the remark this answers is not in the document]
VeronikaSolovei9: FPD support fr AMP endpoint is not supported in PBS-go. In case we will need to do this, then the only way to do this - is to move FPD extraction inside of HoldAction. I will update the code this way.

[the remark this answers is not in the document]
VeronikaSolovei9: Code updated, please re-review

endpoints/openrtb2/auction.go
```diff
@@ -156,15 +155,6 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
 
-	resolvedFPD, fpdErrors := firstpartydata.ExtractFPDForBidders(req)
-	if len(fpdErrors) > 0 {
-		if errortypes.ContainsFatalError(fpdErrors) && writeError(fpdErrors, w, &labels) {
-			return
-		}
-		errL = append(errL, fpdErrors...)
-	}
-	warnings := errortypes.WarningOnly(errL)
-
```
bsardo, line 167: A couple of things here:
1) Do you recall what our line of thinking was on how to handle first party data errors? It looks like some errors are considered fatal and others are not. What types of errors fall into each of these categories?

2) It seems that by moving this logic down into `HoldAuction`, we might be returning a different error code when there is a fatal error. Here we were calling `writeError` which will return either a 400 or 503. When a fatal error occurs during `HoldAuction` FPD processing, it will be returned and handled in this function on line 226 which shows a 500 being returned:
```
if err != nil {
		labels.RequestStatus = metrics.RequestStatusErr
		w.WriteHeader(http.StatusInternalServerError)
		fmt.Fprintf(w, "Critical error while running the auction: %v", err)
		glog.Errorf("/openrtb2/auction Critical error: %v", err)
		ao.Status = http.StatusInternalServerError
		return
	}
```
I'm guessing a 500 doesn't make sense for most, if not all, of the FPD errors.

3) I'm trying to understand what the responsibility is of `Auction` vs `HoldAuction`. It seems like all request validation is done in `Auction` and any 400 errors would be caught there. In that case, it is probably not ideal to move `HoldAuction` down but I understand there are challenges with the other option we discussed so I'm ok with this strategy. I'm just pointing that out here since this just came to mind when I realized the potential http error status code discrepancy described in item 2.

exchange/exchange.go
```diff
@@ -190,6 +191,30 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+			//In order to preserve original request and return it properly in response.ext.debug.resolvedrequest
+			//original request should be copied before FPD execution
+			resolvedBidReq, err := json.Marshal(r.BidRequest)
+			if err != nil {
+				return nil, err
+			}
+			r.ResolvedBidRequest = resolvedBidReq
+		}
+
+		resolvedFPD, fpdErrors := firstpartydata.ExtractFPDForBidders(&openrtb_ext.RequestWrapper{BidRequest: r.BidRequest})
```
bsardo, line 207: Are we taking a step back here performance wise by having to wrap the bid request since the wrapper isn't available in `HoldAuction` at the moment? It seems like we could be unmarshalling fields we already unmarshalled. Something to check.

bsardo (commented):
I'm not finished with my review yet but wanted to share these comments.

endpoints/openrtb2/auction.go
VeronikaSolovei9: Error handling updated

exchange/exchange_test.go
```diff
@@ -3613,60 +3623,6 @@ func TestMakeBidExtJSON(t *testing.T) {
...
-
-	assert.NotNilf(t, outBidResponse, "outBidResponse should not be nil")
-	assert.Nil(t, err, "Error should be nil")
-
-	request := e.adapterMap[openrtb_ext.BidderAppnexus].(*capturingRequestBidder).req
-
-	assert.NotNil(t, request, "Bidder request should not be nil")
-	assert.Equal(t, apnFpd.Site, request.Site, "Site is incorrect")
-	assert.Equal(t, apnFpd.App, request.App, "App is incorrect")
-	assert.Equal(t, apnFpd.User, request.User, "User is incorrect")
```
bsardo, line 3666: I must be missing something, how did this pass before? The test is making sure that all extracted appnexus FPD data in `apnFpd` was merged into `request` but shouldn't just one of app or site merge in? Or actually maybe the validation logic wasn't executing during this test so this test was solely verifying that all three fields `app`, `site` and `user` are able be merged into the bid request.

endpoints/openrtb2/auction_test.go
```diff
@@ -3232,57 +3230,6 @@ func TestParseRequestParseImpInfoError(t *testing.T) {
 	assert.Contains(t, errL[0].Error(), "echovideoattrs of type bool", "Incorrect error message")
 }
 
-func TestAuctionFirstPartyData(t *testing.T) {
```
bsardo, line 3235: Do we need to delete this test? Is there another suitable E2E test for this already?

exchange/exchange.go
```diff
@@ -188,17 +189,46 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+		if responseDebugAllow {
+			//ExtractFPDForBidders removes FPD related structures from request.
+			//In order to preserve original request and return it properly in response.ext.debug.resolvedrequest
+			//original request should be copied before FPD execution
+			resolvedBidReq, err := json.Marshal(r.BidRequestWrapper.BidRequest)
+			if err != nil {
+				return nil, err
+			}
+			r.ResolvedBidRequest = resolvedBidReq
+		}
```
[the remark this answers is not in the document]
bsardo: Discussed offline. Move `if` statement on line 196 above `if` on line 194 so that the resolved bid request is set in the same way no matter what the request type is.

exchange/exchange.go
```diff
@@ -188,17 +189,46 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
-	responseDebugAllow, accountDebugAllow, debugLog := getDebugInfo(r.BidRequest, requestExt, r.Account.DebugAllow, debugLog)
+	responseDebugAllow, accountDebugAllow, debugLog := getDebugInfo(r.BidRequestWrapper.BidRequest, requestExt, r.Account.DebugAllow, debugLog)
+
+	if r.RequestType == metrics.ReqTypeORTB2Web || r.RequestType == metrics.ReqTypeORTB2App {
+		//Extract First party data for auction endpoint only
+		if responseDebugAllow {
+			//ExtractFPDForBidders removes FPD related structures from request.
+			//In order to preserve original request and return it properly in response.ext.debug.resolvedrequest
+			//original request should be copied before FPD execution
+			resolvedBidReq, err := json.Marshal(r.BidRequestWrapper.BidRequest)
```
bsardo, line 200: before marshalling let's call the function that rebuilds the bid request from any unmarshalled exts on the wrapper.

endpoints/openrtb2/sample-requests/valid-whole/supplementary/first-party-data.json
```diff
@@ -1,156 +0,0 @@
-{
```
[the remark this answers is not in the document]
VeronikaSolovei9: We don't call FPD logic from auction endpoint anymore. This test existed just to add coverage, we don't need them anymore. I added some more tests to exchange.

exchange/exchange_test.go
```diff
@@ -3613,60 +3623,6 @@ func TestMakeBidExtJSON(t *testing.T) {
 	}
 }
 
-func TestFPD(t *testing.T) {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, this test existed for coverage purposes only. 
No, not full flow end to end but every part of it is covered in separate test cases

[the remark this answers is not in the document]
VeronikaSolovei9: Right, HoldAuction does more things after we moved FPD extraction in, this test is not relevant anymore.

exchange/exchange_test.go
VeronikaSolovei9: Validation happened before HoldAction, we assumed resolved FPD is right, so we could merge everything. Now this is not the case anymore

exchange/exchange.go
VeronikaSolovei9: Added

exchange/exchange.go
```diff
@@ -870,17 +900,27 @@ func getPrimaryAdServer(adServerId int) (string, error) {
...
+			if err != nil {
+				errList = append(errList, err)
+			}
+		}
 		bidResponseExt.Debug = &openrtb_ext.ExtResponseDebug{
 			HttpCalls:       make(map[openrtb_ext.BidderName][]*openrtb_ext.ExtHttpCall),
-			ResolvedRequest: req,
+			ResolvedRequest: resolvedReq,
 		}
 	}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Modified

endpoints/openrtb2/auction_test.go
VeronikaSolovei9: This tests only checked fpd extraction from request. I don't think we have E2E test due to the way we mock functions. 
I'll check if I can add one. 
We have tests for all parts of FPD, but separately.

VeronikaSolovei9: Added!

exchange/exchange.go
```diff
@@ -188,17 +189,47 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
 
-	responseDebugAllow, accountDebugAllow, debugLog := getDebugInfo(r.BidRequest, requestExt, r.Account.DebugAllow, debugLog)
+	// rebuild/resync the request in the request wrapper.
+	if err := r.BidRequestWrapper.RebuildRequest(); err != nil {
+		return nil, err
+	}
+	responseDebugAllow, accountDebugAllow, debugLog := getDebugInfo(r.BidRequestWrapper.BidRequest, requestExt, r.Account.DebugAllow, debugLog)
+	if responseDebugAllow {
+		//save incoming request with stored requests (if applicable) to return in debug logs
```
AlexBVolcy, line 198: Nitpick: Maybe you could combine this comment with the comment on line 192 to explain what is happening in this code block

exchange/exchange.go
```diff
@@ -188,17 +189,47 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+		resolvedBidReq, err := json.Marshal(r.BidRequestWrapper.BidRequest)
+		if err != nil {
+			return nil, err
```
AlexBVolcy, line 201: Would it be possible to write a test to cover this error case in the code?

exchange/exchange.go
```diff
@@ -264,11 +295,11 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 		if targData != nil {
 			// A non-nil auction is only needed if targeting is active. (It is used below this block to extract cache keys)
-			auc = newAuction(adapterBids, len(r.BidRequest.Imp), targData.preferDeals)
+			auc = newAuction(adapterBids, len(r.BidRequestWrapper.BidRequest.Imp), targData.preferDeals)
 			auc.setRoundedPrices(targData.priceGranularity)
 
 			if requestExt.Prebid.SupportDeals {
-				dealErrs := applyDealSupport(r.BidRequest, auc, bidCategory)
+				dealErrs := applyDealSupport(r.BidRequestWrapper.BidRequest, auc, bidCategory)
```
AlexBVolcy, line 302: Could you write a test to cover this new block of code?

exchange/exchangetest/firstpartydata-multibidder-config-valid.json
```diff
@@ -0,0 +1,227 @@
+{
```
bsardo, line 1: Nitpick: can you use the same naming convention for the two FPD multiple bidder config tests?
Either make them:
- `exchange/exchangetest/firstpartydata-multiple-bidder-configs-valid.json`
- `exchange/exchangetest/firstpartydata-multiple-bidder-configs-invalid.json`
Or:
- `exchange/exchangetest/firstpartydata-multibidder-config-valid.json`
- `exchange/exchangetest/firstpartydata-multibidder-config-invalid.json`

exchange/exchangetest/firstpartydata-multibidder-config-valid.json
```diff
@@ -0,0 +1,227 @@
...
+          "ext": {
+            "prebid": {
+              "data": {
+                "eidpermissions": null,
+                "bidders": [
+                  "appnexus",
+                  "rubicon"
+                ]
+              },
+              "bidderconfig": [
```
bsardo, line 116: This data isn't typically sent in the request right? Is it in the bid requests here because of how something is mocked in this test?

exchange/exchange.go
VeronikaSolovei9: Not sure, those are 2 different pieces of logic.
If you have a vision of a good comment - I'll be happy to update it.

exchange/exchange.go
VeronikaSolovei9: In case of incorrect request such kind of errors will be detected before `HoldAction` execution during request validation functions. I don't see a value in test with broken bid request only to test Marshal and Unmarshal errors. This error handler exists to satisfy error handling approach we have in code already. Technically this error will never happen.

exchange/exchange.go
VeronikaSolovei9: This is beyond of the scope of this PR. I'll need to spend time and understand how to test deal support. There are no json tests for this. I would leave it for another PR and avoid to mix it to this PR, it becomes too complicated already.
We can discuss it with team.

exchange/exchangetest/firstpartydata-multibidder-config-valid.json
VeronikaSolovei9: Good catch! During the implementation I decided this is the set up of unit test. But it's not. Because of change from BidRequest to RequestWrapper any ext changes has to be updated in both RequestWrapper and previously extracted `requestExt`. I added related code in exchange.go. Optimization: RequestWrapper and `requestExt` runs only in case if any fpd was extracted.

exchange/exchange.go
```diff
@@ -192,14 +193,51 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 		_, targData.cacheHost, targData.cachePath = e.cache.GetExtCacheData()
 	}
 
-	responseDebugAllow, accountDebugAllow, debugLog := getDebugInfo(r.BidRequest, requestExt, r.Account.DebugAllow, debugLog)
+	// rebuild/resync the request in the request wrapper.
+	if err := r.BidRequestWrapper.RebuildRequest(); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Moved

exchange/targeting_test.go
```diff
@@ -524,99 +524,6 @@ var TargetingTests []TargetingTestData = []TargetingTestData{
 		},
 		TruncateTargetAttr: &truncateTargetAttrValueNegative,
 	},
-	{
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, good point! Reverted to it's original state

bsardo (approved):
LGTM

AlexBVolcy (approved):
LGTM

PR #2197, opened 2022-03-22
First party data fix for user.data and user.ext.data

VeronikaSolovei9:


firstpartydata/first_party_data.go
```diff
@@ -294,6 +300,19 @@ func mergeUsers(original *openrtb2.User, fpdConfigUser map[string]json.RawMessag
...
+	if userData, present := fpdConfigUser[dataKey]; present {
+		newUserData, err := unmarshalJSONToData(userData)
+		if err != nil {
+			return newUser, err
+		}
+		if len(newUser.Data) > 0 {
+			newUser.Data = append(newUser.Data, newUserData...)
+		} else {
+			newUser.Data = newUserData
+		}
```
SyntaxNode, line 312: `append` gracefully handles `nil` conditions. You can replace:
```
if len(newUser.Data) > 0 {
    newUser.Data = append(newUser.Data, newUserData...)
} else {
    newUser.Data = newUserData
}
```

with

```
newUser.Data = append(newUser.Data, newUserData...)
```

bsardo:
This is looking good. I tested it locally using a couple of sample requests including the request provided in the issue.

TheMediaGrid:
Thank you!
Looks like it will become part of Prebid-server GO 0.205. Will there be any requirements on the PBJS version side so that the user.data is passed from FPD to PBS bidders?

SyntaxNode (approved):

bsardo:
@TheMediaGrid Yes, this should be part of PBS-Go 0.205.0. My team doesn't work on PBJS so I'm not super familiar with the code base, however, from looking at the [Prebid Server and First Party Data, section OpenRTB Interface](https://docs.google.com/document/d/1I2eS40yMYJCgz8XvFljf4f59bEPsbWTrUpnD7a7uJEg/edit#heading=h.io1s8uic128u), I don't think any change is required for PBJS. Those requirements indicate that first party data can be set in the bidderconfig using `setBidderConfig()` so that should suffice. PBS will then analyze the bidderconfig and merge the appropriate pieces of information into the root objects `app`, `site` and `user` for the specified bidders.

PR #2200, opened 2022-03-23
PubMatic: add "acat" parameter

pm-nilesh-chate:
Add support to 'act' request ext param in PubMatic adapter.

Documentation https://docs.prebid.org/dev-docs/pbs-bidders.html#pubmatic (https://github.com/prebid/prebid.github.io/pull/3657)

Prebid JS changes are WIP.


pm-nilesh-chate:
moved to https://github.com/prebid/prebid-server/pull/2201

PR #2202, opened 2022-03-24
Improve Digital adapter: Google additional consent support

jbartek25:
Transform Google additional consent into format understood by the Improve Digital ad server

### DIFF
--- a/adapters/improvedigital/improvedigital.go
+++ b/adapters/improvedigital/improvedigital.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/mxmCherry/openrtb/v15/openrtb2"
 	"github.com/prebid/prebid-server/adapters"
@@ -36,6 +37,8 @@ func (a *ImprovedigitalAdapter) MakeRequests(request *openrtb2.BidRequest, reqIn
 
 func (a *ImprovedigitalAdapter) makeRequest(request openrtb2.BidRequest, imp openrtb2.Imp) (*adapters.RequestData, error) {
 	request.Imp = []openrtb2.Imp{imp}
+
+	request = a.applyAdditionalConsentProvidersArray(request)
 	reqJSON, err := json.Marshal(request)
 	if err != nil {
 		return nil, err
@@ -143,4 +146,78 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
 	return "", &errortypes.BadServerResponse{
 		Message: fmt.Sprintf("Failed to find impression for ID: \"%s\"", impID),
 	}
+}
+
+// This method responsible to clone request and convert additional consent providers string to array when additional consent provider found
+func (a *ImprovedigitalAdapter) applyAdditionalConsentProvidersArray(request openrtb2.BidRequest) openrtb2.BidRequest {
+	// If user/user.ext not defined, no need to parse additional consent
+	if request.User == nil || request.User.Ext == nil {
+		return request
+	}
+
+	var (
+		err     error
+		reqJSON json.RawMessage
+		cpStr   string
+	)
+
+	// Start validating additional consent
+	// Check key exist user.ext.ConsentedProvidersSettings.consented_providers
+	var userExtMap = make(map[string]json.RawMessage)
+	if err = json.Unmarshal(request.User.Ext, &userExtMap); err != nil {
+		return request
+	}
+
+	cpsMapValue, cpsJSONFound := userExtMap["ConsentedProvidersSettings"]
+	if !cpsJSONFound {
+		return request
+	}
+
+	// Check key exist user.ext.ConsentedProvidersSettings.consented_providers
+	var cpMap = make(map[string]json.RawMessage)
+	if err = json.Unmarshal(cpsMapValue, &cpMap); err != nil {
+		return request
+	}
+
+	cpMapValue, cpJSONFound := cpMap["consented_providers"]
+	if !cpJSONFound {
+		return request
+	}
+	// End validating additional consent
+
+	// Check if string contain ~, then substring after ~ to end of string
+	consentStr := string(cpMapValue)
+	var tildaPosition int
+	if tildaPosition = strings.Index(consentStr, "~"); tildaPosition == -1 {
+		return request
+	}
+	cpStr = consentStr[tildaPosition+1 : len(consentStr)-1]
+
+	// Prepare consent providers string for append
+	cpStr = fmt.Sprintf("[%s]", strings.Replace(cpStr, ".", ",", -1))
+	cpMap["consented_providers"] = json.RawMessage(cpStr)
+
+	cpJSON, _ := json.Marshal(cpMap)
+	userExtMap["consented_providers_settings"] = cpJSON
+
+	extJson, extErr := json.Marshal(userExtMap)
+	if extErr != nil {
+		return request
+	}
+
+	// Start request closing
+	reqJSON, err = json.Marshal(request)
+	if err != nil {
+		return request
+	}
+
+	var reqClone openrtb2.BidRequest
+	if err = json.Unmarshal(reqJSON, &reqClone); err != nil {
+		return request
+	}
+	// End request closing
+
+	reqClone.User.Ext = extJson
+
+	return reqClone
 }
--- a/adapters/improvedigital/improvedigital.go
+++ b/adapters/improvedigital/improvedigital.go
@@ -38,7 +38,13 @@ func (a *ImprovedigitalAdapter) MakeRequests(request *openrtb2.BidRequest, reqIn
 func (a *ImprovedigitalAdapter) makeRequest(request openrtb2.BidRequest, imp openrtb2.Imp) (*adapters.RequestData, error) {
 	request.Imp = []openrtb2.Imp{imp}
 
-	request = a.applyAdditionalConsentProvidersArray(request)
+	userExtAddtlConsent := a.getAdditionalConsentProvidersUserExt(request)
+	if len(userExtAddtlConsent) > 0 {
+		userCopy := *request.User
+		userCopy.Ext = userExtAddtlConsent
+		request.User = &userCopy
+	}
+
 	reqJSON, err := json.Marshal(request)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +147,77 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+	var (
+		err     error
+		reqJSON json.RawMessage
+		cpStr   string
+	)
+
+	// Start validating additional consent
+	// Check key exist user.ext.ConsentedProvidersSettings.consented_providers
+	var userExtMap = make(map[string]json.RawMessage)
+	if err = json.Unmarshal(request.User.Ext, &userExtMap); err != nil {
```
AlexBVolcy, line 167: Is there a scenario where we would want to handle the error here differently than just returning the request?

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +147,77 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+	}
+
+	cpsMapValue, cpsJSONFound := userExtMap["ConsentedProvidersSettings"]
+	if !cpsJSONFound {
+		return request
+	}
+
+	// Check key exist user.ext.ConsentedProvidersSettings.consented_providers
+	var cpMap = make(map[string]json.RawMessage)
+	if err = json.Unmarshal(cpsMapValue, &cpMap); err != nil {
```
AlexBVolcy, line 178: Same comment as above, is there a scenario where we would want to handle the error here differently than just returning the request?

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +147,77 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+
+	// Prepare consent providers string for append
+	cpStr = fmt.Sprintf("[%s]", strings.Replace(cpStr, ".", ",", -1))
+	cpMap["consented_providers"] = json.RawMessage(cpStr)
+
+	cpJSON, _ := json.Marshal(cpMap)
+	userExtMap["consented_providers_settings"] = cpJSON
+
+	extJson, extErr := json.Marshal(userExtMap)
+	if extErr != nil {
```
AlexBVolcy, line 204: Same error related comment as above

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +147,77 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+		return request
+	}
+
+	// Start request closing
+	reqJSON, err = json.Marshal(request)
+	if err != nil {
```
AlexBVolcy, line 210: Same error related comment as above

adapters/improvedigital/improvedigital.go
AlexBVolcy: Could you also add a test that triggers this error case here?

AlexBVolcy (commented):
Left some error related comments

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +147,77 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+	consentStr := string(cpMapValue)
+	var tildaPosition int
+	if tildaPosition = strings.Index(consentStr, "~"); tildaPosition == -1 {
+		return request
+	}
+	cpStr = consentStr[tildaPosition+1 : len(consentStr)-1]
+
+	// Prepare consent providers string for append
+	cpStr = fmt.Sprintf("[%s]", strings.Replace(cpStr, ".", ",", -1))
+	cpMap["consented_providers"] = json.RawMessage(cpStr)
```
VeronikaSolovei9, line 198: Please extract all hardcoded string values to constants

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +147,77 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+		return request
+	}
+
+	var reqClone openrtb2.BidRequest
+	if err = json.Unmarshal(reqJSON, &reqClone); err != nil {
+		return request
+	}
+	// End request closing
+
+	reqClone.User.Ext = extJson
```
VeronikaSolovei9, line 220: Instead of Unmarshal request to clone request and then Marshal it back right after this function call, would it be easier to return updated User ext and set it if it was changed after `cpStr` evaluation?

VeronikaSolovei9: It would be better if you return error and data (request or user.ext part of it).

adapters/improvedigital/improvedigital.go
samiul-shanto: applyAdditionalConsentProvidersArray method supposes to handle only additional consent-related logic. If it can't parse necessary data, it just returns the original request and don't interfere with request.

PBS already validates request data (as like if user.ext assigned a string or an array then validation error occurs). So usually it won't go into the return statement (line: 168). but we add an error handler here to avoid any surprise error.

adapters/improvedigital/improvedigital.go
samiul-shanto: We will update hardcoded string to constants

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +147,77 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+	cpStr = consentStr[tildaPosition+1 : len(consentStr)-1]
+
+	// Prepare consent providers string for append
+	cpStr = fmt.Sprintf("[%s]", strings.Replace(cpStr, ".", ",", -1))
+	cpMap["consented_providers"] = json.RawMessage(cpStr)
+
+	cpJSON, _ := json.Marshal(cpMap)
+	userExtMap["consented_providers_settings"] = cpJSON
+
+	extJson, extErr := json.Marshal(userExtMap)
```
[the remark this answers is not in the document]
samiul-shanto: No, we don't have a defined data format. Prebid have ExtUser struct, but it is possible that the user can add additional data and we have to keep all user's data in request.

adapters/improvedigital/improvedigital.go
samiul-shanto: We can't change user.ext of original request (test will fail). So, we must have to masshal/unmarshal it again.

Sure, we will update our code so it will return data and error (request and error)

VeronikaSolovei9: Right, you need to make request to have it's own User copy:
```
newUserExt := a.applyAdditionalConsentProvidersArray(request) <- returns []byte
	if len(newUserExt) > 0 {
		userCopy := *request.User
		userCopy.Ext = newUserExt
		request.User = &userCopy
	}
```
You will not need to create a clone of the entire request.
Here is a link to documentations: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#makerequests , navigate to `Example: Mutating banner shared memory (make a copy).`

adapters/improvedigital/improvedigital.go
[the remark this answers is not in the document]
VeronikaSolovei9: I see. Then no changes are needed here.

samiul-shanto:
@VeronikaSolovei9 We have updated code as per your suggestion, please take a look.

Thanks

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +153,68 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+	if tildaPosition = strings.Index(consentStr, "~"); tildaPosition == -1 {
+		return []byte{}
+	}
+	cpStr = consentStr[tildaPosition+1 : len(consentStr)-1]
+
+	// Prepare consent providers string
+	cpStr = fmt.Sprintf("[%s]", strings.Replace(cpStr, ".", ",", -1))
+	cpMap[consentedProvidersKey] = json.RawMessage(cpStr)
+
+	cpJSON, _ := json.Marshal(cpMap)
```
VeronikaSolovei9, line 211: Please do not ignore errors. May you please return nil instead of []byte and error if any: 
```
if err != nil {
    return nil, err
}
```
Upper level expression `len(userExtAddtlConsent) > 0` handled nil properly.

VeronikaSolovei9 (commented):
One minor comment, other than that looks good

adapters/improvedigital/improvedigital.go
AlexBVolcy: Yeah i second this

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +153,68 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+
+	// Check key exist user.ext.ConsentedProvidersSettings.consented_providers
+	var cpMap = make(map[string]json.RawMessage)
+	if err = json.Unmarshal(cpsMapValue, &cpMap); err != nil {
+		return []byte{}
+	}
+
+	cpMapValue, cpJSONFound := cpMap[consentedProvidersKey]
+	if !cpJSONFound {
+		return []byte{}
```
AlexBVolcy, line 195: Is there a way to test this error case? If so could you update the tests to cover it.

adapters/improvedigital/improvedigital.go
```diff
@@ -144,3 +153,68 @@ func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType,
...
+	}
+	// End validating additional consent
+
+	// Check if string contain ~, then substring after ~ to end of string
+	consentStr := string(cpMapValue)
+	var tildaPosition int
+	if tildaPosition = strings.Index(consentStr, "~"); tildaPosition == -1 {
+		return []byte{}
```
AlexBVolcy, line 203: Is there a way to test this error case? If so could you update the tests to cover it.

AlexBVolcy (commented):
Looks good, just left a few tiny comments

VeronikaSolovei9 (approved):
LGTM.

AlexBVolcy (approved):
LGTM

PR #2203, opened 2022-03-25
Remove Restrictions for UUID Generation For Stored Bid Requests

AlexBVolcy:
This PR is a follow to two previous PRs https://github.com/prebid/prebid-server/pull/1938 and https://github.com/prebid/prebid-server/pull/2000

In those PRs, the goal was to generate a UUID for when an **app** or amp request had a stored bid request and that stored bid request had an ID = {{UUID}} or a config flag `GenerateRequestId` was set to true. 

This update removes the restriction where the auction endpoint would check if the request was an app request, and instead, as long as the request has a stored request, and the macro {{UUID}} or the config flag is set to true, the UUID would be generated.

### DIFF
--- a/endpoints/openrtb2/auction.go
+++ b/endpoints/openrtb2/auction.go
@@ -1661,7 +1661,11 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	resolvedRequest := requestJson
 
 	if hasStoredBidRequest {
-		if deps.cfg.GenerateRequestID || bidRequestID == "{{UUID}}" {
+		isAppRequest, err := checkIfAppRequest(requestJson)
+		if err != nil {
+			return nil, nil, []error{err}
+		}
+		if (deps.cfg.GenerateRequestID && isAppRequest) || bidRequestID == "{{UUID}}" {
 			uuidPatch, err := generateUuidForBidRequest(deps.uuidGenerator)
 			if err != nil {
 				return nil, nil, []error{err}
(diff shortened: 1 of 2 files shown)


endpoints/openrtb2/auction_test.go
```diff
@@ -1879,7 +1879,7 @@ func TestStoredRequestGenerateUuid(t *testing.T) {
 			expectedID:             "ThisID",
 		},
 		{
-			description:            "GenerateRequestID is true, but rawData is a site request, we should not generate uuid",
+			description:            "GenerateRequestID is true, and rawData is a site request, we should generate uuid",
```
hhhjort, line 1882: If we should generate a uuid, then shouldn't expectedID be uuid?

endpoints/openrtb2/auction.go
```diff
@@ -1661,11 +1661,7 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	resolvedRequest := requestJson
 
 	if hasStoredBidRequest {
-		isAppRequest, err := checkIfAppRequest(requestJson)
-		if err != nil {
-			return nil, nil, []error{err}
-		}
-		if isAppRequest && (deps.cfg.GenerateRequestID || bidRequestID == "{{UUID}}") {
+		if deps.cfg.GenerateRequestID || bidRequestID == "{{UUID}}" {
```
SyntaxNode, line 1664: The original feature request changed a few times, but I believe we decided to generate a new bid request id for app and amp auctions. I'm good with expanding this to any request type if the special {{UUID}} token is used in the stored request, but the deps.cfg.GenerateRequestID should still be scoped to app and amp requests, IMHO.


deps.cfg.GenerateRequestID || bidRequestID == "{{UUID}}"

hhhjort (approved):
LGTM

PR #2204, opened 2022-03-25
Apply new Truncate Targeting Config to Key

AlexBVolcy:
This PR is a follow up to https://github.com/prebid/prebid-server/pull/2181

In this PR, I update the `addKeys` function to also apply the truncation to the Targeting Key itself


bsardo (approved):
LGTM

PR #2205, opened 2022-03-25
Colossus Bid Adapter: add new param

HuddledMasses:
- [x] new feature
doc - https://github.com/prebid/prebid.github.io/pull/3663


adapters/colossus/colossus.go
```diff
@@ -29,21 +28,25 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
-		tagID, err = jsonparser.GetString(reqCopy.Imp[0].Ext, "bidder", "TagID")
-		if err != nil {
-			errs = append(errs, err)
-			continue
+		if err = json.Unmarshal(reqCopy.Imp[0].Ext, &bidderExt); err != nil {
+			return nil, append(errs, err)
+		}
+		if err = json.Unmarshal(bidderExt.Bidder, &collossusExt); err != nil {
+			return nil, append(errs, err)
 		}
```
SyntaxNode, line 45: I'm curious, why the change from using `jsonparser.GetString`?

adapters/colossus/colossus.go
```diff
@@ -29,21 +28,25 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
 func (a *ColossusAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
 	var errs []error
 	var err error
-	var tagID string
 
+	var bidderExt adapters.ExtImpBidder
+	var collossusExt openrtb_ext.ExtImpColossus
```
VeronikaSolovei9, line 33: Should it be one `L` in colossus?

HuddledMasses: thanks, fixed

adapters/colossus/colossus.go
HuddledMasses: `jsonparser.GetString` returns an error if the path is missing and `json.Unmarshal` returns an empty string. I thought that focusing on the error is not entirely correct. Therefore, I replaced it. Could you share a possibly better solution? Thanks.

AlexBVolcy (commented):
This looks good! The code coverage is also good. I'm gonna wait and see if there are any suggestions for a better solution regarding `GetString` vs. `Unmarshal` before approval.

adapters/colossus/colossus.go
SyntaxNode: I see. Thank you for explaining. I ask because `jsonparser` is orders of magnitude more efficient when accessing a single value from json. You are right that an error is returned if the path is missing, but luckily that error always has a specific value you could filter out, something like:

```
if tagID, err := jsonparser.GetString(reqCopy.Imp[0].Ext, "bidder", "TagID"); err == nil {
	reqCopy.Imp[0].TagID = tagID
} else if err !=  jsonparser.KeyPathNotFoundError {
	errs = append(errs, err)
	continue
}
```

HuddledMasses: thanks, i like your solution! =)

adapters/colossus/colossus.go
```diff
@@ -27,25 +28,17 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters
...
-			return nil, append(errs, err)
-		}
-
-		if colossusExt.TagID != "" {
-			reqCopy.Imp[0].TagID = colossusExt.TagID
+		if tagID, err := jsonparser.GetString(reqCopy.Imp[0].Ext, "bidder", "TagID"); err == nil {
+			reqCopy.Imp[0].TagID = tagID
+		} else if err != jsonparser.KeyPathNotFoundError {
+			errs = append(errs, err)
+			continue
```
AlexBVolcy, line 41: Would it be feasible to update tests to cover this error case?

AlexBVolcy (commented):
Only one minor comment from me about test coverage

adapters/colossus/colossus.go
HuddledMasses: I'm sorry, but I can't think of a test for this piece of code. Perhaps you can suggest something?

HuddledMasses:
@VeronikaSolovei9 @SyntaxNode @AlexBVolcy please take a look at this pull request

adapters/colossus/colossus.go
AlexBVolcy: I think it's alright to leave this out, since other test coverage looks great

AlexBVolcy (approved):

adapters/colossus/colossus.go
```diff
@@ -122,18 +118,22 @@ func (a *ColossusAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 }
 
 func getMediaTypeForImp(impID string, imps []openrtb2.Imp) (openrtb_ext.BidType, error) {
```
VeronikaSolovei9, line 120: There is one thing for you to consider. I case if imp has multi format this function may return incorrect bid type. We recommend to extract mediaType from response. Here is the example for reference: https://github.com/prebid/prebid-server/blob/74d2199c7dd7e6b2c28d67c26cda2c4405ca5b6f/adapters/algorix/algorix.go#L190

HuddledMasses: thanks, i have updated

VeronikaSolovei9 (approved):
Thank you for update, lgtm

hhhjort (approved):
LGTM

Issue #1817, opened 2021-04-19
Support AMP consentType, gdprApplies, additionalConsent


bretg:
We're a bit behind on this, but as noted in AMP issue 30006, Prebid Server can pull several new values from the AMP interface as documented in https://github.com/ampproject/amphtml/blob/main/extensions/amp-consent/integrating-consent.md#consentmetadata

1. consent_type=CONSENT_METADATA(consentStringType)
1. gdpr_applies=CONSENT_METADATA(gdprApplies)
1. addtl_consent=CONSENT_METADATA(additionalConsent)

e.g.

```
...&consent_type=2&gdpr_applies=true&addtl_consent=aaaaaaaaaaaaaa&...
```

## Details

Note that the AMP parameter gdpr_consent carries both GDPR and CCPA values.

1. If consent_type is provided, it will be an enum containing the following values:
   - 1: The contents of gdpr_consent can be treated as TCF V1. We no longer support TCFv1, so ignore any consent_string provided on the request.
   - 2: The contents of gdpr_consent can be treated as TCF V2. If the consent_string isn't a valid TCF2 string, assume there's no user consent for any purpose as if no gdpr_consent were provided.
   - 3: The contents of gdpr_consent can be treated as a US Privacy String. If the consent_string isn't a valid USP string, assume for the purposes of opt-out-of-sale enforcement that no gdpr_consent was provided.
   - anything else: ignore, log a warning, and assume no gdpr_consent was provided
1. If gdpr_applies is supplied, use its value to set `regs.ext.gdpr`: gdpr_applies=true means regs.ext.gdpr:1, gdpr_applies=false means regs.ext.gdpr:0, any other value means regs.ext.gdpr is not set.
1. If consent_type="2", and gdpr_consent is not empty, then copy gdpr_consent to user.ext.consent
1. If consent_type="3", and gdpr_consent is not empty, then copy gdpr_consent to regs.ext.us_privacy
1. If addtl_consent is supplied, copy its value to `user.ext.ConsentedProvidersSettings.consented_providers`

snahornyi:
@bretg You mentioned that `gdpr_applies=1` this param can be of number value, but from doc we can see that it can be only boolean param(`true` or `false`), so should `0` and `1` also be supported?

bretg:
example fixed - thanks for pointing it out

bretg:
done in PBS-Java with https://github.com/prebid/prebid-server-java/pull/1379

bretg:
[...]

PR #2213, opened 2022-04-04
Yieldlab: Fix GDPR handling for AMP

mirkorean:
We do want to send the consent string even though we are not getting a gdpr flag (We will default to 1 internally and respect the gdpr consent). This problems currently exists when requested via AMP.

### DIFF
--- a/adapters/yieldlab/yieldlab.go
+++ b/adapters/yieldlab/yieldlab.go
@@ -76,8 +76,10 @@ func (a *YieldlabAdapter) makeEndpointURL(req *openrtb2.BidRequest, params *open
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


hhhjort (approved):
Looks good to me. Reviewing how we handle GDPR under the AMP endpoint, it seems that we are not setting `regs.ext.gdpr` on AMP requests that need it. This is something that we need to fix, and should help your AMP GDPR cases once it is done.

mirkorean:
Thanks @hhhjort. The problem with the gpdr-flag (gdpr_applies) with PBS-Go is known and discussed partly here: https://github.com/prebid/prebid-server/issues/1817

AlexBVolcy (approved):
LGTM

Issue #2147, opened 2022-02-03
Listen to unix socket instead of tcp ?


matthieularere-msq:
Is there currently a way to make prebid-server listen to a unix socket instead of tcp port ? My prebid-server isn't exposed but behind a app acting as a proxy to build the openrtb request send to prebid-server, so using unix socket could avoid me to reach tcp limits.

bretg:
@matthieularere-msq - can you please elaborate on this request?

- Prebid Server is HTTP, which is TCP.
- Why can't you just raise the TCP limit? Or lower the keepalive?
- Would HTTP2 connection pooling be a solution?
 
Sounds like you have a server running on the same box and can assume a same-box connection to PBS?

matthieularere-msq:
Indeed, it's a same-box connection to PBS.
I can try fine tune TCP but at the end the issue could reappear later as a limit would remain. That's why I thought about unix socket.
HTTP2 could worth a try. Apart from using http2 package would it requires a lot of code changes ?

bretg:
One of the engineers will have to weigh in on whether there's much work to support HTTP2 inbound for PBS-Go and/or PBS-Java. That seems like the easiest approach here, IMO.

SyntaxNode:
@matthieularere-msq So you are looking to use HTTP over Unix Sockets?

There is not currently an option to listen on Unix Sockets, but I don't see a reason why not to give the Host an option. Are you interested in putting up a PR for this?

I found an [example here](https://gist.github.com/teknoraver/5ffacb8757330715bcbcc90e6d46ac74) for how to bind to Unix Sockets in Go. Looks simple. The appropriate place to implement this in PBS-Go is: https://github.com/prebid/prebid-server/blob/2ac9d20db8667cf75a5e891e73e049d0e80eb4ea/server/server.go#L96

You would need to pass the configuration down to those calls (one for admin endpoints and another for the main endpoints). Are you using Prometheus? I haven't looked into how to add Unix Socket support on that endpoint.

I'm also cool with using HTTP2. I think PBS-Go already has support for HTTP2 incoming connections if you'd like to give it a try.

matthieularere-msq:
@SyntaxNode Thanks, I'll try to see if I manage to make unix socket work, if so I'll submit a PR.

I've tested pbs-go with http2 but it's not supported:
`$ curl -I --http2-prior-knowledge http://localhost:8000/openrtb2/auction
curl: (16) Error in the HTTP2 framing layer`

[...]

PR #2215, opened 2022-04-07
Allow http support for unix_socket

mediasquare-alexandre:
REF : issue #2147 [feature request]

Allow users to listen unix_socket's file of main_server's handlers.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,15 +18,15 @@ import (
 
 // Configuration specifies the static application config.
 type Configuration struct {
-	ExternalURL  string     `mapstructure:"external_url"`
-	Host         string     `mapstructure:"host"`
-	Port         int        `mapstructure:"port"`
-	EnableSocket bool       `mapstructure:"enable_socket"`
-	Socket       string     `mapstructure:"socket"`
-	Client       HTTPClient `mapstructure:"http_client"`
-	CacheClient  HTTPClient `mapstructure:"http_client_cache"`
-	AdminPort    int        `mapstructure:"admin_port"`
-	EnableGzip   bool       `mapstructure:"enable_gzip"`
+	ExternalURL      string     `mapstructure:"external_url"`
+	Host             string     `mapstructure:"host"`
+	Port             int        `mapstructure:"port"`
+	UnixSocketEnable bool       `mapstructure:"unix_socket_enable"`
+	UnixSocketName   string     `mapstructure:"unix_socket_name"`
+	Client           HTTPClient `mapstructure:"http_client"`
+	CacheClient      HTTPClient `mapstructure:"http_client_cache"`
+	AdminPort        int        `mapstructure:"admin_port"`
+	EnableGzip       bool       `mapstructure:"enable_gzip"`
 	// GarbageCollectorThreshold allocates virtual memory (in bytes) which is not used by PBS but
 	// serves as a hack to trigger the garbage collector only when the heap reaches at least this size.
 	// More info: https://github.com/golang/go/issues/48409
@@ -741,8 +741,8 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("external_url", "http://localhost:8000")
 	v.SetDefault("host", "")
 	v.SetDefault("port", 8000)
-	v.SetDefault("enable_socket", false)         // boolean which decide if the socket-server will be started.
-	v.SetDefault("socket", "prebid-server.sock") // path of the socket's file which must be listened.
+	v.SetDefault("unix_socket_enable", false)              // boolean which decide if the socket-server will be started.
+	v.SetDefault("unix_socket_name", "prebid-server.sock") // path of the socket's file which must be listened.
 	v.SetDefault("admin_port", 6060)
 	v.SetDefault("enable_gzip", false)
 	v.SetDefault("garbage_collector_threshold", 0)
--- a/config/config.go
+++ b/config/config.go
@@ -18,14 +18,15 @@ import (
 
 // Configuration specifies the static application config.
 type Configuration struct {
-	ExternalURL string     `mapstructure:"external_url"`
-	Host        string     `mapstructure:"host"`
-	Port        int        `mapstructure:"port"`
-	Socket      string     `mapstructure:"socket"`
-	Client      HTTPClient `mapstructure:"http_client"`
-	CacheClient HTTPClient `mapstructure:"http_client_cache"`
-	AdminPort   int        `mapstructure:"admin_port"`
-	EnableGzip  bool       `mapstructure:"enable_gzip"`
+	ExternalURL  string     `mapstructure:"external_url"`
+	Host         string     `mapstructure:"host"`
+	Port         int        `mapstructure:"port"`
+	EnableSocket bool       `mapstructure:"enable_socket"`
+	Socket       string     `mapstructure:"socket"`
+	Client       HTTPClient `mapstructure:"http_client"`
+	CacheClient  HTTPClient `mapstructure:"http_client_cache"`
+	AdminPort    int        `mapstructure:"admin_port"`
+	EnableGzip   bool       `mapstructure:"enable_gzip"`
 	// GarbageCollectorThreshold allocates virtual memory (in bytes) which is not used by PBS but
 	// serves as a hack to trigger the garbage collector only when the heap reaches at least this size.
 	// More info: https://github.com/golang/go/issues/48409
@@ -740,6 +741,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("external_url", "http://localhost:8000")
 	v.SetDefault("host", "")
 	v.SetDefault("port", 8000)
+	v.SetDefault("enable_socket", false)         // boolean which decide if the socket-server will be started.
 	v.SetDefault("socket", "prebid-server.sock") // path of the socket's file which must be listened.
 	v.SetDefault("admin_port", 6060)
 	v.SetDefault("enable_gzip", false)
--- a/config/config.go
+++ b/config/config.go
@@ -21,6 +21,7 @@ type Configuration struct {
 	ExternalURL string     `mapstructure:"external_url"`
 	Host        string     `mapstructure:"host"`
 	Port        int        `mapstructure:"port"`
+	Socket      string     `mapstructure:"socket"`
 	Client      HTTPClient `mapstructure:"http_client"`
 	CacheClient HTTPClient `mapstructure:"http_client_cache"`
 	AdminPort   int        `mapstructure:"admin_port"`
@@ -739,6 +740,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("external_url", "http://localhost:8000")
 	v.SetDefault("host", "")
 	v.SetDefault("port", 8000)
+	v.SetDefault("socket", "prebid-server.sock") // path of the socket's file which must be listened.
 	v.SetDefault("admin_port", 6060)
 	v.SetDefault("enable_gzip", false)
 	v.SetDefault("garbage_collector_threshold", 0)
--- a/server/server.go
+++ b/server/server.go
@@ -32,7 +32,7 @@ func Listen(cfg *config.Configuration, handler http.Handler, adminHandler http.H
 	adminServer := newAdminServer(cfg, adminHandler)
 	go shutdownAfterSignals(adminServer, stopAdmin, done)
 
-	if cfg.EnableSocket && len(cfg.Socket) > 0 { // start the unix_socket server if config enable-it.
+	if cfg.UnixSocketEnable && len(cfg.UnixSocketName) > 0 { // start the unix_socket server if config enable-it.
 		var (
 			socketListener net.Listener
 			mainServer     = newSocketServer(cfg, handler)
@@ -42,7 +42,7 @@ func Listen(cfg *config.Configuration, handler http.Handler, adminHandler http.H
 			glog.Errorf("Error listening for Unix-Socket connections on path %s: %v for socket server", mainServer.Addr, err)
 			return
 		}
-		go runServer(mainServer, "Socket", socketListener)
+		go runServer(mainServer, "UnixSocket", socketListener)
 	} else { // start the TCP server
 		var (
 			mainListener net.Listener
@@ -112,7 +112,7 @@ func newSocketServer(cfg *config.Configuration, handler http.Handler) *http.Serv
 	}
 
 	return &http.Server{
-		Addr:         cfg.Socket,
+		Addr:         cfg.UnixSocketName,
 		Handler:      serverHandler,
 		ReadTimeout:  15 * time.Second,
 		WriteTimeout: 15 * time.Second,
--- a/server/server.go
+++ b/server/server.go
@@ -117,12 +117,11 @@ func newSocketServer(cfg *config.Configuration, handler http.Handler) *http.Serv
 }
 
 func runServer(server *http.Server, name string, listener net.Listener) (err error) {
-	switch {
-	case server == nil:
+	if server == nil {
 		err = fmt.Errorf(">> Server is a nil_ptr.")
 		glog.Errorf("%s server quit with error: %v", name, err)
 		return
-	case listener == nil:
+	} else if listener == nil {
 		err = fmt.Errorf(">> Listener is a nil.")
 		glog.Errorf("%s server quit with error: %v", name, err)
 		return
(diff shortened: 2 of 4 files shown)


server/server.go
```diff
@@ -29,29 +29,41 @@ func Listen(cfg *config.Configuration, handler http.Handler, adminHandler http.H
...
 	go shutdownAfterSignals(mainServer, stopMain, done)
 
-	mainListener, err := newListener(mainServer.Addr, metrics)
+	socketListener, err := newUnixListener(socketServer.Addr, metrics)
+	if err != nil {
+		glog.Errorf("Error listening for Unix-Socket connections on path %s: %v for socket server", socketServer.Addr, err)
+		return
+	}
+
+	mainListener, err := newTCPListener(mainServer.Addr, metrics)
```
SyntaxNode, line 47: Interesting. Do you have a need to listen both for http and unix sockets?

config/config.go
```diff
@@ -656,6 +657,7 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("external_url", "http://localhost:8000")
 	v.SetDefault("host", "")
 	v.SetDefault("port", 8000)
+	v.SetDefault("socket", "prebid-server.sock") // path of the socket's file which must be listened.
```
SyntaxNode, line 660: I'd also like to see an option for enabling/disabling this functionality, defaulting to disabled.

SyntaxNode (commented):
Thank you for the contribution. I have a few comments in regards to enabling/disabling the socket listener. I understand the lack of unit tests for the listener if it's too hard to simulate, but can you please confirm if you've manually tested this feature end to end?

config/config.go
mediasquare-alexandre: done, commit reference : [d26af7c]

enabling & disabling set into config_file and default_value "false".

mediasquare-alexandre: ```Interesting. Do you have a need to listen both for http and unix sockets?```
I do not commit : [6838230] 
Handle this, if unix_socket enable, the main_server will be start as unix_socket and won't be start in TCP anymore.

```I understand the lack of unit tests for the listener if it's too hard to simulate, but can you please confirm if you've manually tested this feature end to end?```
Sure.

**Comment by mediasquare-alexandre** [comment]
@SyntaxNode : Is everything good for you now ?

**Comment by VeronikaSolovei9** [comment]
Hello, I tested these changes on my local machine and it all works good. Code is simple and straight forward. 
Validation now fails due to insufficient code coverage
```
github.com/prebid/prebid-server/server	0.013s	coverage: 29.1% of statements
Package has less than 30% code coverage. Run ./scripts/coverage.sh --html to see a detailed coverage report, and add tests to improve your coverage
```
Can you please add some tests? 

Also I have a question. How many requests per second or how many concurrent requests socket server can handle? I only needed to create one `/file.sock` and not sure if this server can handle more than one request at a time.

**Comment by mediasquare-alexandre** [comment]
@VeronikaSolovei9 : With commit [41612c6] coverage rate is ~40%. Most of the uncover part was already merge into master and are about chan and no-return function.

```Also I have a question. How many requests per second or how many concurrent requests socket server can handle? I only needed to create one /file.sock and not sure if this server can handle more than one request at a time.``` About this, I still have some tests and documentation to check.

**Comment by AlexBVolcy** [comment]
Just want to point out that these following tests are failing: 

`Test_newMainServer`
`Test_newTCPListener`
`Test_runServer`

**Comment by mediasquare-alexandre** [comment]
@AlexBVolcy [SOLVED] [89dd668], it was due to :
```//chose_your_socket_port
//chose_your_socket_address
//chose_your_host_value``` 
which weren't initialized as usable values for tests.
I check about some more of your code and set those values as some const in test_file.
```const (
		chose_your_socket_port    = 8000                        //chose_your_socket_port
		chose_your_socket_address = "chose_your_socket_address" //chose_your_socket_address
	)
```
For a better maintainability I recommand the use of a conf_file in the server package with a symlink on the main pbs.yaml.

server/server.go
```diff
@@ -87,13 +102,40 @@ func newMainServer(cfg *config.Configuration, handler http.Handler) *http.Server
...
+		Addr:         cfg.Socket,
+		Handler:      serverHandler,
+		ReadTimeout:  15 * time.Second,
+		WriteTimeout: 15 * time.Second,
+	}
+}
+
+func runServer(server *http.Server, name string, listener net.Listener) (err error) {
+	switch {
+	case server == nil:
```
VeronikaSolovei9, line 121: Would it be easier to use `if` instead?

server/server_test.go
```diff
@@ -77,7 +80,132 @@ func forwardSignal(t *testing.T, outbound chan<- struct{}, inbound <-chan os.Sig
...
+		}
+		ret.Close()
+	} else {
+		t.Errorf("[Test_newSocketServer] ret is Nil")
+	}
+}
+
+func Test_newMainServer(t *testing.T) {
+	const (
+		chose_your_socket_port    = 8000                        //chose_your_socket_port
```
VeronikaSolovei9, line 121: Can you please use different variable names? Please omit "chose_your_" part of it.

server/server_test.go
```diff
@@ -77,7 +80,132 @@ func forwardSignal(t *testing.T, outbound chan<- struct{}, inbound <-chan os.Sig
 	var s struct{}
 	sig := <-inbound
 	if sig != os.Interrupt {
-		t.Errorf("Unexpected signal: %s", sig.String())
+		t.Errorf("Unexpected signal: %s\n", sig.String())
 	}
 	outbound <- s
 }
+
+func Test_newSocketServer(t *testing.T) {
```
VeronikaSolovei9, line 88: Please follow general naming approach for tests: `TestNewSocketServer`

mediasquare-alexandre:
@VeronikaSolovei9 Done. All seem good to me. Waiting for your review.

server/server.go
```diff
@@ -32,26 +32,41 @@ func Listen(cfg *config.Configuration, handler http.Handler, adminHandler http.H
...
+		if err != nil {
+			glog.Errorf("Error listening for TCP connections on %s: %v for main server", mainServer.Addr, err)
+			return
+		}
+
+		go runServer(mainServer, "Main", mainListener)
 	}
-	adminListener, err := newListener(adminServer.Addr, nil)
+
+	adminListener, err := newTCPListener(adminServer.Addr, nil)
```
AlexBVolcy, line 59: Is there a valid way to improve the test coverage for this block of code? None of these lines are currently covered.

server/server.go
```diff
@@ -87,13 +102,39 @@ func newMainServer(cfg *config.Configuration, handler http.Handler) *http.Server
...
+		return
+	}
+
 	glog.Infof("%s server starting on: %s", name, server.Addr)
-	err := server.Serve(listener)
-	glog.Errorf("%s server quit with error: %v", name, err)
+	if err = server.Serve(listener); err != nil {
+		glog.Errorf("%s server quit with error: %v", name, err)
+	}
+	return
```
AlexBVolcy, line 134: Is it feasible to add code coverage here as well?

AlexBVolcy (commented):
Overall looks good! Just have a few comments related to test coverage.

server/server.go
SyntaxNode: There might not be a good way to cover this code with unit tests. The TCP code is equally not tested, for likely the same limitations. The author attested to successful local testing.

server/server.go
SyntaxNode: We don't have a good way to test `glog` errors. In the future we've discussed replacing `glog` with an intermediary logging object, but we're not there yet. It's a huge amount of refactoring.

config/config.go
```diff
@@ -17,13 +17,15 @@ import (
...
-	Port        int        `mapstructure:"port"`
-	Client      HTTPClient `mapstructure:"http_client"`
-	CacheClient HTTPClient `mapstructure:"http_client_cache"`
-	AdminPort   int        `mapstructure:"admin_port"`
-	EnableGzip  bool       `mapstructure:"enable_gzip"`
+	ExternalURL  string     `mapstructure:"external_url"`
+	Host         string     `mapstructure:"host"`
+	Port         int        `mapstructure:"port"`
+	EnableSocket bool       `mapstructure:"enable_socket"`
+	Socket       string     `mapstructure:"socket"`
```
SyntaxNode, line 24: Thank you for adding the enable flag. To be a bit nitpicky here, I think the name is ambiguous. Consider renaming to "UnixSocketEnabled" and UnixSocketName", or you can create a sub-object like:

```
UnixSocket  UnixSocketConfig       `mapstructure:"unix_socket"`

type UnixSocketConfig struct {
  Enabled   bool       `mapstructure:"enabled"`
  Name.      string.    `mapstructure:"name"`
}
```

mediasquare-alexandre:
@AlexBVolcy : It was feasible & now done but it required some changes on Listen.
`func Listen(cfg *config.Configuration, handler http.Handler, adminHandler http.Handler, metrics *metricsconfig.DetailedMetricsEngine)`
change to :
`func Listen(cfg *config.Configuration, handler http.Handler, adminHandler http.Handler, metrics *metricsconfig.DetailedMetricsEngine) (err error)`
Listen is now covered for ~33% and overall server.go for ~60%.

@SyntaxNode Yeah, it was nitpicky but at least now its done ;)

server/server_test.go
```diff
@@ -16,7 +21,7 @@ func TestNewAdminServer(t *testing.T) {
 	}
 	server := newAdminServer(cfg, http.HandlerFunc(handler))
 	if server.Addr != "prebid.com:6060" {
-		t.Errorf("Admin server address should be %s. Got %s", "prebid.com:6060", server.Addr)
+		t.Errorf("Admin server address should be %s. Got %s", "prebid.com:6060\n", server.Addr)
```
VeronikaSolovei9, line 24: Super nitpick: can you please remove new lines at the end of error logs? There are couple of them.

VeronikaSolovei9, line 24: Super nitpick: can you please remove new lines at the end of error logs? There are couple of them.

mediasquare-alexandre:
@VeronikaSolovei9 done. [62e84cf] commit reference: super_nitpicky

server/server_test.go
```diff
@@ -81,3 +86,169 @@ func forwardSignal(t *testing.T, outbound chan<- struct{}, inbound <-chan os.Sig
...
+			t.Errorf("[%s] WriteTimeout invalide: %v != %v",
+				func_name, ret.WriteTimeout, mock_server.WriteTimeout)
+		}
+		ret.Close()
+	} else {
+		t.Errorf("[%s] ret is Nil", func_name)
+	}
+}
+
+func TestNewMainServer_(t *testing.T) {
```
AlexBVolcy, line 125: Could you remove the underscore from this test name to be called `TestNewMainServer`

server/server_test.go
```diff
@@ -81,3 +86,169 @@ func forwardSignal(t *testing.T, outbound chan<- struct{}, inbound <-chan os.Sig
...
+	const (
+		func_name    = "TestNewMainServer_"
```
AlexBVolcy, line 127: Is there a reason why this name has an underscore at the end? Is it possible to remove this?

AlexBVolcy, line 127: Is there a reason why this name has an underscore at the end?

server/server_test.go
```diff
@@ -81,3 +86,169 @@ func forwardSignal(t *testing.T, outbound chan<- struct{}, inbound <-chan os.Sig
...
+	if ret, err := newUnixListener(mock_file, nil); err != nil {
+		t.Errorf("[%s] err_ : %s", func_name, err.Error())
+	} else {
+		ret.Close()
+	}
+}
+
+func TestNewAdminServer_(t *testing.T) {
+	const (
+		func_name  = "TestNewAdminServer_"
```
AlexBVolcy, line 188: Same comments as above in regards to the underscore

AlexBVolcy, line 188: Same comments as above in regards to the underscore

AlexBVolcy (commented):
A few more small comments about renaming

mediasquare-alexandre:
@AlexBVolcy : Functions with _ are the one where the code name 'TestXxxxXxxx' already exist.

There are different way of naming when names are already taken :
- 1 : add number at the end of the func name [ref]: https://cs.opensource.google/go/go/+/master:test/closedchan.go
- 2 : use '_' character in the function naming or as name [ref]: https://cs.opensource.google/go/go/+/master:test/goto.go

i.e : 'TestNewMainServer' is already taken.  Firstly I've put 'Test_newMainServer' ; which by the way was one of the only allowed usage of underscore for naming in go ; due to the fact it was a Test on a private package function name 'newMainServer'. Someone told me to do respect the following above... I can't respect this following if you refuse the underscore usage at the end or allowed me to use numbers. Review reference [8 days ago] April 19, 2022 @VeronikaSolovei9 

It is my 1st contribution to this project, is there somewhere a file where I can refered myself for the naming of type test function in case of duplicate naming ? Or can you give me a name for the following functions :
- Test_newMainServer
- Test_newTCPListener
- Test_runServer

Sincerely yours,

SyntaxNode:
Let's for a moment please put aside the issue of an underscore in the name. The purpose of the underscore here is to resolve a naming conflict, so let's focus the conversation on why there is a naming conflict.

There already exists a `TestNewMainServer` to ensure the function `newMainServer` returns a new server with the correctly formed address. This PR adds a new `TestNewMainServer` which also ensures the function `newMainServer` returns a new server with the correctly formed address and also verifies other fields such as the timeouts. 

**I don't see why we need to keep both tests. I'm fine with deleting the original `TestNewMainServer` and keeping the new more comprehensive `TestNewMainServer` version introduced in this PR.**

> It is my 1st contribution to this project

@mediasquare-alexandre Thank you for very much for your first contribution. :)

We try to follow [Go naming conventions](https://go.dev/doc/effective_go#names) whenever possible in our newer code and the convention is to use mixedCase instead of underscores. Hence our nitpicking on that. It doesn't really matter, but we try to keep consistent with new code and might eventually update older code. The area of code you are contributing to is very old (no need to change it in a while) and isn't a good example of the conventions we use today.

Another example of our more modern coding practices is the use of the Testify assert package. That automatically handles some of the features you've included in the tests, such as including the test name. If you'd like to use that library, the equivalent test would look something like:

```
func TestNewMainServer_(t *testing.T) {
	const (
		mockPort    = 8000
		mockAddress = "prebid.com"
	)
	cfg := new(config.Configuration)
	cfg.Port = mockPort
	cfg.Host = mockAddress

	ret := newMainServer(cfg, nil)

	if assert.NotNil(t, ret) {
		assert.Equal(t, "prebid.com:8000", ret.Addr, "addr invalid")
		assert.Equal(t, 15*time.Second, ret.ReadTimeout, "read timeout invalid")
		assert.Equal(t, 15*time.Second, ret.WriteTimeout, "write timeout invalid")
		ret.Close()
	}
}
```

or something like

```
func TestNewMainServer_(t *testing.T) {
	const (
		mockPort    = 8000
		mockAddress = "prebid.com"
	)
	cfg := new(config.Configuration)
	cfg.Port = mockPort
	cfg.Host = mockAddress

	expectedServer := &http.Server{
		Addr:         "prebid.com:8000",
		Handler:      nil,
		ReadTimeout:  15 * time.Second,
		WriteTimeout: 15 * time.Second,
	}

	ret := newMainServer(cfg, nil)
	assert.Equal(t, expectedServer, ret)
}
``

AlexBVolcy (approved):
LGTM

AlexBVolcy (changes requested):
The code looks good, but tests in these test functions are failing:

    TestNewTCPListener
    TestNewUnixListener

Could you please resolve these

mediasquare-alexandre:
@AlexBVolcy Sure it was my bad. All tests I had push were working locally because of some already running servers on the same port. My environment wasn't clean during my last bunch of tests but normally it's ok now. I'm looking for now to see if all tests pass this time. It does locally with a 60.3% coverage rate.

AlexBVolcy (approved):
All looks good! This isn't urgent, but if you could rebase your changes on top of master, it'll get rid of that security failure which would be nice.

VeronikaSolovei9 (approved):

mediasquare-alexandre:
@AlexBVolcy @VeronikaSolovei9 😄 

@VeronikaSolovei9 by the way I did  some test about concurrent requests as you asked. It all depend on your server capacity and the number of maximum socket you can set with command ' ulimit -n '. You can test it with command ' ab ', it's a component of apache2-utils.

PR #2220, opened 2022-04-13
AMX Adapter: support native format

nickjacob:
This PR adds support for native ads & removes unnecessary legacy code that was used for interpolating pixels into ad markup:

- add support for native requests, checking bid type based on bid.ext
- update AMX adapter internal version to 1.2
- add test of bid.ext / bid type logic in MakeBids

### DIFF
--- a/adapters/amx/amx_test.go
+++ b/adapters/amx/amx_test.go
@@ -3,7 +3,6 @@ package amx
 import (
 	"encoding/json"
 	"fmt"
-	"regexp"
 	"testing"
 
 	"github.com/mxmCherry/openrtb/v15/openrtb2"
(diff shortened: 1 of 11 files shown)


VeronikaSolovei9:
These changes looks good to me. However while I was reviewing and testing this PR I noticed you didn't specify required parameters for `imp.ext.amx`. In amx.json file there are 2 fields `imp.ext.amx` can have: tagId and adUnitId. This corresponds to `ExtImpAMX`.
Without specifying required fields I can run request with empty ` imp.ext.amx` and it doesn't throw error, which I think is not a valid bidder config. I may be wrong, I don't know how your system works inside. If empty imp.ext.amx is a valid case, please ignore this comment. If not - add required fields. Please refer to PBS documentation and examples: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#bidder-parameters

nickjacob:
Thanks Veronika!

We do support empty ext.amx. We allow passing the tagId parameter via a custom endpoint that we provide publishers, and adUnitId is not required--that's why we don't require or test for imp.ext.amx.

adapters/amx/amx_test.go
```diff
@@ -174,42 +173,66 @@ func TestMakeRequestsPublisherId(t *testing.T) {
...
+			assert.Len(t, errs, 1)
+			continue
+		}
+
+		if len(errs) > 0 {
+			t.Fatalf("Failed to make bids: %v", errs)
+		}
+
+		assert.Len(t, bids.Bids, 1)
+		assert.Equal(t, bids.Bids[0].BidType, test.bidType)
```
mansinahar, line 235: The order for this needs to be the other way round like:
```
assert.Equal(t, test.bidType, bids.Bids[0].BidType)
```

The assert function signature requires the expected value to be specified first.
Reference: https://pkg.go.dev/github.com/stretchr/testify/assert#Equal

mansinahar:
LGTM except for the above minor comment. Also, please update the docs here: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/amx.md to include `native` as a supported media type

nickjacob:
Thanks @mansinahar -- I made the change to the test & opened this PR: https://github.com/prebid/prebid.github.io/pull/3714

VeronikaSolovei9 (approved):
Thank you for your answers. LGTM

mansinahar (approved):

mansinahar:
> Thanks @mansinahar -- I made the change to the test & opened this PR: [prebid/prebid.github.io#3714](https://github.com/prebid/prebid.github.io/pull/3714)

Thanks @nickjacob. Looks good to me

PR #2232, opened 2022-04-22
sovrn: Add video support

cpabst:


adapters/sovrn/sovrn.go
```diff
@@ -75,71 +100,54 @@ func addHeaderIfNonEmpty(headers http.Header, headerName string, headerValue str
...
-
-	if imp.BidFloor == 0 && sovrnExt.BidFloor > 0 {
-		imp.BidFloor = sovrnExt.BidFloor
-	}
-
-	return imp.TagID, nil
+	return response, nil
 }
 
 func getTagid(sovrnExt openrtb_ext.ExtImpSovrn) string {
```
VeronikaSolovei9, line 153: I'm curious what is the purpose of this function? I know this is not in scope of this PR, but do you think you need it?

.devcontainer/devcontainer.json
```diff
@@ -10,6 +10,7 @@
 			// Options
 			"INSTALL_NODE": "false",
 			"NODE_VERSION": "lts/*",
+			"TEST": "false"
```
VeronikaSolovei9, line 13: Please remove this line, you probably pushed it accidentally.

adapters/sovrn/sovrn.go
```diff
@@ -61,6 +39,53 @@ func (s *SovrnAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+		if imp.BidFloor == 0 && sovrnExt.BidFloor > 0 {
+			imp.BidFloor = sovrnExt.BidFloor
+		}
+
+		// Validate video params if appropriate
+		video := imp.Video
+		if video != nil {
+			if video.MIMEs == nil ||
+				video.MinDuration == 0 ||
+				video.MaxDuration == 0 ||
```
VeronikaSolovei9, line 73: Please review this logic again. For video endpoint there is a possibility to have only maxDuration when `requireExactDuration` if false or not specified. In this case the entire request will be rerjected.

adapters/sovrn/sovrn.go
```diff
@@ -75,71 +100,54 @@ func addHeaderIfNonEmpty(headers http.Header, headerName string, headerValue str
...
-func preprocess(imp *openrtb2.Imp) (string, error) {
-	var bidderExt adapters.ExtImpBidder
-	if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
-		return "", &errortypes.BadInput{
-			Message: err.Error(),
-		}
-	}
+				if request.Imp[0].Video != nil {
+					bidType = openrtb_ext.BidTypeVideo
+				}
```
VeronikaSolovei9, line 140: In case you return bid type in bid.ext we recommend to use this approach:
https://github.com/prebid/prebid-server/blob/c93a0369549f5c087cc41553a31a4bbfe72ddd23/adapters/algorix/algorix.go#L190
It will also help to determine bid type correctly in case if req.imp is multi typed.

adapters/sovrn/sovrn.go
```diff
@@ -61,6 +39,53 @@ func (s *SovrnAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+			errs = append(errs, &errortypes.BadInput{
+				Message: err.Error(),
+			})
+			return nil, errs
+		}
+
+		imp.TagID = getTagid(sovrnExt)
+
+		if imp.BidFloor == 0 && sovrnExt.BidFloor > 0 {
+			imp.BidFloor = sovrnExt.BidFloor
```
VeronikaSolovei9, line 65: Can you please add coverage for this branch?

adapters/sovrn/sovrn.go
cpabst: I have inherited this code from someone who is no longer at Sovrn. But it appears that our original work with Prebid Server used a spelling on our tagid parameter that we quickly decided was wrong. 😄 

So the old version is deprecated, but not removed yet. Not sure how we will go forward with the actual removal, but we will figure that out.

adapters/sovrn/sovrn.go
cpabst: I'll need to review this with product. I'll let you know how that conversation goes.

.devcontainer/devcontainer.json
cpabst: Fixed.

adapters/sovrn/sovrntest/video/multi-banner-video.json
```diff
@@ -0,0 +1,185 @@
...
+  "mockBidRequest": {
+    "id": "test-request-id",
+    "imp": [
+      {
+        "id": "test-imp-id1",
+        "banner": {
+          "format": [
+            {
+              "w": 300,
+              "h": 250
```
mansinahar, line 11: Consider the imp order reversed in this test case where the first imp is a video imp and the second one is the banner imp, in that case the bid type in the response will be incorrectly set to `video`.. Now also consider the imps in the same order as you have in this test, if your bidder service were to send back a bid for the video imp as well, in this case the video bid would have the bid type incorrectly set to `banner`

adapters/sovrn/sovrn.go
```diff
@@ -61,6 +39,53 @@ func (s *SovrnAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+		video := imp.Video
+		if video != nil {
+			if video.MIMEs == nil ||
+				video.MinDuration == 0 ||
+				video.MaxDuration == 0 ||
+				video.Protocols == nil {
+				errs = append(errs, &errortypes.BadInput{
+					Message: "Missing required video parameter",
+				})
+				return nil, errs
```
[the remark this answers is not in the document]
cpabst: Good point. Thanks.

adapters/sovrn/sovrntest/video/multi-banner-video.json
cpabst: I believe this has been solved by using the response `Bid.impid` value.

adapters/sovrn/sovrntest/supplemental/with-only-custom-bidfloor.json
```diff
@@ -0,0 +1,127 @@
...
+              "w": 300,
+              "h": 250
+            },
+            {
+              "w": 300,
+              "h": 600
+            }
+          ]
+        },
+        "bidfloor": 4.2,
```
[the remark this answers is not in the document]
cpabst: Yes, you are correct. This has been fixed.

adapters/sovrn/sovrntest/exemplary/no-bid.json
```diff
@@ -20,7 +20,8 @@
         "bidder": {
           "tagid": "123456"
         }
-      }
+      },
```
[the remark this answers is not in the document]
cpabst: Yep. Figured that out today. Those all should be fixed now.

adapters/sovrn/sovrn.go
```diff
@@ -75,71 +100,54 @@ func addHeaderIfNonEmpty(headers http.Header, headerName string, headerValue str
...
 
-func preprocess(imp *openrtb2.Imp) (string, error) {
-	var bidderExt adapters.ExtImpBidder
-	if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
-		return "", &errortypes.BadInput{
-			Message: err.Error(),
-		}
-	}
+				if request.Imp[0].Video != nil {
+					bidType = openrtb_ext.BidTypeVideo
```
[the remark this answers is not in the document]
cpabst: I'm solving this presently with using the Bid.impid value to set bid type. We will look into using the recommended value in the future.

adapters/sovrn/sovrn.go
cpabst: Done. Fixed a bunch of other tests as well.

cpabst:
@mansinahar @VeronikaSolovei9 I have made quite a few updates based on these comments. This PR is ready for another review.

adapters/sovrn/sovrn.go
```diff
@@ -61,6 +39,69 @@ func (s *SovrnAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+
+		tagId := getTagid(sovrnExt)
+		if tagId == "" {
+			errs = append(errs, &errortypes.BadInput{
+				Message: "Missing required parameter 'tagid'",
+			})
+			continue
+		}
+
+		imp.TagID = getTagid(sovrnExt)
```
mansinahar, line 71: You've already gotten the `tagId` value from the `getTagid` function on L63 so this line should be just:
```
imp.TagID = tagId
```

Also, can you please rename the `getTagid` function to `getTagId`?

adapters/sovrn/sovrn.go
```diff
@@ -61,6 +39,69 @@ func (s *SovrnAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapt
...
+
+		validImps++
+		request.Imp[index] = imp
+	}
+
+	if validImps == 0 {
+		return nil, errs
+	}
+
+	reqJSON, err := json.Marshal(request)
```
mansinahar, line 99: At this point, this request can have both valid and invalid imps because on L92 you're just reassigning the valid imps to the same place in the imps array of the request. You'd actually have to remove the invalid ones as you find them in the for loop so that you're only left with the valid ones or a much easier way would be to just create a new `validImps` array and keep appending each valid imp you find to that array and then outside the for loop, you can just replace `request.Imp` with that new `validImps` array before JSON marshalling

adapters/sovrn/sovrn.go
```diff
@@ -75,71 +116,54 @@ func addHeaderIfNonEmpty(headers http.Header, headerName string, headerValue str
 	}
 }
 
-func (s *SovrnAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
-	if response.StatusCode == http.StatusNoContent {
+func (s *SovrnAdapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.RequestData, responseData *adapters.ResponseData) (*adapters.BidderResponse, []error) {
```
mansinahar, line 119: **Nitpick:** The names `requestData` and `responseData ` don't really convey the proper context here IMHO. How about using, `bidderRequest` and `bidderResponse`, respectively?

adapters/sovrn/sovrn.go
```diff
@@ -150,6 +174,16 @@ func getTagid(sovrnExt openrtb_ext.ExtImpSovrn) string {
...
 }
 
+func getImpIdx(impId string, request *openrtb2.BidRequest) int {
+	for idx, imp := range request.Imp {
+		if imp.ID == impId {
+			return idx
+		}
+	}
+
+	return -1 // hopefully this never happens
```
mansinahar, line 184: We should be returning an error if this does happen.. This would mean that for some reason your bidder bid for an imp that doesn't match an ID with the imps in the request. By returning an explicit error, it will be much easier to debug and fix the issue rather than silently just setting the bid type to `banner`

adapters/sovrn/sovrn.go
cpabst: Both done

adapters/sovrn/sovrn.go
cpabst: Good point. Done.

adapters/sovrn/sovrn.go
cpabst: Sure. 😄

adapters/sovrn/sovrn.go
cpabst: We do not support multi-typed imps yet.

adapters/sovrn/sovrn.go
cpabst: I see your point. I believe I have fixed this correctly now. Please let me know if you feel otherwise.

adapters/sovrn/sovrn.go
mansinahar: @cpabst Were you able to check about this with your team?

adapters/sovrn/sovrntest/video/multi-banner-video.json
```diff
@@ -0,0 +1,184 @@
...
+          },
+          "device": {
+            "ua": "test-user-agent",
+            "ip": "123.123.123.123",
+            "dnt": 0,
+            "language": "en"
+          }
+        }
+      },
+      "mockResponse": {
```
mansinahar, line 141: Please consider adding a response for video as well in this test case. This will allow this test case to verify that responses with multiple bids for various imps are processed correctly and will also verify the getBidType logic to make sure it matches bids to imps properly based on the imp ID

adapters/sovrn/sovrn.go
```diff
@@ -75,78 +117,77 @@ func addHeaderIfNonEmpty(headers http.Header, headerName string, headerValue str
...
 			adm, err := url.QueryUnescape(bid.AdM)
 			if err == nil {
 				bid.AdM = adm
-				bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
+
+				bidType := openrtb_ext.BidTypeBanner
+
+				impIdx, impIdErr := getImpIdx(bid.ImpID, request)
+				if impIdErr != nil {
+					errs = append(errs, impIdErr)
```
mansinahar, line 157: Please add a test case for this error condition

mansinahar:
@cpabst Thanks for addressing the comments. This is looking to be in good shape now. Just have a couple minor comments.

adapters/sovrn/sovrn.go
```diff
@@ -163,25 +168,27 @@ func (s *SovrnAdapter) MakeBids(request *openrtb2.BidRequest, requestData *adapt
...
 	for idx, imp := range request.Imp {
 		if imp.ID == impId {
-			return idx
+			return idx, nil
 		}
 	}
 
-	return -1 // hopefully this never happens
+	return -1, &errortypes.BadInput{
+		Message: fmt.Sprintf("Could not find imp ID '%s' in bid, dropping bid", impId),
```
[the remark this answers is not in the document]
cpabst: Works for me. Done.

adapters/sovrn/sovrntest/video/simple-video.json
```diff
@@ -0,0 +1,153 @@
+{
+  "mockBidRequest": {
+  "id": "test-request-id",
+  "imp": [
```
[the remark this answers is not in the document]
cpabst: This one just has the required parameters, and the other one has all parameters. I like that it shows the minimum viable request, but don't feel strongly if you'd like it to go.

adapters/sovrn/sovrntest/video/multi-banner-video.json
cpabst: Good point. Done.

adapters/sovrn/sovrn.go
cpabst: Done.

adapters/sovrn/sovrntest/video/multi-banner-video.json
cpabst: Good point. Done.

adapters/sovrn/sovrn.go
cpabst: He has been very busy. 😄 
I'd like to keep it this way, and look at this idea as a future enhancement. I see others doing it the same way.

adapters/sovrn/sovrn.go
VeronikaSolovei9: You may not support it, but request may have it. In your code you take imp type from request and match it with response: `request.Imp[impIdx]`

VeronikaSolovei9: You may not support it, but request may have it. In your code you take imp type from request and match it with response: `request.Imp[impIdx]`

mansinahar (approved):

VeronikaSolovei9 (approved):

PR #2244, opened 2022-05-12
GDPR: Refactor AuctionActivitiesAllowed to return struct

bsardo:
This is the first of a several refactor PRs that will eventually lead to the incorporation of the basic enforcement algorithm that can be used instead of the full algorithm for specific purposes. No additional logic has been introduced here. This simply alters the `AuctionActivitiesAllowed` GDPR function signature to return a permissions struct instead of three bool values.

### DIFF
--- a/endpoints/cookie_sync_test.go
+++ b/endpoints/cookie_sync_test.go
@@ -1652,9 +1652,13 @@ func (m *MockGDPRPerms) BidderSyncAllowed(ctx context.Context, bidder openrtb_ex
 	return args.Bool(0), args.Error(1)
 }
 
-func (m *MockGDPRPerms) AuctionActivitiesAllowed(ctx context.Context, bidderCoreName openrtb_ext.BidderName, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal gdpr.Signal, consent string, aliasGVLIDs map[string]uint16) (allowBidReq bool, passGeo bool, passID bool, err error) {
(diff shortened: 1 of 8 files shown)


endpoints/cookie_sync_test.go
```diff
@@ -1652,9 +1652,13 @@ func (m *MockGDPRPerms) BidderSyncAllowed(ctx context.Context, bidder openrtb_ex
...
 
-func (m *MockGDPRPerms) AuctionActivitiesAllowed(ctx context.Context, bidderCoreName openrtb_ext.BidderName, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal gdpr.Signal, consent string, aliasGVLIDs map[string]uint16) (allowBidReq bool, passGeo bool, passID bool, err error) {
+func (m *MockGDPRPerms) AuctionActivitiesAllowed(ctx context.Context, bidderCoreName openrtb_ext.BidderName, bidder openrtb_ext.BidderName, PublisherID string, gdprSignal gdpr.Signal, consent string, aliasGVLIDs map[string]uint16) (permissions gdpr.AuctionPermissions, err error) {
 	args := m.Called(ctx, bidderCoreName, bidder, PublisherID, gdprSignal, consent, aliasGVLIDs)
-	return args.Bool(0), args.Bool(1), args.Bool(2), args.Error(3)
+	return gdpr.AuctionPermissions{
+		AllowBidRequest: args.Bool(0),
+		PassGeo:         args.Bool(1),
+		PassID:          args.Bool(2),
+	}, args.Error(3)
```
SyntaxNode, line 1661: This doesn't look right. Shouldn't the entire `AuctionPermissions` struct be the first response? There are 2 returns, not 4.

bsardo: Yep that can't be right. I'll fix.

Issue #1474, opened 2020-09-03
Debug Metrics


SyntaxNode:
### Motivation
This proposal addresses a discussion in https://github.com/prebid/prebid-server/issues/745 to provide hosts with visibility in how many auctions are running with debugging enabled.

### Feature Request
Add new general metrics and new account metrics to track the number of requests received with debugging enabled through either the test or debug flag.

1. Add a new counter metric `requests_debug` which increments whenever a new request with debugging enabled is received. May be compared with the existing `requests` counter metric to determine percentages.

2. Add a new account metric `account_requests_debug` which increments whenever a new request with debugging enabled for a specific account is received. May be compared with the existing `account_requests` counter metric to determine percentages. Since account cardinality may be very high, this should be controlled with an option.

3. Add a new `metrics.disabled.account_debug` config option to control enabling the account metric. Due to potentially very high account cardinality, this option will default to `true`, matching the default for `metrics.disabled.account_adapter_details`.

SyntaxNode:
Yes, this was completed for PBS-Go in https://github.com/prebid/prebid-server/pull/2246.

bretg:
Confirmed this is not yet ported to PBS-Java

bsardo:
@osulzhenko please classify priority.

PR #2246, opened 2022-05-13
Debug Metrics Implementation

AlexBVolcy:
This PR implements the feature/issue https://github.com/prebid/prebid-server/issues/1474

The goal is to add new metrics to PBS related to debug

The first new debug metric, is one that counts the number of requests we receive that has debug enabled. 

The second new debug metric, is one that counts the number of requests from an account that has debug enabled. 

This second metric is controlled by a config flag  `"metrics.disabled_metrics.account_debug"` because of concerns of high cardinality 

These metrics are implemented for both `prometheus` and `influx`, by creating/implementing a new record function `RecordDebugRequest()` within both metrics' frameworks.

This function is then called within `HoldAuction()` after we receive debug information from the `getDebugInfo()` call. 

For influx, I added two new meters to represent the new metrics in the `Metrics` and `AccountMetrics` struct, `DebugRequestMeter` and `accountMetrics.DebugRequestMeter`

For prometheus, I added two new metrics to the prometheus specific Metrics struct `debugRequests` and `accountDebugRequests`

New tests were included to test these updated metrics

Other metrics engines like nil metrics, multi metrics, and metrics mock, only need to be updated with a simple handling of how they implement/call `RecordDebugRequest()`


endpoints/openrtb2/auction_test.go
```diff
@@ -3702,7 +3702,7 @@ type warningsCheckExchange struct {
 	auctionRequest exchange.AuctionRequest
 }
 
-func (e *warningsCheckExchange) HoldAuction(ctx context.Context, r exchange.AuctionRequest, debugLog *exchange.DebugLog) (*openrtb2.BidResponse, error) {
+func (e *warningsCheckExchange) HoldAuction(ctx context.Context, r exchange.AuctionRequest, debugLog *exchange.DebugLog, labels *metrics.Labels) (*openrtb2.BidResponse, error) {
```
SyntaxNode, line 3705: It is not necessary to add a new `labels` argument. The labels are already present in the `r exchange.AuctionRequest` argument.

metrics/metrics.go
```diff
@@ -8,11 +8,13 @@ import (
...
-	PubID         string // exchange specific ID, so we cannot compile in values
-	CookieFlag    CookieFlag
-	RequestStatus RequestStatus
+	Source           DemandSource
+	RType            RequestType
+	PubID            string // exchange specific ID, so we cannot compile in values
+	CookieFlag       CookieFlag
+	RequestStatus    RequestStatus
+	DebugFlag        bool
+	AccountDebugFlag bool
```
SyntaxNode, line 17: I think you'll be fine with a single debug flag, perhaps naming `DebugEnabed` or `DebugRequested` or perhaps even `Debug`. We don't care _how_ debug is enabled, just that it is _is_ enabled. With the implementation of the debug permissions, I can easily see how you arrived at this though. :)

metrics/prometheus/prometheus.go
```diff
@@ -182,6 +184,11 @@ func NewMetrics(cfg config.PrometheusMetrics, disabledMetrics config.DisabledMet
 		"Count of total requests to Prebid Server labeled by type and status.",
 		[]string{requestTypeLabel, requestStatusLabel})
 
+	metrics.requestsDebug = newCounter(cfg, reg,
+		"requests_debug",
+		"Count of total requests to Prebid Server that have debug enabled labled by type, status.",
+		[]string{requestTypeLabel, requestStatusLabel})
```
SyntaxNode, line 190: I do not believe we need to have any labels on this metric. I personally don't see any value in the status label and only minimal value in the types label. I won't put up a fight it you want to keep the type label, but I'll need convincing on why this is important for the status label.

metrics/metrics.go
AlexBVolcy: Ah this makes sense! I'll update accordingly.

metrics/prometheus/prometheus.go
AlexBVolcy: Yeah I struggled to think of reasons to have the labels, but I initially just matched the format of `metrics.requests`. But I feel comfortable with removing both

endpoints/openrtb2/auction_test.go
AlexBVolcy: Refactored and removed in latest commit

endpoints/openrtb2/amp_auction.go
```diff
@@ -219,6 +220,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 		ao.Errors = append(ao.Errors, err)
 		return
 	}
+	labels.DebugEnabled = debugLog.DebugFlag
```
SyntaxNode, line 223: What is the purpose of this line? ... since it occurs after the `HoldAuction` call. Wouldn't the metrics for an incoming request have been recorded before hitting this line? Same applies to `openrtb/auction.go`

SyntaxNode, line 223: I consider it a "code smell" that `debugLog` is instantiated here just to return the DebugFlag to the outer scope. It would be preferable IMHO to either calculate the debug flag prior to calling `HoldAuction` or include it as a return argument somehow. I know you didn't create this situation, but I believe it's worth an investigation to try and make this flow a bit nicer now.

SyntaxNode, line 223: I consider it a "code smell" that `debugLog` is instantiated here just to return the DebugFlag to the outer scope. It would be preferable IMHO to either calculate the debug flag prior to calling `HoldAuction` or include it as a return argument somehow. I know you didn't create this situation, but I believe it's worth an investigation to try and make this flow a bit nicer now.

SyntaxNode (commented):
Looks good. Let's try to improve the flow of the debug flag, or put it off for a future update if we think of a good idea and decide it's too much work to include here.

endpoints/openrtb2/amp_auction.go
VeronikaSolovei9: If I understand this correctly, metrics will be written at the end of endpoint execution. It's defined in defer section, L122:
```
defer func() {
		deps.metricsEngine.RecordRequest(labels)
		deps.metricsEngine.RecordRequestTime(labels, time.Since(start))
		deps.analytics.LogAmpObject(&ao)
	}()
```
Auction, video and amp endpoints have this approach.

VeronikaSolovei9: Would it be better if labels.DebugEnabled will be passed in `auctionRequest` downstrem to `getDebugInfo` and modified there? 
I cannot think about better solution...

VeronikaSolovei9 (commented):
Changes look good to me. Please resolve Scott's comments.

endpoints/openrtb2/amp_auction.go
mansinahar: I agree with Scott here. This definitely seems odd to instantiate the `debugLog` struct here just to return the `DebugFlag` to the outer scope. How about making `debugLog` a part of the `AuctionRequest` struct? That way you can read and modify it in `HoldAuction` as needed and you can again access it outside of `HoldAuction` for setting the metrics label and such

AlexBVolcy: Yeah exactly like Veronika said above! And when I sent test postman requests, it acted like this as well!

SyntaxNode: Ah. Got it. This code is a bit too clever IMHO, but I agree it's beyond the scope of this PR to fix. Can you add a comment before these lines to make it clear what's going on to my future self?

SyntaxNode: > That way you can read and modify it in HoldAuction as needed and you can again access it outside of HoldAuction for setting the metrics label and such

A step in the right direction, but modifying it as a way to pass information upstream is still an undesirable design. We might have to tackle that later on.

mansinahar:
> A step in the right direction, but modifying it as a way to pass information upstream is still an undesirable design. We might have to tackle that later on.

There's one other thing I think we should do here that is perhaps create a new function for recording debug enabled requests rather than using the current record request method. Then you can just call that record method in `HoldAuction` itself rather than calling it in each of the endpoint handlers. The debug enabled metrics anyway don't care about the source of the request. And this way we wouldn't be modifying it just to be able to pass the information back upstream. Also, debug info is per request so I think it makes sense to add it as part of the `AuctionRequest` struct.

AlexBVolcy:
The latest commit I pushed is the one where I implemented a function `RecordDebugRequest` to handle the recording of the new metrics. This function gets called now in `HoldAuction` so that way we wouldn't have to be concerned with passing the debug information found in `HoldAuction` upstream, and instead can just pass the debug enabled info directly to `RecordDebugRequest`.

I wrongfully titled the name of this commit to be `Debug Log Refactoring`, since the `DebugLog` object is still a separate parameter in `HoldAuction`, and hasn't been combined with the `AuctionRequest` object yet. I discussed this with Mansi, and this will be handled in a future PR.

VeronikaSolovei9 (approved):
LGTM!

endpoints/openrtb2/amp_auction.go
```diff
@@ -206,6 +206,7 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 		GDPRPermissionsBuilder:     aggregatedGDPR.NewPermissions,
 		StoredAuctionResponses:     storedAuctionResponses,
 		StoredBidResponses:         storedBidResponses,
+		PubID:                      labels.PubID,
```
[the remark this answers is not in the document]
AlexBVolcy: Yes thank you! Will update

metrics/go_metrics.go
```diff
@@ -430,6 +434,17 @@ func (me *Metrics) RecordRequest(labels Labels) {
 	am.requestMeter.Mark(1)
 }
 
+func (me *Metrics) RecordDebugRequest(debugEnabled bool, pubID string) {
+	if debugEnabled {
+		me.DebugRequestMeter.Mark(1)
+	}
+
+	am := me.getAccountMetrics(pubID)
+	if !me.MetricsDisabled.AccountDebug && debugEnabled {
```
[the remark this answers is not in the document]
AlexBVolcy: Yes I'll refactor this and add the unknown check as well.

VeronikaSolovei9 (dismissed):

metrics/prometheus/prometheus_test.go
```diff
@@ -156,6 +156,40 @@ func TestRequestMetric(t *testing.T) {
...
+	testCases := []struct {
+		description                      string
+		givenDebugEnabledFlag            bool
+		givenAccountDebugMetricsDisabled bool
+		expectedAccountDebugCount        float64
+		expectedDebugCount               float64
+	}{
+		{
+			description:                      "Boolean flags should result in both metrics being updated",
+			givenDebugEnabledFlag:            true,
```
mansinahar, line 169: Please add a test case where debug is enabled and account debug is disabled and another test case where debug is disabled and account debug is enabled

metrics/go_metrics.go
```diff
@@ -94,6 +95,7 @@ type MarkupDeliveryMetrics struct {
 
 type accountMetrics struct {
 	requestMeter      metrics.Meter
+	debugRequestMeter metrics.Meter
```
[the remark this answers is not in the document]
AlexBVolcy: Good catch: I added a the `DebugRequestMeter` to the struct, but I won't need to add the `accountMetrics.debugRequestMeter`, as the `NewBlankMetrics` just instantiates a blank `accountMetrics` map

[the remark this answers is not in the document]
mansinahar: Oh yeah, I had meant to add this comment only for the generic debugRequestMeter but mistakenly pointed it to the account specific one :D

metrics/prometheus/prometheus.go
```diff
@@ -27,6 +27,7 @@ type Metrics struct {
 	impressionsLegacy            prometheus.Counter
 	prebidCacheWriteTimer        *prometheus.HistogramVec
 	requests                     *prometheus.CounterVec
+	debugRequests                prometheus.Counter
```
mansinahar, line 30: Hmm.. why are the types for `debugRequests` and `accountDebugRequests` different? One is a `Counter` and other is a `CounterVec`

metrics/prometheus/prometheus_test.go
```diff
@@ -156,6 +156,54 @@ func TestRequestMetric(t *testing.T) {
...
+func TestDebugRequestMetric(t *testing.T) {
+	testCases := []struct {
+		description                      string
+		givenDebugEnabledFlag            bool
+		givenAccountDebugMetricsDisabled bool
+		expectedAccountDebugCount        float64
+		expectedDebugCount               float64
+	}{
+		{
+			description:                      "Boolean flags should result in both metrics being updated",
```
mansinahar, line 168: Please specify what the boolean flags are set to in the description. For readability purposes and for when tests fail these descriptions help us easily understand what was passed in and what was expected. I like the descriptions you have for the last two test cases. Could you please update these to have the same format as well? 

I noticed the same thing in the `go_metrics_test.go` file as well

config/config_test.go
```diff
@@ -611,6 +612,7 @@ func TestFullConfig(t *testing.T) {
 	cmpBools(t, "account_required", cfg.AccountRequired, true)
 	cmpBools(t, "auto_gen_source_tid", cfg.AutoGenSourceTID, false)
 	cmpBools(t, "account_adapter_details", cfg.Metrics.Disabled.AccountAdapterDetails, true)
+	cmpBools(t, "account_debug", cfg.Metrics.Disabled.AccountDebug, true)
```
mansinahar, line 615: For the `TestFullConfig` I would suggest setting the value for `account_debug` to be `false` (different from the default value) in the above `fullConfig` value and assert it here to be `false` as well. This just helps us test through both the flow with default values and non-default values.

mansinahar (commented):
I think this is looking pretty close. Thanks for addressing all the feedback. I have one major comment about the different types being used for account vs non-account metric in Prometheus. The other two are quite minor

metrics/prometheus/prometheus.go
mansinahar: Discussed offline. In Prometheus account metrics have an `account` label and so they need to be `CounterVec` than a plain `Counter`

VeronikaSolovei9 (approved):
LGTM!

mansinahar (approved):

PR #2248, opened 2022-05-18
Refactor: Move cookie sync account fetching up a level

bsardo:
Cookie sync endpoint account fetching is moved up a level from the `setLimit` function into the `parseRequest` function so that it can be used for other purposes. An upcoming PR will make use of the fetched account to create a request-specific GDPR permissions object.

### DIFF
--- a/endpoints/cookie_sync.go
+++ b/endpoints/cookie_sync.go
@@ -113,6 +113,15 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
 		return usersync.Request{}, privacy.Policies{}, fmt.Errorf("JSON parsing failed: %s", err.Error())
 	}
 
+	var accountInfo *config.Account
+	var fetchErrs []error
+	if request.Account != "" {
+		accountInfo, fetchErrs = accountService.GetAccount(context.Background(), c.config, c.accountsFetcher, request.Account)
+	}
+	if len(fetchErrs) > 0 {
+		return usersync.Request{}, privacy.Policies{}, combineErrors(fetchErrs)
+	}
+
 	var gdprString string
 	if request.GDPR != nil {
 		gdprString = strconv.Itoa(*request.GDPR)
@@ -132,10 +141,7 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
 		}
 	}
 
-	request, err = c.setLimit(request)
-	if err != nil {
-		return usersync.Request{}, privacy.Policies{}, err
-	}
+	request = c.setLimit(request, accountInfo)
 
 	privacyPolicies := privacy.Policies{
 		GDPR: gdprPrivacy.Policy{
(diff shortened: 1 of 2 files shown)


endpoints/cookie_sync.go
```diff
@@ -192,26 +198,22 @@ func (c *cookieSyncEndpoint) writeParseRequestErrorMetrics(err error) {
 	}
 }
 
-func (c *cookieSyncEndpoint) setLimit(request cookieSyncRequest) (cookieSyncRequest, error) {
-	if request.Account != "" {
-		accountInfo, errs := accountService.GetAccount(context.Background(), c.config, c.accountsFetcher, request.Account)
-		if len(errs) > 0 {
-			return request, combineErrors(errs)
-		}
+func (c *cookieSyncEndpoint) setLimit(request cookieSyncRequest, account *config.Account) cookieSyncRequest {
```
SyntaxNode, line 201: Consider passing in the `account.CookieSync` as an argument instead of the entire `account`.

endpoints/cookie_sync_test.go
```diff
@@ -792,117 +792,104 @@ func TestSetLimit(t *testing.T) {
 	testCases := []struct {
 		description     string
 		givenRequest    cookieSyncRequest
-		givenConfig     config.UserSync
-		expectedError   string
+		giveAccount     *config.Account
```
SyntaxNode, line 795: Typo: `givenAccount` vs `giveAccount`.

endpoints/cookie_sync.go
```diff
@@ -113,6 +113,15 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
...
 	}
 
+	var accountInfo *config.Account
+	var fetchErrs []error
+	if request.Account != "" {
+		accountInfo, fetchErrs = accountService.GetAccount(context.Background(), c.config, c.accountsFetcher, request.Account)
+	}
+	if len(fetchErrs) > 0 {
+		return usersync.Request{}, privacy.Policies{}, combineErrors(fetchErrs)
+	}
```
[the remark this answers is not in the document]
bsardo: 👍 Yep will do. I'll probably use an `if` with assignment since I'm looking to always fetch an account based on our offline team discussion.

AlexBVolcy (dismissed):
LGTM

endpoints/cookie_sync_test.go
```diff
@@ -787,122 +790,221 @@ func TestCookieSyncParseRequest(t *testing.T) {
...
-				Account: "TestAccount",
-				Limit:   42,
+				Limit: 0,
 			},
-			givenConfig: config.UserSync{
-				Cooperative: config.UserSyncCooperative{
-					PriorityGroups: [][]string{{"a", "b", "c"}},
+			givenAccount: &config.Account{
+				CookieSync: config.CookieSync{
+					DefaultLimit: &[]int{20}[0],
```
SyntaxNode, line 806: IMHO: This is a really confusing way to convert an int to a pointer. The array is disorienting.

You can create a variable within the function scope and reference it here, like `DefaultLimit: &int20` or write a simple test helper method similar to what's in the mxmcherry library.

endpoints/cookie_sync.go
```diff
@@ -132,10 +138,8 @@ func (c *cookieSyncEndpoint) parseRequest(r *http.Request) (usersync.Request, pr
 		}
 	}
 
-	request, err = c.setLimit(request)
-	if err != nil {
-		return usersync.Request{}, privacy.Policies{}, err
-	}
+	request = c.setLimit(request, account.CookieSync)
+	request = c.setCooperativeSync(request, account.CookieSync)
```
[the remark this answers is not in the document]
bsardo: Yes it will never be nil. The `account.account.go#GetAccount` function called above this line always returns either an account object or an error. We rely on this behavior in our auction flows as well. I think the `GetAccount` unit test does a good job of ensuring this is the case. I suppose we could add an `if account != nil...return` check just after calling `GetAccount` here and in each of the auction flows to be safe. Alternatively we could make `GetAccount` always return an account object, in which case the function would need to be adjusted to still return an account object even when there is an error...or we could just leave everything as is. Thoughts?

SyntaxNode (approved):

AlexBVolcy (approved):

Issue #1715, opened 2021-02-18
Multi-Bid Support


bretg:
Copying the Prebid Server-related requirements from https://github.com/prebid/Prebid.js/issues/6317 -- go there for the use cases.

## Multi-Bid Requirements

1. PBS must be able to parse the ext.prebid.multibid config and allow the defined number of additional bids on seatbid.bid. If the number of bids allowed is less than 1, treat it as 1. If greater than 9, treat it as 9.
2. The extra bids must have different bidids in ext.prebid.bidid
3. The bid responses should support ad server targeting on the extra bids in the same way as PBJS: the biddercode used for the targeting of the extra bids should be definable as a targetbiddercodeprefix that incorporates a dynamic 'index' starting from 2. e.g. "bidA2, bidA3". Practically speaking, the base of the dynamic aliases will need to be 5 characters or less because the longest base attribute name is 14 chars (hb_cache_host_) and attributes are truncated at 20 chars.
    a) The extra bids receive the 'aliased' targets only if ext.prebid.targeting.includebidderkeys is specified and true
4. The bidresponse 'seat' should be associated with the primary biddercode, but store the resolved targeting biddercode in seatbid.bid.ext.prebid.targetbiddercode.
5. If the multibid config doesn't define a targetbiddercodepattern, no ad targeting should be generated for the extra bids. Just send them as additional seatbid.bid entries under the appropriate imp.
6. PBS analytics adapters may choose how to represent the extra bids: either with the primary biddercode, the targetbiddercode, or both.
7. In the future, we may want to allow account-level multi-bid config, but for now the config is only part of the incoming request or the stored request.
8. A multi-bid response still counts as one response for the purpose of metrics.
9. There are no additional logging or tracing requirements.
10. Extra bids must be cached in PBC when appropriate.
11. The system should support specifying a single bidder, in which case targetbiddercodeprefix is allowed. Or the user can specify multiple bidders, in which case targetbiddercodeprefix is not allowed. This is to support the scenario where multiple bids are allowed but without targeting variables.
12. If targetbiddercodeprefix is not specified, Prebid Server should still emit any additional bids but without setting targeting variables.
13. Validations:
    - If a given biddercode is specified multiple times in bidder and/or bidders, use the first bidder and add warning when in debug mode
    - If maxbids is not specified, ignore whole block and add warning when in debug mode
    - if bidders and targetbiddercodeprefix are both present, ignore targetbiddercodeprefix and add warning when in debug mode
[...]

PR #2254, opened 2022-05-22
[WIP] Multi-Bid Support

pm-nilesh-chate:
Multi-Bid Support #1715
https://github.com/prebid/prebid-server/issues/1715


pm-nilesh-chate:
Hi @bretg,

I'm working on multi-bid support for Go-PBS and have started these code changes in reference to Java's multi-bid implementation at https://github.com/prebid/prebid-server-java/pull/1169.

I needed help on a few questions before I continue with the changes.

- From Java-PBS, what I see is for winning-bid logic, `winningBidsByBidder` now will have all the bids for an impression rather than only one winning bid per impression.
- Is it safe to discard all the bids from an impression that were not selected by sort and maxBids limit. If correct, shouldn't all the bids under each impression qualify as `winningBidsByBidder` candidate? making `winningBidsByBidder` field obsolete.

Note: This is a draft POC, will complete the feature, clean the code, etc before converting it to PR.

mansinahar:
@bretg Can you please help @pm-nilesh-chate with his questions above when you have a chance? Thanks!

bretg:
The first two questions refer to a specific coding detail 'winningBidsByBidder' that I do not understand. I'm going to ignore that, and instead just restate what I believe is already stated in https://github.com/prebid/prebid-server/issues/1715 : 

1) by default, PBS returns all seat bids to the client
2) by default, targeting is only applied to the top bid from each seat (assuming includebidderkeys is set)
3) the multibid feature applies targeting to bids beyond the top bid from each seat (assuming includebidderkeys is set)
4) if multibid is not set and if host/account config auction.default-bid-limit-min is set, PBS will constrain the number bids. (Note: that config name is kinda confusing, but that's what PBS-Java has called it for a long time. Think of it as the 'bid-limit-that-can-be-modified-upwards-by-multibid')

> Existing Go-PBS does not assert 1 bid per impression, do we make it max 1 bid per impression if multi-bid is not support

The purpose of this 'auction.default-bid-limit-min' setting is to bring PBS-Go/Java together on this behavior. It's why Req 15 was added -- https://github.com/prebid/prebid-server/issues/1715#issuecomment-1148681203

pm-nilesh-chate:
Refer https://github.com/prebid/prebid-server/pull/2468 for further updates.
Continuing the changes with the correct account. Apologies for the inconvenience.

Issue #2174, opened 2022-03-04
Conventions and controls for adapters responding with multiple biddercodes


bretg:
The Prebid.js committee has been working through scenarios where a given bid adapter can return bids for biddercodes not defined in the original AdUnit. This issue describes them as "extra" bids -- https://github.com/prebid/Prebid.js/issues/8129

Use cases:
- bid adapters that own multiple exchanges (e.g. triplelift, grid)
- bid adapters that multiplex other bidders behind them

PBJS is proposing that two fields can be used to control and track these scenarios:
1. biddercode - this is the entity making the bid. Currently the PBS 'seat' is used to fill this field. 
2. adaptercode - indicates the bidder code that was in the adunit. This might be an alias of an underlying bid adapter. e.g. it would be "pubmatic" and not "groupm". This is a new field that would be set to the label used by the adunit.

## First phase
1. We extend the metadata object to include an adaptercode field, but have PBS-core fill it rather than relying on the adapter. PBS-core fills seatbid.bid.ext.prebid.meta.adaptercode with the biddercode as defined in imp[].ext.prebid.bidder.BIDDERCODE. Even if it's an alias, that's fine.
2. PBS-Core defines a way for adapters to set the seat (seatbid.seat) and does not initially validate what "seat" is set by the adapter. If the seat is not set, then PBS-core should set it to the biddercode.
3. Targeting values created for the `includebidderkeys` scenario always use biddercode for the suffix. The value of the hb_bidder key for the overall winner is also the biddercode.

## Second phase
There is concern over SPO in these cases (e.g. bid jamming). Or some kind of fraud, e.g. adapterA falsely bidding for adapterB to confuse algorithms. So host companies and publishers may want to control which adapters are trusted to do this, and possibly which biddercodes are returned from each.

PBJS defines parameters for controlling this behavior, so PBS could as well. The proposal is to make it part of account config and default account config:

```
account.alternatebiddercodes.enabled: true/false // the default for adapters not specifically named below
account.alternatebiddercodes.bidders.BIDDER.enabled: true/false
account.alternatebiddercodes.bidders.BIDDER.allowedbiddercodes: ["bidderX", "bidderY"] // defaults to all
```

1. if the bid response seat is different than the expected adapter code:
- check account.alternatebiddercodes.enabled. If it exists:
    - If false, reject. (see below)
    - if true, then check account.alternatebiddercodes.bidders.BIDDER.allowedbiddercodes.
[...]

PR #2257, opened 2022-05-26
Support adapters to set seat for a bid

pm-nilesh-chate:
Implements https://github.com/prebid/prebid-server/issues/2174

- Populate new field `seatbid.bid.ext.prebid.meta.adaptercode` with the adapter code.
- Allow adapters to set `seat` of bid(s).
- PubMatic sets `seat` for bids with value from `bid.ext.marketplace`.
- update existing tests
  - assert bid `seat` for all adapters.
  - update requestBid() mock to include `seat` and `adaptercode`.
- add new tests
  - add exchange test to validate extra bids under new `seat`.
  - add exchange test to validate extra bids and their `adaptercode` under new `seat` of an alias bidder.
  - add test to assert `requestBid()` changes.
  - Add `pubmatic` adapter test to assert extra-bid and a new response ext field `marketplace`.
- Validate extra-bidders as per phase-2 description.
- Do bid adjustment for extra-bid as per phase-2 description.


exchange/exchange.go
```diff
@@ -667,12 +675,12 @@ func (e *exchange) buildBidResponse(ctx context.Context, liveAdapters []openrtb_
 	// Create the SeatBids. We use a zero sized slice so that we can append non-zero seat bids, and not include seatBid
 	// objects for seatBids without any bids. Preallocate the max possible size to avoid reallocating the array as we go.
 	seatBids := make([]openrtb2.SeatBid, 0, len(liveAdapters))
-	for _, a := range liveAdapters {
```
pm-nilesh-chate, line 670: Cannot loop over `liveAdapters` as now, `getAllBids()` could bring extra bidders

exchange/exchange.go
```diff
@@ -553,8 +555,13 @@ func (e *exchange) getAllBids(
...
 
 		//if bidder returned no bids back - remove bidder from further processing
-		if brw.adapterBids != nil && len(brw.adapterBids.bids) != 0 {
-			adapterBids[brw.bidder] = brw.adapterBids
+		for _, seatBid := range brw.adapterSeatBids {
+			if seatBid != nil && len(seatBid.bids) != 0 {
+				// Do we need to address duplicate names here?. What if different BidderNames have exta bids with same seat value.
+				// Ex. 'pubmatic' adapter has extra bids under new 'groupm' seat and 'appnexus' also comes with additional bids under seat 'groupm'
+				// Above example might be more relatable with 'allowUnknownBidderCodes'
+				adapterBids[openrtb_ext.BidderName(seatBid.seat)] = seatBid
```
pm-nilesh-chate, line 563: Do we need to address duplicate names here?

pm-nilesh-chate: Do we clubs the bids in case multiple bidders returns the same alternate bidder code?. Ex. `appnexus` and `Pubmatic` returning `groupm` as a new bidder

@bretg thoughts?

pm-nilesh-chate: updated code to append the bids

pm-nilesh-chate:
Implemented validation in a separate PR https://github.com/PubMatic-OpenWrap/prebid-server/pull/309 against this branch

hhhjort:
I think we don't want to merge and release this without phase 2, as that will give PBS hosts some control over what adapters are allowed to do. Or perhaps include an off switch will prevent adapters from being able to rewrite the seat in seat bids. I can see a potential issue that downstream consumers could be caught off guard on assuming a one to one relation between seat and bidder, and this breaking some logic that depends on that.

As an example, Pub A is using Pubmatic and thus has a relation to them. It is not guaranteed that Pub A has a separate relationship with GroupM. Pub A may have some client logic to generate billing reports based on the seat names of the seat bid. This would result in billing reports for "GroupM" that Pub A does not know what to do with, since they don't have a direct relation there. Pub A can get the information from the adaptercode meta, but that will require them retooling their scripts to account for this new definition of seat. We can't require all publishers to retool their script in lock step with a new PBS release.

pm-nilesh-chate:
> I think we don't want to merge and release this without phase 2, as that will give PBS hosts some control over what adapters are allowed to do. Or perhaps include an off switch will prevent adapters from being able to rewrite the seat in seat bids. I can see a potential issue that downstream consumers could be caught off guard on assuming a one to one relation between seat and bidder, and this breaking some logic that depends on that.
> 
> As an example, Pub A is using Pubmatic and thus has a relation to them. It is not guaranteed that Pub A has a separate relationship with GroupM. Pub A may have some client logic to generate billing reports based on the seat names of the seat bid. This would result in billing reports for "GroupM" that Pub A does not know what to do with, since they don't have a direct relation there. Pub A can get the information from the adaptercode meta, but that will require them retooling their scripts to account for this new definition of seat. We can't require all publishers to retool their script in lock step with a new PBS release.

Merged phase-2 implementation (https://github.com/PubMatic-OpenWrap/prebid-server/pull/309) in this branch. Will address all the comments here :)

config/accounts.go
```diff
@@ -222,3 +225,40 @@ func (a *AccountIntegration) GetByIntegrationType(integrationType IntegrationTyp
...
+	if account.Adapters == nil {
+		return account.Enabled, nil
+	}
+
+	adapterCfg, ok := account.Adapters[bidder]
+	if !((ok && adapterCfg.Enabled) || (!ok && account.Enabled)) {
+		return false, nil
+	}
+
+	if len(adapterCfg.AllowedBidderCodes) == 0 || adapterCfg.AllowedBidderCodes[0] == "*" {
```
pm-nilesh-chate, line 253: Continuing discussion here w.r.t quoted comment at https://github.com/PubMatic-OpenWrap/prebid-server/pull/309#discussion_r888141638
> If we assume that an empty AllowedBidderCodes set means all bidder codes are allowed, which is a change from today's behavior. This requires PBS hosts to make a positive entry for every adapter to disable the new behavior, and continue creating a new entry every time a new adapter is released/enabled. I think we want to require a positive change for each adapter, like setting "*", as enabling this functionality requires additional trust in the behavior of an adapter that was not needed before. Or possibly allowing a default allowBidderCodes that determines what to do with an empty map.

I allowed `len(adapterCfg.AllowedBidderCodes) == 0` based on description on https://github.com/prebid/prebid-server/issues/2174. Quote:

> If true, then check account.alternateBidderCodes.adapters.BIDDER.allowedBidderCodes.
> - If doesn't exist or ["*"], then allow.

I'll update this code, making `AllowedBidderCodes` mandatory to be either `"*"` or a a list of bidders. Sounds good?

hhhjort: Making it mandatory would not be good unless you also put in a default of only allowing each adapter to set itself as the seat (today's behavior) We don't want to break everyone's config when they upgrade, which is what would happen if a new parameter is added that is also mandatory. As far as I know, we only did that once, with the GDPR default setting, after much discussion concluding there was no safe default we could use without risking legal issues.

pm-nilesh-chate: What I meant was:
1. We use `account.alternateBidderCodes.enabled: true/false` as overall toggle for this feature.  (default would be `false` to keep things running as is after upgrade)
     - If `false`, reject the alternate bidder.
     - If `true`, check  `account.alternateBidderCodes.adapters.BIDDER.allowedBidderCodes`. Here,
         - If `allowedBidderCodes` is "*", allow the alternate bidder.
         - Or, if alternate bidder is in `allowedBidderCodes` list, allow the alternate bidder.
         - Else, if alternate bidder is not in `allowedBidderCodes` list, reject the alternate bidder.

2. With this, I don't see the use of `account.alternateBidderCodes.adapters.BIDDER.enabled: true/false`. Since, default empty `allowedBidderCodes` list would act as `account.alternateBidderCodes.adapters.BIDDER.enabled = false`.


Can we please have the description at https://github.com/prebid/prebid-server/issues/2174 updated with correct expectations?

hhhjort (approved):
LGTM

pm-nilesh-chate:
Raised a new PR as this one is from wrong account. New PR: https://github.com/prebid/prebid-server/pull/2266

Apologies for the inconvenience.

PR #2198, opened 2022-03-22
add PGAM server bid adapter

GeneGenie:

PR #2258, opened 2022-05-30
MediaFuse prebid-server initial commit

balajimediafuse:
We have added Prebid server adapter for MediaFuse. Please review and approve it for Production.

Along with this PR, we also wanted the 

1. Installation instruction document for our clients to use MediaFuse Prebid-server adapter
2. How to test the Production MediFuse Prebid-server adapter.


balajimediafuse:
Team,
Please let me know what was causing the merge to be failed, and what change I need to do in the code.

AlexBVolcy:
So it appears you're trying to be an Alias of Appnexus. However, duplicating the code of the appnexus adapter isn't what should be done in this case.

Instead, you'll have to make some simple changes. Some of these changes include adding an entry to the appnexus endpoint in `config.go`, and utilizing the `appnexus Builder`. The only info you'll still be copy/pasting from the appnexus adapter is their yaml file and their bidder params file (both should be updated with the proper renaming to `Mediafuse`).

All changes you'll have to make will follow a very similar pattern to what this PR has done https://github.com/prebid/prebid-server/pull/2198/files -- So this should be a good resource.

You'll want to throw away this current PR, and create a new PR that contains the proper changes needed to become an Alias of Appnexus.

Thanks!

PR #2259, opened 2022-05-31
Add account query parameter to setuid endpoint

bsardo:
This PR adds the optional account query parameter to the setuid endpoint. Each time we hit the endpoint, an attempt is made to fetch the account. If a valid account was not provided, the account fetcher will return a valid account object consisting of the host-specified account defaults.
Note that the fetched account is not used in this PR. It will be used in a follow up PR in the GDPR calculations.

### DIFF
--- a/endpoints/setuid.go
+++ b/endpoints/setuid.go
@@ -79,6 +79,7 @@ func NewSetUIDEndpoint(cfg *config.Configuration, syncersByBidder map[string]use
 		if accountID == "" {
 			accountID = metrics.PublisherUnknown
 		}
+		// the fetched account is not currently used. It'll be used to get account-specific GDPR config in a future PR.
 		_, fetchErrs := accountService.GetAccount(context.Background(), cfg, accountsFetcher, accountID)
 		if len(fetchErrs) > 0 {
 			w.WriteHeader(http.StatusBadRequest)
--- a/endpoints/setuid.go
+++ b/endpoints/setuid.go
@@ -10,10 +10,12 @@ import (
 	"time"
 
 	"github.com/julienschmidt/httprouter"
+	accountService "github.com/prebid/prebid-server/account"
 	"github.com/prebid/prebid-server/analytics"
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/gdpr"
 	"github.com/prebid/prebid-server/metrics"
+	"github.com/prebid/prebid-server/stored_requests"
 	"github.com/prebid/prebid-server/usersync"
 	"github.com/prebid/prebid-server/util/httputil"
 )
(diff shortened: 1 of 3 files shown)


endpoints/setuid.go
```diff
@@ -73,6 +75,19 @@ func NewSetUIDEndpoint(cfg config.HostCookie, syncersByBidder map[string]usersyn
 			return
 		}
 
+		accountID := query.Get("account")
+		if accountID == "" {
+			accountID = metrics.PublisherUnknown
+		}
+		_, fetchErrs := accountService.GetAccount(context.Background(), cfg, accountsFetcher, accountID)
```
AlexBVolcy, line 82: For my knowledge, is this variable with the name `_` the spot where the fetched account will be utilized in the future GDPR PR that you mentioned in the description?

endpoints/setuid.go
```diff
@@ -73,6 +75,19 @@ func NewSetUIDEndpoint(cfg config.HostCookie, syncersByBidder map[string]usersyn
...
+		if len(fetchErrs) > 0 {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte(combineErrors(fetchErrs).Error()))
+			metricsEngine.RecordSetUid(metrics.SetUidBadRequest)
```
AlexBVolcy, line 86: Nitpick question: should we capitalize the `id` in `RecordSetUid` and `SetUidBadRequest` just to be consistent with the function name `NewSetUIDEndpoint`?

AlexBVolcy (commented):
Looks good but just a few minor comments

endpoints/setuid.go
bsardo: Yes, that's correct; that underscore will become `account` in a future PR. I'll then pass `account` into `gdpr.NewTCF2Config` which will create an aggregated GDPR config object consisting of the host and account GDPR settings that can then be used to build a GDPR permissions object.

endpoints/setuid.go
bsardo: We should but I think that's outside of the scope of this PR, especially since this issue has existed for quite some time and because there are a lot of occurrences. We can open another PR to fix this.

endpoints/setuid.go
bsardo: I added a comment so this is evident.

endpoints/setuid.go
```diff
@@ -101,7 +117,7 @@ func NewSetUIDEndpoint(cfg config.HostCookie, syncersByBidder map[string]usersyn
 		}
 
 		setSiteCookie := siteCookieCheck(r.UserAgent())
-		pc.SetCookieOnResponse(w, setSiteCookie, &cfg, cookieTTL)
+		pc.SetCookieOnResponse(w, setSiteCookie, &cfg.HostCookie, cookieTTL)
```
AlexBVolcy, line 120: For my own understanding, why was this function call updated to be `cfg.HostCookie`. How does this impact the overall goal of this PR?

AlexBVolcy (commented):
Looks good. I'm very close to approving, just had one question

endpoints/setuid.go
bsardo: Ah `cfg` here used to be of type `config.HostCookie`. See the `NewSetUIDEndpoint` function signature on line 31. I changed that to be of type `config.Configuration` so I would have access to the entire configuration which is needed by `GetAccount`. As a result, I now need to pass the `HostCookie` field of the `cfg` into `SetCookieOnResponse` here.

AlexBVolcy (approved):

mansinahar (approved):

PR #2263, opened 2022-06-02
Adnuntius: Bug Fix for regs being empty

mikael-lundin:
Just a quick fix that will not look for values in Regs if it's present but empty.

### DIFF
--- a/adapters/adnuntius/adnuntius.go
+++ b/adapters/adnuntius/adnuntius.go
@@ -233,7 +233,7 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, externalRequest *adapte
 func getGDPR(request *openrtb2.BidRequest) (string, string, error) {
 	gdpr := ""
 	var extRegs openrtb_ext.ExtRegs
-	if request.Regs != nil {
+	if request.Regs != nil && request.Regs.Ext != nil {
 		if err := json.Unmarshal(request.Regs.Ext, &extRegs); err != nil {
 			return "", "", fmt.Errorf("failed to parse ExtRegs in Adnuntius GDPR check: %v", err)
 		}
--- a/adapters/adnuntius/adnuntiustest/supplemental/check-gdpr.json
+++ b/adapters/adnuntius/adnuntiustest/supplemental/check-gdpr.json
@@ -1,109 +1,110 @@
 {
-    "mockBidRequest": {
-        "id": "test-request-id",
-        "user": {
-            "id": "1kjh3429kjh295jkl",
-            "ext": {
-                "consent": "CONSENT_STRING"
-            }
-        },
-        "regs": {
-            "ext": {
-                "gdpr": 1
-            }
-        },
-        "imp": [
-            {
-                "id": "test-imp-id",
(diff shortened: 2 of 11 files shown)


adapters/adnuntius/adnuntius.go
```diff
@@ -233,7 +233,7 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, externalRequest *adapte
 func getGDPR(request *openrtb2.BidRequest) (string, string, error) {
 	gdpr := ""
 	var extRegs openrtb_ext.ExtRegs
-	if request.Regs != nil {
+	if request.Regs != nil && request.Regs.Ext != nil {
```
guscarreon, line 236: I believe the `check-gdpr.json` test covers the case where both `request.Regs != nil` and `request.Regs.Ext != nil` evaluate to true. 
image

Can we also add coverage to the scenarios shown below?
| `request.Regs != nil` | `request.Regs.Ext != nil` | `json.Unmarshal(request.Regs.Ext, &extRegs); err != nil` |
|--- | --- | ---|
|false|n/a|n/a|
|true|false|n/a|
|true|true|true|

guscarreon (commented):
Just a test coverage question

adapters/adnuntius/adnuntius.go
mikael-lundin: yes, That's true, the only issue is that one of our customers sends an empty regs object, where our adapter returns an error instead of an ad, and they were unable to change that. That's why we need to not return an error. :)

mikael-lundin: any news about this one ?

guscarreon: > That's why we need to not return an error. 

Cool. May not return an error if you don't believe you need to. The comment above has to do with adding test coverage to your code. Particularly, adding a test case similar to `adapters/adnuntius/adnuntiustest/supplemental/check-gdpr.json` but with an empty `"regs"` object or an empty `"regs.ext"` so we can test those two scenarios. 

Say a `adapters/adnuntius/adnuntiustest/supplemental/empty-regs.json` file with something like:
```
{
    "mockBidRequest": {
        "id": "test-request-id",
        "user": {
            "id": "1kjh3429kjh295jkl",
            "ext": {
                "consent": "CONSENT_STRING"
            }
        },
        "regs": { },
   .
   .
```
And an additional `adapters/adnuntius/adnuntiustest/supplemental/empty-regs-ext.json` file with something like:
```
{
    "mockBidRequest": {
        "id": "test-request-id",
        "user": {
            "id": "1kjh3429kjh295jkl",
            "ext": {
                "consent": "CONSENT_STRING"
            }
        },
        "regs": {
            "ext": { }
        },
   .
   .
```

mikael-lundin: Sure, will fix it right away!

guscarreon (approved):
LGTM. @mikael-lundin thanks for addressing my feedback.

AlexBVolcy (approved):

PR #2264, opened 2022-06-03
GDPR: move request-specific data onto permissions object

bsardo:
~~This PR is in draft. I'll rebase it and put it up for formal review once the open PRs it depends on are merged.~~

This PR moves request-specific GDPR data (signal, consent, alias GVL IDs and publisher ID) onto the permissions object. The auction flow is updated accordingly.

Also the cookie sync and setuid endpoints have been updated as follows:
- the GDPR object is instantiated on a per-request basis in each endpoint rather than being instantiated once for these endpoints at startup
- the fetched account is now used to generate an aggregated config (host GDPR TCF2 config + account GDPR config)
- the aggregated config is passed into the per-request GDPR permissions object

### DIFF
--- a/endpoints/cookie_sync.go
+++ b/endpoints/cookie_sync.go
@@ -41,7 +41,8 @@ var cookieSyncBidderFilterAllowAll = usersync.NewUniformBidderFilter(usersync.Bi
 func NewCookieSyncEndpoint(
 	syncersByBidder map[string]usersync.Syncer,
 	config *config.Configuration,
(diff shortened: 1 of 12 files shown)


VeronikaSolovei9:
Code looks cleaner and easier now. Need another pair of eyes on this PR, I'm not a bid specialist in GDPR.

hhhjort (approved):
Looks good to me.

VeronikaSolovei9 (approved):

PR #2265, opened 2022-06-07
[UNICORN Adapter] support JPY only

faithnh:
Since the currency only supports JPY, I fixed the currency to JPY.

### DIFF
--- a/adapters/unicorn/unicorn.go
+++ b/adapters/unicorn/unicorn.go
@@ -64,6 +64,9 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 		}
 	}
 
+	// Support Cur "JPY" only
+	request.Cur = []string{"JPY"}
+
 	err := modifyImps(request)
 	if err != nil {
 		return nil, []error{err}
--- a/adapters/unicorn/unicorntest/exemplary/banner-app-without-ext.json
+++ b/adapters/unicorn/unicorntest/exemplary/banner-app-without-ext.json
@@ -85,6 +85,7 @@
             "ver": "1.9.0"
           },
           "at": 1,
+          "cur": ["JPY"],
           "device": {
             "connectiontype": 1,
             "ext": {
--- a/adapters/unicorn/unicorntest/exemplary/banner-app-without-placementid.json
+++ b/adapters/unicorn/unicorntest/exemplary/banner-app-without-placementid.json
@@ -28,7 +28,7 @@
         }
       }
     ],
-    "cur": ["USD"],
+    "cur": ["JPY"],
     "app": {
       "publisher": {
         "id": "test"
@@ -96,7 +96,7 @@
             "ver": "1.9.0"
           },
           "at": 1,
-          "cur": ["USD"],
+          "cur": ["JPY"],
           "device": {
             "connectiontype": 1,
             "ext": {


adapters/unicorn/unicorn.go
```diff
@@ -64,6 +64,9 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 		}
 	}
 
+	// Support Cur "JPY" only
+	request.Cur = []string{"JPY"}
```
SyntaxNode, line 68: Do you think it would be a good idea to return an error if a non JPY currency is specified? .. or would you consider trying to use the built-in currency converter?

faithnh: we want to support only the currency JPY, we chose the former (i.e. return an error if a non JPY currency is specified).

But when no currency is specified, we specify currency JPY as default.

https://github.com/prebid/prebid-server/pull/2265/commits/771617d0a57024db1e94f1f72ff5564503d34212

adapters/unicorn/unicorn.go
```diff
@@ -98,6 +107,15 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
 	return []*adapters.RequestData{requestData}, nil
 }
 
+func isOtherThanJPYSpecified(cur []string) bool {
```
AlexBVolcy, line 110: Nitpick: But could this function name be renamed to something like `isCurOtherThanJPY` or `isOtherThanJPY`. I just think the function name could be a little more simple.

AlexBVolcy (commented):
Looks good -- close to approving, just one minor comment.

adapters/unicorn/unicorn.go
```diff
@@ -64,6 +64,15 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapte
...
 	}
 
+	if isOtherThanJPY(request.Cur) {
+		return nil, []error{&errortypes.BadInput{
+			Message: "Only JPY is supported",
+		}}
+	}
+	if len(request.Cur) == 0 {
+		request.Cur = []string{"JPY"}
+	}
```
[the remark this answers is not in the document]
faithnh: Sure, no problem.

Applied the former. 

https://github.com/prebid/prebid-server/pull/2265/commits/90afeb0355cb754fb66ec675896df38c1c2d7c5e

AlexBVolcy (approved):

guscarreon (approved):
@faithnh thank you for addressing my feedback

bretg:
Sorry to re-open the conversation here, but this change might not be not ideal for Unicorn. Bid adapters don't need to drop requests because the publisher's ad server is set up in another currency. My understanding is that the Unicorn endpoint can only accept JPY. No problem. Here's how this could be implemented:

1) When creating the request to the Unicorn endpoint, set cur to ["JPY"]. It doesn't matter what the original currency was. 
2) If you support floors, then use [the provided function](https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#currency) to convert imp.bidfloor to JPY. 
3) On the response, simply tell Prebid Server that the bid is in JPY.

That's it! Prebid Server will take care of converting the JPY bid to whatever currency was in the original request.

If you want to limit bids to people based in Japan, that might be different. Some Prebid Server implementations may set device.geo with the country code, but others might not. So if your goal is to limit traffic to publishers that work only in JPY, this change is ok.

bretg:
My understanding is that the issue is rooted in discrepancies: Prebid Server will convert a JPY bid into the ad server currency based on today's exchange rate as defined in https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json

Wins are reported daily, but if billing happens monthly on an average exchange rate, I can see how there would be discrepancies. If your goal is to completely eliminate reporting issues like this there two options:
1) what you're doing -- just don't bid on inventory that could wind up with discrepancies
2) upgrade the billing system to be able to look up the exchange rates on the day the wins occurred. Prebid uses information available from the EU Central bank.

ctylim:
Hi @bretg I'm also responsible for Prebid Server integration/implementation from UNICORN.

Thanks to your mentioning the issue, I may have some misunderstandings what we implemented. We are going to rollback these commits and support multiple currencies on our endpoint, if my understanding becomes correct.

We didn't consider that PBS forgets the original currency from adserver when we force to delete it in the PBS bid requests. Your concerns will be resolved by simply leaving the original currency in our bid adapter.

> My understanding is that the issue is rooted in discrepancies

Yes actually the issue comes from billing and accounting.
In our unique situation, we, as a DSP, manage exchanges by ourselves because most of our clients charge by JPY. We don't want to use PBS exchange rates if possible because our intending bid price and PBS bid price will differ.

We also request the following be guarranteed:

- We will have to pay our spends by the currency which corresponds to publisher accounts. Original currency in the adserver desirably equals with the accounting currency.
- `imp[].ext.bidder.accountId` field in our bid adapter will be required, which is issued per publisher account by UNICORN. The spend discrepancy tests are also possible per account, by their original currency.

Issue #1491, opened 2020-09-11
Prebid Server Adding an SChain node


bretg:
We haven't actually been asked for this yet, but I predict that someday someone's going to want to have Prebid Server _add_ an schain node to an existing list in source.ext.schain.

Currently, PBS-Go can support _initializing_ the schain with the [default server request](https://docs.prebid.org/prebid-server/features/pbs-default-request.html), but neither version has the ability to add a node to the end of an existing list.

The proposal:
1. Add a new optional server-level config that lets the host company define an schain node.
2. If this config is specified and is valid:
   2a. If source.ext.schain doesn't exist, the configured schain node is added there as the first element in the node array
   2b. If source.ext.schain does exist, the configured schain node is appended to the end of the node array

e.g.
```
auction:
    host-schain-node: '{ "asi":"pbshostcompany.com", "sid":"00001", "rid":"BidRequest", "hp":1 }'
```
Would get added to the end of source.ext.schain.nodes:
```
  "source": {
    "ext": {
      "schain": {
        "ver":"1.0",
        "complete": 1,
        "nodes": [
          {
            "asi":"firstschainnode.com",
            "sid":"00001",
            "rid":"BidRequest1",
            "hp":1
          },
          { "asi":"pbshostcompany.com", "sid":"00001", "rid":"BidRequest", "hp":1 }
        ]     
      } 
    }
  }
```

SyntaxNode:
> We haven't actually been asked for this yet,

Shouldn't we wait until we discover a need to implement such a feature?

AlexBVolcy:
This issue is now being worked on by me! I expect a PR to be out early next week.

AlexBVolcy:
This feature has been implemented and merged in, and will be released in PBS-Go v0.217.0

PR #2272, opened 2022-06-13
Allow Host to Add SChain Node

AlexBVolcy:
This PR addresses this issue https://github.com/prebid/prebid-server/issues/1491

The goal is to give the host a way to optionally define an `schain_node` at the config level. In `config.go`, there's a new member of the Configuration struct, `SChainNode` of type `*openrtb_ext.ExtRequestPrebidSChainSChainNode`. And there's a corresponding `SetDefault`, which will allow the host to define their `schain_node`

If this new schain node is set, the node should be appended to the existing schain nodes that are tied to a specific bidder request at `source.ext.schain`. If `source.ext.schain` is not defined, it needs to be created, and the host node needs to be the first node of that path. The main implementation of this part of the issue is found in `schainwriter.go`

Unit tests were added to accompany the changes

End to end json tests were added as well in `exchange_test.json`

### DIFF
--- a/schain/schainwriter_test.go
+++ b/schain/schainwriter_test.go
@@ -15,6 +15,8 @@ func TestSChainWriter(t *testing.T) {
 	const seller2SChain string = `"schain":{"complete":2,"nodes":[{"asi":"directseller2.com","sid":"00002","rid":"BidRequest2","hp":2}],"ver":"2.0"}`
 	const seller3SChain string = `"schain":{"complete":3,"nodes":[{"asi":"directseller3.com","sid":"00003","rid":"BidRequest3","hp":3}],"ver":"3.0"}`
 	const sellerWildCardSChain string = `"schain":{"complete":1,"nodes":[{"asi":"wildcard1.com","sid":"wildcard1","rid":"WildcardReq1","hp":1}],"ver":"1.0"}`
+	const hostNode string = `{"asi":"pbshostcompany.com","sid":"00001","rid":"BidRequest","hp":1}`
+	const seller1Node string = `{"asi":"directseller1.com","sid":"00001","rid":"BidRequest1","hp":1}`
 
 	tests := []struct {
 		description    string
@@ -157,7 +159,7 @@ func TestSChainWriter(t *testing.T) {
 		{
 			description: "Schain in request, host schain defined, source.ext for bidder request should update with appended host schain",
 			giveRequest: openrtb2.BidRequest{
-				Ext:    json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[{"asi":"appnexus.com","sid":"1792","rid":"6a586290-ae54-44e9-8985-d60821f7bb07","hp":1}],"ver":"1.0"}}]}}`),
+				Ext:    json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[` + seller1Node + `],"ver":"1.0"}}]}}`),
 				Source: nil,
 			},
 			giveBidder: "testbidder",
@@ -165,9 +167,9 @@ func TestSChainWriter(t *testing.T) {
 				ASI: "pbshostcompany.com", SID: "00001", RID: "BidRequest", HP: 1,
 			},
 			wantRequest: openrtb2.BidRequest{
-				Ext: json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[{"asi":"appnexus.com","sid":"1792","rid":"6a586290-ae54-44e9-8985-d60821f7bb07","hp":1}],"ver":"1.0"}}]}}`),
+				Ext: json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[` + seller1Node + `],"ver":"1.0"}}]}}`),
 				Source: &openrtb2.Source{
-					Ext: json.RawMessage(`{"schain":{"complete":1,"nodes":[{"asi":"appnexus.com","sid":"1792","rid":"6a586290-ae54-44e9-8985-d60821f7bb07","hp":1},{"asi":"pbshostcompany.com","sid":"00001","rid":"BidRequest","hp":1}],"ver":"1.0"}}`),
+					Ext: json.RawMessage(`{"schain":{"complete":1,"nodes":[` + seller1Node + `,` + hostNode + `],"ver":"1.0"}}`),
 				},
 			},
 		},
@@ -184,7 +186,7 @@ func TestSChainWriter(t *testing.T) {
 			wantRequest: openrtb2.BidRequest{
 				Ext: nil,
 				Source: &openrtb2.Source{
-					Ext: json.RawMessage(`{"schain":{"complete":0,"nodes":[{"asi":"pbshostcompany.com","sid":"00001","rid":"BidRequest","hp":1}],"ver":"1.0"}}`),
+					Ext: json.RawMessage(`{"schain":{"complete":0,"nodes":[` + hostNode + `],"ver":"1.0"}}`),
 				},
 			},
 		},
--- a/exchange/exchange_test.go
+++ b/exchange/exchange_test.go
@@ -2205,7 +2205,7 @@ func runSpec(t *testing.T, filename string, spec *exchangeSpec) {
 	if spec.BidIDGenerator != nil {
 		*bidIdGenerator = *spec.BidIDGenerator
 	}
-	ex := newExchangeForTests(t, filename, spec.OutgoingRequests, aliases, privacyConfig, bidIdGenerator)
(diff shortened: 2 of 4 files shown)


schain/schainwriter.go
```diff
@@ -57,13 +57,26 @@ func (w SChainWriter) Write(req *openrtb2.BidRequest, bidder string) {
...
+
+	schain := openrtb_ext.ExtRequestPrebidSChain{}
+
+	if selectedSChain != nil {
+		schain.SChain = *selectedSChain
 	}
+
+	if hostSChainInfo != "" {
+		var hostSChainNode *openrtb_ext.ExtRequestPrebidSChainSChainNode
+		if err := json.Unmarshal([]byte(hostSChainInfo), &hostSChainNode); err != nil {
```
bsardo, line 69: I see we are unmarshalling the schain node because it was defined as a string. This leads me to believe it would be better if we defined the schain in the config as a struct. Alternatively we could allow the host to still define it as a string and unmarshal it into a struct at startup which would then be used here so we aren't performing an inefficient unmarshal on every request. I tend to lean though towards allowing the host to define it as a struct in their config. I think it results in leaner config code and is easier for the host to write. @mansinahar thoughts?

exchange/utils.go
```diff
@@ -227,7 +229,10 @@ func getAuctionBidderRequests(auctionRequest AuctionRequest,
 		reqCopy := *req.BidRequest
 		reqCopy.Imp = imps
 
-		sChainWriter.Write(&reqCopy, bidder)
+		err := sChainWriter.Write(&reqCopy, bidder, hostSChainNode)
+		if err != nil {
+			return nil, []error{err}
+		}
```
bsardo, line 235: I'm pretty sure we intentionally left out this error handling because we didn't want invalid schains killing the auction. I'll need to check on that.

schain/schainwriter.go
```diff
@@ -57,13 +57,26 @@ func (w SChainWriter) Write(req *openrtb2.BidRequest, bidder string) {
...
-	schain := openrtb_ext.ExtRequestPrebidSChain{
-		SChain: *selectedSChain,
+
+	schain := openrtb_ext.ExtRequestPrebidSChain{}
+
+	if selectedSChain != nil {
+		schain.SChain = *selectedSChain
 	}
+
+	if hostSChainInfo != "" {
```
bsardo, line 67: For the scenario where the lone schain node is the host defined node, we set the schain `complete` field to `0`, which is the default value for that field when an instance of `openrtb_ext.ExtRequestPrebidSChain` is created. The `ver` field is set to `""` by default but I believe that is required as you can see here [OpenRTB Object: SupplyChain](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/supplychainobject.md#openrtb-object-supplychain). The question is what should it be?

bsardo: After further research, it appears that `ver` should be set to `"1.0"` which is the latest and sole version of `sellers.json` which is linked to from [here](https://iabtechlab.com/sellers-json/). The change log on `sellers.json` shows 1.0.

schain/schainwriter.go
```diff
@@ -33,37 +33,49 @@ type SChainWriter struct {
...
 	bidderSChain := w.sChainsByBidder[bidder]
 
 	// source should not be modified
-	if bidderSChain == nil && wildCardSChain == nil {
-		return
+	if bidderSChain == nil && wildCardSChain == nil && hostSChainNode == nil {
+		return nil
 	}
 
+	selectedSChain = &openrtb_ext.ExtRequestPrebidSChainSChain{}
```
mansinahar, line 48: Why not just set the default `version` here when initializing this struct?

schain/schainwriter.go
```diff
@@ -33,37 +33,49 @@ type SChainWriter struct {
 // Write selects an schain from the multi-schain ORTB 2.5 location (req.ext.prebid.schains) for the specified bidder
 // and copies it to the ORTB 2.5 location (req.source.ext). If no schain exists for the bidder in the multi-schain
 // location and no wildcard schain exists, the request is not modified.
-func (w SChainWriter) Write(req *openrtb2.BidRequest, bidder string) {
+func (w SChainWriter) Write(req *openrtb2.BidRequest, bidder string, hostSChainNode *openrtb_ext.ExtRequestPrebidSChainSChainNode) error {
```
mansinahar, line 36: Why add `error` as a return parameter here? There doesn't seem to be any case where this function will return an error

schain/schainwriter.go
AlexBVolcy: Good idea!

schain/schainwriter.go
AlexBVolcy: You're right, there shouldn't be any error handling in this function

schain/schainwriter.go
AlexBVolcy: Just to keep this updated, we talked during team time a couple days ago, and the type of the host schain node has now been reverted from string to be type `*openrtb_ext.ExtRequestPrebidSChainSChainNode`

schain/schainwriter_test.go
```diff
@@ -152,6 +154,40 @@ func TestSChainWriter(t *testing.T) {
...
+			giveHostSChain: &openrtb_ext.ExtRequestPrebidSChainSChainNode{
+				ASI: "pbshostcompany.com", SID: "00001", RID: "BidRequest", HP: 1,
+			},
+			wantRequest: openrtb2.BidRequest{
+				Ext: nil,
+				Source: &openrtb2.Source{
+					Ext: json.RawMessage(`{"schain":{"complete":0,"nodes":[{"asi":"pbshostcompany.com","sid":"00001","rid":"BidRequest","hp":1}],"ver":"1.0"}}`),
+				},
+			},
+		},
```
bsardo, line 190: Nitpick: to improve readability and continue with the style of this test, I'd suggest defining a couple of schain node constants at the top of this test like so:
```
const hostNode    string = `{"asi":"pbshostcompany.com","sid":"00001","rid":"BidRequest","hp":1}`
const seller1Node string = `{"asi":"directseller1.com","sid":"00001","rid":"BidRequest1","hp":1}`
```
and then using them here so we don't have long inline JSON strings that are hard to read and digest for someone unfamiliar with the logic:
```
@@ -157,7 +160,7 @@ func TestSChainWriter(t *testing.T) {
                {
                        description: "Schain in request, host schain defined, source.ext for bidder request should update with appended host schain",
                        giveRequest: openrtb2.BidRequest{
-                               Ext:    json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[{"asi":"appnexus.com","sid":"1792","rid":"6a586290-ae54-44e9-8985-d60821f7bb07","hp":1}],"ver":"1.0"}}]}}`),
+                               Ext:    json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[` + seller1Node + `],"ver":"1.0"}}]}}`),
                                Source: nil,
                        },
                        giveBidder: "testbidder",
@@ -165,9 +168,9 @@ func TestSChainWriter(t *testing.T) {
                                ASI: "pbshostcompany.com", SID: "00001", RID: "BidRequest", HP: 1,
                        },
                        wantRequest: openrtb2.BidRequest{
-                               Ext: json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[{"asi":"appnexus.com","sid":"1792","rid":"6a586290-ae54-44e9-8985-d60821f7bb07","hp":1}],"ver":"1.0"}}]}}`),
+                               Ext: json.RawMessage(`{"prebid":{"schains":[{"bidders":["testbidder"],"schain":{"complete":1,"nodes":[` + seller1Node + `],"ver":"1.0"}}]}}`),
                                Source: &openrtb2.Source{
-                                       Ext: json.RawMessage(`{"schain":{"complete":1,"nodes":[{"asi":"appnexus.com","sid":"1792","rid":"6a586290-ae54-44e9-8985-d60821f7bb07","hp":1},{"asi":"pbshostcompany.com","sid":"00001","rid":"BidRequest","hp":1}],"ver":"1.0"}}`),
+                                       Ext: json.RawMessage(`{"schain":{"complete":1,"nodes":[` + seller1Node +`,` + hostNode + `],"ver":"1.0"}}`),
                                },
                        },
                },
@@ -184,7 +187,7 @@ func TestSChainWriter(t *testing.T) {
                        wantRequest: openrtb2.BidRequest{
                                Ext: nil,
                                Source: &openrtb2.Source{
-                                       Ext: json.RawMessage(`{"schain":{"complete":0,"nodes":[{"asi":"pbshostcompany.com","sid":"00001","rid":"BidRequest","hp":1}],"ver":"1.0"}}`),
+                                       Ext: json.RawMessage(`{"schain":{"complete":0,"nodes":[` + hostNode + `],"ver":"1.0"}}`),
                                },
                        },
                }
```

bsardo (commented):
This is looking good -- just a couple of minor comments.

bsardo (approved):
LGTM

mansinahar (approved):

PR #2274, opened 2022-06-17
New Adapter: sspBC Bid Adapter

wojciech-bialy-wpm:
Prebid-server version of sspBC Bid Adapter (see https://github.com/prebid/Prebid.js/blob/master/modules/sspBCBidAdapter.md & https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/sspBC.md) 

Supported Media Types - banner only.

Test parameters:
`
var adUnits = [
   {
      code: 'banner-div',
      mediaTypes: {
        banner: {
          sizes: [[300, 250]]
        }
      },
      bids: [{
         bidder: 'sspBC',
         params: {
            id: "006",
            siteId: "235911",
            test: 1
         }
       }]
   }
];
`


adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+
+// Ext data added by proxy
+type SsbcResponseExt struct {
+	AdLabel     string `json:"adlabel"`
+	PublisherId string `json:"pubid"`
+	SiteId      string `json:"siteid"`
+	SlotId      string `json:"slotid"`
+}
+
+type SspbcAdapter struct {
```
SyntaxNode, line 62: Please rename to simply `adapter`. There is no need to repeat the package and no need to export the type by starting with an uppercase letter. Access modifiers in Go are determined by if the first letter of the name if upper case (exported) or lower case (package private). Please change the other types to also start with a lower case letter.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+	adSizes        map[string]int
+	bannerTemplate *template.Template
+}
+
+// ---------------ADAPTER INTERFACE------------------
+// Builder builds a new instance of the sspBC adapter
+func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
+	// find path to template file
+	pathToTemplate := "./adapters/sspbc/bannerTemplate.html"
+	workingDir, err := os.Getwd()
```
SyntaxNode, line 77: Loading the file seems complicated. Please consider hardcoding as a constant (in it's own file if you prefer) and parse the template from the string. Then you don't need to worry about the file system at all or handle the difference between tests and production.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+	if err == nil && strings.HasSuffix(workingDir, "sspbc") {
+		// this is a test running in adapter's directory
+		pathToTemplate = "bannerTemplate.html"
+	} else if err != nil {
+		// this error does not break createBannerAd flow
+		glog.Errorf("SSPBC: Cannot get working directory, assuming default path")
```
SyntaxNode, line 83: Please do not use `glog` from adapter code. Kindly remove all calls and allow the errors to propagate through the core code.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+func (a *SspbcAdapter) MakeRequests(request *openrtb2.BidRequest, requestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var errors []error
+
+	formattedRequest, err := formatSsbcRequest(a, request)
+	if err != nil {
+		glog.Errorf("SSPBC: cannot prepare request")
+		errors = append(errors, err)
+		return nil, errors
```
SyntaxNode, line 107: Nitpick: It would be simpler to create the errors array when necessary, such as `return nil, []error{err}` instead of declaring  and appending to the `errors` variable. It looks cleaner IMHO.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+	requestJSON, err := json.Marshal(formattedRequest)
+	if err != nil {
+		glog.Errorf("SSPBC: cannot marshal request")
+		errors = append(errors, err)
+		return nil, errors
+	}
+
+	requestData := &adapters.RequestData{
+		Method: "POST",
+		Uri:    fmt.Sprintf("%s?bdver=%s&inver=0", a.endpoint, a.version),
```
SyntaxNode, line 119: Consider using the `url` package to append query parameters instead of relying on string templating as that approach is more robust.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+			"sn": "sspbc-test"
+			}
+
+		Note - we cannot read site SN, since response.sn is not defined in
+		openRTB2.BidResponse structure
+
+		For now we set SN as sspbc_go
+
+		Long term SN should be returned in bid.ext
+	*/
```
SyntaxNode, line 167: I like the detail of the comment, but consider moving it to a markdown file in the directory to separate it from the code flow.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+			var BidType openrtb_ext.BidType
+			var BidId = bid.ImpID
+
+			/*
+			  here we should make a call to getBidType method, and based on detected type
+			  make call to createBannerAd, createVideoAd, createNativeAd methods
+
+			  for now we set it to "banner"
+			*/
+			BidType = openrtb_ext.BidTypeBanner
```
SyntaxNode, line 201: We will accept a hardcoded banner value since the yaml file only defines a banner type. If you change that in the future, we'll ask for a proper implementation here. Ideally, the bid response should include the creative media type and we will soon add support for the new mtype field added in OpenRTB 2.6.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+	if err := json.Unmarshal(externalResponse.Body, &response); err != nil {
+		return nil, []error{err}
+	}
+
+	bidResponse := adapters.NewBidderResponseWithBidsCapacity(len(internalRequest.Imp))
+	bidResponse.Currency = response.Cur
+
+	for _, seatBid := range response.SeatBid {
+		for _, bid := range seatBid.Bid {
+			var BidType openrtb_ext.BidType
```
SyntaxNode, line 192: Nitpick: Following Go syntax guides, all variables should start with a lower case letter.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+		byteTemplate := []byte(filledTemplate.String())
+		fmt.Println(base64.StdEncoding.EncodeToString(byteTemplate)) */
+
+	return filledTemplate.String(), nil
+}
+
+func getImpSize(Imp openrtb2.Imp) string {
+	if Imp.Video != nil {
+		return fmt.Sprintf("%dx%d", Imp.Video.W, Imp.Video.H)
+	}
```
SyntaxNode, line 295: Consider removing code specific to video since your adapter only supports banner types, per your yaml file.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+
+	if Imp.Banner != nil {
+		sb := strings.Builder{}
+		areaMax := int64(0)
+		sb.WriteString("1x1")
+		for _, sizeI := range Imp.Banner.Format {
+			areaI := sizeI.W * sizeI.H
+			if areaI > areaMax {
+				areaMax = areaI
+				sb.Reset()
```
SyntaxNode, line 305: If the end result is just a single size, there is no need to use a strings.Builder. That's only appropriate for a continuous long string.

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,407 @@
...
+	var siteId string
+	var isTest int
+
+	// check if adSlots and adSizes maps are initialized
+	if a.adSlots == nil {
+		a.adSlots = make(map[string]AdSlotData)
+	}
+	if a.adSizes == nil {
+		a.adSizes = make(map[string]int)
+	}
```
SyntaxNode, line 327: Why are adSlots and adSizes global to your adapter?

static/bidder-info/sspBC.yaml
```diff
@@ -0,0 +1,12 @@
+maintainer:
+  email: "wojciech.bialy@grupawp.pl "
```
SyntaxNode, line 2: We require a group email address instead of a specific person. Please refer to our new adapter coding guide: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#respect-the-rules

static/bidder-info/sspBC.yaml
```diff
@@ -0,0 +1,12 @@
...
+gvlVendorID: 676
+capabilities:
+  site:
+    mediaTypes:
+      - banner
+userSync:
+  default: iframe
+  iframe:
+    url: https://ssp.wp.pl/bidder/usersync?tcf=2&redirect={{.RedirectURL}}
```
SyntaxNode, line 11: Does your bidding server need privacy policy information such as CCPA and GDPR?

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: OK, renamed.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: Removed all logging from adapter code.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: OK, fixed.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: Ok, changed tu url.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: Added basic detection of bid type (at the moment only for only banner bids). Bids that are not recognized will not be appended to response.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: I've removed all references to Video in current version (to be reintroduced later, when video support is ready).

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: adSlots and adSizes are used between auctions to determine siteid & slotid, if those are not present in bid params.
Basically we are mapping those id's to either adSlot name, or AdSize x AdSizeUsedCount

static/bidder-info/sspBC.yaml
wojciech-bialy-wpm: Changed email to group alias.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: OK,  I've removed unnecessary info from comments.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: Ok, I've changed this part of code, and replaced unnecessary strings.Builder with a simpler Sprintf

adapters/sspbc/sspbc.go
SyntaxNode: Thank you.

static/bidder-info/sspBC.yaml
SyntaxNode: Ok, cool. Just wanted to make sure you got it covered.. and you do.

adapters/sspbc/sspbc.go
SyntaxNode: There is only ever once instance of the adapter in memory. How does the adSlot and adSizes persistent map work across different publishers?

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,354 @@
...
+	return impSize
+}
+
+func (a *adapter) fillAdSlotData(impI openrtb2.Imp, impSize string) adSlotData {
+    var extData adSlotData
+    a.mux.Lock()
+    defer a.mux.Unlock()
+    // save slot data
+    a.adSizes[impSize] = a.adSizes[impSize] + 1
+    if a.adSlots[impI.ID].PbSlot != "" {
```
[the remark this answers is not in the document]
wojciech-bialy-wpm: This structure should help to detect placement on the page, when placement data (site id, placement id) is not provided. 

We do the same on prebid.js side, and if publisher does not supply id's, we will, for instance, send info that this adunit name = X, and  is third time that 300x250 placement was called. The second part (represented by above struct) will help if adUnit name is pseudorandom (as is the case with some publishers).

Buy You're right - while in js adapter case we have one instance of adapter running in browser environment, this will not work if there are multiple PBS instances running in a cluster.

I'm now removing sizeXuse mapping, and in the case of psudorandom placements we will resort to additional first party data supplied by the publisher.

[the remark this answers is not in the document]
wojciech-bialy-wpm: Done - adSizes are now longer kept in the adapter. AdSlots still are - we need that data between request and response - but that has been simplified.

adapters/sspbc/sspbc.go
wojciech-bialy-wpm: Please take a look at below convo. This is the equivalent of additional slot info that we're doing in js adapter, but - as @hhhjort pointed out - it will not work if there's more than one instance of PBS running. This is now simplified (and functionality that it provided will have to be provided as first party data)

adapters/sspbc/sspbc.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// default fallback
+	return impSize
+}
+
+func (a *adapter) fillAdSlotData(impI openrtb2.Imp, impSize string) adSlotData {
+    var extData adSlotData
+    extData.PbSlot = impI.TagID
+    extData.PbSize = impSize
+    a.adSlots[impI.ID] = extData
```
hhhjort, line 261: You are still reading and writing shared data in a thread unsafe way. Also you seem to be writing an entry for every impression, even if `a.adSlots[impI.ID]` already exists, which seems rather wasteful, especially if this interaction is made thread safe.

wojciech-bialy-wpm: OK, I've removed a.adSlots alltogether. What we used to do with it (request->response Imp matching) is done by looking at externalRequest

hhhjort:
You need to run gofmt on your code to get it to pass validation. The best way is to just run validate.sh, which will run gofmt and also run all the unit tests so you can see and verify that it passes rather than waiting for github to do the same.

spanwp:
@hhhjort Thank you. I fixed it because my colleague is on holiday now.

adapters/sspBC/sspbc.go
```diff
@@ -224,70 +221,74 @@ func (a *adapter) createBannerAd(bid openrtb2.Bid, ext responseExt, request *ope
...
+	for _, size := range imp.Banner.Format {
+		area := size.W * size.H
+		if area > areaMax {
+			areaMax = area
+			impSize = fmt.Sprintf("%dx%d", size.W, size.H)
 		}
 	}
 
-	// default fallback
+	if areaMax == 0 {
```
hhhjort, line 267: Given line 256, I don't think this if statement is necessary? Not that it hurts anything.

spanwp: Thanks. You are right. I wanted to explicitly cover the case when one of the size dimensions is zero. But after time, I fully agree with you it is not necessary.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+		SiteId:  ext.SiteId,
+		SlotId:  ext.SlotId,
+		AdLabel: ext.AdLabel,
+		PubId:   ext.PublisherId,
+		Page:    request.Site.Page,
+		Referer: request.Site.Ref,
+		McAd:    base64.URLEncoding.EncodeToString(mcMarshalled),
+	}
+
+	// Prepare banner html, using template file
```
SyntaxNode, line 240: Nitpick: Template is no longer a file.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+			return nil, err
+		}
+		imp.Ext = impExtJSON
+		// save updated imp
+		request.Imp[i] = imp
+	}
+
+	// update site info (ID, of present, and request domain)
+	if siteID != "" {
+		request.Site.ID = siteID
```
SyntaxNode, line 325: `Site` is shared memory and cannot be directly modified. You need to make a copy of it first and then modify, such as:

```
siteCopy := *request.Site
request.Site = &siteCopy
request.Site.ID = siteID
``` 

This would be detected by our json tests if you add a test file where the `site.id` is empty or different.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	// add domain info
+	if siteURL, err := url.Parse(request.Site.Page); err == nil {
+		request.Site.Domain = siteURL.Hostname()
```
SyntaxNode, line 330: Same issue as the comment above. You might be able to arrange the code such that you only make at most one copy. Please add an accompanying test too.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	// set TEST Flag
+	if isTest == 1 {
+		request.Test = 1
```
SyntaxNode, line 335: This is fine. You only need to copy sub-objects. See our [new adapter docs](https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#makerequests) for more detail.

SyntaxNode (commented):
The majority of my comments revolve around json test coverage and shared memory. You'll need to build our your "supplemental" tests. Please look at other adapters for examples.

adapters/sspBC/sspbc.go
hhhjort, line 325: `Site` is a pointer, so you need to verify that it is not null before accessing any of its fields.

hhhjort:
I noticed that you only have a couple of positive test pages. Would be good to test that the adapter does something reasonable in problem cases, such as the bidder returning an error, and possibly a "bad" request that is still good enough to pass the basic validation that is done on requests.

adapters/sspBC/sspbc.go
hhhjort: That should protect against a nil `Site`. May still not be a bad idea to have a check to guard against future expansion should it be decided to expand to app inventory.

SyntaxNode: True. It's not a bad idea to have that protection in code.

SyntaxNode:
Please be aware that PR https://github.com/prebid/prebid-server/pull/2280 was recently merged. This PR updates the package used for the openrtb2 model. You will need to update this PR to use the `github.com/mxmCherry/openrtb/v15/openrtb2` package.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,406 @@
...
+
+	if externalResponse.StatusCode != http.StatusOK {
+		err := &errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Unexpected status code: %d.", externalResponse.StatusCode),
+		}
+		return nil, []error{err}
+	}
+
+	// unmarshall original request - we will need it to recover imp.Id
+	var requestBody openrtb2.BidRequest
```
hhhjort, line 147: I don't believe that you have to unmarshal externalRequst.Body. The internalRequest is the same pointer that MakeBids() was given, so when you moved Imp.ID to Imp.TagID, that change is still there in the internalRequest you have now.

hhhjort:
It is looking much better now. There do seem to be some merge conflicts to resolve still.

adapters/sspBC/sspbc_test.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderConnectAd, config.Adapter{
```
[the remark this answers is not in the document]
spanwp: Thanks.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+	version        string
+	endpoint       string
+	bannerTemplate *template.Template
+}
+
+// ---------------ADAPTER INTERFACE------------------
+// Builder builds a new instance of the sspBC adapter
+func Builder(bidderName openrtb_ext.BidderName, config config.Adapter) (adapters.Bidder, error) {
+	// HTML template used to create banner ads
+	const bannerHTML = `<html><head><title></title><meta charset="UTF-8"><meta name="viewport" content="width=device-width, initial-scale=1.0"><style> body { background-color: transparent; margin: 0; padding: 0; }</style><script> window.rekid = {{.SiteId}}; window.slot = {{.SlotId}}; window.adlabel = '{{.AdLabel}}'; window.pubid = '{{.PubId}}'; window.wp_sn = 'sspbc_go'; window.page = '{{.Page}}'; window.ref = '{{.Referer}}'; window.mcad = JSON.parse(atob('{{.McAd}}'));</script></head><body><div id="c"></div><script async crossorigin nomodule src="//std.wpcdn.pl/wpjslib/wpjslib-inline.js" id="wpjslib"></script><script async crossorigin type="module" src="//std.wpcdn.pl/wpjslib6/wpjslib-inline.js" id="wpjslib6"></script></body></html>`
```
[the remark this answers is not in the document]
spanwp: Thank you, I did this.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+
+	for i, imp := range request.Imp {
+		// read ext data for the impression
+		var extSSP openrtb_ext.ExtImpSspbc
+		var extBidder adapters.ExtImpBidder
+
+		// Read additional data for this imp.
+		// Errors here do not break the flow for this imp, and are ignored
+		if err := json.Unmarshal(imp.Ext, &extBidder); err == nil {
+			_ = json.Unmarshal(extBidder.Bidder, &extSSP)
```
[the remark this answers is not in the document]
spanwp: Thank you, it is done.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+		}
+
+		// store SiteID
+		if extSSP.SiteId != "" {
```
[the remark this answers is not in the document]
spanwp: Thank you, it is done.

adapters/sspBC/sspbc.go
spanwp: Thanks.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+			siteID = extSSP.SiteId
+		}
+
+		// store test info
+		if extSSP.IsTest != 0 {
```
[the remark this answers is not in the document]
spanwp: Thank you, it is done.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	requestURL, err := url.Parse(a.endpoint)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	// add query parameters to request
+	queryParams := requestURL.Query()
+	queryParams.Add("bdver", a.version) // adapter version
```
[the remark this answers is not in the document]
spanwp: Thanks. You are right. It is done.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+		Uri:    requestURL.String(),
+		Body:   requestJSON,
+	}
+
+	return []*adapters.RequestData{requestData}, nil
+}
+
+func (a *adapter) MakeBids(internalRequest *openrtb2.BidRequest, externalRequest *adapters.RequestData, externalResponse *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if externalResponse.StatusCode == http.StatusNoContent {
+		return nil, nil
```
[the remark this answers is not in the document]
spanwp: Thank you, it is done.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	if externalResponse.StatusCode != http.StatusOK {
+		err := &errortypes.BadServerResponse{
```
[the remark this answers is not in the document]
spanwp: Thank you, it is done.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+			var bidType openrtb_ext.BidType
+
+			/*
+			  Determine bid type
+			  At this moment we only check if bid contains Adm property
+
+			  Later updates will check for video & native data
+			*/
+			if bid.AdM != "" {
+				bidType = openrtb_ext.BidTypeBanner
```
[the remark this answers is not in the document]
spanwp: Thank you, it is done.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,339 @@
...
+		// Banner ad is already formatted
+		return bid.AdM, nil
+	}
+
+	// create McAd payload
+	mcad.Id = request.ID
+	mcad.Seat = seat
+	mcad.SeatBid = make([]openrtb2.SeatBid, 1)
+	mcad.SeatBid[0].Bid = make([]openrtb2.Bid, 1)
+	mcad.SeatBid[0].Bid[0] = bid
```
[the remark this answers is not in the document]
spanwp: Thank you, it is done.

adapters/sspBC/sspbc.go
spanwp: You are right. I fixed it. Thanks.

spanwp:
Thanks a lot for all the comments guys.

mansinahar:
@spanwp Can you please resolve the conflicts on this PR when you have a chance for us to take another look?

spanwp:
@mansinahar Thanks, it is done.

adapters/sspBC/sspbc.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	if incompleteImps > 0 {
+		return requestTypeOneCode
+	}
+
+	return requestTypeStandard
+}
+
+func formatSspBcRequest(a *adapter, request *openrtb2.BidRequest) (*openrtb2.BidRequest, error) {
```
SyntaxNode, line 328: The `a` parameter is not used in this method.

adapters/sspBC/sspbc_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"testing"
+
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderSspBC, config.Adapter{
+		Endpoint: "http://ssp.wp.test/bidder/"})
```
SyntaxNode, line 13: There is no need to use a real endpoint here. You may use a fake endpoint to avoid needing to keep the tests in sync with the real url if it changes in the future.

adapters/sspBC/sspbctest/exemplary/banner-fromtemplate.json
```diff
@@ -0,0 +1,152 @@
...
+            "mockResponse": {
+                "status": 200,
+                "body": {
+                    "cur": "USD",
+                    "id": "test-request",
+                    "seatbid": [
+                        {
+                            "bid": [
+                                {
+                                    "adm": "{\"bunch\":\"237503\",\"capping\":\"\",\"server\":\"s\",\"campaign\":\"c\",\"score\":\"1\",\"creations\":[{\"type\":\"image\",\"src\":\"https:\/\/bdr.wpcdn.pl\/tests\/test-300x250.png\",\"trackers\":{},\"width\":300,\"height\":250}],\"code\":\"1\",\"aabVendors\":false,\"slot\":\"005\",\"redir\":\"https:\/\/www.wp.pl\",\"wpclid\":\"1\"}",
```
SyntaxNode, line 96: There is no need or benefit in using a real full length `adm` value. Consider using a shorter string to improve test readability.

SyntaxNode (commented):
Looking very good. I'm impressed with the improvements made since the PR was first opened. I have just a few nitpick comments, but otherwise looks good to me.

static/bidder-info/sspBC.yaml
```diff
@@ -0,0 +1,12 @@
+maintainer:
+  email: "prebid-dev@grupawp.pl"
```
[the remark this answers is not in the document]
spanwp: I got your email. 
obraz

adapters/sspBC/sspbc.go
spanwp: Thanks.

adapters/sspBC/sspbctest/exemplary/banner-fromtemplate.json
spanwp: Thanks.

adapters/sspBC/sspbc_test.go
spanwp: Thanks. I think it is already a fake URL. Can it remain in this form?

hhhjort (approved):
LGTM

Issue #2267, opened 2022-06-07
Interstitial question


KushnerykPavel:
Hello, I received the next error from prebid server. 
`Jun  7 08:26:54 s3036 sleepservice[3695232]: 2022/06/07 08:26:54 http: panic serving 127.0.0.1:42066: runtime error: invalid memory address or nil pointer dereference
Jun  7 08:26:54 s3036 sleepservice[3695232]: goroutine 9565 [running]:
Jun  7 08:26:54 s3036 sleepservice[3695232]: net/http.(*conn).serve.func1(0xc006b83040)
Jun  7 08:26:54 s3036 sleepservice[3695232]: #011/usr/local/go/src/net/http/server.go:1800 +0x139
Jun  7 08:26:54 s3036 sleepservice[3695232]: panic(0xd9f420, 0x16cecf0)
Jun  7 08:26:54 s3036 sleepservice[3695232]: #011/usr/local/go/src/runtime/panic.go:975 +0x3e3
Jun  7 08:26:54 s3036 sleepservice[3695232]: github.com/prebid/prebid-server/endpoints/openrtb2.processInterstitials(0xc0068fb180, 0xc005ecac00, 0xc0067a8b40)
Jun  7 08:26:54 s3036 sleepservice[3695232]: #011/var/src/github.com/prebid/prebid-server/endpoints/openrtb2/interstitial.go:28 +0x113
Jun  7 08:26:54 s3036 sleepservice[3695232]: github.com/prebid/prebid-server/endpoints/openrtb2.(*endpointDeps).parseRequest(0xc0004462d0, 0xc005ecac00, 0xc0068fb180, 0xc005e62ab0, 0x0, 0x0, 0x0, 0x0, 0x0)
Jun  7 08:26:54 s3036 sleepservice[3695232]: #011/var/src/github.com/prebid/prebid-server/endpoints/openrtb2/auction.go:315 +0x629
Jun  7 08:26:54 s3036 sleepservice[3695232]: github.com/prebid/prebid-server/endpoints/openrtb2.(*endpointDeps).Auction(0xc0004462d0, 0x104e820, 0xc00506f340, 0xc005ecac00, 0x0, 0x0, 0x0)
Jun  7 08:26:54 s3036 sleepservice[3695232]: #011/var/src/github.com/prebid/prebid-server/endpoints/openrtb2/auction.go:154 +0x37a
[...]

PR #2276, opened 2022-06-18
handle nil when device ext does not contain prebid settings for interstitial

KushnerykPavel:
Resolve https://github.com/prebid/prebid-server/issues/2267


VeronikaSolovei9 (approved):
LGTM

AlexBVolcy (approved):
Looks good! Thanks for fixing this.

PR #2277, opened 2022-06-20
Unicorn: Revert limitation of JPY only ( add limitation of accountId )

faithnh:
* Revert limitation of JPY only.
* Return error if accountId is not parsed.
* Fix bug that acccountId set zero in `modifyImps` when not have accountId.

https://github.com/prebid/prebid-server/pull/2265#issuecomment-1158509800


adapters/unicorn/unicorn.go
```diff
@@ -191,8 +174,9 @@ func setSourceExt() json.RawMessage {
 func setExt(request *openrtb2.BidRequest) (json.RawMessage, error) {
 	accountID, err := jsonparser.GetInt(request.Imp[0].Ext, "bidder", "accountId")
 	if err != nil {
-		accountID = 0
+		return nil, fmt.Errorf("accountId field is required")
```
VeronikaSolovei9, line 177: Can you please add coverage for this line?

faithnh: I attemped two pattern.

* accountId is not exists.
* accountId is not integer.

But neither was able to pass this line. I think this is probably because [bidder-params](https://github.com/prebid/prebid-server/blob/master/static/bidder-params/unicorn.json#L19) has an accountId that is integer and required.

Do you have any good ideas?

VeronikaSolovei9 (approved):
Thank you for the update, lgtm

PR #2278, opened 2022-06-21
New Adapter: InfyTV

infytvcode:


mansinahar:
@infytvcode Looks like there's a test failure due to URL mismatch in one of your test cases. Can you please fix this when you have a chance?

adapters/infytv/infytv.go
```diff
@@ -0,0 +1,78 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type InfytvAdapter struct {
```
hhhjort, line 15: You can call this simply "Adapter", the "Infytv" identification is already supplied by the package name. As you have it, referencing your adapter from outside the package would be `infytv.InfytvAdapter` which looks a little redundant.

infytvcode: @hhhjort Thanks for the suggestion. Will update it.

infytvcode:
Documentation Pull request https://github.com/prebid/prebid.github.io/pull/3865

adapters/infytv/params_test.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	for _, p := range invalidParams {
+		if err := validator.Validate(openrtb_ext.BidderInfyTV, json.RawMessage(p)); err == nil {
+			t.Errorf("Schema allowed invalid params: %s", p)
+		}
+	}
+}
+
+var validParams = []string{
```
hhhjort, line 36: You may want to include placementId in some of your tests.

infytvcode: @hhhjort I have added placementId in the test. 
Thanks

hhhjort (dismissed):

adapters/infytv/infytv.go
```diff
@@ -0,0 +1,89 @@
...
+	"net/http"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type Adapter struct {
```
SyntaxNode, line 15: Actually, we would prefer the name `adapter`. The beginning upper case letter tells Go to export this type and there is no need for that. Sorry to contradict with @hhhjort here.

static/bidder-info/infytv.yaml
```diff
@@ -0,0 +1,9 @@
+maintainer:
+  email: "tech+hb@infy.tv"
+capabilities:
```
SyntaxNode, line 3: I would like to confirm InfyTV does not have an IAB Europe GVL vendor id.

adapters/infytv/infytv.go
infytvcode: @SyntaxNode Updated to `adapter`.

static/bidder-info/infytv.yaml
infytvcode: @SyntaxNode Yes, InfyTV does not have an IAB Europe GVL vendor id.

hhhjort (approved):

SyntaxNode (approved):

PR #2283, opened 2022-06-28
Read imp[].ext.tid, fix PBJS 7.3.0 compatibility

pm-nilesh-chate:
Added changes to read and mark `imp[].ext.tid` as special field and not a bidder code.

With https://github.com/prebid/Prebid.js/pull/8591 (https://github.com/prebid/Prebid.js/issues/8543) PBJS 7.3.0 sends `imp[].ext.tid` to PBS where PBS interprets `tid` as bidder code. Hence, blocking PBS adapter from using with PBJS 7.3.0.
Below is the error response from PBS.
```
Invalid request: request.imp[0].ext contains unknown bidder: tid. Did you forget an alias in request.ext.prebid.aliases?
```

### DIFF
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -421,8 +421,8 @@ func createSanitizedImpExt(impExt, impExtPrebid map[string]json.RawMessage) (map
 		sanitizedImpExt[openrtb_ext.GPIDKey] = v
 	}
 
-	if v, exists := impExt[string(openrtb_ext.TID)]; exists {
-		sanitizedImpExt[openrtb_ext.TID] = v
+	if v, exists := impExt[string(openrtb_ext.TIDKey)]; exists {
+		sanitizedImpExt[openrtb_ext.TIDKey] = v
 	}
 
 	return sanitizedImpExt, nil
@@ -455,7 +455,8 @@ func isSpecialField(bidder string) bool {
 		bidder == openrtb_ext.FirstPartyDataExtKey ||
 		bidder == openrtb_ext.SKAdNExtKey ||
 		bidder == openrtb_ext.GPIDKey ||
-		bidder == openrtb_ext.PrebidExtKey || bidder == openrtb_ext.TID
+		bidder == openrtb_ext.PrebidExtKey ||
+		bidder == openrtb_ext.TIDKey
 }
 
 // prepareUser changes req.User so that it's ready for the given bidder.
--- a/exchange/utils.go
+++ b/exchange/utils.go
@@ -421,6 +421,10 @@ func createSanitizedImpExt(impExt, impExtPrebid map[string]json.RawMessage) (map
 		sanitizedImpExt[openrtb_ext.GPIDKey] = v
 	}
 
+	if v, exists := impExt[string(openrtb_ext.TID)]; exists {
+		sanitizedImpExt[openrtb_ext.TID] = v
+	}
+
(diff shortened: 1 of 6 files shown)


exchange/utils.go
```diff
@@ -451,7 +455,7 @@ func isSpecialField(bidder string) bool {
 		bidder == openrtb_ext.FirstPartyDataExtKey ||
 		bidder == openrtb_ext.SKAdNExtKey ||
 		bidder == openrtb_ext.GPIDKey ||
-		bidder == openrtb_ext.PrebidExtKey
+		bidder == openrtb_ext.PrebidExtKey || bidder == openrtb_ext.TID
```
SyntaxNode, line 458: Please add a line break after the `||` to match the existing pattern.

exchange/utils_test.go
```diff
@@ -468,6 +470,12 @@ func TestExtractBidderExts(t *testing.T) {
 			givenImpExtPrebidBidders: map[string]json.RawMessage{"bidderA": bidderAJSON},
 			expected:                 map[string]json.RawMessage{"bidderA": bidderAJSON},
 		},
+		{
+			description:              "imp.ext.tid",
+			givenImpExt:              map[string]json.RawMessage{"bidderA": json.RawMessage(`{"shouldBe":"Ignored"}}`), "tid": json.RawMessage(`{`)},
+			givenImpExtPrebidBidders: map[string]json.RawMessage{"bidderA": bidderAJSON},
+			expected:                 map[string]json.RawMessage{"bidderA": bidderAJSON},
+		},
```
SyntaxNode, line 478: I recommend to remove this test case and add tid to the existing "Special Names Ignored - imp.ext.BIDDER" case.

openrtb_ext/bidders_test.go
```diff
@@ -114,6 +114,7 @@ func TestIsBidderNameReserved(t *testing.T) {
 		{"skADN", true},
 		{"SKADN", true},
 		{"notreserved", false},
+		{"tid", true},
```
SyntaxNode, line 117: Please add a test case with mixed case for "tid" and for upper case to test the `EqualFold` method was used for the comparison. Please also move above the "notreserved" case to leave that last for readability (the only case which expected false).

openrtb_ext/request.go
```diff
@@ -20,6 +20,9 @@ const GPIDKey = "gpid"
 // NativeExchangeSpecificLowerBound defines the lower threshold of exchange specific types for native ads. There is no upper bound.
 const NativeExchangeSpecificLowerBound = 500
 
+// TID reserved for Per-Impression Transactions IDs for Multi-Impression Bid Requests.
+const TID = "tid"
```
SyntaxNode, line 24: Please name `TIDKey` to match the existing pattern and move to just below "GPIDKey" to keep the special keys grouped together.

SyntaxNode (commented):
Thank you for proactively solving this problem. I have a few nitpicks, but overall looks great.

exchange/utils_test.go
pm-nilesh-chate: done

openrtb_ext/bidders_test.go
pm-nilesh-chate: done

SyntaxNode (approved):

VeronikaSolovei9 (approved):

PR #2287, opened 2022-06-29
UOE-7905: Do not allow publishers to bypass GDPR

pm-nilesh-chate:
PBS config needed to test this:
```
gdpr:
  enabled: true
  default_value: true
  host_vendor_id: 76
  timeouts_ms:
    init_vendorlist_fetches: 100000
    active_vendorlist_fetch: 100000
```


pm-nilesh-chate:
wrong upstream

PR #2290, opened 2022-07-05
Stored responses metrics

VeronikaSolovei9:
Implementation for Stored bid responses metrics:
1) Number of requests that have stored bid responses or stored auction responses - new metric
2) Time to fetch stored response and fetching errors. These use new metrics and existing functionality to track it: `RecordStoredDataFetchTime` and `RecordStoredDataError `

### DIFF
--- a/config/stored_requests.go
+++ b/config/stored_requests.go
@@ -19,6 +19,7 @@ const (
 	VideoDataType      DataType = "Video"
 	AMPRequestDataType DataType = "AMP Request"
 	AccountDataType    DataType = "Account"
+	ResponseDataType   DataType = "Response"
 )
 
 // Section returns the config section this type is defined in
@@ -29,6 +30,7 @@ func (dataType DataType) Section() string {
 		VideoDataType:      "stored_video_req",
 		AMPRequestDataType: "stored_amp_req",
 		AccountDataType:    "accounts",
+		ResponseDataType:   "stored_responses",
 	}[dataType]
 }
 
@@ -133,6 +135,7 @@ func resolvedStoredRequestsConfig(cfg *Configuration) {
 	cfg.StoredVideo.dataType = VideoDataType
 	cfg.CategoryMapping.dataType = CategoryDataType
 	cfg.Accounts.dataType = AccountDataType
+	cfg.StoredResponses.dataType = ResponseDataType
 }
 
 func (cfg *StoredRequests) validate(errs []error) []error {
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -248,6 +248,10 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 	e.me.RecordRequestPrivacy(privacyLabels)
 
+	if len(r.StoredAuctionResponses) > 0 || len(r.StoredBidResponses) > 0 {
+		e.me.RecordStoredResponse(r.PubID)
+	}
+
 	// If we need to cache bids, then it will take some time to call prebid cache.
 	// We should reduce the amount of time the bidders have, to compensate.
 	auctionCtx, cancel := e.makeAuctionContext(ctx, cacheInstructions.cacheBids)
--- a/metrics/go_metrics.go
+++ b/metrics/go_metrics.go
@@ -454,7 +454,7 @@ func (me *Metrics) RecordDebugRequest(debugEnabled bool, pubID string) {
 
 func (me *Metrics) RecordStoredResponse(pubId string) {
 	me.StoredResponsesMeter.Mark(1)
-	if pubId != PublisherUnknown {
+	if pubId != PublisherUnknown && !me.MetricsDisabled.AccountStoredResponses {
 		me.getAccountMetrics(pubId).storedResponsesMeter.Mark(1)
 	}
 }
--- a/metrics/go_metrics.go
+++ b/metrics/go_metrics.go
@@ -32,6 +32,7 @@ type Metrics struct {
 	AccountCacheMeter              map[CacheResult]metrics.Meter
 	DNSLookupTimer                 metrics.Timer
 	TLSHandshakeTimer              metrics.Timer
+	StoredResponsesMeter           metrics.Meter
 
 	// Metrics for OpenRTB requests specifically. So we can track what % of RequestsMeter are OpenRTB
 	// and know when legacy requests have been abandoned.
@@ -99,7 +100,8 @@ type accountMetrics struct {
 	bidsReceivedMeter metrics.Meter
 	priceHistogram    metrics.Histogram
 	// store account by adapter metrics. Type is map[PBSBidder.BidderCode]
-	adapterMetrics map[openrtb_ext.BidderName]*AdapterMetrics
+	adapterMetrics       map[openrtb_ext.BidderName]*AdapterMetrics
+	storedResponsesMeter metrics.Meter
 }
 
 // NewBlankMetrics creates a new Metrics object with all blank metrics object. This may also be useful for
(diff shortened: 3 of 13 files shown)


AlexBVolcy:
Could you add a description for what this PR is trying to accomplish?

guscarreon (commented):
Looks good. Left couple of questions

exchange/exchange.go
```diff
@@ -248,6 +248,10 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 	e.me.RecordRequestPrivacy(privacyLabels)
 
+	if len(r.StoredAuctionResponses) > 0 || len(r.StoredBidResponses) > 0 {
+		e.me.RecordStoredResponse(r.PubID)
```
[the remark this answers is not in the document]
VeronikaSolovei9: They are different. We discussed this with a team and decided there is no value in 2 metrics for each kind of stored response, instead we just want to know how many requests use any kind of stored responses.

exchange/exchange.go
```diff
@@ -248,6 +248,10 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+	}
+
```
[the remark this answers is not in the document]
VeronikaSolovei9: this is funny, because this is what I did initially and then modified to what it is now.
My only argument is that i prefer to keep it all in one place rather than spread it in code. 
I'm ok with any approaches, what do you think?

metrics/go_metrics_test.go
```diff
@@ -739,6 +740,38 @@ func TestRecordSyncerSet(t *testing.T) {
 	assert.Equal(t, m.SyncerSetsMeter["foo"][SyncerSetUidCleared].Count(), int64(1))
 }
 
+func TestStoredResponses(t *testing.T) {
```
AlexBVolcy, line 743: Is there a test case that you can create where only one of the two metrics gets incremented? And is there a test case you can create where neither are incremented?

metrics/prometheus/prometheus_test.go
```diff
@@ -1626,3 +1641,34 @@ func TestRecordAdapterGDPRRequestBlocked(t *testing.T) {
 			adapterLabel: string(openrtb_ext.BidderAppnexus),
 		})
 }
+
+func TestStoredResponsesMetric(t *testing.T) {
```
AlexBVolcy, line 1645: Is there a test case you could write where neither metric is incremented?

exchange/exchange.go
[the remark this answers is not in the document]
guscarreon: I'm ok with it.

metrics/go_metrics_test.go
VeronikaSolovei9: `test case that you can create where only one of the two metrics gets incremented` - I think this is added, this is second test case. 
`where neither are incremented` - with current logic I don't see a way to do this. 
I'll be happy to add other test cases, but I cannot really come up with input data.
I modified a description for second test case.

metrics/prometheus/prometheus_test.go
VeronikaSolovei9: Logic in Prometheus function is the same. I don't know if I can add more TC for this.

metrics/go_metrics.go
```diff
@@ -447,6 +452,13 @@ func (me *Metrics) RecordDebugRequest(debugEnabled bool, pubID string) {
 	}
 }
 
+func (me *Metrics) RecordStoredResponse(pubId string) {
+	me.StoredResponsesMeter.Mark(1)
+	if pubId != PublisherUnknown {
+		me.getAccountMetrics(pubId).storedResponsesMeter.Mark(1)
+	}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good point Gus. 
I followed a suggestion to add this kind of separation. We should discuss it with the team.

AlexBVolcy (approved):
LGTM

config/stored_requests.go
```diff
@@ -29,6 +30,7 @@ func (dataType DataType) Section() string {
 		VideoDataType:      "stored_video_req",
 		AMPRequestDataType: "stored_amp_req",
 		AccountDataType:    "accounts",
+		ResponseDataType:   "stored_responses",
```
SyntaxNode, line 33: Nitpick: Should this be the singular `stored_response`? What doe PBS-Java use for their metrics?

metrics/prometheus/prometheus.go
```diff
@@ -51,6 +51,9 @@ type Metrics struct {
 	privacyCOPPA                 *prometheus.CounterVec
 	privacyLMT                   *prometheus.CounterVec
 	privacyTCF                   *prometheus.CounterVec
+	storedResponses              prometheus.Counter
+	storedResponsesFetchTimer    *prometheus.HistogramVec
+	storedResponsesErrors        *prometheus.CounterVec
```
SyntaxNode, line 56: Does go_metrics also have the fetch timer and errors metrics? Ideally, we should keep parity between the two.

metrics/prometheus/prometheus.go
```diff
@@ -311,6 +315,21 @@ func NewMetrics(cfg config.PrometheusMetrics, disabledMetrics config.DisabledMet
 			[]string{adapterLabel})
 	}
 
+	metrics.storedResponsesFetchTimer = newHistogramVec(cfg, reg,
+		"stored_response_fetch_time_seconds",
+		"Seconds to fetch stored responses labeled by fetch type",
+		[]string{storedDataFetchTypeLabel},
+		standardTimeBuckets)
```
SyntaxNode, line 322: Time buckets are very expensive (14 metrics I think) but since there are only 2 fetch types, I think we're good here.

metrics/prometheus/prometheus.go
```diff
@@ -499,6 +523,16 @@ func (m *Metrics) RecordDebugRequest(debugEnabled bool, pubID string) {
...
 
+func (m *Metrics) RecordStoredResponse(pubId string) {
+	m.storedResponses.Inc()
+	if !m.metricsDisabled.AccountStoredResponses && pubId != metrics.PublisherUnknown {
+		m.accountStoredResponses.With(prometheus.Labels{
+			accountLabel: pubId,
+		}).Inc()
+	}
+
+}
```
SyntaxNode, line 534: Nitpick: Please remove empty lines.

metrics/prometheus/prometheus.go
VeronikaSolovei9: Yes, I believe these two from go_metrics
```
StoredDataFetchTimer           map[StoredDataType]map[StoredDataFetchType]metrics.Timer
StoredDataErrorMeter           map[StoredDataType]map[StoredDataError]metrics.Meter
```
are metrics similar to 
```
storedResponsesFetchTimer    *prometheus.HistogramVec
storedResponsesErrors        *prometheus.CounterVec
```
from prometheus.go. 
In both cases they work using existing functionality to track stored data. 

go_metrics -> 
StoredResponsesMeter  metrics.Meter 
is similar to 
prometheus.go -> storedResponses  prometheus.Counter

config/stored_requests.go
VeronikaSolovei9: Discussed offline. Decided not to modify due to other metrics names fomat

metrics/prometheus/prometheus.go
VeronikaSolovei9: Removed

guscarreon (approved):
LGTM

AlexBVolcy (approved):

PR #2293, opened 2022-07-07
 AMP targeting query string

VeronikaSolovei9:
Implementation for https://github.com/prebid/prebid-server/issues/1727

### DIFF
--- a/amp/parse.go
+++ b/amp/parse.go
@@ -20,6 +20,7 @@ type Params struct {
 	Slot            string
 	StoredRequestID string
 	Timeout         *uint64
+	Targeting       string
 }
 
 // Size defines size information of an AMP request.
@@ -56,6 +57,7 @@ func ParseParams(httpRequest *http.Request) (Params, error) {
 		Slot:            query.Get("slot"),
 		StoredRequestID: tagID,
 		Timeout:         parseIntPtr(query.Get("timeout")),
+		Targeting:       query.Get("targeting"),
 	}
 	return params, nil
 }
--- a/amp/parse_test.go
+++ b/amp/parse_test.go
@@ -25,7 +25,7 @@ func TestParseParams(t *testing.T) {
 		{
 			description: "All Fields",
 			query: "tag_id=anyTagID&account=anyAccount&curl=anyCurl&consent_string=anyConsent&debug=1&__amp_source_origin=anyOrigin" +
-				"&slot=anySlot&timeout=42&h=1&w=2&oh=3&ow=4&ms=10x11,12x13",
+				"&slot=anySlot&timeout=42&h=1&w=2&oh=3&ow=4&ms=10x11,12x13&targeting=%7B%22gam-key1%22%3A%22val1%22%2C%22gam-key2%22%3A%22val2%22%7D",
 			expectedParams: Params{
 				Account:         "anyAccount",
 				CanonicalURL:    "anyCurl",
@@ -44,6 +44,7 @@ func TestParseParams(t *testing.T) {
 						{W: 10, H: 11}, {W: 12, H: 13},
 					},
 				},
+				Targeting: `{"gam-key1":"val1","gam-key2":"val2"}`,
 			},
 		},
 		{
--- a/amp/parse_test.go
+++ b/amp/parse_test.go
@@ -24,6 +24,7 @@ func TestParseParams(t *testing.T) {
 		},
 		{
 			description: "All Fields",
+			// targeting data is encoded string that looks like this: {"gam-key1":"val1","gam-key2":"val2"}
 			query: "tag_id=anyTagID&account=anyAccount&curl=anyCurl&consent_string=anyConsent&debug=1&__amp_source_origin=anyOrigin" +
 				"&slot=anySlot&timeout=42&h=1&w=2&oh=3&ow=4&ms=10x11,12x13&targeting=%7B%22gam-key1%22%3A%22val1%22%2C%22gam-key2%22%3A%22val2%22%7D",
 			expectedParams: Params{
--- a/endpoints/openrtb2/amp_auction.go
+++ b/endpoints/openrtb2/amp_auction.go
@@ -453,6 +453,44 @@ func (deps *endpointDeps) overrideWithParams(ampParams amp.Params, req *openrtb2
 		req.TMax = int64(*ampParams.Timeout) - deps.cfg.AMPTimeoutAdjustment
 	}
 
+	if len(ampParams.Targeting) > 0 {
+		err := setTargeting(req, ampParams.Targeting)
+		if err != nil {
+			return []error{err}
+		}
+	}
+
+	return nil
+}
+
+func setTargeting(req *openrtb2.BidRequest, targeting string) error {
+	var targetingData map[string]interface{}
+	if err := json.Unmarshal([]byte(targeting), &targetingData); err != nil {
+		return fmt.Errorf("unable to unmarshal targeting data: %s", err.Error())
+	}
+	_, _, offset, err := jsonparser.Get(req.Imp[0].Ext, "data")
+	if err != nil && err != jsonparser.KeyPathNotFoundError {
+		return errors.New("error parsing imp ext")
+	}
+	if offset != -1 {
+		return errors.New("field 'data' already present in imp ext")
+	}
+
+	var impExt map[string]interface{}
+	if len(req.Imp[0].Ext) > 0 {
+		if err := json.Unmarshal(req.Imp[0].Ext, &impExt); err != nil {
+			return fmt.Errorf("unable to unmarshal imp.ext: %s", err.Error())
+		}
+	} else {
+		impExt = make(map[string]interface{})
+	}
+
+	impExt["data"] = targetingData
+	newImpExt, err := json.Marshal(impExt)
+	if err != nil {
+		return fmt.Errorf("unable to marshal imp.ext: %s", err.Error())
+	}
+	req.Imp[0].Ext = newImpExt
 	return nil
 }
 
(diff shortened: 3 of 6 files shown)


endpoints/openrtb2/amp_auction.go
```diff
@@ -453,6 +453,44 @@ func (deps *endpointDeps) overrideWithParams(ampParams amp.Params, req *openrtb2
 		req.TMax = int64(*ampParams.Timeout) - deps.cfg.AMPTimeoutAdjustment
 	}
 
+	if len(ampParams.Targeting) > 0 {
+		err := setTargeting(req, ampParams.Targeting)
+		if err != nil {
+			return []error{err}
+		}
```
SyntaxNode, line 460: Nitpick: Can be written a bit more condensed:

```
if err := setTargeting(req, ampParams.Targeting); err != nil {
	return []error{err}
}
```

endpoints/openrtb2/amp_auction.go
```diff
@@ -453,6 +453,44 @@ func (deps *endpointDeps) overrideWithParams(ampParams amp.Params, req *openrtb2
...
+		impExt = make(map[string]interface{})
+	}
+
+	impExt["data"] = targetingData
+	newImpExt, err := json.Marshal(impExt)
+	if err != nil {
+		return fmt.Errorf("unable to marshal imp.ext: %s", err.Error())
+	}
+	req.Imp[0].Ext = newImpExt
 	return nil
```
SyntaxNode, line 494: IMHO the flow of this function is a bit hard for me to follow. I see you already group sections of the logic, so perhaps small "header" comments to help guide the reader?

endpoints/openrtb2/amp_auction.go
```diff
@@ -453,6 +453,44 @@ func (deps *endpointDeps) overrideWithParams(ampParams amp.Params, req *openrtb2
...
+	if err := json.Unmarshal([]byte(targeting), &targetingData); err != nil {
+		return fmt.Errorf("unable to unmarshal targeting data: %s", err.Error())
+	}
+	_, _, offset, err := jsonparser.Get(req.Imp[0].Ext, "data")
+	if err != nil && err != jsonparser.KeyPathNotFoundError {
+		return errors.New("error parsing imp ext")
+	}
+	if offset != -1 {
+		return errors.New("field 'data' already present in imp ext")
+	}
```
SyntaxNode, line 477: I don't think it's worth it to use jsonparser to check for existence of the `data` field when the very next thing we do is unmarshal it. Consider moving the "already present" check after unmarshaling the map.

amp/parse.go
```diff
@@ -56,6 +57,7 @@ func ParseParams(httpRequest *http.Request) (Params, error) {
 		Slot:            query.Get("slot"),
 		StoredRequestID: tagID,
 		Timeout:         parseIntPtr(query.Get("timeout")),
+		Targeting:       query.Get("targeting"),
```
bsardo, line 60: Can you add a non-empty targeting string to the `amp/parse_test.go#TestParseParams` test case `description: "All Fields"`?

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1449,3 +1449,107 @@ func TestAmpAuctionResponseHeaders(t *testing.T) {
...
+			errorMessage:   "",
+		},
+		{
+			description:    "imp ext has data, valid targeting data",
+			bidRequest:     openrtb2.BidRequest{Imp: []openrtb2.Imp{{Ext: []byte(`{"data":{"placementId":123}}`)}}},
+			targeting:      `{"gam-key1":"val1", "gam-key2":"val2"}`,
+			expectedImpExt: ``,
+			wantError:      true,
+			errorMessage:   "field 'data' already present in imp ext",
+		},
```
bsardo, line 1525: I have limited knowledge of the targeting functionality. However, it seems odd that this scenario would result in an error. My guess is that any key-value pairs specified in the request should be merged into `imp.ext.data` if it already exists, with the values specified as a query parameter taking precedence if the value is already defined in the stored request. I don't see anything in the related issue that specifies what should happen in this scenario. Perhaps we should see what PBS-Java does and check in with Bret?

bsardo (commented):
Just a couple of additional comments. I agree with @SyntaxNode's suggestions as well.

endpoints/openrtb2/amp_auction_test.go
SyntaxNode: I had the same question, just within the code instead of the tests. :)

endpoints/openrtb2/amp_auction.go
```diff
@@ -453,6 +453,44 @@ func (deps *endpointDeps) overrideWithParams(ampParams amp.Params, req *openrtb2
...
+	var targetingData map[string]interface{}
+	if err := json.Unmarshal([]byte(targeting), &targetingData); err != nil {
+		return fmt.Errorf("unable to unmarshal targeting data: %s", err.Error())
+	}
+	_, _, offset, err := jsonparser.Get(req.Imp[0].Ext, "data")
+	if err != nil && err != jsonparser.KeyPathNotFoundError {
+		return errors.New("error parsing imp ext")
+	}
+	if offset != -1 {
+		return errors.New("field 'data' already present in imp ext")
```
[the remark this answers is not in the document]
VeronikaSolovei9: Confirmed with Bret. Replaced to merge.

endpoints/openrtb2/amp_auction_test.go
VeronikaSolovei9: Confirmed with Bret and replaced to merge.

amp/parse.go
VeronikaSolovei9: Good point, added!

exchange/utils.go
```diff
@@ -805,3 +805,11 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
 	}
 	return allBidderRequests
 }
+
+func BuildExtData(data []byte) []byte {
```
hhhjort, line 809: This function name seems a little vague. Perhaps WrapJSONInData() as this is not specific to .ext fields, it just places the argument in a JSON `data` object.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1449,3 +1449,123 @@ func TestAmpAuctionResponseHeaders(t *testing.T) {
...
+			wantError:      false,
+			errorMessage:   "",
+		},
+		{
+			description:    "imp ext has invalid format, valid targeting data",
+			bidRequest:     openrtb2.BidRequest{Imp: []openrtb2.Imp{{Ext: []byte(`{123:{}`)}}},
+			targeting:      `{"gam-key1":"val1", "gam-key2":"val2"}`,
+			expectedImpExt: ``,
+			wantError:      true,
+			errorMessage:   "unable to merge imp.ext with targeting data, check targeting data is correct: Invalid JSON Document",
```
mansinahar, line 1548: In case of AMP, if we do have an invalid `imp.ext`, this error message seems to be a little misleading as it suggests to look at the targeting data being passed in. However, I am assuming in case of amp if the imp.ext was invalid, we'd return an error earlier while fetching the stored request itself, right?

amp/parse_test.go
```diff
@@ -25,7 +25,7 @@ func TestParseParams(t *testing.T) {
 		{
 			description: "All Fields",
 			query: "tag_id=anyTagID&account=anyAccount&curl=anyCurl&consent_string=anyConsent&debug=1&__amp_source_origin=anyOrigin" +
-				"&slot=anySlot&timeout=42&h=1&w=2&oh=3&ow=4&ms=10x11,12x13",
+				"&slot=anySlot&timeout=42&h=1&w=2&oh=3&ow=4&ms=10x11,12x13&targeting=%7B%22gam-key1%22%3A%22val1%22%2C%22gam-key2%22%3A%22val2%22%7D",
```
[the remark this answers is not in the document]
VeronikaSolovei9: I checked the test and it will require a lot of refactoring to implement this, including parts I didn't touch. 
I still think this is not necessary, because it shows exactly how input data will look like. This is useful for future reference. Expected result specified in the `expectedParams` one line below.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1449,3 +1449,123 @@ func TestAmpAuctionResponseHeaders(t *testing.T) {
...
+		empty_fetcher.EmptyFetcher{},
+		&config.Configuration{MaxRequestSize: maxSize},
+		&metricsConfig.NilMetricsEngine{},
+		analyticsConf.NewPBSAnalytics(&config.Analytics{}),
+		nil,
+		nil,
+		openrtb_ext.BuildBidderMap(),
+		empty_fetcher.EmptyFetcher{},
+	)
+	request, err := http.NewRequest("GET", "/openrtb2/auction/amp?tag_id=1&targeting=%7B%22gam-key1%22%3A%20%22val1%22%2C%22gam-key2%22%3A%20%22val2%22%7D", nil)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Modified. It makes sense here.

endpoints/openrtb2/amp_auction_test.go
VeronikaSolovei9: Right, is json in stored request is invalid, `parseAmpRequest` will throw an error while unmarshalling the request

mansinahar (approved):

amp/parse_test.go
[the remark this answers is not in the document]
mansinahar: Yeah, I see what you mean. We will have to store them as a map and then add them inside the for loop one by one. Do you think it might be helpful to add a comment with the decoded query string on the line above to make it easier to read? Have approved the PR anyway since it LGTM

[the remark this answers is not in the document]
VeronikaSolovei9: Thank you Mansi, I added comment.

PR #2294, opened 2022-07-07
Kargo Adapter: Add userSync to bidder-info

jsadwith:
Adds userSync details to Kargo bidder-info


jsadwith:
@SyntaxNode should I replicate this to the `prebid-server-java` repo, or is that handled by Prebid core devs?

jsadwith:
Thanks @SyntaxNode - it's easy enough, so I'll open a PR.

PR #2302, opened 2022-07-13
New Adapter: Oftmedia (alias of adtelligent)

GeneGenie:
Adding new adapter alias

Doc: https://github.com/prebid/prebid-server/pull/2302

### DIFF
--- a/static/bidder-info/oftmedia.yaml
+++ b/static/bidder-info/oftmedia.yaml
@@ -1,0 +1,17 @@
+maintainer:
+  email: "alejo@152media.com"
+gvlVendorID: 1111
+capabilities:
+  app:
+    mediaTypes:
+      - banner
+      - video
+  site:
+    mediaTypes:
+      - banner
+      - video
+userSync:
+  # oftmedia supports user syncing, but requires configuration by the host. contact this
+  # bidder directly at the email address in this file to ask about enabling user sync.
+  supports:
+    - iframe
--- a/static/bidder-info/oftmedia.yaml
+++ b/static/bidder-info/oftmedia.yaml
@@ -1,5 +1,5 @@
 maintainer:
-  email: "alejo@152media.com"
+  email: "admin@152media.com"
 gvlVendorID: 1111
 capabilities:
   app:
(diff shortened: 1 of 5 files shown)


static/bidder-info/oftmedia.yaml
```diff
@@ -0,0 +1,17 @@
+maintainer:
+  email: "alejo@152media.com"
```
SyntaxNode, line 2: Please use a group email address rather than an individual.

AlexBVolcy:
Looks good to me, just waiting on e-mail address update before approval

static/bidder-info/oftmedia.yaml
GeneGenie: done

SyntaxNode:
@GeneGenie Thank you for changing the email. Last request: Could you please update the [adapter docs ](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/oftmedia.md) to indicate the PBS adapter is an alias of Adtelligent?

Looks good otherwise.

GeneGenie:
> @GeneGenie Thank you for changing the email. Last request: Could you please update the [adapter docs ](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/oftmedia.md) to indicate the PBS adapter is an alias of Adtelligent?
> 
> Looks good otherwise.

The only question - how to make this indication. Coz client module, aliasing another bidAdapter code.

AlexBVolcy:
@GeneGenie I think you would just add an `aliasCode: alias_name` section to your adapter docs, but I think you've already done that correctly. @SyntaxNode can you confirm? 

I looked at another adapter doc that is an alias, and they followed that same format. https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/janet.md

SyntaxNode:
This was asked in Prebid Slack and the answer given is:
> You'll need to have two versions: oftmedia.md and oftmedia_server.md.

GeneGenie:
Added doc

SyntaxNode (approved):

AlexBVolcy (approved):

PR #2306, opened 2022-07-14
Adnuntius Bid Adapter: Read dimensions from request

mikael-lundin:
Previously we could not override the dimensions in the request. we used to setup the sizes inside of the ad server. Now we allow the client to override the set sizes for the request. This PR is just adding that functionality in prebid server

Also changes were made to allow users to pass no cookie in the request to tell Adnuntius ad server not to set cookies in the browser. Updated documentation can be found here:

https://github.com/prebid/prebid.github.io/pull/3905


VeronikaSolovei9 (approved):
LGTM

adapters/adnuntius/adnuntius.go
```diff
@@ -109,13 +121,41 @@ func makeEndpointUrl(ortbRequest openrtb2.BidRequest, a *adapter) (string, []err
...
+		}
+
+		return sizes
+	}
+
+	if imp.Banner.W != nil && imp.Banner.H != nil {
+		size := make([][]int64, 1)
+		size[0] = []int64{*imp.Banner.W, *imp.Banner.H}
+		return size
+	}
```
AlexBVolcy, line 154: I noticed that your tests don't seem to cover this code block, could you update the tests so that this is covered?

adapters/adnuntius/adnuntius.go
```diff
@@ -109,13 +121,41 @@ func makeEndpointUrl(ortbRequest openrtb2.BidRequest, a *adapter) (string, []err
...
 	q.Set("tzo", fmt.Sprint(tzo))
 	q.Set("format", "json")
 
 	url := a.endpoint + "?" + q.Encode()
 	return url, nil
 }
 
+func getImpSizes(imp openrtb2.Imp) [][]int64 {
+	if imp.Banner == nil {
+		return nil
```
AlexBVolcy, line 138: Could you add a test where imp.Banner is nil?

AlexBVolcy (commented):
Looks good, just a question about test coverage

mikael-lundin:
Sure, will adf this:) Thanks

adapters/adnuntius/adnuntius.go
mikael-lundin: Added test for this! :)

adapters/adnuntius/adnuntius.go
mikael-lundin: Added test for this as well,

adapters/adnuntius/adnuntius.go
VeronikaSolovei9: Thank you for adding tests. 
This part of code is still not covered even after you added a test with impression without banner. Test exits on Line 175, there is another check if banner is nil.  Can you remove this line here? It's redundant.

mikael-lundin: Myt bad, should be fixed with the latest pull request.

AlexBVolcy (approved):

VeronikaSolovei9 (approved):

Issue #2186, opened 2022-03-17
Support server-side configuration for client ad units


YuriyVelichkoPI:
## Motivation

The deployment cycle of the mobile SDK is pretty long and could be even infinite. To supply any changes to the user’s app we should:

- Release a new version of the SDK
- The publisher should update the SDK version 
- The publisher should release the new version of the application
- The User should update the application to the latest version

If the user doesn’t update the app it will still use the old SDK version, so changes may never reach a part of the audience. 

However, it’s often needed to make some changes on the user side ASAP in order to fix some behavior or improve UX. It is possible to do this via the server-side settings for the client (SDK). 

Using the server-side configuration publishers will be able to control the SDK properties remotely.

## Background

In the nearest releases, we are going to provide publishers the ability to customize some display properties e.g: 

- Video asset selection policy 
- Application Muted 
- Close Icon 
- Skip Option 
- Skip Delay 

The list of customized properties will grow eventually. 

From the SDK perspective, the best way to support such customization is to receive configuration properties from the PBS and apply them at runtime. So publishers can rollout or rollback changes on the flight. 

## Objectives

The publisher should be able to store the properties in the format of a JSON object with custom key-value pairs on the PBS. For instance:

```
{
    skip_offset_sec: 10 
}
```

The SDK should be able to request these properties via an existing or new endpoint. 

### Put properties into the stored requests

The configuration properties could be added directly to the stored requests. With respective paths `ext.prebid.ad_unit_config` for stored impression, and `ext.prebid.sdk_config` for stored requests. 

### Put properties into the separated stored configs

The configuration properties could be added as separated stored configs. Like stored requests. 

### Return configs in the bid response

The fields `bid.ext.prebid.ad_unit_config` and `bid.ext.prebid.sdk_config` could be used for providing the configuration properties. 

In order not to send configs in each bid response, PBS may require the additional flag in the bid request e.g. `includeclientconfig`. If PBS receives the bid request with `includeclientconfig` it should attach the configs for the stored request and stored impression to the bid response. If there is no flag `includeclientconfig` in the request, configs should be ignored. SDK in its own turn may request configs once per hour, for example. Or it also can be customizable via `ext.prebid.sdk_config`.

### Introduce a new endpoint 

`GET /stored_config` with config ID can return respective configuration properties.

bretg:
Suggestion - how about two generic pass-throughs:

[...]

PR #2308, opened 2022-07-15
Add Request Level and Imp Level Passthrough Value to Bid Response

AlexBVolcy:
This PR addresses the issue outlined here https://github.com/prebid/prebid-server/issues/2186

The goal of the PR is to allow for a `passthrough` value to be defined at `bidrequest.ext.prebid.passthrough` and for that value to be added to the bid response at `bidresponse.ext.prebid.passthrough`

The other goal is to allow another `passthrough` value to be defined at the imp level in the request at `bidrequest.imp.ext.prebid.passthrough` and have that value be added to the bid response at `bidresponse.seatbid[].bid[].ext.prebid.passthrough ` where the `impId` of the `passthrough` value matches the `impId` of the `bid`

To main way we accomplish the first goal is to alter `makeExtBidResponse()` which is called during the `HoldAuction()` flow. We alter the `ExtRequestPrebid` struct to include the `passthrough` value of type `json.RawMessage`, and then we just pass the `requestExt.Prebid` object to the `makeExtBidResponse()` function and apply the `passthrough` value to the `bidresponseExt.Prebid `struct. 

To accomplish the imp level passthrough implementation, we just have to update the `processStoredRequests()` function, which parses through `ImpExtInfo` to build an `ImpExtInfoMap`, which maps the `impId` to the `ExtInfo` tied to that imp. We first update the `ExtImpPrebid` struct to include the `passthrough` value of type `json.RawMessage`, and when we get to the part of the function where the `ImpExtInfoMap` is being updated, we just also include the `passthrough` value in that mapping, so that we can use it later on when building the response.

So when the response is being built, `makeBidExtJSON()` is called, and here is where `ImpExtInfoMap` is used. We simply take the `ext.Prebid` object that is being populated, and add the `passthrough` value to the object using the map, `ImpExtInfoMap[impId].passthrough`.

Unit tests and end to end tests were updated or added to accompany the changes through `exchange_test.go 
`


endpoints/openrtb2/sample-requests/invalid-whole/imp-empty-object.json
```diff
@@ -10,5 +10,5 @@
     }
   },
   "expectedReturnCode": 400,
-  "expectedErrorMessage": "Invalid request: request.imp[0] missing required field: \"id\"\n"
+  "expectedErrorMessage": "Invalid request: Key path not found"
```
bsardo, line 13: Why is this error message changing? The new message hides the reason the request failed. Can we keep this the way it was?

exchange/exchangetest/passthrough.json
```diff
@@ -0,0 +1,122 @@
...
+                    "prebid": {
+                        "passthrough": {
+                            "passthrough_val": 20
+                        }
+                    }
+                }
+            }],
+            "ext": {
+                "prebid": {
+                    "passthrough": {"passthrough_val":20}
```
bsardo, line 27: Perhaps the two occurrences of `passthrough_val` in the incoming request (on lines 20 and 27) should have different values so we can be sure they are being copied to the right places in the outgoing request and response?

exchange/exchange_test.go
```diff
@@ -2215,15 +2216,31 @@ func runSpec(t *testing.T, filename string, spec *exchangeSpec) {
...
+	// Passthrough JSON Testing
+	var passthrough json.RawMessage
+	var err error
+	impExtInfoMap := make(map[string]ImpExtInfo)
+	if spec.PassthroughFlag {
+		passthrough, err = getPassthrough(&openrtb_ext.RequestWrapper{BidRequest: &spec.IncomingRequest.OrtbRequest})
+		if err != nil {
+			t.Errorf("%s: Exchange returned an unexpected error. Got %s", filename, err.Error())
+		}
+		impExtInfoMap["my-imp-id"] = ImpExtInfo{Passthrough: passthrough}
```
bsardo, line 2228: I think we should shy away from introducing hidden requirements in our test framework to get a particular piece of our test framework to function. As a result, in order to make the pass through feature more dynamic, I suggest removing the `"my-imp-id"` dependency for the pass through feature and instead read the imp id from `spec.IncomingRequest.OrtbRequest` and use that as the index into `impExtInfoMap`.

exchange/exchangetest/passthrough.json
```diff
@@ -0,0 +1,122 @@
...
+                        }
+                    }
+                }]
+            }]
+        },
+        "ext": {
+            "prebid": {
+                "passthrough": {"passthrough_val":20}
+            }
+        }
```
bsardo, line 120: The test framework doesn't appear to be attempting to validate this portion of the response. If I remove this `ext` block the test still passes.

endpoints/openrtb2/sample-requests/invalid-whole/imp-empty-object.json
bsardo: The team discussed this offline and decided that it would be best if we check if the error is a `KeyPathNotFoundError`, in which case we would generate an error with the message `Invalid request: request.imp[0] missing required field: \"id\"\n"`.

exchange/exchangetest/passthrough.json
bsardo: This still appears to be a problem but I think my suggested fix here should fix it: https://github.com/prebid/prebid-server/pull/2308#discussion_r930549815

endpoints/openrtb2/auction.go
```diff
@@ -1711,9 +1706,24 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
 
+func setImpExtInfoMap(imp json.RawMessage, impData ImpExtPrebidData, impExtInfoMap map[string]exchange.ImpExtInfo) error {
+	impId, err := jsonparser.GetString(imp, "id")
+	if err != nil {
+		if err.Error() == "Key path not found" {
+			return errors.New("request.imp[0] missing required field: \"id\"\n")
+		}
+		return err
+	}
+	echoVideoAttributes := false
```
mansinahar, line 1717: I noticed that this part of reading and setting `echoVideoAttributes` in the `ImpExtInfo` was only done for imps that had a stored Imp associated with it, in the old code but, now we are setting this for all imps regardless of whether it has a stored imp associated with it or not. Was this something that was discovered missing for imps that didn't have a stored imp associated with it, when you touched this code?

endpoints/openrtb2/auction.go
```diff
@@ -1711,9 +1706,24 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
...
+	if impData.ImpExtPrebid.Options != nil {
+		echoVideoAttributes = impData.ImpExtPrebid.Options.EchoVideoAttrs
+	}
+	impExtInfoMap[impId] = exchange.ImpExtInfo{EchoVideoAttrs: echoVideoAttributes, Passthrough: impData.ImpExtPrebid.Passthrough}
```
mansinahar, line 1721: Before, we were also setting `StoredImp` value in the `ImpExtInfo` struct. Is that not needed anymore?

endpoints/openrtb2/auction.go
```diff
@@ -1711,9 +1706,24 @@ func (deps *endpointDeps) processStoredRequests(ctx context.Context, requestJson
 	return resolvedRequest, impExtInfoMap, nil
 }
 
+func setImpExtInfoMap(imp json.RawMessage, impData ImpExtPrebidData, impExtInfoMap map[string]exchange.ImpExtInfo) error {
+	impId, err := jsonparser.GetString(imp, "id")
+	if err != nil {
+		if err.Error() == "Key path not found" {
+			return errors.New("request.imp[0] missing required field: \"id\"\n")
```
mansinahar, line 1713: We shouldn't be hardcoding `request.imp[0]` in the error message here. We should pass in the imp element number in the imp array to this function and add that to the error message

exchange/exchange.go
```diff
@@ -1056,20 +1063,22 @@ func makeBidExtJSON(ext json.RawMessage, prebid *openrtb_ext.ExtBidPrebid, impEx
...
+				return nil, err
+			}
+			//handler for case where EchoVideoAttrs is true, but video data is not found
+			if len(videoData) > 0 {
+				extMap[openrtb_ext.StoredRequestAttributes] = json.RawMessage(videoData)
+			}
 		}
 	}
-
+	extMap[openrtb_ext.PrebidExtKey] = prebid
```
mansinahar, line 1081: Any reason for moving this line after imp video attributes are set?

exchange/exchangetest/passthrough.json
```diff
@@ -0,0 +1,122 @@
...
+                            "price": 0.3,
+                            "w": 200,
+                            "h": 250,
+                            "crid": "creative-1",
+                            "ext": {
+                                "someField": "someValue",
+                                "origbidcpm": 0.3,
+                                "prebid": {
+                                    "willBeOverwritten": "by core logic",
+                                    "passthrough": {
```
mansinahar, line 78: The `mockResponse` is the mocked response from the bidder. This might not have the `passthrough` object in the bid back. This only needs to be asserted in the response from Prebid

endpoints/openrtb2/auction.go
AlexBVolcy: You're right that the reading/setting of `echoVideoAttributes` and `StoredImp` only happened for imps that had a stored imp associated with it. The only thing that's needed to be read/set when there isn't a stored request, is the `passthrough` value found from the imp.

So initially I created a new function to handle the setting of these values cause I thought there was enough duplicate code between when there is a stored request vs. when there isn't, but now I think it makes more sense to remove the new function all together, and leave the original code for when there is a stored request, and when there isn't, I'll just read and set the `passthrough` value to the `impExtInfoMap`.

exchange/exchangetest/passthrough.json
AlexBVolcy: Yeah you're right, I'll remove this piece

exchange/exchange.go
AlexBVolcy: Where the line was originally, the `prebid` object might be updated with a `passthrough` value after the `extMap` is already set. So I moved it to ensure that the `extMap` is set with the `prebid` object with all the correct information present.

exchange/exchange_test.go
```diff
@@ -4211,6 +4240,7 @@ type exchangeSpec struct {
 	StartTime         int64                  `json:"start_time_ms,omitempty"`
 	BidIDGenerator    *mockBidIDGenerator    `json:"bidIDGenerator,omitempty"`
 	RequestType       *metrics.RequestType   `json:"requestType,omitempty"`
+	PassthroughFlag   bool                   `json:"passthrough_ext_flag,omitempty"`
```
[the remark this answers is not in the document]
AlexBVolcy: I think renaming it to passthrough_flag would be enough for me!

exchange/exchange.go
```diff
@@ -1070,20 +1077,22 @@ func makeBidExtJSON(ext json.RawMessage, prebid *openrtb_ext.ExtBidPrebid, impEx
...
-		videoData, _, _, err := jsonparser.Get(impExtInfo.StoredImp, "video")
-		if err != nil && err != jsonparser.KeyPathNotFoundError {
-			return nil, err
-		}
-		//handler for case where EchoVideoAttrs is true, but video data is not found
-		if len(videoData) > 0 {
-			extMap[openrtb_ext.StoredRequestAttributes] = json.RawMessage(videoData)
+	// ext.prebid.storedrequestattributes and ext.prebid.passthrough
+	if impExtInfo, ok := impExtInfoMap[impId]; ok {
+		prebid.Passthrough = impExtInfoMap[impId].Passthrough
```
bsardo, line 1083: Will this create a `passthrough` key at `imp.ext` if there is non-passthrough data for a given imp in `impExtInfoMap`? I think we should avoid doing that. Perhaps we should check if `impExtInfoMap[impId].Passthrough` is set here before performing this assignment. We should probably also have a test for this scenario.

exchange/exchangetest/passthrough.json
```diff
@@ -0,0 +1,116 @@
+{
```
bsardo, line 1: Nitpick: can we rename these test files to something like:
`passthrough.json` --> `passthrough_root_and_imp.json`
`passthrough_multi_imp.json` --> `passthrough_imp_only.json`
`passthrough_only_request.json` --> `passthrough_root_only.json`

exchange/exchange.go
```diff
@@ -1070,20 +1077,24 @@ func makeBidExtJSON(ext json.RawMessage, prebid *openrtb_ext.ExtBidPrebid, impEx
...
-	extMap[openrtb_ext.PrebidExtKey] = prebid
 
-	// ext.storedrequestattributes
-	if impExtInfo, ok := impExtInfoMap[impId]; ok && impExtInfo.EchoVideoAttrs {
-		videoData, _, _, err := jsonparser.Get(impExtInfo.StoredImp, "video")
-		if err != nil && err != jsonparser.KeyPathNotFoundError {
-			return nil, err
+	// ext.prebid.storedrequestattributes and ext.prebid.passthrough
+	if impExtInfo, ok := impExtInfoMap[impId]; ok {
+		if impExtInfoMap[impId].Passthrough != nil {
```
bsardo, line 1083: Hm in retrospect you might not need this check. When the `extMap` is marshaled at the end of this function I think the `Passthrough` field will be omitted from the marshalled json if it is nil or empty.

bsardo (approved):
LGTM

mansinahar (approved):

PR #2315, opened 2022-07-21
IX: Add multi-format ad unit support

oronno:
This PR add support for multi-format ad unit in IX Adapter.  
Sample Impression Object for multi-format request could be like:

```
    "imp": [
        {
            "id": "1_0",
            "banner": {
                "topframe": 1,
                "format": [
                    {
                        "w": 300,
                        "h": 250,
                        "ext": {
                            "siteID": "..",
                            "sid": "300x250"
                        }
                    }
                ]
            },
            "video": {
                "mimes": [
                    "video/mp4"
                ],
                "maxduration": 60,
                "protocols": [
                    2,
                    3,
                    5,
                    6
                ],
                "w": 300,
                "h": 250
            },
            "native": {
                    "request": "{}",
                    "ver": "1.1"
            },
            "ext": {
                "ix": {
                    "siteId": ".."
                }
            }
        }
    ]
```

Exchange may return multiple bids on different type with same impression id. PBS will generate targeting for winning bid.


rokon12 (dismissed):

adapters/ix/ix.go
```diff
@@ -222,6 +224,37 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
...
+	switch bid.MType {
+	case openrtb2.MarkupBanner:
+		return openrtb_ext.BidTypeBanner, nil
+	case openrtb2.MarkupVideo:
+		return openrtb_ext.BidTypeVideo, nil
+	case openrtb2.MarkupAudio:
+		return openrtb_ext.BidTypeAudio, nil
+	case openrtb2.MarkupNative:
+		return openrtb_ext.BidTypeNative, nil
+	}
```
SyntaxNode, line 237: Nice use of the new OpenRTB 2.6 `MType` field. I would like to build support directly in PBS-Core for MType. What behavior would you like ideally to see from how PBS-Core handles this field? Maybe a new interface callback func to get the media type if MType doesn't exist? Maybe a bid error where it's rejected if MType doesn't exist and BidType isn't provided? I'm looking for suggestions.

adapters/ix/ix.go
```diff
@@ -222,6 +224,37 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
...
+			}
+		}
+	}
+
+	bidType, ok := impMediaTypeReq[bid.ImpID]
+	if ok {
+		return bidType, nil
+	} else {
+		return "", fmt.Errorf("unmatched impression id: %s", bid.ImpID)
+	}
```
SyntaxNode, line 255: Nitpick: Can be written a little more compactly:

```
if bidType, ok := impMediaTypeReq[bid.ImpID]; ok {
	return bidType, nil
} else {
	return "", fmt.Errorf("unmatched impression id: %s", bid.ImpID)
}
```

adapters/ix/ix.go
```diff
@@ -222,6 +224,37 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
...
+		return openrtb_ext.BidTypeNative, nil
+	}
+
+	if bid.Ext != nil {
+		var bidExt openrtb_ext.ExtBid
+		err := json.Unmarshal(bid.Ext, &bidExt)
+		if err == nil && bidExt.Prebid != nil {
+			prebidType := string(bidExt.Prebid.Type)
+			if prebidType != "" {
+				return openrtb_ext.ParseBidType(prebidType)
```
SyntaxNode, line 245: Could you please add a json test for this case?

adapters/ix/ix.go
SyntaxNode, line 237: There's a json test covering the `MarkupVideo` case. I'm not going to ask for coverage of all other media types since the code is super simple and will be moved up into PBS-Core logic soon, although I welcome opinions for how to handle the fallback case.

oronno: > Maybe a bid error where it's rejected if MType doesn't exist  

Not sure if we should do this, because mtype is not a required field in openrtb 2.6 and I believe, most of the exchanges don't support it yet.    
IMO, PBS-Core can implement similar logic as I made here - take mtype if present, else fallback to ext.prebid.type; otherwise bid request's type as last resort (for non-multiformat). In that case, Adapter don't need to handle type anymore.

adapters/ix/ix.go
oronno: added

adapters/ix/ix.go
oronno: updated. thanks

mansinahar (approved):

SyntaxNode (approved):

PR #2316, opened 2022-07-21
Triplelift: Expand Video Formats

SyntaxNode:
Created on behalf of Triplelift.

### DIFF
--- a/adapters/triplelift/triplelift.go
+++ b/adapters/triplelift/triplelift.go
@@ -26,7 +26,7 @@ type TripleliftRespExt struct {
 
 func getBidType(ext TripleliftRespExt) openrtb_ext.BidType {
 	t := ext.Triplelift.Format
-	if t == 11 {
+	if t == 11 || t == 12 || t == 17 {
 		return openrtb_ext.BidTypeVideo
 	}
 	return openrtb_ext.BidTypeBanner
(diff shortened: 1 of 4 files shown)


nllerandi3lift:
Thanks looks good.

adapters/triplelift/triplelift.go
```diff
@@ -26,7 +26,7 @@ type TripleliftRespExt struct {
 
 func getBidType(ext TripleliftRespExt) openrtb_ext.BidType {
 	t := ext.Triplelift.Format
-	if t == 11 {
+	if t == 11 || t == 12 || t == 17 {
```
VeronikaSolovei9, line 29: Would it be better to extract these magic numbers to constants?

bsardo (approved):
LGTM

adapters/triplelift/triplelift.go
SyntaxNode: Ideally, I agree. I do not know what these format values represent. I don't think it matters too much in this context and I'm personally ok with magic numbers in this instance. Do you feel strongly about this suggestion or is this ok?

nllerandi3lift: These are just different ids indicating instream, outstream, and one other video format. I don't think it's super relevant to have that indicated here.

VeronikaSolovei9 (approved):

nllerandi3lift:
@bretg this resolved a rendering issue we were seeing when testing Outstream (recently added format). Wondering if you can help port to Java.

bretg:
Adding @net-burst for FYI. Oleksandr, please consider this one of the higher priority porting tasks. Thanks.

Net-burst:
Done in https://github.com/prebid/prebid-server-java/pull/1996

PR #2325, opened 2022-07-26
New Adapter: Seeding Alliance

sag-celejd:
https://github.com/prebid/prebid.github.io/pull/3927


static/bidder-info/seedingAlliance.yaml
```diff
@@ -0,0 +1,12 @@
...
+gvlVendorID: 371
+capabilities:
+  site:
+    mediaTypes:
+      - native
+      - banner
+userSync:
+  redirect:
+    url: https://dmp.nativendo.de/set-uuid?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&redirect_url={{.RedirectURL}}
+    userMacro: $UID
```
VeronikaSolovei9, line 12: Please make sure this URL is correct and returns redirect with user macro. 
I tested it with encoded Prebid Server url (https://prebid.adnxs.com/pbs/v1/setuid?bidder=3DseedingAlliance&gdpr=&gdpr_consent=&f=i&uid=$UID): `https://dmp.nativendo.de/set-uuid?gdpr=&gdpr_consent=&redirect_url=https%3A%2F%2Fprebid.adnxs.com%2Fpbs%2Fv1%2Fsetuid%3Fbidder%3DseedingAlliance%26gdpr%3D%26gdpr_consent%3D%26f%3Di%26uid%3D%24UID` and expect to see redirect, but see  http 400 error code instead. 

Example of a working URL https://js.adscale.de/pbsync.html?gdpr=&gdpr_consent=&redirect=https%3A%2F%2Fprebid.adnxs.com%2Fpbs%2Fv1%2Fsetuid%3Fbidder%3Dbliink%26gdpr%3D%26gdpr_consent%3D%26f%3Di%26uid%3D
It returns http 200 and then http 302 statuses

static/bidder-params/seedingAlliance.json
```diff
@@ -0,0 +1,12 @@
...
+    "title": "Seeding Alliance Adapter Params",
+    "description": "A schema which validates params accepted by the Seeding Alliance adapter",
+    "type": "object",
+    "properties": {
+        "adUnitId": {
+            "type": "string",
+            "description": "Ad Unit ID"
+        }
+    }
+}
```
VeronikaSolovei9, line 12: Just to confirm: adUnitId is not a required parameter and can be omitted, right?

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,142 @@
+package seedingAlliance
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/mxmCherry/openrtb/v15/openrtb2"
```
VeronikaSolovei9, line 10: While merging with main Prebid Server branch, please set it ti v16.

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/errortypes"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type SeedingAllianceAdapter struct {
```
VeronikaSolovei9, line 17: Please rename to `type adapter struct {`, there is no need to duplicate a package name here.

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,142 @@
...
+func addTagID(imp *openrtb2.Imp) error {
+	var ext adapters.ExtImpBidder
+	var extSA openrtb_ext.ImpExtSeedingAlliance
+
+	if err := json.Unmarshal(imp.Ext, &ext); err != nil {
+		return fmt.Errorf("could not unmarshal adapters.ExtImpBidder: %w", err)
+	}
+
+	if err := json.Unmarshal(ext.Bidder, &extSA); err != nil {
+		if v, ok := ext.Prebid.Bidder["seedingAlliance"]; ok {
```
VeronikaSolovei9, line 132: There is no need for this check. Prebid Server gives it to adapter in the same way regardless if imp.ext looks like this 
```
"ext": {
                "seedingAlliance": {
                    "adunitid": "test1"
                }
            }
```
or like this:
```
"ext": {
                "prebid": {
                    "bidder": {
                        "seedingAlliance": {
                            "adunitid": "test1"
                        }
                    }
                }
            }
```

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,142 @@
...
+	requestJSON, err := json.Marshal(request)
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	requestData := &adapters.RequestData{
+		Method: http.MethodPost,
+		Uri:    a.endpoint,
+		Body:   requestJSON,
+	}
```
VeronikaSolovei9, line 46: Just to confirm: are there any http headers expected?

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (a *SeedingAllianceAdapter) MakeRequests(request *openrtb2.BidRequest, extraRequestInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	for i := range request.Imp {
+		if err := addTagID(&request.Imp[i]); err != nil {
+			return nil, []error{err}
+		}
+	}
+
+	request.Cur = append(request.Cur, "EUR")
```
VeronikaSolovei9, line 35: Need to confirm with a team if this is a correct way to set currency like this

VeronikaSolovei9 (commented):
Added some comments, please merge your branch with latest Prebid Server master branch

static/bidder-params/seedingAlliance.json
sag-celejd: Thanks for noticing that.  Should have been actually set as "required".

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: Yes, no headers are expected.

static/bidder-info/seedingAlliance.yaml
sag-celejd: The URL itself is fine but we expect a non nil value for **gdpr** on our end.

Meaning that the following URL should work fine:
`https://dmp.nativendo.de/set-uuid?gdpr=0&gdpr_consent=&redirect_url=https://prebid.adnxs.com/pbs/v1/setuid?bidder=seedingAlliance&gdpr=&gdpr_consent=&f=i&uid=$UID`

Is this an issue?

VeronikaSolovei9: Oh I see, please make sure you support encoded redirect URL, this is how PBS will send it.

adapters/seedingAlliance/seedingAlliance.go
VeronikaSolovei9: Discussed with team. This is ok from our perspective.  For bidder there are can be two "EUR". Do you want to check this array has it already before append?

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,140 @@
...
+	var extSA openrtb_ext.ImpExtSeedingAlliance
+
+	if err := json.Unmarshal(imp.Ext, &ext); err != nil {
+		return fmt.Errorf("could not unmarshal adapters.ExtImpBidder: %w", err)
+	}
+
+	if err := json.Unmarshal(ext.Bidder, &extSA); err != nil {
+		if err := json.Unmarshal(ext.Prebid.Bidder["seedingAlliance"], &extSA); err != nil {
+			return fmt.Errorf("could not unmarshal openrtb_ext.ImpExtSeedingAlliance: %w", err)
+		}
```
VeronikaSolovei9, line 134: Lines 132-134 are not necessary, I mentioned this in one of my previous comments.

static/bidder-info/seedingAlliance.yaml
sag-celejd: Just made sure that we actually support encoded redirect URLs. 👍

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: Just checked again - you are correct about that during regular usage but our tests fail when we remove line 132-134 since **ext.Bidder** appears to be **nil/empty** when running _TestJsonSamples(t *testing.T)_.

Is this intended?

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: Thanks for confirming it!
That's a good point - just pushed a commit with the check.

adapters/seedingAlliance/seedingAlliance.go
VeronikaSolovei9: You use this format of imp.ext in unite tests: 
```
 "prebid": {
     "bidder": {
          "seedingAlliance": {
                 "adUnitId": "example-tag-id"
           }
       }
}
```
Modify it to 
```
 "prebid": {
       "bidder": {
               "adUnitId": "example-tag-id"
        }
 }

```
Please refer to other bidders examples:
https://github.com/prebid/prebid-server/blob/3b008e46db95a7d31fdfabea975315209feb03c8/adapters/appnexus/appnexustest/exemplary/simple-banner.json#L22

SyntaxNode: Bidder parameters are specified in the request at either `imp.ext.prebid.bidders.BIDDER` or the legacy `imp.ext.prebid.BIDDER` locations. PBS moves bidder parameters to `imp.ext.prebid.bidder` for the request. The test cases represent what PBS sends to the bidder code, so the later is the correct format.

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,148 @@
...
+		if err := addTagID(&request.Imp[i]); err != nil {
+			return nil, []error{err}
+		}
+	}
+
+	eurExists := false
+	for _, cur := range request.Cur {
+		if cur == "EUR" {
+			eurExists = true
+		}
```
VeronikaSolovei9, line 39: Can you please extract lines 35-40 to a separate function? You can optimize code a little by adding `break` after `eurExists = true`, so it's found already and you don't need to iterate over other elements.

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: Thank you both, 
`prebid.bidder.adUnitId` didn't work but `ext.bidder.adUnitId` did the trick!

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: Done 👍

VeronikaSolovei9 (commented):
Thank you for addressing comments. Code looks good to me now. 
I still have http 400 when I test user sync URL, will confirm with a team if it's ok and approve this PR.

static/bidder-info/seedingAlliance.yaml
VeronikaSolovei9: We cannot merge this PR if User sync URL doesn't work. It should work with encoded redirect URL and return http 200 and http 302 statuses.

VeronikaSolovei9:
Added comment above about UserSync URL

static/bidder-info/seedingAlliance.yaml
sag-celejd: https://dmp.nativendo.de/set-uuid?gdpr=0&gdpr_consent=&redirect_url=https%3A%2F%2Fprebid.adnxs.com%2Fpbs%2Fv1%2Fsetuid%3Fbidder%3DseedingAlliance%26gdpr%3D%26gdpr_consent%3D%26f%3Di%26uid%3D%24UID

That 400 response is not coming from our DMP. That status (incl. the following error message: `The bidder name provided is not supported by Prebid Server`) shows up after the redirect since I guess the adapter isn't merged yet.

SyntaxNode: @sag-celejd The user sync url fails if the gdpr field is empty. Prebid Server will set the `gdpr` parameter to either empty, 0, or 1 and all must be supported. An empty gdpr is a lack of a signal versus an explicit 0 or 1 signal. If the gdpr is empty we receive a 400 and thats is what @VeronikaSolovei9 is referring to.

sag-celejd: Alright, I just changed the behaviour on our end:

https://dmp.nativendo.de/set-uuid?gdpr=&gdpr_consent=&redirect_url=https%3A%2F%2Fprebid.adnxs.com%2Fpbs%2Fv1%2Fsetuid%3Fbidder%3DseedingAlliance%26gdpr%3D%26gdpr_consent%3D%26f%3Di%26uid%3D%24UID

Should work fine now.

VeronikaSolovei9: Yes, works good now, thank you for addressing this!

VeronikaSolovei9 (dismissed):
Thank you for addressing changes, LGTM

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func getMediaTypeForImp(impId string, imps []openrtb2.Imp) openrtb_ext.BidType {
+	var UnsupportedBidType openrtb_ext.BidType = "unsupported"
+	var UnknownBidType openrtb_ext.BidType = "unknown"
+
+	for _, imp := range imps {
+		if imp.ID == impId {
+			switch {
+			case imp.Native != nil:
```
mansinahar, line 108: Determining media type of the bid by looking at the media type in the imp is not recommended because in case of multi-format imps, this can lead to an incorrect value depending on how your bidder handles multi-format imps. For example, consider an imp that has both `Native` and `Banner` specified. Now if your bidder returned a `banner` bid then this code will wrongly assign `Native` as the bid type. 

We recommend using `bid.ext.prebid.type` to send this information from your bidder to Prebid Server.

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,149 @@
...
+				return openrtb_ext.BidTypeNative
+			case imp.Banner != nil:
+				return openrtb_ext.BidTypeBanner
+			case imp.Video != nil:
+				return UnsupportedBidType
```
mansinahar, line 113: Also, I am not sure if these cases for video and audio are needed. The way Prebid Server Core code functions, it won't let through an imp to the adapter if the adapter doesn't support the imp's media type. So unless you have video and audio listed as supported media types in the `seedingAlliance.json` file, this is pretty much non-reachable code. Moreover, in a case where your bidder does receive a request with types other than banner or native, I am assuming it will return an error, no? Again leading to this code being non-reachable.

adapters/seedingAlliance/seedingAlliancetest/exemplary/native.json
```diff
@@ -0,0 +1,84 @@
...
+                            "seat": "123"
+                        }
+                    ],
+                    "bidid": "8141327771600527856",
+                    "cur": "EUR"
+                }
+            }
+        }
+    ],
+    "expectedBidResponses": []
```
mansinahar, line 83: Why is the `expectedBidResponses` empty here? We should be expecting a native bid, no?

adapters/seedingAlliance/seedingAlliancetest/supplemental/status_bad_request.json
```diff
@@ -0,0 +1,72 @@
...
+                "banner": {
+                    "format": [
+                        {
+                            "w": 300,
+                            "h": 250
+                        },
+                        {
+                            "w": 300,
+                            "h": 600
+                        }
```
mansinahar, line 16: Why would this be considered as a bad request?

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -0,0 +1,149 @@
...
+
+	if err := json.Unmarshal(imp.Ext, &ext); err != nil {
+		return fmt.Errorf("could not unmarshal adapters.ExtImpBidder: %w", err)
+	}
+
+	if err := json.Unmarshal(ext.Bidder, &extSA); err != nil {
+		return fmt.Errorf("could not unmarshal openrtb_ext.ImpExtSeedingAlliance: %w", err)
+	}
+
+	imp.TagID = extSA.AdUnitID
```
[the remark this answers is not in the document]
sag-celejd: I added the minLength property - thanks for the hint.

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: That's not a change we would want to do at this point. We are fine with current function and aware of the potential issues. Seems also like that's how many other adapters are doing it.

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: I am aware that those switch cases should be unreachable code. It's rather a style thing (exhaustive switches etc.).
See my last commit - I just removed the unreachable cases.

adapters/seedingAlliance/seedingAlliance.go
mansinahar: Sounds good 👍

mansinahar:
@sag-celejd Thanks for addressing the comments. The changes look good. Can you please also check out my comments about the JSON tests and address those when you have a chance?

adapters/seedingAlliance/seedingAlliancetest/exemplary/native.json
sag-celejd: Thanks for pointing that out - expectedBidResponses is no longer empty.

adapters/seedingAlliance/seedingAlliancetest/supplemental/status_bad_request.json
sag-celejd: That request itself is trivial/meaningless - I'm just mocking a _Bad Request_ response to test the _MakeBids_ method to accomplish a higher code test coverage.

mansinahar (approved):

VeronikaSolovei9 (approved):

PR #2327, opened 2022-07-27
Request Wrapper: Only Create Elements If Needed

SyntaxNode:
This PR change the behavior of the RequestWrapper to only create new objects if there a non-empty extension object. I also fixed some misc marhsal issues that I discovered through the rebuild ext methods.

### DIFF
--- a/openrtb_ext/request_wrapper.go
+++ b/openrtb_ext/request_wrapper.go
@@ -262,16 +262,21 @@ func (rw *RequestWrapper) rebuildSiteExt() error {
 }
 
 func (rw *RequestWrapper) rebuildSourceExt() error {
-	if rw.Source == nil && rw.sourceExt != nil && rw.sourceExt.Dirty() {
-		rw.Source = &openrtb2.Source{}
+	if rw.sourceExt == nil || !rw.sourceExt.Dirty() {
+		return nil
 	}
-	if rw.sourceExt != nil && rw.sourceExt.Dirty() {
-		sourceJson, err := rw.sourceExt.marshal()
-		if err != nil {
-			return err
-		}
+
+	sourceJson, err := rw.sourceExt.marshal()
+	if err != nil {
+		return err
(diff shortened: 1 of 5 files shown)


openrtb_ext/request_wrapper.go
```diff
@@ -780,7 +832,7 @@ func (re *RegExt) SetUSPrivacy(uSPrivacy string) {
 type SiteExt struct {
 	ext      map[string]json.RawMessage
 	extDirty bool
-	amp      int8
+	amp      *int8
```
AlexBVolcy, line 835: We spoke about this yesterday, but I just wanted to confirm: this was changed to a pointer so we could check if this `amp` variable is empty?

AlexBVolcy (commented):
Looks good! Just left some minor comments

openrtb_ext/request_wrapper.go
```diff
@@ -41,6 +41,8 @@ type RequestWrapper struct {
 	sourceExt  *SourceExt
 }
 
+const jsonEmptyObjectLength = 2
```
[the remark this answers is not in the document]
SyntaxNode: Yes, exactly. It's the only reason a json object will be 2 characters.

openrtb_ext/request_wrapper.go
```diff
@@ -780,7 +832,7 @@ func (re *RegExt) SetUSPrivacy(uSPrivacy string) {
 type SiteExt struct {
```
mansinahar, line 832: Not really related to your changes but wondering if there's any difference between `SiteExt` that's defined here vs `ExtSite` defined in `openrtb_ext/site.go`? The latter doesn't really seem to be used anywhere.. perhaps we can remove it as part of a separate PR?

mansinahar (approved):

AlexBVolcy (approved):

PR #2330, opened 2022-07-29
GDPR: add enforce_algo config flag and convert enforce_purpose to bool

bsardo:
This PR makes two config changes to both the host config and account configs:
1) The `gdpr.tcf2.purpose{1-10}.enforce_algo` field is introduced
2) The `gdpr.tcf2.purpose{1-10}.enforce_purpose` field is converted from a string to bool.

The `enforce_purpose` field was originally a string that could either be set to `"no"` or `"full"` which indicated whether the GDPR full enforcement algorithm or no enforcement algorithm should be executed for a given purpose. If it was set to `"full"` we assumed we were enforcing the purpose and when set to `"no"` we weren't enforcing the purpose. These assumptions were ok under the current implementation but a `"basic"` algorithm enforcement option will be introduced shortly which will limit hosts a bit. By converting to bool we will give hosts some more flexibility as they will be able to explicitly state whether they want to run the full or basic algorithm for a given purpose with or without enforcing the purpose. If we kept it as a string this would not be possible.

**Old implementation:**
```
enforce_purpose  |  assumption
     "no"        |   purpose is not enforced
     "basic"     |   purpose is enforced
     "full"      |   purpose is enforced
```

**New implementation:**
```
enforce_purpose  |  enforce_algo
   false         |     "basic"
   false         |     "full"
   true          |     "basic"
   true          |     "full"

*no assumptions - hosts can explicitly state whether or not to enforce a purpose for a given algorithm
```

**Conversion logic:**
On startup we call the functions `migrateConfigTCF2EnforcePurposeFlags` and `migrateConfigTCF2PurposeEnabledFlags` in succession.
`migrateConfigTCF2EnforcePurposeFlags` attempts to set `enforce_algo` based on `enforce_purpose` if `enforce_algo` is not set and `enforce_purpose` is set to a string value. It will also convert an `enforce_purpose` string value to bool. Various warnings are emitted to notify the host that something in their config has been deprecated.
 `migrateConfigTCF2PurposeEnabledFlags` attempts to set `enforce_purpose` based on the previously deprecated `enabled` field if `enforce_purpose` is not set but `enabled` is set. This logic has been simplified a bit as it now assumes that if `enforce_purpose` is set that it is a bool since the function `migrateConfigTCF2EnforcePurposeFlags` called before it will convert any string value to bool.

For every account fetched, we perform similar conversion logic to what is done in the host config. An `enforce_purpose` string type is detected when the fetched account config JSON is unmarshaled and an attempt is made to apply the conversion logic to see if it resolves the unmarshal error. If it does not, a malformed account config error is reported. This type of error is reported as a metric for visibility.


VeronikaSolovei9 (dismissed):
Looks good to me.

config/accounts_test.go
```diff
@@ -246,11 +246,14 @@ func TestAccountIntegrationGetByIntegrationType(t *testing.T) {
...
+	True := true
+	False := false
+
 	tests := []struct {
 		description          string
 		givePurposeConfigNil bool
-		givePurpose1Enforced string
-		givePurpose2Enforced string
+		givePurpose1Enforced *bool
+		givePurpose2Enforced *bool
```
AlexBVolcy, line 256: Could you explain why the type change here?

config/config.go
```diff
@@ -1259,6 +1269,47 @@ func migrateConfigSpecialFeature1(v *viper.Viper) {
 	}
 }
 
+func migrateConfigTCF2PurposeFlags(v *viper.Viper) {
+	migrateConfigTCF2EnforcePurposeFlags(v)
+	migrateConfigTCF2PurposeEnabledFlags(v)
+}
+
+func migrateConfigTCF2EnforcePurposeFlags(v *viper.Viper) {
+	for i := 1; i <= 10; i++ {
```
AlexBVolcy, line 1278: How do we know this for loop needs to run 10 times specifically?

errortypes/errortypes.go
```diff
@@ -165,6 +165,25 @@ func (err *BidderTemporarilyDisabled) Severity() Severity {
 	return SeverityWarning
 }
 
+// MalformedAcct should be used when the retrieved account config cannot be unmarshaled
+//
```
AlexBVolcy, line 169: Nitpick: Could you remove this unused comment?

config/config.go
bsardo: There are ten purposes, each of which can be configured in the host config and account config. Each purpose config can be specified as `gdpr.tcf2.purposeX` where X is 1-10.

config/accounts_test.go
bsardo: Sure, the function here under test is the account config method `PurposeEnforced` which indicates whether a given purpose is enforced. That function looks at whether `EnforcePurpose` is set on the account config. `EnforcePurpose` used to be a string value and now, as of this PR, it's a bool so these two values have changed from string to bool since they are used to setup the config for the test. They are also pointers just like the `EnforcePurpose` field on the `AccountGDPRPurpose` object because it allows us to detect presence. Before we were able to detect presence when it was just a string (no pointer) by looking for an empty string since the only valid values were non-empty values "full" and "no".

AlexBVolcy (commented):
Looks good, one minor comment

config/config.go
```diff
@@ -259,11 +260,11 @@ func (cfg *GDPR) validatePurposes(errs []error) []error {
...
 	for i := 0; i < len(purposeConfigs); i++ {
-		enforcePurposeValue := purposeConfigs[i].EnforcePurpose
-		enforcePurposeField := fmt.Sprintf("gdpr.tcf2.purpose%d.enforce_purpose", (i + 1))
+		enforceAlgoValue := purposeConfigs[i].EnforceAlgo
+		enforceAlgoField := fmt.Sprintf("gdpr.tcf2.purpose%d.enforce_algo", (i + 1))
 
-		if enforcePurposeValue != TCF2NoEnforcement && enforcePurposeValue != TCF2FullEnforcement {
-			errs = append(errs, fmt.Errorf("%s must be \"no\" or \"full\". Got %s", enforcePurposeField, enforcePurposeValue))
+		if enforceAlgoValue != TCF2FullEnforcement {
+			errs = append(errs, fmt.Errorf("%s must be \"full\". Got %s", enforceAlgoField, enforceAlgoValue))
```
VeronikaSolovei9, line 267: Comment for L 288 - no usages for `TCF2NoEnforcement   = "no"`, which is a good sign. Do you think this constant should live in code still?

VeronikaSolovei9 (approved):
Looks good to me

account/account.go
```diff
@@ -120,3 +136,79 @@ func setDerivedConfig(account *config.Account) {
...
+	}
+
+	patchConfig, err := json.Marshal(newAccount)
+	if err != nil {
+		return nil, err
+	}
+
+	newConfig, err = jsonpatch.MergePatch(config, patchConfig)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
bsardo: Hm I don't see a way to test this. For the merge patch to fail either the config or the patch has to be malformed. A malformed config will never make it this far because an error will be caught at the top of the function when calling `jsonparser.Get(config, "gdpr")` which I already have a test case for. And the patch cannot be malformed because we are correctly programmatically constructing it here. The function would need to be refactored to achieve this and that probably isn't worth it.

config/config.go
bsardo: 👍 removed

account/account.go
```diff
@@ -54,9 +55,16 @@ func GetAccount(ctx context.Context, cfg *config.Configuration, fetcher stored_r
 
 			// unmarshal fetched account to determine if it is well-formed
 			if _, ok := err.(*json.UnmarshalTypeError); ok {
-				return nil, []error{&errortypes.MalformedAcct{
-					Message: fmt.Sprintf("The prebid-server account config for account id \"%s\" is malformed. Please reach out to the prebid server host.", accountID),
-				}}
+				// attempt to convert deprecated GDPR enforce purpose fields to resolve issue
+				completeJSON, err = ConvertGDPREnforcePurposeFields(completeJSON)
+				// unmarshal again to check if unmarshal error still exists after GDPR field conversion
+				err = json.Unmarshal(completeJSON, account)
```
bsardo, line 61: Add note here that this logic exists for backwards compatibility reasons and will eventually be removed.

AlexBVolcy (approved):

VeronikaSolovei9 (approved):

Issue #2321, opened 2022-07-22
eidpermissions Exposed to Bidders


PDXandr:
Please confirm if eidpermissions are intended to be passed along to bidders.  I would think publishers would not want that information exposed as it informs bidders which competitors the publisher is using, and which have access to first-party data.

Example of eidpermissions here but this does not confirm the behavior: https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#post-openrtb2auction

In testing at Xandr, request to bidders included:

`{\"eidpermissions\":[{\"source\":\"intTestSource2\",\"bidders\":[\"pubmatic\",\"appnexus\"]}]}`

bretg:
Good call @PDXandr . Here's the [table of all the various extensions and whether bidders should see them](https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#prebid-server-ortb2-extension-summary) -- both teams should review all of them. Have opened an internal ticket for the PBS-Java team.

PDXandr:
Apologies for missing that clarification at the bottom of the documentation @bretg .  @SyntaxNode let me know if you need anything else for the Go side.

SyntaxNode:
Looks like the entire `eidpermissions` should not be sent to the bidder.

We'll work on this for PBS-Go. I agree with removing everything in ext.prebid and imp.ext.prebid unless specifically allowed. This looks to currently be:
- `imp[].ext.prebid.bidder` (only their value)
- `imp[].ext.prebid.is_rewarded_inventory`
- `ext.prebid.integration`
- `ext.prebid.channel`
- `ext.prebid.debug`
- `ext.prebid.currency.rates`
- `ext.prebid.currency.usepbsrates`
- `ext.prebid.interstitial`
- `ext.prebid.multibid` (only their value)
- ext.prebid.auctiontimestamp
- ext.prebid.options. echovideoattrs

SyntaxNode:
Fixed in PBS-Go 0.222.0.

bretg:
Moving to https://github.com/prebid/prebid-server/issues/2383. That one is a broader general review of ext.prebid and what's sent to bidders including EIDs.

PR #2331, opened 2022-08-01
Ext Fields Exposed to Bidders

SyntaxNode:
Resolves https://github.com/prebid/prebid-server/issues/2321


adapters/33across/33across.go
```diff
@@ -48,7 +48,7 @@ type bidExt struct {
 }
 
 type bidTtxExt struct {
-	MediaType string `json:mediaType,omitempty`
+	MediaType string `json:"mediaType,omitempty"`
```
SyntaxNode, line 51: Unrelated issue discovered during this PR.

exchange/exchangetest/aliases.json
```diff
@@ -58,11 +58,7 @@
             }
           ],
           "ext": {
-            "prebid": {
-              "aliases": {
-                "districtm": "appnexus"
-              }
-            }
+            "prebid": {}
```
SyntaxNode, line 61: There is an existing behavior where if `ext.prebid` is empty the empty object is written to the request. It's easy to remove this field instead. Is that safe? I imagine the current functionality is an accident, but it does provide a signal downstream that the request was processed by PBS. Do you think it's important to preserve that behavior? I'd personally love to make the request as compact/small as possible.

exchange/exchangetest/ccpa-nosale-specific-bidder.json
```diff
@@ -6,17 +6,19 @@
...
+                        "mimes": ["video/mp4"]
+                    },
+                    "ext": {
+                        "appnexus": {
+                            "placementId": 1
+                        }
                     }
                 }
-            }],
+            ],
```
SyntaxNode, line 21: I used a different json formatter, so there will be some of these minor changes. The VS Code Beautify plugin is no longer maintained.

openrtb_ext/request.go
```diff
@@ -38,26 +38,25 @@ type ExtRequestPrebid struct {
...
-	BidderParams         json.RawMessage           `json:"bidderparams,omitempty"`
 
 	// NoSale specifies bidders with whom the publisher has a legal relationship where the
 	// passing of personally identifiable information doesn't constitute a sale per CCPA law.
 	// The array may contain a single sstar ('*') entry to represent all bidders.
 	NoSale []string `json:"nosale,omitempty"`
-
-	CurrencyConversions *ExtRequestCurrency `json:"currency,omitempty"`
-	BidderConfigs       []BidderConfig      `json:"bidderconfig,omitempty"`
-	Experiment          *Experiment         `json:"experiment,omitempty"`
```
SyntaxNode, line 60: I alphabetized the list for easy of reading, except for NoSale since it has a length commend which breaks the code flow.

schain/schainwriter.go
```diff
@@ -79,11 +79,5 @@ func (w SChainWriter) Write(req *openrtb2.BidRequest, bidder string) {
...
 // extPrebidSChainExists checks if an schain exists in the ORTB 2.5 req.ext.prebid.schain location
 func extPrebidSChainExists(reqExt *openrtb_ext.ExtRequest) bool {
-	if reqExt == nil {
-		return false
-	}
-	if reqExt.Prebid.SChains == nil {
-		return false
-	}
-	return true
+	return reqExt != nil && reqExt.Prebid.SChains != nil
```
SyntaxNode, line 82: Simplifying while I came across it.

VeronikaSolovei9 (dismissed):
LGTM

exchange/utils.go
```diff
@@ -396,17 +399,28 @@ func splitImps(imps []openrtb2.Imp) (map[string][]openrtb2.Imp, error) {
 }
 
 func createSanitizedImpExt(impExt, impExtPrebid map[string]json.RawMessage) (map[string]json.RawMessage, error) {
-	sanitizedImpExt := make(map[string]json.RawMessage, 3)
+	sanitizedImpExt := make(map[string]json.RawMessage, 8)
+	sanitizedImpPrebidExt := make(map[string]json.RawMessage, 8)
```
bsardo, line 403: Shouldn't the size of these be 6 and 2 respectively instead of 8 and 8?

exchange/utils.go
```diff
@@ -396,17 +399,28 @@ func splitImps(imps []openrtb2.Imp) (map[string][]openrtb2.Imp, error) {
...
+	// marshal sanitized imp[].ext.prebid
+	if len(sanitizedImpPrebidExt) > 0 {
+		if impExtPrebidJSON, err := json.Marshal(sanitizedImpPrebidExt); err == nil {
 			sanitizedImpExt[openrtb_ext.PrebidExtKey] = impExtPrebidJSON
 		} else {
 			return nil, fmt.Errorf("cannot marshal ext.prebid: %v", err)
 		}
 	}
 
+	// copy reserved imp[].ext fields known to bot be bidder names
```
bsardo, line 423: Super nitpick: typo "bot be"

bsardo (commented):
Looks good. Just a couple of comments.
Also I see on the corresponding issue that the following fields are also permitted:
```
ext.prebid.interstitial
ext.prebid.multibid (only their value)
ext.prebid.auctiontimestamp
```
I don't see any evidence that we support these at the moment, however, I do see that multibid is currently being worked on.

exchange/utils.go
SyntaxNode: Yes

SyntaxNode:
> Looks good. Just a couple of comments. Also I see on the corresponding issue that the following fields are also permitted:
> 
> ```
> ext.prebid.interstitial
> ext.prebid.multibid (only their value)
> ext.prebid.auctiontimestamp
> ```
> 
> I don't see any evidence that we support these at the moment, however, I do see that multibid is currently being worked on.

`ext.prebid.interstitial` is a typo in the docs and should actually be `device.ext.prebid.interstitial` which does not need to be pruned because it's the only field in that ext.

`ext.prebid.multibid` is not yet implemented in PBS-Go.

`ext.prebid.auctiontimestamp` also doesn't exist in PBS-Go. Not sure why, we should look into that - but I think that's best handled separately.

bsardo (approved):
LGTM

VeronikaSolovei9 (approved):

PR #2333, opened 2022-08-02
AdsCert 2.0: Added logger bridge to adscert library

VeronikaSolovei9:


experiment/adscert/SignerLogger.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"github.com/golang/glog"
+)
+
+type SignerLogger struct {
+}
+
+func (sl *SignerLogger) Debugf(format string, args ...interface{}) {
+	//there is no Debug level in glog
+	glog.Infof(format, args...)
```
mansinahar, line 13: Is there any advantage of using `glog` vs the default level logger that comes with the `adscert/logger` package?

experiment/adscert/signer.go
```diff
@@ -22,6 +23,7 @@ func (ns *NilSigner) Sign(destinationURL string, body []byte) (string, error) {
 }
 
 func NewAdCertsSigner(experimentAdCertsConfig config.ExperimentAdsCert) (Signer, error) {
+	logger.SetLoggerImpl(&SignerLogger{})
```
mansinahar, line 26: I don't see the `logger` being used anywhere. Is there going to be a separate PR that makes use of this logger to log messages? Or am I missing something here?

mansinahar (approved):

AlexBVolcy (approved):

PR #2334, opened 2022-08-02
pubmatic adapter: support native ads

pm-nilesh-chate:


adapters/pubmatic/pubmatic.go
```diff
@@ -401,6 +402,13 @@ func (a *PubmaticAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externa
 				}
 				bidType = getBidType(bidExt)
 			}
+
+			if bidType == openrtb_ext.BidTypeNative {
+				if value, _, _, err := jsonparser.Get([]byte(bid.AdM), string(openrtb_ext.BidTypeNative)); err == nil {
+					bid.AdM = string(value)
+				}
+			}
```
VeronikaSolovei9, line 410: It looks like you never add any data in `errs` array. Can you add append statements where they are needed? Also if error occurs do you still want to add this bid to response?

pm-nilesh-chate: We had an observation that a few bidders send the native response content at non-ORTB location. i.e. A few bidders place their native response at `bid.adm.native` instead of `bid.adm`.

Hence, we check if there is a `bid.adm.native` field and move its content a level up at `bid.adm`. We ignore the error assuming that `bid.adm.native` is absent and the native response content is already at correct location. i.e `bid.adm`.

VeronikaSolovei9: I see what you mean. 
My concern is about unused `errs` variable. General recommendation - is to return unmarshalling errors too. I know this is beyond of the scope of this PR, but would be nice if you fix it.

VeronikaSolovei9 (commented):
Added a recommendation about `errs` array handling. Other than this code changes look good to me.

adapters/pubmatic/pubmatic.go
AlexBVolcy: Yeah I agree with @VeronikaSolovei9 that I'd like to see an appending to the `errs` variable if `err != nil`

VeronikaSolovei9 (approved):
Thank you for the changes. LGTM.

AlexBVolcy (approved):

adapters/pubmatic/pubmatic.go
pm-nilesh-chate: Addressed one of the `err != nil` case in this PR. WIP to handle other `err` reporting https://github.com/prebid/prebid-server/pull/2346

PR #2337, opened 2022-08-03
Add account-specific error metrics to setuid endpoint

bsardo:
This PR makes the setuid endpoint consistent with the cookie sync endpoint in how it reports account-related error metrics. Instead of always reporting "bad request" when there is an account error, we now indicate whether the account was blocked, or invalid before defaulting to "bad request".
Also a few changes were made to ensure any errors encountered by the setuid endpoint are appropriately logged when calling the analytic adapter `LogSetUIDObject` function.

### DIFF
--- a/endpoints/setuid.go
+++ b/endpoints/setuid.go
@@ -61,6 +61,7 @@ func NewSetUIDEndpoint(cfg *config.Configuration, syncersByBidder map[string]use
 			w.WriteHeader(http.StatusBadRequest)
 			w.Write([]byte(err.Error()))
 			metricsEngine.RecordSetUid(metrics.SetUidSyncerUnknown)
+			so.Errors = []error{err}
 			so.Status = http.StatusBadRequest
 			return
 		}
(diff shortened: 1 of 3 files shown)


endpoints/setuid.go
```diff
@@ -82,8 +84,17 @@ func NewSetUIDEndpoint(cfg *config.Configuration, syncersByBidder map[string]use
...
 		if len(fetchErrs) > 0 {
 			w.WriteHeader(http.StatusBadRequest)
-			w.Write([]byte(combineErrors(fetchErrs).Error()))
-			metricsEngine.RecordSetUid(metrics.SetUidBadRequest)
+			err := combineErrors(fetchErrs)
+			w.Write([]byte(err.Error()))
+			switch err {
+			case errCookieSyncAccountBlocked:
+				metricsEngine.RecordSetUid(metrics.SetUidAccountBlocked)
+			case errCookieSyncAccountInvalid:
```
bsardo, line 92: I figured it was ok to use the cookie sync `errCookieSyncAccountBlocked` and `errCookieSyncAccountInvalid` errors here since setuid is related to cookie sync instead of refactoring `combineErrors` or creating a similar function for use by setuid.

mansinahar (approved):

AlexBVolcy (approved):

PR #2338, opened 2022-08-03
AMP Endpoint: support parameters consentType and gdprApplies

guscarreon:
This pull request implements two out of the three AMP endpoint API GET fields described in issue #1817 
1. `consent_type=CONSENT_METADATA(consentStringType)`
2. `gdpr_applies=CONSENT_METADATA(gdprApplies)`

From the Details section of issue #1817 's description, this PR implements the following:
1. If `consent_type` is provided, it will be an enum containing the following values:
 - The contents of `gdpr_consent` can be treated as TCF V1. We no longer support TCFv1, so ignore any `consent_string` provided on the request.
 - The contents of `gdpr_consent` can be treated as TCF V2. If the consent_string isn't a valid TCF2 string, assume there's no user consent for any purpose as if no `gdpr_consent` were provided.
 - The contents of gdpr_consent can be treated as a US Privacy String. If the `consent_string` isn't a valid USP string, assume for the purposes of opt-out-of-sale enforcement that no `gdpr_consent` was provided.
 - Anything else: ignore, log a warning, and assume no `gdpr_consent` was provided

2. If `gdpr_applies` is supplied, use its value to set `regs.ext.gdpr`: `gdpr_applies=true` means `regs.ext.gdpr:1`, `gdpr_applies=false` means `regs.ext.gdpr:0`, any other value means `regs.ext.gdpr` is not set.
3. If `consent_type="2"`, and `gdpr_consent` is not empty, then copy `gdpr_consent` to `user.ext.consent`
4. If `consent_type="3"`, and `gdpr_consent` is not empty, then copy `gdpr_consent` to `regs.ext.us_privacy`

### Summary
The modifications shown below make up the entire logic behind this PR.
```
35   // ParseParams parses the AMP parameters from a HTTP request.
36   func ParseParams(httpRequest *http.Request) (Params, error) {
37       query := httpRequest.URL.Query()
38  
39       tagID := query.Get("tag_id")
40       if len(tagID) == 0 {
41           return Params{}, errors.New("AMP requests require an AMP tag_id")
42       }
43  
44       params := Params{
45           Account:      query.Get("account"),
46           CanonicalURL: query.Get("curl"),
47           Consent:      chooseConsent(query.Get("consent_string"), query.Get("gdpr_consent")),
48           Debug:        query.Get("debug") == "1",
49           Origin:       query.Get("__amp_source_origin"),
50           Size: Size{
51               Height:         parseInt(query.Get("h")),
52               Multisize:      parseMultisize(query.Get("ms")),
53               OverrideHeight: parseInt(query.Get("oh")),
54               OverrideWidth:  parseInt(query.Get("ow")),
55               Width:          parseInt(query.Get("w")),
56           },
57           Slot:            query.Get("slot"),
58           StoredRequestID: tagID,
59           Timeout:         parseIntPtr(query.Get("timeout")),
   +         ConsentType:     parseInt(query.Get("consent_type")),
   +         GdprApplies:     parseBoolPtr(query.Get("gdpr_applies")),
60           Targeting:       query.Get("targeting"),
61       }
62       return params, nil
63   }
amp/parse.go
```

```
 52 func ReadPolicy(ampParams Params, req *openrtb2.BidRequest, pbsConfigGDPREnabled bool) (privacy.PolicyWriter, error) {
 53
 54     if len(ampParams.Consent) == 0 {
 55         return privacy.NilPolicyWriter{}, nil
 56     }
 57
 58     var rv privacy.PolicyWriter = privacy.NilPolicyWriter{}
 59     var warning error
 60     var errMsg string
 61
 62     switch ampParams.ConsentType {
 63     case ConsentTCF2:
 64         if pbsConfigGDPREnabled {
 65             // Even if consent is invalid, we write the consent string to req.user.ext.consent
 66             rv = gdpr.ConsentWriter{ampParams.Consent}
 67
 68             if tcf2.IsConsentV2(ampParams.Consent) {
 69                 if _, err := tcf2.ParseString(ampParams.Consent); err != nil {
 70                     errMsg = err.Error()
 71                 }
 72             } else {
 73                 errMsg = fmt.Sprintf("Consent string '%s' is not a valid TCF2 consent string.", ampParams.Consent)
 74             }
 75         }
 76     case ConsentUSPrivacy:
 77         if ccpa.ValidateConsent(ampParams.Consent) {
 78             rv = ccpa.ConsentWriter{ampParams.Consent}
 79         } else {
 80             errMsg = fmt.Sprintf("Consent string '%s' is not a valid CCPA consent string.", ampParams.Consent)
 81         }
 82     case ConsentDenied:
 83         errMsg = "Consent denied. Consent string ignored."
 84     case ConsentTCF1:
 85         errMsg = "TCF1 consent is deprecated and no longer supported."
 86     default:
 87         errMsg = fmt.Sprintf("Consent '%d' is not recognized as either CCPA or GDPR TCF2.", ampParams.ConsentType)
 88     }
 89
 90     if len(errMsg) > 0 {
 91         warning = &errortypes.Warning{
 92             Message:     errMsg,
 93             WarningCode: errortypes.InvalidPrivacyConsentWarningCode,
 94         }
 95     }
 96
 97     return rv, warning
 98 }
amp/parse.go
endpoints/openrtb2/amp_auction.go
```
Everything else was written to complement, support, and test these changes.


endpoints/openrtb2/sample-requests/valid-whole/supplementary/aliased-buyeruids.json
```diff
@@ -1,5 +1,6 @@
 {
   "description": "Amp request where root.user.ext.prebid.buyeruids field makes use of alias defined in root.ext.prebid.aliases",
+  "query": "tag_id=101",
```
guscarreon, line 3: New JSON test field

endpoints/openrtb2/sample-requests/amp-query-values/gdpr-tcf2-consent-through-query.json
```diff
@@ -0,0 +1,90 @@
...
+        },
+        "ext": {
+          "appnexus": {
+            "placementId": 12883451
+          }
+        }
+      }
+    ]
+  },
+  "expectedValidatedBidRequest": {
```
guscarreon, line 31: New JSON test framework field

endpoints/openrtb2/test_utils.go
```diff
@@ -1086,12 +1089,35 @@ func (tc *testConfigValues) getAdaptersConfigMap() map[string]config.Adapter {
 	return adaptersConfig
 }
 
+// exchangeTestWrapper is a wrapper that asserts the openrtb2 bid request just before the HoldAuction call
```
guscarreon, line 1092: In the presence of `"expectedValidatedBidRequest"` in a given JSON test, this wrapper will assert this field right before passing the bidRequest to `exchange.HoldAuction()`

privacy/consentvalidator.go
```diff
@@ -0,0 +1,14 @@
+package privacy
+
+type ConsentValidator interface {
+	ValidateConsent(consent string) bool
```
guscarreon, line 4: Helps to create mocks in order to make the `readPolicy()` function's unit tests more robust

amp/parse.go
```diff
@@ -20,7 +20,11 @@ type Params struct {
 	Slot            string
 	StoredRequestID string
 	Timeout         *uint64
-	Targeting       string
+
```
SyntaxNode, line 23: Please remove the line break and alphabetize the list with the new fields added by this PR.

amp/parse.go
```diff
@@ -32,6 +36,14 @@ type Size struct {
 	Width          int64
 }
 
+// Policy consent types
+const (
+	NO_CONSENT = 0
+	TCF1       = 1
+	TCF2       = 2
+	CCPA       = 3
```
SyntaxNode, line 44: These constants should be prefaces with something which ties them to the consent type and Go does not use SCREAMING_CAPS for constant names. Perhaps something like: ConsentNon/ConsentUndefined, ConsentTCF1, ConsentTCF2, ConsentUSPrivacy

amp/parse.go
```diff
@@ -57,6 +69,9 @@ func ParseParams(httpRequest *http.Request) (Params, error) {
 		Slot:            query.Get("slot"),
 		StoredRequestID: tagID,
 		Timeout:         parseIntPtr(query.Get("timeout")),
+		ConsentType:     parseInt(query.Get("consent_type")),
+		GdprApplies:     parseBoolPtr(query.Get("gdpr_applies")),
+		AddtlConsent:    query.Get("addtl_consent"),
```
SyntaxNode, line 74: Please alphabetize this list.

amp/parse.go
```diff
@@ -76,6 +91,20 @@ func parseInt(value string) int64 {
 	return 0
 }
 
+func parseBoolPtr(value string) *bool {
```
SyntaxNode, line 94: Change to incorporate `strconv.ParseBool(x)` and return an error if there's a problem. We can accept empty string or a boolean value, anything else is an error.

amp/parse_test.go
```diff
@@ -94,6 +94,44 @@ func TestParseParams(t *testing.T) {
...
+
+	testCases := []struct {
+		desc     string
+		in       string
+		expected *bool
+	}{
+		{
+			"Input is an empty string",
+			"",
+			nil,
```
SyntaxNode, line 109: Please avoid using unnamed field setting, it's too hard to read.

amp/parse_test.go
```diff
@@ -94,6 +94,44 @@ func TestParseParams(t *testing.T) {
...
+		{
+			"Input is the word 'true', expect a reference pointing to true value",
+			"true",
+			&boolTrue,
+		},
+	}
+	for _, tc := range testCases {
+		actual := parseBoolPtr(tc.in)
+
+		// assertions
```
SyntaxNode, line 130: Nitpick: The comment is not necessary.

endpoints/openrtb2/amp_auction.go
```diff
@@ -443,7 +444,11 @@ func (deps *endpointDeps) overrideWithParams(ampParams amp.Params, req *openrtb2
 		req.Imp[0].TagID = ampParams.Slot
 	}
 
-	policyWriter, policyWriterErr := readPolicy(ampParams.Consent)
+	policy := policyReader{
```
SyntaxNode, line 447: I think it's good to make the policy reader more open, but please define it earlier and pass it down as a dependency instead of creating it right here just once when we could reuse for every auction.

SyntaxNode, line 447: Nitpick: The name `policy` is confusing when it's really a privacy reader.

endpoints/openrtb2/amp_auction.go
```diff
@@ -578,21 +588,46 @@ func setAmpExt(site *openrtb2.Site, value string) {
...
+	reqWrap := &openrtb_ext.RequestWrapper{BidRequest: req}
+	if regsExt, err := reqWrap.GetRegExt(); err == nil {
+		regsExt.SetGDPR(&gdpr)
+	} else {
+		return err
+	}
+	return reqWrap.RebuildRequest()
+}
+
+func (pr policyReader) readPolicy(ampParams amp.Params, req *openrtb2.BidRequest) (privacy.PolicyWriter, error) {
```
SyntaxNode, line 611: This is now amp specific, so please move it into the amp package.

endpoints/openrtb2/amp_auction.go
```diff
@@ -578,21 +588,46 @@ func setAmpExt(site *openrtb2.Site, value string) {
...
+
+	if err := setRegExtGDPR(ampParams.GdprApplies, req); err != nil {
```
SyntaxNode, line 613: I'd like to see writing the GDPR flag incorporated into the consentwriter. This will help to ensure we do not write the gdpr flag unless the we know this is a TCF1 or TCF2 consent type.

endpoints/openrtb2/amp_auction.go
```diff
@@ -578,21 +588,46 @@ func setAmpExt(site *openrtb2.Site, value string) {
 	}
 }
 
-func readPolicy(consent string) (privacy.PolicyWriter, error) {
-	if len(consent) == 0 {
+// setRegExtGDPR sets regs.ext.gdpr to either 0 or 1 only if it was set in the request query
+func setRegExtGDPR(gdprApplies *bool, req *openrtb2.BidRequest) error {
```
SyntaxNode, line 592: This code should live in the privacy/gdpr package - ideally within the consent writer.

endpoints/openrtb2/amp_auction.go
```diff
@@ -578,21 +588,46 @@ func setAmpExt(site *openrtb2.Site, value string) {
...
+		return privacy.NilPolicyWriter{}, err
+	}
+
+	if len(ampParams.Consent) == 0 {
 		return privacy.NilPolicyWriter{}, nil
 	}
 
-	if gdpr.ValidateConsent(consent) {
-		return gdpr.ConsentWriter{consent}, nil
+	if pr.gdpr.ValidateConsent(ampParams.Consent) || ampParams.ConsentType == amp.TCF2 {
```
SyntaxNode, line 621: Please update this logic to:

```
if ampParams.ConsentType == none/unknown/unspecified {
  ... code as it always was with existing error and with the gdpr consent writer now also writing the gdpr flag ...
} else {
  switch ampParams.ConsentType {
     case amp.TCF1: ... ignore or error, what does PBS-Java do? ....
     case amp.TCF2: ... validate consent, if error, return error ....
     case amp.USPrivacy: ... validate consent, if error, return error ....
     default: .. error or ignore, unknown consent string type ...
  }
}
```

If the consent type is specified, it should be strictly followed such that TCF1 is ignored or results in an error, TCF2

endpoints/openrtb2/auction_benchmark_test.go
```diff
@@ -155,7 +156,7 @@ func BenchmarkValidWholeExemplary(b *testing.B) {
 				AccountRequired:    test.Config.AccountRequired,
 			}
 
-			auctionEndpointHandler, mockBidServers, mockCurrencyRatesServer, err := buildTestEndpoint(test, cfg)
+			auctionEndpointHandler, mockBidServers, mockCurrencyRatesServer, err := buildTestEndpoint(&testing.T{}, test, cfg)
```
SyntaxNode, line 159: The creation of a new testing object looks suspicious. Can you get this from the benchmarking object?

openrtb_ext/request_wrapper.go
```diff
@@ -755,6 +755,8 @@ type RegExt struct {
 	extDirty       bool
 	usPrivacy      string
 	usPrivacyDirty bool
+	GDPR           *int8
```
SyntaxNode, line 758: Both new fields must be lower case since we don't want it exported.

openrtb_ext/request_wrapper.go
```diff
@@ -791,6 +797,19 @@ func (re *RegExt) marshal() (json.RawMessage, error) {
 		re.usPrivacyDirty = false
 	}
 
+	if re.GDPRDirty {
+		gdprJson, err := json.Marshal(*re.GDPR)
```
SyntaxNode, line 801: This can result in a panic. You should only call Marshal if `re.gdpr` is not nil, otherwise delete it from the map. This is a pattern we follow in other cases so you look around this file for an example.

privacy/ccpa/parsedpolicy.go
```diff
@@ -23,8 +23,10 @@ const (
 
 const allBiddersMarker = "*"
 
+type CCPAConsentValidator struct{}
```
SyntaxNode, line 26: This should just be called `Validator`.

privacy/consentvalidator.go
SyntaxNode: This is test code, so please only keep it in test files so it's not compiled with the app for release.

privacy/consentvalidator.go
```diff
@@ -0,0 +1,14 @@
...
+}
+
+// MockConsentValidator implements ConsentValidator and is useful for testing purposes
+type MockConsentValidator struct {
```
SyntaxNode, line 8: Nitpik: This is not a fake, not a mock. Please rename it accordingly. Let's keep the mock naming for when we use the testify mock library and provide the ability to set and test expectations.

privacy/gdpr/policy.go
```diff
@@ -10,8 +10,10 @@ type Policy struct {
 	Consent string
 }
 
+type GDPRConsentValidator struct{}
```
SyntaxNode, line 13: This should just be called `Validator`.

privacy/gdpr/policy.go
```diff
@@ -10,8 +10,10 @@ type Policy struct {
...
+
 // ValidateConsent returns true if the consent string is empty or valid per the IAB TCF spec.
-func ValidateConsent(consent string) bool {
+func (GDPRConsentValidator) ValidateConsent(consent string) bool {
```
SyntaxNode, line 16: Do we need the parentheses around the type name?

endpoints/openrtb2/amp_auction.go
guscarreon: Good catch. Corrected

endpoints/openrtb2/amp_auction.go
guscarreon: Good catch. Done

endpoints/openrtb2/amp_auction.go
guscarreon: That's the reason I wanted to bring this changes up for discussion in a PR. According to my interpretation of the specs, I understood this to be independent. Paraphrasing the relevant section as described in issue  #1817: `"2. If gdpr_applies is supplied, use its value to set regs.ext.gdpr: gdpr_applies=true means regs.ext.gdpr:1, gdpr_applies=false means regs.ext.gdpr:0, any other value means regs.ext.gdpr is not set."`. This seems to be TCF1 and TCF2 independent. Am I wrong?

endpoints/openrtb2/amp_auction.go
guscarreon: If we move to the consent writer, we'll need the `GDPR.Validate(ampParams.Consent)` call to be true and according to the issue #1817 specs, we don't need the consent string to be true in order to set `regs.ext.gdpr`, hence the reason it belongs on this package (I believe). Thoughts?

endpoints/openrtb2/amp_auction.go
guscarreon: Yep, this part was totally wrong. Logic has been corrected

endpoints/openrtb2/amp_auction.go
SyntaxNode: It might be... I'm really not sure.

endpoints/openrtb2/amp_auction.go
guscarreon: Moved to the `gdpr` package.

privacy/gdpr/policy.go
guscarreon: No need. This was removed.

privacy/gdpr/policy.go
guscarreon: This type has been removed

privacy/consentvalidator.go
guscarreon: This file has been removed

privacy/ccpa/parsedpolicy.go
guscarreon: Type has been removed

endpoints/openrtb2/amp_auction.go
guscarreon: > I'd like to see writing the GDPR flag incorporated into the consentwriter. This will help to ensure we do not write the gdpr flag unless the we know this is a TCF1 or TCF2 consent type.
The GDPR flag is independent of the consent according to issue #1817. Let me know if you agree with the approach of keeping the `setRegExtGDPR()` in the `amp_auction.go` endpoint as part of the request parsing process.

amp/parse.go
```diff
@@ -32,6 +41,62 @@ type Size struct {
 	Width          int64
 }
 
+// Policy consent types
+const (
+	ConsentDenied = iota
+	ConsentTCF1
+	ConsentTCF2
+	ConsentUSPrivacy
```
SyntaxNode, line 49: Please set the values directly instead of relying on `iota` because this map is defined by an external source and `iota` implies the values have no specific meaning. You may want to include in the comment that these values are specified externally from the AMP project.

amp/parse.go
```diff
@@ -32,6 +41,62 @@ type Size struct {
...
+)
+
+func ReadPolicy(ampParams Params, req *openrtb2.BidRequest, pbsConfigGDPREnabled bool) (privacy.PolicyWriter, error) {
+
```
SyntaxNode, line 53: Nitpick: Please remove leading blank line.

amp/parse.go
```diff
@@ -32,6 +41,62 @@ type Size struct {
...
+	var warning error
+	var errMsg string
+
+	switch ampParams.ConsentType {
+	case ConsentTCF2:
+		if pbsConfigGDPREnabled {
+			// Even if consent is invalid, we write the consent string to req.user.ext.consent
+			rv = gdpr.ConsentWriter{ampParams.Consent}
+
+			if tcf2.IsConsentV2(ampParams.Consent) {
```
SyntaxNode, line 68: Why do we validate the consent string here if we always put in the request? If we omit this, will downstream code put in a similar warning and/or error?

amp/parse.go
```diff
@@ -32,6 +41,62 @@ type Size struct {
...
+			}
+		}
+	case ConsentUSPrivacy:
+		if ccpa.ValidateConsent(ampParams.Consent) {
+			rv = ccpa.ConsentWriter{ampParams.Consent}
+		} else {
+			errMsg = fmt.Sprintf("Consent string '%s' is not a valid CCPA consent string.", ampParams.Consent)
+		}
+	case ConsentDenied:
+		errMsg = "Consent denied. Consent string ignored."
```
SyntaxNode, line 83: I'm curious, what does consent denied mean?

endpoints/openrtb2/amp_auction.go
```diff
@@ -443,7 +441,11 @@ func (deps *endpointDeps) overrideWithParams(ampParams amp.Params, req *openrtb2
 		req.Imp[0].TagID = ampParams.Slot
 	}
 
-	policyWriter, policyWriterErr := readPolicy(ampParams.Consent)
+	if err := setRegExtGDPR(ampParams.GdprApplies, req); err != nil {
+		return []error{err}
+	}
```
SyntaxNode, line 446: Can we incorporate the gdpr field into the ReadPolicy method?

endpoints/openrtb2/amp_auction.go
```diff
@@ -578,23 +580,24 @@ func setAmpExt(site *openrtb2.Site, value string) {
...
+func setRegExtGDPR(gdprApplies *bool, req *openrtb2.BidRequest) error {
+	if gdprApplies == nil {
+		return nil
 	}
 
-	if ccpa.ValidateConsent(consent) {
-		return ccpa.ConsentWriter{consent}, nil
+	gdpr := int8(0)
+	if *gdprApplies {
+		gdpr++
```
SyntaxNode, line 591: IMHO: It's better to set this to explicitly 1 instead of using the increment helper, since the value 1 has a specific  meaning.

amp/parse.go
```diff
@@ -32,6 +41,62 @@ type Size struct {
...
+	case ConsentTCF1:
+		errMsg = "TCF1 consent is deprecated and no longer supported."
+	default:
+		errMsg = fmt.Sprintf("Consent '%d' is not recognized as either CCPA or GDPR TCF2.", ampParams.ConsentType)
```
SyntaxNode, line 87: If there is consent type specified, we should fall back to the same logic of trying to parse it as TCF2 and CCPA. We don't want to just return an error message.

amp/parse.go
```diff
@@ -76,6 +143,14 @@ func parseInt(value string) int64 {
 	return 0
 }
 
+func parseBoolPtr(value string) *bool {
+	rv, err := strconv.ParseBool(value)
+	if err != nil {
+		return nil
```
SyntaxNode, line 149: We may want to reconsider this error handling. If the value is empty a return of nil makes sense, but if the value is not a boolean then perhaps a fatal error message is more appropriate.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -129,7 +228,7 @@ func TestAMPPageInfo(t *testing.T) {
 }
 
 func TestGDPRConsent(t *testing.T) {
-	consent := "BOu5On0Ou5On0ADACHENAO7pqzAAppY"
+	consent := "CPdiPIJPdiPIJACABBENAzCv_____3___wAAAQNd_X9cAAAAAAAA"
```
SyntaxNode, line 231: Why did you need to change this test consent string?

endpoints/openrtb2/amp_auction.go
SyntaxNode: Looks like you're missing the first if condition, to fallback to previous logic when the consent type is not provided.

endpoints/openrtb2/amp_auction_test.go
hhhjort: If I recall my consent strings, the old is TCF1 and the new is TCF2

amp/parse.go
hhhjort: Do you mean if there is no consent type specified? If the consent type is specified, then it should be an error if a different type of consent string is provided.

amp/parse.go
guscarreon: Makes sense. Done

amp/parse.go
guscarreon: Done

amp/parse.go
guscarreon: Renamed to `ConsentNone`

endpoints/openrtb2/amp_auction.go
guscarreon: Done

endpoints/openrtb2/amp_auction_test.go
guscarreon: After the logic change the consent string that was removed was no longer passing validation. I believe @hhhjort s right, it must be a TCF1 string.

endpoints/openrtb2/auction_benchmark_test.go
guscarreon: Done

openrtb_ext/request_wrapper.go
guscarreon: Makes sense. Corrected.

openrtb_ext/request_wrapper.go
guscarreon: Corrected

amp/parse.go
guscarreon: Agree Brian. The sole reason I included a GDPR string validation call is to return a warning. Let me know your take on the new logic.

endpoints/openrtb2/amp_auction.go
guscarreon: Done

amp/parse.go
guscarreon: Great catch. Implemented.

amp/parse.go
guscarreon: Done. Modified both `parseBoolPtr` and `parseIntPtr`

openrtb_ext/request_wrapper.go
```diff
@@ -791,6 +797,23 @@ func (re *RegExt) marshal() (json.RawMessage, error) {
...
+		if re.gdpr != nil {
+			gdprJson, err := json.Marshal(*re.gdpr)
+			if err != nil {
+				return nil, err
+			}
+			if len(gdprJson) > 0 {
+				re.ext["gdpr"] = json.RawMessage(gdprJson)
+			} else {
+				delete(re.ext, "gdpr")
+			}
```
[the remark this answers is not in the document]
guscarreon: Yes. Good catch. Removede unnecessary `if/else`

amp/parse.go
guscarreon: @bsardo the logic has been updated. Let me know your thoughts.

amp/parse.go
guscarreon: > Why do we validate the consent string here if we always put in the request?

In the discussions we had on how to implement this, we decided to leave invalid strings, issue a warning, and let the `consent` string malformedness end up not granting the permissions downstream.

>  If we omit this, will downstream code put in a similar warning and/or error?

It won't. This is the only warning on this regard. In `exchange/utils.go` we resolve the `gdprApplies` value and parse the permissions found in the consent string, but at no point do we exit with an error if the `consent` string is malformed.

amp/parse.go
```diff
@@ -32,6 +41,114 @@ type Size struct {
...
+	case ConsentTCF2:
+		if pbsConfigGDPREnabled {
+			rv = buildGdprTCF2ConsentWriter(ampParams)
+			// Log warning if GDPR consent string is invalid
+			warningMsg = validateTCf2ConsentString(ampParams.Consent)
+		}
+	case ConsentUSPrivacy:
+		rv = ccpa.ConsentWriter{ampParams.Consent}
+		if ccpa.ValidateConsent(ampParams.Consent) {
+			if parseGdprApplies(ampParams.GdprApplies) == 1 {
```
hhhjort, line 76: I was pondering, since gdpr is opt out and requires careful tracking of consent, do we want to default to no consent in case of this mismatch?

guscarreon: I think we should. Would make perfect sense. Should I still log the warning?

hhhjort: I spoke with Julien Delhommeau and he suggests that `consent_type` is reasonable to take as authoritative as to which consent framework applies to the request. So we are good with the current code.

hhhjort (approved):
LGTM

endpoints/openrtb2/sample-requests/valid-whole/supplementary/aliased-buyeruids.json
bsardo, line 3: Can we move these AMP specific tests into a new directory or rename one of the existing folders in the path so we know these are AMP tests?

bsardo (commented):
This is looking good. Just a few nitpicks and questions.

amp/parse.go
```diff
@@ -32,6 +41,114 @@ type Size struct {
...
+				// Log warning because AMP request comes with both a valid CCPA string and gdpr_applies set to true
+				warningMsg = "AMP request gdpr_applies value was ignored in account of provided consent string found to be CCPA and not GDPR."
+			}
+		} else {
+			// Log warning if CCPA string is invalid
+			warningMsg = fmt.Sprintf("Consent string '%s' is not a valid CCPA consent string.", ampParams.Consent)
+		}
+	case ConsentNone:
+		fallthrough
```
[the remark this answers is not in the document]
guscarreon: That's right. It's merely for readability. But I can remove if you want.

amp/parse.go
```diff
@@ -32,6 +41,114 @@ type Size struct {
...
+// GDPR applies, even if field gdpr_applies wasn't set in the AMP endpoint query.
+func buildGdprTCF2ConsentWriter(ampParams Params) gdpr.ConsentWriter {
+	writer := gdpr.ConsentWriter{Consent: ampParams.Consent}
+
+	// If gdpr_applies was not set, regs.ext.gdpr must equal 1
+	var gdprValue int8 = 1
+	if ampParams.GdprApplies != nil {
+		// set regs.ext.gdpr if non-nil gdpr_applies was set to true
+		gdprValue = parseGdprApplies(ampParams.GdprApplies)
+	}
```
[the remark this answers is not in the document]
guscarreon: > here it appears we default to a gdprValue of 1 if it is nil and in the other cases we are defaulting to a value of 0 since the nil check in the function will return 0. Is that correct?

Correct. That's the reason I added the `nil` check here and not in the other instances. From the specs and previous discussions we agreed that if `gdpr_applies` was not set, the `gdpr.ConsentWriter` must default to write `regs.ext.gdpr` with a value of 1. On the other hand, if `gdpr_applies` was set/non-nil, the `gdpr.ConsentWriter` must write into `regs.ext.gdpr` whatever value `gdpr_applies` holds.

endpoints/openrtb2/sample-requests/valid-whole/supplementary/aliased-buyeruids.json
guscarreon: Done. `sample-requests/valid-whole/supplementary/aliases.json` was duplicated into: `sample-requests/amp/valid-supplementary/aliases.json`. The latter comes with the `"query"` field, the former, doesn't.

hhhjort (approved):

bsardo (approved):
LGTM

Issue #1162, opened 2020-01-09
Prebid Server Price Floor Feature


bszekely1:
# Overview
The document describes the Prebid Server implementation of the Price Floor Feature as defined [here](https://github.com/prebid/Prebid.js/issues/4273). Prebid Server will contain its own set of requirements given the additional flexibility PBS has to execute code in a server-side environment.

Intended workflow will remain the same in PBS as in pbjs:

1. Publisher signs with a Price Floor provider
2. Price Floor provider makes a JSON file available with the data
3. Publisher configures Prebid to resolve and enforce the price floor

PBS should support Prebid.js, Prebid Server, and Prebid SDK.

Please see the [Product Requirements Document (PRD)](https://docs.google.com/document/d/1YAt00RfQaU2Ugciqw1LnirLgjyUz9ibfyjfxWUHR_P0/edit#) for detail.

hhhjort:
Not really specified above, but my feelings towards account scope data is as follows:

For options that are controlled by the publisher or by the PBS host's business logic, it is best to use an object stored directly in the request, and have that data filled in by the stored request backend. That will then hook the configuration into the existing publisher config system rather than adding a second location where some publisher configs will live.

For options that the PBS host will need to control directly without option for the publisher to override (for example account white/blacklists), PBS configuration option rather than request object.

So for the above flags, I would recommend all the flags be settable in the request, which would by default allow all of them to be auction specific in theory, while the stored request solution could limit some to an account level setting. At least outside of a publisher overriding what the stored request system supplied using a properly crafted request.

albertmolinermrf:
Is the list of supported fields for floor schema final? For example, the examples contain "impId" and "imp.id", but none is explicitly listed there. Additionally, are the values case-sensitive? If so, can we print them explicitly?

Is there an expected resolution algorithm when multiple keys are match? First wins, last wins, most specific wins...

bszekely1:
@hhhjort Thank you, i added better definition in the description to keep context of host-wide flags or stored request level flags. If you have any feedback, I am open to suggestions.

@albertmolinermrf 
> Is the list of supported fields for floor schema final?

The list is not final. This is simply in the intent to implement phase and has not started. The idea is to make the system flexible to fit most use cases. In this first pass we wanted to start with some basic fields that made sense. We didn't want to overload the development from the start.

>  are the values case-sensitive?

[...]

PR #2341, opened 2022-08-04
[Draft] floors feature implementation

PubMatic-OpenWrap:
Started development for floors feature as mentioned in PRD https://github.com/prebid/prebid-server/issues/1162.
This PR contains changes for floors feature and mainly covers basic implementation of Signalling and Enforcement.


floors/floors.go
```diff
@@ -0,0 +1,135 @@
...
+
+// IsRequestEnabledWithFloor will check if floors is enabled in request
+func IsRequestEnabledWithFloor(Floors *openrtb_ext.PriceFloorRules) bool {
+
+	if Floors != nil && Floors.Enabled != nil && !*Floors.Enabled {
+		return *Floors.Enabled
+	}
+
+	return true
+}
```
SyntaxNode, line 68: This might be better expressed as a method on `*openrtb_ext.PriceFloorRules` and perhaps named `GetEnabled`.

floors/floors.go
```diff
@@ -0,0 +1,135 @@
...
+
+// UpdateImpsWithFloors will validate floor rules, based on request and rules prepares various combinations
+// to match with floor rules and selects appripariate floor rule and update imp.bidfloor and imp.bidfloorcur
+func UpdateImpsWithFloors(floorExt *openrtb_ext.PriceFloorRules, request *openrtb2.BidRequest, conversions currency.Conversions) []error {
```
SyntaxNode, line 72: Nitpick: Might be better named `ModifyImpsWithFloors` to make it absolutely clear it's modifying input.

floors/floors.go
```diff
@@ -0,0 +1,135 @@
...
+	var (
+		floorErrList      []error
+		floorModelErrList []error
+		floorVal          float64
+	)
+	floorData := floorExt.Data
+	if floorData == nil {
+		return floorModelErrList
```
SyntaxNode, line 80: This is a clever ways of saying `return nil`, so perhaps just return that directly.

floors/floors.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	floorModelErrList = validateFloorSkipRates(floorExt)
+	if len(floorModelErrList) > 0 {
+		return floorModelErrList
+	}
+
+	floorData.ModelGroups, floorModelErrList = validateFloorModelGroups(floorData.ModelGroups)
+	if len(floorData.ModelGroups) == 0 {
```
SyntaxNode, line 89: Shouldn't this be checking if floorModelErrList is not empty?

floors/floors.go
```diff
@@ -0,0 +1,135 @@
...
+		return floorModelErrList
+	} else if len(floorData.ModelGroups) > 1 {
```
SyntaxNode, line 91: Nitpick: No need to us else if the other case is a return . :)

floors/floors.go
```diff
@@ -0,0 +1,135 @@
...
+	if floorData.ModelGroups[0].Schema.Delimiter == "" {
+		floorData.ModelGroups[0].Schema.Delimiter = DEFAULT_DELIMITER
+	}
+
+	floorExt.Skipped = new(bool)
+	if shouldSkipFloors(floorExt.Data.ModelGroups[0].SkipRate, floorExt.Data.SkipRate, floorExt.SkipRate, rand.Intn) {
+		*floorExt.Skipped = true
+		floorData.ModelGroups = nil
+		return floorModelErrList
+	}
```
SyntaxNode, line 104: Just wanted to say that I really like this code block.

floors/rule.go
```diff
@@ -0,0 +1,382 @@
...
+	floorCur := "USD"
+	if floorExt.Data.Currency != "" {
+		floorCur = floorExt.Data.Currency
+	}
+
+	if floorExt.Data.ModelGroups[0].Currency != "" {
+		floorCur = floorExt.Data.ModelGroups[0].Currency
+	}
+	return floorCur
+}
```
SyntaxNode, line 43: Looks like this could be used to help resolve my earlier comment about USD currency.

floors/rule.go
```diff
@@ -0,0 +1,382 @@
...
+		if winWeight <= 0 {
+			modelGroups[0], modelGroups[i] = modelGroups[i], modelGroups[0]
+			modelGroups[0].DebugWeight = debugWeight
+			return modelGroups[:1]
+		}
+	}
+	return modelGroups[:1]
+}
+
+func shouldSkipFloors(ModelGroupsSkipRate, DataSkipRate, RootSkipRate int, f func(int) int) bool {
```
SyntaxNode, line 91: Nitpick: Per Go naming convention, all function variables should start with a lower case letter.

floors/rule.go
```diff
@@ -0,0 +1,382 @@
...
+func findRule(RuleValues map[string]float64, delimiter string, desiredRuleKey []string, numFields int) string {
+
+	ruleKeys := prepareRuleCombinations(desiredRuleKey, numFields, delimiter)
+	for i := 0; i < len(ruleKeys); i++ {
+		if _, ok := RuleValues[ruleKeys[i]]; ok {
+			return ruleKeys[i]
+		}
+	}
+	return ""
+}
```
SyntaxNode, line 113: Consider returning a second bool variable for if it was found, similar to how Go does it for map key lookup. This removes the information implicitly carried by the empty string.

floors/rule.go
```diff
@@ -0,0 +1,382 @@
...
+		for k := 0; k < len(comb[i]); k++ {
+			wt1 += 1 << (segNum - comb[i][k])
+		}
+
+		wt2 := 0
+		for k := 0; k < len(comb[j]); k++ {
+			wt2 += 1 << (segNum - comb[j][k])
+		}
+		return wt1 < wt2
+	})
```
SyntaxNode, line 379: Consider extracting the sort algorithm and testing it separately.

openrtb_ext/floors.go
```diff
@@ -0,0 +1,45 @@
+package openrtb_ext
+
+type PriceFloorSchema struct {
+	Fields    []string `json:"fields,omitempty"`
+	Delimiter string   `json:"delimiter,omitempty"`
+}
```
SyntaxNode, line 6: Please be consistent with a line break between structs.

openrtb_ext/floors.go
```diff
@@ -0,0 +1,45 @@
+package openrtb_ext
+
+type PriceFloorSchema struct {
```
SyntaxNode, line 3: Please add comments for where within the openrtb request these apply to.

SyntaxNode (commented):
Thanks for the contribution. This is a large PR, so I have a large amount of comments. :)

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/golang/glog"
+	"github.com/mxmCherry/openrtb/v16/openrtb2"
+
+	"github.com/prebid/prebid-server/currency"
+	"github.com/prebid/prebid-server/floors"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func EnforceFloorToBids(bidRequest *openrtb2.BidRequest, seatBids map[openrtb_ext.BidderName]*pbsOrtbSeatBid, conversions currency.Conversions, enforceDealFloors bool) (map[openrtb_ext.BidderName]*pbsOrtbSeatBid, []string) {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: For accessing seatbids from each bidder, pbsOrtbSeatBid is required and pbsOrtbSeatBid is private in exchange package, hence could not move into floors package.
If we make pbsOrtbSeatBid public then we can this function into floors package.

config/config.go
```diff
@@ -131,6 +140,9 @@ func (cfg *Configuration) validate(v *viper.Viper) []error {
 	if cfg.AccountDefaults.Events.Enabled {
 		glog.Warning(`account_defaults.events will currently not do anything as the feature is still under development. Please follow https://github.com/prebid/prebid-server/issues/1725 for more updates`)
 	}
+	if cfg.PriceFloors.Enabled {
+		glog.Warning(`PriceFloors.Enabled will enforce floor feature which is still under development.`)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: we can move into Experiment struct

config/config.go
```diff
@@ -97,6 +97,15 @@ type Configuration struct {
...
 	// Experiment configures non-production ready features.
 	Experiment Experiment `mapstructure:"experiment"`
+	// Floors feature configuration
+	PriceFloors PriceFloors `mapstructure:"price_floors"`
+}
+
+type PriceFloors struct {
+	Enabled           bool `mapstructure:"enabled"`
+	UseDynamicData    bool `mapstructure:"use_dynamic_data"`
+	EnforceFloorsRate int  `mapstructure:"enforce_floors_rate"`
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: This global level config, which defines percent chance that PBS should suppress any bids below the matched floor from entering the auction. 
e.g. if EnforceFloorsRate = 100, floors enforcement would be applied to all the bids, 
if EnforceFloorsRate = 50, floors enforcement would be applied to 50% of the bids

config/config_test.go
```diff
@@ -386,6 +392,11 @@ request_validation:
     ipv4_private_networks: ["1.1.1.0/24"]
     ipv6_private_networks: ["1111::/16", "2222::/16"]
 generate_bid_id: true
+price_floors:
+   enabled: true
+   use_dynamic_data: false
+   enforce_floors_rate: 100
+   enforce_deal_floors: true
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Default config is tested in TestDefaults

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	var rejections []string
+	impMap := make(map[string]bidFloor)
+
+	//Maintaining BidRequest Impression Map
+	for i := range bidRequest.Imp {
+		var bidfloor bidFloor
+		bidfloor.bidFloorCur = bidRequest.Imp[i].BidFloorCur
+		if bidfloor.bidFloorCur == "" {
+			bidfloor.bidFloorCur = "USD"
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: In case floor currency not provided, can we set it to req.curr[0]  to avoid hardcoding ?

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/prebid/prebid-server/floors"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func EnforceFloorToBids(bidRequest *openrtb2.BidRequest, seatBids map[openrtb_ext.BidderName]*pbsOrtbSeatBid, conversions currency.Conversions, enforceDealFloors bool) (map[openrtb_ext.BidderName]*pbsOrtbSeatBid, []string) {
+
+	type bidFloor struct {
+		bidFloorCur string
+		bidFloor    float64
+	}
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: This map is introduced to map each ImpID from request to imp in SeatBids which is  map[openrtb_ext.BidderName]*pbsOrtbSeatBid.
Did not get any direct mapping, let us know if better approach can be implemented

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+	var rejections []string
+	impMap := make(map[string]bidFloor)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: added length

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+	var errs []error
+	requestExt, err := r.BidRequestWrapper.GetRequestExt()
+	if err != nil {
+		errs = append(errs, err)
+		return errs
+	}
+	prebidExt := requestExt.GetPrebid()
+	if floor != nil && floor.Enabled() && floors.IsRequestEnabledWithFloor(prebidExt.Floors) && prebidExt.Floors != nil {
+		errs = floors.UpdateImpsWithFloors(prebidExt.Floors, r.BidRequestWrapper.BidRequest, conversions)
+		requestExt.SetPrebid(prebidExt)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: prebidExt provided in request might contain multiple floor modules.
After floors selection process selectFloorModelGroup(), only selected module is updated into prebidExt.

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+		err := r.BidRequestWrapper.RebuildRequest()
+		if err != nil {
+			errs = append(errs, err)
+		}
+		updatedBidReq, _ := json.Marshal(r.BidRequestWrapper.BidRequest)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Agree, 
This is introduced to know details about selected floor module from given module, Only selected module would be present in prebidExt. 
json.Marshall() can be done only in case of debug.

exchange/exchange.go
```diff
@@ -955,6 +969,7 @@ func (e *exchange) makeExtBidResponse(adapterBids map[openrtb_ext.BidderName]*pb
 		bidResponseExt.Debug = &openrtb_ext.ExtResponseDebug{
 			HttpCalls:       make(map[openrtb_ext.BidderName][]*openrtb_ext.ExtHttpCall),
 			ResolvedRequest: r.ResolvedBidRequest,
+			UpdatedRequest:  r.UpdatedBidRequest,
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: ResolvedRequest contains request resolved by prebid server, however UpdatedRequest contains information after floors module selection and information updated after floors Enforcement.

floors/enforce.go
```diff
@@ -0,0 +1,45 @@
+package floors
+
+import (
+	"github.com/mxmCherry/openrtb/v16/openrtb2"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func ShouldEnforceFloors(bidRequest *openrtb2.BidRequest, floorExt *openrtb_ext.PriceFloorRules, configEnforceRate int, f func(int) int) bool {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: updated

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+
+		}
+		seatBids[bidderName].bids = eligibleBids
+
+	}
+
+	return seatBids, rejections
+}
+
+func SignalFloors(r *AuctionRequest, floor floors.Floor, conversions currency.Conversions, responseDebugAllow bool) []error {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: added description

floors/floors.go
```diff
@@ -0,0 +1,135 @@
...
+	SKIP_RATE_MIN          int    = 0
+	SKIP_RATE_MAX          int    = 100
+	MODEL_WEIGHT_MAX_VALUE int    = 100
+	MODEL_WEIGHT_MIN_VALUE int    = 0
+	ENFORCE_RATE_MIN       int    = 0
+	ENFORCE_RATE_MAX       int    = 100
+)
+
+type FloorConfig struct {
+	FloorEnabled      bool
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Updated

floors/floors.go
PubMatic-OpenWrap: added GetEnabled() method

floors/floors.go
PubMatic-OpenWrap: updated to ModifyImpsWithFloors()

floors/floors.go
PubMatic-OpenWrap: there is possibility that out of multiple modelGroups, some modelGroups can be invalid or some of the parameters can be invalid.
There can be case where in some model groups got rejected due to invalid parameters (i.e. Errors are present in floorModelErrList) and one or more model groups are valid, in this case we need to process valid model groups.
If all the model groups are rejected due to validation in that case we would be returning with floorModelErrList

floors/floors.go
PubMatic-OpenWrap: if more than one model groups are present then single model group needs to be selected, 
Hence else if is added

floors/rule.go
PubMatic-OpenWrap: Earlier comment was in Enforcement and this logic is for getting currency from floor json

exchange/floors.go
```diff
@@ -0,0 +1,131 @@
...
+			bidfloor.bidFloorCur = "USD"
+		}
+		bidfloor.bidFloor = bidRequest.Imp[i].BidFloor
+		impMap[bidRequest.Imp[i].ID] = bidfloor
+	}
+
+	for bidderName, seatBid := range seatBids {
+		eligibleBids := make([]*pbsOrtbBid, 0)
+		for bidInd := range seatBid.bids {
+			bid := seatBid.bids[bidInd]
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: This block compares each bid received against bid floor price considering bid floor currency normalisation and 
Enforcement of bid floor based on enforceDealFloors flag

exchange/floors.go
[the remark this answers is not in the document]
PubMatic-OpenWrap: Added comment

floors/floors.go
PubMatic-OpenWrap: Addressed review comment

floors/floors.go
PubMatic-OpenWrap: Thank you

floors/rule.go
PubMatic-OpenWrap: Renamed function

floors/rule.go
PubMatic-OpenWrap: This is getting tested in TestPrepareRuleCombinations(),
Let us know if it serves purpose.

openrtb_ext/floors.go
PubMatic-OpenWrap: Addressed

openrtb_ext/floors.go
PubMatic-OpenWrap: Added comment

config/config.go
```diff
@@ -1217,6 +1218,11 @@ func SetupViper(v *viper.Viper, filename string) {
 	v.SetDefault("experiment.adscert.inprocess.domain_renewal_interval_seconds", 30)
 	v.SetDefault("experiment.adscert.remote.url", "")
 	v.SetDefault("experiment.adscert.remote.signing_timeout_ms", 5)
+
+	v.SetDefault("experiment.price_floors.enabled", true)
```
SyntaxNode, line 1222: Please set this to `false` by default.

config/config.go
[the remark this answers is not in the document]
SyntaxNode: Gotcha. Can you please add a comment explaining that it's a random chance from 0 to 100.

exchange/exchange.go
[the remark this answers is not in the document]
SyntaxNode: Does PBS-Java also add an updated request to the debug info?

exchange/floors.go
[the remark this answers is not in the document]
SyntaxNode: A map of impressions by impression ID is a good approach. I don't think we need the `bidFloor` struct. Can this be `impMap := make(map[string]openrtb2.Imp, len(...)) instead?

exchange/floors.go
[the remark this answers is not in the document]
SyntaxNode: Please consider extracting the code within the loop to it's own method to make it easer to read / maintain / test.

exchange/exchange.go
```diff
@@ -293,6 +301,12 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 	var bidResponseExt *openrtb_ext.ExtBidResponse
 	if anyBidsReturned {
 
+		//If floor enforcement config enabled then filter bids
+		adapterBids, bidRejections := enforceFloors(&r, adapterBids, e.floor, conversions, responseDebugAllow)
```
SyntaxNode, line 305: If it's easier to return a slice of errors instead of bidRejections, feel free to go for it.

exchange/floors.go
```diff
@@ -0,0 +1,134 @@
...
+	if floor != nil && floor.IsFloorEnabled() && prebidExt.Floors.GetEnabled() {
+		if floors.ShouldEnforce(r.BidRequestWrapper.BidRequest, prebidExt.Floors, floor.GetEnforceRate(), rand.Intn) {
+			var enforceDealFloors bool
+			if prebidExt != nil && prebidExt.Floors != nil && prebidExt.Floors.Enforcement != nil && prebidExt.Floors.Enforcement.FloorDeals != nil {
+				enforceDealFloors = *prebidExt.Floors.Enforcement.FloorDeals && floor.EnforceDealFloor()
+			}
+			seatBids, rejections = enforceFloorToBids(r.BidRequestWrapper.BidRequest, seatBids, conversions, enforceDealFloors)
+		}
+		requestExt.SetPrebid(prebidExt)
+		err = r.BidRequestWrapper.RebuildRequest()
```
SyntaxNode, line 121: `enforceFloors` is only called from `HoldAuction`. It's odd to see a change to the request happening here after it's already been sent to the bidder. I also find it confusing that `prebidExt` would be modified in `floors.ShouldEnforce`.

All code which modifies a request for a specific bidder is contained in `cleanOpenRTBRequest`. Could you please refactor this code to move this logic there? ... except for `enforceFloorToBids` as since that operates on the bids it needs to stay in this location.

floors/enforce.go
```diff
@@ -0,0 +1,45 @@
...
+		var floorInRequest bool
+		for i := range bidRequest.Imp {
+			if bidRequest.Imp[i].BidFloor > 0 {
+				floorInRequest = true
+				break
+			}
+		}
+		if !floorInRequest {
+			return floorInRequest
+		}
```
SyntaxNode, line 20: Consider extracting to a method, like:

```
func requestHasFloors(bidRequest *openrtb2.BidRequest) bool {
	for i := range bidRequest.Imp {
		if bidRequest.Imp[i].BidFloor > 0 {
			return true
		}
	}
	
	return false
}
```

Then you can use return statements, describe the functionality via the function name, and make it easier to isolate for testing.

floors/enforce.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	if floorExt.Enforcement == nil {
+		floorExt.Enforcement = new(openrtb_ext.PriceFloorEnforcement)
+	}
+
+	if floorExt.Enforcement.EnforcePBS == nil {
+		floorExt.Enforcement.EnforcePBS = new(bool)
+	}
+	*floorExt.Enforcement.EnforcePBS = shouldEnforce
```
SyntaxNode, line 43: Why does this update the floorExt in addition to returning the value? Can we just use the return value and not modify the request?

floors/floors.go
SyntaxNode: Wouldn't we also want to report validation errors for the model groups which were rejected? We could send them down as warnings.

floors/validate.go
```diff
@@ -0,0 +1,57 @@
+package floors
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+func validateFloorRules(Schema openrtb_ext.PriceFloorSchema, delimiter string, RuleValues map[string]float64) []error {
```
SyntaxNode, line 10: Please rename to something which indicates the rule values are modified. I would normally expect this to just return a list of errors, but it's also modifying the floors and filtering our invalid ones. Better yet, consider returning a new map of rules instead of modifying the original collection - and reassign after.

floors/floors.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	floorData.ModelGroups, floorModelErrList = validateFloorModelGroups(floorData.ModelGroups)
+	if len(floorData.ModelGroups) == 0 {
+		return floorModelErrList
+	} else if len(floorData.ModelGroups) > 1 {
+		floorData.ModelGroups = selectFloorModelGroup(floorData.ModelGroups, rand.Intn)
+	}
+
+	if floorData.ModelGroups[0].Schema.Delimiter == "" {
```
SyntaxNode, line 85: Consider assigning a model group variable, since there is only one left at this point in the code. Something like:
```
modelGroup := floorData.ModelGroups[0]
```

floors/rule.go
```diff
@@ -0,0 +1,381 @@
...
+		floorExt.FloorMinCur != floorCur {
+		rate, err = conversions.GetRate(floorExt.FloorMinCur, floorCur)
+		floorMin = rate * floorExt.FloorMin
+	}
+	return floorMin, floorCur, err
+}
+
+func updateImpExtWithFloorDetails(matchedRule string, imp *openrtb2.Imp, floorVal float64) {
+	imp.Ext, _ = jsonparser.Set(imp.Ext, []byte(`"`+matchedRule+`"`), "prebid", "floors", "floorRule")
+	imp.Ext, _ = jsonparser.Set(imp.Ext, []byte(fmt.Sprintf("%.4f", floorVal)), "prebid", "floors", "floorRuleValue")
```
SyntaxNode, line 61: You may not use `jsonparser.Set`. it is not production ready code and has caused panics for us in the past.

openrtb_ext/floors.go
```diff
@@ -0,0 +1,56 @@
...
+	Skipped     *bool                  `json:"skipped,omitempty"`
+}
+
+// GetEnabled will check if floors is enabled in request
+func (Floors *PriceFloorRules) GetEnabled() bool {
+	if Floors != nil && Floors.Enabled != nil && !*Floors.Enabled {
+		return *Floors.Enabled
+	}
+	return true
+}
```
SyntaxNode, line 56: Please reorganized these structs such that the "parent" struct is listed before the "child" struct. I would expect to see `PriceFloorRules` declared first and for `PriceFloorSchema` to be declared after `PriceFloorModelGroup`, etc..

SyntaxNode (commented):
Thank for replying to the feedback so far. Sorry for the delay in a follow up review. There are more PRs submitted than usual this summer.

I focused this review on logic flow. I left out nitpicks and tests for now. I'd like to see alignment of the floorsExt setting with how bidder specific requests are typically created within the codebase. Please see me other comments for details.

config/config.go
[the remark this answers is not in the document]
PubMatic-OpenWrap: added

exchange/floors.go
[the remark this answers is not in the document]
PubMatic-OpenWrap: yes we can access req.curr[0]

exchange/floors.go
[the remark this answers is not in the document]
PubMatic-OpenWrap: removed bidFloor struct.

exchange/floors.go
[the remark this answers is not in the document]
PubMatic-OpenWrap: restructured function

config/config.go
PubMatic-OpenWrap: Updated

config/experiment.go
```diff
@@ -87,5 +98,9 @@ func (cfg *Experiment) validate(errs []error) []error {
 			errs = append(errs, fmt.Errorf("%s: %d", ErrMsgInvalidRemoteSignerSigningTimeout, cfg.AdCerts.Remote.SigningTimeoutMs))
 		}
 	}
+
+	if cfg.PriceFloors.Enabled {
+		glog.Warning(`PriceFloors.Enabled will enforce floor feature which is still under development.`)
+	}
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Added validation

exchange/floors.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		seatBids[bidderName].bids = eligibleBids
+	}
+	return seatBids, rejections
+}
+
+// signalFloors function does singanlling of floors,
+// Internally validation of floors parameters and validation of rules is done,
+// Based on number of modelGroups and modelWeight, one model is selected and imp.bidfloor and imp.bidfloorcur is updated
+func signalFloors(r *AuctionRequest, floor floors.Floor, conversions currency.Conversions, responseDebugAllow bool) []error {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: renamed function to selectFloorsAndModifyImp()

exchange/exchange.go
PubMatic-OpenWrap: returned slice of errors

floors/enforce.go
PubMatic-OpenWrap: Thank you for suggestion, 
Even I thought of adding same function 😄

floors/enforce.go
PubMatic-OpenWrap: This update is done to handle below case:
When Enforcement.EnforcePBS  is not defined and based on EnforceRate enforcement needs to be done or skipped, 
In that case this value is updated with appropriate value. 
And to update prebidExt due to this case requestExt.SetPrebid(prebidExt) is done in enforceFloors()

exchange/floors.go
PubMatic-OpenWrap: refer https://github.com/prebid/prebid-server/pull/2341#discussion_r958319263

floors/validate.go
PubMatic-OpenWrap: Renamed function to selectValidFloorModelGroups()

floors/rule.go
```diff
@@ -0,0 +1,381 @@
...
+	skipRate := 0
+
+	if ModelGroupsSkipRate > 0 {
+		skipRate = ModelGroupsSkipRate
+	} else if DataSkipRate > 0 {
+		skipRate = DataSkipRate
+	} else {
+		skipRate = RootSkipRate
+	}
+	return skipRate > f(SKIP_RATE_MAX+1)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: corrected

floors/floors.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (fc *Config) GetEnforceRate() int {
+	return fc.EnforceRate
+}
+
+func (fc *Config) EnforceDealFloor() bool {
+	return fc.EnforceDealFloors
+}
+
+type Floor interface {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: no other implementation. 
If we do not use Interface, then we might have to return config struct member values

floors/validate.go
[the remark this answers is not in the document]
PubMatic-OpenWrap: updated

floors/validate.go
```diff
@@ -0,0 +1,57 @@
...
+	var errs []error
+	for key, val := range RuleValues {
+		parsedKey := strings.Split(key, delimiter)
+		if len(parsedKey) != len(Schema.Fields) {
+			// Number of fields in rule and number of schema fields are not matching
+			errs = append(errs, fmt.Errorf("Invalid Floor Rule = '%s' for Schema Fields = '%v'", key, Schema.Fields))
+			delete(RuleValues, key)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: nice catch

floors/rule.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func findRule(RuleValues map[string]float64, delimiter string, desiredRuleKey []string, numFields int) (string, bool) {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: done

exchange/floors.go
```diff
@@ -0,0 +1,134 @@
...
+
+	var errs []error
+	requestExt, err := r.BidRequestWrapper.GetRequestExt()
+	if err != nil {
+		errs = append(errs, err)
+		return errs
+	}
+	prebidExt := requestExt.GetPrebid()
+	if floor != nil && floor.IsFloorEnabled() && prebidExt.Floors.GetEnabled() && prebidExt.Floors != nil {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: updated

exchange/exchange.go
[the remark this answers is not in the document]
PubMatic-OpenWrap: To include changes made by floors feature into resolved request, code https://github.com/prebid/prebid-server/blob/288e22c862cb4dff41dcac093160f3c54852618f/exchange/exchange.go#L216 needs to be moved after processing of floors data.

floors/floors.go
```diff
@@ -0,0 +1,124 @@
...
+	CATCH_ALL              string = "*"
+	SKIP_RATE_MIN          int    = 0
+	SKIP_RATE_MAX          int    = 100
+	MODEL_WEIGHT_MAX_VALUE int    = 100
+	MODEL_WEIGHT_MIN_VALUE int    = 0
+	ENFORCE_RATE_MIN       int    = 0
+	ENFORCE_RATE_MAX       int    = 100
+)
+
+type Config struct {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: To implement below interface this structure was introduced

exchange/floors.go
```diff
@@ -0,0 +1,145 @@
...
+	requestExt, err := r.BidRequestWrapper.GetRequestExt()
+	if err != nil {
+		errs = append(errs, err)
+		return errs
+	}
+	prebidExt := requestExt.GetPrebid()
+	if floor.Enabled && prebidExt.Floors != nil && prebidExt.Floors.GetEnabled() {
+		errs = floors.ModifyImpsWithFloors(prebidExt.Floors, r.BidRequestWrapper.BidRequest, conversions)
+		requestExt.SetPrebid(prebidExt)
+		err := r.BidRequestWrapper.RebuildRequest()
```
SyntaxNode, line 98: FYI: This RebuildRequest call is necessary for now. We'll be able to remove it in the near future. I'm working on bringing in the full use of the request wrapper into hold auction code.

floors/enforce.go
SyntaxNode: No other code in this PR check for `EnforcePBS`. Is this therefore a signal downstream to bidders so they know what behavior PBS performed? @bretg checking on the spec intention here.

floors/floors.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	DEFAULT_DELIMITER      string = "|"
+	CATCH_ALL              string = "*"
+	SKIP_RATE_MIN          int    = 0
+	SKIP_RATE_MAX          int    = 100
+	MODEL_WEIGHT_MAX_VALUE int    = 100
+	MODEL_WEIGHT_MIN_VALUE int    = 0
+	ENFORCE_RATE_MIN       int    = 0
+	ENFORCE_RATE_MAX       int    = 100
```
SyntaxNode, line 21: Nitpick: In Go, it's common to use CamelCase for constants instead of CAPS. ie DEFAULT_DELIMITER -> DefaultDelimiter. This is largely due to the capital letter at the start of the variable name indicating if the type is package private or exported. If no other package needs these constants, please start with a lower case letter.

floors/rule.go
SyntaxNode: You need to change this to use the standard Go json library. We've tried to use `jsonparser.Set`before and it seg faulted on us in certain cases.

floors/rule.go
SyntaxNode: @bsardo what is your opinion?

floors/validate.go
```diff
@@ -0,0 +1,58 @@
...
+	var errs []error
+	for key, val := range ruleValues {
+		parsedKey := strings.Split(key, delimiter)
+		if len(parsedKey) != len(schema.Fields) {
+			// Number of fields in rule and number of schema fields are not matching
+			errs = append(errs, fmt.Errorf("Invalid Floor Rule = '%s' for Schema Fields = '%v'", key, schema.Fields))
+			delete(ruleValues, key)
+			continue
+		}
+		delete(ruleValues, key)
```
SyntaxNode, line 20: Since delete will always happen, should it be moved up before the if statement?

floors/validate.go
```diff
@@ -0,0 +1,58 @@
...
+	if floorExt.Data != nil && (floorExt.Data.SkipRate < SKIP_RATE_MIN || floorExt.Data.SkipRate > SKIP_RATE_MAX) {
+		errs = append(errs, fmt.Errorf("Invalid SkipRate at data level = '%v'", floorExt.Data.SkipRate))
+		return errs
+	}
+
+	if floorExt.SkipRate < SKIP_RATE_MIN || floorExt.SkipRate > SKIP_RATE_MAX {
+		errs = append(errs, fmt.Errorf("Invalid SkipRate at root level = '%v'", floorExt.SkipRate))
+	}
+	return errs
+}
```
SyntaxNode, line 39: validateFloorSkipRates will only return a single error. I recommend to change the return type to the singular `error` and the overall flow to:

```
if ..
  return fmt.Errorf(..))

if ...
  return fmt.Errorf(..))

return nil
```

SyntaxNode (commented):
Thank you for the updates and your patience for a re-review. Looking a lot better IMHO after the changes. I'm in the middle of reviewing tests and comparing the expected behavior against the spec. I hope to finish that mid next week.

SyntaxNode:
We recently fixed a security issue with a dependency. Please merge from master to resolve.

PubMatic-OpenWrap:
> Thank you for the updates and your patience for a re-review. Looking a lot better IMHO after the changes. I'm in the middle of reviewing tests and comparing the expected behavior against the spec. I hope to finish that mid next week.

Thank you for reviewing changes, 
Let us know once done with test case review so that can push all the changes in single commit

PubMatic-OpenWrap:
> Hello @pm-jaydeep-mohite, thanks for contributing! Just for future reference, if you don't mind, please merge with master instead of force pushing after your PR has been assigned for review. Force pushing makes it hard to figure out what has changed since the initial review which leads to a lot of extra time spent by our team on the delta reviews, especially for larger PRs. 🙂

Will merge changes from master in future

bsardo:
Hi @PubMatic-OpenWrap, we're familiar with rebasing and recommend using it when working on a PR in draft mode. However, once the review process begins, we strongly prefer merging with master because it keeps the PR history in tact and makes it much easier for reviewers to identify what has changed since their last review. This is especially important with larger PRs. As the PR get closer to its final state, the last thing we want to do is have to re-review 3k lines of code because of a force push when we may only be expecting a few small changes based on recent comments that can easily be observed with a merge with master. I hope this helps.

exchange/floors.go
PubMatic-OpenWrap: ok, noted

floors/floors.go
PubMatic-OpenWrap: updated to use CamelCase

floors/validate.go
PubMatic-OpenWrap: Done

floors/validate.go
PubMatic-OpenWrap: yes

exchange/floors.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/golang/glog"
+	"github.com/mxmCherry/openrtb/v16/openrtb2"
+
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/currency"
+	"github.com/prebid/prebid-server/floors"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// Check for Floors enforcement for deals,
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Updated

exchange/floors.go
```diff
@@ -0,0 +1,145 @@
...
+	errs := []error{}
+	impMap := make(map[string]openrtb2.Imp, len(bidRequest.Imp))
+
+	//Maintaining BidRequest Impression Map
+	for i := range bidRequest.Imp {
+		impMap[bidRequest.Imp[i].ID] = bidRequest.Imp[i]
+	}
+
+	for bidderName, seatBid := range seatBids {
+		eligibleBids := make([]*pbsOrtbBid, 0)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: updated

openrtb_ext/floors.go
PubMatic-OpenWrap: reorganised structures

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+func getDeviceType(request *openrtb2.BidRequest) string {
+	value := CATCH_ALL
+	if isMobileDevice(request.Device.UA) {
+		value = Phone
+	} else if isTabletDevice(request.Device.UA) {
+		value = Tablet
+	}
+	return value
```
bsardo, line 156: According to the requirements, `"*"` should only be returned if `request.Device.UA` does not exist so perhaps there should be an early return statement here:
```
if request.Device == nil || len(request.Device.UA) == 0 {
    return CATCH_ALL
}
```
That nil check is probably needed anyway otherwise `isMobileDevice` and `isTabletDevice` could potentially try to access a field on a `nil` value.

bsardo, line 156: The requirements also indicate that if `request.Device.UA` is present but it is not mobile or tablet then it should return `"desktop"`:
```
if device.ua is present, resolve deviceType to:

'phone' if UA matches one of these patterns: "Phone", "iPhone", "Android.*Mobile", "Mobile.*Android"
else 'tablet' if UA matches one of these: "tablet", "iPad", "Windows NT.*touch", "touch.*Windows NT", "Android"
else 'desktop'
```

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+
+	storedrequestID, err := jsonparser.GetString(imp.Ext, "prebid", "storedrequest", "id")
+	if err == nil && storedrequestID != "" {
+		return storedrequestID
+	}
+	return adUnitCode
+}
+
+func isMobileDevice(userAgent string) bool {
+	isMobile, err := regexp.MatchString("(?i)Phone|iPhone|Android|Mobile", userAgent)
```
bsardo, line 316: This regexp doesn't appear to match the requirements, particularly the android patterns:
`'phone' if UA matches one of these patterns: "Phone", "iPhone", "Android.*Mobile", "Mobile.*Android"`

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+	if err != nil {
+		return false
+	}
+	return isMobile
+}
+
+func isTabletDevice(userAgent string) bool {
+	isTablet, err := regexp.MatchString("(?i)tablet|iPad|Windows NT", userAgent)
```
bsardo, line 324: This regexp doesn't appear to match the requirements, particularly the Windows NT and Android patterns:
`else 'tablet' if UA matches one of these: "tablet", "iPad", "Windows NT.*touch", "touch.*Windows NT", "Android"`

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+func getgptslot(imp openrtb2.Imp) string {
+	var value string
+	adsname, err := jsonparser.GetString(imp.Ext, "data", "adserver", "name")
+	if err == nil && adsname == "gam" {
+		gptSlot, _ := jsonparser.GetString(imp.Ext, "data", "adserver", "adslot")
+		if gptSlot != "" {
+			value = gptSlot
+		}
```
bsardo, line 257: What are we supposed to return as the implied else condition here if `gptSlot` is an empty string? Currently we are returning an empty string but is that correct? I see that `getpbsadslot(imp)` returns `"*"` so should we be doing the same here?
I see that most of these functions will default to returning `"*"` which it appears you're doing so these values would at least match the wildcard in the generated rule combinations.

bsardo (commented):
Hi @PubMatic-OpenWrap,
I left a bunch of comments mostly around meeting the requirements. I'd also like to direct your attention to one comment in particular that involves the design. Please see my comment about O(2^N) complexity. I believe this should probably be resolved before making some of the other changes to the signaling logic.

Also it seems like you might have made some changes locally and forgot to push them to the PR based on your answers to some of @SyntaxNode's comments. You'll need to merge with master as well to resolve some recently introduced conflicts.

I still need to give the enforcement piece a closer look. I will do that shortly but I wanted to share this other feedback with you in the interim.

Thanks!

floors/floors.go
PubMatic-OpenWrap: Added condition to check if at least one model group is present

exchange/floors.go
```diff
@@ -0,0 +1,145 @@
...
+func selectFloorsAndModifyImp(r *AuctionRequest, floor config.PriceFloors, conversions currency.Conversions, responseDebugAllow bool) []error {
+
+	var errs []error
+	requestExt, err := r.BidRequestWrapper.GetRequestExt()
+	if err != nil {
+		errs = append(errs, err)
+		return errs
+	}
+	prebidExt := requestExt.GetPrebid()
+	if floor.Enabled && prebidExt.Floors != nil && prebidExt.Floors.GetEnabled() {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Changes are done, will be pushed in next commit

openrtb_ext/floors.go
```diff
@@ -0,0 +1,55 @@
+package openrtb_ext
+
+// PriceFloorRules defines the contract for bidrequest.ext.prebid.floors
+type PriceFloorRules struct {
+	FloorMin    float64                `json:"floormin,omitempty"`
+	FloorMinCur string                 `json:"floormincur,omitempty"`
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: according to v2 of schema float data type is mentioned, 
should we use float32 or float64 ?

openrtb_ext/floors.go
```diff
@@ -0,0 +1,55 @@
...
+	SkipRate            int                    `json:"skiprate,omitempty"`
+	FloorsSchemaVersion string                 `json:"floorsschemaversion,omitempty"`
+	ModelTimestamp      int                    `json:"modeltimestamp,omitempty"`
+	ModelGroups         []PriceFloorModelGroup `json:"modelgroups,omitempty"`
+}
+
+type PriceFloorModelGroup struct {
+	Currency     string             `json:"currency,omitempty"`
+	ModelWeight  int                `json:"modelweight,omitempty"`
+	DebugWeight  int                `json:"debugweight,omitempty"` // Added for Debug purpose, shall be removed
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: yes

openrtb_ext/floors.go
```diff
@@ -0,0 +1,55 @@
...
+type PriceFloorEnforcement struct {
+	EnforcePBS    *bool `json:"enforcepbs,omitempty"`
+	FloorDeals    *bool `json:"floordeals,omitempty"`
+	BidAdjustment bool  `json:"bidadjustment,omitempty"`
+	EnforceRate   int   `json:"enforcerate,omitempty"`
+}
+
+// GetEnabled will check if floors is enabled in request
+func (Floors *PriceFloorRules) GetEnabled() bool {
+	if Floors != nil && Floors.Enabled != nil && !*Floors.Enabled {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: already simplified, changes will be pushed in next commit

exchange/floors.go
```diff
@@ -0,0 +1,145 @@
...
+		requestExt.SetPrebid(prebidExt)
+		err := r.BidRequestWrapper.RebuildRequest()
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		if responseDebugAllow {
+			updatedBidReq, _ := json.Marshal(r.BidRequestWrapper.BidRequest)
+			//save updated request after floors signalling
+			r.UpdatedBidRequest = updatedBidReq
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: we would be modifying  resolved request

exchange/exchange.go
```diff
@@ -247,6 +250,13 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 	bidAdjustmentFactors := getExtBidAdjustmentFactors(requestExt)
 
+	// Get currency rates conversions for the auction
+	conversions := e.getAuctionCurrencyRates(requestExt.Prebid.CurrencyConversions)
+
+	// If floors feature is enabled at server and request level, Update floors values in impression object
+	floorErrs := selectFloorsAndModifyImp(&r, e.floor, conversions, responseDebugAllow)
+	errs = append(errs, floorErrs...)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: we can append floors related errors after `cleanOpenRTBRequests`

floors/floors.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// ModifyImpsWithFloors will validate floor rules, based on request and rules prepares various combinations
+// to match with floor rules and selects appripariate floor rule and update imp.bidfloor and imp.bidfloorcur
+func ModifyImpsWithFloors(floorExt *openrtb_ext.PriceFloorRules, request *openrtb2.BidRequest, conversions currency.Conversions) []error {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: updated

floors/floors.go
```diff
@@ -0,0 +1,89 @@
...
+	var (
+		floorErrList      []error
+		floorModelErrList []error
+		floorVal          float64
+	)
+	floorData := floorExt.Data
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: added  nil check

floors/floors.go
```diff
@@ -0,0 +1,89 @@
...
+
+	modelGroup := floorData.ModelGroups[0]
+	if modelGroup.Schema.Delimiter == "" {
+		modelGroup.Schema.Delimiter = DEFAULT_DELIMITER
+	}
+
+	floorExt.Skipped = new(bool)
+	if shouldSkipFloors(floorExt.Data.ModelGroups[0].SkipRate, floorExt.Data.SkipRate, floorExt.SkipRate, rand.Intn) {
+		*floorExt.Skipped = true
+		floorData.ModelGroups = nil
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: As per section 4.3
Non-selected models are removed when skipRate is false, in case of skipRate is true we would not be either signalling or enforcing floors, hence removed all modelgroups. 
If its required for extra detailed information, we can keep as it in case of skipRate is true

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+		floorCur = floorExt.Data.Currency
+	}
+
+	if floorExt.Data.ModelGroups[0].Currency != "" {
+		floorCur = floorExt.Data.ModelGroups[0].Currency
+	}
+	return floorCur
+}
+
+func getMinFloorValue(floorExt *openrtb_ext.PriceFloorRules, conversions currency.Conversions) (float64, string, error) {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: yes, this requirement is added recently, 
Changes shall be done in upcoming commits

floors/validate.go
```diff
@@ -0,0 +1,58 @@
...
+		ruleValues[newKey] = val
+	}
+	return errs
+}
+
+func validateFloorSkipRates(floorExt *openrtb_ext.PriceFloorRules) []error {
+	var errs []error
+
+	if floorExt.Data != nil && (floorExt.Data.SkipRate < SKIP_RATE_MIN || floorExt.Data.SkipRate > SKIP_RATE_MAX) {
+		errs = append(errs, fmt.Errorf("Invalid SkipRate at data level = '%v'", floorExt.Data.SkipRate))
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Done

floors/validate.go
```diff
@@ -0,0 +1,58 @@
...
+		return errs
+	}
+
+	if floorExt.SkipRate < SKIP_RATE_MIN || floorExt.SkipRate > SKIP_RATE_MAX {
+		errs = append(errs, fmt.Errorf("Invalid SkipRate at root level = '%v'", floorExt.SkipRate))
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Done

floors/validate.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return errs
+}
+
+func selectValidFloorModelGroups(modelGroups []openrtb_ext.PriceFloorModelGroup) ([]openrtb_ext.PriceFloorModelGroup, []error) {
+	var errs []error
+	var validModelGroups []openrtb_ext.PriceFloorModelGroup
+	for _, modelGroup := range modelGroups {
+		if modelGroup.SkipRate < SKIP_RATE_MIN || modelGroup.SkipRate > SKIP_RATE_MAX {
+			errs = append(errs, fmt.Errorf("Invalid Floor Model = '%v' due to SkipRate = '%v'", modelGroup.ModelVersion, modelGroup.SkipRate))
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Done

floors/validate.go
```diff
@@ -0,0 +1,58 @@
...
+			continue
+		}
+
+		if modelGroup.ModelWeight < MODEL_WEIGHT_MIN_VALUE || modelGroup.ModelWeight > MODEL_WEIGHT_MAX_VALUE {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Done

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	return value
+}
+
+func getMediaType(imp openrtb2.Imp) string {
+	value := CATCH_ALL
+	if imp.Banner != nil {
+		value = string(openrtb_ext.BidTypeBanner)
+	} else if imp.Video != nil {
+		value = string(openrtb_ext.BidTypeVideo)
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Will be added in next commit

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+	return value
+}
+
+func getSizeValue(imp openrtb2.Imp) string {
+	size := CATCH_ALL
+	width := int64(0)
+	height := int64(0)
+	if imp.Banner != nil {
+		if len(imp.Banner.Format) > 0 {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: yes, this logic only applies when there aren't multiple supported formats.

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+			width = imp.Banner.Format[0].W
+			height = imp.Banner.Format[0].H
+		} else if imp.Banner.W != nil && imp.Banner.H != nil {
+			width = *imp.Banner.W
+			height = *imp.Banner.H
+		}
+	} else {
+		width = imp.Video.W
+		height = imp.Video.H
+	}
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: corrected

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+func getMinFloorValue(floorExt *openrtb_ext.PriceFloorRules, conversions currency.Conversions) (float64, string, error) {
+	var err error
+	var rate float64
+	floorMin := floorExt.FloorMin
+	floorCur := getFloorCurrency(floorExt)
+
+	if floorExt.FloorMin > 0.0 && floorExt.FloorMinCur != "" && floorCur != "" &&
+		floorExt.FloorMinCur != floorCur {
+		rate, err = conversions.GetRate(floorExt.FloorMinCur, floorCur)
+		floorMin = rate * floorExt.FloorMin
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Done

floors/rule.go
PubMatic-OpenWrap: added

floors/rule.go
PubMatic-OpenWrap: If `gptSlot` is an empty string then we are returning` "*"`

floors/rule.go
PubMatic-OpenWrap: added desktop

openrtb_ext/floors.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+type PriceFloorEndpoint struct {
+	URL string `json:"url,omitempty"`
+}
+
+type PriceFloorData struct {
+	Currency            string                 `json:"currency,omitempty"`
+	SkipRate            int                    `json:"skiprate,omitempty"`
+	FloorsSchemaVersion string                 `json:"floorsschemaversion,omitempty"`
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: added validation for `FloorsSchemaVersion`

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+func getDeviceCountry(request *openrtb2.BidRequest) string {
+	value := CATCH_ALL
+	if request.Device != nil && request.Device.Geo != nil {
+		value = request.Device.Geo.Country
+	}
+	return value
+}
+
+func getMediaType(imp openrtb2.Imp) string {
+	value := CATCH_ALL
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: added case to handle multi-format impression

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	if width != 0 && height != 0 {
+		size = fmt.Sprintf("%dx%d", width, height)
+	}
+	return size
+}
+
+func getDomain(request *openrtb2.BidRequest) string {
+	value := CATCH_ALL
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: removed

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+	var desiredkeys [][]string
+	var ruleKeys []string
+
+	segNum := 1 << numSchemaFields
+	for i := 0; i < numSchemaFields; i++ {
+		subset = append(subset, strings.ToLower(keys[i]))
+		comb = append(comb, i)
+	}
+	desiredkeys = append(desiredkeys, subset)
+	for numWildCart := 1; numWildCart <= numSchemaFields; numWildCart++ {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: corrected

floors/rule.go
```diff
@@ -0,0 +1,404 @@
...
+		newComb := generateCombinations(comb, numWildCart, segNum)
+		for i := 0; i < len(newComb); i++ {
```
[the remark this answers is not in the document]
PubMatic-OpenWrap: Time complexity is O(2^N) complexity where N is the number of schema fields, in actual use case N could be less than 12.
It would be great if we get ideas on  more efficient algorithm.
We can prepare combination set for each N  in first request and put it in cache and in 2nd request onwards we can use cached combination, however rule keys combination should be done to each request.

Issue #1812, opened 2021-04-15
Allow gzip compress for BidRequest backend communication with bidders


josepowera:
Some exchanges/DSPs have implemented support for compressed json BidRequest. This change could be system level, since it would be useful for many bidders in order to reduce bandwidth / latency from Prebid Server to DSP bidServer.  This option would have to be enabled per bidder.

Same goes for BidResponse from DSP back to Prebid Server.  In this case Prebid Server can decide based on  Content-Encoding: gzip header.

Is there any settins already available to enable this?  (this is similar to gzip handler to compress BidResponse from PrebidServer back to browser)

bretg:
This is a good call @josepowera . We'll discuss in this week's PBS committee meeting.

bretg:
We discussed today. The team agrees this is a worthwhile effort. We'll tackle it in three parts:

1) support compression on requests coming into PBS, with a length limit (any input from the community on the max size welcome)
2) support a bidder-config for outgoing requests. We may initially only support gzip.
3) verify that PBS bid adapter can send compressed requests

bretg:
Proposing that we make the length limit configurable with a default of 50KB uncompressed. That seems like more than enough.

SyntaxNode:
> Proposing that we make the length limit configurable with a default of 50KB uncompressed. That seems like more than enough.

PBS-Go (and likely PBS-Java) already include a configurable maximum request size with a default of 256KB. I'd like us to use the same configuration value for uncompressing compressed payloads, unless someone has a good reason for having separate values? We could always lower the default value is 256KB is too high.

bretg:
Discussed in committee:
- start with gzip. Ignore deflate. We would like to support Brotli.
- the default length limit is still open, but we're leaning towards 256KB

albertmolinermrf:
Marginally related: PBS-Java allows to [request compression from bidders](https://github.com/prebid/prebid-server-java/issues/288).

rpanchyk:
@bretg how many bidders are planned to support compressed `BidRequest`s ?
Just want to clarify, if not so many of them - would it it be better to add only bidder-specific config? Something like we have for Conversant - `adapters.conversant.generate-bid-id`.

bretg:
This fell through the cracks. Yes, I think there should be a bid adapter-level config that lets PBS know that the endpoint accepts certain encoding. e.g. 
```
adapters.CODE.endpoint-compression: gzip
```
[...]

PR #2347, opened 2022-08-09
Add GZIP Compression

AlexBVolcy:
This PR addresses this issue here: https://github.com/prebid/prebid-server/issues/1812

The goal of the PR is to allow bidders to GZIP compress the request data this being utilized to create http requests that are being generated in by `doRequestImpl()`

To accomplish this, I've added a new string variable `EndpointCompression`, which can be set to tell PBS what type of compression a bidder may want (i.e. `EndpointCompression = "GZIP"`). 

This `EndpointCompression` object has been added as a member of the `BidderInfo` struct so that this compression information can be passed to `AdaptBidder()`. 

In `AdaptBidder()`, the `EndpointCompression` is then added to a `bidderAdapterConfig` object, so that we can access the compression info in `requestBid()`.

Once at requestBid(), `doRequest()` and subsequently `doRequestImpl()` is called with the `bidderAdapterConfig` info present, so it's here we check if `EndpointCompression == "GZIP"`, and if so, we compress the request data before calling `http.NewRequest()`.

A single unit test was written for this update, and may require manual end-to-end testing given the nature of the feature requiring compressed data to be created and verified.


exchange/bidder.go
```diff
@@ -501,7 +507,20 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
...
-	httpReq, err := http.NewRequest(req.Method, req.Uri, bytes.NewBuffer(req.Body))
+	var requestBody []byte
+
+	switch bidder.config.EndpointCompression {
+	case "GZIP":
+		var b bytes.Buffer
+		w := gzip.NewWriter(&b)
+		w.Write([]byte(req.Body))
+		w.Close()
+		requestBody = b.Bytes()
```
VeronikaSolovei9, line 518: Lines 514-518 look like a good candidates to be extracted to a separate function. 
`requestBody = compressToGZIP(req.Body)` (or better name for it :))

config/bidderinfo.go
```diff
@@ -22,6 +22,11 @@ type BidderInfo struct {
 	GVLVendorID             uint16               `yaml:"gvlVendorID"`
 	Syncer                  *Syncer              `yaml:"userSync"`
 	Experiment              BidderInfoExperiment `yaml:"experiment"`
+	CompressionType         *CompressionType     `yaml:"compressionType"`
+}
+
+type CompressionType struct {
+	EndpointCompression string `yaml:"endpointCompression"`
```
shahbaztariq, line 29: Would it be better to be more specific here so that it is understood that this compression is for outgoing requests only?

There is already confusion around how Go handles GZIP in it's default httpclient.

My thoughts are that if we're more specific here the config will be better understood and easier to reason with.

Something like `RequestCompressionType`? Or something along those lines?

VeronikaSolovei9 (changes requested):

exchange/bidder.go
```diff
@@ -501,7 +507,20 @@ func (bidder *bidderAdapter) doRequest(ctx context.Context, req *adapters.Reques
 }
 
 func (bidder *bidderAdapter) doRequestImpl(ctx context.Context, req *adapters.RequestData, logger util.LogMsg) *httpCallInfo {
-	httpReq, err := http.NewRequest(req.Method, req.Uri, bytes.NewBuffer(req.Body))
+	var requestBody []byte
+
+	switch bidder.config.EndpointCompression {
+	case "GZIP":
+		var b bytes.Buffer
```
VeronikaSolovei9, line 514: In config "GZIP" can be also specified as "gzip" or even "gZip". This case will never work.
Please replace switch statement to switch strings.ToUpper(bidder.config.EndpointCompression) {.
Also I recommend to extract "GZIP" to constant.

VeronikaSolovei9:
For some reason it shows weird difference in bidder_test.go file. 
For line 275 where you compare debug http calls, can you add a verification if debug http calls are present, check the `Content-Encoding` header is present as well and it has `gzip` value in it? I understand it's hard to test, with this check you may ensure that Switch case worked. 
Also please add a check that request body in debug http calls is not gziped. Basically just assert request body equals to `{"key":"val"}`

Line 189 `Body:    []byte("{\"key\":\"val\"}"),` modify to `Body:    []byte(`{"key":"val"}`),`

VeronikaSolovei9:
Code looks good to me, added some minor comments

config/bidderinfo.go
AlexBVolcy: I understand your reasoning, but we are trying to be consistent with variable/object naming between PBS-Go and PBS-Java, so I think to accomplish this this object can stay being named this way. I'll see if other team members think differently though!

VeronikaSolovei9 (approved):
Thank you for addressing the comments, LGTM

config/bidderinfo.go
guscarreon, line 29: Can we add a small description for `CompressionType`? Please feel free to reword as you see fit.
```
27 + // CompressionType if set, determines the type of compression the bid request will undergo before being sent to the corresponding bid server
28   type CompressionType struct {
29       EndpointCompression string `yaml:"endpointCompression"`
30   }
31  
```

guscarreon (commented):
Can we add an additional `TestReadFullConfig()` test case inside `config/bidderinfo_test.go`? Something similar to what we already have in `config/config_test.go` but with a hardcoded sample YAML file instead of a hardcoded JSON, and it comes with all the fields supported by the `BidderInfo` struct:
```
11  
12   const testInfoFilesPath = "./test/bidder-info"
13   const testSimpleYAML = `
14   maintainer:
15     email: "some-email@domain.com"
16   gvlVendorID: 42
17   `
   + const fullBidderYAMLConfig = `
   + maintainer:
   +   email: "some-email@domain.com"
   + capabilities:
   +   app:
   +     mediaTypes: 
   +       - banner
   +   site:
   +     mediaTypes:
   +       - video
   +       - native
   + modifyingVastXmlAllowed: true
   + debug:
   +   allow: true
   + gvlVendorID: 42
   + userSync:
   +   iframe:
   +     url: "someURL"
   +     userMacro: "aValue"
   +   redirect:
   +     url: anotherURL
   +     userMacro: anotherValue
   +   key: "akey"
   +   supports:
   +     - item
   +     - item2
   +   external_url: oneMoreUrl
   +   support_cors: true
   + experiment:
   +   adsCert:
   +     enabled: true
   + compressionType: GZIP
   + `
   +
   + func TestLoadBidderInfoFromDisk(t *testing.T) {
   +     // Read the fullBidderYAMLConfig
   +     expected := BidderInfos{
   +     	bidder: {
   +     		Enabled: true,
   +     		Maintainer: &MaintainerInfo{
             . 
             . 
             . 
   +            CompressionType: "GZIP",
   +        },
   +     }
   +     // Read the fullBidderYAMLConfig
   +     // assert
   + }
18  
config/bidderinfo_test.go
```

guscarreon (approved):
LGTM

guscarreon (approved):
LGTM

VeronikaSolovei9 (dismissed):

AlexBVolcy:
Please don't merge this PR, until I'm done with my end to end manual testing!

AlexBVolcy:
Manual end to end testing is completed! This is ready for a final review/merge -- I just changed the naming of one yaml variable

guscarreon (approved):
LGTM

VeronikaSolovei9 (approved):

PR #2351, opened 2022-08-12
AdsCert signer metrics

VeronikaSolovei9:
New metrics are added to track AdsCert signing:
`RecordAdsCertReq` - record requests with AdsCert header, tracks number of successful and failed requests.
`RecordAdsCertSignTime` - time to sign AdsCert request

Local test with Prometheus passed, all metrics are correctly displayed in Prometheus data

### DIFF
--- a/exchange/bidder.go
+++ b/exchange/bidder.go
@@ -176,13 +176,16 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, bidderRequest Bidde
 				reqData[i].Headers.Add("Sec-GPC", reqInfo.GlobalPrivacyControlHeader)
 			}
 			if bidRequestOptions.addCallSignHeader {
+				startSignRequestTime := time.Now()
 				signatureMessage, err := adsCertSigner.Sign(reqData[i].Uri, reqData[i].Body)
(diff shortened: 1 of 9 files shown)


exchange/bidder.go
```diff
@@ -176,13 +176,16 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, bidderRequest Bidde
 				reqData[i].Headers.Add("Sec-GPC", reqInfo.GlobalPrivacyControlHeader)
 			}
 			if bidRequestOptions.addCallSignHeader {
+				startSignRequestTime := time.Now()
 				signatureMessage, err := adsCertSigner.Sign(reqData[i].Uri, reqData[i].Body)
+				bidder.me.RecordAdsCertSignTime(time.Since(startSignRequestTime))
```
hhhjort, line 181: Do we expect that the time taken to generate a signature to be significant to warrant tracking as a metric?

VeronikaSolovei9: When we enable this feature it can help to show if this feature can be a potential bottleneck in request handling. It can be especially valuable when PBS uses remote signatory service.  
@SyntaxNode please confirm

hhhjort (approved):

mansinahar (approved):

PR #2353, opened 2022-08-18
Adapter aliases: moved adapter related configs to bidder.yaml files

VeronikaSolovei9:
- all adapter related configs are moved to static/bidder-info/{bidder}.yaml files.
- Apapter struct now carries only adapter needed data
- all adapter related configs are removed from config.go
- bidder list now loads from the file system

Docs update PR: https://github.com/prebid/prebid.github.io/pull/4008


config/bidderinfo.go
```diff
@@ -176,94 +164,366 @@ type SyncerEndpoint struct {
...
+		}
+	}
+
+	return nil
 }
 
-type infoReaderFromDisk struct {
-	path string
+func applyBidderInfoConfigOverrides(configBidderInfos BidderInfos, fsBidderInfos BidderInfos) (BidderInfos, error) {
+	for bidderName, bidderInfo := range configBidderInfos {
```
VeronikaSolovei9, line 365: Will add unit tests for this function when logic is confirmed

AlexBVolcy:
Also just wanted to comment highlighting the merge conflicts with bidderinfo., bidderinfo_test.go, adapter_util_test.go

config/bidderinfo.go
```diff
@@ -176,94 +164,366 @@ type SyncerEndpoint struct {
...
+	return fsBidderInfos, nil
 }
 
-func (r infoReaderFromDisk) Read(bidder string) ([]byte, error) {
-	path := fmt.Sprintf("%v/%v.yaml", r.path, bidder)
-	return ioutil.ReadFile(path)
+// Override returns a new Syncer object where values in the original are replaced by non-empty/non-default
+// values in the override, except for the Supports field which may not be overridden. No changes are made
+// to the original or override Syncer.
+func (s *Syncer) Override(original *Syncer) *Syncer {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Right, this is a part of merging function, will add it in next push

config/bidderinfo.go
```diff
@@ -176,94 +164,366 @@ type SyncerEndpoint struct {
...
-// bidders are omitted from the result.
-func (infos BidderInfos) ToGVLVendorIDMap() map[openrtb_ext.BidderName]uint16 {
-	m := make(map[openrtb_ext.BidderName]uint16, len(infos))
-	for name, info := range infos {
-		if info.Enabled && info.GVLVendorID != 0 {
-			m[openrtb_ext.BidderName(name)] = info.GVLVendorID
-		}
+// Override returns a new SyncerEndpoint object where values in the original are replaced by non-empty/non-default
+// values in the override. No changes are made to the original or override SyncerEndpoint.
+func (s *SyncerEndpoint) Override(original *SyncerEndpoint) *SyncerEndpoint {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Will do!

config/bidderinfo.go
```diff
@@ -176,94 +164,366 @@ type SyncerEndpoint struct {
...
 
-	if s.URL != "" {
-		copy.URL = s.URL
+	return infos, nil
+}
+
+// ToGVLVendorIDMap transforms a BidderInfos object to a map of bidder names to GVL id. Disabled
+// bidders are omitted from the result.
+func (infos BidderInfos) ToGVLVendorIDMap() map[openrtb_ext.BidderName]uint16 {
+	m := make(map[openrtb_ext.BidderName]uint16, len(infos))
```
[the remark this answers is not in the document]
VeronikaSolovei9: This is moved from another place, I didn't change content. Renamed for future convenience.

config/bidderinfo.go
```diff
@@ -177,94 +166,367 @@ type SyncerEndpoint struct {
...
+	errs = validateBidderInfos(bidderInfos)
+
+	return bidderInfos, errs
+}
+
+// LoadBidderInfoFromDisk parses all static/bidder-info/{bidder}.yaml files from the file system.
+func LoadBidderInfoFromDisk(path string) (BidderInfos, error) {
+	return loadBidderInfo(path)
+}
+
```
hhhjort, line 191: Why do we have `LoadBidderInfoFromDisk(path)` when `loadBidderInfo(path)` is also necessarily from disk? I could see the utility if `loadBidderInfo()` was more general in some sense, but I don't see what we gain from it in this instance.

config/config.go
```diff
@@ -908,186 +908,6 @@ func SetupViper(v *viper.Viper, filename string) {
 	// macro substitution. it is important for the uid to be the last query parameter.
 	v.SetDefault("user_sync.redirect_url", "{{.ExternalURL}}/setuid?bidder={{.SyncerKey}}&gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&f={{.SyncType}}&uid={{.UserMacro}}")
 
-	for _, bidder := range openrtb_ext.CoreBidderNames() {
```
hhhjort, line 911: without any defaults, how will viper know to look for adapter config overrides in the environment variables. Viper does not use reflect to interrogate the config struct for all possible fields that might exist. I do think viper has another function other than `SetDefault()` that will clue it in to a field's existence without actually setting a default value.

config/config_test.go
```diff
@@ -1379,53 +1261,6 @@ func TestValidateAccountsConfigRestrictions(t *testing.T) {
 	assert.Contains(t, errs, errors.New("accounts.postgres: retrieving accounts via postgres not available, use accounts.files"))
 }
 
-func TestUserSyncFromEnv(t *testing.T) {
```
hhhjort, line 1382: Why is this test being removed? Is it intended that it should no longer be possible to set usersync parameters from environment variables?

VeronikaSolovei9: There will be no adapter related config in environment variables. All bidder related data is loading from {bidder}.yaml files or pbs.json or pbs.yaml files. 
`cfg.Adapters` were removed from Configuration struct. 

`assert.Nil(t, cfg.Adapters["appnexus"].Syncer.IFrame)` - is not a valid structure anymore.

config/config.go
VeronikaSolovei9: This also relates to Configuration -> map[string]Adapter which was removed. There is no way to set adapterDefaults to adapters. Unless i understand something wrong. We need to discuss it with the team.

config/bidderinfo.go
VeronikaSolovei9: Per my understanding it was done specifically for unit tests to mock data reading from file system. Implementations for "infoReader" are `infoReaderFromDisk` and `fakeInfoReader`. 

Do we want to support other types of data loading besides file system?

config/config_test.go
hhhjort: It is common for kubernettes deployments to use environment variables for configuration rather than config files, particularly when you have deployments in multiple datacenters that have datacenter specific parameters (such as the bidder endpoint). Is this breaking change already documented in an issue? Otherwise we should certainly document it as such and give hosts ample opportunity to rework their setups to continue to work properly when this feature is removed.

config/config.go
hhhjort: See my comments below on environment variable config support being removed.

config/bidderinfo.go
hhhjort: My confusion rests more in that if `LoadBidderInfoFromDisk()` is an `infoReader` interface, then `loadBidderInfo()` is as well, as they have the same signature. Normally when I see a pattern like this, it is of the form:
```f1(A) X {
  return f2(A,B)
  }```
  Where B is some constant, and `f2` is some more general function than `f1` is supposed to be. Such as if `loadBidderInfo()` is intended to support other sources than disk some day, but where `LoadBidderInforFromDisk()` is called, will only ever be loading from disk. Using the pattern you have used here would allow for expanding `loadBidderInfo()` in the future without having to update the parts of code that currently call it. Since I am not aware of any plans to support other data sources, and `loadBidderInfo` is also an implementation of `infoReader`, I still do not understand why both functions exist like this.

File: config/bidderinfo.go [hunk 8]
```diff
@@ -177,94 +166,344 @@ type SyncerEndpoint struct {
...
+			return nil, err
+		}
+
+		info := BidderInfo{}
+		if err := yaml.Unmarshal(data, &info); err != nil {
+			return nil, fmt.Errorf("error parsing yaml for bidder %s: %v", fileName, err)
+		}
+
+		bidderName := strings.Split(fileName, ".yaml")
+		infos[(bidderName[0])] = info
```
AlexBVolcy, line 195: Is it guaranteed that there will be data present at index 0 here? Just want to check

config/bidderinfo.go
```diff
@@ -177,94 +166,344 @@ type SyncerEndpoint struct {
...
-type infoReader interface {
-	Read(bidder string) ([]byte, error)
+func validateSyncer(bidderInfo BidderInfo) error {
+	if bidderInfo.Syncer == nil {
+		return nil
+	}
+
+	for _, v := range bidderInfo.Syncer.Supports {
+		if !strings.EqualFold(v, "iframe") && !strings.EqualFold(v, "redirect") {
+			return fmt.Errorf("syncer could not be created, invalid supported endpoint: %s", v)
```
AlexBVolcy, line 336: Could this variable `v` be more descriptive? I understand if you want to leave it the way it is, but I tend to like a little more from variable names

AlexBVolcy (commented):
Very minor comments

config/bidderinfo.go
VeronikaSolovei9: good point. Added check if file is a yaml file and array len.

config/bidderinfo.go
VeronikaSolovei9: Renamed

config/config.go
```diff
@@ -701,6 +702,12 @@ func New(v *viper.Viper) (*Configuration, error) {
 	// Migrate combo stored request config to separate stored_reqs and amp stored_reqs configs.
 	resolvedStoredRequestsConfig(&c)
 
+	mergedBidderInfos, err := applyBidderInfoConfigOverrides(c.BidderInfos, bidderInfos)
```
hhhjort, line 705: Would it make sense to insert the `bidderInfos` loaded from disk into the `config` struct before the viper load step, so viper does the merge automatically? And then perhaps a validate infos step to handle the extra logic in the Overrides function.

VeronikaSolovei9: I tried it and found interesting behavior. In `setBidderDefaults` bidder infos in viper initialized with empty values. If any field for adapter was set through env variables or pbs.json or pbs.yaml file - Viper sets this value to bidder info.
If I set bidder info from file system to config before unmarshalling - Viper replaces everything for bidders that were "extra" configured through env variables. These bidders only have data from env variables and nothing from file system. Bidders that were not configured through env variables have data from file system. 

Hope I understand you correctly, basically you suggest to do something like this: 
```
// New uses viper to get our server configurations.
func New(v *viper.Viper, bidderInfos BidderInfos) (*Configuration, error) {
	var c Configuration
	c.BidderInfos = bidderInfos ---> set loaded bidder info from filesystem here
	if err := v.Unmarshal(&c); err != nil {
		return nil, fmt.Errorf("viper failed to unmarshal app config: %v", err)
	}
      .....
```

VeronikaSolovei9: Could you please add more details about `validate infos step to handle the extra logic in the Overrides function`? Move bidder infos validation to `applyBidderInfoConfigOverrides`?

hhhjort: The above is what I was thinking. So you discovered that if, for example, a bidder's `endpoint` was set in an environment variable, viper would override the initial setting with that from the environment variable. However if it was set in `pbs.yaml`, viper would ignore it and keep what was originally set before the unmarshal? If this is true, there is some sort of inconsistency with viper. I don't think it will be worthwhile debugging that inconsistency, my suggestion was mainly in hopes of keeping the code simpler and easier to read, there should be no real functional advantage in doing it this way.

AlexBVolcy:
This looks good! I want to confirm that exchange about bidderInfos and viper is resolved before approving!

config/config.go
VeronikaSolovei9: Discussed this offline. I didn't find better and easier way to achieve same result using Viper functionality.

config/config_test.go
hhhjort: Did you add back in tests for environment variables setting adapter parameters somewhere else?

VeronikaSolovei9: Good point! Added and fixed mapstructures in bidder info and added remaining bidder info properties to `setBidderDefaults`

hhhjort (dismissed):
LGTM

AlexBVolcy (approved):

config/bidderinfo.go
```diff
@@ -177,94 +166,345 @@ type SyncerEndpoint struct {
...
 		}
+	}
+	return nil
+}
 
-		info.Enabled = isEnabledByConfig(adapterConfigs, bidder)
-		infos[bidder] = info
+func validatePlatformInfo(info *PlatformInfo) error {
+	if len(info.MediaTypes) == 0 {
+		return errors.New("mediaTypes should be an array with at least one string element")
```
SyntaxNode, line 318: Nitpick: There is no need to include the wording "should be an array". Of course it's an array. Consider changing to an error along the line of "at least one meda type needs to be specified".

config/bidderinfo.go
```diff
@@ -177,94 +166,345 @@ type SyncerEndpoint struct {
...
+			return fmt.Errorf("syncer could not be created, invalid supported endpoint: %s", supports)
+		}
+	}
+
+	return nil
 }
 
-type infoReaderFromDisk struct {
-	path string
+func applyBidderInfoConfigOverrides(configBidderInfos BidderInfos, fsBidderInfos BidderInfos) (BidderInfos, error) {
```
SyntaxNode, line 344: This is no longer necessary. We can let Viper take care of this work. You need to only give Viper the bidder infos before calling unmarshal and it will take care of the overrides:

```
func New(v *viper.Viper, bidderInfos BidderInfos) (*Configuration, error) {
	var c Configuration
	c.BidderInfos = bidderInfos

	if err := v.Unmarshal(&c); err != nil {
		return nil, fmt.Errorf("viper failed to unmarshal app config: %v", err)
	}
	
	...
```

This breaks is the legacy user sync url, which we can either choose to remove as part of this work or we alert `applyBidderInfoConfigOverrides` to only apply the legacy user sync url override. 

This also removes our ability to disallow certain parts of the config to be changed, like the maintainer or media types. The later can cause a problem if it's less restrictive.

I think we can either:
1. Let Viper handle everything and remove the override code. This allows hosts to "shoot themselves in the foot", but there's nothing stop them from editing the bidder info files directly?
2. Keep control of overrides and create a new BidderInfoOverride with less fields and leave the override code.

Thoughts?

config/config.go
```diff
@@ -1292,16 +1124,23 @@ func migrateConfigTCF2PurposeEnabledFlags(v *viper.Viper) {
...
+	v.BindEnv(adapterCfgPrefix+".debug.allow", "")
+	v.BindEnv(adapterCfgPrefix+".gvlVendorID", "")
+	v.BindEnv(adapterCfgPrefix+".usersync_url", "")
+	v.BindEnv(adapterCfgPrefix+".experiment.adsCert.enabled", "")
+	v.BindEnv(adapterCfgPrefix+".platform_id", "")
+	v.BindEnv(adapterCfgPrefix+".app_secret", "")
+	v.BindEnv(adapterCfgPrefix+".xapi.username", "")
+	v.BindEnv(adapterCfgPrefix+".xapi.password", "")
+	v.BindEnv(adapterCfgPrefix+".xapi.tracker", "")
+	v.BindEnv(adapterCfgPrefix+".endpointCompression", "")
```
SyntaxNode, line 1143: There be in danger in allowing unrestricted overrides of `capabilities.app.mediaTypes` and `capabilities.site.mediaTypes`.

config/config_test.go
```diff
@@ -764,9 +682,203 @@ func TestMigrateConfigFromEnv(t *testing.T) {
...
+
+	assert.Equal(t, true, cfg.BidderInfos["bidder1"].Experiment.AdsCert.Enabled)
+	assert.Equal(t, "username_override", cfg.BidderInfos["bidder1"].XAPI.Username)
+	assert.Equal(t, "password_override", cfg.BidderInfos["bidder1"].XAPI.Password)
+	assert.Equal(t, "tracker_override", cfg.BidderInfos["bidder1"].XAPI.Tracker)
+
+	assert.Equal(t, "platform_id_override", cfg.BidderInfos["bidder1"].PlatformID)
+	assert.Equal(t, "app_secret_override", cfg.BidderInfos["bidder1"].AppSecret)
+	assert.Equal(t, "GZIP", cfg.BidderInfos["bidder1"].EndpointCompression)
 }
```
SyntaxNode, line 882: IMHO we don't need to test all of the env variables, just a few is fine to provide viper binding works ok. This seems a little too verbose IMHO.

config/test/bidder-info-invalid/someBidder-invalid.yaml
```diff
@@ -0,0 +1,3 @@
+maintainer:
+  email: 123
+gvlVendorID: "42"
```
SyntaxNode, line 3: Why do we need do add this? Unit tests should cover this already.

endpoints/events/vtrack.go
```diff
@@ -267,7 +267,7 @@ func isAllowVastForBidder(bidder string, bidderInfos *config.BidderInfos, allowU
 	// check if bidder is configured
 	if b, ok := (*bidderInfos)[bidder]; bidderInfos != nil && ok {
 		// check if bidder is enabled
-		return b.Enabled && b.ModifyingVastXmlAllowed
+		return !b.Disabled && b.ModifyingVastXmlAllowed
```
SyntaxNode, line 270: I recommend adding a helper method to bidder info so we can call .IsEnabled() instead of negating the disabled field. The later is harder to read and easier to miss.

config/bidderinfo.go
VeronikaSolovei9: Modified

config/config.go
VeronikaSolovei9: removed support of the capabilities override

config/config_test.go
VeronikaSolovei9: Removed some of checks

config/test/bidder-info-invalid/someBidder-invalid.yaml
VeronikaSolovei9: removed. Just wanted to test it with actual yaml file.

config/bidderinfo.go
VeronikaSolovei9: This is what you showed me on our very first discussion about the approach to support env variables. 
We also had a discussion about it with Hans: https://github.com/prebid/prebid-server/pull/2353#discussion_r963124663

If I set `c.BidderInfos = bidderInfos` and have env variable for at least one field for a bidder (for example endpoint) - it overrides the entire object for this bidder and sets "endpoint" only. Other data from yaml file is not present for this bidder after config umnarshalling. 

I think we need to control overrides. I removed `capabilities` from overrides, we can discuss what other fields should be removed. But I also think this is not obvious why some of the fields can be overridden and some not. Please advice.

config/bidderinfo.go
```diff
@@ -315,7 +319,7 @@ func validateCapabilities(info *CapabilitiesInfo, bidderName string) error {
 
 func validatePlatformInfo(info *PlatformInfo) error {
 	if len(info.MediaTypes) == 0 {
-		return errors.New("mediaTypes should be an array with at least one string element")
+		return errors.New("at least one meda type needs to be specified")
```
hhhjort, line 322: media has a typo 'meda'

AlexBVolcy (dismissed):

AlexBVolcy (approved):
LGTM

hhhjort (approved):

Issue #2139, opened 2022-01-24
Prebid Server OpenRTB 2.6 support


bretg:
OpenRTB 2.6 is open for comment at https://iabtechlab.com/wp-content/uploads/2022/04/OpenRTB-2-6_FINAL.pdf

A field-by-field list of changes below, but here are the high-level changes for Prebid Server:
1. Logically, stored requests need to be merged before doing the ORTB normalization. Or it needs to be done twice. We shouldn't assume that PBS-host companies have converted all of their stored requests at initial release.
2. Expand JSON validation. We propose keeping the removed fields so PBS validation is done on a superset of 2.5 and 2.6.
3. Transition support for moved objects: schain, eids, rewarded video. Clients will be placing this data on the original extensions for years to come. We could make this transition support modular or configurable so host companies can avoid the cost at some point.
    a. The committee has agreed to _move_ extended locations to the new locations. e.g. source.ext.schain to source.schain.
    b. If both new and old locations are seen in a request (e.g. post SRID merge), prefer the new location.
4. Updates to the [ortb2 blocking module](https://docs.prebid.org/prebid-server/pbs-modules/ortb2-blocking.html) to support cattax fields
5. Port long-form video functionality to the /ortb2/auction endpoint using the new pod-related fields.
6. If a bidder only supports 2.5:
    1. PBS-core should copy these objects from the 2.6 location to the 2.5 location: schain, eids, regs, user.consent, rewarded video.
    2. For fields that don't exist in 2.5 (e.g. imp.ssai), PBS-core may either remove them or leave them in place. Theoretically bid adapters should ignore them.
7. Bidder-specific first party data must support specifying all the new fields.

These changes don't all have to come at once.

## BidRequest changes in 2.6
- New field $.wlangb. pass it through to bid adapters
- New fields \\$.{site,app}.content.langb, \\$.device.langb. Pass through to bid adapters. ORTB 2.6 says "Only one of language or langb should be present." but PBS does not need to enforce this.
- New field $.cattax: clarifies .bcat. Will need to update the [ortb2 blocking module](https://docs.prebid.org/prebid-server/pbs-modules/ortb2-blocking.html) to support this field
- New fields $.site.cattax. Support as First Party Data fields and pass through to bidders.
- New field $.app.cattax. Support as First Party Data fields and pass through to bidders.
- New field $.{}.publisher.cattax. Support as First Party Data fields and pass through to bidders.
[...]

PR #2354, opened 2022-08-18
OpenRTB 2.6: Up + Down Conversions

SyntaxNode:
Implements another part of https://github.com/prebid/prebid-server/issues/2139. The converts are not yet called by the core logic of Prebid Server, but will in a follow-up PR in attempt to keep the reviews smaller / contained.


openrtb_ext/request_wrapper.go
```diff
@@ -458,26 +460,34 @@ func (re *RequestExt) unmarshal(extJson json.RawMessage) error {
...
-	if err != nil {
+
+	if err := json.Unmarshal(extJson, &re.ext); err != nil {
 		return err
 	}
+
 	prebidJson, hasPrebid := re.ext["prebid"]
 	if hasPrebid {
 		re.prebid = &ExtRequestPrebid{}
-		err = json.Unmarshal(prebidJson, re.prebid)
```
SyntaxNode, line 472: Found some issues with error handling introduced over time.

openrtb_ext/request_wrapper.go
```diff
@@ -761,23 +783,48 @@ func (re *RegExt) unmarshal(extJson json.RawMessage) error {
...
-	if err != nil {
+
+	if err := json.Unmarshal(extJson, &re.ext); err != nil {
 		return err
 	}
+
+	gdprJson, hasGDPR := re.ext["gdpr"]
+	if hasGDPR {
+		if err := json.Unmarshal(gdprJson, &re.gdpr); err != nil {
+			return errors.New("gdpr must be an integer")
```
SyntaxNode, line 800: Specialized error message to get a similar error message like we had before. This how runs prior to value validation, which wasn't previously the case.

openrtb_ext/convert_up_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	for _, test := range testCases {
+		w := &RequestWrapper{BidRequest: &test.givenRequest}
+		moveSupplyChainFrom24To25(w)
+		assert.NoError(t, w.RebuildRequest(), test.description)
+		assert.Equal(t, test.expectedRequest, *w.BidRequest, test.description)
+	}
+}
+
+func TestConvertSupplyChainFrom25To26(t *testing.T) {
```
bsardo, line 192: Should we add another test case here to cover the scenario where the source object has to be created before moving the 2.5 schain to the 2.6 location? You already have a test like this for moving from 2.4 to 2.5.

openrtb_ext/convert_up.go
```diff
@@ -0,0 +1,150 @@
...
+// consent string (req.regs.ext.usprivacy) to the OpenRTB 2.6 location (req.regs.usprivacy).
+// If the OpenRTB 2.6 location is already present the OpenRTB 2.5 consent string is dropped.
+func moveUSPrivacyFrom25To26(r *RequestWrapper) {
+	// read and clear 2.5 location
+	regsExt, _ := r.GetRegExt()
+	usPrivacy25 := regsExt.GetUSPrivacy()
+	regsExt.SetUSPrivacy("")
+
+	// move to 2.6 location
+	if usPrivacy25 != "" && r.Regs.USPrivacy == "" {
```
bsardo, line 132: This is not necessarily the focus of this PR but I just noticed that you're doing a string compare on `usPrivacy25` here but for all other fields like `consent25` in the commit above it is a nil check. I took a look at the request wrapper and for every field we provide a `Get` and `Set` for, the field is a pointer. It seems like maybe this field should also be a pointer. We should investigate. Perhaps there's no impact but it is probably something worth looking into.

openrtb_ext/convert_up_test.go
```diff
@@ -0,0 +1,382 @@
...
+				Regs:   &openrtb2.Regs{GDPR: openrtb2.Int8Ptr(1), USPrivacy: "3"},
+				User:   &openrtb2.User{Consent: "1", EIDs: []openrtb2.EID{{Source: "42"}}},
+			},
+		},
+		{
+			description: "2.4 -> 2.6 + Other Ext Fields",
+			givenRequest: openrtb2.BidRequest{
+				ID:     "anyID",
+				Ext:    json.RawMessage(`{"schain":{"complete":1,"nodes":[],"ver":"2"},"other":"otherExt"}`),
+				Source: &openrtb2.Source{Ext: json.RawMessage(`{"other":"otherSource","schain":{"complete":1,"nodes":[],"ver":"2"}}`)},
```
bsardo, line 45: Maybe you want to remove `"schain":{"complete":1,"nodes":[],"ver":"2"}` from the `Source.Ext` object since that is a 2.5 schain and this test is focused on the up convert from 2.4 to 2.6?

openrtb_ext/convert_down.go
```diff
@@ -0,0 +1,147 @@
...
+	sourceExt, err := r.GetSourceExt()
+	if err != nil {
+		return err
+	}
+	sourceExt.SetSChain(schain26)
+
+	return nil
+}
+
+// moveConsentFrom26To25 modifies the request to move the OpenRTB 2.6 supply chain
```
bsardo, line 57: Typos: `moveConsentFrom26To25` and "supply chain"

openrtb_ext/convert_down.go
```diff
@@ -0,0 +1,147 @@
...
+	regExt, err := r.GetRegExt()
+	if err != nil {
+		return err
+	}
+	regExt.SetGDPR(gdpr26)
+
+	return nil
+}
+
+// moveConsentFrom26To25 modifies the request to move the OpenRTB 2.6 supply chain
```
bsardo, line 80: Typo: "supply chain"

openrtb_ext/convert_down.go
```diff
@@ -0,0 +1,147 @@
...
+	userExt, err := r.GetUserExt()
+	if err != nil {
+		return err
+	}
+	userExt.SetConsent(&consent26)
+
+	return nil
+}
+
+// moveUSPrivacyFrom26To25 modifies the request to move the OpenRTB 2.6 supply chain
```
bsardo, line 103: Typo: "supply chain"

openrtb_ext/convert_down.go
```diff
@@ -0,0 +1,147 @@
...
+	regExt, err := r.GetRegExt()
+	if err != nil {
+		return err
+	}
+	regExt.SetUSPrivacy(usprivacy26)
+
+	return nil
+}
+
+// moveEIDFrom26To25 modifies the request to move the OpenRTB 2.6 supply chain object
```
bsardo, line 126: Typo: "supply chain"

openrtb_ext/convert_down_test.go
```diff
@@ -0,0 +1,354 @@
...
+		schain1     = &openrtb2.SupplyChain{Complete: 1, Nodes: []openrtb2.SupplyChainNode{}, Ver: "1"}
+		schain1Json = json.RawMessage(`{"schain":{"complete":1,"nodes":[],"ver":"1"}}`)
+		schain2Json = json.RawMessage(`{"schain":{"complete":1,"nodes":[],"ver":"2"}}`)
+	)
+
+	testCases := []struct {
+		description     string
+		givenRequest    openrtb2.BidRequest
+		expectedRequest openrtb2.BidRequest
+		expectdErr      string
```
bsardo, line 115: 20 occurrences of typo `expectdErr` 🙂

bsardo (commented):
This is looking good. Pretty much just typos 🙂.

bsardo:
Also I see this in the corresponding issue:
`New field: $.imp[].rwdd. This replaces the extension imp.ext.prebid.is_rewarded_inventory. PBS-core and/or several bid adapters will need to be updated to support a transition period where either value could be present. Adapters looking for the extension flag: algorix, bidmachine, pangle, openx, rubicon, unicorn.`

Is there something here for us to address in PBS-Core?

openrtb_ext/convert_up_test.go
SyntaxNode: I prefer not to cover that case in the tests for `moveSupplyChainFrom25To26` and keep coverage in `TestConvertUpTo26` because this is an emergent behavior combining the two together due to the nuances of the request wrapper. I think an integration test is the most appropriate location for that coverage. Cool with you?

openrtb_ext/convert_up.go
SyntaxNode: Good point. I'll review the mxmcherry/openrtb field pointers. There's also reports that other fields should be pointers too, so I'll take a general review pass.

openrtb_ext/convert_up_test.go
SyntaxNode: Removed.

openrtb_ext/convert_up_test.go
bsardo: Good point, cool with me.

SyntaxNode:
> Also I see this in the corresponding issue: `New field: $.imp[].rwdd. This replaces the extension imp.ext.prebid.is_rewarded_inventory. PBS-core and/or several bid adapters will need to be updated to support a transition period where either value could be present. Adapters looking for the extension flag: algorix, bidmachine, pangle, openx, rubicon, unicorn.`
> 
> Is there something here for us to address in PBS-Core?

Yes. I'll handle that in a separate PR since we need to extend the request wrapper to impressions, otherwise we'll need to add a lot more marshalling.

bsardo (approved):

AlexBVolcy (approved):

PR #2355, opened 2022-08-19
Remove video endpoint and all dependencies

VeronikaSolovei9:
In this PR video endpoint and all related dependencies were removed including video endpoint code, video fetcher, structs and metrics. 

Besides these changes more components should be modified.
- Prebid Server configs: all entries with `STORED_VIDEO_REQ` should be deleted
- Prebid Server documentation
- Prebid UI: remove "video" related tab
- DataBase: remove `videoconfig_config`


adapters/smaato/smaato.go
```diff
@@ -84,7 +83,7 @@ func (adapter *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adap
 		return nil, []error{err}
 	}
 
-	isVideoEntryPoint := reqInfo.PbsEntryPoint == metrics.ReqTypeVideo
+	isVideoEntryPoint := reqInfo.PbsEntryPoint == "video" //never true
```
VeronikaSolovei9, line 86: Not sure what to do with this case. It will be never true because `eqInfo.PbsEntryPoint` cannot be `video`

bsardo:
This was an exploratory PR. We were looking to see what would need to be added to the auction endpoint to support video in accordance with ORTB 2.6. The endpoint will be removed sometime in the future.

PR #2357, opened 2022-08-22
Fix the default prefix for Prometheus metrics

nachovalen:
When no _metrics.prometheus.namespace_ and no _metrics.prometheus.subsystem_ are set, it adds "__" as prefix. eg: "__go_goroutines" (Actual behaviour).

But to make it compatible with standard Go dashboards (in Grafana), we expect metrics to be exported without any prefix. eg: "go_goroutines".

This PR adds the namespace or the subsystem only if they are set.
Please, let me know if it's not the right approach.
Thanks

### DIFF
--- a/metrics/prometheus/prometheus.go
+++ b/metrics/prometheus/prometheus.go
@@ -425,7 +425,13 @@ func NewMetrics(cfg config.PrometheusMetrics, disabledMetrics config.DisabledMet
 
 	metrics.Gatherer = reg
 
-	metricsPrefix := fmt.Sprintf("%s_%s_", cfg.Namespace, cfg.Subsystem)
+	metricsPrefix := "";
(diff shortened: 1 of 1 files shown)


SyntaxNode:
Thank you for the submission. Could you please run gofmt on the changes? The build caught this error:
`gofmt needs to be run, 1 files have issues.  Below is a list of files to review:`

nachovalen:
Thanks! PR updated

guscarreon (approved):
LGTM

SyntaxNode:
@nachovalen I'd like to share your contribution at the next Prebid Server Webinar. May I ask what company you work for / contributed this change on behalf of?

nachovalen:
I work for a company called Softonic.
Thanks

PR #2370, opened 2022-09-06
Pass Host Config Info to Bid Adapters

AlexBVolcy:
This PR addresses this issue: https://github.com/prebid/prebid-server/issues/1739

The goal of the PR is to pass specific host info to bid adapters in two different ways

This host info data we want passed consists of: `externalUrl`, `gvlID`, and `datacenter`

One way is to pass this info in the auction request itself at the path `ext.prebid.server` (i.e. `ext.prebid.server.datacenter`)

The second way is to pass this info as a configuration data structure

When it comes to adding the info to the auction request, all bid adapters get a copy of the request, so as long as I properly update the request to include these new values (and ensure they don't get cleaned), this should suffice in accomplishing the first way of passing the info.

To do this, what I needed was to create a new request object in `openrtb_ext/request.go` called `ExtRequestPrebidServer`, which will hold the three host info values we want to support. This object is then added as another parameter of the `ExtRequestPrebid` object. 

To ensure the new host info wouldn't be scrubbed from the request that's sent to bid adapters, I had to update `cleanOpenRTBRequests()` --> `buildRequestExtForBidder()` and add the new host info Server object to the `ExtRequestPrebid` object that's being built there.

When it comes to passing this info as config data structure, I created a new data structure in `config.go` called `Server` which holds the three host info values, and that object is added to the `Configuration` object. And then I updated `SetupViper()` so that the host could set the values of this object.

In order for this config object to be available to adapters, I had to update the `Builder()` function that all adapters use to build a new instance of their adapter. Simply, the `Builder()` function is updated to accept the config `Server` object (which is taken from the `Server` object defined by `Viper`), and that `Server` object is then added to that particular adapter object.

This change had to be applied to all adapters, and all adapters tests, so that's where the bulk of the files changed comes from.


adapters/33across/33across.go
```diff
@@ -15,6 +15,7 @@ import (
 
 type TtxAdapter struct {
 	endpoint string
+	Server   config.Server
```
SyntaxNode, line 18: You're not going to like this after already completing the work, but I recommend not to add the server variable to the adapter structs. Let's leave it to adapters to capture that information in the future if they'd like. For now, I think we're good with just the updates to the Builder methods.

config/config.go
```diff
@@ -597,6 +598,12 @@ type Debug struct {
 	OverrideToken       string              `mapstructure:"override_token"`
 }
 
+type Server struct {
+	ExternalUrl string `mapstructure:"external_url"`
+	GdprID      string `mapstructure:"gdpr_id"`
+	Datacenter  string `mapstructure:"datacenter"`
+}
```
SyntaxNode, line 605: The `Server` struct should just be used for communicating values down to adapters. The External URL should be copied from `Configuration.ExternalURL` and the GDPR ID should be copied from `Configuration.GDPR.HostVendorID`. The data center is the only new field we need to add to the config, which can just be at the top level IMHO.

openrtb_ext/request.go
```diff
@@ -57,7 +57,8 @@ type ExtRequestPrebid struct {
 	// NoSale specifies bidders with whom the publisher has a legal relationship where the
 	// passing of personally identifiable information doesn't constitute a sale per CCPA law.
 	// The array may contain a single sstar ('*') entry to represent all bidders.
-	NoSale []string `json:"nosale,omitempty"`
+	NoSale []string                `json:"nosale,omitempty"`
+	Server *ExtRequestPrebidServer `json:"server,omitempty"`
```
SyntaxNode, line 61: Please add this alphabetically in the above section.

adapters/33across/33across.go
AlexBVolcy: No worries, I've got the muscle memory down for updating the adapters! I'll remove this

config/config.go
AlexBVolcy: Awesome, sounds good, will update!

adapters/33across/33across.go
SyntaxNode: :joy:

exchange/utils.go
```diff
@@ -292,6 +292,7 @@ func buildRequestExtForBidder(bidder string, requestExt json.RawMessage, request
 		Debug:               requestExtParsed.Prebid.Debug,
 		CurrencyConversions: requestExtParsed.Prebid.CurrencyConversions,
 		BidderParams:        bidderParams,
+		Server:              requestExtParsed.Prebid.Server,
```
SyntaxNode, line 295: Good on adding the server info here. Please update the tests in `TestBuildRequestExtForBidder` to include test data for the new server field. Is this information being set? That is the job of PBS core logic and is not left up to the bidders.

exchange/adapter_util.go
```diff
@@ -26,7 +27,7 @@ func BuildAdapters(client *http.Client, cfg *config.Configuration, infos config.
 	return exchangeBidders, nil
 }
 
-func buildBidders(adapterConfig map[string]config.Adapter, infos config.BidderInfos, builders map[openrtb_ext.BidderName]adapters.Builder) (map[openrtb_ext.BidderName]adapters.Bidder, []error) {
+func buildBidders(adapterConfig map[string]config.Adapter, infos config.BidderInfos, builders map[openrtb_ext.BidderName]adapters.Builder, hostInfo config.Server) (map[openrtb_ext.BidderName]adapters.Bidder, []error) {
```
SyntaxNode, line 30: I recommend to rename `hostInfo` to `server` to match the name in the Builder definition and to consider keeping the builders argument as the last one in the signature.

SyntaxNode (commented):
Looks good. Thank you for the updates. I noticed two more things in this review pas.

bsardo (dismissed):

exchange/exchange.go
```diff
@@ -202,6 +204,7 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 	if err != nil {
 		return nil, err
 	}
+	requestExt.Prebid.Server = &openrtb_ext.ExtRequestPrebidServer{ExternalUrl: e.server.ExternalUrl, GvlID: e.server.GvlID, Datacenter: e.server.Datacenter}
```
bsardo, line 207: Should we only be setting this if we have at least one piece of server data? Otherwise we end up with an empty `prebid.server` object in our bid requests.

SyntaxNode: Probably a good idea. An `Empty` method on `ExtRequestPrebidServer` might be nice to check for this condition.

AlexBVolcy: That's a good check to add! Will update

config/config.go
```diff
@@ -729,7 +735,7 @@ func (cfg *Configuration) AccountDefaultsJSON() json.RawMessage {
 	return cfg.accountDefaultsJSON
 }
 
-//Allows for protocol relative URL if scheme is empty
+// Allows for protocol relative URL if scheme is empty
```
SyntaxNode, line 738: Nipick: If you're going to fix the comment, please change it to start with the function name, ie
`// GetBaseURL allows for...`

openrtb_ext/request.go
```diff
@@ -108,6 +109,12 @@ type ExtRequestPrebidCache struct {
 	VastXML *ExtRequestPrebidCacheVAST `json:"vastxml"`
 }
 
+type ExtRequestPrebidServer struct {
+	ExternalUrl string `json:"externalurl"`
+	GvlID       int    `json:"gvlid"`
+	Datacenter  string `json:"datacenter"`
```
SyntaxNode, line 115: Nitpick: Consider changing field name to `DataCenter`.

exchange/exchange.go
```diff
@@ -204,7 +204,9 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 	if err != nil {
 		return nil, err
 	}
-	requestExt.Prebid.Server = &openrtb_ext.ExtRequestPrebidServer{ExternalUrl: e.server.ExternalUrl, GvlID: e.server.GvlID, Datacenter: e.server.Datacenter}
+	if requestExt.Prebid.Server.Empty() {
+		requestExt.Prebid.Server = &openrtb_ext.ExtRequestPrebidServer{ExternalUrl: e.server.ExternalUrl, GvlID: e.server.GvlID, DataCenter: e.server.DataCenter}
+	}
```
bsardo, line 209: Ah my earlier thought here wasn't about checking if there was any server data specified in the request, but rather about checking if there was any host data so that we only set `requestExt.Prebid.Server` if `e.server` is non-empty in order to avoid creating an object at `ext.prebid.server` without any real data in it.

My understanding is that server info specified in the host config should always overwrite what is specified in the request.

@SyntaxNode, your suggestion was that it might make sense to put an `Empty` method on `ExtRequestPrebidServer`. Were you thinking the logic would then be something like this?
```
serverInfo := openrtb_ext.ExtRequestPrebidServer{ExternalUrl: e.server.ExternalUrl, GvlID: e.server.GvlID, DataCenter: e.server.DataCenter}
if !extPrebidServer.Empty() {
    requestExt.Prebid.Server = &serverInfo
}
```
Or should the `Empty` method go on type `Server` defined in `config/config.go` instead?

AlexBVolcy: Oh yes, I think I messed this up! I think you're right in saying we should only set `requestExt.Prebid.Server` when `e.server` is non-empty. And I could add an `Empty` method to check if `e.server` is non-empty.

bsardo (dismissed):
LGTM, but need confirmation from @SyntaxNode regarding handling an empty host config.

exchange/exchangetest/aliases.json
```diff
@@ -58,7 +58,13 @@
...
           ],
           "ext": {
-            "prebid": {}
+            "prebid": {
+              "server": {
+                "datacenter": "Datacenter",
+                "externalurl": "http://hosturl.com",
+                "gvlid": 1
+            }
+            }
```
SyntaxNode, line 67: Nitpick: Looks good, just please run a formatter on the updates json files. There shouldn't be too closing braces at the same indentation level. There are several json files with this, just likely from copying and pasting the new server object.

bsardo (approved):

SyntaxNode (approved):

PR #2374, opened 2022-09-12
GDPR: Basic enforcement

bsardo:
This PR adds basic enforcement. The basic enforcement algorithm may now be set as the enforcement algorithm of choice for any purpose in the host config or in an account config, with the account config setting taking precedence. If the basic algorithm is specified in a config for a given purpose or if the conditions hold to downgrade from full to basic for a particular bidder, we will execute the basic enforcement algorithm instead of the full enforcement algorithm for that bidder and purpose.

As part of this PR I did some refactoring while leaving all of the GDPR integration tests in `impl_test.go` in place. As part of the refactor:
- `allowActivities`, which was previously called from two places in `AuctionActivitiesAllowed`, has been deleted with the logic moved into `AuctionActivitiesAllowed` and reorganized so this related code is consolidated and a bit flatter.
- separate helper functions `allowID`, `allowBidRequest` and `allowGeo` were created
- the full enforcement algorithm was separated out into its own file to improve testability

Notes on the design:
- an interface `PurposeEnforcer` was created which is implemented by both the full and basic enforcement algorithm objects `FullEnforcement` and `BasicEnforcement` and these algorithms compute legal basis for a given purpose and bidder using the `LegalBasis` method
- when the `allowSync`, `allowID` and `allowBidRequest` functions are checking if a corresponding activity is allowed, they use the `purposeEnforcerBuilder` to determine which algorithm object implementing the `PurposeEnforcer` interface should be used to determine legal basis for a purpose.
- The `purposeEnforcerBuilder` minimizes the number of `PurposeEnforcer` objects that are created for a given request by caching any newly created objects for a given purpose for the life of the request. Any subsequent attempt during the life of the request to create a particular type of `PurposeEnforcer` object for a particular purpose will retrieve what is stored in memory instead of constructing a new object.


config/accounts.go
```diff
@@ -131,6 +131,18 @@ func (a *AccountGDPR) PurposeEnforced(purpose consentconstants.Purpose) (value,
 	return *a.PurposeConfigs[purpose].EnforcePurpose, true
 }
 
+// PurposeEnforcementAlgo checks the purpose enforcement algo for a given purpose by first
+// looking at the account settings, and if not set there, defaulting to the host configuration.
+func (a *AccountGDPR) PurposeEnforcementAlgo(purpose consentconstants.Purpose) (value string, exists bool) {
```
SyntaxNode, line 136: I am a bit confused as to what an algp is in this context and why the value is always returned as TCF2FullEnforcement.

config/accounts.go
```diff
@@ -131,6 +131,18 @@ func (a *AccountGDPR) PurposeEnforced(purpose consentconstants.Purpose) (value,
...
+	if a.PurposeConfigs[purpose] == nil {
+		return TCF2FullEnforcement, false
+	}
+	if a.PurposeConfigs[purpose].EnforceAlgo == TCF2FullEnforcement {
+		return TCF2FullEnforcement, true
+	}
```
SyntaxNode, line 142: I recommend using Go's response for map lookups and reuse the value response. Something like:
```
c, exists := a.PurposeConfigs[purpose]

if exists {
	return TCF2FullEnforcement, c.EnforceAlgo == TCF2FullEnforcement
}

return TCF2FullEnforcement, false
```

config/accounts.go
```diff
@@ -156,6 +168,17 @@ func (a *AccountGDPR) PurposeVendorException(purpose consentconstants.Purpose, b
...
+// PurposeVendorExceptions returns the vendor exception map for a given purpose.
+func (a *AccountGDPR) PurposeVendorExceptions(purpose consentconstants.Purpose) (value map[openrtb_ext.BidderName]struct{}, exists bool) {
+	if a.PurposeConfigs[purpose] == nil {
+		return nil, false
+	}
+	if a.PurposeConfigs[purpose].VendorExceptionMap == nil {
+		return nil, false
+	}
+	return a.PurposeConfigs[purpose].VendorExceptionMap, true
+}
```
SyntaxNode, line 180: Please see my earlier comment for re-using map responses. There could be 2 extra lookups that we don't to do in this code.

config/config.go
```diff
@@ -335,6 +326,14 @@ func (t *TCF2) PurposeEnforced(purpose consentconstants.Purpose) (value bool) {
 	return t.PurposeConfigs[purpose].EnforcePurpose
 }
 
+// PurposeEnforcementAlgo
```
SyntaxNode, line 329: IMHO it's better to omit a comment then include a blank one.

gdpr/aggregated_config.go
```diff
@@ -10,13 +10,16 @@ import (
...
+	BasicEnforcementVendors() map[string]struct{}
 	FeatureOneEnforced() bool
 	FeatureOneVendorException(openrtb_ext.BidderName) bool
 	IntegrationEnabled(config.IntegrationType) bool
 	IsEnabled() bool
 	PurposeEnforced(consentconstants.Purpose) bool
+	PurposeEnforcementAlgo(consentconstants.Purpose) string
 	PurposeEnforcingVendors(consentconstants.Purpose) bool
 	PurposeVendorException(consentconstants.Purpose, openrtb_ext.BidderName) bool
+	PurposeVendorExceptions(consentconstants.Purpose) map[openrtb_ext.BidderName]struct{}
```
SyntaxNode, line 22: IMHO the difference of names between `BasicEnforcementVendor` and `BasicEnforcementVendors` is subtle. How are they related to each other?

gdpr/aggregated_config.go
```diff
@@ -62,6 +65,15 @@ func (tc *tcf2Config) PurposeEnforced(purpose consentconstants.Purpose) bool {
 	return value
 }
 
+// PurposeEnforcementAlgo checks the purpose enforcement algo for a given purpose by first
+// looking at the account settings, and if not set there, defaulting to the host configuration.
+func (tc *tcf2Config) PurposeEnforcementAlgo(purpose consentconstants.Purpose) string {
+	if value, exists := tc.AccountConfig.PurposeEnforcementAlgo(purpose); exists {
+		return value
+	}
+	return tc.HostConfig.PurposeEnforcementAlgo(purpose)
```
SyntaxNode, line 74: Ah yes. Like this. :)

gdpr/full_enforcement.go
```diff
@@ -0,0 +1,89 @@
...
+type FullEnforcement struct {
+	cfg purposeConfig
+}
+
+// NewFullEnforcement creates a FullEnforcement object
+func NewFullEnforcement(cfg purposeConfig) *FullEnforcement {
+	return &FullEnforcement{
+		cfg: cfg,
+	}
+}
```
SyntaxNode, line 29: Why do we need a `NewFullEnforcement` method here? Can we create the `FullEnforcement` directly?

config/accounts.go
hhhjort: Is the first return value anticipating a future PR? It seems that it must always be `TCF2FullEnforcement` in this code.

gdpr/aggregated_config.go
hhhjort: If one is returning a map and the other is returning one entry from that map, would it be better to just have the map returning function and just do the lookup locally to keep the interface cleaner?

gdpr/basic_enforcement.go
```diff
@@ -0,0 +1,52 @@
...
+	if be.cfg.EnforcePurpose && consent.PurposeLITransparency(be.cfg.PurposeID) && overrides.allowLITransparency {
+		return true
+	}
+	if be.cfg.EnforcePurpose && !consent.PurposeAllowed(be.cfg.PurposeID) {
+		return false
+	}
+	if !be.cfg.EnforceVendors {
+		return true
+	}
+	if consent.VendorConsent(vendorInfo.vendorID) {
```
hhhjort, line 48: could these last two returns be simplified as `return consent.VendorConsent(vendorInfo.vendorID)` ?

config/accounts.go
bsardo: Yeah I confused myself. Originally this PR was just going to be the refactoring part and basic enforcement was going to be introduced in another PR but I ended up including basic enforcement here but never added the basic config option. I'll add it now.

config/config.go
bsardo: Oops I meant to add a description here

gdpr/full_enforcement.go
bsardo: We don't need it; we can create it directly.

config/accounts.go
bsardo: This is a mistake. The function now returns whichever algorithm (full or basic) is set on the account, and if it isn't set, it returns the default full.

gdpr/aggregated_config.go
bsardo: 👍  I removed the `BasicEnforcementVendor` and `PurposeVendorException` functions from the interface and the various config objects and now we just have the interface functions that return the maps with lookups performed locally.

gdpr/impl.go
```diff
@@ -109,120 +155,59 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, vendor
...
-const pubRestrictRequireLegitInterest = 2
-
-func (p *permissionsImpl) checkPurpose(consent tcf2.ConsentMetadata, vendor api.Vendor, vendorID uint16, purpose consentconstants.Purpose, enforceVendors, vendorException, weakVendorEnforcement bool) bool {
-	if consent.CheckPubRestriction(uint8(purpose), pubRestrictNotAllowed, vendorID) {
-		return false
-	}
+// allowBidRequest computes legal basis for a given bidder using the enforcement algorithms selected
+// by the purpose enforcer builder
+func (p *permissionsImpl) allowBidRequest(bidder openrtb_ext.BidderName, consentMeta tcf2.ConsentMetadata, vendorInfo VendorInfo) bool {
+	purpose := consentconstants.Purpose(2)
```
hhhjort, line 171: Nitpick: `purpose` is only used in the next line. Might it be better to just put this constant in as the argument directly?

gdpr/purpose_config.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"github.com/prebid/go-gdpr/consentconstants"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// purposeConfig represents all of the config info selected from the host and account configs for
+// a particular purpose needed to determine legal basis using one of the GDPR enforcement algorithms
+type purposeConfig struct {
+	PurposeID                  consentconstants.Purpose
+	EnforceAlgo                string
```
hhhjort, line 12: we seem to do a lot of comparisons with the string, and one instance of printing it out. It should only take on 2 values (3 if you count null/empty). Would it be better to set up some constants with iota, and then have a string lookup map or array to get the human readable string the one time we need it, rather than having multiple string comparisons executed for each request?

bsardo: Agreed. I added an enforcement algorithm integer representation using iota to the host config and account config. The host config is populated on startup and the account config value is set each time a request is fetched. This significantly reduces the number of string compares per request down to 10 (one per purpose) when an account is fetched. We can get the number down to 0 in the future when we refactor fetchers to cache structs instead of json.

gdpr/impl.go
```diff
@@ -61,18 +72,51 @@ func (p *permissionsImpl) AuctionActivitiesAllowed(ctx context.Context, bidderCo
...
+	basicEnforcementVendors := p.cfg.BasicEnforcementVendors()
+	_, weakVendorEnforcement := basicEnforcementVendors[string(bidder)]
+
+	if !vendorFound && !weakVendorEnforcement {
+		return DenyAll, nil
+	}
+
+	parsedConsent, vendor, err := p.parseVendor(ctx, vendorID, p.consent)
+	if err != nil {
+		return DenyAll, err
```
AlexBVolcy, line 86: This might be out of the scope of this PR, but I know you moved some stuff around for code cleanliness, but this line isn't covered by tests, could it be updated?

gdpr/impl.go
```diff
@@ -61,18 +72,51 @@ func (p *permissionsImpl) AuctionActivitiesAllowed(ctx context.Context, bidderCo
...
+			return DenyAll, nil
+		}
+	}
 
-	if id, ok := p.resolveVendorId(bidderCoreName, bidder); ok {
-		return p.allowActivities(ctx, id, bidderCoreName, weakVendorEnforcement)
-	} else if weakVendorEnforcement {
-		return p.allowActivities(ctx, 0, bidderCoreName, weakVendorEnforcement)
+	if !p.cfg.IsEnabled() {
+		return AllowBidRequestOnly, nil
```
AlexBVolcy, line 99: This line isn't covered by tests, could it be updated?

gdpr/basic_enforcement.go
```diff
@@ -0,0 +1,42 @@
...
+	if be.cfg.EnforcePurpose && consent.PurposeLITransparency(be.cfg.PurposeID) && overrides.allowLITransparency {
+		return true
+	}
+	if be.cfg.EnforcePurpose && !consent.PurposeAllowed(be.cfg.PurposeID) {
+		return false
+	}
+	if !be.cfg.EnforceVendors {
+		return true
+	}
+	return consent.VendorConsent(vendorInfo.vendorID)
```
AlexBVolcy, line 41: To clarify, is this section of if statements, a representation of those truth tables you showed me?

config/accounts.go
```diff
@@ -131,6 +120,17 @@ func (a *AccountGDPR) PurposeEnforced(purpose consentconstants.Purpose) (value,
 	return *a.PurposeConfigs[purpose].EnforcePurpose, true
 }
 
+// PurposeEnforcementAlgo checks the purpose enforcement algo for a given purpose by first
+// looking at the account settings, and if not set there, defaulting to the host configuration.
+func (a *AccountGDPR) PurposeEnforcementAlgo(purpose consentconstants.Purpose) (value TCF2EnforcementAlgo, exists bool) {
+	c, exists := a.PurposeConfigs[purpose]
```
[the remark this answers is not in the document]
bsardo: We could but I think following the Go standard of using very short names makes sense here given the limited scope of the variable, the proximity of the variable references to each other without any other unrelated logic interspersed and the size of the function. I'd argue the readability is improved with a short name here as it makes the code less verbose while making it clear that we are working with a config object called `c`.

gdpr/impl.go
bsardo: Ah yes we haven't had good test coverage in `impl.go` for a long time. I'm aware of this. I was planning to address the test coverage holes in `impl.go` in a future PR. Also a lot of the tests there will be rewritten. I wanted to keep those tests as is for now to increase confidence that my refactors aren't breaking something.

gdpr/basic_enforcement.go
bsardo: Yes that's correct. The requirements document contains truth tables for both basic and full enforcement. The `LegalBasis` method in `basic_enforcement.go` and `full_enforcement.go` implement those truth tables.

config/accounts.go
```diff
@@ -143,17 +143,14 @@ func (a *AccountGDPR) PurposeEnforcingVendors(purpose consentconstants.Purpose)
...
-	if a.PurposeConfigs[purpose] == nil {
-		return false, false
-	}
-	if a.PurposeConfigs[purpose].VendorExceptionMap == nil {
-		return false, false
-	}
-	_, found := a.PurposeConfigs[purpose].VendorExceptionMap[bidder]
+// PurposeVendorExceptions returns the vendor exception map for a given purpose.
+func (a *AccountGDPR) PurposeVendorExceptions(purpose consentconstants.Purpose) (value map[openrtb_ext.BidderName]struct{}, exists bool) {
+	c, exists := a.PurposeConfigs[purpose]
```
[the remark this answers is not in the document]
bsardo: See response regarding `c` usage in `PurposeEnforcementAlgo`.

config/config.go
```diff
@@ -335,6 +335,14 @@ func (t *TCF2) PurposeEnforced(purpose consentconstants.Purpose) (value bool) {
 	return t.PurposeConfigs[purpose].EnforcePurpose
 }
 
+// PurposeEnforcementAlgo returns the default enforcement algorithm for a given purpose
+func (t *TCF2) PurposeEnforcementAlgo(purpose consentconstants.Purpose) (value TCF2EnforcementAlgo) {
+	if c, exists := t.PurposeConfigs[purpose]; exists {
```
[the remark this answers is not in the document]
bsardo: See https://github.com/prebid/prebid-server/pull/2374#discussion_r987117224

config/config.go
```diff
@@ -344,17 +352,15 @@ func (t *TCF2) PurposeEnforcingVendors(purpose consentconstants.Purpose) (value
...
-func (t *TCF2) PurposeVendorException(purpose consentconstants.Purpose, bidder openrtb_ext.BidderName) (value bool) {
-	if t.PurposeConfigs[purpose] == nil {
-		return false
-	}
-	if _, ok := t.PurposeConfigs[purpose].VendorExceptionMap[bidder]; ok {
-		return true
+// PurposeVendorExceptions returns the vendor exception map for a given purpose if it exists, otherwise it returns
+// an empty map of vendor exceptions
+func (t *TCF2) PurposeVendorExceptions(purpose consentconstants.Purpose) (value map[openrtb_ext.BidderName]struct{}) {
+	c, exists := t.PurposeConfigs[purpose]
```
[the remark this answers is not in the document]
bsardo: See https://github.com/prebid/prebid-server/pull/2374#discussion_r987117224

gdpr/impl.go
```diff
@@ -109,120 +155,58 @@ func (p *permissionsImpl) allowSync(ctx context.Context, vendorID uint16, vendor
...
-	if consent.CheckPubRestriction(uint8(purpose), pubRestrictNotAllowed, vendorID) {
-		return false
-	}
+// allowBidRequest computes legal basis for a given bidder using the enforcement algorithms selected
+// by the purpose enforcer builder
+func (p *permissionsImpl) allowBidRequest(bidder openrtb_ext.BidderName, consentMeta tcf2.ConsentMetadata, vendorInfo VendorInfo) bool {
+	enforcer := p.purposeEnforcerBuilder(consentconstants.Purpose(2), bidder)
 
-	if vendorException {
+	if enforcer.LegalBasis(vendorInfo, bidder, consentMeta, Overrides{}) {
```
hhhjort, line 173: Isn't this just the same as
`return enforcer.LegalBasis(vendorInfo, bidder, consentMeta, Overrides{})` ?

gdpr/purpose_config.go
```diff
@@ -0,0 +1,42 @@
...
+	BasicEnforcementVendorsMap map[string]struct{}
+}
+
+// basicEnforcementVendor returns true if a given bidder is configured as a basic enforcement vendor
+// for the purpose
+func (pc *purposeConfig) basicEnforcementVendor(bidder openrtb_ext.BidderName) bool {
+	if pc.BasicEnforcementVendorsMap == nil {
+		return false
+	}
+	if _, found := pc.BasicEnforcementVendorsMap[string(bidder)]; found {
```
hhhjort, line 26: Could also be:
`_, found := pc.BasicEnforcementVendorsMap[string(bidder)]`
`return found`

gdpr/purpose_config.go
```diff
@@ -0,0 +1,40 @@
...
+	return found
+}
+
+// vendorException returns true if a given bidder is configured as a vendor exception
+// for the purpose
+func (pc *purposeConfig) vendorException(bidder openrtb_ext.BidderName) bool {
+	if pc.VendorExceptionMap == nil {
+		return false
+	}
+	if _, found := pc.VendorExceptionMap[bidder]; found {
```
hhhjort, line 36: Sorry I didn't call this one out explicitly, can be simplified the same as just above. Otherwise ready to merge I think.

AlexBVolcy (dismissed):
LGTM!

hhhjort (approved):
LGTM

AlexBVolcy (dismissed):

hhhjort (dismissed):

AlexBVolcy (dismissed):

AlexBVolcy (approved):

hhhjort (approved):

PR #2377, opened 2022-09-13
TheMediaGrid: support native mediaType

TheMediaGrid:
doc update PR - https://github.com/prebid/prebid.github.io/pull/4017


adapters/grid/grid.go
```diff
@@ -312,6 +359,9 @@ func (a *GridAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapte
 	request.Imp = validImps
 
 	reqJSON, err := json.Marshal(request)
+
+	fixedReqJSON, err := fixNative(reqJSON)
```
VeronikaSolovei9, line 363: Please do not ignore error from the previous statement

adapters/grid/grid.go
```diff
@@ -282,6 +300,35 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
...
+			if gridImp, ok := imp.(map[string]interface{}); ok {
+				native, hasNative := maputil.ReadEmbeddedMap(gridImp, "native")
+				if hasNative {
+					request, hasRequest := maputil.ReadEmbeddedString(native, "request")
+					if hasRequest {
+						delete(native, "request")
+						if err := json.Unmarshal([]byte(request), &parsedRequest); err == nil {
+							native["request_native"] = parsedRequest
+						} else {
+							native["request_native"] = request
```
VeronikaSolovei9, line 321: At this point you want to set native.request back even if `err!=nil`. What data other than do json you expect here? 
Here you want to replace `imp.native.request` to `imp.native.request_native`, that is not a part of openRTB, right?

adapters/grid/grid.go
```diff
@@ -282,6 +300,35 @@ func setImpExtData(imp openrtb2.Imp) openrtb2.Imp {
 	return imp
 }
 
+func fixNative(req json.RawMessage) (json.RawMessage, error) {
+	var gridReq map[string]interface{}
+	var parsedRequest map[string]interface{}
+
+	if err := json.Unmarshal(req, &gridReq); err != nil {
```
VeronikaSolovei9, line 307: Right before executing this function you `reqJSON, err := json.Marshal(request)`. Do you think it will be easier to pass `openrtb2.BidRequest` here instead of `req json.RawMessage`?

adapters/grid/grid.go
TheMediaGrid: That's right, but even if we got errors when trying to parse json , we still want to get native.request to be able to see what the value is

adapters/grid/grid.go
TheMediaGrid: I guess it would be easier to operate on abstract json since we need to change its structure

adapters/grid/grid.go
VeronikaSolovei9: Thank you for the reply. 
Line 302 is not covered by unit tests and the way to do this - is to add test similar to "simple-native.json" and set `req.imp[].native.request` to string like "hello".  It will throw the Unmarshal error and L304 will be executed. 
Please add this kind of test.

VeronikaSolovei9:
Thank you for the updates. 
I didn't find a better way to rename json field, so I'm ok with current implementation. Please add unit test I described in previous  comment.

adapters/grid/grid.go
VeronikaSolovei9: I see it now, ok!

AlexBVolcy (approved):
LGTM

VeronikaSolovei9 (approved):

Issue #2178, opened 2022-03-08
Request imp format can generate errors for adapters that don't support a certain format.


hhhjort:
PBS core has a filter step that removes any formats from an imp sent to an adapter that the adapter does not support. This is good, except we also generate an error in the response indicating that an adapter does not support a particular format. This can add unnecessary noise to analytics, in some cases such errors are unavoidable.

For example, adapter A supports banner and native. Adapter B banner only. Publisher has a multiformat imp that supports both banner and native. To avoid the error from being generated for adapter B, the publisher will have to send a banner only request, missing out on any native demand available from adapter A.

I suggest we downgrade this from an error to a warning, or perhaps forgo it completely as long as there is at least one adapter that supports each format supplied. We probably still want an error if there are no adapters able to service a format requested in a request.

bretg:
Got a report that PBS-Java also have an issue in this area. It's not enforcing the mediatype rules at all.e.g. if a bid adapter's config doesn't list 'native', it's not removing imp.native from the request before sending to the adapter. Will open an internal ticket.

coreyb-cbs:
We recently hit this error with one of our bidders (PBS-Java). We pulled the bidder until a resolution is found. To me, it only looks to be an issue if the bidder test checks for unsupported media types. I say this because we have other s2s bidders in play that do not support native but are not producing an error because they do not check for such in their bidder test. Ideally, PBS core would remove unsupported media types from the imp, but to me it looks like the current structure of the imp wouldn't allow this because the media types are not specific to the bidder - they are applied to the slot.

bretg:
@coreyb-cbs - PBS-core makes an adapter-specific version of the ORTB request. It will remove the mediatypes from that copy of the object. Will move that fix up in the priority a bit.

coreyb-cbs:
@bretg following up... was a fix/PR ever created for this?

bretg:
A PBS-Java update is in the works. Would guess it can be released with v1.90. i.e. not the next release (1.89), but the one after.

bsardo:
@AlexBVolcy will be handling this for PBS-Go

bretg:
Released with PBS-Java 1.92

SyntaxNode:
Implemented in PBS-Go 0.228.0.

PR #2385, opened 2022-09-20
Downgrade errors to warnings when adapter doesn't support imp format

AlexBVolcy:
This PR addresses this issue: https://github.com/prebid/prebid-server/issues/2178

The goal of this PR is: In the situation when an adapter doesn't support the format indicated from the request imp, an error is thrown that we want downgraded to a warning since this error is unnecessary to be written to analytics

This error is thrown from the bidder wrapper `InfoAwareBidder`'s call too `MakeRequests()`. So for all the error checks that check  if an adapter doesn't support the imp format (i.e. `{&errortypes.BadInput{Message: "this bidder does not support site requests"}`), we downgrade those to a warning (i.e. `{&errortypes.Warning{Message: "this bidder does not support site requests"}`)

Tests were updated with this change.

### DIFF
--- a/adapters/infoawarebidder.go
+++ b/adapters/infoawarebidder.go
@@ -13,12 +13,12 @@ import (
 // media types defined in the static/bidder-info/{bidder}.yaml file.
 //
 // It adjusts incoming requests in the following ways:
-//   1. If App or Site traffic is not supported by the info file, then requests from
-//      those sources will be rejected before the delegate is called.
-//   2. If a given MediaType is not supported for the platform, then it will be set
-//      to nil before the request is forwarded to the delegate.
-//   3. Any Imps which have no MediaTypes left will be removed.
-//   4. If there are no valid Imps left, the delegate won't be called at all.
+//  1. If App or Site traffic is not supported by the info file, then requests from
+//     those sources will be rejected before the delegate is called.
+//  2. If a given MediaType is not supported for the platform, then it will be set
+//     to nil before the request is forwarded to the delegate.
+//  3. Any Imps which have no MediaTypes left will be removed.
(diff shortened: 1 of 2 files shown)


adapters/infoawarebidder.go
```diff
@@ -78,19 +78,19 @@ func pruneImps(imps []openrtb2.Imp, allowedTypes parsedSupports) (int, []error)
 	for i := 0; i < len(imps); i++ {
 		if !allowedTypes.banner && imps[i].Banner != nil {
 			imps[i].Banner = nil
-			errs = append(errs, &errortypes.BadInput{Message: fmt.Sprintf("request.imp[%d] uses banner, but this bidder doesn't support it", i)})
+			errs = append(errs, &errortypes.Warning{Message: fmt.Sprintf("request.imp[%d] uses banner, but this bidder doesn't support it", i)})
```
SyntaxNode, line 81: We may want to add a new warning code for this case where a media type is not supported.

bsardo:
@AlexBVolcy was any consideration given to @hhhjort's suggestion on the issue that we may still want an error if there are no adapters able to service a format requested in a request? At a glance, it looks like this would require a refactor of the core bidder loop or could perhaps introduce added complexity that we don't feel is worth it but I figured I should ask.

bsardo:
This was discussed offline. The suggestion to throw an error if there are no adapters able to service a requested format is a behavior change outside of the scope of this PR.

bsardo (approved):
LGTM

SyntaxNode (approved):
Looks good. The return statements haven't changed so the logic flow is the same, it's just the error types that changed. Tested locally and new warnings look good.

PR #2386, opened 2022-09-21
Seeding Alliance: Support ext.prebid.type and endpoint update

sag-celejd:


adapters/seedingAlliance/seedingAlliance_test.go
```diff
@@ -42,32 +42,44 @@ func TestResolvePriceMacro(t *testing.T) {
...
+		}
+
+		if err := json.Unmarshal(bid.Ext, &extBid); err != nil {
+			t.Fatalf("could not unmarshal extBid: %v", err)
+		}
+
+		got, gotErr := getMediaTypeForBid(bid.Ext)
 		assert.Equal(t, test.want, got)
+		if gotErr != nil && !test.wantErr {
+			t.Fatalf("wantErr: %v, gotErr: %v", test.wantErr, gotErr)
```
AlexBVolcy, line 81: Could you update this test, for when you want an error, that you assert it with the error you're anticipating?

sag-celejd: Sure, check the latest commit.

adapters/seedingAlliance/seedingAlliance_test.go
```diff
@@ -42,32 +42,109 @@ func TestResolvePriceMacro(t *testing.T) {
...
+
+		if err := json.Unmarshal(bid.Ext, &extBid); err != nil {
+			t.Fatalf("could not unmarshal extBid: %v", err)
+		}
+
+		got, gotErr := getMediaTypeForBid(bid.Ext)
 		assert.Equal(t, test.want, got)
+
+		if gotErr != nil {
+			if test.wantErr {
```
VeronikaSolovei9, line 142: `if test.wantErr` should be executed first. This will guarantee errors will be checked.
In case `gotErr` is nil where error should be returned - test will not catch it

adapters/seedingAlliance/seedingAlliance.go
```diff
@@ -77,42 +77,41 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
...
-	for _, imp := range imps {
-		if imp.ID == impId {
-			switch {
-			case imp.Native != nil:
-				return openrtb_ext.BidTypeNative
-			case imp.Banner != nil:
-				return openrtb_ext.BidTypeBanner
-			}
-		}
+	if err := json.Unmarshal(ext, &bidExt); err != nil || bidExt.Prebid == nil {
```
VeronikaSolovei9, line 110: Nitpick: `bidExt.Prebid == nil` should be a separate check. 
Please refer to this code for cleaner example: https://github.com/prebid/prebid-server/blob/f3e440389e3d57305be796c39359ff1fb09784b1/adapters/adf/adf.go#L143

adapters/seedingAlliance/seedingAlliance_test.go
sag-celejd: Good point, see my last commit.

adapters/seedingAlliance/seedingAlliance.go
sag-celejd: Don't think that example is any cleaner/more readable - actually way worse since the happy path is nested.
I separated the nil check though in the last commit.

adapters/seedingAlliance/seedingAlliance_test.go
VeronikaSolovei9: Thank you!

adapters/seedingAlliance/seedingAlliance.go
VeronikaSolovei9: Ok, this works too!

VeronikaSolovei9 (approved):

AlexBVolcy (approved):

PR #2390, opened 2022-09-26
Changed FS bidder names to mixed case

VeronikaSolovei9:


config/bidderinfo.go
```diff
@@ -170,38 +170,66 @@ func (bi BidderInfo) IsEnabled() bool {
...
 }
 
-// LoadBidderInfo parses all static/bidder-info/{bidder}.yaml files from the file system.
-func LoadBidderInfo(path string) (BidderInfos, error) {
-	bidderConfigs, err := ioutil.ReadDir(path)
+type InfoReader interface {
+	Read() (map[string][]byte, error)
+}
+
+type InfoReaderFromDisk struct {
```
SyntaxNode, line 177: Yay! I think this will make the code more testable.

config/bidderinfo.go
```diff
@@ -170,38 +170,66 @@ func (bi BidderInfo) IsEnabled() bool {
...
-	infos := BidderInfos{}
-
+	bidderInfos := make(map[string][]byte)
 	for _, bidderConfig := range bidderConfigs {
 		if bidderConfig.IsDir() {
 			continue //ignore directories
 		}
 		fileName := bidderConfig.Name()
-		filePath := fmt.Sprintf("%v/%v", path, fileName)
+		filePath := fmt.Sprintf("%v/%v", r.Path, fileName)
```
SyntaxNode, line 193: Nitpick: This should really be `filepath.Join(r.Path, fileName)`. The original author of that code got it wrong. :-P

config/bidderinfo.go
```diff
@@ -170,38 +170,66 @@ func (bi BidderInfo) IsEnabled() bool {
...
 		bidderName := strings.Split(fileName, ".")
 		if len(bidderName) == 2 && bidderName[1] == "yaml" {
-			// all bidder names loaded from the file system should be in lowercase
-			// to match bidder names in Viper config bidder infos
-			// this will prevent from missing bidder config overrides
-			// in func applyBidderInfoConfigOverrides
-			infos[(strings.ToLower(bidderName[0]))] = info
+			normalizedBidderName, bidderNameExists := openrtb_ext.NormalizeBidderName(bidderName[0])
+			if !bidderNameExists {
+				return nil, fmt.Errorf("error finding bidder %s in bidders list", fileName)
```
SyntaxNode, line 230: Nitpick: Consider wording along the lines of "error parsing config for bidder %s: unknown bidder"

config/bidderinfo.go
```diff
@@ -170,38 +170,66 @@ func (bi BidderInfo) IsEnabled() bool {
...
 		}
 
 		bidderName := strings.Split(fileName, ".")
 		if len(bidderName) == 2 && bidderName[1] == "yaml" {
-			// all bidder names loaded from the file system should be in lowercase
-			// to match bidder names in Viper config bidder infos
-			// this will prevent from missing bidder config overrides
-			// in func applyBidderInfoConfigOverrides
-			infos[(strings.ToLower(bidderName[0]))] = info
+			normalizedBidderName, bidderNameExists := openrtb_ext.NormalizeBidderName(bidderName[0])
```
SyntaxNode, line 228: Consider accepting `NormalizeBidderName` as a parameter so it can be faked for tests. I see you updated a lot of test files for this change. Perhaps it would be best to keep the tests separated from this dependency.

config/bidderinfo.go
```diff
@@ -351,7 +379,11 @@ func validateSyncer(bidderInfo BidderInfo) error {
 
 func applyBidderInfoConfigOverrides(configBidderInfos BidderInfos, fsBidderInfos BidderInfos) (BidderInfos, error) {
 	for bidderName, bidderInfo := range configBidderInfos {
-		if fsBidderCfg, exists := fsBidderInfos[bidderName]; exists {
+		normalizedBidderName, bidderNameExists := openrtb_ext.NormalizeBidderName(bidderName)
+		if !bidderNameExists {
+			return nil, fmt.Errorf("error finding bidder in bidders list by bidder name loaded in configs for bidder %s. Configs include data in environment variables and pbs.json or pbs.yaml files", bidderName)
```
SyntaxNode, line 384: Nitpick: Suggested wording simplification (and similarly for the other error message):
`error setting configuration for bidder %s: unknown bidder`

config/bidderinfo.go
VeronikaSolovei9: Done!

config/bidderinfo_test.go
```diff
@@ -36,14 +41,14 @@ endpointCompression: "GZIP"
 `
 
 func TestLoadBidderInfoFromDisk(t *testing.T) {
-	// should appear in result in lowercase
-	bidder := "somebidder"
+	// should appear in result in mixed case
+	bidder := "stroeerCore"
```
SyntaxNode, line 45: For historical reasons: `stroeerCore` was chosen at random just because we needed a real mixed case bidder name. There is no other significance to this choice. Our usual real bidder choices do not have mixed upper and lower case in their name.

SyntaxNode (approved):

config/bidderinfo.go
```diff
@@ -170,38 +171,66 @@ func (bi BidderInfo) IsEnabled() bool {
...
 		if len(bidderName) == 2 && bidderName[1] == "yaml" {
-			// all bidder names loaded from the file system should be in lowercase
-			// to match bidder names in Viper config bidder infos
-			// this will prevent from missing bidder config overrides
-			// in func applyBidderInfoConfigOverrides
-			infos[(strings.ToLower(bidderName[0]))] = info
+			normalizedBidderName, bidderNameExists := normalizeBidderName(bidderName[0])
+			if !bidderNameExists {
+				return nil, fmt.Errorf("error parsing config for bidder %s: unknown bidder", fileName)
+			}
```
guscarreon, line 232: A map check is less expensive than unmarshalling the `yaml` file. A small optimization would occur if we switch the order of the operations:
```
213   func processBidderInfos(reader InfoReader, normalizeBidderName func(string) (openrtb_ext.BidderName, bool)) (BidderInfos, error) {
214       bidderConfigs, err := reader.Read()
215       if err != nil {
216           return nil, fmt.Errorf("error loading bidders data")
217       }
218   
219       infos := BidderInfos{}
220   
221       for fileName, data := range bidderConfigs {
222 -         info := BidderInfo{}
223 -         if err := yaml.Unmarshal(data, &info); err != nil {
224 -             return nil, fmt.Errorf("error parsing config for bidder %s: %v", fileName, err)
225 -         }
226 - 
227           bidderName := strings.Split(fileName, ".")
228           if len(bidderName) == 2 && bidderName[1] == "yaml" {
229               normalizedBidderName, bidderNameExists := normalizeBidderName(bidderName[0])
230               if !bidderNameExists {
231                   return nil, fmt.Errorf("error parsing config for bidder %s: unknown bidder", fileName)
232               }
    + 
    +             info := BidderInfo{}
    +             if err := yaml.Unmarshal(data, &info); err != nil {
    +                 return nil, fmt.Errorf("error parsing config for bidder %s: %v", fileName, err)
    +             }
    + 
233               infos[string(normalizedBidderName)] = info
234           }
235       }
236       return infos, nil
config/bidderinfo.go
```

main.go
```diff
@@ -59,7 +60,7 @@ const infoDirectory = "./static/bidder-info"
 func loadConfig(bidderInfos config.BidderInfos) (*config.Configuration, error) {
 	v := viper.New()
 	config.SetupViper(v, configFileName, bidderInfos)
-	return config.New(v, bidderInfos)
+	return config.New(v, bidderInfos, openrtb_ext.NormalizeBidderName)
```
guscarreon, line 63: Do we need to pass `openrtb_ext.NormalizeBidderName` from `main.go` all the way down to `config/bidderinfo.go`'s `applyBidderInfoConfigOverrides(...)` function? I see the need for `bidderInfos` to be passed from `main.go` and also the convenience of passing a mock function to `applyBidderInfoConfigOverrides(...)` for testing purposes but, I'm wondering if maybe we can simplify `config.New(...)` by not adding that third paramenter:
```
619   // New uses viper to get our server configurations.
620 - func New(v *viper.Viper, bidderInfos BidderInfos) (*Configuration, error) {
    + func New(v *viper.Viper, bidderInfos BidderInfos, normalizeBidderName func(string) (openrtb_ext.BidderName, bool)) (*Configuration, error) {
621       var c Configuration
622       if err := v.Unmarshal(&c); err != nil {
623   *-- 78 lines: return nil, fmt.Errorf("viper failed to unmarshal app config: %v", err)------------------------------------
701       resolvedStoredRequestsConfig(&c)
702   
703 -     mergedBidderInfos, err := applyBidderInfoConfigOverrides(c.BidderInfos, bidderInfos, normalizeBidderName)
    +     mergedBidderInfos, err := applyBidderInfoConfigOverrides(c.BidderInfos, bidderInfos, openrtb_ext.NormalizeBidderName)
704       if err != nil {
705           return nil, err
706       }
config/config.go
```
Therefore we could rollback to:
```
60   func loadConfig(bidderInfos config.BidderInfos) (*config.Configuration, error) {
61       v := viper.New()
62       config.SetupViper(v, configFileName, bidderInfos)
63 -     return config.New(v, bidderInfos, openrtb_ext.NormalizeBidderName)
   +     return config.New(v, bidderInfos)
64   }
main.go
```
Unless passing it from `main.go` makes something happen behind the scenes that I'm missing.

guscarreon (commented):
Looks pretty good. I left a couple of questions but feel like approving either way.

config/bidderinfo.go
SyntaxNode: Sounds like a good idea. I would like to mention this code runs once at startup so performance is not a critical concern with this flow.

main.go
SyntaxNode: It is necessary for `New` to accept the normalize function so unit tests are free to use fake bidder names instead of being forced to use real ones. The bidder name flow will continue to evolve and in the near future the normalize func will likely no longer be required. We're focused on restoring functionality to fix the bug and will design a better solution as a follow-up.

VeronikaSolovei9: Yes, I was thinking about this too. We also need this parameter in config_test.go -> `newDefaultConfig`

guscarreon (approved):
LGTM

SyntaxNode (approved):

PR #2392, opened 2022-09-27
add freewheelssp go adapter 

mwang-sticky:


adapters/freewheelssp/freewheelssp.go
```diff
@@ -0,0 +1,119 @@
...
+type adapter struct {
+	endpoint string
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	if len(request.Imp) == 0 {
+		return nil, []error{&errortypes.BadInput{
+			Message: "No impression in bid request.",
+		}}
+	}
```
SyntaxNode, line 28: This check is not needed. The core logic of Prebid Server will never call `MakeRequests` without at least one impression.

adapters/freewheelssp/freewheelssp.go
```diff
@@ -0,0 +1,119 @@
...
+
+	for i := 0; i < len(request.Imp); i++ {
+		imp := &request.Imp[i]
+		var bidderExt adapters.ExtImpBidder
+		if err := json.Unmarshal(imp.Ext, &bidderExt); err != nil {
+			return nil, []error{&errortypes.BadInput{
+				Message: fmt.Sprintf("Invalid imp.ext, %s", err.Error()),
```
SyntaxNode, line 35: Consider adding the imp id or index to the error message to give the publisher a better idea of where the mistake is made. (both here and on line 42)

adapters/freewheelssp/freewheelssp.go
```diff
@@ -0,0 +1,119 @@
...
+		if imp.Ext, err = json.Marshal(requestImpExt); err != nil {
+			return nil, []error{&errortypes.BadInput{
+				Message: fmt.Sprintf("Unable to transfer requestImpExt to Json fomat, %s", err.Error()),
+			}}
+		}
+	}
+
+	requestJSON, err := json.Marshal(request)
+	headers := http.Header{}
+	headers["componentId"] = []string{"prebid-go"}
```
SyntaxNode, line 59: Please move the header declaration to after the error check.

SyntaxNode, line 59: Consider using: `headers.Add("componentId", "prebid-go")`.

SyntaxNode, line 59: While this is perfectly acceptable, did you know that PBS will automatically add a x-prebid version header which includes if it's go or java? That's relatively new behavior.

adapters/freewheelssp/freewheelssp.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
+		}}
+	}
+
+	var bidResp openrtb2.BidResponse
+	if err := json.Unmarshal(response.Body, &bidResp); err != nil {
+		return nil, []error{err}
+	}
+	cur := currency.EUR.String()
```
SyntaxNode, line 91: Just want to verify you wish to default to EUR. The default for OpenRTB is USD.

adapters/freewheelssp/freewheelssp.go
```diff
@@ -0,0 +1,119 @@
...
+
+	for _, seatBid := range bidResp.SeatBid {
+		for _, bid := range seatBid.Bid {
+			bidTmp := bid
+			bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
+				Bid:     &bidTmp,
+				BidType: bidType,
+			})
+		}
+	}
```
SyntaxNode, line 110: Please update the seat bid loop to match our latest recommended approach per https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#adapter-code

The use of an index rather than the bid iteration variable is an important distinction.

```
    for i := range seatBid.Bid {
      b := &adapters.TypedBid{
        Bid:     &seatBid.Bid[i],
        BidType: bidType,
      }
  }
```

adapters/freewheelssp/freewheelssp_test.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/prebid/prebid-server/adapters/adapterstest"
+	"github.com/prebid/prebid-server/config"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"testing"
+)
+
+func TestJsonSamples(t *testing.T) {
+	bidder, buildErr := Builder(openrtb_ext.BidderFreewheelSSP, config.Adapter{
+		Endpoint: "https://ads.stickyadstv.com/openrtb/dsp"})
```
SyntaxNode, line 12: There is no need for this to be a real endpoint. It's just for the tests.

exchange/adapter_builders.go
```diff
@@ -68,6 +68,7 @@ import (
 	"github.com/prebid/prebid-server/adapters/engagebdr"
 	"github.com/prebid/prebid-server/adapters/eplanning"
 	"github.com/prebid/prebid-server/adapters/epom"
+	freewheelssp "github.com/prebid/prebid-server/adapters/freewheelssp"
```
SyntaxNode, line 71: The package alias should not be necessary.

adapters/freewheelssp/freewheelssp.go
mwang-sticky: thanks, I removed this

adapters/freewheelssp/freewheelssp.go
mwang-sticky: got it

adapters/freewheelssp/freewheelssp.go
mwang-sticky: thanks, I used the new one

mwang-sticky: Thanks for your information, we also need this key to unify different request's type

adapters/freewheelssp/freewheelssp.go
mwang-sticky: we changed to USD, thanks

adapters/freewheelssp/freewheelssp.go
mwang-sticky: got it, thanks.

adapters/freewheelssp/freewheelssp.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (a *adapter) MakeBids(internalRequest *openrtb2.BidRequest, externalRequest *adapters.RequestData, response *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if response.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
```
AlexBVolcy, line 76: Is there a way to cover this error case with tests?

AlexBVolcy (commented):
Could you please link your docs PR?

SyntaxNode (commented):
Looks good. Thank you for the fixes.

FYI: Prebid Server has 6 characters of guaranteed bidder name uniqueness. For `freehweelssp` that would be `freehw`. I'm sharing this because your adapter name is `freewheelssp` instead of just `freewheel`, so if you plan to add other adapters within the freewheel family in the future, the targeting keys may not be unique between them. Not a problem, just wanted you to be aware.

bsardo:
Hi @mwang-sticky, we'll check out your changes shortly. Also, if you don't mind, could you please avoid force pushing after the PR has been reviewed? We prefer you just push another commit with your changes as this helps the reviewers save a lot of time since they often only need to review the delta. And if conflicts arise they can be resolved by merging with master instead of rebasing with a force push 🙂. Thanks again for your contribution.

AlexBVolcy (commented):
This looks good, but I still don't see a link to your docs PR, could you please add this as a comment on this PR

mwang-sticky:
@bsardo @AlexBVolcy 
Sorry for force commit to resolve some conflict,
I add the doc request: https://github.com/prebid/prebid.github.io/pull/4094
however our prebid js still use freeweel-ssp which is not supported by GAM, then prebid-go has changed to new biddercode freewheelssp, please tell me how can I solve this difference, does it impact prebid integration?
I look forward your help.

bsardo:
Hi @mwang-sticky, we strongly prefer prebid.js and PBS use the same bidder codes. If you use different biddercodes it confuses publishers and leads to confusing documentation.

It sounds like you want to move away from `freewheel-ssp` to `freewheelssp` with one of the main reasons being the new name is supported by GAM. Can you confirm?

If your long term goal is to standardize on `freewheelssp`, then we recommend one of the following two courses of action:
1. If you plan to update prebid.js with the new name `freewheelssp` in the immediate future, then you should be ok just supporting `freewheelssp` in PBS since you'll effectively be forcing publishers to update to the new name even if they just want to use prebid.js with an imminent prebid.js update.
2. If you plan to update prebid.js with the new name eventually (say in 2023 or beyond), then we recommend adding an alias in PBS so publishers can easily switch between the two names until prebid.js catches up. In order to do this, you would create a hardcoded alias where you have two bidder names of type `openrtb_ext.BidderName` and a single adapter of type `adapters.Builder` that maps to each of the two bidder names. See [PR#2302](https://github.com/prebid/prebid-server/pull/2302) as an example of how to add a hardcoded alias in PBS.

I hope this helps. If I misunderstood your intention, let me know.

mwang-sticky:
> It sounds like you want to move away from `freewheel-ssp` to `freewheelssp` with one of the main reasons being the new name is supported by GAM. Can you confirm?
Yes, some of our publishers asked us to change the old biddercode. 
> If your long term goal is to standardize on `freewheelssp`, then we recommend one of the following two courses of action:

Thanks for your good suggestion. I add the old bidder code in go adapter.
Our js team has add the new bidder name, however because some of our clients has used the old prebid js, then we don't want to impact a lot. https://github.com/prebid/Prebid.js/pull/9093

Does this change work for prebid go/js integration?

bsardo:
Hi @mwang-sticky, I'm sorry I'm not able to confirm your PBS and prebid.js integration works as that is outside of the scope of our team. I do believe though that as long as prebid.js and PBS support the same bidder parameters for your bidder and you each has support for both bidder codes, the integration should work.

mwang-sticky:
@bsardo 
As the biddercode is conflicted, how can I to solve this error

`
bidders_validate_test.go:62: 
        	Error Trace:	bidders_validate_test.go:62
        	Error:      	"10" is not less than or equal to "6"
        	Test:       	`TestBidderUniquenessGatekeeping`
`
Thanks,

SyntaxNode:
> @bsardo As the biddercode is conflicted, how can I to solve this error
> 
> ` bidders_validate_test.go:62: Error Trace: bidders_validate_test.go:62 Error: "10" is not less than or equal to "6" Test: `TestBidderUniquenessGatekeeping` ` Thanks,

This is to ensure the bidder name is unique enough in targeting keys. The uniqueness constraint is not usually important for aliases. So long as a bid from `freewheelssp` and `freewheel-ssp` are the same, then you can add an exception along side Triplelift and Adkernel. ie:

```
if bidder != BidderTripleliftNative && bidder != BidderAdkernelAdn && bidder != BidderFreewheelSSPOld {
```

mwang-sticky:
> if bidder != BidderTripleliftNative && bidder != BidderAdkernelAdn && bidder != BidderFreewheelSSPOld {
I updated this, should I update other info？

AlexBVolcy (commented):
I'm ready to approve, but @bsardo I just want to confirm how we want to go about PRs that are throwing this gofmt error in the Github validation?

bsardo:
Hi @mwang-sticky, can you merge with master when you get a chance? Your code is failing the go 1.19 validation checks due to `go fmt` changes that were introduced with the recent upgrade to go 1.19 on master.

mwang-sticky:
@bsardo I synced the master to my branch, thanks for your reminder.

AlexBVolcy (approved):

VeronikaSolovei9 (approved):
Thank you for addressing comments, lgtm

AlexBVolcy (approved):
LGTM

PR #2402, opened 2022-10-03
consumable adapter: add eid support

jpiros:
This PR adds support for User Ids

### DIFF
--- a/adapters/consumable/consumable.go
+++ b/adapters/consumable/consumable.go
@@ -173,7 +173,7 @@ func (a *ConsumableAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *
 			body.GDPR = &gdpr
 
 			for i := 0; i < len(extUser.Eids); i++ {
-				if extUser.Eids[i].Source == "adserver.org" && len(extUser.Eids[i].UIDs) > 0 {
+				if len(extUser.Eids[i].UIDs) > 0 {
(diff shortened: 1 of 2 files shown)


adapters/consumable/consumable.go
```diff
@@ -170,6 +171,15 @@ func (a *ConsumableAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *
...
 			body.GDPR = &gdpr
+
+			for i := 0; i < len(extUser.Eids); i++ {
+				if len(extUser.Eids[i].UIDs) > 0 {
+					if extUser.Eids[i].UIDs[0].ID != "" {
+						body.User.Eids = extUser.Eids
+					}
+					break
+				}
+			}
```
SyntaxNode, line 182: Might be better expressed as a method, something like:

```
if eidsNotEmpty(extUser.Eids) {
	body.User.Eids = extUser.Eids
}

...

func eidsNotEmpty(eids []openrtb2.EID) bool {
	for i := 0; i < len(eids); i++ {
		if len(eids[i].UIDs) > 0 && eids[i].UIDs[0].ID != "" {
			return true
		}
	}
	return false
}
```

jpiros: updated, thanks for the suggestion

SyntaxNode (approved):

AlexBVolcy (approved):

Issue #2044, opened 2021-10-13
Integration vs channel


bretg:
It was pointed out that some variable names refer to "integration", but are actually referring to "channel" values.

Follow up to https://github.com/prebid/prebid-server/issues/1428

In config/accounts.go
```
// IntegrationType enumerates the values of integrations Prebid Server can configure for an account
type IntegrationType string

// Possible values of integration types Prebid Server can configure for an account
const (
	IntegrationTypeAMP   IntegrationType = "amp" // should be "ChannelAMP"
	IntegrationTypeApp   IntegrationType = "app"  // should be "ChannelApp"
	IntegrationTypeVideo IntegrationType = "video"   // should be "ChannelVideo"
	IntegrationTypeWeb   IntegrationType = "web"  // should be "ChannelWeb"
)
```
Similarly, in PBS-Java config, privacy.*.integration-enabled should be "channel-enabled"
```
          integration-enabled:     // "channel-enabled" please
            video: true
            web: true
            app: true
            amp: true
```

This change would be one that host companies would need to be aware of. Will discuss in the next PBS meeting.

bretg:
The first phase of this released with PBS-Java 1.79. PBS-Java users should consider checking their databases and migrating "integration-enabled" to "channel-enabled". At some point in the near future, this update will be required.

bsardo:
`IntegrationType` has been changed to `ChannelType` in PBS-Go and the corresponding instances have been renamed:
```
// ChannelType enumerates the values of integrations Prebid Server can configure for an account
type ChannelType string

// Possible values of channel types Prebid Server can configure for an account
const (
	ChannelAMP   ChannelType = "amp"
	ChannelApp   ChannelType = "app"
	ChannelVideo ChannelType = "video"
	ChannelWeb   ChannelType = "web"
	ChannelDOOH  ChannelType = "dooh"
)
```

Both `integration-enabled` and `channel-enabled` have been supported for about a year at this point and we recently removed support for `integration-enabled` as part of the 2.0 release initiative.

PR #2404, opened 2022-10-04
Integration to Channel Name Changes

AlexBVolcy:
This PR addresses this issue https://github.com/prebid/prebid-server/issues/2044


config/accounts.go
```diff
@@ -45,13 +45,16 @@ type CookieSync struct {
 type AccountCCPA struct {
 	Enabled            *bool              `mapstructure:"enabled" json:"enabled,omitempty"`
 	IntegrationEnabled AccountIntegration `mapstructure:"integration_enabled" json:"integration_enabled"`
+	ChannelEnabled     AccountIntegration `mapstructure:"channel_enabled" json:"channel_enabled"`
```
bsardo, line 48: It appears we are keeping both `IntegrationEnabled` and `ChannelEnabled` in the account config for now so that hosts have an opportunity to migrate from `integration_enabled` to `channel_enabled`, is that correct? I see this approach described in the associated issue and that PBS-Java has taken a similar approach where they eventually removed `integration_enabled` completely after the transition period which I assume we will also do in a future release.

config/accounts.go
```diff
@@ -12,10 +12,10 @@ type IntegrationType string
```
bsardo, line 12: I think we should change this to `type ChannelType string` since `IntegrationType` will eventually go away.

exchange/utils.go
```diff
@@ -25,19 +25,19 @@ import (
 )
 
 var integrationTypeMap = map[metrics.RequestType]config.IntegrationType{
```
bsardo, line 27: Let's use the proper naming here: `var channelTypeMap = map[metrics.RequestType]config.ChannelType{`

config/accounts.go
```diff
@@ -45,13 +45,16 @@ type CookieSync struct {
 type AccountCCPA struct {
 	Enabled            *bool              `mapstructure:"enabled" json:"enabled,omitempty"`
 	IntegrationEnabled AccountIntegration `mapstructure:"integration_enabled" json:"integration_enabled"`
+	ChannelEnabled     AccountIntegration `mapstructure:"channel_enabled" json:"channel_enabled"`
 }
 
 // EnabledForIntegrationType indicates whether CCPA is turned on at the account level for the specified integration type
 // by using the integration type setting if defined or the general CCPA setting if defined; otherwise it returns nil
 func (a *AccountCCPA) EnabledForIntegrationType(integrationType IntegrationType) *bool {
```
bsardo, line 53: I think all references to `IntegrationType` across the repo should be changed to the correct name `ChannelType`, whether that be in function names, types, variable names, etc.

config/accounts.go
```diff
@@ -45,13 +45,16 @@ type CookieSync struct {
...
 	if integrationEnabled := a.IntegrationEnabled.GetByIntegrationType(integrationType); integrationEnabled != nil {
 		return integrationEnabled
+	} else if channelEnabled := a.ChannelEnabled.GetByIntegrationType(integrationType); channelEnabled != nil {
+		return channelEnabled
 	}
```
bsardo, line 58: It looks like this is the `if-else if` block that will be altered in the follow up PR when we completely phase out support for `integration-enabled` in the account config.

config/accounts.go
```diff
@@ -211,13 +217,13 @@ func (a *AccountIntegration) GetByIntegrationType(integrationType IntegrationTyp
 	var integrationEnabled *bool
```
bsardo, line 217: This function has various references to integration that should be channel.

gdpr/aggregated_config.go
```diff
@@ -13,6 +13,7 @@ type TCF2ConfigReader interface {
 	FeatureOneEnforced() bool
 	FeatureOneVendorException(openrtb_ext.BidderName) bool
 	IntegrationEnabled(config.IntegrationType) bool
```
bsardo, line 15: Now that we have `ChannelEnabled` we should be able to delete `IntegrationEnabled` from the interface.

gdpr/aggregated_config.go
```diff
@@ -50,6 +51,13 @@ func (tc *tcf2Config) IntegrationEnabled(integrationType config.IntegrationType)
 	return tc.HostConfig.Enabled
```
bsardo, line 51: You should be able to delete this function `IntegrationEnabled` and let's rename `config/config.go#IntegrationEnabled` to `config/config.go#ChannelEnabled`.

gdpr/aggregated_config.go
```diff
@@ -50,6 +51,13 @@ func (tc *tcf2Config) IntegrationEnabled(integrationType config.IntegrationType)
...
 }
 
+func (tc *tcf2Config) ChannelEnabled(integrationType config.IntegrationType) bool {
```
bsardo, line 54: Please use the comment for the function `IntegrationEnabled` just above this new function and alter it accordingly.

config/accounts.go
```diff
@@ -45,13 +45,16 @@ type CookieSync struct {
 type AccountCCPA struct {
 	Enabled            *bool              `mapstructure:"enabled" json:"enabled,omitempty"`
 	IntegrationEnabled AccountIntegration `mapstructure:"integration_enabled" json:"integration_enabled"`
```
bsardo, line 47: It might be nice to provide some kind of deprecation warning if `integration_enabled` is found in a fetched account config but that could really bog down the logs. We should give some thought to whether we should bother to do something like that, and if so, how.

bsardo (commented):
I left a few comments here. I think the overarching theme is that all references to integration type should become channel type. That will set us up nicely so that the follow up PR that completely removes support for integration type from account configs just involves deleting a few lines of code.

config/accounts.go
bsardo: Discussed offline. We can't currently do this without taking a major performance hit. Throwing warnings with respect to JSON configs is something that can be introduced when we move from caching JSON to caching structs.

config/accounts.go
AlexBVolcy: Yes this is correct!

config/accounts.go
SyntaxNode: You can add a metric to give the host an idea of legacy account parameters are being used.

config/accounts_test.go
```diff
@@ -8,64 +8,64 @@ import (
...
+			wantEnabled:        nil,
 		},
 	}
 
 	for _, tt := range tests {
 		account := Account{
 			GDPR: AccountGDPR{
 				Enabled: tt.giveGDPREnabled,
-				IntegrationEnabled: AccountIntegration{
+				IntegrationEnabled: AccountChannel{
```
bsardo, line 62: This test case is still currently setting the `IntegrationEnabled` field when it should be setting the `ChannelEnabled` field. We also probably need a couple of additional test cases for when the deprecated `IntegrationEnabled` field is set but `ChannelEnabled` is not and also to verify precedence when both fields are set.
If you run code coverage you should see that we don't have full coverage of the function.

config/accounts_test.go
```diff
@@ -76,64 +76,64 @@ func TestAccountGDPREnabledForIntegrationType(t *testing.T) {
...
+			wantEnabled:        nil,
 		},
 	}
 
 	for _, tt := range tests {
 		account := Account{
 			CCPA: AccountCCPA{
 				Enabled: tt.giveCCPAEnabled,
-				IntegrationEnabled: AccountIntegration{
+				IntegrationEnabled: AccountChannel{
```
bsardo, line 130: Same comment as above regarding test coverage.

config/accounts.go
bsardo: @AlexBVolcy I like @SyntaxNode's idea of adding a metric. I suggest doing this in a separate PR if possible.

bsardo (approved):
LGTM

VeronikaSolovei9 (commented):
Just to confirm these 2 field are not related:
accounts.go -> `type Account struct` -> `DefaultIntegration`
and 
request.go -> `ExtRequestPrebid` -> `Integration`

AlexBVolcy:
@VeronikaSolovei9 Yes those two fields are not related!

PR #2407, opened 2022-10-06
pubmatic: pass alternatebiddercodes to its adserver

pm-nilesh-chate:
- PubMatic adapter: Pass alternatebiddercodes list to SSP
- PubMatic adapter: Bug fix, do not omit request.ext.prebid

### DIFF
--- a/adapters/pubmatic/pubmatic.go
+++ b/adapters/pubmatic/pubmatic.go
@@ -384,35 +384,11 @@ func getAlternateBidderCodesFromRequestExt(reqExt *openrtb_ext.ExtRequest) []str
 			if pmABC.AllowedBidderCodes == nil || (len(pmABC.AllowedBidderCodes) == 1 && pmABC.AllowedBidderCodes[0] == "*") {
 				return []string{"all"}
 			}
-			// keeping this off for now. Makes more sense to move this to PBS-Core
-			// return append(allowedBidders, assertBidderCodes(pmABC.AllowedBidderCodes)...)
 			return append(allowedBidders, pmABC.AllowedBidderCodes...)
 		}
 	}
 
 	return allowedBidders
-}
-
-// assertBidderCodes typical string assertions: remove duplicates, remove whitespaces, all lowercase case biddercode
-func assertBidderCodes(bidders []string) []string {
-	// nil not expected as it is a special case to allow all bidders
-	if bidders == nil {
-		return nil
-	}
-
-	checkedBidders := map[string]bool{"pubmatic": true}
-	newBiddersList := []string{}
-	for _, bidder := range bidders {
-		tmpBidder := strings.ToLower(bidder)
-		if _, ok := checkedBidders[tmpBidder]; !ok {
-			checkedBidders[tmpBidder] = true
-			tmpBidder = strings.TrimSpace(tmpBidder)
-			if len(tmpBidder) > 0 {
-				newBiddersList = append(newBiddersList, tmpBidder)
-			}
-		}
-	}
-	return newBiddersList
 }
 
 func addKeywordsToExt(keywords []*openrtb_ext.ExtImpPubmaticKeyVal, extMap map[string]interface{}) {
--- a/adapters/pubmatic/pubmatic.go
+++ b/adapters/pubmatic/pubmatic.go
@@ -366,14 +366,14 @@ func extractPubmaticExtFromRequest(request *openrtb2.BidRequest) (extRequestAdSe
 		pmReqExt.Acat = acat
 	}
 
-	if allowedBidders := getAlternateBidderCodesFromRequest(reqExt); allowedBidders != nil {
+	if allowedBidders := getAlternateBidderCodesFromRequestExt(reqExt); allowedBidders != nil {
 		pmReqExt.Marketplace = &marketplaceReqExt{AllowedBidders: allowedBidders}
 	}
 
 	return pmReqExt, nil
 }
 
-func getAlternateBidderCodesFromRequest(reqExt *openrtb_ext.ExtRequest) []string {
+func getAlternateBidderCodesFromRequestExt(reqExt *openrtb_ext.ExtRequest) []string {
 	if reqExt == nil || reqExt.Prebid.AlternateBidderCodes == nil {
 		return nil
 	}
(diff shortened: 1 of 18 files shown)


adapters/pubmatic/pubmatic.go
```diff
@@ -320,31 +323,96 @@ func parseImpressionObject(imp *openrtb2.Imp, extractWrapperExtFromImp, extractP
...
+	}
+
+	allowedBidders := []string{"pubmatic"}
+	if reqExt.Prebid.AlternateBidderCodes.Enabled {
+		if pmABC, ok := reqExt.Prebid.AlternateBidderCodes.Bidders["pubmatic"]; ok && pmABC.Enabled {
+			if pmABC.AllowedBidderCodes == nil || (len(pmABC.AllowedBidderCodes) == 1 && pmABC.AllowedBidderCodes[0] == "*") {
+				return []string{"all"}
+			}
+			// keeping this off for now. Makes more sense to move this to PBS-Core
+			// return append(allowedBidders, assertBidderCodes(pmABC.AllowedBidderCodes)...)
```
bsardo, line 388: Just curious, were you planning to propose such a change to PBS-Core? Perhaps we should delete this commented out code and the corresponding function if you're not going to use it.

adapters/pubmatic/pubmatic.go
```diff
@@ -320,31 +323,96 @@ func parseImpressionObject(imp *openrtb2.Imp, extractWrapperExtFromImp, extractP
...
+			return append(allowedBidders, pmABC.AllowedBidderCodes...)
+		}
+	}
+
+	return allowedBidders
+}
+
+// assertBidderCodes typical string assertions: remove duplicates, remove whitespaces, all lowercase case biddercode
+func assertBidderCodes(bidders []string) []string {
+	// nil not expected as it is a special case to allow all bidders
```
bsardo, line 398: Is this comment accurate? It seems as though `bidders` would not be nil when allowing all bidders. Wouldn't it have a single entry `"all"`?

adapters/pubmatic/pubmatic.go
```diff
@@ -320,31 +323,96 @@ func parseImpressionObject(imp *openrtb2.Imp, extractWrapperExtFromImp, extractP
...
+	if bidders == nil {
+		return nil
+	}
+
+	checkedBidders := map[string]bool{"pubmatic": true}
```
bsardo, line 403: Should `checkedBidders` just be an empty map? It appears that when you call this you would have already added `"pubmatic"` to the bidders array so it would just be handled here when you loop over the bidders.

adapters/pubmatic/pubmatic_test.go
```diff
@@ -359,3 +395,120 @@ func TestPubmaticAdapter_MakeBids(t *testing.T) {
...
+				bidders: []string{"Pubmatic", "Groupm", "groupm", " ", "", " appnexus "},
+			},
+			want: []string{"groupm", "appnexus"},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := assertBidderCodes(tt.args.bidders); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("assertBidderCodes() = %v, want %v", got, tt.want)
+			}
```
bsardo, line 511: I think it would be better to use the testify package instead of reflect. You could do the following:
```
result := assertBidderCodes(tt.args.bidders)
assert.ElementsMatch(t, result, tt.want, tt.name)
```

adapters/pubmatic/pubmatic_test.go
```diff
@@ -359,3 +395,120 @@ func TestPubmaticAdapter_MakeBids(t *testing.T) {
...
+				err := json.Unmarshal(tt.args.bidRequest.Ext, &reqExt)
+				if err != nil {
+					t.Errorf("getAlternateBidderCodesFromRequest() = %v", err)
+				}
+			}
+
+			got := getAlternateBidderCodesFromRequestExt(reqExt)
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("getAlternateBidderCodesFromRequest() = %v, want %v", got, tt.want)
+			}
```
bsardo, line 478: I suggest using testify instead of reflect:
```
assert.ElementsMatch(t, got, tt.want, tt.name)
```

adapters/pubmatic/pubmatic_test.go
```diff
@@ -158,23 +159,39 @@ func TestExtractPubmaticExtFromRequest(t *testing.T) {
 		request *openrtb2.BidRequest
```
bsardo, line 159: As a unit test, I think that this test could use one more test case that exercises the line that sets `pmReqExt.Marketplace`.

adapters/pubmatic/pubmatic.go
pm-nilesh-chate: Yes, had an initial plan to submit it as part of PBS-Core but feels like unnecessary validation. Removed the function from adapter code.

adapters/pubmatic/pubmatic.go
pm-nilesh-chate: removed the comment w.r.t above code

adapters/pubmatic/pubmatic.go
pm-nilesh-chate: removed the comment w.r.t above comments

adapters/pubmatic/pubmatic_test.go
pm-nilesh-chate: removed the comment w.r.t above comments

adapters/pubmatic/pubmatic_test.go
pm-nilesh-chate: done

adapters/pubmatic/pubmatic_test.go
pm-nilesh-chate: added test case for **Valid Pubmatic marketplace ext**

pm-nilesh-chate:
Hi @bsardo, Thanks for the comments. I've addressed them. Could you please review the same.

adapters/pubmatic/pubmatic.go
```diff
@@ -320,31 +323,72 @@ func parseImpressionObject(imp *openrtb2.Imp, extractWrapperExtFromImp, extractP
...
+	if err != nil {
+		return pmReqExt, fmt.Errorf("error decoding Request.ext : %s", err.Error())
+	}
+	pmReqExt.ExtRequest = *reqExt
+
 	reqExtBidderParams, err := adapters.ExtractReqExtBidderParamsMap(request)
 	if err != nil {
-		return nil, acat, err
+		return pmReqExt, err
 	}
```
bsardo, line 345: I suggest removing the todo comment on line 334 and handling the repeated request ext unmarshal by replacing this block of code with something like the following:
```
	reqExtBidderParams := make(map[string]json.RawMessage)
	if reqExt.Prebid.BidderParams != nil {
		err = json.Unmarshal(reqExt.Prebid.BidderParams, &reqExtBidderParams)
		if err != nil {
			return pmReqExt, err
		}
	}
```

bsardo (commented):
Looks good, just an optimization suggestion

adapters/pubmatic/pubmatic.go
pm-nilesh-chate: done

bsardo (approved):

VeronikaSolovei9 (approved):

PR #2408, opened 2022-10-06
AndBeyond.Media adapter: rename bidder

AndBeyondMediaHB:


static/bidder-info/beyondmedia.yaml
```diff
@@ -1,4 +1,3 @@
-endpoint: "http://backend.andbeyond.media/pserver"
```
SyntaxNode, line 1: The endpoint is required, unless if you want to change the adapter to be disabled by default.

SyntaxNode (commented):
Please also update `GetDisabledBiddersErrorMessages` in `adapter_util.go` to include your original. bidder name, otherwise publishers currently using your older adapter name will receive errors.

You may also wish to ad "andbeyondmedia" as an alias of your new name to allow for a transition period.

AndBeyondMediaHB:
> adapter name

no one is integrated with the old adapter name at the moment

AndBeyondMediaHB:
> Please also update `GetDisabledBiddersErrorMessages` in `adapter_util.go` to include your original. bidder name, otherwise publishers currently using your older adapter name will receive errors.
> 
> You may also wish to ad "andbeyondmedia" as an alias of your new name to allow for a transition period.

no one is integrated with the old adapter name at the moment. We are not interested in the old name.

static/bidder-info/beyondmedia.yaml
AndBeyondMediaHB: @SyntaxNode Thanks, updated

bsardo (commented):
I see references to the `andbeyond` subdomain in `static/bidder-info/beyondmedia.yaml` and various test files in your adapter directory. Is that subdomain accurate? I'm less concerned about the test files, though I think you should probably update them to fake uris. especially if the subdomain is no longer valid as it can be misleading.

Also you have a couple of references to `AndBeyond` in `static/bidder-params/beyondmedia.json`.

And to confirm, no one has ever integrated with your adapter yet? I just want to make sure we don't need to add AndBeyond to the disabled bidders error list.

Thanks

SyntaxNode:
> no one is integrated with the old adapter name at the moment. We are not interested in the old name.

That's fine. There's no need to use an alias to keep it active. I do ask that you make the change in `adapter_util.go` to avoid breaking any edge case publishers. This change is not an alias. It just gracefully provides an error indicating the name change. We ask the same for all adapters and you'll see a few entries already.

AndBeyondMediaHB:
> > no one is integrated with the old adapter name at the moment. We are not interested in the old name.
> 
> That's fine. There's no need to use an alias to keep it active. I do ask that you make the change in `adapter_util.go` to avoid breaking any edge case publishers. This change is not an alias. It just gracefully provides an error indicating the name change. We ask the same for all adapters and you'll see a few entries already.

@SyntaxNode done

AndBeyondMediaHB:
> I see references to the `andbeyond` subdomain in `static/bidder-info/beyondmedia.yaml` and various test files in your adapter directory. Is that subdomain accurate? I'm less concerned about the test files, though I think you should probably update them to fake uris. especially if the subdomain is no longer valid as it can be misleading.
> 
> Also you have a couple of references to `AndBeyond` in `static/bidder-params/beyondmedia.json`.
> 
> And to confirm, no one has ever integrated with your adapter yet? I just want to make sure we don't need to add AndBeyond to the disabled bidders error list.
> 
> Thanks

@bsardo endpoint `http://backend.andbeyond.media/pserver` with domain `backend.andbeyond.media` is valid and **AndBeyond.Media** is the actual name. We also updated `disabledBidders`

bsardo (approved):
LGTM

SyntaxNode (approved):

SyntaxNode:
@AndBeyondMediaHB Looks good to us. There is a merge conflict. Please resolve the conflict and we'll merge it in.

SyntaxNode (approved):

bsardo:
@AndBeyondMediaHB this looks good. Could you please open a docs PR and then we can merge. Thanks!

AndBeyondMediaHB:
@bsardo I don't quite understand you. Do you mean [this](https://github.com/prebid/prebid.github.io/pull/4018/files) pull request?

bsardo:
@AndBeyondMediaHB my mistake. I see that the docs are accurate, particularly the `biddercode` field is correct.

PR #2416, opened 2022-10-11
New Adapter: Bidstack

arthurlataks:
Related docs PR: https://github.com/prebid/prebid.github.io/pull/4071


AlexBVolcy:
There is a PR https://github.com/prebid/prebid-server/pull/2370 that is being merged today that will cause conflicts to new adapter PRs. 

To resolve these conflicts, you'll need to do the following:

In `bidstack.go`, update the `Builder()` function signature to include a new parameter `server config.Server`

In `bidstack_test.go`, update any test that includes a call to `Builder()` to pass a `config.Server` object, to that call (i.e. `config.Server{ExternalUrl: "http://hosturl.com", GvlID: 1, DataCenter: "datacenter"}`)

After the merge, you can look at other adapters to see how the server object is implemented as an example.

adapters/bidstack/bidstack.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return bidResponse, nil
+}
+
+func prepareHeaders(request *openrtb2.BidRequest) (headers http.Header, err error) {
+	bidderParams, err := adapters.ExtractReqExtBidderParamsMap(request)
+	if err != nil {
+		return nil, fmt.Errorf("extract bidder params: %v", err)
+	}
+	publisherID := strings.ReplaceAll(string(bidderParams["publisherId"]), "\"", "")
```
VeronikaSolovei9, line 113: Lines 109-113: consider to refactor this to use imp.ext, not byte array. It will make code much simpler. You will need to add `openrtb_ext/imp_{bidder}.go`. Please follow the instruction in dev docs: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html
It also looks like you only need to execute it once, make sure it's not in a loop over imp array.

adapters/bidstack/bidstack_test.go
```diff
@@ -0,0 +1,71 @@
...
+	})
+
+	t.Run("malformed extensions", func(t *testing.T) {
+		expected := (http.Header)(nil)
+		actual, err := prepareHeaders(&openrtb2.BidRequest{Ext: []byte(`{`)})
+
+		assert.EqualError(t, err, "extract bidder params: error decoding Request.ext : unexpected end of JSON input")
+		assert.Equal(t, expected, actual)
+	})
+}
```
VeronikaSolovei9, line 71: With imp_{bidder}.go part of these test cases will not be needed. Imp.ext will be validated by Prebid Server core code based on `bidder-params/bidstack.json`

VeronikaSolovei9 (commented):
Thank you for submitting the PR. Please address comments.

adapters/bidstack/bidstack.go
arthurlataks: i was considering this during the implementation but decided to use the root.ext of the bid request.
Since, as you noticed, I need only one instance of the key, there is no need to pass it with each impression.

adapters/bidstack/bidstack.go
```diff
@@ -0,0 +1,123 @@
...
+	headers, err := prepareHeaders(request)
+	if err != nil {
+		return nil, []error{fmt.Errorf("headers prepare: %v", err)}
+	}
+
+	for _, imp := range request.Imp {
+		if imp.BidFloor > 0 && imp.BidFloorCur != "" && strings.ToUpper(imp.BidFloorCur) != currencyUSDISO4217 {
+			convertedValue, err := reqInfo.ConvertCurrency(imp.BidFloor, imp.BidFloorCur, currencyUSDISO4217)
+			if err != nil {
+				return nil, []error{fmt.Errorf("currency convert: %v", err)}
```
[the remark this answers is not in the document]
arthurlataks: This sample is taken from the documentation, and I believe this is a better way rather than skipping an impression because otherwise, this error case will become silent and untestable.

arthurlataks:
@VeronikaSolovei9 Hello, thank you for the review, it was very helpful. I have addressed all the comments.

adapters/bidstack/bidstack.go
arthurlataks: Updated the code in favour of using `openrtb_ext.Imp` while extracting the publisher ID from the bid request.

adapters/bidstack/bidstack_test.go
arthurlataks: Updated the code to remove redundant test cases.

adapters/bidstack/bidstack_test.go
```diff
@@ -0,0 +1,50 @@
...
+	bidder, buildErr := Builder(openrtb_ext.BidderBidstack, config.Adapter{Endpoint: "http://mock-adserver.url"}, config.Server{})
+
+	if buildErr != nil {
+		t.Fatalf("Builder returned unexpected error %v", buildErr)
+	}
+
+	adapterstest.RunJSONBidderTest(t, "bidstacktest", bidder)
+}
+
+func Test_prepareHeaders(t *testing.T) {
```
VeronikaSolovei9, line 25: Nitpick: please follow a camelcase naming in all test functions.  Test_prepareHeaders  ->  TestPrepareHeaders

VeronikaSolovei9 (commented):
Thank you for addressing comments. Added one minor nitpick about unit tests names

arthurlataks:
@VeronikaSolovei9 Hello, minor nitpick resolved

hhhjort (approved):
LGTM

VeronikaSolovei9 (approved):

Issue #2300, opened 2022-07-13
Parsing oRTB.site.domain when no http referer is provided


vfedoseev:
Hi,

While checking the Prebid server integrations on several sites, we've noticed an issue with 'no-referrer' meta tag.

The /auction endpoint is currently checking the following values in bid request:
- bidReq.Site
- bidReq.Site.Page
- bidReq.Site.Domain

If any of them is not presented, then the Http.Referer header is parsed and the results are passed to the 'Site' object:
https://github.com/prebid/prebid-server/blob/master/endpoints/openrtb2/auction.go#L1525

Sometimes the publishers use the following meta tag on the sites - it prevents browsers from sending Http.Referer to any requests:
`<meta name="referrer" content="no-referrer">`

If the 'site.domain' is also missing in the bid request payload, then it won't be parsed and passed to the adapters.
Example request below:
```
{
    "id": "ddd21d93-6c5d-458a-9cc2-0452330702e6",
    "source": {
        "tid": "ddd21d93-6c5d-458a-9cc2-0452330702e6"
    },
    "tmax": 3000,
    "imp": [
        ....
    ],
    "ext": {
        "prebid": {
            "auctiontimestamp": 1657713316889,
            "targeting": {
                "includewinners": true,
                "includebidderkeys": false
            },
            "channel": {
                "name": "pbjs",
                "version": "v5.20.4"
            }
        }
    },
    "cur": [
        "EUR"
    ],
    "site": {
        "publisher": {
            "id": "1"
        },
        "page": "https://test.somepage.com"
    }
}
```

So, the final request to /auction enpoint:
- does non contain Http.Referer and Site.Domain
- contains Site.Page

We suggest to parse the Site.Page - if Http.Referer is empty and Site.Page is available - to extract the Site.Domain value.

bretg:
PBS-Java sets site.domain and site.publisher.domain:

site.domain: full domain. e.g. www.example.com. or sports.usatoday.com
site.publisher.domain: 'rounded off' site.domain. e.g. example.com or usatoday.com

We'll review how these values are set and make sure it's in sync with Go

bsardo:
**PBS-Go current logic:**
```
if site.page is not set {
	if http.referer is set and http.referer is a valid URL {
		site.page = http.referer
	}
}
if site.domain is not set {
	if http.referer is set and http.referer is a valid URL {
		site.domain = http.referer.host   (i.e. the example.com portion of http://cool.example.com)
	}
}
```
[...]

PR #2422, opened 2022-10-13
Set Site.Publisher.Domain and setSiteImplicitly() Refactoring

AlexBVolcy:
This PR addresses the final piece of this issue https://github.com/prebid/prebid-server/issues/2300

The main portion of this issue was already addressed here https://github.com/prebid/prebid-server/pull/2301

The goal of this PR is to set `site.publisher.domain` if `site.domain` is set

The value that we set `site.publisher.domain` to is extracted from the highest level domain in http referrer value

To accomplish this, only one function is needed to be updated in `openrtb2/auction.go,` `setSiteImplicitly()` 

`setSiteImplicitly()` was also refactored to be simplified to have various nested if conditions extracted into their own functions

Tests were updated to accompany the changes

Note: I had to have gofmt updates in order for my tests to run in Github, so 13 files comments were updated slightly for spacing. The only files that were updated in relation to set `site.publisher.domain` were `auction.go` and `auction_test.go`


endpoints/openrtb2/auction.go
```diff
@@ -1565,8 +1565,15 @@ func setSiteImplicitly(httpReq *http.Request, r *openrtb_ext.RequestWrapper) {
 				if r.Site == nil {
 					r.Site = &openrtb2.Site{}
 				}
+				if r.Site.Publisher == nil {
+					r.Site.Publisher = &openrtb2.Publisher{}
+				}
+
 				if r.Site.Domain == "" {
 					r.Site.Domain = domain
+					r.Site.Publisher.Domain = domain
```
hhhjort, line 1574: At this point, is it not possible that r.Site.Publisher.Domain is populated, and getting overwritten here? If for example, r.Site.Domain is "", but r.Site.Publisher.Domain has a value, you should be able to reach this line and overwrite the existing value.

mansinahar (approved):

hhhjort (approved):
LGTM

PR #2423, opened 2022-10-14
Request Wrapper Usage In HoldAuction - Phase 1

SyntaxNode:
Pushes usage of the Request Wrapper in `HoldAuction` down to right before the `cleanOpenRTBRequests` call, which is the bit which breaks down the main request into bidder specific requests.


exchange/exchange.go
```diff
@@ -335,9 +343,9 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 		var bidCategory map[string]string
 		//If includebrandcategory is present in ext then CE feature is on.
-		if requestExt.Prebid.Targeting != nil && requestExt.Prebid.Targeting.IncludeBrandCategory != nil {
+		if requestExtLegacy.Prebid.Targeting != nil && requestExtLegacy.Prebid.Targeting.IncludeBrandCategory != nil {
 			var rejections []string
-			bidCategory, adapterBids, rejections, err = applyCategoryMapping(ctx, requestExt, adapterBids, e.categoriesFetcher, targData, &randomDeduplicateBidBooleanGenerator{})
+			bidCategory, adapterBids, rejections, err = applyCategoryMapping(ctx, requestExtLegacy, adapterBids, e.categoriesFetcher, targData, &randomDeduplicateBidBooleanGenerator{})
```
hhhjort, line 348: From my scan, this is the only `requestExtLegacy` that is not followed by a `.Prebid`. Looking into `applyCategoryMapping()` it only makes use of the `Prebid` component of `requestExtLegacy` I think we can simplify this down to just extracting `requestExtPrebid` from `requestExt` and getting rid of `extractBidRequestExt()` and the extra unmarshal that comes with it.

exchange/exchangetest/aliases.json
```diff
@@ -64,16 +64,7 @@
                 }
               }
             }
-          ],
```
hhhjort, line 67: Simplifying tests by not hardcoding in a server object to the test exchange?

exchange/exchange.go
SyntaxNode: Did you check if `requestExtLegacy.Prebid` usage is purely read-only? I didn't yet check the usage below this line. Currently, `requestExtLegacy ` is a separate copy from the request where a write would be ignored. But if there is no write, then yup - we're safe to make this change.

exchange/exchangetest/aliases.json
SyntaxNode: Exactly. The server object tests are now isolated to their own cases. Values default to empty and can be specified in the test case json file for those which need it. IMHO the number of tests affected by changes to this code is too much.

exchange/exchange.go
hhhjort: Doesn't matter. `GetPrebid()` returns a copy of the `Prebid` object. Thus `SetPrebid(Prebid)` is required to make the change active, and allows wrapper to set the dirty bit.

exchange/exchange.go
```diff
@@ -214,32 +214,42 @@ type BidderRequest struct {
...
+
 	if !e.server.Empty() {
-		requestExt.Prebid.Server = &openrtb_ext.ExtRequestPrebidServer{ExternalUrl: e.server.ExternalUrl, GvlID: e.server.GvlID, DataCenter: e.server.DataCenter}
+		if requestExtPrebid == nil {
+			requestExtPrebid = &openrtb_ext.ExtRequestPrebid{}
+		}
+		requestExtPrebid.Server = &openrtb_ext.ExtRequestPrebidServer{
+			ExternalUrl: e.server.ExternalUrl,
+			GvlID:       e.server.GvlID,
+			DataCenter:  e.server.DataCenter}
```
guscarreon, line 235: I realize this is probably outside the scope of this PR but, could `ExtRequestPrebidServer` be used?
```
57   type exchange struct {
58       adapterMap               map[openrtb_ext.BidderName]AdaptedBidder
59       bidderInfo               config.BidderInfos
60       bidderToSyncerKey        map[string]string
61       me                       metrics.MetricsEngine
62       cache                    prebid_cache_client.Client
63       cacheTime                time.Duration
64       gdprPermsBuilder         gdpr.PermissionsBuilder
65       tcf2ConfigBuilder        gdpr.TCF2ConfigBuilder
66       currencyConverter        *currency.RateConverter
67       externalURL              string
68       gdprDefaultValue         gdpr.Signal
69       privacyConfig            config.Privacy
70       categoriesFetcher        stored_requests.CategoryFetcher
71       bidIDGenerator           BidIDGenerator
72       hostSChainNode           *openrtb2.SupplyChainNode
73       adsCertSigner            adscert.Signer
74 -     server                   config.Server
   +     server                   *openrtb_ext.ExtRequestPrebidServer
75       bidValidationEnforcement config.Validations
76       requestSplitter          requestSplitter
77   }
exchange/exchange.go
```
so we could maybe simplify to:
```
228     if !e.server.Empty() {
229         if requestExtPrebid == nil {
230             requestExtPrebid = &openrtb_ext.ExtRequestPrebid{}
231         }
232 -       requestExtPrebid.Server = &openrtb_ext.ExtRequestPrebidServer{
233 -           ExternalUrl: e.server.ExternalUrl,
234 -           GvlID:       e.server.GvlID,
235 -           DataCenter:  e.server.DataCenter}
    +       requestExtPrebid.Server = *e.server
236         requestExt.SetPrebid(requestExtPrebid)
237     }
exchange/exchange.go
```

SyntaxNode:
Per feedback, updated the rest of the HoldAuction to avoid an unmarshal call. There is an assumption that the splitter does not modify the prebid ext.

exchange/exchange.go
SyntaxNode: Please be careful here. It's a shallow copy. Deep pointers can still cause trouble.

exchange/exchangetest/server-info-without-ext.json
```diff
@@ -0,0 +1,68 @@
+{
+    "server": {
```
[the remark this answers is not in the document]
SyntaxNode: I will change the 2-space back to 4-space. VS Code's reformatter wasn't working as expected. Found and fixed.

exchange/exchangetest/server-info-with-ext.json
```diff
@@ -0,0 +1,74 @@
+{
+    "server": {
```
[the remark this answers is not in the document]
SyntaxNode: I will change the 2-space back to 4-space. VS Code's reformatter wasn't working as expected. Found and fixed.

exchange/exchange.go
```diff
@@ -214,32 +214,42 @@ type BidderRequest struct {
...
+		requestExt.SetPrebid(requestExtPrebid)
 	}
 
-	cacheInstructions := getExtCacheInstructions(requestExt)
-	targData := getExtTargetData(requestExt, &cacheInstructions)
+	cacheInstructions := getExtCacheInstructions(requestExtPrebid)
+
+	targData := getExtTargetData(requestExtPrebid, cacheInstructions)
```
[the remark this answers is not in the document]
SyntaxNode: I think the better question is why were we passing it by reference before? That actually introduced the chance of a nil reference panic. i thought it better to keep it as a value.

exchange/exchange.go
SyntaxNode: I'm not sure. I prefer to leave that out for now. I'm not yet convinced `config.Server` should be replaced wholesale.

exchange/events.go
```diff
@@ -31,12 +31,19 @@ func getEventTracking(requestExtPrebid *openrtb_ext.ExtRequestPrebid, ts time.Ti
...
 		externalURL:        externalURL,
 	}
 }
 
+func getIntegrationType(requestExtPrebid *openrtb_ext.ExtRequestPrebid) string {
+	if requestExtPrebid != nil {
+		return requestExtPrebid.Integration
+	}
+	return ""
+}
```
SyntaxNode, line 45: Added nil safety check. Error condition exposed during testing.

exchange/exchange.go
```diff
@@ -428,15 +439,36 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
+			multiBidMap[multiBid.Bidder] = *multiBid
+		} else {
+			for _, bidder := range multiBid.Bidders {
+				multiBidMap[bidder] = *multiBid
+			}
+		}
+	}
+
+	return multiBidMap
+}
```
SyntaxNode, line 460: Refactor from removing `extractBidRequestExt`. Usage is contained in this conditional block, so there is no need to reference it in the ext object.

exchange/exchange.go
```diff
@@ -261,28 +273,26 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
...
 
-	recordImpMetrics(r.BidRequestWrapper.BidRequest, e.me)
+	recordImpMetrics(r.BidRequestWrapper, e.me)
 
 	// Make our best guess if GDPR applies
-	gdprDefaultValue := e.parseGDPRDefaultValue(r.BidRequestWrapper.BidRequest)
+	gdprDefaultValue := e.parseGDPRDefaultValue(r.BidRequestWrapper)
+
+	// rebuild/resync the request in the request wrapper.
+	if err := r.BidRequestWrapper.RebuildRequest(); err != nil {
```
guscarreon, line 286: A non-dirty `RequestWrapper` will rapidly exit `request.RebuildRequest()`, but still do at least two levels of stack allocations because of function calls. Because no `BidRequest` fields seem to be modified between line 252 and here, we could slightly optimize by doing a `responseDebugAllow` check instead of an extra `request.RebuildRequest()` call.
```
283     gdprDefaultValue := e.parseGDPRDefaultValue(r.BidRequestWrapper)
284
    +   if !responseDebugAllow {
285         // rebuild/resync the request in the request wrapper.
286         if err := r.BidRequestWrapper.RebuildRequest(); err != nil {
287             return nil, err
288         }
    +   }
296
297     e.me.RecordRequestPrivacy(privacyLabels)
exchange/exchange.go
```
Unless I'm missing something, these are the outcomes I could identify:

**As written**
| ExecuteProcessedAuctionStage calls `RebuildRequest()` | `responseDebugAllow` | times `responseDebugAllow()` gets called |
| ---------------------------- | ------------------ | ---------------------- |
|        no                    |     false          | 1 time. Line 286 |
|        no                    |     true           | 2 times. Lines 252 and 286 |
|       yes                    |     false          | 2 times. Lines 217 and 286 |
|       yes                    |     true           | 3 times. Lines 217, 252 and 286 |

**Adding the `!responseDebugAllow` check**
| ExecuteProcessedAuctionStage calls `RebuildRequest()` | `responseDebugAllow` | times `responseDebugAllow()` gets called |
| ---------------------------- | ------------------ | ---------------------- |
|        no                    |     false          | 1 time on line 286  |
|        no                    |     true           | 1 time. Line 252 |
|       yes                    |     false          | 2 times. Lines 217 and 286 |
|       yes                    |     true           | 2 times. Lines 217 and 252 |


On the other hand, I admit the second `responseDebugAllow` does look a little verbose. What do you think?

guscarreon (commented):
Looks great. Makes things look cleaner. Left a couple of nitpicks to consider.

hooks/hookexecution/executor.go
```diff
@@ -165,6 +169,11 @@ func (e *hookExecutor) ExecuteProcessedAuctionStage(request *openrtb2.BidRequest
 	e.saveModuleContexts(contexts)
 	e.pushStageOutcome(outcome)
 
+	// remove type information if there is no rejection
+	if reject == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
SyntaxNode: I prefer to leave this in because returning a `nil` reject is not the same as returning `nil`. `return reject` contains type information which caused problems during tests. I don't think it would cause problems during normal runtime. I added a comment above to try to explain this situation, but based on your comment that's not a good comment. Perhaps you can recommend a different wording?

exchange/exchange.go
SyntaxNode: Thank you for the review of this potential optimization. I appreciate the level of detail you captured. 

As you noted, this optimization relies on there being no modification to the request wrapper after line 252. What if that changes in the future? Do you perhaps have an idea for how to codify this assumption somehow? .. such as with a unit test? Or is this really not a practical concern?

I don't really know why the request wrapper is rebuilt at that specific spot. I wonder if that may be moved entirely.. but now I think I'm getting too complicated. :)

guscarreon: Agree. I guess the `ReqestWrapper` is a work in progress and in phases to follow we can decide where should we trigger the rebuilds.

hooks/hookexecution/executor.go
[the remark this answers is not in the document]
guscarreon: > `return reject` contains type information which caused problems during tests.

Wow 😮 . I guess I need to get more familiar with the modules package. No, your comment is fine, it totally went over me, sorry.

guscarreon (approved):
LGTM

exchange/exchange.go
SyntaxNode: > I guess the ReqestWrapper is a work in progress

Yes. Absolutely. @hhhjort is working on a follow-up PR to continue that work.

PR #2425, opened 2022-10-17
Smaato: Add Native

el-chuck:
Add native to Smaato Adapter

### DIFF
--- a/adapters/smaato/smaato.go
+++ b/adapters/smaato/smaato.go
@@ -17,7 +17,7 @@ import (
 	"github.com/prebid/prebid-server/util/timeutil"
 )
 
-const clientVersion = "prebid_server_0.4"
+const clientVersion = "prebid_server_0.5"
 
 type adMarkupType string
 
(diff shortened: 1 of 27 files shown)


adapters/smaato/smaato.go
```diff
@@ -187,20 +188,29 @@ func (adapter *adapter) makeIndividualRequests(request *openrtb2.BidRequest) ([]
...
 
 func splitImpressionsByMediaType(imp *openrtb2.Imp) ([]openrtb2.Imp, error) {
-	if imp.Banner == nil && imp.Video == nil {
-		return nil, &errortypes.BadInput{Message: "Invalid MediaType. Smaato only supports Banner and Video."}
+	if imp.Banner == nil && imp.Video == nil && imp.Native == nil {
+		return nil, &errortypes.BadInput{Message: "Invalid MediaType. Smaato only supports Banner, Video and Native."}
 	}
 
-	imps := make([]openrtb2.Imp, 0, 2)
+	imps := make([]openrtb2.Imp, 0, 3)
```
hhhjort, line 195: I am pondering if this should be `make([]openrtb2.Imp, 0 ,3*len(imp))`, as the maximum number of imps you could see would be the orginal number of imps times 3, which would be the case if each imp was multiformat with all three allowed formats. Or more likely just 2, as I don't expect you would see a multiformat imp with all three of banner, video, and native very often, if at all.

hhhjort (approved):
LGTM

guscarreon (approved):
LGTM

PR #2428, opened 2022-10-18
Update Go version to 1.19

4lexvav:


.devcontainer/devcontainer.json
```diff
@@ -6,7 +6,7 @@
 		"dockerfile": "Dockerfile",
 		"args": {
 			// Update the VARIANT arg to pick a version of Go
-			"VARIANT": "1.16",
+			"VARIANT": "1.19",
```
guscarreon, line 9: @AlexBVolcy I don't use VS Code so I can't assert this line to be correct

guscarreon (commented):
Couldn't verify `.devcontainer/devcontainer.json` because I'm not familiar with the VS Code IDE. Otherwise, the change looks good to me with the exception of the comment found in `endpoints/cookie_sync_test.go` shown below: 
```
1829 // ErrReader returns an io.Reader that returns 0, err from all Read calls. This is added in
1830 // Go 1.16. Copied here for now until we switch over.
1831 func ErrReader(err error) io.Reader {
1832     return &errReader{err: err}
1833 }
endpoints/cookie_sync_test.go
```
Can we modify or, if needed, remove?

AlexBVolcy (commented):
This looks good. I'm just waiting on @SyntaxNode to verify .devcontainer/devcontainer.json before approval

4lexvav:
@guscarreon thanks for bringing my attention to the comment in file `endpoints/cookie_sync_test.go`, I removed this code and changed its usage to `ErrReader` from the [standard library](https://pkg.go.dev/testing/iotest#ErrReader).

SyntaxNode:
> This looks good. I'm just waiting on @SyntaxNode to verify .devcontainer/devcontainer.json before approval

The tests failed. Turns out, Go 1.18 added a new feature where the binary includes information about the git repository. This breaks in the dev container since git isn't really available. I've implemented (and tested) a fix along with updates to now-outdated config.

Please update to this file:

```
// For format details, see https://aka.ms/vscode-remote/devcontainer.json or this file's README at:
// https://github.com/microsoft/vscode-dev-containers/tree/v0.112.0/containers/go
{
	"name": "Go",
	"build": {
		"dockerfile": "Dockerfile",
		"args": {
			// Update the VARIANT arg to pick a version of Go
			"VARIANT": "1.19",
			// Options
			"INSTALL_NODE": "false",
			"NODE_VERSION": "lts/*"
		}
	},
	"containerEnv": {
		"GOPRIVATE": "${localEnv:GOPRIVATE}",
		"GOFLAGS": "-buildvcs=false",
		"PBS_GDPR_DEFAULT_VALUE": "0"
	},
	"runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],

	// Set *default* container specific settings.json values on container create.
	"settings": {
		"terminal.integrated.profiles.linux": {
			"bash": {
				"path": "/usr/bin/bash"
			}
		},
		"terminal.integrated.defaultProfile.linux": "bash",
		"go.toolsManagement.checkForUpdates": "off",
		"go.gopath": "/go",
	},

	// Add the IDs of extensions you want installed when the container is created.
	"extensions": [
		"golang.Go",
		"ms-azuretools.vscode-docker",
		"redhat.vscode-xml",
		"redhat.vscode-yaml",
		"eamodio.gitlens",
	],

	// Use 'forwardPorts' to make a list of ports inside the container available locally.
	"forwardPorts": [8000,8001,6060],

	// Use 'postCreateCommand' to run commands after the container is created.
	"postCreateCommand": "mkdir ~/.ssh; ssh-keyscan github.com > ~/.ssh/known_hosts",

	// Uncomment to connect as a non-root user. See https://aka.ms/vscode-remote/containers/non-root.
	"remoteUser": "vscode"
}
```

AlexBVolcy:
Waiting for update based on @SyntaxNode comment before approval.

SyntaxNode (approved):

guscarreon (approved):
Thank you for addressing our feedback @4lexvav . Changes look good to me

AlexBVolcy (approved):

PR #2433, opened 2022-10-25
ExtRequestPrebidCache: Unnecessary UnmarshalJSON

SyntaxNode:
I moved the logic in this method to validation code in the ImpWrapper PR, but left the unmarshal code because I wasn't sure of the bit about infinite loops. I understand now. This code does nothing and can be removed.


AlexBVolcy (approved):

guscarreon (approved):
LGTM

PR #2436, opened 2022-10-26
Triplelift: adding native endpoint

patrickloughrey:
Adding Triplelift's native endpoint to our config

cc: @nllerandi3lift

### DIFF
--- a/static/bidder-info/triplelift_native.yaml
+++ b/static/bidder-info/triplelift_native.yaml
@@ -1,4 +1,4 @@
-endpoint: "https://tlx.3lift.com/s2sn/auction?&supplier_id=20"
+endpoint: "https://tlx.3lift.com/s2sn/auction?supplier_id=20"
 disabled: true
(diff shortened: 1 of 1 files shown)


static/bidder-info/triplelift_native.yaml
```diff
@@ -1,3 +1,4 @@
+endpoint: "https://tlx.3lift.com/s2sn/auction?supplier_id=20"
 disabled: true
```
VeronikaSolovei9, line 2: Just to confirm: do you want to enable this adapter as well?

nllerandi3lift: No. We just want to make the endpoint available should any publisher/PBS host want to leverage it. This is not our main adapter so it is seldom used. We're just looking to save an email asking us what the endpoint should be should anyone want to turn this on.

VeronikaSolovei9 (approved):

guscarreon (approved):
LGTM

Issue #2429, opened 2022-10-21
Optimising the account service in go implementation - Why are we always performing patching and merging operation per account request


ShriprasadM:
I can see we can optimise the patch and merge operation done in account service by doing it one time and storing it in the cache. This is already discussed with @mansinahar . We have done the changes and about to raise the PR

Flow what I have understood :
1. fetcher.FetchAccount - returns accountJSON either from backend persistent storage or from cache
2. in Else part : It merges AccountsDefaultsJSON which creates completeJSON
3. setDeriveConfig  - Do something around GDPR
4. Finally we return account object5. 

**Question**: Why are we always performing patching and merging operation of account per request? Shouldn't it be part of cache , where fetcher push the accountJSON  ?

Instead of current Implementation can we have following? I feel it will avoid unnecessary repetitive operation of patch and merge per request

Possibly  the  changes will be expected here
https://github.com/prebid/prebid-server/blob/bdc3190c741bd858b868be21d51a08617e114326/stored_requests/fetcher.go#L213

**Proposed Change:**
>accountJSON := cache.get(account) 
 if   accountJSON == nil {
     accountJSON = fetcher.getAccount() // from backend persistent storage
     /* Have merge and patch logic here*/
     completeJSON = jsonpatch.MergePatch(cfg.AccountDefaultsJSON(), accountJSON)
     accountJSON = completeJSON
     cache.save(accountJSON, ttl)
} 
return accountJSON

SyntaxNode:
We refer to these parts of code as "fetchers". We currently have fetchers for:
- Stored Requests
- Stored Requests - AMP Specific
- Stored Requests.- Video Specific
- Stored Responses
- Account Config
- Category Taxonomy Translations
- TCF GVL (different implementation, same concept)

The fetcher code is 5+ years old and could very much benefit from some attention. A complete refactor is on our radar, but there's a lot going on right now and we haven't been able to devote our time. If you're interested, I'll leave some of our initial design thoughts down below. Not sure when we'll get it, but community contributors are welcomed to work with us (as always) to move it along.

### Q&A

> Why are we always performing patching and merging operation of account per request?

The cache layer currently works on the json data, not the parsed and resolved config object. A more efficient path would be to cache the resolved object. I didn't look into the details of your request, but if you're 100% sure there are no side affects (proven with excellent test coverage) then sure.

> Shouldn't it be part of cache , where fetcher push the accountJSON ?

Yes.

[...]

PR #2440, opened 2022-10-28
Account Service Optimisation

Pubmatic-Dhruv-Sonone:
Related Issue: https://github.com/prebid/prebid-server/issues/2429


bsardo:
@Pubmatic-Dhruv-Sonone it looks like you'll need to merge with master to resolve the format issues you're seeing during the go 1.19 validation step.

account/account.go
```diff
@@ -47,39 +48,6 @@ func GetAccount(ctx context.Context, cfg *config.Configuration, fetcher stored_r
...
-				// unmarshal again to check if unmarshal error still exists after GDPR field conversion
-				err = json.Unmarshal(completeJSON, account)
-
-				if _, ok := err.(*json.UnmarshalTypeError); ok {
-					return nil, []error{&errortypes.MalformedAcct{
-						Message: fmt.Sprintf("The prebid-server account config for account id \"%s\" is malformed. Please reach out to the prebid server host.", accountID),
-					}}
-				}
-			}
-		}
```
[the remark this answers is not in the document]
Pubmatic-Dhruv-Sonone: Missed this during master merge. Added this logic.

Pubmatic-Dhruv-Sonone:
@bsardo  Can you please review the PR? I have addressed above comments.

stored_requests/fetcher.go
```diff
@@ -28,7 +29,7 @@ type Fetcher interface {
 
 type AccountFetcher interface {
 	// FetchAccount fetches the host account configuration for a publisher
-	FetchAccount(ctx context.Context, accountID string) (json.RawMessage, []error)
+	FetchAccount(ctx context.Context, accountDefaultJSON json.RawMessage, accountID string) (*config.Account, []error)
```
mansinahar, line 32: I think we should leave the return type for `FetchAccount` to be a raw JSON string as our cache is currently configured to store stuff as JSON strings rather than objects. This is something we want ideally but IMHO it doesn't make sense until we change the cache code to accept objects as well.

stored_requests/backends/http_fetcher/fetcher.go
```diff
@@ -163,7 +167,39 @@ func (fetcher *HttpFetcher) FetchAccount(ctx context.Context, accountID string)
 			DataType: "Account",
 		}}
 	}
-	return accountJSON, nil
+	// accountID resolved to a valid account, merge with AccountDefaults for a complete config
+	account := &config.Account{}
+	completeJSON, err := jsonpatch.MergePatch(accountDefaultsJSON, accountJSON)
```
mansinahar, line 172: With leaving the return type as raw JSON for `FetchAccount`, you can just return after this step and move the rest of the GDPR related backwards compatibility logic in the original `GetAccount` function itself. This will minimize the duplicate code needed in each of the fetcher implementation.

stored_requests/fetcher.go
```diff
@@ -210,18 +211,27 @@ func (f *fetcherWithCache) FetchResponses(ctx context.Context, ids []string) (da
...
+		}
 		return account, errs
 	} else {
 		f.metricsEngine.RecordAccountCacheResult(metrics.CacheMiss, 1)
 	}
-	account, errs = f.fetcher.FetchAccount(ctx, accountID)
+	account, errs = f.fetcher.FetchAccount(ctx, accountDefaultJSON, accountID)
 	if len(errs) == 0 {
-		f.cache.Accounts.Save(ctx, map[string]json.RawMessage{accountID: account})
+		accountJSON, err := json.Marshal(account)
```
mansinahar, line 230: Once you make those above changes, you won't have to do JSON marshalling here before saving it in the cache as you'll already have it as raw JSON string from `FetchAccount`

mansinahar (commented):
@Pubmatic-Dhruv-Sonone Thanks for taking the effort to make Prebid Server more efficient. I like the motivation behind this. However, given that we do have fetchers refactor on our radar some time for next year, I think we can keep the changes minimal while achieving what you're aiming for. Please see my comments below.

bsardo:
@Pubmatic-Dhruv-Sonone just curious what your thoughts are regarding the comments we left above.

stored_requests/backends/http_fetcher/fetcher.go
Pubmatic-Dhruv-Sonone: Done

stored_requests/fetcher.go
Pubmatic-Dhruv-Sonone: Reverted the account marshalling changes.

Pubmatic-Dhruv-Sonone:
> @Pubmatic-Dhruv-Sonone just curious what your thoughts are regarding the comments we left above.

@bsardo Sorry for the delay, I was on leave last few weeks. I have addressed the above comments.

stored_requests/backends/http_fetcher/fetcher_test.go
```diff
@@ -249,7 +249,7 @@ func newAccountHandler(t *testing.T, expectAccIDs []string, jsonifier func(strin
 
 		for _, accID := range gotAccIDs {
 			if accID != "" {
-				accIDResponse[accID] = jsonifier(accID)
+				accIDResponse[accID] = json.RawMessage(`{"id":"` + accID + `"}`)
```
mansinahar, line 252: Why not just use the existing `jsonifier` function that's been sent as parameter to `newAccountHandler`?

stored_requests/backends/file_fetcher/fetcher.go
```diff
@@ -49,7 +50,12 @@ func (fetcher *eagerFetcher) FetchAccount(ctx context.Context, accountID string)
 			DataType: "Account",
 		}}
 	}
-	return accountJSON, nil
+
+	completeJSON, err := jsonpatch.MergePatch(accountDefaultsJSON, accountJSON)
```
mansinahar, line 54: We should add test cases to cover this new change where default account information is provided to `FetchAccount` and it is properly merged and sent back.

stored_requests/fetcher_test.go
```diff
@@ -243,7 +243,7 @@ func TestAccountCacheHit(t *testing.T) {
 		})
 
 	metricsEngine.On("RecordAccountCacheResult", metrics.CacheHit, 1)
-	account, errs := aFetcherWithCache.FetchAccount(ctx, "known")
+	account, errs := aFetcherWithCache.FetchAccount(ctx, json.RawMessage("{}"), "known")
```
mansinahar, line 246: Might be a good idea to add another test case with some default account JSON

endpoints/openrtb2/auction_test.go
```diff
@@ -62,10 +62,6 @@ func TestJsonSampleRequests(t *testing.T) {
 			"Makes sure we handle (default) aliased bidders properly",
 			"aliased",
 		},
-		{
```
mansinahar, line 65: May I ask what's the reason behind removing the malformed error test cases?

mansinahar (commented):
@Pubmatic-Dhruv-Sonone Thanks for making the changes. I have a couple more minor comments but other than that it looks great.

stored_requests/backends/file_fetcher/fetcher.go
Pubmatic-Dhruv-Sonone: Added new test cases.

stored_requests/fetcher_test.go
Pubmatic-Dhruv-Sonone: This test cases is for account cache. The flow does not reach the account json merge code, as account fetcher is mocked.

endpoints/openrtb2/auction_test.go
Pubmatic-Dhruv-Sonone: Reverted

stored_requests/backends/http_fetcher/fetcher_test.go
Pubmatic-Dhruv-Sonone: We need a valid account json for this test case, the existing value was fails in json patching.  Removed jsonifier as I am already creating a valid json here and jsonifier adds escape characters to json string, which causes unmarshalling error.

mansinahar (approved):

VeronikaSolovei9 (approved):
LGTM

PR #2378, opened 2022-09-14
AMP additionalConsent

guscarreon:
This pull request implements the second part of issue #1817. Particularly, the `addtl_consent` query field described in point number 5 of the issue's description:

5. If `addtl_consent` is supplied, copy its value to `user.ext.ConsentedProvidersSettings.consented_providers`

According to specs, is set independently of the `consent_type`. This pull request's core change happens inside the `endpoints/openrtb2/amp_auction.go` and is shown below:

```
436     setAmpExt(req.Site, "1")
437
438     setEffectiveAmpPubID(req, ampParams.Account)
439
440     if ampParams.Slot != "" {
441         req.Imp[0].TagID = ampParams.Slot
442     }
443
    +   if err := setConsentedProviders(req, ampParams); err != nil {
    +       return []error{err}
    +   }
    +
444     policyWriter, policyWriterErr := amp.ReadPolicy(ampParams, deps.cfg.GDPR.Enabled)
445     if policyWriterErr != nil {
446         return []error{policyWriterErr}
447     }
endpoints/openrtb2/amp_auction.go
```

All other changes are to support or test the addition shown above.

PR #2443, opened 2022-11-01
Declare request wrapper keys as constants

guscarreon:
In pull request #2378 we decided to put request wrapper keys into constants. This is the second part of said endevour, coverts `"consent"`, `"amp"`, `"eids"`, `"gdpr"`, `"prebid"`, `"schain"` and `"us_privacy"`.

### DIFF
--- a/openrtb_ext/request_wrapper.go
+++ b/openrtb_ext/request_wrapper.go
@@ -44,8 +44,8 @@ type RequestWrapper struct {
 
 const (
 	jsonEmptyObjectLength               = 2
-	ConsentedProvidersSettingsStringKey = "ConsentedProvidersSettings"
-	ConsentedProvidersSettingsListKey   = "consented_providers_settings"
+	consentedProvidersSettingsStringKey = "ConsentedProvidersSettings"
+	consentedProvidersSettingsListKey   = "consented_providers_settings"
 	consent                             = "consent"
 	amp                                 = "amp"
 	eids                                = "eids"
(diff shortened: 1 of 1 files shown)


SyntaxNode:
Why? Does this provide a performance gain of some kind?

guscarreon:
> Why? Does this provide a performance gain of some kind?

None that I can think of. We thought the constants are easier to manage.

VeronikaSolovei9 (dismissed):
Looks like a good change

openrtb_ext/request_wrapper.go
```diff
@@ -44,8 +44,15 @@ type RequestWrapper struct {
 
 const (
 	jsonEmptyObjectLength               = 2
-	ConsentedProvidersSettingsStringKey = "ConsentedProvidersSettings"
-	ConsentedProvidersSettingsListKey   = "consented_providers_settings"
+	consentedProvidersSettingsStringKey = "ConsentedProvidersSettings"
+	consentedProvidersSettingsListKey   = "consented_providers_settings"
+	consent                             = "consent"
```
mansinahar, line 49: @guscarreon I like the change from hardcoded strings to constants. Thanks for doing that. However, one minor comment, for these constant names we should decide on whether we want the `Key` suffix for them or not and keep it consistent among all constant names that represent a hardcoded string. I personally lean towards having the suffix but don't feel too strongly about it so will leave it up to you.

VeronikaSolovei9 (approved):

Issue #2411, opened 2022-10-07
Define how server-side adapters should pass fledge auction config


bretg:
Related to https://github.com/prebid/Prebid.js/issues/9080

We need to define how server-side adapters should pass fledge auction config. 

We're working out the PBS-PBJS protocol in the other issue. This one is for tracking how adapters specify the value.

bretg:
Here's the format of the wire protocol as discussed in https://github.com/prebid/Prebid.js/issues/9080
```
{
  ext: {
    fledge: {
      auctionconfigs: [{
          impid: "123",
          bidder: "bidderA",
          config: { ... }
     },{
          impid: "123",
          bidder: "bidderB",
          config: { ... }
     }]
  }
}
```

So I think PBS-core will have the following responsibilities:
1. Pick up the fledge auctionconfig from wherever bid adapters are placing it
2. Make sure that bidders aren't submitting fledge auctionconfig for other bidders. (subject to the "multiple biddercode" rules)
3. Make sure that fledge auctionconfig is associated with a valid imp object

No need to validate the body of the fledge auctionconfig

I'm too fuzzy on the bid adapter interface to suggest the path-forward here. Need @SyntaxNode and @net-burst to weigh in on this aspect.

bretg:
We discussed fledge in the PBJS committee. The JSON format is approved, but might change in the future if the IAB comes up with a different format.

The preference is that "bidder" is the biddercode that's in imp.ext.prebid.bidders... i.e. 'pubm_ss' rather than 'pubmatic'. Not clear to me this is available to bid adapters. 

If the bid adapter is aware of how it was invoked, then it should use that code. This might differ if the adapter is fronting other bidders, i.e. in "allowUnknownBidderCodes" mode. In that case, it seems that the adapter will need to return the best biddercode it has available.

laurb9:
The bidder code could (and I would argue, should) be added by the framework. In PBS-Go bidders respond with a [structure](https://codinginadtech.com/prebid-live-tests/fledge/gpt-simple.html) that has Bids and Currency, we can leverage that to pass the auction config array.

I may have some bandwidth so I can look into implementing this once the prebid.js prebid-server adapter convention is sorted out.

laurb9:
We should also discuss how the "auction environment" parameter `ae` should be passed. There are two options: 
- `imp.ext.ae`/`imp.ext.auction_environment` hoping to be eventually standardized, or 
[...]

PR #2445, opened 2022-11-02
FLEDGE server support

laurb9:
- Request must include `imp[].ext.ae=1`
- Bidders receive `ExtImpBidder.AuctionEnvironment` 
- Bidders return `BidResponse.FledgeAuctionConfigs` as a `[]FledgeAuctionConfig{ImpId, Bidder, Config}`
- Adaptor code copies it to `pbsOrtbSeatBid.fledgeAuctionConfigs` field
- exchange bid postprocessing in `getAllBids` splits fledgeAuctionConfigs from bids and merges them into what will become `ext.prebid.fledge.auctionconfigs[]`

Potential nice to haves ? listed for tracking and discussion but they don't hold back the functionality implemented.
- [x] change to `imp.ext.ae`
- [ ] validation on impid matching
- [ ] validation that imp request had ae: 1
- [x] bidderName validation or override
- [x] handle no-bids with fledge response, as they are currently discarded
- [ ] dedupe responses from same bidder for same impid ? not sure how
- [x] support stored responses with fledge payload

PBS-Java PR: https://github.com/prebid/prebid-server-java/pull/2092


bsardo:
@laurb9 if you don't mind, can you please open a separate PR for your adapter changes so that they are separate from the PBS Core changes when this is ready for review?

laurb9:
> if you don't mind, can you please open a separate PR for your adapter changes so that they are separate from the PBS Core changes when this is ready for review?

will do!

bretg:
related to issue https://github.com/prebid/prebid-server/issues/2411

hhhjort:
One possible deduping solution would be to choose the "first" or "last" config seen for an seat/ImpID pair, and log a warning that the extra fledge config(s) were dropped. First and last are in quotes, as I don't think there is any guarantee on what order the fledge configs would be processed in.

In addition, it may be an idea to track bidder/seatID/ImpID, as one source of duplication could be two bidders both bidding into the same seat ID. This will allow for a descriptive warning that would pinpoint the source of the issue (one bidder providing duplicate fledges vs. two bidders trying to use the same seat)

hhhjort:
The three unchecked items. Are you planning on doing them in a separate PR, or are you still planning on handling them in this PR?

laurb9:
> The three unchecked items. Are you planning on doing them in a separate PR, or are you still planning on handling them in this PR?

I considered leaving them out of this PR and address later if we encounter issues in actual use. Open for discussion of course. The thought process leading to this is that FLEDGE is an emerging proposal, supporting it minimally enables changes to happen faster. There's no harm at the moment if PBS returns all responses as-is and lets upstream make sense of them. Prebid.js might have to do the same type of work while processing responses from PBS and the other bidders anyway. Thoughts ?

hhhjort (approved):
Looks good for the functionality included.

hhhjort (approved):

laurb9:
Found through manual integration that `validateImpExt` now rejects imp.ext.ae as unknown bidder  (_request.imp[0].ext.prebid.bidder contains unkown bidder: ae_). Looking into it and should have a fix shortly.

bsardo:
@laurb9 it looks like you'll need to merge with master to resolve conflicts. We can give this another look after that.

exchange/exchange_test.go
```diff
@@ -2297,6 +2297,8 @@ func runSpec(t *testing.T, filename string, spec *exchangeSpec) {
 			assert.JSONEq(t, expectedPassthough, actualPassthrough, "Expected bid response extension is incorrect")
 		}
 
+	} else if spec.Response.Ext != nil {
+		assert.JSONEq(t, string(spec.Response.Ext), string(bid.Ext), "ext mismatch")
```
laurb9, line 2301: This unconditional `Response.Ext` check uncovered a "false pass" in the json tests added with #2450 
`spec.BidValidationEnforcement` is never set so the section below does not run, and now the `ext` check I added picked up on the difference. I'll add a comment on that PR.

After the fixes we will still have three separate places that check `ext` - bid validation, passthrough and fledge. Would be nice to consolidate - the issue for me was the ever-present "debugging disabled for account" message would require touching too many unrelated (for me) files to fix.

AlexBVolcy: Hey @laurb9 I just opened a PR https://github.com/prebid/prebid-server/pull/2529, to address the false pass in relation to `spec.BidValidationEnforcement`

hhhjort (approved):

exchange/bidder.go
```diff
@@ -275,6 +275,16 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, bidderRequest Bidde
 					}
 				}
 
+				// FLEDGE auctionconfig responses are sent separate from bids
+				if bidResponse.FledgeAuctionConfigs != nil {
+					// TODO: only if ae was set in incoming request (imp.ext.ae etc)
```
AlexBVolcy, line 280: Can we remove this TODO comment?

exchange/exchange.go
```diff
@@ -615,7 +628,26 @@ func (e *exchange) getAllBids(
...
+		if fledge == nil {
+			fledge = &openrtb_ext.Fledge{
+				AuctionConfigs: make([]*openrtb_ext.FledgeAuctionConfig, 0, len(seatBid.FledgeAuctionConfigs)),
+			}
+		}
+		for _, config := range seatBid.FledgeAuctionConfigs {
+			fledge.AuctionConfigs = append(fledge.AuctionConfigs, &openrtb_ext.FledgeAuctionConfig{
+				Bidder:  bidderName.String(),
+				Adapter: config.Bidder,
+				ImpId:   config.ImpId, // TODO: validate impId
```
AlexBVolcy, line 645: Can we remove this TODO comment?

exchange/exchange.go
```diff
@@ -593,19 +598,27 @@ func (e *exchange) getAllBids(
...
 	for i := 0; i < len(bidderRequests); i++ {
 		brw := <-chBids
 
 		//if bidder returned no bids back - remove bidder from further processing
 		for _, seatBid := range brw.adapterSeatBids {
-			if seatBid != nil && len(seatBid.Bids) != 0 {
-				if _, ok := adapterBids[openrtb_ext.BidderName(seatBid.Seat)]; ok {
-					adapterBids[openrtb_ext.BidderName(seatBid.Seat)].Bids = append(adapterBids[openrtb_ext.BidderName(seatBid.Seat)].Bids, seatBid.Bids...)
+			if seatBid == nil {
+				continue
```
AlexBVolcy, line 610: This line isn't covered in tests, would you be able to update tests to cover it?

exchange/exchange.go
```diff
@@ -1254,20 +1293,50 @@ func buildStoredAuctionResponse(storedAuctionResponses map[string]json.RawMessag
...
+							AuctionConfigs: make([]*openrtb_ext.FledgeAuctionConfig, 0, len(auctionConfigs)),
+						}
+					}
+					for _, config := range auctionConfigs {
+						newConfig := &openrtb_ext.FledgeAuctionConfig{
+							ImpId:  impId,
+							Bidder: string(bidderName),
+							Config: config.Config,
+						}
+						fledge.AuctionConfigs = append(fledge.AuctionConfigs, newConfig)
```
AlexBVolcy, line 1317: This chunk of code is not covered by tests. Can you update test to cover this?

exchange/bidder.go
```diff
@@ -275,6 +275,16 @@ func (bidder *bidderAdapter) requestBid(ctx context.Context, bidderRequest Bidde
...
+					if fledgeAuctionConfigs := seatBidMap[bidderRequest.BidderName].FledgeAuctionConfigs; fledgeAuctionConfigs != nil {
+						seatBidMap[bidderRequest.BidderName].FledgeAuctionConfigs = append(fledgeAuctionConfigs, bidResponse.FledgeAuctionConfigs...)
+					} else {
+						seatBidMap[bidderRequest.BidderName].FledgeAuctionConfigs = bidResponse.FledgeAuctionConfigs
```
AlexBVolcy, line 284: This chunk of code isn't covered by tests. Can you update tests to cover this section?

hhhjort (approved):

AlexBVolcy (approved):
LGTM

PR #2449, opened 2022-11-07
IX Adapter: Pass version information in Request.Ext field

ccorbo:
This update to the IX Adapter is to pass Prebid Server Version information and PBJS version information as part of a new "IXDiag" field. This field will be used to pass any diagnostic information as needed.

Changes include changese to ix.go and ix_test.go.

### DIFF
--- a/adapters/ix/ix.go
+++ b/adapters/ix/ix.go
@@ -337,14 +337,12 @@ func BuildIxDiag(request *openrtb2.BidRequest) error {
 	ixdiag := &IxDiag{}
 
 	if extRequest.Prebid != nil {
-		ixdiag = &IxDiag{
-			PbjsV: extRequest.Prebid.Channel.Version,
-		}
+		ixdiag.PbjsV = extRequest.Prebid.Channel.Version
 	}
 
 	// Slice commit hash out of version
 	if strings.Contains(version.Ver, "-") {
-		ixdiag.PbsV = version.Ver[:strings.LastIndex(version.Ver, "-")]
+		ixdiag.PbsV = version.Ver[:strings.Index(version.Ver, "-")]
 	} else if version.Ver != "" {
 		ixdiag.PbsV = version.Ver
 	}
@@ -352,7 +350,10 @@ func BuildIxDiag(request *openrtb2.BidRequest) error {
 	// Only set request.ext if ixDiag is not empty
 	if *ixdiag != (IxDiag{}) {
 		extRequest.IxDiag = ixdiag
-		requestExtJson, _ := json.Marshal(extRequest)
+		requestExtJson, err := json.Marshal(extRequest)
+		if err != nil {
+			return err
+		}
 		request.Ext = requestExtJson
 	}
 	return nil
--- a/adapters/ix/ix.go
+++ b/adapters/ix/ix.go
@@ -11,6 +11,7 @@ import (
 	"github.com/prebid/prebid-server/config"
 	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/version"
 
 	"github.com/prebid/openrtb/v17/native1"
 	native1response "github.com/prebid/openrtb/v17/native1/response"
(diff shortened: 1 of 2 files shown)


adapters/ix/ix.go
```diff
@@ -309,3 +326,34 @@ func marshalJsonWithoutUnicode(v interface{}) (string, error) {
...
+	if request.Ext != nil {
+		if err := json.Unmarshal(request.Ext, &extRequest); err != nil {
+			return err
+		}
+	}
+	ixdiag := &IxDiag{}
+
+	if extRequest.Prebid != nil {
+		ixdiag = &IxDiag{
+			PbjsV: extRequest.Prebid.Channel.Version,
```
guscarreon, line 341: If condition in line 339 is true, I believe we are overwritting `ixdiag`'s original value. Can we assign `PbjsV`'s value as we do in lines 347 and 349 for `PbsV` instead?
```
337     ixdiag := &IxDiag{}
338
339     if extRequest.Prebid != nil {
340 -       ixdiag = &IxDiag{
341 -           PbjsV: extRequest.Prebid.Channel.Version,
342 -       }
    +       ixdiag.PbjsV = extRequest.Prebid.Channel.Version
343     }
344
345     // Slice commit hash out of version
346     if strings.Contains(version.Ver, "-") {
347         ixdiag.PbsV = version.Ver[:strings.LastIndex(version.Ver, "-")]
348     } else if version.Ver != "" {
349         ixdiag.PbsV = version.Ver
350     }
adapters/ix/ix.go
```

adapters/ix/ix.go
```diff
@@ -309,3 +326,34 @@ func marshalJsonWithoutUnicode(v interface{}) (string, error) {
...
+	if strings.Contains(version.Ver, "-") {
+		ixdiag.PbsV = version.Ver[:strings.LastIndex(version.Ver, "-")]
+	} else if version.Ver != "" {
+		ixdiag.PbsV = version.Ver
+	}
+
+	// Only set request.ext if ixDiag is not empty
+	if *ixdiag != (IxDiag{}) {
+		extRequest.IxDiag = ixdiag
+		requestExtJson, _ := json.Marshal(extRequest)
```
guscarreon, line 355: I realize it's very unlikely that the `Marshal` call returns an error here but, can we still add an error check?

adapters/ix/ix.go
```diff
@@ -309,3 +326,34 @@ func marshalJsonWithoutUnicode(v interface{}) (string, error) {
...
+		ixdiag = &IxDiag{
+			PbjsV: extRequest.Prebid.Channel.Version,
+		}
+	}
+
+	// Slice commit hash out of version
+	if strings.Contains(version.Ver, "-") {
+		ixdiag.PbsV = version.Ver[:strings.LastIndex(version.Ver, "-")]
+	} else if version.Ver != "" {
+		ixdiag.PbsV = version.Ver
```
guscarreon, line 349: `version.Ver` gets populated in runtime with the return value of `git describe --tags` that may return a string with two hyphens:
```
╰─➤  git describe --tags
0.23.1-3-g4ee257d8
```
Can we please modify this logic to correctly parse a string like this?

adapters/ix/ix_test.go
```diff
@@ -101,3 +103,99 @@ func TestIxMakeBidsWithCategoryDuration(t *testing.T) {
...
+				ID:  "1",
+				Ext: json.RawMessage(`{"prebid":"channel":{"name":"web","version":"7.20"}}}`),
+			},
+			expectedRequest: &openrtb2.BidRequest{
+				ID:  "1",
+				Ext: nil,
+			},
+			expectError: true,
+			pbsVersion:  "1.880-abcdef",
+		},
```
guscarreon, line 178: Can we please add a test case where a string comes with two hyphens?
```
╰─➤  git describe --tags
0.23.1-3-g4ee257d8
```

adapters/ix/ix_test.go
```diff
@@ -101,3 +103,99 @@ func TestIxMakeBidsWithCategoryDuration(t *testing.T) {
...
+	}
+
+	for _, test := range testCases {
+		version.Ver = test.pbsVersion
```
guscarreon, line 182: Can we reset `version.Ver` back to its original empty `""` value please?
```
181       for _, test := range testCases {
182           version.Ver = test.pbsVersion
183           err := BuildIxDiag(test.request)
184           if test.expectError {
185               assert.NotNil(t, err)
186           } else {
187               assert.Equal(t, test.request, test.expectedRequest)
188               assert.Nil(t, err)
189           }
190       }
    +         version.Ver = ""
191   }
```

guscarreon (commented):
Looks pretty good. Left a couple of comments

adapters/ix/ix_test.go
ccorbo: yep updated

adapters/ix/ix_test.go
ccorbo: yep done

adapters/ix/ix.go
ccorbo: thanks, updated

guscarreon (approved):
@ccorbo thank you for addressing my feedback. Change looks good to me.

VeronikaSolovei9 (approved):
LGTM

PR #2462, opened 2022-11-22
Modularity: Hook invocation Process [Entrypoint Stage]

legendko:
1. Added implementation for module/hook invocation. Complete flow is ready for the `Entrypoint stage` for `/openrtb2/auction` and `/openrtb2/amp` endpoints. The execution of stages consists of multiple generic functions which will allow to reuse the invocation process for the next stages.
2. The modules may `reject` the request entirely returning an `empty bid response` with corresponding `NBR code`.
3. `StageOutcome` will be used for storing the details of the modules' execution results and in case debug is enabled they will be added to the response.


endpoints/openrtb2/amp_auction.go
```diff
@@ -139,6 +144,11 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 	w.Header().Set("Access-Control-Expose-Headers", "AMP-Access-Control-Allow-Source-Origin")
 	w.Header().Set("X-Prebid", version.BuildXPrebidHeader(version.Ver))
 
+	if _, reject := deps.hookExecutor.ExecuteEntrypointStage(r, nil); reject != nil {
```
SyntaxNode, line 147: I recommend adding a comment explaining there is no body for AMP requests to explain the nil argument. Or perhaps use a private package scoped variable called `nilBody` set to nil and pass that in here.

endpoints/openrtb2/amp_auction.go
```diff
@@ -215,12 +225,13 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
 		PubID:                      labels.PubID,
+		HookExecutor:               deps.hookExecutor,
 	}
 
 	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
 	ao.AuctionResponse = response
-
-	if err != nil {
+	reject := hookexecution.FindReject([]error{err})
+	if err != nil && reject == nil {
```
SyntaxNode, line 234: The `FindReject` function is allocating memory just to check the type, which is not ideal for performance. I also don't believe it needs to return a pointer. Please implement something like a new `CastRejectErr` method:

```
func FindFirstRejectOrNil(errors []error) RejectError {
	for _, err := range errors {
		if rejectErr, ok := CastRejectErr(err); ok {
			return rejectErr
		}
	}
	return nil
}

func CastRejectErr(err error) (RejectError, bool) {
    return err.(RejectError)
}
```

and update the call side to:

```
rejectErr := hookexecution.CastRejectErr(err)
```

endpoints/openrtb2/amp_auction.go
```diff
@@ -240,6 +251,41 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
+	reject hookexecution.RejectError,
+	w http.ResponseWriter,
+	reqWrapper *openrtb_ext.RequestWrapper,
+	labels metrics.Labels,
+	ao analytics.AmpObject,
+	errs []error,
+) (metrics.Labels, analytics.AmpObject) {
+	nbr := openrtb3.NoBidReason(reject.NBR)
+	response := &openrtb2.BidResponse{
+		NBR: &nbr,
```
SyntaxNode, line 272: You can simplify this with: `NBR: reject.NBR.Ptr()`

endpoints/openrtb2/amp_auction.go
```diff
@@ -298,12 +344,14 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
 
 	// add debug information if requested
-	if reqWrapper.Test == 1 && eRErr == nil {
-		if extResponse.Debug != nil {
-			ampResponse.Debug = extResponse.Debug
-		} else {
-			glog.Errorf("Test set on request but debug not present in response: %v", err)
-			ao.Errors = append(ao.Errors, fmt.Errorf("Test set on request but debug not present in response: %v", err))
+	if reqWrapper != nil {
+		if reqWrapper.Test == 1 && eRErr == nil {
```
SyntaxNode, line 348: You can remove an `if` statement by combining the top two conditions:
`if reqWrapper != nil  && reqWrapper.Test == 1 && eRErr == nil {`

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1929,3 +1934,69 @@ func TestSetTargeting(t *testing.T) {
 
 	}
 }
+
+func TestValidAmpResponseWhenRequestRejected(t *testing.T) {
+	nbr := openrtb3.NoBidReason(123)
+	reject := hookexecution.RejectError{
+		int(nbr),
+		hookexecution.HookID{"foobar", "foo"},
+		hooks.StageEntrypoint.String(),
```
SyntaxNode, line 1943: Please always include labels on object initializers.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1929,3 +1934,69 @@ func TestSetTargeting(t *testing.T) {
...
+		description         string
+		expectedAmpResponse AmpResponse
+		hookExecutor        hookexecution.HookStageExecutor
+	}{
+		{
+			"Assert correct AmpResponse when request rejected at entrypoint stage",
+			AmpResponse{Targeting: map[string]string{}},
+			rejectableHookExecutor{entrypointReject: &reject},
+		},
+	}
```
SyntaxNode, line 1956: There is no need to use a table driven test if there is only 1 condition. Will there be more conditions in the near future?

endpoints/openrtb2/auction.go
```diff
@@ -218,12 +227,14 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
 		BidderImpReplaceImpID:      bidderImpReplaceImp,
 		PubID:                      labels.PubID,
+		HookExecutor:               deps.hookExecutor,
 	}
 	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
 	ao.Request = req.BidRequest
 	ao.Response = response
 	ao.Account = account
-	if err != nil {
+	reject := hookexecution.FindReject([]error{err})
```
SyntaxNode, line 236: Please use the same new reject method here as in AMP to avoid allocating an array/slice just to perform the type check.

endpoints/openrtb2/auction.go
```diff
@@ -285,6 +331,15 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
...
 	}
 
+	requestJson, reject := deps.hookExecutor.ExecuteEntrypointStage(httpRequest, requestJson)
+	if reject != nil {
+		errs = []error{reject}
+		if err = json.Unmarshal(requestJson, req.BidRequest); err != nil {
+			glog.Errorf("Failed to unmarshal BidRequest during entrypoint rejection: %s", err)
+		}
+		return
+	}
```
SyntaxNode, line 341: Why are we trying to unmarshal if the module has issues a rejection?

endpoints/openrtb2/auction_test.go
```diff
@@ -4650,6 +4654,69 @@ func TestValidateStoredResp(t *testing.T) {
 	}
 }
 
+func TestValidResponseWhenRequestRejected(t *testing.T) {
+	nbr := openrtb3.NoBidReason(123)
+	reject := hookexecution.RejectError{
+		int(nbr),
+		hookexecution.HookID{"foobar", "foo"},
+		hooks.StageEntrypoint.String(),
```
SyntaxNode, line 4662: Please always include labels on object initializers.

endpoints/openrtb2/auction_test.go
```diff
@@ -4650,6 +4654,69 @@ func TestValidateStoredResp(t *testing.T) {
...
+
+	testCases := []struct {
+		description         string
+		expectedBidResponse openrtb2.BidResponse
+		hookExecutor        hookexecution.HookStageExecutor
+	}{
+		{
+			"Assert correct BidResponse when request rejected at entrypoint stage",
+			openrtb2.BidResponse{ID: "some-request-id", NBR: &nbr},
+			rejectableHookExecutor{entrypointReject: &reject},
```
SyntaxNode, line 4673: Please always include labels on object initializers. If there is only one test case, please remove the table driven structure.

endpoints/openrtb2/auction_test.go
```diff
@@ -4650,6 +4654,69 @@ func TestValidateStoredResp(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.description, func(t *testing.T) {
```
SyntaxNode, line 4678: Good use of `t.Run`. That's something we're missing from most of our tests. Glad to see if being added in by your tests.

hooks/hookexecution/errors.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+
+	"github.com/prebid/prebid-server/errortypes"
+)
+
+// TimeoutError indicates exceeding of the max execution time allotted for hook.
+type TimeoutError struct{}
+
+func (e TimeoutError) Error() string {
+	return fmt.Sprint("Hook execution timeout")
```
SyntaxNode, line 13: There is no need to use `Sprint` in this case.

hooks/hookexecution/errors.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// FailureError indicates expected error occurred during hook execution on the module-side.
+type FailureError struct{}
+
+func (e FailureError) Error() string {
+	return fmt.Sprint("Hook execution failed")
```
SyntaxNode, line 20: There is no need to use `Sprint` in this case.

hooks/hookexecution/execution.go
```diff
@@ -0,0 +1,233 @@
...
+		wg.Add(1)
+		go func(hw hooks.HookWrapper[H], moduleCtx hookstage.ModuleInvocationContext) {
+			defer wg.Done()
+			executeHook(moduleCtx, hw, payload, hookHandler, group.Timeout, resp, done)
+		}(hook, mCtx)
+	}
+
+	go func() {
+		wg.Wait()
+		close(resp)
```
SyntaxNode, line 75: Why is the `resp` channel being closed here? This introduces a race condition where `resp` channel would be closed before the `executeHook` have finished sending their results. Better to remove the wait group and add a defer to close the `resp` channel.

hooks/hookexecution/execution.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func executeGroup[H any, P any](
+	executionCtx executionContext,
+	group hooks.Group[H],
+	payload P,
+	hookHandler hookHandler[H, P],
+) (GroupOutcome, P, groupModuleContext, *RejectError) {
+	var wg sync.WaitGroup
+	done := make(chan struct{})
```
SyntaxNode, line 61: Better to rename `done` to its true purpose of `rejected`. It is not a way for `executeHook` to signal work has completed, but rather for `collectHookResponses` to exit pending `executeHook` calls early when a rejection is first encountered.

hooks/hookexecution/execution.go
```diff
@@ -0,0 +1,233 @@
...
+	payload P,
+	hookHandler hookHandler[H, P],
+	timeout time.Duration,
+	resp chan<- hookResponse[P],
+	done <-chan struct{},
+) {
+	hookRespCh := make(chan hookResponse[P], 1)
+
+	select {
+	case <-done:
```
SyntaxNode, line 95: This case can be removed. There is an incredibly low change a rejection would've occurred at this point where the code for this optimization isn't worth it IMHO.

hooks/hookexecution/execution.go
```diff
@@ -0,0 +1,233 @@
...
+			groupOutcome.InvocationResults[i].Errors = append(groupOutcome.InvocationResults[i].Errors, r.Err.Error())
+			switch r.Err.(type) {
+			case TimeoutError:
+				groupOutcome.InvocationResults[i].Status = StatusTimeout
+			case FailureError:
+				groupOutcome.InvocationResults[i].Status = StatusFailure
+			default:
+				groupOutcome.InvocationResults[i].Status = StatusExecutionFailure
+			}
+			continue
```
SyntaxNode, line 180: The code flow is hard to follow with all of the continue statements. What does it mean to continue within this loop? Is there a way to rewrite this loop and/or use helper methods to make the code flow more obvious?

hooks/hookexecution/executor.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/prebid/prebid-server/hooks"
+	"github.com/prebid/prebid-server/hooks/hookstage"
+)
+
+const (
+	EndpointAuction = "/openrtb2/auction"
+	EndpointAmp     = "/openrtb2/amp"
+)
+
+type entity string
```
SyntaxNode, line 20: What is an entity?

hooks/hookexecution/executor.go
```diff
@@ -0,0 +1,186 @@
...
+		for moduleName, moduleCtx := range moduleCtxs {
+			e.moduleContexts.put(moduleName, moduleCtx)
+		}
+	}
+}
+
+func (e *hookExecutor) pushStageOutcome(outcome StageOutcome) {
+	e.Lock()
+	e.stageOutcomes = append(e.stageOutcomes, outcome)
+	defer e.Unlock()
```
SyntaxNode, line 152: Please move the defer up one line to be directly below the `Lock` call.

hooks/hookstage/mutation.go
```diff
@@ -0,0 +1,52 @@
+package hookstage
+
+type MutationType int
+
+const (
+	MutationUpdate MutationType = iota
+	MutationDelete
```
SyntaxNode, line 7: Do we also need "MutationAdd"?

endpoints/openrtb2/amp_auction_test.go
4lexvav: We are going to extend this with new test cases in the following PRs, to check rejections made at other stages.

endpoints/openrtb2/auction.go
4lexvav: You are right, this makes no sense within this PR, so I've removed it.
In one of the next PRs, we are adding debug info to `BidResponse`, but only if the debug mode enabled, which is determined from `BidRequest`.

endpoints/openrtb2/amp_auction.go
4lexvav: Sure, we will add `CastRejectErr` func to avoid extra memory allocation. However, since we work with the concrete type `RejectError`, we are forced to return a pointer if we want to return `nil` when `RejectError` is not found.

endpoints/openrtb2/auction_test.go
4lexvav: We will add new test cases in the next PRs here too.

hooks/hookexecution/execution.go
4lexvav: Perhaps I’m missing something, but in my understanding `wg.Wait()` is a blocking operation, so when `close(resp)` is called, all goroutines have already finished their execution and it’s safe to close the channel.

hooks/hookexecution/executor.go
4lexvav: An entity specifies the type of object that was processed during the execution of the stage.
For most stages, the entity will be equal to one of the constant values. But for the `bidder-request` and `raw-bidder-response` stages, the entity is set to the bidder name. This is similar to what PBS-Java does.

endpoints/openrtb2/amp_auction.go
SyntaxNode: Ah, ok. Got it. Makes sense.

endpoints/openrtb2/amp_auction_test.go
SyntaxNode: Gotcha. Sounds good.

hooks/hookexecution/execution.go
SyntaxNode: Yes, `wg.Wait()` is blocking and will not unblock until every `executeHook` completes. Let me take another look.

hooks/hookexecution/executor.go
SyntaxNode: Thank you. I think a comment will make it clear.

endpoints/openrtb2/amp_auction.go
```diff
@@ -240,6 +254,37 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
 
+	if isRejectErr {
+		labels, ao = rejectAmpRequest(*rejectErr, w, reqWrapper, labels, ao, errL)
+		return
+	}
+
+	labels, ao = sendAmpResponse(w, response, reqWrapper, labels, ao, errL)
+}
+
+func rejectAmpRequest(
```
AlexBVolcy, line 265: Nitpick: Could you re-format this function so the parameters are all on one line to follow the patterns of other functions in this file?

endpoints/openrtb2/amp_auction.go
```diff
@@ -240,6 +254,37 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
...
+	errs []error,
+) (metrics.Labels, analytics.AmpObject) {
+	response := &openrtb2.BidResponse{NBR: openrtb3.NoBidReason(rejectErr.NBR).Ptr()}
+	ao.AuctionResponse = response
+	ao.Errors = append(ao.Errors, rejectErr)
+
+	return sendAmpResponse(w, response, reqWrapper, labels, ao, errs)
+}
+
+func sendAmpResponse(
```
AlexBVolcy, line 280: Nitpick: Could you re-format this function so the parameters are all on one line to follow the patterns of other functions in this file?

endpoints/openrtb2/auction.go
```diff
@@ -235,8 +246,38 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		ao.Status = http.StatusInternalServerError
 		ao.Errors = append(ao.Errors, err)
 		return
+	} else if isRejectErr {
+		labels, ao = rejectAuctionRequest(*rejectErr, w, req.BidRequest, labels, ao)
+		return
+	}
```
AlexBVolcy, line 252: Can you update tests to cover these lines if possible?

endpoints/openrtb2/auction.go
```diff
@@ -235,8 +246,38 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+
+	labels, ao = sendAuctionResponse(w, response, labels, ao)
+}
+
+func rejectAuctionRequest(
```
AlexBVolcy, line 257: Same parameter one line re-formatting comment

endpoints/openrtb2/auction.go
```diff
@@ -235,8 +246,38 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+		response.ID = request.ID
 	}
 
+	ao.Response = response
+	ao.Errors = append(ao.Errors, rejectErr)
+
+	return sendAuctionResponse(w, response, labels, ao)
+}
+
+func sendAuctionResponse(
```
AlexBVolcy, line 275: Same parameter one line re-formatting comment

hooks/hookexecution/context.go
```diff
@@ -0,0 +1,71 @@
...
+	accountId      string
+	account        *config.Account
+	moduleContexts *moduleContexts
+}
+
+func (ctx executionContext) getModuleContext(moduleName string) hookstage.ModuleInvocationContext {
+	moduleInvocationCtx := hookstage.ModuleInvocationContext{Endpoint: ctx.endpoint}
+	if ctx.moduleContexts != nil {
+		if mc, ok := ctx.moduleContexts.get(moduleName); ok {
+			moduleInvocationCtx.ModuleContext = mc
```
AlexBVolcy, line 24: Is it possible to update tests to cover this line?

hooks/hookexecution/context.go
```diff
@@ -0,0 +1,71 @@
...
+func (mc *moduleContexts) put(moduleName string, mCtx hookstage.ModuleContext) {
+	mc.Lock()
+	defer mc.Unlock()
+
+	newCtx := mCtx
+	if existingCtx, ok := mc.ctxs[moduleName]; ok && existingCtx != nil {
+		for k, v := range mCtx {
+			existingCtx[k] = v
+		}
+		newCtx = existingCtx
```
AlexBVolcy, line 54: Is it possible to update tests to cover these lines?

hooks/hookexecution/execution.go
AlexBVolcy: Yeah I would agree with this comment

hooks/hookexecution/execution.go
```diff
@@ -0,0 +1,227 @@
...
+			if !stage.IsRejectable() {
+				groupOutcome.InvocationResults[i].Status = StatusExecutionFailure
+				groupOutcome.InvocationResults[i].Errors = append(
+					groupOutcome.InvocationResults[i].Errors,
+					fmt.Sprintf(
+						"Module (name: %s, hook code: %s) tried to reject request on the %s stage that does not support rejection",
+						r.HookID.ModuleCode,
+						r.HookID.HookCode,
+						executionCtx.stage,
+					),
```
AlexBVolcy, line 187: Is it possible to update tests to cover these lines?

hooks/hookexecution/executor.go
```diff
@@ -0,0 +1,187 @@
...
+	return nil
+}
+
+func (e *hookExecutor) ExecuteAllProcessedBidResponsesStage(responses []*adapters.BidderResponse) {
+	//TODO: implement
+}
+
+func (e *hookExecutor) ExecuteAuctionResponseStage(response *openrtb2.BidResponse) {
+	//TODO: implement
+}
```
AlexBVolcy, line 140: Is it necessary to have these functions defined if there is no code actually present in them yet?

hooks/hookexecution/execution.go
4lexvav: Sure, we will add more tests to ensure 100% coverage.

hooks/hookexecution/execution.go
4lexvav: We've done some refactoring of this function, please check if it's clearer now.

endpoints/openrtb2/amp_auction.go
4lexvav: Please clarify, do you wish us to re-format functions to one line only in files we edited or is it a whole project convention and we should use the same format across all functions we created? Because we tried to keep the line length under 120 characters long in all functions we created.

endpoints/openrtb2/auction.go
4lexvav: Actually, in this PR we do not invoke any hooks within the `exchange.HoldAuction` method, so there will be no rejection error and this code is not unreachable right now. But it will be added in next PR with corresponding tests, so if you think it is better to get out of this code from current PR - we can remove it for now.

hooks/hookexecution/executor.go
4lexvav: Yes, we can remove it from this PR and add it in subsequent PRs with implementation of these methods.

hooks/hookexecution/context.go
4lexvav: The purpose of this logic is for communication between different stages, so that module can pass some data to itself from earlier stages to later ones. Right now we have implementation for `entrypoint` stage only, but we will definitely cover this logic with implementations of new stages.

hooks/hookexecution/execution.go
4lexvav: Unfortunately, right now we have implementation only for the `entrypoint` stage which is rejectable. But we are going to cover this logic with tests within the PRs for `all-processed-bid-responses` and `auction-response` stages, for which rejection is not allowed.

endpoints/openrtb2/amp_auction.go
bsardo: Hi @4lexvav, the team discussed this offline and decided that we want to favor readability over consistency in this case so we prefer you stick with your approach of breaking long one-line function signatures up so that they span multiple lines. 120 characters seems like a good rule of thumb.
@AlexBVolcy, there are some cases where we already do this and we will continue to make this change throughout our code base over time.

hooks/hookexecution/execution.go
4lexvav: We added tests to cover more cases, but right now we cannot achieve 100% code coverage because some of the logic is reachable at later stages only. But we will add more tests in PRs with new stages.

endpoints/openrtb2/amp_auction.go
4lexvav: Sure, I restored the multiline format of function signatures.

endpoints/openrtb2/auction_test.go
```diff
@@ -4650,6 +4654,69 @@ func TestValidateStoredResp(t *testing.T) {
...
+				empty_fetcher.EmptyFetcher{},
+				test.hookExecutor,
+			}
+
+			reqBody := validRequest(t, "../exemplary/simple.json")
+			req := httptest.NewRequest("POST", "/openrtb2/auction", strings.NewReader(reqBody))
+			recorder := httptest.NewRecorder()
+
+			deps.Auction(recorder, req, nil)
+			assert.Equal(t, recorder.Code, http.StatusOK, "Endpoint should return 200 OK.")
```
bsardo, line 4707: According to the spec, a rejection in the entrypoint stage should result in a 204 or 4XX:

> For /openrtb2/auction and /openrtb2/video modules must be able to modify the original POST request that comes into the server or reject the request entirely. Rejection at this stage may result in HTTP 204 or a 4xx error and a metric associating the rejection with the module.

In that case, the body should be empty if using 204 or contain an error message if using 4XX.

What is PBS-Java doing?

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1929,3 +1933,68 @@ func TestSetTargeting(t *testing.T) {
...
+				hardcodedResponseIPValidator{response: true},
+				empty_fetcher.EmptyFetcher{},
+				test.hookExecutor,
+			}
+
+			req := httptest.NewRequest("POST", "/openrtb2/amp?tag_id=1", nil)
+			recorder := httptest.NewRecorder()
+
+			deps.AmpAuction(recorder, req, nil)
+			assert.Equal(t, recorder.Code, http.StatusOK, "Endpoint should return 200 OK.")
```
bsardo, line 1989: According to the spec, a rejection at the entrypoint stage should result in a 204:

> For /openrtb2/amp and /setuid, modules must be able to modify the GET query string or reject the request entirely. Rejection results in HTTP 204 and a metric associating the rejection with the module.

In that case, the body should be empty if using 204.

hooks/hookstage/mutation.go
```diff
@@ -0,0 +1,54 @@
+package hookstage
+
+type MutationType int
+
+const (
+	MutationAdd MutationType = iota
```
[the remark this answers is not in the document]
4lexvav: Right now the hooks framework does not provide implementations for the `add`, `update` or `delete` operations. Instead, it all falls on the shoulders of the module that will create and return mutations from its hooks. Therefore, the module itself is responsible for choosing the right type for its mutations and defining the cases in which it should return an error. But we do not force modules to follow some algorithm for mutation types. The errors returned from mutations are logged as the warning messages and such mutations are skipped: https://github.com/prebid/prebid-server/blob/cfd8e8a8ba865def1fbabea2ceb5074a48701656/hooks/hookexecution/execution.go#L247-L254

But with introduction of module we will see more use cases and will be able to develop the `ChangeSet` concept to make it possible to provide implementations for different operations, define their behaviour and edge cases (in which case should the mutation fail or proceed?).

hooks/plan.go
```diff
@@ -8,26 +8,37 @@ import (
...
-	StageRawBidResponse           = "rawbidresponse"
-	StageAllProcessedBidResponses = "allprocbidresponses"
-	StageAuctionResponse          = "auctionresponse"
+	StageEntrypoint               Stage = "entrypoint"
+	StageRawAuction               Stage = "raw-auction"
+	StageProcessedAuction         Stage = "processed-auction"
+	StageBidderRequest            Stage = "bidder-request"
+	StageRawBidderResponse        Stage = "raw-bidder-response"
+	StageAllProcessedBidResponses Stage = "all-processed-bid-responses"
+	StageAuctionResponse          Stage = "auction-response"
```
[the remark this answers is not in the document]
legendko: updated

hooks/hookexecution/executor.go
```diff
@@ -0,0 +1,130 @@
...
+
+type HookStageExecutor interface {
+	StageExecutor
+	SetAccount(account *config.Account)
+	GetOutcomes() []StageOutcome
+}
+
+type hookExecutor struct {
+	account        *config.Account
+	accountId      string
```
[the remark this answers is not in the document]
legendko: done

hooks/hookexecution/outcome.go
```diff
@@ -0,0 +1,70 @@
...
+type Action string
+
+const (
+	ActionUpdate Action = "update"    // the hook returned mutations that were successfully applied
+	ActionReject Action = "reject"    // the hook decided to reject the stage
+	ActionNone   Action = "no_action" // the hook does not want to take any action
+)
+
+// Messages in format: {"module": {"hook": ["msg1", "msg2"]}}
+type Messages map[string]map[string][]string
```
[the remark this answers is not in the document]
legendko: Yes, this map will be used in the outcomes later for enriching the response. Removed from this PR.

hooks/hookexecution/outcome.go
```diff
@@ -0,0 +1,70 @@
...
+	Message       string                  `json:"message"` // arbitrary string value returned from hook execution
+	DebugMessages []string                `json:"debugmessages"`
+	Errors        []string                `json:"-"`
+	Warnings      []string                `json:"-"`
+}
+
+// HookID points to the specific hook defined by the hook execution plan.
+type HookID struct {
+	ModuleCode string `json:"module-code"`
+	HookCode   string `json:"hook-impl-code"`
```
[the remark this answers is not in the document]
legendko: Sure, we updated them with underscores. There was not any particular reason to use the flat case.

hooks/hookexecution/outcome_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func assertEqualStageOutcomes(t *testing.T, expected StageOutcome, actual StageOutcome) {
+	t.Helper()
+
+	require.Equal(t, len(actual.Groups), len(expected.Groups), "Stage outcomes contain different number of groups")
```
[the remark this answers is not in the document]
legendko: done

hooks/hookexecution/execution.go
```diff
@@ -0,0 +1,268 @@
...
+	case <-time.After(timeout):
+		resp <- hookResponse[P]{
+			Err:           TimeoutError{},
+			ExecutionTime: time.Since(startTime),
+			HookID:        hookId,
+			Result:        hookstage.HookResult[P]{},
+		}
+	case <-rejected:
+		return
+	}
```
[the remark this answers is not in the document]
legendko: Correct, we use this behavior for stopping the processing ASAP according to the PRD requirements. Other hooks from this group will finish their work in their own goroutines if they already started working, but their results wont have any effect. The next groups of hooks won't even start processing and we exit immediately.

endpoints/openrtb2/auction_test.go
legendko: Yes, we've read this but PBS-Java is always sending a 200 response with an empty bid response, so we decided to do the same for PBS-Go

endpoints/openrtb2/amp_auction_test.go
legendko: Same here

bsardo (approved):

AlexBVolcy (approved):

PR #2465, opened 2022-11-24
Rubicon: Support `bidonmultiformat` parameter

SerhiiNahornyi:


adapters/rubicon/rubicon.go
```diff
@@ -238,35 +233,18 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter, server co
...
 	errs := make([]error, 0, len(request.Imp))
-	var err error
 	requestData := make([]*adapters.RequestData, 0, numRequests)
 	headers := http.Header{}
 	headers.Add("Content-Type", "application/json;charset=utf-8")
 	headers.Add("Accept", "application/json")
 	headers.Add("User-Agent", "prebid-server/1.0")
 
-	requestImpCopy := request.Imp
+	impsToExtMap := prepareImpsToExtMap(createImpsToExtMap(request.Imp, errs))
```
hhhjort, line 242: I do not believe passing the `errs` slice in this way will update the slice in the current context. Like the append function returns a slice to make sure it is properly updated, you will have to do the same here to capture any errors generated within this function.

hhhjort (approved):
LGTM

bretg (changes requested):
sorry, but have to ask for a change here: we would like to rename the parameter from `multiformat` to `bidonmultiformat`

static/bidder-params/rubicon.json
```diff
@@ -38,6 +38,10 @@
         "type": "array"
       }
     },
+    "bidonmultiformat": {
```
mansinahar, line 41: @SerhiiNahornyi This PR LGTM. Can you please open up a docs PR for this new parameter if you already haven't done so? Thanks!

bretg: We don't really want general pubs to use this param yet, but I added a note about it as a "beta" parameter in https://github.com/prebid/prebid.github.io/pull/4212

bretg (approved):

bretg:
Note that this also hardcodes the secure flag to the Magnite exchange to `1`, which is an interim way to address https://github.com/prebid/prebid-server/issues/2478

hhhjort (approved):
LGTM

PR #2473, opened 2022-11-29
Stored responses: return nil if no stored reponses was found for ids

### LINKED ISSUE #2470: StoredResponse: unexpected json error if stored response for id is empty
Hi!

If the vendor running prebid server does not support storedResponses (for what ever reason) and the client sets storedAuctionResponse for placements, the auction will error with something like

```
E1128 11:08:52.434519   16648 auction.go:232] /openrtb2/auction Critical error: unexpected end of JSON input
```

as the `buildStoredAuctionResponse` will fail to unmarshal the empty body of the storedResponse. 

I propose that if the len of storedResponses returned by ProcessStoredResponse is 0, do a early return:
 
```
Author: Hugo Lindström <hugo.lindstrom@relevant-digital.com>
Date:   Mon Nov 28 10:43:55 2022 +0100

    prebid: ignore empty stored response for id

diff --git a/stored_responses/stored_responses.go b/stored_responses/stored_responses.go
index 533ff2c7..adbd7ee0 100644
--- a/stored_responses/stored_responses.go
+++ b/stored_responses/stored_responses.go
@@ -4,6 +4,7 @@ import (
        "context"
        "encoding/json"
        "fmt"
+
        "github.com/buger/jsonparser"
        "github.com/mxmCherry/openrtb/v16/openrtb2"
@@ -140,7 +141,7 @@ func ProcessStoredResponses(ctx context.Context, requestJson []byte, storedRespF

        if len(storedResponsesIds) > 0 {
                storedResponses, errs := storedRespFetcher.FetchResponses(ctx, storedResponsesIds)
-               if len(errs) > 0 {
+               if len(storedResponses) == 0 || len(errs) > 0 {
                        return nil, nil, nil, errs
                }
                bidderImpIdReplaceImp := flipMap(impBidderReplaceImp) 

```

If this a valid path forward, I am happy to submit a PR.

hugolm84:
If `storedbidresponse` id(s) was provided in the `imp` request, but no stored response(s) was found, return nil instead of error out on unmarshalling empty string into ExtImpPrebid.

Additionally, the mocked fetcher returns all mocked stored responses, which I believe is incorrect. It should return the responses requested, if any.

Closes #2470

### DIFF
--- a/stored_responses/stored_responses.go
+++ b/stored_responses/stored_responses.go
@@ -141,7 +141,7 @@ func ProcessStoredResponses(ctx context.Context, requestJson []byte, storedRespF
 
 	if len(storedResponsesIds) > 0 {
 		storedResponses, errs := storedRespFetcher.FetchResponses(ctx, storedResponsesIds)
-		if len(errs) > 0 {
+		if len(storedResponses) == 0 || len(errs) > 0 {
 			return nil, nil, nil, errs
 		}
 		bidderImpIdReplaceImp := flipMap(impBidderReplaceImp)
--- a/stored_responses/stored_responses_test.go
+++ b/stored_responses/stored_responses_test.go
@@ -350,6 +350,27 @@ func TestProcessStoredAuctionAndBidResponses(t *testing.T) {
 			expectedBidderImpReplaceImpID: BidderImpReplaceImpID{},
 		},
 		{
+			description: "Stored auction response one imp with no stored response",
+			requestJson: []byte(`{"imp": [
+				{
+					"id": "imp-id1",
+					"ext": {
+						"appnexus": {
+							"placementId": 123
+						},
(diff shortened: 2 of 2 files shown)


VeronikaSolovei9 (commented):
Need to confirm this change with the team. This update modifies the behavior and imp with stored response actually executes like a "real" imp if this stored response is not found.

VeronikaSolovei9:
Hello @hugolm84,
please refer to this discussion related to the changes in this PR: https://github.com/prebid/prebid-server/issues/2487
It appears the whole request needs to be rejected if stored bid response is not found. We will implement it later.
I'm sorry, we need to close this PR. 
Thank you for the contribution!

bsardo:
Closing. We'll implement the change described in #2487 shortly instead.

PR #2476, opened 2022-11-30
IX Adapter: Fix panic when ext.prebid.channel is nil

oronno:
Recent changes at https://github.com/prebid/prebid-server/pull/2449 cause panic when ext.prebid.channel is nil.
Fixing this with nil check and added test case to cover.


bsardo (approved):
LGTM

hhhjort (approved):
LGTM

PR #2480, opened 2022-12-05
adapter sspBC: Update to v5.8

spanwp:
Changes:
- Mcad argument handling has been simplified. Now it is passed as an object to the template and there is no need to call additional JS functions. 
- Increased version number.
- Updated tests.


hhhjort (approved):
Looks like a straightforward, small change.

spanwp:
@hhhjort Thanks for your quick review. Yes, it's a simple change. It allows as eliminate an error with the atob function.

AlexBVolcy (approved):

PR #2484, opened 2022-12-07
Modularity: Metrics infrastructure and collection

legendko:


metrics/metrics.go
```diff
@@ -49,6 +49,12 @@ type PrivacyLabels struct {
 	LMTEnforced    bool
 }
 
+type ModuleLabels struct {
+	Module string
+	Stage  string
+	PubID  string // account ID
```
SyntaxNode, line 55: Please use standard Go comment style, in this case something like "// PubID is the account id", or just rename `PubID` to `AccountID` if you'd like.

hooks/hookexecution/execution.go
```diff
@@ -166,8 +170,13 @@ func handleHookResponse[P any](
 	ctx executionContext,
 	payload P,
 	hr hookResponse[P],
+	metricEngine metrics.MetricsEngine,
 ) (P, HookOutcome, *RejectError) {
 	var rejectErr *RejectError
+	labels := metrics.ModuleLabels{Module: hr.HookID.ModuleCode, Stage: ctx.stage, PubID: ctx.accountId}
+	metricEngine.RecordModuleCalled(labels)
+	metricEngine.RecordModuleDuration(labels, hr.ExecutionTime)
```
SyntaxNode, line 178: I recommend combining `RecordModuleCalled` and `RecordModuleDuration` since they're only called here once as a pair and it makes sense IMHO for a call to be directly associated with a duration.

metrics/go_metrics.go
```diff
@@ -215,13 +238,25 @@ func NewBlankMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderNa
...
+	mn := make([]string, len(ms))
+
+	i := 0
+	for n := range ms {
+		mn[i] = n
+		i++
+	}
+
+	return mn
+}
```
SyntaxNode, line 251: I recommend changing the `mn` variable to `names` for improved readability.

metrics/go_metrics.go
```diff
@@ -781,3 +873,142 @@ func (me *Metrics) RecordAdsCertReq(success bool) {
...
+
+func (me *Metrics) RecordModuleDuration(labels ModuleLabels, duration time.Duration) {
+	mm, err := me.getModuleMetric(labels)
+	if err != nil {
+		return
+	}
+
+	// Module metrics
+	mm.DurationTimer.Update(duration)
+	// Account-Module metrics
```
SyntaxNode, line 885: Please add a line break before the Account-Module comment throughout to improve readability.

metrics/prometheus/preload.go
```diff
@@ -207,6 +208,40 @@ func preloadLabelValues(m *Metrics, syncerKeys []string) {
 			adapterLabel: adapterValues,
 		})
 	}
+
+	for module := range moduleStageNames {
+		preloadLabelValuesForHistogram(m.moduleDuration[module], map[string][]string{
+			stageLabel: stageValues,
```
SyntaxNode, line 214: Shouldn't this look more like:

```
for module, stageValues := range moduleStageNames {
  preloadLabelValuesForHistogram(m.moduleDuration[module], map[string][]string{
    stageLabel: stageValues,
```

This way we only pre-create labels which are needed for the particular module.

.. or else just use all possible stage values instead.

metrics/prometheus/type_conversion.go
```diff
@@ -130,3 +130,20 @@ func tcfVersionsAsString() []string {
...
+		for _, stage := range stageNames {
+			if _, ok := foundStages[stage]; !ok {
+				stages = append(stages, stage)
+				foundStages[stage] = struct{}{}
+			}
+		}
+	}
+
+	return stages
+}
```
SyntaxNode, line 149: Please see my earlier comment. This method is likely no longer necessary.

modules/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+		if !added {
+			return nil, fmt.Errorf(`hook "%s" does not implement any supported hook interface`, id)
+		}
+	}
+
+	return moduleStageNameCollector, nil
+}
+
+func addModuleStageName(moduleStageNameCollector map[string][]string, id string, stage string) map[string][]string {
+	replacer := strings.NewReplacer(".", "_", "-", "_")
```
SyntaxNode, line 67: Can you move the replacer to a top level package scoped variable to save on some allocations?

endpoints/openrtb2/amp_auction.go
```diff
@@ -77,7 +77,7 @@ func NewAmpEndpoint(
 		IPv6PrivateNetworks: cfg.RequestValidation.IPv6PrivateNetworksParsed,
 	}
 
-	hookExecutor := hookexecution.NewHookExecutor(hookExecutionPlanBuilder, hookexecution.EndpointAmp)
+	hookExecutor := hookexecution.NewHookExecutor(hookExecutionPlanBuilder, hookexecution.EndpointAmp, met)
```
AlexBVolcy, line 80: Would it be possible to update this name to something more descriptive like `metrics` or `metricsEngine`?

endpoints/openrtb2/auction.go
```diff
@@ -86,7 +86,7 @@ func NewEndpoint(
 		IPv6PrivateNetworks: cfg.RequestValidation.IPv6PrivateNetworksParsed,
 	}
 
-	hookExecutor := hookexecution.NewHookExecutor(hookExecutionPlanBuilder, hookexecution.EndpointAuction)
+	hookExecutor := hookexecution.NewHookExecutor(hookExecutionPlanBuilder, hookexecution.EndpointAuction, met)
```
AlexBVolcy, line 89: Would it be possible to update this name to something more descriptive like `metrics` or `metricsEngine`?

metrics/go_metrics.go
AlexBVolcy: Yeah I agree with the need to change the variable names here

metrics/go_metrics.go
```diff
@@ -311,6 +346,10 @@ func NewMetrics(registry metrics.Registry, exchanges []openrtb_ext.BidderName, d
 	newMetrics.AdsCertRequestsFailure = metrics.GetOrRegisterMeter("ads_cert_requests.failed", registry)
 	newMetrics.adsCertSignTimer = metrics.GetOrRegisterTimer("ads_cert_sign_time", registry)
 
+	for m, stages := range moduleStageNames {
```
AlexBVolcy, line 349: Could you rename `m` to be `modules`

metrics/go_metrics.go
```diff
@@ -342,6 +381,28 @@ func makeBlankAdapterMetrics(disabledMetrics config.DisabledMetrics) *AdapterMet
 	return newAdapter
 }
 
+func makeBlankModuleStageMetrics(stages []string) map[string]*ModuleMetrics {
+	mm := map[string]*ModuleMetrics{}
```
AlexBVolcy, line 385: Could this be renamed for readability?

metrics/go_metrics.go
```diff
@@ -781,3 +873,142 @@ func (me *Metrics) RecordAdsCertReq(success bool) {
...
+
+func (me *Metrics) RecordModuleCalled(labels ModuleLabels) {
+	mm, err := me.getModuleMetric(labels)
+	if err != nil {
+		return
+	}
+
+	// Module metrics
+	mm.CallCounter.Inc(1)
+	// Account-Module metrics
```
AlexBVolcy, line 901: Agree there should be a line break between line 900 and 901

metrics/go_metrics.go
```diff
@@ -781,3 +873,142 @@ func (me *Metrics) RecordAdsCertReq(success bool) {
...
+
+func (me *Metrics) RecordModuleFailed(labels ModuleLabels) {
+	mm, err := me.getModuleMetric(labels)
+	if err != nil {
+		return
+	}
+
+	// Module metrics
+	mm.FailureCounter.Inc(1)
+	// Account-Module metrics
```
AlexBVolcy, line 917: Same line break needed

metrics/go_metrics.go
```diff
@@ -781,3 +873,142 @@ func (me *Metrics) RecordAdsCertReq(success bool) {
...
+
+func (me *Metrics) RecordModuleSuccessNooped(labels ModuleLabels) {
+	mm, err := me.getModuleMetric(labels)
+	if err != nil {
+		return
+	}
+
+	// Module metrics
+	mm.SuccessNoopCounter.Inc(1)
+	// Account-Module metrics
```
AlexBVolcy, line 933: Same line break needed

AlexBVolcy (commented):
Minor comments related to variable re-naming and small changes

metrics/metrics.go
legendko: renamed to `AccountID`

hooks/hookexecution/execution.go
legendko: makes sense, updated

metrics/prometheus/preload.go
legendko: I agree with you. I made that all possible stages are added for every module, but that's unnecessary. Fixed

modules/helpers.go
legendko: Done

legendko:
> Minor comments related to variable re-naming and small changes

Updated

modules/modules.go
```diff
@@ -50,18 +52,25 @@ func (m *builder) Build(cfg config.Modules, client *http.Client) (hooks.HookRepo
...
-				return nil, fmt.Errorf(`failed to init "%s" module: %s`, id, err)
+				return nil, nil, fmt.Errorf(`failed to init "%s" module: %s`, id, err)
 			}
 
 			modules[id] = module
 		}
 	}
 
-	return hooks.NewHookRepository(modules)
+	coll, err := createModuleStageNamesCollection(modules)
```
AlexBVolcy, line 68: Nitpick: Could you re-name `coll` to be more descriptive

AlexBVolcy (commented):
Looks good, almost ready to approve, just left one small comment

metrics/go_metrics.go
```diff
@@ -383,6 +444,30 @@ func registerAdapterMetrics(registry metrics.Registry, adapterOrAccount string,
...
+	}
+}
+
+func registerAccountModuleMetrics(registry metrics.Registry, id string, module string, mm *ModuleMetrics) {
+	mm.DurationTimer = metrics.GetOrRegisterTimer(fmt.Sprintf("account.%s.modules.module.%s.duration", id, module), registry)
+	mm.CallCounter = metrics.GetOrRegisterCounter(fmt.Sprintf("account.%s.modules.module.%s.call", id, module), registry)
+	mm.FailureCounter = metrics.GetOrRegisterCounter(fmt.Sprintf("account.%s.modules.module.%s.failure", id, module), registry)
+	mm.SuccessNoopCounter = metrics.GetOrRegisterCounter(fmt.Sprintf("account.%s.modules.module.%s.success.noop", id, module), registry)
+	mm.SuccessUpdateCounter = metrics.GetOrRegisterCounter(fmt.Sprintf("account.%s.modules.module.%s.success.update", id, module), registry)
+	mm.SuccessRejectCounter = metrics.GetOrRegisterCounter(fmt.Sprintf("account.%s.modules.module.%s.reject", id, module), registry)
```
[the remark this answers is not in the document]
SyntaxNode: Good catch. It should be `success.reject`. I can see how `success.reject` looks strange, but it makes sense in the module ran successfully and then decided to reject the request.

modules/modules.go
legendko: done

metrics/go_metrics.go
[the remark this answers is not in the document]
legendko: eagle eye :)

legendko:
Latest updates:
1) Removed metrics gathering for `unknown` accounts
2) Added condition to determine whether the module was really successful
3) Fixes for your comments

hooks/hookexecution/execution.go
```diff
@@ -264,6 +296,16 @@ func handleHookMutations[P any](payload P, hr hookResponse[P], hookOutcome *Hook
...
+	}
+
+	// if at least one mutation from a given module was successfully applied
+	// we consider that the module was processed successfully
+	if successfulMutations > 0 {
+		metricEngine.RecordModuleSuccessUpdated(labels)
+	} else {
+		hookOutcome.Status = StatusExecutionFailure
+		metricEngine.RecordModuleExecutionError(labels)
 	}
```
bsardo, line 309: I don't see anything in the PRD about making this determination. Is this also PBS-Java's behavior?

bsardo: I guess this makes sense since this logic will only execute if there's at least one mutation and if at least one mutation is successfully applied we consider that an update. If there were also failed updates, they would be wrapped up in the hook outcome `Warnings`.

legendko: No, this is our own assumption based on our previous discussions on the topic what should we do if a module can not apply a mutation because another module deleted some property. We discussed that the client is responsible for a correct module configuration and should check by himself that the modules do not operate with the same fields. In case such event occurs we discussed that we should simply log a warning. But we realized that we could have a rare occasion when a module has only 1 mutation and it can not be applied (or N mutations and all of them can not be applied), then it seems strange that we send a successful metric and record a successful status for that module. If you think we should revert the changes or it should have another behavior please correct us.

AlexBVolcy (approved):
LGTM

PR #2485, opened 2022-12-07
Improve Digital adapter: adding publisher id as endpoint param

jbartek25:


adapters/improvedigital/improvedigital.go
```diff
@@ -285,3 +293,15 @@ func getImpExtWithRewardedInventory(imp openrtb2.Imp) ([]byte, error) {
 
 	return nil, nil
 }
+
+func (a *ImprovedigitalAdapter) buildEndpointURL(imp openrtb2.Imp) string {
+	publisherEndpoint := ""
+	var impBidder ImpExtBidder
+
+	err := json.Unmarshal(imp.Ext, &impBidder)
```
hhhjort, line 301: It may be more efficient to return the `bidder` entry from the ext in `getImpExtWithRewardedInventory()`, and passing that on to this function. This will save some work compared to unmarshaling the entire imp.ext from scratch.

samiul-shanto: `getImpExtWithRewardedInventory()` is an optional feature, whereas `buildEndpointURL()` is mandatory for all requests to generate a Publisher endpoint. `buildEndpointURL()` shouldn't be dependable on `getImpExtWithRewardedInventory()` return data. 

And getImpExtWithRewardedInventory() usage for un-structure data set, that's why we use map (`make(map[string]json.RawMessage)`) instead of struct. but, for buildEndpointURL() have a struct. 

isn't possible to use the same unmarshal data for both `getImpExtWithRewardedInventory` and `buildEndpointURL`? Yes, but it will be more complex to use the exact data for two different features which require two different data sets,

hhhjort (approved):
LGTM

static/bidder-info/improvedigital.yaml
```diff
@@ -1,4 +1,4 @@
-endpoint: "http://ad.360yield.com/pbs"
+endpoint: "http://ad.360yield.com/{PublisherId}pbs"
```
mansinahar, line 1: Quick question: Why not add the `/` as part of the URL itself rather than appending it in the `buildEndpointURL` method

samiul-shanto: When no publisher ID is defined, URL should be `http://ad.360yield.com/pbs`. So, if I add `/` as part of URL, the URL would be `http://ad.360yield.com//pbs`.

mansinahar (approved):

PR #2488, opened 2022-12-08
Outbrain adapter: add video support

markkuhar:


mansinahar (commented):
LGTM. Can you please also update the docs [here](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/outbrain.md) with the added video support

markkuhar:
Thanks for review. For when is the next release planned?

bsardo:
@markkuhar this should make it into the next release, 0.238.0, which is tentatively planned for early next week.

mansinahar (approved):

PR #2489, opened 2022-12-09
bliink: bidder parameters added

LouisBillaut:
Bliink bidder parameters added

### DIFF
--- a/static/bidder-params/bliink.json
+++ b/static/bidder-params/bliink.json
@@ -13,5 +13,6 @@
       "type": "string",
       "description": "The image url on which the ad is displayed in case of in-image ad"
     }
-  }
+  },
+  "required": ["tagId"]
 }
--- a/static/bidder-params/bliink.json
+++ b/static/bidder-params/bliink.json
@@ -4,5 +4,14 @@
   "description": "A schema which validates params accepted by the Bliink adapter",
   "type": "object",
 
-  "properties": {}
+  "properties": {
+    "tagId": {
+      "type": "string",
+      "description": "The TagID from BLIINK"
+    },
+    "imageUrl": {
+      "type": "string",
+      "description": "The image url on which the ad is displayed in case of in-image ad"
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


LouisBillaut:
Hi, I'm leaving a comment because I've opened this PR 2 weeks ago.
Did I forgot something in it ? 
Thanks for your time 😊

dnsdrs (dismissed):

static/bidder-params/bliink.json
```diff
@@ -4,5 +4,14 @@
   "description": "A schema which validates params accepted by the Bliink adapter",
   "type": "object",
 
-  "properties": {}
+  "properties": {
```
hhhjort, line 7: This JSON schema documents that two possible properties exist, and if they do their values must be strings. But they do not need to be present in a request. If this is correct, then all is good. If there are further requirements, we should probably discuss how best to get them in.

LouisBillaut:
The field TagId is required, but not the imageUrl field.

static/bidder-params/bliink.json
LouisBillaut: The field TagId is required, but not the imageUrl field.

hhhjort: I would suggest some code in your adapter to check for the existence of TagId in the request, and return an error if it is not found. Then, at a later date put in a PR that sets TagId to required in the JSON schema. PBS made a design choice that if any of the adapter parameters fails the schema check, the whole request fails.

The reasoning for this strictness was the adapter parameter schema will rarely change, if ever. However a publisher with a typo in one of the adapter's parameter may not notice anything amiss if at least some prebid adapters are returning ads. So failing completely is more likely to get the publisher's attention to fix the typo.

Once enough time has passed for publishers to fix their prebid configs for your new rrquirements, you can set your schema to flag TagId as required and have the full benefit of the schema enforcement.

LouisBillaut: Done.
I directly set the field "tagId" to required because for the moment, we don't have any publishers using our adapter.

hhhjort (approved):
LGTM

AlexBVolcy (approved):
LGTM

PR #2491, opened 2022-12-09
Modularity: module builders generator

4lexvav:
Usage:
- run from the root directory `go generate modules/modules.go` before building the app
- or just `make build-modules`


modules/generator/builder.tmpl
```diff
@@ -0,0 +1,19 @@
+package modules
+
+import (
+    {{- range .}}
+    {{.Vendor}}{{.Module | Title}} "github.com/prebid/prebid-server/modules/{{.Vendor}}/{{.Module}}"
+    {{- end}}
+)
```
SyntaxNode, line 7: Please wrap the import statement around an if block to allow for the case of no modules:

```
{{if .}}
import (
    {{- range .}}
    {{.Vendor}}{{.Module | Title}} "github.com/prebid/prebid-server/modules/{{.Vendor}}/{{.Module}}"
    {{- end}}
)
{{end}}
```

modules/generator/buildergen.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("failed to create %s file: %s", outName, err))
+	}
+	defer f.Close()
+
+	if err = t.Execute(f, modules); err != nil {
+		panic(fmt.Sprintf("failed to generate %s file content: %s", outName, err))
+	}
+
+	fmt.Printf("%s file successfully generated\n", outName)
```
SyntaxNode, line 62: Please run gofmt after code generation to ensure proper formatting. This is especially important for the case of no modules to collapse the curly braces. I tested it locally:

```
b := &bytes.Buffer{}
if err = t.Execute(b, modules); err != nil {
	panic(fmt.Sprintf("failed to generate %s file content: %s", outName, err))
}

bf := b.Bytes()
if bf, err = format.Source(bf); err != nil {
	panic(fmt.Sprintf("failed to generate %s file content: %s", outName, err))
}

if err = os.WriteFile(outName, bf, 0666); err != nil {
	panic(fmt.Sprintf("failed to create %s file: %s", outName, err))
}

fmt.Printf("%s file successfully generated\n", outName)
```

PR #2495, opened 2022-12-15
Modularity: Raw auction stage implementation

legendko:
Account retrieval was moved to earlier phase of request processing.
We load the stored requests and if we actually have one we try to get an account from there, otherwise we search the account in the original request. After the raw auction hook execution we retrieve the stored data once more as related request data could be changed by the hook. Finally, we merge the stored request data.


endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1934,67 +1935,68 @@ func TestSetTargeting(t *testing.T) {
...
-				openrtb_ext.BuildBidderMap(),
-				nil,
-				nil,
-				hardcodedResponseIPValidator{response: true},
-				empty_fetcher.EmptyFetcher{},
-				test.hookExecutor,
-			}
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			test := testCase{}
```
hhhjort, line 1973: I don't understand why you are creating a second test case for use with the current test case. Also you are unmarshalling `fileData` multiple times, it seems excessive.

legendko: Most likely we overlooked to remove it after refactoring or after some merge conflict, thanks

endpoints/openrtb2/auction.go
```diff
@@ -179,26 +190,12 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 	}
 
 	usersyncs := usersync.ParseCookieFromRequest(r, &(deps.cfg.HostCookie))
-	if req.App != nil {
-		labels.Source = metrics.DemandApp
-		labels.RType = metrics.ReqTypeORTB2App
-		labels.PubID = getAccountID(req.App.Publisher)
-	} else { //req.Site != nil
-		labels.Source = metrics.DemandWeb
+	if req.App == nil {
```
SyntaxNode, line 193: Might read a little better if this is `req.Site != nil`, We'd need to make this change anyway in the near future when we add DOOH support.

endpoints/openrtb2/auction.go
```diff
@@ -342,18 +339,63 @@ func (deps *endpointDeps) parseRequest(httpRequest *http.Request) (req *openrtb_
...
+	if rejectErr != nil {
+		errs = []error{rejectErr}
+		if err = json.Unmarshal(requestJson, req.BidRequest); err != nil {
+			glog.Errorf("Failed to unmarshal BidRequest during raw auction stage rejection: %s", err)
+		}
+		return
+	}
+
+	// retrieve storedRequests and storedImps once more in case stored data was changed by the raw auction hook
+	impInfo, errs = parseImpInfo(requestJson)
```
SyntaxNode, line 381: Do we have a way to know if any raw auction stage hooks actually ran? If not, there is no need to do this and we can save on CPU.

hhhjort (approved):
LGTM

VeronikaSolovei9 (approved):

PR #2496, opened 2022-12-16
Update to support reading GDPR and US Privacy from GPP string

hhhjort:


exchange/gdpr.go
```diff
@@ -11,6 +13,15 @@ import (
...
 	var err error
+
+	if bidRequest.Regs != nil && len(bidRequest.Regs.GPPSID) > 0 {
+		for _, id := range bidRequest.Regs.GPPSID {
+			if id == int8(gppConstants.SectionTCFEU2) {
+				return gdpr.SignalYes, nil
+			}
+		}
+		return gdpr.SignalNo, nil
+	}
```
SyntaxNode, line 24: There be dragons.

Should a non-nil GPP SID array be treated the same as a missing one? If there is an array without any elements, might this be interpreted as an explicit `SignalNo`? The len check on line 17 would cause this to fall through and perhaps land as `SignalAmbiguous`.

exchange/gdpr.go
```diff
@@ -21,7 +32,13 @@ func extractGDPR(bidRequest *openrtb2.BidRequest) (gdpr.Signal, error) {
...
 
 // ExtractConsent will pull the consent string from an openrtb request
-func extractConsent(bidRequest *openrtb2.BidRequest) (consent string, err error) {
+func extractConsent(bidRequest *openrtb2.BidRequest, gpp gpplib.GppContainer) (consent string, err error) {
+	for i, id := range gpp.SectionTypes {
+		if id == gppConstants.SectionTCFEU2 {
+			consent = gpp.Sections[i].GetValue()
+			return
+		}
+	}
```
SyntaxNode, line 41: FYI: I noticed this pattern appears a lot. Consider adding a new helper to the GPP library in the iteration, something like:

```
func (g GppContainer) GetSection(c constants.SectionID) (Section, bool) {
..
```

Where the last bool indicates if the section was found.

We could decide to optimize this lookup with a map as an implementation detail of `GetSection`.

exchange/gdpr.go
```diff
@@ -3,6 +3,8 @@ package exchange
 import (
 	"encoding/json"
 
+	gpplib "github.com/prebid/go-gpp"
+	gppConstants "github.com/prebid/go-gpp/constants"
```
SyntaxNode, line 7: FYI: In the next GPP lib iteration, I wonder if the constants package is really necessary. Perhaps that should be brough up into the gpp level.

exchange/gdpr_test.go
```diff
@@ -47,28 +49,56 @@ func TestExtractGDPR(t *testing.T) {
...
+			giveRegs:    &openrtb2.Regs{GPPSID: []int8{6}},
+			wantGDPR:    gdpr.SignalNo,
+		},
 	}
 
 	for _, tt := range tests {
-		bidReq := openrtb2.BidRequest{
-			Regs: tt.giveRegs,
-		}
+		t.Run(tt.description, func(t *testing.T) {
```
SyntaxNode, line 80: Nice.

exchange/gdpr_test.go
```diff
@@ -47,28 +49,56 @@ func TestExtractGDPR(t *testing.T) {
...
+			bidReq := openrtb2.BidRequest{
+				Regs: tt.giveRegs,
+			}
 
-		result, err := extractGDPR(&bidReq)
-		assert.Equal(t, tt.wantGDPR, result, tt.description)
+			result, err := extractGDPR(&bidReq)
+			assert.Equal(t, tt.wantGDPR, result, tt.description)
```
SyntaxNode, line 86: There is no longer a need to include the descriptions as arguments to the assert methods now that the test is contained in a t.Run call.

endpoints/openrtb2/auction.go
```diff
@@ -645,7 +646,17 @@ func (deps *endpointDeps) validateRequest(req *openrtb_ext.RequestWrapper, isAmp
...
 
-	if ccpaPolicy, err := ccpa.ReadFromRequestWrapper(req); err != nil {
+	var gpp gpplib.GppContainer
+	if req.BidRequest.Regs != nil && len(req.BidRequest.Regs.GPP) > 0 {
+		gpp, err = gpplib.Parse(req.BidRequest.Regs.GPP)
+		if err != nil {
+			errL = append(errL, &errortypes.Warning{
+				Message:     fmt.Sprintf("GPP consent string is invalid and will be ignored. (%v)", err),
+				WarningCode: errortypes.InvalidPrivacyConsentWarningCode})
+		}
```
SyntaxNode, line 656: To match the CCPA and TCF behavior, we should set `req.BidRequest.Regs.GPP` to empty if the string is invalid. Right now, it's not so much ignored.

privacy/ccpa/policy.go
```diff
@@ -14,21 +16,30 @@ type Policy struct {
...
-func ReadFromRequestWrapper(req *openrtb_ext.RequestWrapper) (Policy, error) {
+func ReadFromRequestWrapper(req *openrtb_ext.RequestWrapper, gpp gpplib.GppContainer) (Policy, error) {
 	var consent string
 	var noSaleBidders []string
 
-	if req == nil {
-		return Policy{}, nil
+	for i, id := range gpp.SectionTypes {
+		if id == gppConstants.SectionUSPV1 {
+			//if id == 6 {
```
SyntaxNode, line 25: Left over debugging comment?

privacy/ccpa/policy.go
```diff
@@ -14,21 +16,30 @@ type Policy struct {
 }
 
 // ReadFromRequestWrapper extracts the CCPA regulatory information from an OpenRTB bid request.
-func ReadFromRequestWrapper(req *openrtb_ext.RequestWrapper) (Policy, error) {
+func ReadFromRequestWrapper(req *openrtb_ext.RequestWrapper, gpp gpplib.GppContainer) (Policy, error) {
```
SyntaxNode, line 19: Note For Future: We should move the gpp container into the request wrapper.

privacy/ccpa/policy.go
```diff
@@ -44,7 +55,17 @@ func ReadFromRequestWrapper(req *openrtb_ext.RequestWrapper) (Policy, error) {
...
 
 func ReadFromRequest(req *openrtb2.BidRequest) (Policy, error) {
-	return ReadFromRequestWrapper(&openrtb_ext.RequestWrapper{BidRequest: req})
+	var gpp gpplib.GppContainer
+	var err error
+	if req.Regs != nil && len(req.Regs.GPP) > 0 {
+		gpp, err = gpplib.Parse(req.Regs.GPP)
+		if err != nil {
+			// zero out bad container
+			gpp = gpplib.GppContainer{}
```
SyntaxNode, line 64: This behavior isn't present in the auction.go code. Should it be? .. or would it be good to ensure that gpp is always empty if a Parse fails? Should it be nil? Just looking for consistency with the two call sites.

privacy/ccpa/policy.go
```diff
@@ -44,7 +55,17 @@ func ReadFromRequestWrapper(req *openrtb_ext.RequestWrapper) (Policy, error) {
 }
 
 func ReadFromRequest(req *openrtb2.BidRequest) (Policy, error) {
```
bsardo, line 57: Can we add some tests for this function now that it is no longer a one-liner?

privacy/ccpa/policy_test.go
```diff
@@ -153,13 +156,38 @@ func TestReadFromRequest(t *testing.T) {
...
+			request: &openrtb2.BidRequest{
+				Regs: &openrtb2.Regs{Ext: json.RawMessage(`{"us_privacy":"ABC"}`)},
+				Ext:  json.RawMessage(`{"prebid":{"nosale":["a", "b"]}}`),
+			},
+			giveGPP: gpplib.GppContainer{Version: 1, SectionTypes: []gppConstants.SectionID{6}, Sections: []gpplib.Section{&upsv1Section}},
+			expectedPolicy: Policy{
+				Consent:       "1YNY",
+				NoSaleBidders: []string{"a", "b"},
+			},
+		},
```
bsardo, line 181: Can we add a test case here for when a different GPP section is set ensuring that the consent object has a consent string of empty string and no error is detected?

exchange/gdpr.go
hhhjort: That has been on my mind. If any of the sections need access to the constants, it needs to be separate. I figured it was easier to merge it later than split it later once the need, or lack thereof, was clearer.

privacy/ccpa/policy.go
hhhjort: Left over from when `SectionUSPV1` was missing from the constants. So not exactly debugging, but needs to go.

exchange/gdpr.go
hhhjort: GGPSID is a slice, not a pointer. Is a nil value different from a zero length slice?

exchange/gdpr.go
hhhjort: Agreed

endpoints/openrtb2/auction.go
hhhjort: Where do we set the TCF string to empty if it is invalid?

privacy/ccpa/policy.go
hhhjort: The container as defined there is a struct rather than a pointer, so cannot be nil. It is a valid point about being inconsistent, I think I just wanted to not ignore an error if it appeared. We may still want to empty out the container if the parse fails in the gpp library itself.

exchange/gdpr.go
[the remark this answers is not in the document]
hhhjort: That is correct. Scott did have a good suggestion for adding a dictionary to optimize this lookup inside the container itself.

privacy/ccpa/policy_test.go
hhhjort: Will do

privacy/ccpa/policy.go
hhhjort: Will need to construct some GPP strings to enable that.

endpoints/openrtb2/auction.go
bsardo: We aren't. Should we? I guess it makes sense and would be easier for bidders to digest if they care to inspect it but is there a possibility they would like access to the original value?
I don't see anything in the spec on this subject.

exchange/gdpr.go
bsardo: I would think that for at least as long as we're supporting `regs.ext` as a fallback, the absence of `GPPSID` would be handled differently from a zero length `GPPSID` with the former being considered an ambiguous signal and the latter being an explicit no. If it's ambiguous, we then look in `regs.ext`.
Thoughts?
I poked around the prebid spec and the IAB spec but I don't see any clear direction on this.

SyntaxNode: Yes, GGPSID is a slice. While slices are not technically pointers, their main purpose is to point to a backing array. Therefore, I suppose, if a slice has a nil backing array it is considered nil and `slice == nil` evaluates to `true`.

The problem is I don't know how we should handle the case where a GPP consent string is provided but GPPSID is nil.

exchange/gdpr.go
SyntaxNode: Ok, cool.

endpoints/openrtb2/auction.go
SyntaxNode: I thought we treated invalid GDPR strings the same as we do for CCPA. I'm fine with matching GDPR behavior. We can consider this separately.

privacy/ccpa/policy.go
SyntaxNode: Gotcha, Works for me at this stage. We can reconsider in Phase 2.

bsardo (commented):
LGTM. I just need clarification on how to handle the scenario described in this thread: https://github.com/prebid/prebid-server/pull/2496#discussion_r1052451403

bsardo (approved):

exchange/utils.go
```diff
@@ -177,9 +187,9 @@ func ccpaEnabled(account *config.Account, privacyConfig config.Privacy, requestT
 	return privacyConfig.CCPA.Enforce
 }
 
-func extractCCPA(orig *openrtb2.BidRequest, privacyConfig config.Privacy, account *config.Account, aliases map[string]string, requestType config.ChannelType) (privacy.PolicyEnforcer, error) {
+func extractCCPA(orig *openrtb2.BidRequest, privacyConfig config.Privacy, account *config.Account, aliases map[string]string, requestType config.ChannelType, gpp gpplib.GppContainer) (privacy.PolicyEnforcer, error) {
 	// Quick extra wrapper until RequestWrapper makes its way into CleanRequests
-	ccpaPolicy, err := ccpa.ReadFromRequestWrapper(&openrtb_ext.RequestWrapper{BidRequest: orig})
+	ccpaPolicy, err := ccpa.ReadFromRequestWrapper(&openrtb_ext.RequestWrapper{BidRequest: orig}, gpp)
```
SyntaxNode, line 192: I think we need to account for the GPPSID. If the SID is missing the CCPA section identifier, I believe we should return the `NilPolicyEnforcer`.

bsardo: Shouldn't the behavior be the same as for GDPR where if the SID is missing the relevant section identifier we fall back to checking the old regs field?
I thought that behavior was implemented for CCPA in `ReadFromRequestWrapper`. If the SID isn't found, the consent string is empty and a `Policy` object is created that may or may not contain a consent string depending on whether it was successfully read from the old regs location and that consent string will determine whether or not we should enforce.

hhhjort: This was my thinking as well. Take it from the GPP framework if it exists, and if not, fall back on old behavior. We could revisit in January or later if we should upgrade the old logic for the reality of the GPP framework, but it seems best short term to preserve the old behavior as much as possible as the GPP starts to roll out to the world.

bsardo: Ah my mistake. @SyntaxNode and I just discussed this.
I was looking at this and thinking that was a SID check but it is not:
```
for i, id := range gpp.SectionTypes {
	if id == gppConstants.SectionUSPV1 {
		consent = gpp.Sections[i].GetValue()
	}
}
```

SyntaxNode (approved):
LGTM. Thank you for the test enhancements.

bsardo (approved):

PR #2500, opened 2022-12-20
Modularity: All processed bid responses stage implementation

legendko:
During the implementation of All Processed Bid Responses Stage we faced the issue of cycle import because the hook needs a map of exchange.pbsOrtbSeatBids as parameter in hookexecution.hookExecutor function. But we already import the hookexecution package in exchange package in order to call the hooks.
There were 2 solutions for this issue:

1. Move the exchange.pbsOrtbSeatBid and related exchange.pbsOrtbBid to a new package imported both by exchange and hookexecution which resulted in exporting these entities and their fields. The fields would be exported anyway in order to be able to use them in the modules.
2. Write 2 interfaces for exchange.pbsOrtbSeatBid and exchange.pbsOrtbBid respectively and use them in the hook call functions but this approach had some pitfalls. Since we need to pass a map of exchange.pbsOrtbSeatBids to the hook call function, we can not pass it directly as a map of structs which implement our interface, because they do not have the same representation in memory. This would require to copy all values from map[openrtb_ext.BidderName]pbsOrtbSeatBid to a map[openrtb_ext.BidderName]PbsOrtbSeatBidInterface in order to be able to pass them as an argument and use inside the hook execution function. Moreover each pbsOrtbSeatBid has a slice of pbsOrtbBids which will require the same copy operation for all of its elements because of the same reason.

We decided to use the first approach as a much simpler and more effective solution without unnecessary copy operations.


legendko:
**One more important change:**
We changed the location of current hook call from the agreed https://github.com/prebid/prebid-server/blob/master/exchange/exchange.go#L324 to https://github.com/prebid/prebid-server/blob/master/exchange/exchange.go#L312
The main reason is to put the hook call before category mapping. In PBS-Java they have such order. If you think PBS-Go has to act differently please tell us.

legendko:
> > **One more important change:** We changed the location of current hook call from the agreed https://github.com/prebid/prebid-server/blob/master/exchange/exchange.go#L324 to https://github.com/prebid/prebid-server/blob/master/exchange/exchange.go#L312 The main reason is to put the hook call before category mapping. In PBS-Java they have such order. If you think PBS-Go has to act differently please tell us.
> 
> UPD: this change also influenced on the order of bid Id generation and vast generation. Now they are called after the hook. We should decide which order should be used for these processes.

Copying my comment from slack, so it does not get lost in the threads:

At the beginning when we were planning the locations for each hook we decided to call the all_processed_bid_responses hook after bid Id generation and vast generation because PBS-Java has a similar flow. But at that time we did not took into account the process of category mapping which in Java is called after the hook and we had it before the hook because of the current implementation. A few days ago we noticed this difference but forgot about the original reasons of the primal location and moved the `all_processed_bid_responses` hook before category mapping like in PBS-Java, but this resulted in changing the order for bid Id generation and vast generation as well.
Is category mapping critical for all_processed_bid_responses hook? I did not find any information about it in the PRD. If this hook is not intended to change that information, I suggest to move back the location of the hook as we agreed on the very beginning: after bid id generation and vast generation. If category mapping can be changed during this stage then the only way is to refactor the existing func at https://github.com/prebid/prebid-server/blob/master/exchange/exchange.go#L742 so category mapping is called after the hook, but bid Id generation and vast generation before.
To simplify the understanding I'll write a small schema

```
PBS-Java:
bid Id generation and vast generation => Hook => Category mapping

Original agreement:
Category mapping => bid Id generation and vast generation => Hook

Updated code:
Hook => Category mapping => bid Id generation and vast generation
```

hooks/hookexecution/mocks_test.go
```diff
@@ -84,10 +85,14 @@ func (e mockRejectHook) HandleProcessedAuctionHook(_ context.Context, _ hookstag
...
 
+func (e mockRejectHook) HandleAllProcessedBidResponsesHook(_ context.Context, _ hookstage.ModuleInvocationContext, _ hookstage.AllProcessedBidResponsesPayload) (hookstage.HookResult[hookstage.AllProcessedBidResponsesPayload], error) {
+	return hookstage.HookResult[hookstage.AllProcessedBidResponsesPayload]{Reject: true}, nil
+}
+
 type mockTimeoutHook struct{}
 
 func (e mockTimeoutHook) HandleEntrypointHook(_ context.Context, _ hookstage.ModuleInvocationContext, _ hookstage.EntrypointPayload) (hookstage.HookResult[hookstage.EntrypointPayload], error) {
-	time.Sleep(2 * time.Millisecond)
+	time.Sleep(3 * time.Millisecond)
```
AlexBVolcy, line 95: Could you explain why this value was changed from 2 to 3?

AlexBVolcy (commented):
This looks good, have one comment before approval.

hooks/hookexecution/mocks_test.go
legendko: I guess, it was just a typo. Reverted to 2 milliseconds.

bsardo:
@legendko, the `applyCategoryMapping` function may remove some bids but given that the intent of this hook is for creative malware scans, it is ok to run this function before executing the hook since there is no reason to perform such a scan on a bid that will be removed anyway. As such, I think we should go with the original agreement:
`Category mapping => bid Id generation and vast generation => Hook`

bsardo:
@legendko I think you mentioned the original plan was to put it after bid generation and VAST generation as well so maybe move it down a few lines just after this:
```
evTracking := getEventTracking(&requestExt.Prebid, r.StartTime, &r.Account, e.bidderInfo, e.externalURL)
adapterBids = evTracking.modifyBidsForEvents(adapterBids)
```

legendko:
> @legendko I think you mentioned the original plan was to put it after bid generation and VAST generation as well so maybe move it down a few lines just after this:
> 
> ```
> evTracking := getEventTracking(&requestExt.Prebid, r.StartTime, &r.Account, e.bidderInfo, e.externalURL)
> adapterBids = evTracking.modifyBidsForEvents(adapterBids)
> ```

@bsardo sorry, looked and inserted in the wrong place.

bsardo:
No worries. Thanks!

AlexBVolcy (approved):
LGTM! Thank you for being so prompt with the changes to when the hook is executed

hooks/hookexecution/mocks_test.go
```diff
@@ -287,3 +315,26 @@ func (e mockUpdateBidderResponseHook) HandleRawBidderResponseHook(_ context.Cont
...
+
+func (e mockFailedMutationHook) HandleAllProcessedBidResponsesHook(_ context.Context, _ hookstage.ModuleInvocationContext, _ hookstage.AllProcessedBidResponsesPayload) (hookstage.HookResult[hookstage.AllProcessedBidResponsesPayload], error) {
+	changeSet := &hookstage.ChangeSet[hookstage.AllProcessedBidResponsesPayload]{}
+	changeSet.AddMutation(func(payload hookstage.AllProcessedBidResponsesPayload) (hookstage.AllProcessedBidResponsesPayload, error) {
+		return payload, errors.New("key not found")
+	}, hookstage.MutationUpdate, "some-bidder", "bids[0]", "deal_priority")
+
+	return hookstage.HookResult[hookstage.AllProcessedBidResponsesPayload]{ChangeSet: changeSet}, nil
+}
+
```
bsardo, line 327: Nitpick: can you move this up so it is near the other `mockFailedMutationHook` functions just above line 305 `type mockUpdateBidderResponseHook struct{}`?

exchange/exchange_test.go
```diff
@@ -4539,9 +4540,9 @@ type capturingRequestBidder struct {
 	req *openrtb2.BidRequest
 }
 
-func (b *capturingRequestBidder) requestBid(ctx context.Context, bidderRequest BidderRequest, conversions currency.Conversions, reqInfo *adapters.ExtraRequestInfo, adsCertSigner adscert.Signer, bidRequestMetadata bidRequestOptions, alternateBidderCodes openrtb_ext.ExtAlternateBidderCodes, executor hookexecution.StageExecutor) (seatBid []*pbsOrtbSeatBid, errs []error) {
```
bsardo, line 4542: Can you re-add the `executor hookexecution.StageExecutor` parameter so `capturingRequestBidder` satisfies the `AdaptedBidder` interface?

bsardo (commented):
Looks good, just two minor comments.

bsardo (approved):

AlexBVolcy (approved):
LGTM!

PR #2501, opened 2022-12-21
VISX: Add string support to UID

deubaka:
Updating UID to support both integer and string.

- Updated schema to support string and integer types
- Added regex to schema for UID format validation
- Added new test params

### DIFF
--- a/adapters/visx/params_test.go
+++ b/adapters/visx/params_test.go
@@ -2,8 +2,9 @@ package visx
 
 import (
 	"encoding/json"
-	"github.com/prebid/prebid-server/openrtb_ext"
 	"testing"
+
+	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
 func TestValidParams(t *testing.T) {
@@ -34,6 +35,7 @@ func TestInvalidParams(t *testing.T) {
 
 var validParams = []string{
 	`{"uid":13245}`,
+	`{"uid":"13245"}`,
 	`{"uid":13245, "size": [10,5]}`,
 	`{"uid":13245, "other_optional": true}`,
 }
--- a/adapters/visx/params_test.go
+++ b/adapters/visx/params_test.go
@@ -47,7 +47,10 @@ var invalidParams = []string{
 	`2`,
 	`{"size":12345678}`,
 	`{"size":""}`,
-	`{"uid": "1"}`,
+	`{"uid": "-1"}`,
+	`{"uid": "232af"}`,
+	`{"uid": "af213"}`,
+	`{"uid": "af"}`,
 	`{"size": true}`,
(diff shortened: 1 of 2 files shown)


adapters/visx/params_test.go
```diff
@@ -47,7 +47,10 @@ var invalidParams = []string{
 	`2`,
 	`{"size":12345678}`,
 	`{"size":""}`,
-	`{"uid": "1"}`,
+	`{"uid": "-1"}`,
+	`{"uid": "232af"}`,
+	`{"uid": "af213"}`,
+	`{"uid": "af"}`,
```
guscarreon, line 53: Nitpicks: 
1. Can we add an `int` that comes with a `char` in here?
```
43   var invalidParams = []string{
44       `null`,
45       `nil`,
46       ``,
47       `[]`,
48       `true`,
49       `2`,
50       `{"size":12345678}`,
51       `{"size":""}`,
52       `{"uid": "-1"}`,
53       `{"uid": "232af"}`,
   +     `{"uid": 232af}`,
54       `{"uid": "af213"}`,
55       `{"uid": "af"}`,
56       `{"size": true}`,
57       `{"uid": true, "size":"1234567"}`,
58   }
```

2. And also add a `string` unit tests on the `validParams` side?
```
35   var validParams = []string{
36       `{"uid":13245}`,
   +     `{"uid":"13245"}`,
37       `{"uid":13245, "size": [10,5]}`,
38       `{"uid":13245, "other_optional": true}`,
39   }
```

deubaka: Hi @guscarreon,

Thanks for the feedback.

On #1, I'm not sure if it's still necessary given that case would be an invalid JSON all together, and parsing _should_ fail already, without reaching schema validation. 

As for #2, make sense, just to make sure it's covered as well.

hhhjort (approved):
LGTM

guscarreon (commented):
LGTM. @deubaka thank you for addressing our feedback

guscarreon (approved):

static/bidder-params/visx.json
```diff
@@ -5,7 +5,8 @@
   "type": "object",
   "properties": {
     "uid": {
-      "type": "integer",
+      "type": ["integer", "string"],
```
guscarreon, line 8: @deubaka can we reflect this change in the [bid-param section of the docs](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/visx.md#bid-params) please?

deubaka: Good catch, thanks! We've added the corresponding updates in this PR, https://github.com/prebid/prebid.github.io/pull/4273.

Issue #2367, opened 2022-09-02
Return All Bid Status


bretg:
Client-side analytics adapters don't have access to what actually happened with PBS bids unless they're successful.

The ORTB protocol was designed to carry just actual successful bids, not information about the status of all bid attempts. However, with the Prebid world being part client-side and part server-side, holisitic analytics is made more difficult. It's quite complicated to log auction info/results separately: some client-side, some server-side and then sew them all together at scale on the backend. 

e.g. say a request contains calls to bidderA, bidderB, and bidderC. If only bidderA responds, then the resulting seatbid[].bid[] contains only bidderA. The client doesn't know what happened to bidderB or bidderC: no bid? error? GDPR removal? Price floor removal?

So we would like to propose a new optional request flag that instructs PBS to include the status of all bid attempts in the response.

## Proposal
1) Support a new request flag ext.prebid.returnallbidstatus: true - this causes PBS to create special entries in the response for every imp+bidder in the request, even nobids and errors.

This is independent from the debug/test option, though it might be useful to assume this flag is true if debug is true?

2) When the flag is set and a bidder doesn't have a bid for any reason, create a 'non-bid' response just before "Stage 8 - Auction Response"
```
ext.prebid.seatnonbid: [{ - generally follows the syntax of seatbid
  seat: "bidderB"
  nonbid: [{
    impid is set to $.imp[].id
    no id field
    no price field
    ext.status.code is set to "nobid", "error", "removed"
      ext.status.details[].code is set to a code
      ext.status.details[].message is set to a string version of the code
  }]
}]
```
We considered just collating this data in seatbid.bid, but we can't be sure that that all clients will know what to do with the extra info: e.g. SDK, SSAI servers, etc. Sure, they shouldn't be passing ext.prebid.returnallbidstatus if they can't handle the response, but it seems safer to place the non-bids in a different location.

3) The bids list included should be the list of imps+bidders present just before Stage 3 - Processed Auction Request.

4) Some reasons a bidder might not have a bid response include:
- nobid
- bidder error
- GDPR removed the bidder
- bid removed due to pricefloor 
- a module removed the bid request before being sent out
- a module removed the bid response after the request
Each of these scenarios should be considered test cases.

bretg:
Here'a a proposed design. If the ext.prebid.returnallbidstatus flag is true, then as part of the ORTB response generation

```
Loop through the imps in the original request
    Loop through each bidder in imp (imp[n].ext.prebid.bidder.*
[...]

PR #2505, opened 2022-12-23
I2466 - Introducing seatnonbid support

ShriprasadM:
@SyntaxNode and @AlexBVolcy : Just want to put few updates there with latest changes

### Change in NonBidExt.Prebid.Bid struct

Old Struct was 
```go
type Prebid struct {
  Bid openrtb2.Bid  `json:"bid"`
}
```
New struct is : Limited exposure of all openrtb2.Bid. 

UPDATE :  Removed `openrtb2.Bid` instead introduced only required fields from `openrtb2.Bid` as per code review comments. Hence, updating the following structure

```go
// Bid is Wrapper around original/proxy bid object
type Bid struct {
  /* Removed openrtb2.Bid 
  openrtb2.Bid
  ID    string `json:"id,omitempty"`    // added omitempty
  ImpID string `json:"impid,omitempty"` // added omitempty */

  Price   float64                 `json:"price,omitempty"`
  ADomain []string                `json:"adomain,omitempty"`
  CatTax  adcom1.CategoryTaxonomy `json:"cattax,omitempty"`
  Cat     []string                `json:"cat,omitempty"`
  DealID  string                  `json:"dealid,omitempty"`
  W       int64                   `json:"w,omitempty"`
  H       int64                   `json:"h,omitempty"`
  Dur     int64                   `json:"dur,omitempty"`
  MType   openrtb2.MarkupType     `json:"mtype,omitempty"`

  OriginalBidCPM float64 `json:"originalbidcpm,omitempty"`
  OriginalBidCur string  `json:"originalbidcur,omitempty"`
}

type Prebid struct {
  Bid Bid `json:"bid"`
}
```
Sample output 
```json
{
	"ext": {
		"prebid": {
			"seatnonbid": [{
				"nonbid": [{
					"impid": "imp_1",
					"statuscode": 303,
					"ext": {
						"prebid": {
							"bid": {
								"price": 3,
								"adomain": ["ford.com"],
								"dealid": "456",
								"originalbidcpm": 3,
								"originalbidcur": "USD"
							}
						}
					}
				}],
				"seat": "appnexus",
				"ext": null
			}, {
				"nonbid": [{
					"impid": "imp_1",
					"statuscode": 303,
					"ext": {
						"prebid": {
							"bid": {
								"price": 3,
								"adomain": ["ford.com"],
								"dealid": "456",
								"originalbidcpm": 3,
								"originalbidcur": "USD"
							}
						}
					}
				}, {
					"impid": "imp_2",
					"statuscode": 303,
					"ext": {
						"prebid": {
							"bid": {
								"price": 10,
								"originalbidcpm": 10,
								"originalbidcur": "USD"
							}
						}
					}
				}],
				"seat": "pubmatic",
				"ext": null
			}]
		}
	}
}
```
2.  With this change, instead of putting entire openrtb2.bid object, we are only keeping following fields popoulated (here is the part of implementation captured)

```go
// SeatNonBid[].NonBid[].Ext.Prebid.Bid
Bid: openrtb2.Bid{. 
          Price:   bid.Bid.Price,   // here bid.Bid is entities.PbsOrtbBid
          ADomain: bid.Bid.ADomain,
          CatTax:  bid.Bid.CatTax,
          Cat:     bid.Bid.Cat,
          DealID:  bid.Bid.DealID,
          W:       bid.Bid.W,
          H:       bid.Bid.H,
          Dur:     bid.Bid.Dur,
          MType:   bid.Bid.MType,
        },
       OriginalBidCPM: bid.OriginalBidCPM,
       OriginalBidCur: bid.OriginalBidCur,
}
``` 
### Do we need any provisions of `returnallbidstatus` in case of `video` and `amp` endpoints ?


bsardo:
@ShriprasadM would you mind merging with master to resolve the conflicts? Thanks!

ShriprasadM:
> @ShriprasadM would you mind merging with master to resolve the conflicts? Thanks!

@bsardo : I have resolved the conflicts. Thanks

AlexBVolcy (commented):
This is looking good! Just left some thoughts that stuck out to me. And please merge with master when you get the chance to resolve the conflict.

openrtb_ext/response.go
```diff
@@ -95,3 +99,24 @@ const (
...
+// type NonBid struct {
+// 	openrtb2.Bid
+// 	StatusCode int
+// }
+
+// PROPOSAL2 - Introduce Bid Object
+type NonBid struct {
+	ImpId      string
+	StatusCode int
+	Bid        openrtb2.Bid
```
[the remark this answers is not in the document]
bretg: Comment 2367 originally had seatbid, but bid would be more correct. I suggest leaving it as nonbid.ext.prebid.bid.price. What's one more level between friends?

We're still cleaning up from the imp.ext.BIDDER mess...

analytics/core.go
```diff
@@ -93,3 +96,8 @@ type NotificationEvent struct {
 	Request *EventRequest   `json:"request"`
 	Account *config.Account `json:"account"`
 }
+
+// LogObject is common loggable object
+type LogObject struct {
+	SeatNonBid []openrtb_ext.SeatNonBid
+}
```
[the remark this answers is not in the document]
ShriprasadM: `LogObject` is introduced because of following use case
1. Exchange (exchange/exchange.go) suppose to update seat non-bid scenarios to Prebid Analytics Module
2. As per the current design (master) `exchange/exchange.go` has no access to `AuctionObject`, `AmpObject` or `VideoObject`. So, it will not be able to update unless these objects made available to it
3. Today,  `AuctionObject`, `AmpObject` or `VideoObject` are constructed by `auction.go`, `amp_auction.go` and `video_auction.go` respectively
4.  So, I have passed them over `AuctionRequest`. For this I have introduced `LogObject`, the more generalised form Of `AuctionObject`, `AmpObject` or `VideoObject here` [here](https://github.com/PubMatic-OpenWrap/prebid-server/blob/042398562aad7680056520a5ed6243a8062bd24e/exchange/exchange.go#L192) - . Otherwise, we will need to keep 3 different objects within `AuctionRequest`, which I think bad design.

openrtb_ext/response.go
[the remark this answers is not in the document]
ShriprasadM: @SyntaxNode / @AlexBVolcy / @bretg :
What is your thought on PROPOSAL1 mentioned above -  Extending NonBid with Bid Object itself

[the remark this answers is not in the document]
bretg: Will repeat what was said in issue 2367 and in the comment just above this one: yes, I'm ok with adding the bid object as ext.seatnonbid.nonbid.ext.prebid.bid.

analytics/core.go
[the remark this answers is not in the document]
AlexBVolcy: I see, this makes it more clear as to why `LogObject` could be valuable, but it also leaves me wondering if we still could just have `SeatNonBid` live in all of the places that `LogObject` currently occupies. I'm going to speak with @SyntaxNode about it, and get back to you!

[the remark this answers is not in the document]
ShriprasadM: @AlexBVolcy : Have you got a chance to check around it with @SyntaxNode ?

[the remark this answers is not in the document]
SyntaxNode: Yes. We reviewed the approach together and I understand your intention with introducing the LogObject. Thank you for sharing your use case, that was very useful context.

I agree with your second point that the exchange package has no access to the analytics package. This proposal changes that by adding a direct reference to analytics.LogObject. The exchange package (and HoldAuction in particular) is a weak point of PBS' design as it does too much work and takes too many parameters. The AuctionRequest struct was created as an attempt to curb the growth of HoldAuction, but even that has grow beyond comfort in the last year. For this reason, I ask you to consider a different approach to avoid entangling HoldAuction with more parts of the code.

Instead of passing the log object to HoldAuction, how about we enhance the return of HoldAuction to include the non bid information and handle sharing that with the analytics package in each endpoint? This would align well with future refactor plans, such that each endpoint constructs its own response for its own format. The AMP endpoint does extra work to form it's response which we can simplify in a future refactor, and sending back parts of the response as a strongly typed object would be both a great first step and avoid adding a dependency to the exchange.

What do you think?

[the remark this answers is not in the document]
ShriprasadM: Yes, I think, it would be good to return of HoldAuction to include the non bid information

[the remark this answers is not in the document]
ShriprasadM: Thanks to @viral for pointing this out. 
@SyntaxNode : 

**Please provide your views on inteoducing `ResponseWrapper`**

As per the thought, he have we can introduced `ResponseWrapper`, wherein `openrtb2.BidResponse` will be embedded. `ResponseWrapper` will contain the unmarshalled version of `response extension`. This will help us in sharing the `SeatNonBid` object to analytics, which has build by the exchange it self. Structure will look like as follows
```go
type ResponseWrapper struct {
	openrtb2.BidResponse
	responseExt *openrtb_ext.ExtBidResponse
}


func (rw *ResponseWrapper) GetSeatNonBid() []SeatNonBid {
	// return seatnonbid using responseExt
}

// Change HoldAuction signature to return ResponseWrapper instead of openrtb2.BidResponse
HoldAuction(ctx context.Context, r AuctionRequest, debugLog *DebugLog) (*openrb_ext.ResponseWrapper, error)
```
**Benefits of this approach**
1. No need to unmarshal again for obtaining seatnonbid
2. SeatNonBid will be shared, so no new object will be allocated on stack/heap
3. ResponseWrapper is extensible. 


**ReturnAllBidStatus** @SyntaxNode 
1. As per current PR's implementation, exchange.makeExtBidResponse is setting `SeatNonBid` within prebid extension only if   `ReturnAllBidStatus` is true.
2. With current approach, I strongly feel that this needs to move at `auction.go`, `amp_auction.go` and `video_auction.go` level. Otherwise, SeatNonBid can not be made available to Analytics if `ReturnAllBidStatus` is `false`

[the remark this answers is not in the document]
SyntaxNode: I like your thought process. I am most intrigued with your last arguments pointing to a "code smell" - a design flaw with the current implementation. The auction, amp, and video endpoints each have their own response structure but HoldAuction is built just for auction (the first endpoint we built). We use a hack for the amp and video endpoints where a lot of work is necessary to reformat the json response.

While the ResponseWrapper is a great idea and may be part of the final design, it doesn't fully solve the problem - as you realize with the mention of the `ReturnAllBidStatus` issue.

I agree with your ideal solution of moving exchange.makeExtBidResponse out of HoldAuction and into the auction, amp, and video endpoints. Instead of returning the bid response data structure, HoldAuction can return the parts needed to build the response, including all seat non bid data.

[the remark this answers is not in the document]
ShriprasadM: Thanks @SyntaxNode  for your review,  I will do the needful and update here

SyntaxNode:
I'm continuing the discussion of "Consider ReturnAllBidStatus after holdAuction" from the comment thread. The recommendation from @ShriprasadM is:
> With current approach, I strongly feel that this needs to move at auction.go, amp_auction.go and video_auction.go level. 

I agree. The bid response is currently constructed in `buildBidResponse` at the end of the `HoldAuction` function. It is returned as-is for /auction and repackaged for /amp (requiring extra work).  The `buildBidResponse` function is complex - combining bid validation, filtering, and response object generation.

What if we returned a different structure from `HoldAuction` and allowed each endpoint to package it as needed? This could result in a response something like:

```
type exchange.Response struct {
  ID
  NBR
  SeatBids ResponseSeat
  Debug
}

type exchange.ResponseSeat struct {
  Seat
  Bids []ResponseBid
  NonBids []
}

type exchange.ResponseBid struct {
  Bid openrtb2.Bid
  CacheInfo
  DealPriority
  Targeting
  OriginalBidCPM
  Meta
  ...
}

type exchange.ResponseNonBid struct {
  Bid openrtb2.Bid
  StatusCode
}
```

A new function for each endpoint will accept `exchange.Response` and format it either to an OpenRTB response or an AMP response, and has strongly typed information to share with analytics adapters.

ShriprasadM:
> I'm continuing the discussion of "Consider ReturnAllBidStatus after holdAuction" from the comment thread. The recommendation from @ShriprasadM is:
> 
> > With current approach, I strongly feel that this needs to move at auction.go, amp_auction.go and video_auction.go level.
> 
> I agree. The bid response is currently constructed in `buildBidResponse` at the end of the `HoldAuction` function. It is returned as-is for /auction and repackaged for /amp (requiring extra work). The `buildBidResponse` function is complex - combining bid validation, filtering, and response object generation.
> 
> What if we returned a different structure from `HoldAuction` and allowed each endpoint to package it as needed? This could result in a response something like:
> 
> ```
> type exchange.Response struct {
>   ID
>   NBR
>   Seats ResponseSeat
>   Debug
> }
> 
> type exchange.ResponseSeat struct {
>   SeatName
>   Bids []ResponseBid
>   NonBids []ResponseNonBid
> }
> 
> type exchange.ResponseBid struct {
>   Bid openrtb2.Bid
>   CacheInfo
>   DealPriority
>   Targeting
>   OriginalBidCPM
>   Meta
>   ...
> }
> 
> type exchange.ResponseNonBid struct {
>   Bid openrtb2.Bid
>   StatusCode
> }
> ```
> 
> A new function for each endpoint will accept `exchange.Response` and format it either to an OpenRTB response or an AMP response, and has strongly typed information to share with analytics adapters.

@SyntaxNode : Thanks for an update. 

As per your thought mentioned above, What I have understood that we are trying to go away from `HoldAuction` to return `openrtb2/BidResponse` object. Instead it will return  collection of `Seats`, containing logical separation of `Bids` and `NonBids`. 

But, before your comment, I had change the implementation and have introduced `AuctionResponse` object. `HoldAuction` response returns now `AuctionResponse`, which composed of `openrtb2.BidResponse` and `openrtb_ext.BidResponseExt`.

```go
// AuctionResponse contains OpenRTB Bid Response object and its extension (un-marshalled) object
type AuctionResponse struct {
	*openrtb2.BidResponse                             // BidResponse defines the contract for openrtb bidresponse
	ExtBidResponse        *openrtb_ext.ExtBidResponse // ExtBidResponse defines the contract for bidresponse.ext
}
```

Please provide your thoughts around the `AuctionResponse`. Also I have kept `ResponseWrapper` for future use. Please let me know, can we keep it as a part of the code. I have added lot of unit tests around `ResposneWrapper` too.

analytics/core.go
```diff
@@ -47,17 +49,21 @@ type AmpObject struct {
...
-	StartTime     time.Time
+	Status         int
+	Errors         []error
+	Request        *openrtb2.BidRequest
+	Response       *openrtb2.BidResponse
+	VideoRequest   *openrtb_ext.BidRequestVideo
+	VideoResponse  *openrtb_ext.BidResponseVideo
+	StartTime      time.Time
+	SeatNonBid     []openrtb_ext.SeatNonBid    `json:"-"`
+	RequestWrapper *openrtb_ext.RequestWrapper `json:"-"`
```
AlexBVolcy, line 66: What is the purpose of these json paths tied to these variables

endpoints/openrtb2/amp_auction.go
```diff
@@ -795,3 +808,22 @@ func setTrace(req *openrtb2.BidRequest, value string) (err error) {
 
 	return nil
 }
+
+// setSeatNonBid populates bidresponse.ext.prebid.seatnonbid if  bidrequest.ext.prebid.returnallbidstatus is true
+func setSeatNonBid(finalExtBidResponse *openrtb_ext.ExtBidResponse, request *openrtb_ext.RequestWrapper, aucResponse *exchange.AuctionResponse) bool {
+	if aucResponse == nil || request == nil {
+		return false
```
AlexBVolcy, line 815: Can you cover this line of code

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,33 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
-
+	setSeatNonBidRaw(req, auctionResponse)
 	labels, ao = sendAuctionResponse(w, deps.hookExecutor, response, req.BidRequest, account, labels, ao)
 }
 
+// setSeatNonBidRaw is temporary introduced function for setting SeatNonBid inside bidResponse.Ext
+//
+//	populates bidresponse.ext.prebid.seatnonbid if  bidrequest.ext.prebid.returnallbidstatus is true using setSeatNonBid()
+//
+// This function will go away in next PR of https://github.com/prebid/prebid-server/pull/2505/
```
AlexBVolcy, line 268: I'm confused by this comment. It sounds like you're saying this function will go away in this PR because the link you provided in this comment is this PR.

I think I need more clarity as to why this function is needed, and why `setSeatNonBid()` can't just be used in its place?

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,33 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
 		labels, ao = rejectAuctionRequest(*rejectErr, w, deps.hookExecutor, req.BidRequest, account, labels, ao)
 		return
 	}
-
+	setSeatNonBidRaw(req, auctionResponse)
 	labels, ao = sendAuctionResponse(w, deps.hookExecutor, response, req.BidRequest, account, labels, ao)
 }
 
+// setSeatNonBidRaw is temporary introduced function for setting SeatNonBid inside bidResponse.Ext
+//
```
AlexBVolcy, line 265: Can you remove this blank comment

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,33 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+//	populates bidresponse.ext.prebid.seatnonbid if  bidrequest.ext.prebid.returnallbidstatus is true using setSeatNonBid()
+//
```
AlexBVolcy, line 267: Can you remove this blank comment

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,33 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+// This function will go away in next PR of https://github.com/prebid/prebid-server/pull/2505/
+func setSeatNonBidRaw(request *openrtb_ext.RequestWrapper, aucResponse *exchange.AuctionResponse) bool {
```
AlexBVolcy, line 269: I'm confused by the use of the word `raw` in this function name? Could you explain this?

endpoints/openrtb2/amp_auction.go
```diff
@@ -795,3 +808,22 @@ func setTrace(req *openrtb2.BidRequest, value string) (err error) {
...
+	}
+	reqExt, err := request.GetRequestExt()
+	if err == nil {
```
AlexBVolcy, line 818: I think you need to handle the case where `err != nil` as well.

endpoints/openrtb2/amp_auction.go
```diff
@@ -422,6 +432,9 @@ func getExtBidResponse(
 		}
 	}
 
+	// Set seat non-bid if requested
```
AlexBVolcy, line 435: I think you can remove this comment

endpoints/openrtb2/amp_auction.go
```diff
@@ -795,3 +808,22 @@ func setTrace(req *openrtb2.BidRequest, value string) (err error) {
...
+	}
+	reqExt, err := request.GetRequestExt()
+	if err == nil {
+		prebid := reqExt.GetPrebid()
+		if prebid != nil && prebid.ReturnAllBidStatus {
+			if finalExtBidResponse.Prebid == nil {
+				finalExtBidResponse.Prebid = &openrtb_ext.ExtResponsePrebid{}
+			}
+			finalExtBidResponse.Prebid.SeatNonBid = aucResponse.GetSeatNonBid()
+			return true
```
AlexBVolcy, line 825: Can you cover this block of code?

analytics/core.go
ShriprasadM: I am unable to follow the question but, 

1. These fields are ignored from marshalling / un-marshalling, with the thought that `PubStack` module is marshalling entire AuctionObject. So, avoiding marshalling of these two fields because - 
      a. It many increase the network packet size, as SeatNonBid will capture all non-positive bids, whereas `RequestWrapper` is just a helper for keeping the unmarshalled version of extensions

endpoints/openrtb2/amp_auction.go
ShriprasadM: 100% Covered using `TestSetSeatNonBid`

endpoints/openrtb2/auction.go
ShriprasadM: Because, 
 1. today exchange.HoldAuction prepares and marshals some piece of response.Ext which is then used by auction.go, amp_auction.go and video_auction.go
 2. As per discussion with @SyntaxNode  we are planning to move away from - HoldAuction building openrtb2.BidResponse. instead respective auction modules will build this object - https://github.com/prebid/prebid-server/pull/2505#issuecomment-1460978275 
 3. So, we will need this method to do first,  unmarshalling of response.Ext

endpoints/openrtb2/auction.go
ShriprasadM: Done

endpoints/openrtb2/auction.go
ShriprasadM: Done

endpoints/openrtb2/auction.go
ShriprasadM: https://github.com/prebid/prebid-server/pull/2505/files#discussion_r1150031661 
Added `Raw` as suffix as there is already function with name `setSeatNonBid`

endpoints/openrtb2/amp_auction.go
ShriprasadM: Done

endpoints/openrtb2/amp_auction.go
ShriprasadM: Done

analytics/core.go
AlexBVolcy: What I'm wondering is why do we need ``json:"-"`` to be next to these fields if they are going to be ignored anyway? Can't the json part of it be deleted, and instead these lines just look like:

```
SeatNonBid     []openrtb_ext.SeatNonBid
RequestWrapper *openrtb_ext.RequestWrapper
```

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,34 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+	// by HoldAuction
+	response := aucResponse.BidResponse
+	respExt := new(openrtb_ext.ExtBidResponse)
+
+	if json.Unmarshal(response.Ext, &respExt) == nil && setSeatNonBid(respExt, request, aucResponse) {
+		// marshal again
+		if respExtJson, err := json.Marshal(respExt); err == nil {
+			response.Ext = respExtJson
+			return true
+		}
```
AlexBVolcy, line 283: I think I'd want to see more error handling here for the unmarshalling/marshalling in the event where the error is not equal to nil. So if either unmarshal/marshal results in an error, I'd want to see that passed upstream rather than it being ignored.

ShriprasadM: Done

AlexBVolcy:
This is looking good. I just want to bring up this comment I left yesterday that you might've missed. https://github.com/prebid/prebid-server/pull/2505#discussion_r1151241944

analytics/core.go
ShriprasadM: @AlexBVolcy : These fields will not be ignored by `PubStack` module, if `json:"-"` is not mentioned. 

> I am unable to follow the question but,
> 
> 1. These fields are ignored from marshalling / un-marshalling, with the thought that `PubStack` module is marshalling entire AuctionObject. So, avoiding marshalling of these two fields because -
>    a. It many increase the network packet size, as SeatNonBid will capture all non-positive bids, whereas `RequestWrapper` is just a helper for keeping the unmarshalled version of extensions

exchange/seat_non_bids.go
```diff
@@ -0,0 +1,68 @@
...
+	seatNonBidsMap map[string][]openrtb_ext.NonBid
+}
+
+func newSeatNonBids() seatNonBids {
+	return seatNonBids{
+		seatNonBidsMap: make(map[string][]openrtb_ext.NonBid),
+	}
+}
+
+func (snb *seatNonBids) add(bid *entities.PbsOrtbBid, nonBidReason int, seat string) {
```
SyntaxNode, line 19: Consider using a value receiver here instead of a pointer in receiver. In Go, it's good practice to always default to a value unless you need to set state or the structure is really large - neither which apply in this case.

exchange/seat_non_bids.go
```diff
@@ -0,0 +1,68 @@
...
+	if bid == nil || bid.Bid == nil {
+		return
+	}
+	if snb.seatNonBidsMap == nil {
+		snb.seatNonBidsMap = make(map[string][]openrtb_ext.NonBid)
+	}
+	nonBid := openrtb_ext.NonBid{
+		ImpId:      bid.Bid.ImpID,
+		StatusCode: nonBidReason, //
```
SyntaxNode, line 28: Nitpick: Remove unused comment characters.

exchange/seat_non_bids.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+	}
+
+	nonBids := snb.seatNonBidsMap[seat]
+
+	if nonBids == nil {
+		snb.seatNonBidsMap[seat] = []openrtb_ext.NonBid{nonBid}
+	} else {
+		snb.seatNonBidsMap[seat] = append(nonBids, nonBid)
+	}
```
SyntaxNode, line 54: This condition isn't necessary. The `append` method will gracefully handle nil slices. This could just be `snb.seatNonBidsMap[seat] = append(nonBids, nonBid)`

exchange/seat_non_bids.go
SyntaxNode, line 19: Nitpick: Consider naming this `addBid` since it's adding a bid and not a seatNonBid.

exchange/seat_non_bids.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (snb *seatNonBids) get() []openrtb_ext.SeatNonBid {
+	var seatNonBid []openrtb_ext.SeatNonBid
+	if snb.seatNonBidsMap != nil {
```
SyntaxNode, line 59: Do you need the nil check? The `range` loop should gracefully handle nil or empty maps.

openrtb_ext/response.go
```diff
@@ -96,3 +100,33 @@ const (
...
+
+// Bid is Wrapper around original/proxy bid object
+type Bid struct {
+	openrtb2.Bid
+	ID    string `json:"id,omitempty"`    // added omitempty
+	ImpID string `json:"impid,omitempty"` // added omitempty
+
+	OriginalBidCPM float64 `json:"originalbidcpm,omitempty"`
+	OriginalBidCur string  `json:"originalbidcur,omitempty"`
+}
```
SyntaxNode, line 112: Why does this embed `openrtb2.Bid` rather than including only the fields we want to represent in our prebid ext?

openrtb_ext/response.go
```diff
@@ -96,3 +100,33 @@ const (
...
+
+type Prebid struct {
+	Bid Bid `json:"bid"`
+}
```
SyntaxNode, line 116: `Prebid` is too generic of a name for this package. Please name it `ExtResponseNonBidPrebid` to keep with the established naming convention.

static/bidder-info/appnexus.yaml
```diff
@@ -1,4 +1,4 @@
-endpoint: "http://ib.adnxs.com/openrtb2"
+endpoint: "http://tech-stack-mgmt.pubmatic.com/owtools/keywords-based/owtools/getbid"
```
SyntaxNode, line 1: Please revert your test urls.

SyntaxNode (commented):
Thanks for the hard work. I have many comments, but the overall approach is great and I'm happy with the direction we're going in.

exchange/auction_response.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"github.com/prebid/openrtb/v17/openrtb2"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+// AuctionResponse contains OpenRTB Bid Response object and its extension (un-marshalled) object
+type AuctionResponse struct {
+	*openrtb2.BidResponse                             // BidResponse defines the contract for openrtb bidresponse
+	ExtBidResponse        *openrtb_ext.ExtBidResponse // ExtBidResponse defines the contract for bidresponse.ext
+}
```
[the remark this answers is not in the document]
ShriprasadM: @SyntaxNode : embedded `openrtb2.BidResponse`, because, I was looking at  `AuctionResponse` is extended version of `openrtb2.BidResponse`. So, it would be easy for consumer to access fields of  `openrtb2.BidResponse` directly, without doing `auctionResopnse.Bid` or similar

openrtb_ext/response.go
ShriprasadM: embedded `openrtb2.Bid` with the thought that,
1. `openrtb_ext.Bid` is wrapper around `openrtb2.Bid`
2. It will be easy for adding new fields whenever required just by assigning the values to fields present in `openrtb2.Bid`
3. If we not kept `openrtb2.Bid` object embedded then will need to ensure field names are inline with `openrtb2.Bid` object. Othewise, consumer will need to keep mapping between `openrtb2.Bid` and `openrtb_ext.Bid`. e.g. If consumer wants to log `openrtb2.Bid.SlotInPod` but the field declared under `openrtb_ext.Bid` object was `PodSlot`

bretg: @ShriprasadM - I pretty strongly disagree with adding the bid object. The issue https://github.com/prebid/prebid-server/issues/2367#issuecomment-1375775940 lays out the fields that are needed in the nonbid. If there are fields of interest to analytics that are missing, please feel to suggest.

Specifically, I will not accept passing `adm` through the nonbid. That field can be many KB large and would be a waste of space. To a lesser concern, the various url fields (burl, nurl, etc) would also a waste of network bandwidth.

Because of those fields, I think it's unwise to pretend that we can "just pass the whole seatbid.bid object".  But I will add it to the agenda for the next meeting. If @SyntaxNode and the committee think it's acceptable to just strip the long-and-useless fields, I will abide.

exchange/auction_response.go
```diff
@@ -0,0 +1,20 @@
...
+
+// GetSeatNonBid returns array of seat non-bid if present. nil otherwise
+func (ar *AuctionResponse) GetSeatNonBid() []openrtb_ext.SeatNonBid {
+	if ar.ExtBidResponse != nil && ar.ExtBidResponse.Prebid != nil {
```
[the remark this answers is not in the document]
ShriprasadM: Done

exchange/exchange.go
```diff
@@ -117,6 +117,8 @@ type deduplicateChanceGenerator interface {
 
 type randomDeduplicateBidBooleanGenerator struct{}
 
+var sNonBids = newSeatNonBids()
```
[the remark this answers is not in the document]
ShriprasadM: Done

exchange/seat_non_bids.go
ShriprasadM: Done. Removed nil check

openrtb_ext/response.go
ShriprasadM: Done

bsardo:
@ShriprasadM can you resolve the conflicts when you get a chance? Thanks!

bretg:
Note: we're still trying iron out ext.seatnonbid.nonbid.ext.prebid.bid in https://github.com/prebid/prebid-server/issues/2367

--------

Updated ext.seatnonbid.nonbid.ext.prebid.bid to include origbidcpm and origbidcur. Note `orig` and not `original` to align them with the extension PBS supports in the seatbid. https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#original-bid-cpm

Specifically, in this PR, origbidcpm and origbidcur are described as being outside the bid object. The existing origbidcpm and origbidcur extensions are actually in bid.ext. Are you ok with just placing them in ext.seatnonbid.nonbid.ext.prebid.bid ?

endpoints/openrtb2/amp_auction.go
```diff
@@ -236,8 +236,14 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 		HookExecutor:               deps.hookExecutor,
 	}
 
-	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
+	aucResponse, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
+	var response *openrtb2.BidResponse
+	if aucResponse != nil {
+		response = aucResponse.BidResponse
+		ao.SeatNonBid = aucResponse.GetSeatNonBid()
+	}
```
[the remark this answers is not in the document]
ShriprasadM:
```go
ao.AuctionResponse = auctionResponse.BidResponse
```
@SyntaxNode :  Just worrying about above code snippet. I believe it will fail in case `auctionResponse` is `nil`.

Other changes can be done. 
1. Removed `response` variable
2. convert receiver of `GetSeatNonBid()` to pointer and added nil check there

endpoints/openrtb2/amp_auction.go
```diff
@@ -236,8 +236,14 @@ func (deps *endpointDeps) AmpAuction(w http.ResponseWriter, r *http.Request, _ h
 		HookExecutor:               deps.hookExecutor,
 	}
 
-	response, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
+	aucResponse, err := deps.ex.HoldAuction(ctx, auctionRequest, nil)
```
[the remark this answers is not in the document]
ShriprasadM: Done, will push the changes

ShriprasadM:
> @ShriprasadM can you resolve the conflicts when you get a chance? Thanks!

@bsardo , yes conflicts are resolved. Working on comments given by @SyntaxNode

ShriprasadM:
> Note: we're still trying iron out ext.seatnonbid.nonbid.ext.prebid.bid in #2367
> 
> Updated ext.seatnonbid.nonbid.ext.prebid.bid to include origbidcpm and origbidcur. Note `orig` and not `original` to align them with the extension PBS supports in the seatbid. https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#original-bid-cpm
> 
> Specifically, in this PR, origbidcpm and origbidcur are described as being outside the bid object. The existing origbidcpm and origbidcur extensions are actually in bid.ext. Are you ok with just placing them in ext.seatnonbid.nonbid.ext.prebid.bid ?

@bretg : Updated the PR with  `origbidcpm` and `origbidcur`. 
@SyntaxNode : I have removed `Bid` object from the structure. Can you review that change. I am working on other review comments you have added

openrtb_ext/response.go
```diff
@@ -96,3 +101,44 @@ const (
...
+	CatTax  adcom1.CategoryTaxonomy `json:"cattax,omitempty"`
+	Cat     []string                `json:"cat,omitempty"`
+	DealID  string                  `json:"dealid,omitempty"`
+	W       int64                   `json:"w,omitempty"`
+	H       int64                   `json:"h,omitempty"`
+	Dur     int64                   `json:"dur,omitempty"`
+	MType   openrtb2.MarkupType     `json:"mtype,omitempty"`
+
+	OriginalBidCPM float64 `json:"origbidcpm,omitempty"`
+	OriginalBidCur string  `json:"origbidcur,omitempty"`
```
ShriprasadM, line 120: @bretg : Update `json` variables as per your comments. Please review.

endpoints/openrtb2/amp_auction.go
```diff
@@ -281,19 +287,21 @@ func rejectAmpRequest(
...
 	hookExecutor hookexecution.HookStageExecutor,
-	response *openrtb2.BidResponse,
+	aucResponse *exchange.AuctionResponse,
 	reqWrapper *openrtb_ext.RequestWrapper,
 	account *config.Account,
 	labels metrics.Labels,
 	ao analytics.AmpObject,
 	errs []error,
 ) (metrics.Labels, analytics.AmpObject) {
+	// assuming aucResponse is never nil here
```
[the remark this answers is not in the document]
ShriprasadM: added nil check

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,39 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+// 2. As per discussion with Scott we are planning to move away from - HoldAuction building openrtb2.BidResponse. instead respective auction modules will build this object
+// 3. So, we will need this method to do first,  unmarshalling of response.Ext
+func setSeatNonBidRaw(request *openrtb_ext.RequestWrapper, aucResponse *exchange.AuctionResponse) error {
+	if aucResponse == nil || aucResponse.BidResponse == nil {
+		return nil
+	}
+	// unmarshalling is required here, until we are moving away from bidResponse.Ext, which is populated
+	// by HoldAuction
+	response := aucResponse.BidResponse
+	respExt := new(openrtb_ext.ExtBidResponse)
```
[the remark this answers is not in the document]
ShriprasadM: Done

exchange/auction_response.go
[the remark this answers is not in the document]
ShriprasadM: Done

exchange/seat_non_bids.go
```diff
@@ -0,0 +1,68 @@
+package exchange
+
+import (
+	"github.com/prebid/openrtb/v17/openrtb2"
+	"github.com/prebid/prebid-server/exchange/entities"
+	"github.com/prebid/prebid-server/openrtb_ext"
+)
+
+type seatNonBids struct {
+	seatNonBidsMap map[string][]openrtb_ext.NonBid
```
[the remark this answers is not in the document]
ShriprasadM: Done

exchange/seat_non_bids.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func newSeatNonBids() seatNonBids {
+	return seatNonBids{
+		seatNonBidsMap: make(map[string][]openrtb_ext.NonBid),
+	}
+}
```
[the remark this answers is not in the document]
ShriprasadM: Done

exchange/seat_non_bids.go
ShriprasadM: Done

exchange/seat_non_bids.go
ShriprasadM: Done

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,39 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+	}
 	labels, ao = sendAuctionResponse(w, deps.hookExecutor, response, req.BidRequest, account, labels, ao)
 }
 
+// setSeatNonBidRaw is temporary introduced function for setting SeatNonBid inside bidResponse.Ext
+// Because,
+// 1. today exchange.HoldAuction prepares and marshals some piece of response.Ext which is then used by auction.go, amp_auction.go and video_auction.go
+// 2. As per discussion with Scott we are planning to move away from - HoldAuction building openrtb2.BidResponse. instead respective auction modules will build this object
+// 3. So, we will need this method to do first,  unmarshalling of response.Ext
+func setSeatNonBidRaw(request *openrtb_ext.RequestWrapper, aucResponse *exchange.AuctionResponse) error {
```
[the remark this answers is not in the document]
ShriprasadM: Done

exchange/seat_non_bids_test.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"testing"
+
+	"github.com/prebid/openrtb/v19/openrtb2"
+	"github.com/prebid/prebid-server/exchange/entities"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSeatNonBidsAdd(t *testing.T) {
```
AlexBVolcy, line 12: Repeating a comment left by @SyntaxNode, but could you update the descriptions of these tests to not include spaces? (e.g. `nil seatNonBidsMap` --> `nil-seatNonBidsMap`)

exchange/seat_non_bids_test.go
```diff
@@ -0,0 +1,109 @@
...
+			snb := &nonBids{
+				seatNonBidsMap: tt.fields.seatNonBidsMap,
+			}
+			snb.addBid(tt.args.bid, tt.args.nonBidReason, tt.args.seat)
+			assert.Equalf(t, tt.want, snb.seatNonBidsMap, "expected seatNonBidsMap not nil")
+		})
+	}
+}
+
+func TestSeatNonBidsGet(t *testing.T) {
```
AlexBVolcy, line 57: Could you add an extra test case here where seatNonBids is equal to nil

endpoints/openrtb2/auction.go
```diff
@@ -249,10 +257,39 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+	respExt := new(openrtb_ext.ExtBidResponse)
+	if err := json.Unmarshal(response.Ext, &respExt); err != nil {
+		return err
+	}
+	if setSeatNonBid(respExt, request, aucResponse) {
+		// marshal again
+		if respExtJson, err := json.Marshal(respExt); err == nil {
+			response.Ext = respExtJson
+			return nil
+		}
```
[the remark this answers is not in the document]
ShriprasadM: Not returning error, considering, if marshalling failure due to `seatNonBid` context, should not fail entire request and may respond without `seatNonBid` information. 
If you see we can fail the request in case failed to marshal `seatNonBid` information

exchange/seat_non_bids_test.go
ShriprasadM: Done

exchange/seat_non_bids_test.go
ShriprasadM: Done

analytics/core.go
```diff
@@ -35,6 +35,8 @@ type AuctionObject struct {
 	Account              *config.Account
 	StartTime            time.Time
 	HookExecutionOutcome []hookexecution.StageOutcome
+	SeatNonBid           []openrtb_ext.SeatNonBid    `json:"-"`
+	RequestWrapper       *openrtb_ext.RequestWrapper `json:"-"`
```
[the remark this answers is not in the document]
SyntaxNode: I empathize with wanting to avoid unmarshalling extension objects within each analytics adapter. We need to balance with the danger of memory corruption since the RequestWrapper is not thread safe and even the Get methods modify memory (so this is a different situation than enforcing no writes to the Request object). There is now the ability to clone a RequestWrapper , so we could clone it before passing it to each adapter - but then we need to consider if the cost of memory copying is worth the unmarshal savings.

If we replace the Request with the RequestWrapper the json marshal would change. However, I am thinking now that the file system and pubstack adapters make a mistake by using the data objects in this way. We could solve that problem by creating separate objects for each of those adapters to isolate the changes. This would also remove the need to specify  `json:"-"`.

Thoughts @hhhjort @bsardo ?

endpoints/openrtb2/amp_auction.go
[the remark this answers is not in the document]
SyntaxNode: That's a good point. We could also solve this problem by always returning an AuctionResponse value instead of a pointer, but that may be a large change and there is a pleasantness to returning `nil` when there is an error.

I am good with either leaving this as (skeptically) returning a value instead of a pointer.

analytics/core.go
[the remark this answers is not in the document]
SyntaxNode: I opened https://github.com/prebid/prebid-server/pull/2739 to separate the data models so you don't need to add `json:"-"` here and if we decide to replace Request with RequestWrapper, the analytics adapters can handle the change.

endpoints/openrtb2/auction.go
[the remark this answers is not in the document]
SyntaxNode: The only reason why a Marshal call should fail is due to memory corruption. If that's happening, it's appropriate to fail the entire request and send the error to the log. This current logic is burying the error.

exchange/exchange.go
```diff
@@ -1548,3 +1560,19 @@ func setErrorMessageSecureMarkup(validationType string) string {
 	}
 	return ""
 }
+
+// setSeatNonBid  adds SeatNonBids within bidResponse.Ext.Prebid.SeatNonBid
```
SyntaxNode, line 1564: Nitpick: Please remove the extra space in the comment after the function name.

openrtb_ext/response.go
```diff
@@ -96,3 +101,44 @@ const (
 	UserSyncIframe UserSyncType = "iframe"
 	UserSyncPixel  UserSyncType = "pixel"
 )
+
+// Bid is subset of Bid object with exact json signature
+// defined at https://github.com/prebid/openrtb/blob/v17.1.0/openrtb2/bid.go.
+// It also contains the custom fields
+type Bid struct {
```
SyntaxNode, line 108: Looks great to me. Thank you. I would recommend to change the name to something which clearly associates it with a nonBid. You're welcomed to come up with something clever, but I'm fine with something a bit obvious, if not a little silly, like`nonBidBid`.

analytics/core.go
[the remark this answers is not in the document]
hhhjort: If analytics adapters are only going to use the wrapper to read the extension, and not write anything, then it should be fairly safe to pass them the wrapper. The only thing that could happen on read is on the GetXExt() call where you get a wrapper Ext object. If one doesn't already exist on the wrapper, the wrapper will spin up a new one to return. If two adapters call this at the same time, you could get two copies of the ext object. The copies will be identical clones, and each adapter will consistently use the one it started with. After the race condition window, the wrapper will only have one of them. The other copy will persist until the adapter finishes with it and it is garbage collected.

So as long as we police the analytics adapters to not write to the request (which we should do anyway), and PBS core is not modifying the request wrapper while the analytics adapters are doing their thing, it should be okay to let the adapters have the request wrapper.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -2317,3 +2316,69 @@ func (e errorResponseWriter) Write(bytes []byte) (int, error) {
...
+			want: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := setSeatNonBid(tt.args.finalExtBidResponse, tt.args.request, tt.args.aucResponse); got != tt.want {
+				t.Errorf("setSeatNonBid() = %v, want %v", got, tt.want)
+			}
+		})
+	}
```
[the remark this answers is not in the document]
ShriprasadM: Done

openrtb_ext/response.go
ShriprasadM: I have renamed it to `NonBidObject`

analytics/core.go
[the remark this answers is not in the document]
bsardo: @ShriprasadM, we discussed this offline and we're ok with you using the wrapper in analytics adapters given the constraints @hhhjort outlined. As mentioned above, we will need to be on the lookout for attempts to write to the request. I'm going to open up an issue asking for new race tests to help us catch potential writes to the wrapper from within the analytics adapters so it is not just a manual code review effort.

bsardo:
@ShriprasadM is this ready for re-review now? The test failure is due to a flakey test that we've got a fix for and will merge with master shortly.

ShriprasadM:
> @ShriprasadM is this ready for re-review now? The test failure is due to a flakey test that we've got a fix for and will merge with master shortly.

@bsardo Yes PR is ready for review. I have addressed following review comments given by @SyntaxNode  

1. remove the json annotations on the analytics objects.
2. Remove the request from the analytics objects since now the request wraper is provided. we don’t need both.
3. would you like to resolve //TODO: move 303 to constant and maintain list of Non Bid Reason with creating the enum?

endpoints/openrtb2/amp_auction.go
```diff
@@ -809,3 +824,23 @@ func setTrace(req *openrtb2.BidRequest, value string) (err error) {
 
 	return nil
 }
+
+// setSeatNonBid populates bidresponse.ext.prebid.seatnonbid if  bidrequest.ext.prebid.returnallbidstatus is true
```
AlexBVolcy, line 828: Nitpick: could you remove extra space in between if and bidrequest in this comment

endpoints/openrtb2/auction.go
```diff
@@ -257,12 +262,44 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+	}
+	// unmarshalling is required here, until we are moving away from bidResponse.Ext, which is populated
+	// by HoldAuction
+	response := auctionResponse.BidResponse
+	respExt := &openrtb_ext.ExtBidResponse{}
+	if err := json.Unmarshal(response.Ext, &respExt); err != nil {
+		return err
+	}
+	if setSeatNonBid(respExt, request, auctionResponse) {
+		// marshal again
```
AlexBVolcy, line 292: Nitpick: This comment can be removed

endpoints/openrtb2/auction.go
```diff
@@ -257,12 +262,44 @@ func (deps *endpointDeps) Auction(w http.ResponseWriter, r *http.Request, _ http
...
+		if respExtJson, err := json.Marshal(respExt); err == nil {
+			response.Ext = respExtJson
+			return nil
+		} else {
+			return err
```
AlexBVolcy, line 297: This code block isn't covered by tests, could you update this?

exchange/exchange.go
```diff
@@ -1573,3 +1585,19 @@ func setErrorMessageSecureMarkup(validationType string) string {
...
+	}
+	if bidResponseExt == nil {
+		bidResponseExt = &openrtb_ext.ExtBidResponse{}
+	}
+	if bidResponseExt.Prebid == nil {
+		bidResponseExt.Prebid = &openrtb_ext.ExtResponsePrebid{}
+	}
+
+	bidResponseExt.Prebid.SeatNonBid = seatNonBids.get()
+	return bidResponseExt
```
AlexBVolcy, line 1602: These lines of code aren't covered by tests, can this be updated?

exchange/seat_non_bids_test.go
```diff
@@ -0,0 +1,111 @@
...
+	type fields struct {
+		// seatNonBidsMap map[string][]openrtb_ext.NonBid
```
AlexBVolcy, line 59: Can this comment be removed?

exchange/exchange.go
ShriprasadM: image
Done. But unable to cover else part as its hard to produce `json.Marshal` error scenario

endpoints/openrtb2/auction.go
ShriprasadM: image
Done. But unable to cover else part as its hard to produce `json.Marshal` error scenario

exchange/exchange.go
ShriprasadM: Done. 
image

exchange/seat_non_bids_test.go
ShriprasadM: Done

endpoints/openrtb2/auction.go
ShriprasadM: Done

ShriprasadM:
@SyntaxNode / @bsardo / @AlexBVolcy :  `addbid` method is not thread safe. So, we should avoid multiple threads calling this method. Added method comment for the same
 https://github.com/prebid/prebid-server/pull/2505/commits/68f744290ac56514f90691a263446768960ceec4

SyntaxNode:
> Please tell, if we should add mutex here ?

I think we're fine with `addBid` not being thread safe. Right now, there's just one call to that method in the video endpoint which does not run in parallel. You'll find that most of the bid handling runs sequentially. We use a channel receiver loop to collect the bids which adapts from the parallel calls into sequential order.

You may encounter a need to add concurrency protection in the future, but I recommend to try and avoid it for now.

This is a good though, and many others may have the same concern. I recommend you add a comment to `addBid` which clearly states the method is not thread safe and cannot be called in parallel.

ShriprasadM:
> > Please tell, if we should add mutex here ?
> 
> I think we're fine with `addBid` not being thread safe. Right now, there's just one call to that method in the video endpoint which does not run in parallel. You'll find that most of the bid handling runs sequentially. We use a channel receiver loop to collect the bids which adapts from the parallel calls into sequential order.
> 
> You may encounter a need to add concurrency protection in the future, but I recommend to try and avoid it for now.
> 
> This is a good though, and many others may have the same concern. I recommend you add a comment to `addBid` which clearly states the method is not thread safe and cannot be called in parallel.

@SyntaxNode : I have already updated comment there. Here is commit - https://github.com/prebid/prebid-server/pull/2505/commits/68f744290ac56514f90691a263446768960ceec4

SyntaxNode (approved):

AlexBVolcy (approved):

PR #2507, opened 2022-12-27
TLS connection for MySql and Postgres drivers

victorshevtsov:
Some cloud providers allow only a secure connection to a hosted MySql instance.

So we need to be able to pass certificates to the prebid-server in the configuration as follows:
```
stored_requests:
  database:
    connection:
      driver: mysql
      server_ca: /mysql/ca.pem
      client_cert: /mysql/client-cert.pem
      client_key: /mysql/client-key.pem
```


holms:
@VeronikaSolovei9 could you please take a look at this, as you've previously reviewed huge PR for mysql support, maybe this one would be quicker for you because you're more familiar with this piece of code. 

We have a problem that digitalocean doesn't support non-ssl connections anymore for their managed databases. So we kind of blocked by this feature.

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +84,65 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+		if err := setupTLSConfig(provider); err != nil {
+			return "", err
+		}
+		buffer.WriteString("?tls=prebid-tls")
+	}
+
+	return buffer.String(), nil
+}
+
+func setupTLSConfig(provider *MySqlDbProvider) error {
```
VeronikaSolovei9, line 98: Thank you for your contribution. Changes in this PR make sense to me, I found related documentation here: 
https://pkg.go.dev/github.com/go-sql-driver/mysql#RegisterTLSConfig
https://cloud.google.com/sql/docs/mysql/samples/cloud-sql-mysql-databasesql-connect-tcp-sslcerts

I checked it works with my local MySQL and Postgres DBs. 
However I din't find how to configure my local MySQL server to support certificates.

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +84,65 @@ func (provider *MySqlDbProvider) ConnString() string {
 		buffer.WriteString(provider.cfg.Database)
 	}
 
-	return buffer.String()
+	// TSL connection
+	if provider.cfg.ServerCa != "" && provider.cfg.ClientCert != "" && provider.cfg.ClientKey != "" {
+		if err := setupTLSConfig(provider); err != nil {
+			return "", err
+		}
+		buffer.WriteString("?tls=prebid-tls")
```
VeronikaSolovei9, line 92: Should it be a config value: `prebid-tls`?

victorshevtsov: Thee are 3 new settings in prebid configuration to setup TLS connection- `server_ca`, `client_cert` and `client_key`:
```
stored_requests:
  database:
    connection:
      driver: mysql
      host: 172.18.0.3
      port: 3306
      user: root
      password: PreBidPwd
      dbname: prebid
      server_ca: /home/user/ca.pem
      client_cert: /home/user/client-cert.pem
      client_key: /home/user/client-key.pem
```

Here you can find how to configure MySQL to use encrypted connection:
https://dev.mysql.com/doc/refman/8.0/en/using-encrypted-connections.html

VeronikaSolovei9: Yes, understand. I'm only talking about `tls=prebid-tls`. It's weird to see it hardcoded. We will discuss with the team.

victorshevtsov: The `prebid-tls` string is hardcoded in two places:
- when registering the TLS config
- when referencing it in the connection string
It could be any string.

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +84,65 @@ func (provider *MySqlDbProvider) ConnString() string {
 		buffer.WriteString(provider.cfg.Database)
 	}
 
-	return buffer.String()
+	// TSL connection
+	if provider.cfg.ServerCa != "" && provider.cfg.ClientCert != "" && provider.cfg.ClientKey != "" {
```
bsardo, line 88: If we created a TLS struct as I described in an earlier comment, and all three of these are needed for TLS, then we could put a receiver method on that TLS struct that checks if TLS is explicitly enabled.

bsardo, line 88: If we're expecting either none or all three of these to be set, we could add some validation at startup that warns if only one or two of these are set.

config/stored_requests.go
```diff
@@ -190,12 +190,15 @@ func (cfg *DatabaseConfig) validate(dataType DataType, errs []error) []error {
 // DatabaseConnection has options which put types to the Database Connection string. See:
 // https://godoc.org/github.com/lib/pq#hdr-Connection_String_Parameters
 type DatabaseConnection struct {
```
bsardo, line 192: @VeronikaSolovei9 were you able to test mysql and postgres with and without TLS locally?

stored_requests/backends/db_provider/mysql_dbprovider.go
bsardo: Should we move this to a constant defined at the top of the file?

stored_requests/backends/db_provider/mysql_dbprovider_test.go
```diff
@@ -83,7 +83,7 @@ func TestConnStringMySql(t *testing.T) {
 			cfg: cfg,
 		}
 
-		connString := provider.ConnString()
+		connString, _ := provider.ConnString()
```
bsardo, line 86: Can we add a few additional test cases to exercise the new logic in `ConnString` and its helper function `setupTLSConfig`? It looks like we would just need to generate a couple of fake certificates.

config/stored_requests.go
```diff
@@ -190,12 +190,15 @@ func (cfg *DatabaseConfig) validate(dataType DataType, errs []error) []error {
...
-	Password string `mapstructure:"password"`
+	Driver     string `mapstructure:"driver"`
+	Database   string `mapstructure:"dbname"`
+	Host       string `mapstructure:"host"`
+	Port       int    `mapstructure:"port"`
+	Username   string `mapstructure:"user"`
+	Password   string `mapstructure:"password"`
+	ServerCa   string `mapstructure:"server_ca"`
+	ClientCert string `mapstructure:"client_cert"`
+	ClientKey  string `mapstructure:"client_key"`
```
bsardo, line 201: Also, I just realized that we don't currently support TLS for postgres. I see that when we build the postgres connection string we disable SSL: `buffer.WriteString("sslmode=disable")`.
We should eventually add TLS support for postgres in another PR.

For now, I'd like us to carefully consider the config variable names here since they will ultimately be used for other database TLS configurations.
I'm looking at https://pkg.go.dev/github.com/lib/pq@v1.10.4 and I see `sslcert`, `sslkey` and `sslrootcert` which should correspond to what we have here now right?
As for the query param in the postgres connection string, I think we would just set that to one of the other values described in the docs but we would still need all of these certs configured here.

@VeronikaSolovei9 any preferences on naming?

bsardo (commented):
Thanks for taking this on and thanks for your patience. Please see my comments when you get a moment.

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +84,65 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+	if err != nil {
+		return err
+	}
+
+	clientCert = append(clientCert, certs)
+	mysql.RegisterTLSConfig("prebid-tls", &tls.Config{
+		RootCAs:               rootCertPool,
+		Certificates:          clientCert,
+		InsecureSkipVerify:    true,
+		VerifyPeerCertificate: verifyPeerCertFunc(rootCertPool),
```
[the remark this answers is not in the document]
holms: Can we maybe make `InsecureSkipVerify` parameter available to set in config? We had situation that we couldn't connect without  this param even though we using their certificate which digital ocean provided themselves. So it's handy to be able to set this param for anyone with env var.

config/stored_requests.go
VeronikaSolovei9: I tested it with local MySQL server and Production Postgres. Both without TLS configs

config/stored_requests.go
VeronikaSolovei9: Good point! 
Hm, I would not combine all possible DB providers to the same config. So far we have generic host, post, user and password. Now we are about to add TLS support for MySql. For Postgres similar parameters are required, but have different names. 

Idea 1: 
Instead of these 3 variables add a `map[string]string` with extra DB parameters. It can lead to complications in setup, parsing, and figuring out what is what and for what DB provider. 

Idea 2:
Just add generic names like it defined in documentation: 
https://cloud.google.com/sql/docs/mysql/samples/cloud-sql-mysql-databasesql-connect-tcp-sslcerts

"DB_ROOT_CERT" // e.g., '/path/to/my/server-ca.pem'
"DB_CERT" // e.g. '/path/to/my/client-cert.pem'
"DB_KEY" // e.g. '/path/to/my/client-key.pem'

And rethink about it if something else will be needed in future.

bsardo: 👍 idea 2 is what I had in mind as well, though I think we should use names that don't include `DB`, `SSL` or `TLS` in them since we should already have that context from the config structure: `stored_requests.database.connections.tls.VARIABLE`.

In that case, how about:
```
root_cert
client_cert
client_key
```

As for @holms's [request](https://github.com/prebid/prebid-server/pull/2507#discussion_r1081648100) regarding adding `InsecureSkipVerify` to the config, I think we can do that but we should consider whether it makes sense for this to add more generic variable names or use the `map[string]string` format presented as idea 1 for DB-specific parameters. I still prefer using generic names over a map for this as well if possible. For postgres there is a query param `sslmode` that can be set to one of six values, one of which might map to `InsecureSkipVerify`. @holms which, if any, of the postgres values sounds like it matches the `InsecureSkipVerify` behavior?
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

holms: @bsardo  Perhaps `sslmode=verify-ca` would be the closest match? It doesn't check hostname, so probably self generated certs would pass with this param.

@victorshevtsov  I don't understand why there are three params in the config? You just need to provide a client key, which is pem certificate. I've just quickly googled how would I connect to mysql using tls with python for example, and all you need is just one certificate, also as digital ocean also provides only one certificate. Other params are specifically for mysql server, not client.

holms: > Also, I just realized that we don't currently support TLS for postgres. I see that when we build the postgres connection string we disable SSL: `buffer.WriteString("sslmode=disable")`. We should eventually add TLS support for postgres in another PR.
> 
> For now, I'd like us to carefully consider the config variable names here since they will ultimately be used for other database TLS configurations. I'm looking at https://pkg.go.dev/github.com/lib/pq@v1.10.4 and I see `sslcert`, `sslkey` and `sslrootcert` which should correspond to what we have here now right? As for the query param in the postgres connection string, I think we would just set that to one of the other values described in the docs but we would still need all of these certs configured here.
> 
> @VeronikaSolovei9 any preferences on naming?

Yeah I've actually planned to request Victor for another PR, so that TLS would be supported in postgesql too. Azure soon will be deprecating non-tls connections. Digital Ocean already doesn't allow to connect without tls to their postgresql managed service.

victorshevtsov: 1. Personally, I would go with the second idea, and use these names for the settings:
```
root_cert
client_cert
client_key
```

2. As @holms said, the client's cert and key parameters are optional. So, it is possible to pass only `root_cert` or all the 3 parameters.

VeronikaSolovei9:
I was able to configure my local MySQL server to accept secured connections. I tested this PR with all three parameters `server_ca`, `client_cert` and  `client_key` and with only one `server_ca`. Both tests worked successfully.

holms:
@VeronikaSolovei9 great :) are we good for the merge?

VeronikaSolovei9:
Do you want to add a support for just one `server_ca`? Now all 3 parameters are required

VeronikaSolovei9:
> Since these are TLS-specific configurations, perhaps we put these in their own struct:
> 
> ```
> TLS TLS `mapstructure:"tls"`
> 
> type TLS struct {
>     ServerCa   string `mapstructure:"server_ca"`
> 	ClientCert string `mapstructure:"client_cert"`
> 	ClientKey  string `mapstructure:"client_key"`
> }
> ```
> 
> Thoughts?


What is the decision about this suggestion? This idea looks good to me.

VeronikaSolovei9:
`root_cert` or `server_ca`?

holms:
> Do you want to add a support for just one `server_ca`? Now all 3 parameters are required

We decided that only `server_ca` is required. Waiting for a commit from Victor.

VeronikaSolovei9:
Right, but if other 2 are specified we need to add them to connection parameters

holms:
> Right, but if other 2 are specified we need to add them to connection parameters

Also when 2 of 3 is specified would be wise to crash, because in such case all 3 are required?

holms:
> I also think this hardcoded value should be extracted to constant: `prebid-tls`

I agree. This should be a constant.

VeronikaSolovei9:
> Also when 2 of 3 is specified would be wise to crash, because in such case all 3 are required?
Then maybe extra logic should be added to determine what parameters to use

VeronikaSolovei9:
Do we want to add `sslmode` parameter? For Postgres it's hardcoded to "disable"

victorshevtsov:
The more I think about the possibility to create a universal configuration for database connection, the more I understand that the best way is to have less settings and let the db driver do it's job:
- `driver` - mysql | postgres
- `connection_string` - a connection string specific for the driver

Such approach will make all the things much easier and we will not try to reimplement the drivers' logic to validate the parameters.

Please take a look at the documentation for both drivers:
https://pkg.go.dev/github.com/go-sql-driver/mysql#readme-parameters
https://pkg.go.dev/github.com/lib/pq#hdr-Connection_String_Parameters

In a case if we need to register a custom TLS config we can use a TLS structure as suggested earlier in the discussion.
See https://pkg.go.dev/github.com/go-sql-driver/mysql#RegisterTLSConfig for the details.

So, the connection config will look like the following:
```
stored_requests:
  database:
    connection:
      driver: mysql
      connection_string: user@tcp(localhost:3306)/prebid-db?tls=custom-tls
      tls:
        name: custom-tls
        root_cert: /home/user/prebid/root-cert.pem
        client_cert: /home/user/prebid/client-cert.pem
        client_key: /home/user/prebid/client-key.pem
```

What do you think?
@bsardo @VeronikaSolovei9 @holms

VeronikaSolovei9:
Trying to think about potential pros and cons of connection string.

Cons:

Connection string is usually created by DB library and follows DB specific format. I feel like it's a potential place to make a mistake and crafting this string itself is harder than just specifying username, password, url and other parameters separately. 
Also based on DB type this connection string can have different formats (i.e., connection strings are not portable).

For MySQL it can be a connection string and connection URL and both have different formats: 
Connection URL: `mysql://localhost:3306/test?username=username&password=password`
Connection string: `user@tcp(localhost:3306)/test?tls=custom`

For Postgres: 
Connection string:  `user=pqgotest dbname=pqgotest sslmode=verify-full`
Connection URL: `postgres://pqgotest:password@localhost/pqgotest?sslmode=verify-full`

In case the driver will need a new parameter we will need to research how to pass it in a connection string rather than pass this value as a config.
In the documentation there are separate fields for every parameter https://pkg.go.dev/github.com/go-sql-driver/mysql#Config

Pros:
Definitely will simplify other DB types integrations. It will solve the problem here with parameters we want to incorporate to the connection.

My gut feeling is that we should _not_ use the connection string. I'm interested to know other opinions.

About custom tls. Are there any cases where we may need more than one tls config? Or what is the reason to have it?

holms:
Connection string as an option definitely nice to have as a last resort. But validating it would be hard, and question is it even necessary. Maybe just try/catch when making connection? And then maybe driver would inform about this in a stacktrace of stdout?

Also I've been in situation when I've been using postgresql read replicas and you need to specify each node separately in a single connection string. Hardly anyone even supports such configuration. So connection string could be a solution for using sql clusters. 

I also don't see a reason to have name for tls, or have multiple tls'ses?

victorshevtsov:
Here are examples of connection strings for MySQL and ProstgreSQL:
https://www.connectionstrings.com/mysql/
https://www.connectionstrings.com/postgresql/

And here are the full docs:
https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

Ideally, Prebid server should not limit a customer to only the wrapped/mapped connection options, but let the customer use all the power of the selected DB driver.

holms:
> Here are examples of connection strings for MySQL and ProstgreSQL:
> 
> https://www.connectionstrings.com/mysql/
> 
> https://www.connectionstrings.com/postgresql/
> 
> 
> 
> And here are the full docs:
> 
> https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html
> 
> https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
> 
> 
> 
> Ideally, Prebid server should not limit a customer to only the wrapped/mapped connection options, but let the customer use all the power of the selected DB driver.

Maybe leaving some default mapping when connection string is not in use for backward compatibility at least..? Also why tls structure has name param? Should be able to define certs only once?

holms:
In my opinion some limited mapping would be nice to have. If someone like me will need more detailed confirmation I'd ussr connection string.

Also is it possible to pass a content of the cert? Because we hosting pbs in kubernetes, and cert would be passed from a kubernetes secret as a content. Not sure if it's possible to mount a secret to a file.

victorshevtsov:
> In my opinion some limited mapping would be nice to have. If someone like me will need more detailed confirmation I'd ussr connection string.
> 
> Also is it possible to pass a content of the cert? Because we hosting pbs in kubernetes, and cert would be passed from a kubernetes secret as a content. Not sure if it's possible to mount a secret to a file.

Yes, it's possible: https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod

holms:
@VeronikaSolovei9 so I've got all my questions answered. So what do you think about using connection strings, with some limited mapping for both of db's? Prebid-server would become less database dependent and it's quite a popular practice to use them in any backend service out there.

VeronikaSolovei9:
We briefly discussed it with the team. 
@SyntaxNode brought up a good point. With the connection string we will need to explicitly write username and password to the connection string. Now username and password are not exposed to the config along with other DB parameters, but with the connection string they will be.

Also migration to connection string will require a drastic changes in configs for all PBS hosts. 

@bsardo  @SyntaxNode please share your thoughts.

holms:
> We briefly discussed it with the team. 
> 
> @SyntaxNode brought up a good point. With the connection string we will need to explicitly write username and password to the connection string. Now username and password are not exposed to the config along with other DB parameters, but with the connection string they will be.
> 
>


I'm not sure how connection string will expose username and password because currently I'm currently passing username and password with env vars. Connection string would be also passed as env var. So in any case user and password are not exposed.


> Also migration to connection string will require a drastic changes in configs for all PBS hosts. 

True, having only connection string for db config would be a bit overkill, this would lead to a complex configuration. But having it as an additional way to configure connection of database looks like a wise approach, so that we wouldn't have to map all possible db connections params in the future. Imagine when we'll send a request for some mongodb support in near future? Mapping all possible mongo connection params which probably won't even exist in MySQL and/or pgsql would be a true hassle. What I'm saying is just having an ability to pass extra args in a format of connection string would be great feature. Prebid server is used in a massive scales and fine tuning it's db connection configuration in such way seems very a productive measure. 

Mapping tls structure which would be used in the connection string is a bit contradicting to my point.. but due to nature of certificates that's probably the cleanest way to implement tls config.

We already have access to your slack, we can discuss in there if it's more convenient for you.

bsardo:
@holms, @victorshevtsov, thanks again for all of your contributions here. I agree that we should allow for more advanced configuration leveraging the full power of the database. However, I think we can accomplish this without a `connection_string` config option.

I understand that a `connection_string` option does give the host full control but I have a couple of issues with it:

1. I don't consider the entire connection string to be sensitive but rather just the username and password. I assume some hosts store these two pieces of information in some secure storage with extra safeguards and then load them into environment variables. If we used a connection string option, we'd have to allow these pieces of information to be injected into the connection string which then introduces macros into the equation, or the host could store the entire connection string in secure storage but I don't think that makes sense.
2. I'd still want to provide the current config for both backwards compatibility and for ease of use. This then requires start-up logic to decide which config to use if the connection string and one or more pieces of the mapped config are specified. That's doable of course but it may introduce some unnecessary complexity.

If we just focus on RDBMS's for now and look at the most popular ones on the market (Postgres, MySQL, SQL Server, Oracle, SQLite), they all allow connection string options to be specified via a query string. If we just introduce a mapped `query_string` parameter, the `DbProvider` interface implementation for a given driver can simply append `query_string` to the end of the connection string it generates.
In this case, we'd end up with the following config:
```
stored_requests:
  database:
    connection:
      driver: "postgres"
      dbname: "prebid-db"
      host: "localhost"
      port: 3306
      user:     "SECRET"
      password: "SECRET"
      query_string: "?sslmode="
      tls:
        root_cert: path/to/root-cert.pem
        client_cert: path/to/client-cert.pem
        client_key: path/to/client-key.pem
```

If `tls` is provided, `sslmode` for postgres would default to something (e.g. `verify-full`) if not overridden in `query_string`. Similar default behavior would exist for MySQL.

This solution would also allow us the flexibility to introduce mapped parameters for the advanced configuration specified through the query string should we ever want to explore this approach more thoroughly in the future.

Also, I think that at this juncture, any discussion about extensibility should be focused on RDBMS. Adding support for NoSQL solutions will be a significant investment as they have their own set of challenges that we would need to take into account and I don't see the demand warranting such an undertaking right now.

Thoughts?

holms:
Oh yeah query strings, this actually solves our particular case. Thank you for this great suggestion.

@victorshevtsov I'm very positive to use query string, any issues implementing this approach?

victorshevtsov:
Yes, the suggested approach looks good for me.

SyntaxNode:
Tangent: @victorshevtsov I'd like to share your contribution at the next Prebid Server Webinar. May I ask what company you work for / contributed this change on behalf of?

holms:
> Tangent: @victorshevtsov I'd like to share your contribution at the next Prebid Server Webinar. May I ask what company you work for / contributed this change on behalf of?

Greetings. Victor contributed this PR on behalf of our company "[Setupad](setupad.com)". We are your partners, signed a contract this month.

stored_requests/backends/db_provider/postgres_dbprovider.go
```diff
@@ -78,8 +82,41 @@ func (provider *PostgresDbProvider) ConnString() string {
...
+	if !strings.Contains(provider.cfg.QueryString, "sslmode=") {
+		buffer.WriteString("sslmode=")
+		buffer.WriteString(sslmode)
+	}
+
+	// Rest of connection string parameters passed through query_string
+	if provider.cfg.QueryString != "" {
+		buffer.WriteString(" ")
+		buffer.WriteString(provider.cfg.QueryString)
+	}
```
bsardo, line 117: It appears postgres supports two connection string formats: key/val strings and URIs ([34.1.1 Connection Strings](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)) We're using the key/val format in this function but we are then appending the query string to the end of it. I guess technically a host could specify the query string as key/val with spaces separating the pairs but given that the name of the field is `query_string`, I would expect them to be defined in a URI format with `&` separating the pairs.
I think we should do one of the following:
1. move to a URI format (I can't imagine there are any drawbacks to doing so but we should do more research first. I think this is my preferred path forward as it should be simplest)
2. gracefully handle a URI-formatted `query_string` parameter by converting `&` to a space
3. validate the query string parameter and provide some indication of how it should be formatted

I guess we could always just let the connection fail on startup and the host could figure it out but that's not great. We are doing that to some extent already though I believe.

Thoughts?

bsardo (commented):
We could also use some additional test coverage of the connection string functions but I imagine you're trying to nail down the logic first before doing that which is totally fine.

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +86,92 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+			buffer.WriteString("&")
+			buffer.WriteString(provider.cfg.QueryString)
+		}
+	}
+
+	// Rest of connection string parameters passed through query_string
+	if provider.cfg.QueryString != "" {
+		buffer.WriteString("?")
+		buffer.WriteString(provider.cfg.QueryString)
+	}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Do we even need this  `tls=prebid-tls` parameter? It's a hardcoded value that adds complications to assembling the connection string. 
Would this work instead: 

stored_request.go: make TLS a pointer. It's simpler to check if it's nil: 
```
TLS         *TLS   `mapstructure:"tls"`
```
And then just:
```
tlss := provider.cfg.TLS != nil
	if tlss {
		if err := setupTLSConfig(provider); err != nil {
			return "", err
		}
	}
if provider.cfg.QueryString != "" {
		buffer.WriteString("?")
		buffer.WriteString(provider.cfg.QueryString)
	}
```
Do I miss something?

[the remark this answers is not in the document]
victorshevtsov: I tried the suggested approach and it didn't work without passing the same custom tls key that used to call `mysql.RegisterTLSConfig` method.
Unfortunately, that doesn't work because of the way how the mysql driver configures a secure connection:
https://pkg.go.dev/github.com/go-sql-driver/mysql#section-readme

image

image

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +86,92 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+
+	return buffer.String(), nil
+}
```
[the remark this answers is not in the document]
victorshevtsov: @bsardo thank you for the pointing out all the said above about MySql driver configuration.
I followed your suggested changes, and also added tests for the connection string with TLS support.

stored_requests/backends/db_provider/mysql_dbprovider.go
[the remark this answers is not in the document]
VeronikaSolovei9: Oh I see. Ok, I'll test it locally next time.

stored_requests/backends/db_provider/postgres_dbprovider.go
victorshevtsov: Moved to a URI format as per the preferred option of the suggested ones.

stored_requests/backends/db_provider/mysql_dbprovider.go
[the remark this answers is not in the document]
VeronikaSolovei9: I see why now! If this line was executed `mysql.RegisterTLSConfig(customTLSKey, &tls.Config{` then in connection string tls **must** be equal to `customTLSKey`. 
With current code if only `RootCert` is specified, then `tls=true` goes to connection string and it fails to connect to DB.
I tested this locally, it works with/without QueryString and with/without TLS:

```
func (provider *MySqlDbProvider) ConnString() (string, error) {
	buffer := bytes.NewBuffer(nil)

	if provider.cfg.Username != "" {
		buffer.WriteString(provider.cfg.Username)
		if provider.cfg.Password != "" {
			buffer.WriteString(":")
			buffer.WriteString(provider.cfg.Password)
		}
		buffer.WriteString("@")
	}

	buffer.WriteString("tcp(")
	if provider.cfg.Host != "" {
		buffer.WriteString(provider.cfg.Host)
	}

	if provider.cfg.Port > 0 {
		buffer.WriteString(":")
		buffer.WriteString(strconv.Itoa(provider.cfg.Port))
	}
	buffer.WriteString(")")

	buffer.WriteString("/")

	if provider.cfg.Database != "" {
		buffer.WriteString(provider.cfg.Database)
	}

	if provider.cfg.QueryString != "" {
		buffer.WriteString("?")
		buffer.WriteString(provider.cfg.QueryString)
	}

	if provider.cfg.TLS != nil {

		if err := setupTLSConfig(provider); err != nil {
			return "", err
		}
		tlsParam := "tls=" + customTLSKey

		addTls := true

		if provider.cfg.QueryString != "" {
			if strings.Contains(provider.cfg.QueryString, "tls=") {
				if strings.Contains(provider.cfg.QueryString, tlsParam) {
					//query string has correct tls param
					addTls = false
				} else {
					//error - tls name doesn't match
					return "", errors.New("Incorrect query string")
				}
			} else {
				//ad tls after query string
				buffer.WriteString("&")
				buffer.WriteString(tlsParam)
				addTls = false
			}
		}
		if addTls {
			//no query string, add tls only
			buffer.WriteString("?")
			buffer.WriteString(tlsParam)
		}
	}

	return buffer.String(), nil
}

```

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +86,97 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+	if tls != "" {
+		if len(provider.cfg.QueryString) == 0 {
+			return "tls=" + tls, nil
+		}
+		if !strings.Contains(provider.cfg.QueryString, "tls=") {
+			return "tls=" + tls + "&" + provider.cfg.QueryString, nil
+		}
+	}
+
+	return provider.cfg.QueryString, nil
```
VeronikaSolovei9, line 179: Please refer to thiscomment: https://github.com/prebid/prebid-server/pull/2507#discussion_r1114716636 
`tls` can only be `customTLSKey`.

victorshevtsov: @VeronikaSolovei9 thank you for pointing out this missed case.
Fixed and added corresponding tests.

bsardo: @victorshevtsov I think there still might be a hole in the logic. I don't think that it is necessarily true that `tls` can only ever be `customTLSKey`. I think that if `tls` is defined in the `query_string` we should use that value instead of using the default `customTLSKey`. The version of `generateQueryString` function I originally shared was written with that intent. The issue I see is that the first parameter passed to `RegisterTLSConfig` must match the constructed query string `tls` value and that may either be the `tls` value provided in `query_string` by the host or `customTLSKey` if `tls` was not provided in `query_string` by the host.

In that case, I think some adjustment to `generateQueryString` might fix this (I didn't test this):
```
func (provider *MySqlDbProvider) generateQueryString() (string, error) {
        // parse the query string
	queryParams, err := url.ParseQuery(provider.cfg.QueryString)
	if err != nil {
		return "", err
	}
        // determine if TLS is in the query string and get the value
	tlsValues, tlsInQueryStr := queryParams["tls"]
	tls := ""
	if len(tlsValues) > 1 {
		return "", error(TODO)
	}
	if len(tlsValues) == 1 {
		tls = tlsValues[0]
	}

        // if there is a root cert we know we want to use TLS
	if provider.cfg.TLS.RootCert != "" {
                // no TLS was found in the query string so use the default key
		if !tlsInQueryStr {
			tls = customTLSKey
		}

                 // pass the key (either from query string or the default) down to set up TLS
		if err := setupTLSConfig(provider, tls); err != nil {
			return "", err
		}
	}

	if tls != "" {
		if len(provider.cfg.QueryString) == 0 {
			return "tls=" + tls, nil
		}
		if !tlsInQueryStr {
			return "tls=" + tls + "&" + provider.cfg.QueryString, nil
		}
	}

        // I think this line would have returned the wrong string in the current code because if tls != "" but there was a query string already containing the key tls, we would have registered with customTLSKey but returned a string containing whatever tls was put in the query string by the host
	return provider.cfg.QueryString, nil
}
```

Thoughts?

victorshevtsov: How should we handle a case If the tls option is passed by the query string it has a special meaning as describe in the docs?
image

bsardo: Does PBS have to do anything special in these cases? It didn't seem like it to me. For the invalid / self-signed cert case it's on the host to correctly generate that and provide it as a root cert to PBS. The initial ping by PBS should fail if they set it up incorrectly.
Wouldn't these values also be passed as the first param to `RegisterTLSConfig`?
Are you just referring to validation of those values? I don't think I would worry about that because technically any name is valid.

victorshevtsov: Done.

VeronikaSolovei9 (commented):
Thank you for the changes!
Added minor comments. Local MySQL testing looks good with and without certificates. 
Local Postgres without certificates looks good too. I'm going to work on local Postgres certificates set up and Postgres connection testing.

VeronikaSolovei9:
While testing these changes with Postgres without certificates (our current prod set up) I hit this exact issue: 
https://github.com/jackc/pgx/issues/1285
Note: but it worked with my local PostgresDB! 
I reverted changes to one of your previous commits where a query string looks like `"user=postgres password=mypass12# host=127.0.0.1 port=5432 dbname=postgres"` and it worked!

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +86,105 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+
+		opts := x509.VerifyOptions{Roots: pool}
+		if _, err = cert.Verify(opts); err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+func (provider *MySqlDbProvider) generateQueryString() (string, error) {
```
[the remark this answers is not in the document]
victorshevtsov: Changed to call `setupTLSConfig` outside `generateQueryString`.

bsardo:
> @VeronikaSolovei9, what is your suggestion? Does revert of the commit [c6d3f21](https://github.com/prebid/prebid-server/pull/2507/commits/c6d3f213aabcfaf189370cf32895fe63d7937caa) solves the issue and makes the PR ready to go?

👀 I'd rather not roll back to using a KV-formatted string unless we absolutely have to because it establishes an inconsistency. We'll look into this a bit further and get back to you.

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -22,8 +29,12 @@ func (provider *MySqlDbProvider) Config() config.DatabaseConnection {
 }
 
 func (provider *MySqlDbProvider) Open() error {
```
bsardo, line 31: Just confirming, I assume you've tested this with a remote database instance right? We haven't done that on our end for MySQL; we only tested it locally.

The documentation does appear to indicate that only the `tls` query param is needed.

bsardo (commented):
We're getting close. Thanks again for your patience.

I'd also like to point out that there are a couple of query param values for MySQL that might require escaping as pointed out in [Go SQL Driver MySQL README parameters section](https://pkg.go.dev/github.com/go-sql-driver/mysql#readme-parameters). However, I'm not sure if what's listed there is necessary as [RFC3986 Section 3.4](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4) indicates the `/` need not be escaped inside the query string.

There may also be a few for postgres, particularly the password param options, as well.

For now, we could just put the onus on the host to specify an escaped query string so we can get this merged in sooner. Having PBS escape it can always be added in a follow-up PR. Thoughts?

bsardo:
> > @VeronikaSolovei9, what is your suggestion? Does revert of the commit [c6d3f21](https://github.com/prebid/prebid-server/pull/2507/commits/c6d3f213aabcfaf189370cf32895fe63d7937caa) solves the issue and makes the PR ready to go?
> 
> 👀 I'd rather not roll back to using a KV-formatted string unless we absolutely have to because it establishes an inconsistency. We'll look into this a bit further and get back to you.

Circling back here. See my suggestion in this [comment](https://github.com/prebid/prebid-server/pull/2507#discussion_r1132480050).

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +86,108 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+
+		opts := x509.VerifyOptions{Roots: pool}
+		if _, err = cert.Verify(opts); err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+func (provider *MySqlDbProvider) generateQueryString() (string, error) {
```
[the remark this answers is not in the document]
victorshevtsov: Yes, you are right. Fixed.

stored_requests/backends/db_provider/postgres_dbprovider.go
```diff
@@ -45,41 +49,74 @@ func (provider *PostgresDbProvider) Ping() error {
...
 	}
 
-	if provider.cfg.Database != "" {
-		buffer.WriteString("dbname=")
-		buffer.WriteString(provider.cfg.Database)
-		buffer.WriteString(" ")
+	return buffer.String(), nil
+}
+
+func (provider *PostgresDbProvider) generateQueryString() (string, error) {
```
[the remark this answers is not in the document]
victorshevtsov: Yes, you are right. Fixed.

stored_requests/backends/db_provider/mysql_dbprovider.go
```diff
@@ -75,7 +86,108 @@ func (provider *MySqlDbProvider) ConnString() string {
...
+func setupTLSConfig(provider *MySqlDbProvider) error {
+	rootCertPool := x509.NewCertPool()
+
+	pem, err := ioutil.ReadFile(provider.cfg.TLS.RootCert)
+	if err != nil {
+		return err
+	}
+
+	if ok := rootCertPool.AppendCertsFromPEM(pem); !ok {
+		return err
```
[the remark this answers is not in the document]
victorshevtsov: Fixed.

stored_requests/backends/db_provider/postgres_dbprovider.go
```diff
@@ -45,41 +49,74 @@ func (provider *PostgresDbProvider) Ping() error {
...
 
-func (provider *PostgresDbProvider) ConnString() string {
+func (provider *PostgresDbProvider) ConnString() (string, error) {
 	buffer := bytes.NewBuffer(nil)
+	buffer.WriteString("postgresql://")
+
+	if provider.cfg.Username != "" {
+		buffer.WriteString(provider.cfg.Username)
+		if provider.cfg.Password != "" {
+			buffer.WriteString(":")
```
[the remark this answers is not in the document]
victorshevtsov: I think so, as the postgresql documentation says:
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
image

stored_requests/backends/db_provider/postgres_dbprovider.go
```diff
@@ -45,41 +49,74 @@ func (provider *PostgresDbProvider) Ping() error {
...
+			buffer.WriteString(provider.cfg.Password)
```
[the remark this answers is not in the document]
victorshevtsov: Added percent encoding for postgers connection string.

Note: Tested MySQL connection using `PreBidPwd:/?#[]@!$&()*+,;=` as a password and PBS establishes a connection just fine without percent encoding, but fails when the password is encoded. So, I left MySQL as is without changes.

stored_requests/backends/db_provider/postgres_dbprovider_test.go
```diff
@@ -66,24 +75,124 @@ func TestConnStringPostgres(t *testing.T) {
...
+				password:    "somepassword",
+				QueryString: "param=value&sslmode=prefer",
+				TLS: TLS{
+					RootCert:   "root-cert.pem",
+					ClientCert: "client-cert.pem",
+					ClientKey:  "client-key.pem",
+				},
+			},
+			connString: "postgresql://someuser:somepassword@example.com:20/TestDB?sslrootcert=root-cert.pem&sslcert=client-cert.pem&sslkey=client-key.pem&param=value&sslmode=prefer",
 		},
```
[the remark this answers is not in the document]
victorshevtsov: Done.

stored_requests/backends/db_provider/postgres_dbprovider.go
```diff
@@ -45,41 +49,74 @@ func (provider *PostgresDbProvider) Ping() error {
...
+}
+
+func (provider *PostgresDbProvider) generateQueryString() (string, error) {
+	sslmode := "disable"
+	sslrootcert := ""
+	sslcert := ""
+	sslkey := ""
+	queryString := ""
+
+	if provider.cfg.TLS.RootCert != "" {
```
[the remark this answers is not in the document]
victorshevtsov: I would not overcomplicate the logic of the `generateQueryString()` function by combining different cases when at the same time some of the mentioned parameters passed by a `qyeryString` and the `tls` struct.

Perhaps, the easier solution is to fail if there is a mixture of the ways of how the parameters passed.
In other words, all the parameters (`sslrootcert`, `sslcert` and `sslkey`) have to be passed either by corresponding fields of the `tls` struct or by a `query_string` field of the `connection` struct, but not mixed.

If you agree with my suggestion please provide a proper error message for that case.

stored_requests/backends/db_provider/mysql_dbprovider.go
victorshevtsov: Yes, I've tested it with a remote database hosted at DigitalOcean and with a local database running in a docker container.

bsardo (commented):
Everything looks good from my perspective except for how we handle the scenario where TLS certs are in both the query string and config object. I agree with your suggested approach on that.

victorshevtsov:
Done.

VeronikaSolovei9 (approved):
Code and local testing looks good. Good work!

bsardo:
@victorshevtsov thanks again for your contribution. Nice work!

holms:
Thank you for great work everybody! 🍾

PR #2508, opened 2022-12-28
Rename synacormedia adapter to imds to reflect ownership change

ecammit:
This rename synacormedia adapter to imds to reflect ownership change, while leaving the synacormedia adapter in place referencing the imds adapter code to allow existing implementations to continue to function under the old name.

See also:
https://github.com/prebid/Prebid.js/pull/9381
https://github.com/prebid/prebid.github.io/pull/4237


guscarreon (dismissed):
LGTM

adapters/imds/imds.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/prebid/prebid-server/openrtb_ext"
 )
 
-type SynacorMediaAdapter struct {
+type ImdsAdapter struct {
```
SyntaxNode, line 17: Please name this type `adapter`. There is no need to repeat the namespace name. There is no need to export the type by starting with an uppercase character. It's likely the SynacorMedia adapter predated this advice.

static/bidder-info/imds.yaml
```diff
@@ -0,0 +1,18 @@
...
+    mediaTypes:
+      - banner
+      - video
+  site:
+    mediaTypes:
+      - banner
+      - video
+userSync:
+  # imds supports user syncing, but requires configuration by the host. contact this
+  # bidder directly at the email address in this file to ask about enabling user sync.
```
SyntaxNode, line 15: There is a default iframe user sync defined, but then you've added a comment that configuration is required. Should you remove the default endpoint in that case? I wouldn't expect a host company to read this message if user syncing works out of the box.

static/bidder-info/imds.yaml
```diff
@@ -0,0 +1,18 @@
...
+capabilities:
+  app:
+    mediaTypes:
+      - banner
+      - video
+  site:
+    mediaTypes:
+      - banner
+      - video
+userSync:
```
SyntaxNode, line 13: It is possible for two bidders to share the same user sync key (formerly called the "cookie family"). The default key is the bidder name, where imds and synacormedia are considered separate. You have three options:

- Keep the user sync keys separate. synacormedia would continue using the existing values and imds would start to build up a new identity map from scratch, separate and distinct from synacormedia. This takes extra space in the sync cookie.
- Use the imds key for both imds and synacormedia. imds would still start to build up a new identity map from scratch, but would be shared by both bidder names and use minimal space in the sync cookie.
- Use the synacormedia key for both and continue to build on the existing identity map.

We do not have a way for cleanly migrating from one user sync key to another.

ecammit: All set. Updated with changes.

static/bidder-info/imds.yaml
```diff
@@ -0,0 +1,21 @@
...
+  key: imds
```
guscarreon, line 14: I believe the key value must be surrounded by quotes
```
14   userSync:
15 -   key: imds
   +   key: "imds"
16     supportCors: true
```

static/bidder-info/synacormedia.yaml
```diff
@@ -11,6 +12,11 @@ capabilities:
       - banner
       - video
 userSync:
+  key: imds
```
guscarreon, line 15: I believe the key value must be surrounded by quotes
```
14   userSync:
15 -   key: imds
   +   key: "imds"
16     supportCors: true
```

guscarreon (dismissed):
Did a bit more research and apparently in YAML quotes are sort of optional and only really needed when we want to escape quotes or other characters. Sorry 😅 

LGTM

ecammit:
> Did a bit more research and apparently in YAML quotes are sort of optional and only really needed when we want to escape quotes or other characters. Sorry 😅
> 
> LGTM

Not a problem. Happy to quote them anyhow to be safe.

static/bidder-info/imds.yaml
```diff
@@ -11,8 +11,10 @@ capabilities:
       - banner
       - video
 userSync:
-  # imds supports user syncing, but requires configuration by the host. contact this
-  # bidder directly at the email address in this file to ask about enabling user sync.
+  cookie-family-name: imds
```
SyntaxNode: I believe `cookie-family-name` is what PBS-Java decided to call this value. In PBS-Go it is simply `key`. You don't need to specify the `key` in this file since it's the same as the default. You only need it for the synacormedia alias.

ecammit: Removed the `key` from imds yaml file.

guscarreon (approved):
LGTM

SyntaxNode (approved):

Issue #2479, opened 2022-12-05
Add instance level enable_events config


shahinrahbariasl:
Currently events such as imp and win  can be enabled in Prebid Server using stored accounts (by setting `enable_events`) on accounts level. This requires a new stored account to be created for every new account. However we can move `enable_events` config to instance level. That way if hosting PBS, events tracking can be enabled/disabled for all accounts. 

More info:
PBS event handler checks for events being enabled: https://github.com/prebid/prebid-server/blob/5cc5da91f517b79b6b7d2182994bb57d38813918/endpoints/events/event.go#L108

bretg:
@shahinrahbariasl - I'm ok with a host-level flag, but it would be overridden by account-level config. The order of precedence we use in PBS is

1. request-level params
2. account-level config
3. host-level config
4. hardcoded defaults

If you want to deviate from this convention for this use case, please explain the scenario

ShriprasadM:
@bretg  we are working on https://github.com/prebid/prebid-server/issues/1725. As a part of this thing we have introduced events.enabled flag and planning to deprecate accounts.events_enabled flag (As per #3 - https://docs.google.com/document/d/1tH6I9Vho-_KaUIGNCMum9JwbxHDc0FyG2wD8T-h3wkY/edit mentioned here))

shahinrahbariasl:
Hi @bretg and @ShriprasadM , I was thinking more like the other way which would be instance level flag taking precedence since the use case I'm trying to tackle is, when hosting PBS, we would want to enable events for all accounts without the need to update the existing stored accounts or create new ones. I had a draft PR which I just mentioned in this issue and the logic I was thinking was more like:
```
if !account.EventsEnabled && !e.Cfg.Event.Enabled {
		...
                // not enabled
		return
	}
```

If we want to make stored accounts override instance level, the current config is `enable_event` so I think that should be negated to have a config that disables events but that's a bit out of the scope that I was hoping for and also it would require changing the existing stored accounts. Or we could add a new field to stored account to disable events? but again, that'd be confusing with the existing `enable_event` account config.

With the above logic, in case someone needs to disable events for an account, they can then remove the instance level config and manage the events per account with stored accounts.

Let me know what you think or if I'm missing something.

shahinrahbariasl:
Ok was able to make stored account override instance level by changing stored account event flag to *bool instead of bool so that we can distinguish between nil and false values: https://github.com/prebid/prebid-server/pull/2509

[...]

PR #2509, opened 2022-12-29
Pb 1407 instance level event flag

shahinrahbariasl:
Addresses #2479.

Adds instance level config to enable PBS analytics events (`event.enabled`)

- Instance level config overrides account level config. This is because with the current `event_enable` config at account level, we need to create a stored account for every account that we want to enable events for. To avoid that, instance level config can override that. This way if hosting PBS, we can enable events for all accounts without the need for creating stored accounts for each. 
- When using PBS analytics events, I don't see many use cases that events would need to be disabled for an account, but if there's any, instance level `event.enabled` should be set to `false` or removed and events enablement should be managed through stored accounts then.

Example `pbs.yml` config to enable events at instance level:
```
event:
  enabled: true
```

Screen Shot 2022-12-16 at 11 55 22 AM

Screen Shot 2022-12-16 at 11 56 37 AM

Screen Shot 2022-12-16 at 11 56 32 AM

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -469,6 +469,7 @@ type VTrack struct {
 
 type Event struct {
 	TimeoutMS int64 `mapstructure:"timeout_ms"`
+	Enabled   bool  `mapstructure:"enabled"`
 }
 
 type HostCookie struct {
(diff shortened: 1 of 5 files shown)


mansinahar:
> Hey @bsardo, I'm not sure why the validate jobs failed for this. Do you have any idea? Thanks!

@shahinrahbariasl Looks like there's some formatting issue in `exchange/exchange.go`. Can you please run `go fmt` and upload the changes to this PR? That should likely fix it.

shahinrahbariasl:
> > Hey @bsardo, I'm not sure why the validate jobs failed for this. Do you have any idea? Thanks!
> 
> @shahinrahbariasl Looks like there's some formatting issue in `exchange/exchange.go`. Can you please run `go fmt` and upload the changes to this PR? That should likely fix it.

@mansinahar Thank you! I accidentally rebased this PR and that caused too many file changes for review so I closed this one and opened another PR: https://github.com/prebid/prebid-server/pull/2533

PR #2511, opened 2022-12-31
New Adapter: CWire

espen-j:
Introduces a new CWire prebid server adapter supporting ORTB2 protocol.

Documentation:
https://github.com/prebid/prebid.github.io/pull/4242


MatejaMaric:
Hey @SyntaxNode, can you re-run workflows? I fixed the issues, it should work now.

MatejaMaric:
@SyntaxNode Oops, it seems that I forgot to fix the tests as well... 😅 You can re-run workflows again, hopefully it will work now...

go.sum
```diff
@@ -29,18 +29,12 @@ cloud.google.com/go v0.94.1/go.mod h1:qAlAugsXlC+JWO+Bke5vCtc9ONxjQT3drlTTnAplMW
 cloud.google.com/go v0.97.0/go.mod h1:GF7l59pYBVlXQIBLx3a761cZ41F9bBH3JUlihCt2Udc=
 cloud.google.com/go v0.98.0/go.mod h1:ua6Ush4NALrHk5QXDWnjvZHN93OuF0HfuEPq9I1X0cM=
 cloud.google.com/go v0.99.0/go.mod h1:w0Xx2nLzqWJPuozYQX+hFfCSI8WioryfRDzkoI/Y2ZA=
-cloud.google.com/go v0.100.2/go.mod h1:4Xra9TjzAeYHrl5+oeLlzbM2k3mjVhZh4UqTZ//w99A=
```
guscarreon, line 32: Nitpick: Can we revert the changes made on this file? We try to keep dependency updates off adapter PRs unless strictly necessary.

static/bidder-info/cwire.yaml
```diff
@@ -0,0 +1,10 @@
+endpoint: "https://embed.cwi.re/prebid/ortb2"
+maintainer:
+  email: "mateja.maric@cwire.com"
+  endpointCompression: gzip
```
guscarreon, line 4: I believe `endpointCompression` should be at the same level of `endpoint` and `gvlVendorID` because `maintainer` doesn't have an `endpointCompression` field.

adapters/cwire/cwiretest/exemplary/banner.json
```diff
@@ -0,0 +1,125 @@
...
+    ],
+    "imp": [
+      {
+        "id": "102",
+        "banner": {
+          "h": 250,
+          "w": 300,
+          "pos": 0
+        }
+      }
```
guscarreon, line 15: Can we add an extra test case or JSON test file that comes with more `req.imp[i].ext` fields?

adapters/cwire/cwire.go
```diff
@@ -0,0 +1,105 @@
...
+func (a *CWireAdapter) MakeBids(bidReq *openrtb2.BidRequest, unused *adapters.RequestData, httpRes *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if httpRes.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	var resp openrtb2.BidResponse
+	if err := json.Unmarshal(httpRes.Body, &resp); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Error while decoding response, err: %s", err),
+		}}
```
guscarreon, line 90: Can we add more JSON test cases that cover this scenarios? One with a malformed response body and another with a `204` response

image

MatejaMaric:
Hey @guscarreon, thanks for the review! I'll see to it that all your recommendations are implemented.

adapters/cwire/cwire.go
```diff
@@ -0,0 +1,105 @@
...
+
+- The adapters.Builder method to create a new instance of the adapter based on
+  the host configuration
+
+- The adapters.Bidder interface consisting of the MakeRequests method to create
+  outgoing requests to your bidding server and the MakeBids method to create bid
+  responses.
+*/
+
+type CWireAdapter struct {
```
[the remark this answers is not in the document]
MatejaMaric: Done

go.sum
MatejaMaric: Done

static/bidder-info/cwire.yaml
MatejaMaric: Fixed

adapters/cwire/params_test.go
```diff
@@ -0,0 +1,53 @@
...
+		if err := validator.Validate(openrtb_ext.BidderCWire, json.RawMessage(invalidParam)); err == nil {
+			t.Errorf("Schema allowed unexpected params: %s", invalidParam)
+		}
+	}
+}
+
+var validParams = []string{
+	`{"placementId":123,"pageId":321}`,
+	`{"placementId":123,"pageId":321,"cwcreative":"3746"}`,
+}
```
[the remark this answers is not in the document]
MatejaMaric: Added

adapters/cwire/cwiretest/exemplary/banner.json
MatejaMaric: Added

adapters/cwire/cwire.go
MatejaMaric: Added

MatejaMaric:
Hey @guscarreon, I implemented your recommendations, can you review them? Thanks!

adapters/cwire/cwire.go
```diff
@@ -0,0 +1,105 @@
...
+
+This method is called for each response received from your bidding server
+within the bidding time window (request.tmax). If there are no requests or if
+all requests time out, the MakeBids method will not be called.
+*/
+func (a *adapter) MakeBids(bidReq *openrtb2.BidRequest, unused *adapters.RequestData, httpRes *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if httpRes.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
```
mansinahar, line 85: I'd recommend adding another check to make sure the returned response code was 200 and if it wasn't 200 then return an error appropriately before trying to unmarshal the response.

adapters/cwire/cwire.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := json.Unmarshal(httpRes.Body, &resp); err != nil {
+		return nil, []error{&errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Error while decoding response, err: %s", err),
+		}}
+	}
+
+	bidderResponse := adapters.NewBidderResponse()
+	bidderResponse.Currency = resp.Cur
+	for _, sb := range resp.SeatBid {
+		for _, bid := range sb.Bid {
```
mansinahar, line 96: Please use an index variable while iterating over `sb.Bid` because `bid` variable is getting a copy of each entry, not the entry itself. So when you take the address of the entry on L98, you end up with a pointer pointing at the LAST element in the `sb.Bid` slice (since the iteration stops at that point).

static/bidder-info/cwire.yaml
```diff
@@ -0,0 +1,10 @@
+endpoint: "https://ortb.cwi.re/v1/bid"
+maintainer:
+  email: "mateja.maric@cwire.com"
```
mansinahar, line 3: Please use a group email here. We strongly discourage using individual emails because there might be situations where a person say moves to a diff company, we would have no way to reach them.

static/bidder-params/cwire.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "cwfeatures": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "A string array of CWire features"
+    }
+  },
+  "required": []
```
mansinahar, line 31: In the docs I see that all these parameters are said to be required but here they aren't. Is that intentional?

adapters/cwire/cwire.go
MatejaMaric: Done

static/bidder-info/cwire.yaml
MatejaMaric: Fixed

static/bidder-params/cwire.json
MatejaMaric: Yeah, it's intentional. We also have Prebid.js adapter and it does require those fields (PBS adapter doesn't).
Anyway, my colleague is currently working on Prebid.js adapter and soon we should have those requirements aligned between the two.

mansinahar: @MatejaMaric Sounds good. In this case it sounds like you want the parameters to be optional in both PBJS and Prebid Server so this sounds fine but if you'd be looking at making any of these required any time in the future then we'd recommend making all of them required to start with and then move to making some of them optional based on your need. Because it's easier to go from required to optional but difficult to do the other way round without breaking existing setups of Publishers.

adapters/cwire/cwire.go
```diff
@@ -0,0 +1,111 @@
...
+This method is called for each response received from your bidding server
+within the bidding time window (request.tmax). If there are no requests or if
+all requests time out, the MakeBids method will not be called.
+*/
+func (a *adapter) MakeBids(bidReq *openrtb2.BidRequest, unused *adapters.RequestData, httpRes *adapters.ResponseData) (*adapters.BidderResponse, []error) {
+	if httpRes.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if httpRes.StatusCode != http.StatusOK {
```
mansinahar, line 86: @MatejaMaric Thanks for adding this check. I'd recommend updating the error message to 
```
fmt.Errorf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode)}
```
This will guide publishers do some troubleshooting on their end in case of any issues

adapters/cwire/cwire.go
```diff
@@ -0,0 +1,111 @@
...
+// Builder builds a new instance of the CWire adapter for the given bidder with the given config.
+func Builder(bidderName openrtb_ext.BidderName, config config.Adapter, server config.Server) (adapters.Bidder, error) {
+	bidder := &adapter{
+		endpoint: config.Endpoint,
+	}
+	return bidder, nil
+}
+
+/*
+The MakeRequests method is responsible for returning none, one, or many HTTP
```
mansinahar, line 39: This comment seems to be a copy paste of the interface definition of this function. Can you please modify it to just provide a description of the CWire implementation of `MakeRequests`? Similar comment for `MakeBids`

adapters/cwire/cwiretest/supplemental/badresponse.json
```diff
@@ -0,0 +1,89 @@
...
+        }
+      }
+    }
+  ],
+  "expectedMakeBidsErrors": [
+    {
+      "value": "Error while decoding response, err: .*",
+      "comparison": "regex"
+    }
+  ]
```
mansinahar, line 88: With the added check for any other non-200 responses, can you please also add a JSON test for a non-200 response that's not 204?

openrtb_ext/imp_cwire.go
```diff
@@ -0,0 +1,10 @@
+package openrtb_ext
+
+// ImpExtCwire defines the contract for MakeRequests `request.imp[i].ext.bidder`
+type ImpExtCWire struct {
```
mansinahar, line 4: I missed this in my review yesterday but you don't need this at all. The imp ext struct is only needed for adapters that unmarshal the imp ext to retrieve the bidder params and do some conversions based off of that before sending a value to their bidder service. In your case looks like you support openrtb2 format and are sending the original marshalled request as is so you can delete this definition completely.

mansinahar (commented):
@MatejaMaric Thanks for addressing the comments. I have a few more minor comments, can you please look at these when you have a chance? Once these are addressed, I should be happy to approve.

adapters/cwire/cwire.go
mansinahar: Yeah this is definitely a tricky one. Thanks for fixing this :)

adapters/cwire/cwire.go
MatejaMaric: Done

openrtb_ext/imp_cwire.go
MatejaMaric: It true that we don't need it currently, but it doesn't hurt to have it. If I remember correctly from the docs, it can be used to check if the parameters are malformed and if they are, we can ignore the impression (aka not make an HTTP call). We don't do that currently, but maybe we will. All in all, I personally think they also can be seen as a part of documentation just like `cwire.json`.

mansinahar (approved):

guscarreon (approved):
LGTM

PR #2514, opened 2023-01-04
IX Adapter: remove disable config from ix adapter configs

shahinrahbariasl:
Historically bidders were disabled by default. That's not the case anymore. This change is to remove `disabled` config from ix adapter configs.

More info: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#prebid-server---new-bid-adapter-go

### DIFF
--- a/static/bidder-info/ix.yaml
+++ b/static/bidder-info/ix.yaml
@@ -1,4 +1,3 @@
-disabled: true
 maintainer:
   email: "pdu-supply-prebid@indexexchange.com"
 gvlVendorID: 10


static/bidder-info/ix.yaml
```diff
@@ -1,4 +1,3 @@
-disabled: true
```
SyntaxNode, line 1: You must provide a default endpoint for an adapter to be enabled by default.

shahinrahbariasl: @SyntaxNode Thanks for the review, discussed internally, will close this PR for now.

Issue #2469, opened 2022-11-25
Price bucket calculation for Custom Price Granularity


Pubmatic-Dhruv-Sonone:
While using custom price granularity, prebid server return pb values lesser than min value of bucket
Example : For  the CPM =2 and  price granularity:
```
 {
 	Min:       0.0,
 	Max:       1.5,
 	Increment: 1.0,
 },
 {
 	Min:       1.5,
 	Max:       2.5,
 	Increment: 1.2,
 }
```
I expected pb value to be 1.5 but prebid-server calculates it as 1.2.
why is min bucket value not being taken into consideration for computing pb value?

Also Prebid-server seems to be using  different logic than prebid.js for price bucket calcuation.

Below is the logic used by Prebid.js.
```
var pow = Math.pow(10, precision + 2);
var cpmToFloor = (cpm * pow - bucketMin * pow) / (increment * pow);
var cpmTarget = Math.floor(cpmToFloor) * increment + bucketMin;
```

bretg:
You misunderstand min and max.

Custom price granularities are defined as ranges. So another way to write the example you posted is

- for prices between 0 and 1.5, round down to the nearest multiple of 1.0
- for prices between 1.5 and 2.5, round down to the nearest multiple of 1.2

Prebid Server is doing it correctly in this case, and I believe it's in sync with what Prebid.js is doing except that the javascript now supports an arbitrary javascript function for rounding, which is of course impossible in server. If you have an example where there's a difference, please post the details.

Also worth noting that "min" is not supported. See https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#targeting

```
Top end of this range of price buckets. The bottom end is 0 or the max of the previous bucket. Note: in order to prevent ranges with gaps, there’s no ‘min’ attribute.
```

bretg:
Ok, you're right that there's a difference. The Prebid.js cpmBucketManager.js implements getCpmStringValue saying: "start increments at the bucket min"

Do you have a real-world example where this causes a problem? We'd like to use it as another test case.

Since this difference has existed for 5 years, it's hard for me to consider this an urgent issue. We'll fix it, but can't currently promise a timeline.

To the PBS-Go and Java developers who take this on, here's the goal: make PBS resolve hb_pb the same way as Prebid.js.

With this config:
```
{
  "ext": {
    "prebid": {
      "targeting": {
        "pricegranularity": {
          "precision": 2,
          "ranges": [{
[...]

PR #2515, opened 2023-01-05
Changed price bucket calculation logic to start increment at bucketMin

Pubmatic-Dhruv-Sonone:
Fix for https://github.com/prebid/prebid-server/issues/2469

### DIFF
--- a/exchange/price_granularity.go
+++ b/exchange/price_granularity.go
@@ -13,7 +13,6 @@ func GetPriceBucket(cpm float64, config openrtb_ext.PriceGranularity) string {
 	bucketMax := 0.0
 	increment := 0.0
 	precision := config.Precision
-	bucketFloor := 0.0
 	bucketMin := 0.0
 	for i := 0; i < len(config.Ranges); i++ {
 		if config.Ranges[i].Max > bucketMax {
@@ -22,9 +21,7 @@ func GetPriceBucket(cpm float64, config openrtb_ext.PriceGranularity) string {
 		// find what range cpm is in
 		if cpm >= config.Ranges[i].Min && cpm <= config.Ranges[i].Max {
 			increment = config.Ranges[i].Increment
-			bucketMin = bucketFloor
-		} else {
-			bucketFloor = config.Ranges[i].Max
+			bucketMin = config.Ranges[i].Min
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


exchange/price_granularity.go
```diff
@@ -13,14 +13,18 @@ func GetPriceBucket(cpm float64, config openrtb_ext.PriceGranularity) string {
...
 	for i := 0; i < len(config.Ranges); i++ {
 		if config.Ranges[i].Max > bucketMax {
 			bucketMax = config.Ranges[i].Max
 		}
 		// find what range cpm is in
 		if cpm >= config.Ranges[i].Min && cpm <= config.Ranges[i].Max {
 			increment = config.Ranges[i].Increment
+			bucketMin = bucketFloor
+		} else {
+			bucketFloor = config.Ranges[i].Max
```
mansinahar, line 27: You don't need to keep track of the prev max and assign that to `bucketMin`. This is already done [here](https://github.com/prebid/prebid-server/blob/master/openrtb_ext/request.go#L241). You can just simply keep track of the smallest `bucketMin`, similar to how `bucketMax` is tracked and pass that value to the `getCpmTarget` function. With this change you won't need `bucketFloor` at all.

Something like:

```
	bucketMax := 0.0
	bucketMin := math.MaxFloat64
	increment := 0.0
	precision := config.Precision

	for i := 0; i < len(config.Ranges); i++ {
		if config.Ranges[i].Min < bucketMin {
			bucketMin = config.Ranges[i].Min
		}
		if config.Ranges[i].Max > bucketMax {
			bucketMax = config.Ranges[i].Max
		}
```

Pubmatic-Dhruv-Sonone: @mansinahar 
We want to keep track for the bucketMin  for the range from which increment is selected, not the smallest bucketMin. smallest bucketMin will always be zero as done [here](https://github.com/prebid/prebid-server/blob/f29682a87293010a346aeed4a5870b4fe4c37eee/openrtb_ext/request.go#L232).

mansinahar: @Pubmatic-Dhruv-Sonone Ah! You're right. That was my mistake. So you only need bucketMin = config.Ranges[i].Min on L25 and we can remove bucketFloor altogether as the validation code already makes sure that the previous max is set as min for the next range. Can you please make that change? Thanks!

Pubmatic-Dhruv-Sonone: Done. Made the change to set ` bucketMin = config.Ranges[i].Min`

hhhjort (approved):
LGTM

mansinahar (approved):

PR #2516, opened 2023-01-05
Modularity: Enrich Bid Response with hook execution results

4lexvav:


endpoints/openrtb2/amp_auction.go
```diff
@@ -289,31 +297,62 @@ func sendAmpResponse(
...
-					targets[key] = value
+	if response != nil {
+		for _, seatBids := range response.SeatBid {
+			for _, bid := range seatBids.Bid {
+				if bytes.Contains(bid.Ext, byteCache) {
+					// Looking for cache_id to be set, as this should only be set on winning bids (or
+					// deal bids), and AMP can only deliver cached ads in any case.
+					// Note, this could cause issues if a targeting key value starts with "hb_cache_id",
+					// but this is a very unlikely corner case. Doing this so we can catch "hb_cache_id"
+					// and "hb_cache_id_{deal}", which allows for deal support in AMP.
```
AlexBVolcy, line 308: Just want to confirm this comment is necessary or if it can be shortened at all?

hooks/hookexecution/enricher.go
```diff
@@ -0,0 +1,191 @@
...
+
+	if ext != nil {
+		response, err = jsonpatch.MergePatch(ext, response)
+	}
+
+	return response, err
+}
+
+// GetModulesJSON returns debug and trace information produced from executing hooks.
+//
```
AlexBVolcy, line 68: Nitpick: This empty comment line can be removed

endpoints/openrtb2/amp_auction.go
```diff
@@ -289,31 +297,62 @@ func sendAmpResponse(
...
+					bidExt := &openrtb_ext.ExtBid{}
+					err := json.Unmarshal(bid.Ext, bidExt)
+					if err != nil {
+						w.WriteHeader(http.StatusInternalServerError)
+						fmt.Fprintf(w, "Critical error while unpacking AMP targets: %v", err)
+						glog.Errorf("/openrtb2/amp Critical error unpacking targets: %v", err)
+						ao.Errors = append(ao.Errors, fmt.Errorf("Critical error while unpacking AMP targets: %v", err))
+						ao.Status = http.StatusInternalServerError
+						return labels, ao
```
AlexBVolcy, line 317: This chunk of code isn't covered by tests, can the test file be updated to cover this block?

endpoints/openrtb2/amp_auction.go
```diff
@@ -289,31 +297,62 @@ func sendAmpResponse(
...
+	// Fixes #231
+	enc := json.NewEncoder(w)
+	enc.SetEscapeHTML(false)
+
+	// If an error happens when encoding the response, there isn't much we can do.
+	// If we've sent _any_ bytes, then Go would have sent the 200 status code first.
+	// That status code can't be un-sent... so the best we can do is log the error.
+	if err := enc.Encode(ampResponse); err != nil {
+		labels.RequestStatus = metrics.RequestStatusNetworkErr
+		ao.Errors = append(ao.Errors, fmt.Errorf("/openrtb2/amp Failed to send response: %v", err))
```
AlexBVolcy, line 342: Can these lines of code be covered by tests?

endpoints/openrtb2/amp_auction.go
```diff
@@ -333,38 +372,34 @@ func sendAmpResponse(
...
-		} else {
-			glog.Errorf("Test set on request but debug not present in response.")
-			ao.Errors = append(ao.Errors, fmt.Errorf("test set on request but debug not present in response"))
+	if reqWrapper != nil {
+		if reqWrapper.Test == 1 && eRErr == nil {
+			if extResponse.Debug != nil {
+				extBidResponse.Debug = extResponse.Debug
+			} else {
+				glog.Errorf("Test set on request but debug not present in response.")
+				ao.Errors = append(ao.Errors, fmt.Errorf("test set on request but debug not present in response"))
```
AlexBVolcy, line 387: Can these lines be covered by tests?

metrics/prometheus/type_conversion.go
```diff
@@ -130,3 +130,17 @@ func tcfVersionsAsString() []string {
...
+
+	i := 0
+	for m, s := range moduleStageNames {
+		mods[i] = m
+		stages = append(stages, s...)
+		i++
+	}
+
+	return mods, stages
+}
```
bsardo, line 146: I don't see this used anywhere. Can we delete it?

amp/parse.go
```diff
@@ -31,6 +31,7 @@ type Params struct {
 	StoredRequestID   string
 	Targeting         string
 	Timeout           *uint64
+	Trace             string
```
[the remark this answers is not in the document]
4lexvav: No, it looks like in PBS-Java this parameter can only be set in the stored request. But you cannot modify it by the amp parameters retrieved from HTTP request. Do you think it is better to remove this ability from PBS-Go?

endpoints/openrtb2/amp_auction.go
```diff
@@ -42,10 +42,12 @@ const defaultAmpRequestTimeoutMillis = 900
...
-	Debug     *openrtb_ext.ExtResponseDebug                             `json:"debug,omitempty"`
-	Errors    map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage `json:"errors,omitempty"`
-	Warnings  map[openrtb_ext.BidderName][]openrtb_ext.ExtBidderMessage `json:"warnings,omitempty"`
+	Targeting map[string]string `json:"targeting"`
+	ORTB2     ORTB2             `json:"ortb2"`
+}
+
+type ORTB2 struct {
+	Ext openrtb_ext.ExtBidResponse `json:"ext"`
 }
```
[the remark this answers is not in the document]
4lexvav: Right, I wasn't aware of the mentioned issue, therefore I just changed the structure of the response, without introducing new fields.

endpoints/openrtb2/amp_auction.go
4lexvav: This is actually an old comment introduced by someone else. Within this PR I just slightly changed this part of code.
So unfortunately I'm not sure how important this comment is to decide to shorten it.

endpoints/openrtb2/amp_auction.go
4lexvav: I added a separate test for this code.

openrtb_ext/response.go
```diff
@@ -41,6 +41,7 @@ type ExtResponseSyncData struct {
 type ExtResponsePrebid struct {
 	AuctionTimestamp int64           `json:"auctiontimestamp,omitempty"`
 	Passthrough      json.RawMessage `json:"passthrough,omitempty"`
+	Modules          json.RawMessage `json:"modules,omitempty"`
```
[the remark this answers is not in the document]
4lexvav: Actually, at the moment I cannot remember any specific reason for choosing the `json.RawMessage` over a nested structure.
Most likely because the module data is only meant to be sent with a response so there is no need to maintain one more nested struct. Also, the value for the `Modules` field is set only after the `BidResponse` has already been created, thus we do not have access to the parsed `ExtBidResponse` structure and we deal with the `ext` field in JSON format: https://github.com/prebid/prebid-server/blob/9c9a40cf70da1375ab4feb4b475191bd52060005/endpoints/openrtb2/auction.go#L286

analytics/core.go
```diff
@@ -27,23 +28,25 @@ type PBSAnalyticsModule interface {
...
-	Response  *openrtb2.BidResponse
-	Account   *config.Account
-	StartTime time.Time
+	Status               int
+	Errors               []error
+	Request              *openrtb2.BidRequest
+	Response             *openrtb2.BidResponse
+	Account              *config.Account
+	StartTime            time.Time
+	HookExecutionOutcome []hookexecution.StageOutcome
```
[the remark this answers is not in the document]
4lexvav: Well, this part in PBS-Java is different from PBS-Go (different structure of objects, different fields), but the general concept looks similar: PBS-Java also sets hook execution outcomes to the object, which is later passed to analytics adapters.

hooks/hookexecution/outcome.go
```diff
@@ -51,7 +82,7 @@ type HookOutcome struct {
 	Status        Status                  `json:"status"`
 	Action        Action                  `json:"action"`
 	Message       string                  `json:"message"` // arbitrary string value returned from hook execution
-	DebugMessages []string                `json:"debug_messages"`
+	DebugMessages []string                `json:"debug_messages,omitempty"`
 	Errors        []string                `json:"-"`
 	Warnings      []string                `json:"-"`
```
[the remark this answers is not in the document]
4lexvav: That's right, they are only visible at the `prebid.modules.{errors/warning}` level.

metrics/prometheus/type_conversion.go
4lexvav: My fault, missed it during the merge. Deleted.

endpoints/openrtb2/auction_test.go
```diff
@@ -4667,29 +4667,37 @@ func TestValidResponseWhenRequestRejected(t *testing.T) {
...
-			planBuilder: mockPlanBuilder{rawBidderResponsePlan: makeRejectPlan[hookstage.RawBidderResponse](mockRejectionHook{nbr})},
+			file:        "sample-requests/hooks/auction_bidder_response_reject.json",
+			planBuilder: mockPlanBuilder{rawBidderResponsePlan: makePlan[hookstage.RawBidderResponse](mockRejectionHook{nbr})},
+		},
+		{
+			description: "Assert correct BidResponse with debug information from modules added to ext.prebid.modules",
+			file:        "sample-requests/hooks/auction.json",
+			planBuilder: mockPlanBuilder{
+				entrypointPlan: makeEntrypointPlan(),
+				rawAuctionPlan: makeRawAuctionPlan(),
```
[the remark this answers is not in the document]
4lexvav: Agree with you, a plan constructed in this way is self-descriptive.

hooks/hookexecution/enricher.go
```diff
@@ -0,0 +1,191 @@
...
+	return json.Marshal(modulesOutcome)
+}
+
+func getDebugContext(bidRequest *openrtb2.BidRequest, account *config.Account) (trace, bool) {
+	var traceLevel string
+	var isDebugEnabled bool
+
+	if bidRequest != nil {
+		traceLevel, _ = jsonparser.GetString(bidRequest.Ext, "prebid", "trace")
+		isDebug, _ := jsonparser.GetBoolean(bidRequest.Ext, "prebid", "debug")
```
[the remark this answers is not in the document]
4lexvav: Good point, I extended the return values of this function with warnings. Eventually, I add them as errors to `Auction/AmpObject`.

AlexBVolcy (approved):
LGTM

bsardo (approved):

PR #2519, opened 2023-01-06
Increase module plan timeouts and sleeps for flaky tests

bsardo:


SyntaxNode (approved):
Looks good for a short term fix. The long term fix will need to factor away dependence on a time.Sleep call.

hhhjort (approved):
LGTM. Agree with Scott.

PR #2524, opened 2023-01-10
Unruly: Add support for "siteId" bidder parameter (different casing)

arielunruly:
Currently, unruly.json supported only "siteid" but our ssp now supports "siteId" with a capital I.
We need to add this property to the unruly.json.

### DIFF
--- a/openrtb_ext/imp_unruly.go
+++ b/openrtb_ext/imp_unruly.go
@@ -1,6 +1,6 @@
 package openrtb_ext
 
 type ExtImpUnruly struct {
-	SiteIDOld int  `json:"siteid"`
-	SiteID int `json:"siteId"`
+	SiteIDOld int `json:"siteid"`
+	SiteID    int `json:"siteId"`
 }
--- a/openrtb_ext/imp_unruly.go
+++ b/openrtb_ext/imp_unruly.go
@@ -1,5 +1,6 @@
 package openrtb_ext
 
 type ExtImpUnruly struct {
+	SiteIDOld int  `json:"siteid"`
 	SiteID int `json:"siteId"`
 }
--- a/openrtb_ext/imp_unruly.go
+++ b/openrtb_ext/imp_unruly.go
@@ -1,5 +1,5 @@
 package openrtb_ext
 
 type ExtImpUnruly struct {
-	SiteID int `json:"siteid"`
+	SiteID int `json:"siteId"`
 }
--- a/static/bidder-params/unruly.json
+++ b/static/bidder-params/unruly.json
@@ -10,7 +10,7 @@
     },
     "siteId": {
       "type": "integer",
-      "description": "ID for publisher site"
+      "description": "ID for publisher site,preferred"
     }
   },
   "oneOf": [
--- a/static/bidder-params/unruly.json
+++ b/static/bidder-params/unruly.json
@@ -1,13 +1,20 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema#",
-    "title": "Unruly Adapter Params",
-    "description": "A schema which validates params accepted by the Unruly adapter",
-    "type": "object",
-    "properties": {
-      "siteid": {
-        "type": "integer",
-        "description": "ID for publisher site"
-      }
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Unruly Adapter Params",
+  "description": "A schema which validates params accepted by the Unruly adapter",
+  "type": "object",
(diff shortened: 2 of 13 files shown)


hhhjort:
Won't you drop `siteId` when you unmarshal your adapter parameters and then remarshal them?

AlexBVolcy:
I'm also wondering the same thing about `siteId` being dropped during unmarshal/marshaling

arielunruly:
Both siteId and siteid exist for backward compatibility.

hhhjort:
> Both siteId and siteid exist for backward compatibility.

Yes, but when you unmarshal into the struct, it will only see `siteid`, and drop `siteId`. Then when you remashal, only `siteid` can exist in the final JSON. You need to modify the struct to have both fields if you want to be able to pass both possibilities through your adapter.

hhhjort:
You should also add a test with the alternate case to confirm the code works and generates the expected request to your servers.

arielunruly:
@hhhjort , I have updated the struct and mocks.

openrtb_ext/imp_unruly.go
```diff
@@ -1,5 +1,5 @@
 package openrtb_ext
 
 type ExtImpUnruly struct {
-	SiteID int `json:"siteid"`
```
hhhjort, line 4: You want to keep the old `siteid` entry if you want to keep supporting the old configuration. (Highly recommended for atleast a transition period). So something like:
```
type ExtImpUnruly struct {
	SiteIDOld int `json:"siteid"`
        SiteID int `json:"siteId"`
}
```
And probably have some of each type in your tests to verify that both the old and the new casings work with the current code.

AlexBVolcy: Yeah agreed, having both the old configuration and the new casing is what I'd recommend.

static/bidder-params/unruly.json
```diff
@@ -1,13 +1,20 @@
...
+  "type": "object",
+  "properties": {
+    "siteid": {
+      "type": "integer",
+      "description": "ID for publisher site"
     },
-  "required": ["siteid"]
+    "siteId": {
+      "type": "integer",
+      "description": "ID for publisher site"
```
SyntaxNode, line 13: Which casing do you prefer to use? We have a convention of ending the description of that property with `, preferred.` as a hint to publishers.

arielunruly:
I see the tests passed except the Code scanning results Trivy, i am sorry but i haven't touched this part. what should i do ? pull again from master?

static/bidder-params/unruly.json
arielunruly: Do you mean to add a `,` after the end of the `description`?

bsardo:
@arielunruly nothing is required on your part. We recently merged a change into master that will make the Trivy scan pass. We'll get another reviewer on this shortly. Thanks for your patience.

arielunruly:
Hello @bsardo , any news with the code scanning results? :-)

static/bidder-params/unruly.json
AlexBVolcy: I think @SyntaxNode is wondering if you could update the description of the bidder param you prefer between `siteId` and `siteid`

So if `siteId` is the preferred casing, you can update its description to be like `"ID for publisher site, preferred"`

AlexBVolcy:
I'll be ready to approve after you update the description of your preferred bidder param!

Sonali-More-Xandr (commented):
Can you please also update the [dev-docs](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/unruly.md) to reflect these changes in the parameter

arielunruly:
> Can you please also update the [dev-docs](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/unruly.md) to reflect these changes in the parameter

updated, please advise.

AlexBVolcy:
I'm not seeing any updates to the dev-docs, or a PR for one? Could you link the changes here?

arielunruly:
> dev-docs

yes i think i forgot to do a PR , here is the link now : 
https://github.com/prebid/prebid.github.io/pull/4381

arielunruly:
> I'm not seeing any updates to the dev-docs, or a PR for one? Could you link the changes here?

I guess i forgot to do a PR , here is the link now : 
https://github.com/prebid/prebid.github.io/pull/4381

Sonali-More-Xandr (approved):

AlexBVolcy (approved):

PR #2529, opened 2023-01-12
Bid Response Validation Test Fix

AlexBVolcy:
As pointed out by @laurb9 in the Bid Response Validation PR https://github.com/prebid/prebid-server/pull/2450, on line 2302 in `exchange/exchange_test.go`, the validation flag that is used for the JSON tests, wasn't being set. So we were getting a false pass in the tests, without actually asserting what's within the validation flag code block.

The reason it wasn't being set is because I changed how I wanted my validation flag to work, and implemented this change elsewhere in this test file, but forgot to change this line. To fix this, I updated the line so that the proper validation object is being utilized.

This validation object is set based on the info that's present in the `bid_response` json files specified at `host_bid_validations` within the json file.

### DIFF
--- a/exchange/exchange_test.go
+++ b/exchange/exchange_test.go
@@ -2299,7 +2299,7 @@ func runSpec(t *testing.T, filename string, spec *exchangeSpec) {
 
 	}
 
-	if spec.BidValidationEnforcement == config.ValidationEnforce {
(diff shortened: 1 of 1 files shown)


VeronikaSolovei9 (approved):

guscarreon (approved):
LGTM

PR #2533, opened 2023-01-19
Deprecate events_enabled in favor of events.enabled

shahinrahbariasl:
Addresses #2479.

Adds instance level config to enable PBS analytics events (`event.enabled`)

- Instance level config overrides account level config. This is because with the current `event_enable` config at account level, we need to create a stored account for every account that we want to enable events for. To avoid that, instance level config can override that. This way if hosting PBS, we can enable events for all accounts without the need for creating stored accounts for each. 
- When using PBS analytics events, I don't see many use cases that events would need to be disabled for an account, but if there's any, instance level `event.enabled` should be removed and events enablement should be managed through stored accounts then.

Example `pbs.yml` config to enable events at instance level:
```
event:
  enabled: true
```

Note: had to make stored account event enable flag pointer bool in order to be able to distinguish between nil stored account (in case stored account is not defined for the account therefore its enable flag being nil) and stored account with enable event flag set to false (for example, in case overriding instance level flag).

Screen Shot 2022-12-16 at 11 55 22 AM

Screen Shot 2022-12-16 at 11 56 37 AM

Screen Shot 2022-12-16 at 11 56 32 AM

### DIFF
--- a/account/account.go
+++ b/account/account.go
@@ -261,8 +261,11 @@ func useCCPAChannelEnabled(account *config.Account) bool {
 // This function favors "events.enabled" field over deprecated "events_enabled" field, if values for both are set.
 // If only deprecated "events_enabled" field is set then it sets the same value to "events.enabled" field.
 func deprecateEventsEnabledField(account *config.Account) {
-
 	if account != nil && account.Events.Enabled == nil {
 		account.Events.Enabled = account.EventsEnabled
+	}
+	if account != nil {
+		// assign the old value to the new value so old and new are always the same even though the new value is what is used in the application code.
+		account.EventsEnabled = account.Events.Enabled
 	}
 }
--- a/account/account.go
+++ b/account/account.go
@@ -102,6 +102,10 @@ func GetAccount(ctx context.Context, cfg *config.Configuration, fetcher stored_r
 		})
 		return nil, errs
 	}
+
+	// set the value of events.enabled field based on deprecated events_enabled field and ensure backward compatibility
+	deprecateEventsEnabledField(account)
+
 	return account, nil
 }
 
@@ -251,4 +255,14 @@ func useGDPRChannelEnabled(account *config.Account) bool {
 
 func useCCPAChannelEnabled(account *config.Account) bool {
 	return account.CCPA.ChannelEnabled.IsSet() && !account.CCPA.IntegrationEnabled.IsSet()
+}
+
+// deprecateEventsEnabledField is responsible for ensuring backwards compatibility of "events_enabled" field.
+// This function favors "events.enabled" field over deprecated "events_enabled" field, if values for both are set.
+// If only deprecated "events_enabled" field is set then it sets the same value to "events.enabled" field.
+func deprecateEventsEnabledField(account *config.Account) {
+
+	if account != nil && account.Events.Enabled == nil {
+		account.Events.Enabled = account.EventsEnabled
+	}
 }
--- a/account/account.go
+++ b/account/account.go
@@ -261,10 +261,10 @@ func useCCPAChannelEnabled(account *config.Account) bool {
 // This function favors "events.enabled" field over deprecated "events_enabled" field, if values for both are set.
 // If only deprecated "events_enabled" field is set then it sets the same value to "events.enabled" field.
 func deprecateEventsEnabledField(account *config.Account) {
-	if account != nil && account.Events.Enabled == nil {
-		account.Events.Enabled = account.EventsEnabled
-	}
 	if account != nil {
+		if account.Events.Enabled == nil {
+			account.Events.Enabled = account.EventsEnabled
+		}
 		// assign the old value to the new value so old and new are always the same even though the new value is what is used in the application code.
 		account.EventsEnabled = account.Events.Enabled
 	}
--- a/account/account_test.go
+++ b/account/account_test.go
@@ -582,14 +582,38 @@ func TestAccountUpgradeStatusGetAccount(t *testing.T) {
 }
 
 func TestDeprecateEventsEnabledField(t *testing.T) {
-
 	testCases := []struct {
 		name    string
 		account *config.Account
 		want    *bool
 	}{
 		{
-			name: "events.enabled is nil",
+			name:    "account is nil",
+			account: nil,
+			want:    nil,
+		},
+		{
+			name: "account.EventsEnabled is nil, account.Events.Enabled is nil",
+			account: &config.Account{
+				EventsEnabled: nil,
+				Events: config.Events{
+					Enabled: nil,
+				},
+			},
+			want: nil,
+		},
+		{
+			name: "account.EventsEnabled is nil, account.Events.Enabled is non-nil",
+			account: &config.Account{
+				EventsEnabled: nil,
+				Events: config.Events{
+					Enabled: ptrutil.ToPtr(true),
+				},
+			},
+			want: ptrutil.ToPtr(true),
+		},
+		{
+			name: "account.EventsEnabled is non-nil, account.Events.Enabled is nil",
 			account: &config.Account{
 				EventsEnabled: ptrutil.ToPtr(true),
 				Events: config.Events{
@@ -599,12 +623,12 @@ func TestDeprecateEventsEnabledField(t *testing.T) {
 			want: ptrutil.ToPtr(true),
 		},
 		{
-			name: "events.enabled is non-nil",
+			name: "account.EventsEnabled is non-nil, account.Events.Enabled is non-nil",
 			account: &config.Account{
+				EventsEnabled: ptrutil.ToPtr(false),
 				Events: config.Events{
 					Enabled: ptrutil.ToPtr(true),
 				},
-				EventsEnabled: ptrutil.ToPtr(false),
 			},
 			want: ptrutil.ToPtr(true),
 		},
@@ -613,7 +637,9 @@ func TestDeprecateEventsEnabledField(t *testing.T) {
 	for _, test := range testCases {
 		t.Run(test.name, func(t *testing.T) {
 			deprecateEventsEnabledField(test.account)
-			assert.Equal(t, test.want, test.account.Events.Enabled, test.name)
+			if test.account != nil {
+				assert.Equal(t, test.want, test.account.Events.Enabled)
+			}
 		})
 	}
 }
--- a/account/account_test.go
+++ b/account/account_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/prebid/prebid-server/metrics"
 	"github.com/prebid/prebid-server/openrtb_ext"
 	"github.com/prebid/prebid-server/stored_requests"
+	"github.com/prebid/prebid-server/util/ptrutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
@@ -576,6 +577,43 @@ func TestAccountUpgradeStatusGetAccount(t *testing.T) {
 				_, _ = GetAccount(context.Background(), cfg, fetcher, accountID, metrics)
 			}
 			metrics.AssertNumberOfCalls(t, "RecordAccountUpgradeStatus", test.expectedMetricCount)
+		})
+	}
+}
+
+func TestDeprecateEventsEnabledField(t *testing.T) {
+
+	testCases := []struct {
+		name    string
+		account *config.Account
+		want    *bool
+	}{
+		{
+			name: "events.enabled is nil",
+			account: &config.Account{
+				EventsEnabled: ptrutil.ToPtr(true),
+				Events: config.Events{
+					Enabled: nil,
+				},
+			},
+			want: ptrutil.ToPtr(true),
+		},
+		{
+			name: "events.enabled is non-nil",
+			account: &config.Account{
+				Events: config.Events{
+					Enabled: ptrutil.ToPtr(true),
+				},
+				EventsEnabled: ptrutil.ToPtr(false),
+			},
+			want: ptrutil.ToPtr(true),
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			deprecateEventsEnabledField(test.account)
+			assert.Equal(t, test.want, test.account.Events.Enabled, test.name)
 		})
 	}
 }
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -22,7 +22,6 @@ const (
 )
 
 // Account represents a publisher account configuration
-// EventsEnabled is defined as pointer bool so that we can distinguish between nil and false.
 type Account struct {
 	ID                      string                               `mapstructure:"id" json:"id"`
 	Disabled                bool                                 `mapstructure:"disabled" json:"disabled"`
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -22,11 +22,12 @@ const (
 )
 
 // Account represents a publisher account configuration
+// EventsEnabled is defined as pointer bool so that we can distinguish between nil and false.
 type Account struct {
 	ID                      string                               `mapstructure:"id" json:"id"`
 	Disabled                bool                                 `mapstructure:"disabled" json:"disabled"`
 	CacheTTL                DefaultTTLs                          `mapstructure:"cache_ttl" json:"cache_ttl"`
-	EventsEnabled           bool                                 `mapstructure:"events_enabled" json:"events_enabled"`
+	EventsEnabled           *bool                                `mapstructure:"events_enabled" json:"events_enabled"`
 	CCPA                    AccountCCPA                          `mapstructure:"ccpa" json:"ccpa"`
 	GDPR                    AccountGDPR                          `mapstructure:"gdpr" json:"gdpr"`
 	DebugAllow              bool                                 `mapstructure:"debug_allow" json:"debug_allow"`
--- a/config/accounts.go
+++ b/config/accounts.go
@@ -26,7 +26,7 @@ type Account struct {
 	ID                      string                               `mapstructure:"id" json:"id"`
 	Disabled                bool                                 `mapstructure:"disabled" json:"disabled"`
 	CacheTTL                DefaultTTLs                          `mapstructure:"cache_ttl" json:"cache_ttl"`
-	EventsEnabled           *bool                                `mapstructure:"events_enabled" json:"events_enabled"`
+	EventsEnabled           *bool                                `mapstructure:"events_enabled" json:"events_enabled"` // Deprecated: Use events.enabled instead.
 	CCPA                    AccountCCPA                          `mapstructure:"ccpa" json:"ccpa"`
 	GDPR                    AccountGDPR                          `mapstructure:"gdpr" json:"gdpr"`
 	DebugAllow              bool                                 `mapstructure:"debug_allow" json:"debug_allow"`
--- a/config/config.go
+++ b/config/config.go
@@ -1380,7 +1380,8 @@ func migrateConfigDatabaseConnection(v *viper.Viper) {
 
 // migrateConfigEventsEnabled is responsible for ensuring backward compatibility of events_enabled field.
 // This function copies the value of newField "events.enabled" and set it to the oldField "events_enabled".
-// This is necessary to achieve the desired order of precedence favoring the account values over the host values given the account fetcher JSON merge mechanics.
+// This is necessary to achieve the desired order of precedence favoring the account values over the host values
+// given the account fetcher JSON merge mechanics.
 func migrateConfigEventsEnabled(oldFieldValue *bool, newFieldValue *bool) (updatedOldFieldValue, updatedNewFieldValue *bool) {
 	newField := "account_defaults.events.enabled"
 	oldField := "account_defaults.events_enabled"
--- a/config/config.go
+++ b/config/config.go
@@ -17,7 +17,6 @@ import (
 	"github.com/prebid/go-gdpr/consentconstants"
 	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/openrtb_ext"
-	"github.com/prebid/prebid-server/util/ptrutil"
 )
 
 // Configuration specifies the static application config.
@@ -687,10 +686,15 @@ func New(v *viper.Viper, bidderInfos BidderInfos, normalizeBidderName func(strin
 	// Update account defaults and generate base json for patch
 	c.AccountDefaults.CacheTTL = c.CacheURL.DefaultTTLs // comment this out to set explicitly in config
 
-	// update deprecated and new events enabled values for account defaults
+	// Update the deprecated and new events enabled values for account defaults.
+	// Given the current fetchers behavior where a fetched account is JSON merge patched with the host config account
+	// defaults, in order to maintain the typical PBS order of precedence of account config over host config, we do
+	// some unorthodox work here to set the resolved host config events enabled value on the deprecated field only while
+	// clearing the new field. This ensures that when the fetcher performs a JSON merge patch operation we don't lose any
+	// information needed to determine if events should be enabled.
 	accountDefaultsEventsEnabled := IsAccountEventEnabled(c.AccountDefaults.EventsEnabled, c.AccountDefaults.Events.Enabled)
 	*c.AccountDefaults.EventsEnabled = accountDefaultsEventsEnabled
-	c.AccountDefaults.Events.Enabled = ptrutil.ToPtr(false)
+	c.AccountDefaults.Events.Enabled = nil
 
 	if err := c.MarshalAccountDefaults(); err != nil {
 		return nil, err
--- a/config/config.go
+++ b/config/config.go
@@ -1487,21 +1487,11 @@ func isValidCookieSize(maxCookieSize int) error {
 // the `account.events_enabled` field, which is deprecated, is still used. We want to use its value whenever
 // the `account.events.enabled` field is not found.
 func IsAccountEventEnabled(deprecated_events_enabled *bool, events_enabled *bool) bool {
-	if deprecated_events_enabled != nil {
-		if *deprecated_events_enabled {
-			return true // case 4, 5, 6
-		} else {
-			if events_enabled != nil {
-				return *events_enabled // case 2, 8
-			} else {
-				return false // case 3
-			}
-		}
+	if events_enabled != nil {
+		return *events_enabled
+	} else if deprecated_events_enabled != nil {
+		return *deprecated_events_enabled
 	} else {
-		if events_enabled != nil {
-			return *events_enabled // case 1, 7
-		} else {
-			return false // case 9
-		}
+		return false
 	}
 }
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,7 @@ import (
 	"github.com/prebid/go-gdpr/consentconstants"
 	"github.com/prebid/prebid-server/errortypes"
 	"github.com/prebid/prebid-server/openrtb_ext"
+	"github.com/prebid/prebid-server/util/ptrutil"
 )
 
 // Configuration specifies the static application config.
@@ -689,8 +690,7 @@ func New(v *viper.Viper, bidderInfos BidderInfos, normalizeBidderName func(strin
 	// update deprecated and new events enabled values for account defaults
 	accountDefaultsEventsEnabled := IsAccountEventEnabled(c.AccountDefaults.EventsEnabled, c.AccountDefaults.Events.Enabled)
 	*c.AccountDefaults.EventsEnabled = accountDefaultsEventsEnabled
-	boolFalse := false
-	c.AccountDefaults.Events.Enabled = &boolFalse
+	c.AccountDefaults.Events.Enabled = ptrutil.ToPtr(false)
 
 	if err := c.MarshalAccountDefaults(); err != nil {
 		return nil, err
@@ -1483,9 +1483,9 @@ func isValidCookieSize(maxCookieSize int) error {
 	return nil
 }
 
-// IsAccountEventEnabled checks if event is enabled for account or not.
-// account.events_enabled/accountDefaults.events_enabled, which is deprecated, is still in use in some incoming requests so we'd like to still use its value whenever account.events.enabled/accountDefaults.events.enabled is not found.
-// Once we've identified if events are enabled or not, we can set the resulting boolean value into account.Events.Enabled/accountDefaults.Events.Enabled and set account.EventsEnabled/accountDefaults.EventsEnabled to nil.
+// IsAccountEventEnabled checks whether events are enabled for an account. In some incoming requests,
+// the `account.events_enabled` field, which is deprecated, is still used. We want to use its value whenever
+// the `account.events.enabled` field is not found.
 func IsAccountEventEnabled(deprecated_events_enabled *bool, events_enabled *bool) bool {
 	if deprecated_events_enabled != nil {
 		if *deprecated_events_enabled {
--- a/config/config.go
+++ b/config/config.go
@@ -145,7 +145,7 @@ func (cfg *Configuration) validate(v *viper.Viper) []error {
 	}
 
 	if len(cfg.AccountDefaults.Events.VASTEvents) > 0 {
-		errs = append(errs, errors.New("account_defaults.Events.VASTEvents will currently not do anything as the feature is still under development. Please follow https://github.com/prebid/prebid-server/issues/1725 for more updates"))
+		errs = append(errs, errors.New("account_defaults.Events.VASTEvents has no effect as the feature is under development."))
 	}
 
 	errs = cfg.Experiment.validate(errs)
@@ -1386,6 +1386,7 @@ func migrateConfigEventsEnabled(oldFieldValue *bool, newFieldValue *bool) (updat
 	newField := "account_defaults.events.enabled"
 	oldField := "account_defaults.events_enabled"
 
+	updatedOldFieldValue = oldFieldValue
 	if oldFieldValue != nil {
 		glog.Warningf("%s is deprecated and should be changed to %s", oldField, newField)
 	}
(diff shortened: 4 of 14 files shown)


bretg:
In the interest of open source harmony, I've updated the [PRD](https://docs.google.com/document/d/1a-U1DytiCTtacM27qH3OZwhPa7bDwQzfpVRHDhyikWw/edit#) to add Req 2.1.19


> This is because with the current event_enable config at account level, we need to create a stored account for every account that we want to enable events for. 

The assumption is that the majority of PBS host companies run in "validate account ID" mode, which requires creation of account info per customer anyhow.

> When using PBS analytics events, I don't see many use cases that events would need to be disabled for an account

Prebid's philosophy has been that publishers should be in control of what happens on their pages. In many scenarios, analytics is taking place from javascript analytics adapters, so by globally enabling events on the server, you run the risk of double-counting wins. In addition, events are an extra request the browser has to invoke, and pubs should be in control of when that happens.

> Instance level config overrides account level config. 

This is unacceptable. The standard is that account-level config always overrides host-level config, and this directly conflicts with the PRD
```
enabled: enable/disable event tracking for this account/installation.
Overrides the configured host level default.
```

The way PBS-Java has implemented account config is that there's a `default-account-config` . This is specifically for account defaults of all types so account-level overrides can be constrained to exceptions from the default. That's a reasonable place to put this flag. I agree that the disjoined `events-enabled` vs `enabled` seems unnecessary, so I've removed the `events` in the default.

shahinrahbariasl:
Hi @bretg, thank you for your detailed feedback! Instance level flag overriding account level was from my previous approach in a different PR which I closed, before I learnt about account level standard. That's updated now. 

Regarding account default configs, seems like that feature is not available in PBS-GO yet if I'm not mistaken https://github.com/prebid/prebid-server/blob/da478e319ea6c2c4a723fc752a150de401909535/config/config.go#L136

In PBS-GO, because `account.EventsEanbled` flag is a boolean, if it doesn't exist it evaluates to false therefore /event calls getting rejected https://github.com/prebid/prebid-server/blob/5cc5da91f517b79b6b7d2182994bb57d38813918/endpoints/events/event.go#L108

Hmm, but even if we have default account configs I'd assume by default we have to make events disabled based on the Prebid philosophy you mentioned and possible double-counting of wins. Or maybe we can come up with a solution to avoid that.

hhhjort (approved):
This is looking good. A relatively simple change to enable better account level control.

exchange/events.go
```diff
@@ -53,7 +54,7 @@ func (ev *eventTracking) modifyBidsForEvents(seatBids map[openrtb_ext.BidderName
 
 // isModifyingVASTXMLAllowed returns true if this bidder config allows modifying VAST XML for event tracking
 func (ev *eventTracking) isModifyingVASTXMLAllowed(bidderName string) bool {
-	return ev.bidderInfos[bidderName].ModifyingVastXmlAllowed && (ev.enabledForAccount || ev.enabledForRequest)
+	return ev.bidderInfos[bidderName].ModifyingVastXmlAllowed && ((ev.enabledForAccount != nil && *ev.enabledForAccount) || ev.enabledForRequest || ev.enabledForInstance)
```
guscarreon, line 57: Under the current logic, if `ev.enabledForAccount` is `false` and `ev.enabledForInstance` is `true`, the `ev.enabledForInstance` `true` value prevails despite being lower in the order of precedence @bretg mentioned in issue #2479:
1. request-level params
2. account-level config
3. host-level config
4. hardcoded defaults

I believe we should swap `ev.enabledForInstance` to the forefront of the second part of the `&&` operand and and make sure `ev.enabledForAccount` equals `nil` so the short-circuit OR doesn't omit an account override. In other words:
```
55   // isModifyingVASTXMLAllowed returns true if this bidder config allows modifying VAST XML for event tracking
56   func (ev *eventTracking) isModifyingVASTXMLAllowed(bidderName string) bool {
57 -     return ev.bidderInfos[bidderName].ModifyingVastXmlAllowed && ((ev.enabledForAccount != nil && *ev.enabledForAccount) || ev.enabledForRequest || ev.enabledForInstance)
   +     return ev.bidderInfos[bidderName].ModifyingVastXmlAllowed && ((ev.enabledForAccount == nil && ev.enabledForInstance) || (ev.enabledForAccount != nil && *ev.enabledForAccount) || ev.enabledForRequest)
58   }
exchange/events.go
```
Also, given that this boolean expression is used in three different places, can we put it in it's own function and write a table-driven test case for it?
```
55   // isModifyingVASTXMLAllowed returns true if this bidder config allows modifying VAST XML for event tracking
56   func (ev *eventTracking) isModifyingVASTXMLAllowed(bidderName string) bool {
57 -     return ev.bidderInfos[bidderName].ModifyingVastXmlAllowed && ((ev.enabledForAccount != nil && *ev.enabledForAccount) || ev.enabledForRequest || ev.enabledForInstance)
   +     return ev.bidderInfos[bidderName].ModifyingVastXmlAllowed && isEventAllowed(ev.enabledForRequest, ev.enabledForAccount, ev.enabledForInstance)
58   }
   + 
   + func isEventAllowed(requestEnabled bool, accountEnabled *bool, hostEnabled bool) bool {
   +     return (accountEnabled == nil && hostEnabled) ||  (accountEnabled != nil && *accountEnabled) || requestEnabled
   + }
exchange/events.go
```

config/accounts.go
```diff
@@ -21,11 +21,12 @@ const (
 )
 
 // Account represents a publisher account configuration
+// EventsEnabled is defined as pointer bool so that we can distinguish between nil and false.
 type Account struct {
```
guscarreon, line 25: Can we please add a test case similar to `TestMigrateConfigDatabaseQueryParams` found inside `config/config_test.go`  to assert that this `Account` struct gets unmarshalled by Viper correctly?

endpoints/events/event.go
```diff
@@ -104,11 +104,19 @@ func (e *eventEndpoint) Handle(w http.ResponseWriter, r *http.Request, _ httprou
...
-		return
+	// account does not support events and events are not enabled at instance level (account config takes precedence)
+	if account.EventsEnabled != nil {
+		if !*account.EventsEnabled {
+			w.WriteHeader(http.StatusUnauthorized)
+			w.Write([]byte(fmt.Sprintf("Account '%s' doesn't support events", eventRequest.AccountID)))
+			return
+		}
+	} else {
+		if !e.Cfg.Event.Enabled {
```
guscarreon, line 115: Nitpick: can we put the conditional in an `else if` to avoid nesting?
```
117 -   } else {
118 -       if !e.Cfg.Event.Enabled {
    +   } else if !e.Cfg.Event.Enabled {
119         w.WriteHeader(http.StatusUnauthorized)
120         w.Write([]byte(fmt.Sprintf("Account '%s' or server doesn't support events", eventRequest.AccountID)))
121         return
122 -       }
123     }
endpoints/events/event.go
```

config/config.go
```diff
@@ -470,6 +470,7 @@ type VTrack struct {
 
 type Event struct {
 	TimeoutMS int64 `mapstructure:"timeout_ms"`
+	Enabled   bool  `mapstructure:"enabled"`
```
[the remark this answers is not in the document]
shahinrahbariasl: > @bretg I see you added requirement 2.1.19 to indicate when events should be considered enabled:
> 
> > 19. Events should be considered 'enabled' for a given account when any of the following conditions are true:
> >     a. host-level events are enabled and account-level events are enabled
> >     b. host-level events are enabled and the host-level default-account-config has events enabled.
> 
> If I'm not mistaken, from looking at the PBS-Java code and the discussion here, it appears PBS-Java's host-level events enabled flag is actually the host-level default-account-config events enabled flag. In other words, you don't have both a host flag and a default account config flag right? There's no need to have both.
> 
> @shahinrahbariasl PBS-Go actually does have the host-level account defaults config option so we could use that instead of adding this host-level flag. See field `AccountDefaults Account` in the `Configuration` struct in `config/config.go`.

Hi @bsardo, I was just addressing @guscarreon 's comments. You're right I can just do:
```
account_defaults:
  events_enabled: true
```
and that should enable events by default. Didn't know about that flag, I guess there's no need for this PR then...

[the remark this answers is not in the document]
bretg: Very well, updated the doc to remove references to host-level config, clarifying that it's host-level default-account-config

config/accounts.go
```diff
@@ -25,7 +25,7 @@ type Account struct {
 	ID                      string                               `mapstructure:"id" json:"id"`
 	Disabled                bool                                 `mapstructure:"disabled" json:"disabled"`
 	CacheTTL                DefaultTTLs                          `mapstructure:"cache_ttl" json:"cache_ttl"`
-	EventsEnabled           bool                                 `mapstructure:"events_enabled" json:"events_enabled"`
+	EventsEnabled           *bool                                `mapstructure:"events_enabled" json:"events_enabled"`
```
guscarreon, line 28: Can we add an assertion for the default value? And also for the 3 possible values including `nil`?
```
145   func TestDefaults(t *testing.T) {
146       cfg, _ := newDefaultConfig(t)
147  
148       cmpInts(t, "port", cfg.Port, 8000)
149   *-- 31 lines: cmpInts(t, "admin_port", cfg.AdminPort, 6060)-------------------------------------------------------------
180  
181       cmpBools(t, "account_defaults.price_floors.enabled", cfg.AccountDefaults.PriceFloors.Enabled, false)
182       cmpInts(t, "account_defaults.price_floors.enforce_floors_rate", cfg.AccountDefaults.PriceFloors.EnforceFloorsRate, 100)
183       cmpBools(t, "account_defaults.price_floors.adjust_for_bid_adjustment", cfg.AccountDefaults.PriceFloors.AdjustForBidAdjustment, true)
184       cmpBools(t, "account_defaults.price_floors.enforce_deal_floors", cfg.AccountDefaults.PriceFloors.EnforceDealFloors, false)
185       cmpBools(t, "account_defaults.price_floors.use_dynamic_data", cfg.AccountDefaults.PriceFloors.UseDynamicData, false)
    +     cmpBools(t, "account_defaults.events_enabled", cfg.AccountDefaults.EventsEnabled, false)
186  
config/config_test.go
```

shahinrahbariasl:
Hey @guscarreon is there any other feedback I can address for this? Please let me know, thanks!

endpoints/events/event.go
```diff
@@ -107,10 +107,16 @@ func (e *eventEndpoint) Handle(w http.ResponseWriter, r *http.Request, _ httprou
...
-	// account does not support events
-	if !account.EventsEnabled {
+	// account does not support events and events are not enabled by default for accounts
+	if account.EventsEnabled != nil {
+		if !*account.EventsEnabled {
+			w.WriteHeader(http.StatusUnauthorized)
+			w.Write([]byte(fmt.Sprintf("Account '%s' doesn't support events", eventRequest.AccountID)))
+			return
+		}
+	} else if e.Cfg.AccountDefaults.EventsEnabled == nil || !*e.Cfg.AccountDefaults.EventsEnabled {
```
guscarreon, line 117: Hi @shahinrahbariasl sorry for not coming back to you sooner. Your code looks pretty good but, after discussing this feature behind the scenes we notice that `"events_enabled"`, which is deprecated, is still in use in some incoming requests so we'd like to still use its value whenever  `"events.enabled"` is not found. In other words, we should prioritize `account.Events.Enabled` over `account.EventsEnabled`. Can we please code this logic, put it into its own function and write unit tests for it? Please set the resulting boolean value into `account.Events.Enabled` and set `account.EventsEnabled` to `nil`.
| `"events_enabled"` | `"events.enabled"` | output |
| ------ | ------ | ------ |
| `false` | `false` | `false` |
| `false` | `true` | `true` |
| `false` | not set | `false` |
| `true` | `false` | `false` |
| `true` | `true` | `true` |
| `true` | not set | `true` |
| not set | `false` | `false` |
| not set | `true` | `true` |
| not set | not set | `true` |

We should use the same logic shown above for `e.Cfg.AccountDefaults.EventsEnabled` and `e.Cfg.AccountDefaults.Events.Enabled` before the `MarshalAccountDefaults()` function call in line 691 of `config/config.go` but, in this instance, set the resulting value to the deprecated field `e.Cfg.AccountDefaults.EventsEnabled` while setting the new field `e.Cfg.AccountDefaults.Events.Enabled` to `nil`.
```
673 // New uses viper to get our server configurations.
674 func New(v *viper.Viper, bidderInfos BidderInfos, normalizeBidderName func(string) (openrtb_ext.BidderName, bool)) (*Configuration, error) {
675     var c Configuration
676     if err := v.Unmarshal(&c); err != nil {
677         return nil, fmt.Errorf("viper failed to unmarshal app config: %v", err)
678     }
679
680     if err := c.RequestValidation.Parse(); err != nil {
681         return nil, err
682     }
683
684     if err := isValidCookieSize(c.HostCookie.MaxCookieSizeBytes); err != nil {
685         glog.Fatal(fmt.Printf("Max cookie size %d cannot be less than %d \n", c.HostCookie.MaxCookieSizeBytes, MIN_COOKIE_SIZE_BYTES))
686         return nil, err
687     }
688
689     // Update account defaults and generate base json for patch
690     c.AccountDefaults.CacheTTL = c.CacheURL.DefaultTTLs // comment this out to set explicitly in config
691     if err := c.MarshalAccountDefaults(); err != nil {
692         return nil, err
693     }
694
config/config.go
```
In summary we want to maintain the following order of precedence:
1. account config new field `account.Events.Enabled`
2. account config deprecated field `account.EventsEnabled`
3. host config new field `e.Cfg.AccountDefaults.Events.Enabled`
4. host config deprecated field `e.Cfg.AccountDefaults.EventsEnabled`
Does this make sense?

guscarreon (commented):
The `Events` field that represents the new `"accounts.events.enabled"` flag, also needs to be declared as a pointer.
```
63   type Events struct {
64 -     Enabled    bool        `mapstructure:"enabled" json:"enabled"`
   +     Enabled    *bool       `mapstructure:"enabled" json:"enabled"`
65       DefaultURL string      `mapstructure:"default_url" json:"default_url"`
66       VASTEvents []VASTEvent `mapstructure:"vast_events" json:"vast_events,omitempty"`
67   }
config/events.go
```

If not, the legacy field `"events_enabled"` will always be ignored. Notice how the truth table changes if `"events.enabled"` is not a pointer:

| r | `"events_enabled"` | `"events.enabled"` | output  |
| - | ------- | ------- | ------- |
| 1 | `false` | `false` | `false` |
| 2 | `false` | `true` | `true`  |
| 3 | `true` | `false` | `false` |
| 4 | `true` | `true` | `true`  |
| 5 | not set | `false`| `false` |
| 6 | not set | `true` | `true`  |

The output column above is identical to the `"events.enabled"` value everytime. The correct truth table is shown below (corrected as prer @bsardo's comment). Under this logic the only scenarios where the legacy field `"events_enabled"` is taken into account are rows 3 and 6 when `"events.enabled"` is not set. In other words, if the new `Events` field is not set, we should use the legacy value `"events_enabled"` instead as we expect some incoming requests to still use the deprecated `"events_enabled"` flag and not the new one. Also notice row 9: if neither flags are set (both pointers `nil`), we should default to `false`.

| r | `"events_enabled"` | `"events.enabled"` | output |
| - | ------ | ------ | ------ |
| 1 | `false` | `false` | `false` |
| 2 | `false` | `true` | `true` |
| 3 | `false` | not set | `false` |
| 4 | `true` | `false` | `false` |
| 5 | `true` | `true` | `true` |
| 6 | `true` | not set | `true` |
| 7 | not set | `false` | `false` |
| 8 | not set | `true` | `true` |
| 9 | not set | not set | `false` |

I think we should declare `Events` as a `*bool` and modify `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` acordingly.

Also: can  `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` be one function? They look, and after the changes will still look, very similar.

shahinrahbariasl:
> The `Events` field that represents the new `"accounts.events.enabled"` flag, also needs to be declared as a pointer.
> 
> ```
> 63   type Events struct {
> 64 -     Enabled    bool        `mapstructure:"enabled" json:"enabled"`
>    +     Enabled    *bool       `mapstructure:"enabled" json:"enabled"`
> 65       DefaultURL string      `mapstructure:"default_url" json:"default_url"`
> 66       VASTEvents []VASTEvent `mapstructure:"vast_events" json:"vast_events,omitempty"`
> 67   }
> config/events.go
> ```
> 
> If not, the legacy field `"events_enabled"` will always be ignored. Notice how the truth table changes if `"events.enabled"` is not a pointer:
> 
> r	`"events_enabled"`	`"events.enabled"`	output
> 1	`false`	`false`	`false`
> 2	`false`	`true`	`true`
> 3	`true`	`false`	`false`
> 4	`true`	`true`	`true`
> 5	not set	`false`	`false`
> 6	not set	`true`	`true`
> The output column above is identical to the `"events.enabled"` value everytime. The correct truth table is shown below (corrected as prer @bsardo's comment). Under this logic the only scenarios where the legacy field `"events_enabled"` is taken into account are rows 3 and 6 when `"events.enabled"` is not set. In other words, if the new `Events` field is not set, we should use the legacy value `"events_enabled"` instead as we expect some incoming requests to still use the deprecated `"events_enabled"` flag and not the new one. Also notice row 9: if neither flags are set (both pointers `nil`), we should default to `false`.
> 
> r	`"events_enabled"`	`"events.enabled"`	output
> 1	`false`	`false`	`false`
> 2	`false`	`true`	`true`
> 3	`false`	not set	`false`
> 4	`true`	`false`	`false`
> 5	`true`	`true`	`true`
> 6	`true`	not set	`true`
> 7	not set	`false`	`false`
> 8	not set	`true`	`true`
> 9	not set	not set	`false`
> I think we should declare `Events` as a `*bool` and modify `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` acordingly.
> 
> Also: can `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` be one function? They look, and after the changes will still look, very similar.

Hey @guscarreon, that's a good call out. Just to be clear, you mean the `Enabled` field in Events object to be a pointer right?

guscarreon:
> > The `Events` field that represents the new `"accounts.events.enabled"` flag, also needs to be declared as a pointer.
> > ```
> > 63   type Events struct {
> > 64 -     Enabled    bool        `mapstructure:"enabled" json:"enabled"`
> >    +     Enabled    *bool       `mapstructure:"enabled" json:"enabled"`
> > 65       DefaultURL string      `mapstructure:"default_url" json:"default_url"`
> > 66       VASTEvents []VASTEvent `mapstructure:"vast_events" json:"vast_events,omitempty"`
> > 67   }
> > config/events.go
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If not, the legacy field `"events_enabled"` will always be ignored. Notice how the truth table changes if `"events.enabled"` is not a pointer:
> > r	`"events_enabled"`	`"events.enabled"`	output
> > 1	`false`	`false`	`false`
> > 2	`false`	`true`	`true`
> > 3	`true`	`false`	`false`
> > 4	`true`	`true`	`true`
> > 5	not set	`false`	`false`
> > 6	not set	`true`	`true`
> > The output column above is identical to the `"events.enabled"` value everytime. The correct truth table is shown below (corrected as prer @bsardo's comment). Under this logic the only scenarios where the legacy field `"events_enabled"` is taken into account are rows 3 and 6 when `"events.enabled"` is not set. In other words, if the new `Events` field is not set, we should use the legacy value `"events_enabled"` instead as we expect some incoming requests to still use the deprecated `"events_enabled"` flag and not the new one. Also notice row 9: if neither flags are set (both pointers `nil`), we should default to `false`.
> > r	`"events_enabled"`	`"events.enabled"`	output
> > 1	`false`	`false`	`false`
> > 2	`false`	`true`	`true`
> > 3	`false`	not set	`false`
> > 4	`true`	`false`	`false`
> > 5	`true`	`true`	`true`
> > 6	`true`	not set	`true`
> > 7	not set	`false`	`false`
> > 8	not set	`true`	`true`
> > 9	not set	not set	`false`
> > I think we should declare `Events` as a `*bool` and modify `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` acordingly.
> > Also: can `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` be one function? They look, and after the changes will still look, very similar.
> 
> Hey @guscarreon, that's a good call out. Just to be clear, you mean the `Enabled` field in Events object to be a pointer similar to `event_enabled` right?

  Yes @shahinrahbariasl

shahinrahbariasl:
> > > The `Events` field that represents the new `"accounts.events.enabled"` flag, also needs to be declared as a pointer.
> > > ```
> > > 63   type Events struct {
> > > 64 -     Enabled    bool        `mapstructure:"enabled" json:"enabled"`
> > >    +     Enabled    *bool       `mapstructure:"enabled" json:"enabled"`
> > > 65       DefaultURL string      `mapstructure:"default_url" json:"default_url"`
> > > 66       VASTEvents []VASTEvent `mapstructure:"vast_events" json:"vast_events,omitempty"`
> > > 67   }
> > > config/events.go
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > If not, the legacy field `"events_enabled"` will always be ignored. Notice how the truth table changes if `"events.enabled"` is not a pointer:
> > > r	`"events_enabled"`	`"events.enabled"`	output
> > > 1	`false`	`false`	`false`
> > > 2	`false`	`true`	`true`
> > > 3	`true`	`false`	`false`
> > > 4	`true`	`true`	`true`
> > > 5	not set	`false`	`false`
> > > 6	not set	`true`	`true`
> > > The output column above is identical to the `"events.enabled"` value everytime. The correct truth table is shown below (corrected as prer @bsardo's comment). Under this logic the only scenarios where the legacy field `"events_enabled"` is taken into account are rows 3 and 6 when `"events.enabled"` is not set. In other words, if the new `Events` field is not set, we should use the legacy value `"events_enabled"` instead as we expect some incoming requests to still use the deprecated `"events_enabled"` flag and not the new one. Also notice row 9: if neither flags are set (both pointers `nil`), we should default to `false`.
> > > r	`"events_enabled"`	`"events.enabled"`	output
> > > 1	`false`	`false`	`false`
> > > 2	`false`	`true`	`true`
> > > 3	`false`	not set	`false`
> > > 4	`true`	`false`	`false`
> > > 5	`true`	`true`	`true`
> > > 6	`true`	not set	`true`
> > > 7	not set	`false`	`false`
> > > 8	not set	`true`	`true`
> > > 9	not set	not set	`false`
> > > I think we should declare `Events` as a `*bool` and modify `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` acordingly.
> > > Also: can `IsAccountEventEnabled` and `IsAccountDefaultsEventEnabled` be one function? They look, and after the changes will still look, very similar.
> > 
> > 
> > Hey @guscarreon, that's a good call out. Just to be clear, you mean the `Enabled` field in Events object to be a pointer similar to `event_enabled` right?
> 
> Yes @shahinrahbariasl

I think we can also remove this warning since we're now technically taking `cfg.AccountDefaults.Events.Enabled` into consideration as well
https://github.com/prebid/prebid-server/blob/master/config/config.go#L141

guscarreon:
Clicked the wrong button and closed by accident. Sorry.

guscarreon:
> I think we can also remove this warning since we're now technically taking `cfg.AccountDefaults.Events.Enabled` into consideration as well https://github.com/prebid/prebid-server/blob/master/config/config.go#L141

Good point we should remove that warning.

shahinrahbariasl:
> > I think we can also remove this warning since we're now technically taking `cfg.AccountDefaults.Events.Enabled` into consideration as well https://github.com/prebid/prebid-server/blob/master/config/config.go#L141
> 
> Good point we should remove that warning.

Agreed and addressed.

config/config_test.go
```diff
@@ -3266,77 +3268,3 @@ func TestIsAccountEventEnabled(t *testing.T) {
 func boolPointer(b bool) *bool {
 	return &b
```
guscarreon, line 3269: Nitpick: To avoid repeated code in the codebase, please use the `ToPtr()` function found in `util/ptrutil/ptrutil.go`

config/config_test.go
```diff
@@ -3190,64 +3190,66 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
 }
 
 func TestIsAccountEventEnabled(t *testing.T) {
+	boolFalse := false
+	boolTrue := true
```
guscarreon, line 3194: Nitpick: We don't need these two if we simply use the `ToPtr()` function found in `util/ptrutil/ptrutil.go`

config/config_test.go
```diff
@@ -3190,64 +3190,66 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
 		deprecatedEventsEnabled *bool
-		eventsEnabled           bool
+		eventsEnabled           *bool
 		expectedIsEventEnabled  bool
 	}{
 		{
 			name:                    "events disabled, events.enabled false",
 			deprecatedEventsEnabled: boolPointer(false),
-			eventsEnabled:           false,
+			eventsEnabled:           &boolFalse,
```
guscarreon, line 3204: Nitpick: Can we simplify by using the `ToPtr()` function from `util/ptrutil/ptrutil.go`?
```
3203 -           deprecatedEventsEnabled: boolPointer(false),
3204 -           eventsEnabled:           &boolFalse,
     +           deprecatedEventsEnabled: ptrutil.ToPtr(false),
     +           eventsEnabled:           ptrutil.ToPtr(false),
```

config/config_test.go
```diff
@@ -3190,64 +3190,66 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
 			expectedIsEventEnabled:  false,
 		},
 		{
 			name:                    "events disabled, events.enabled true",
 			deprecatedEventsEnabled: boolPointer(false),
-			eventsEnabled:           true,
+			eventsEnabled:           &boolTrue,
```
guscarreon, line 3210: Nitpick: Can we simplify by using the `ToPtr()` function from `util/ptrutil/ptrutil.go`?
```
3209 -           deprecatedEventsEnabled: boolPointer(false),
3210 -           eventsEnabled:           &boolTrue,
     +           deprecatedEventsEnabled: ptrutil.ToPtr(false),
     +           eventsEnabled:           ptrutil.ToPtr(true),
```

guscarreon (commented):
Looking pretty good. I believe we're close

config/config.go
```diff
@@ -1463,3 +1467,26 @@ func isValidCookieSize(maxCookieSize int) error {
 	}
 	return nil
 }
+
+// IsAccountEventEnabled checks whether events are enabled for an account. In some incoming requests,
+// the `account.events_enabled` field, which is deprecated, is still used. We want to use its value whenever
+// the `account.events.enabled` field is not found.
+func IsAccountEventEnabled(deprecated_events_enabled *bool, events_enabled *bool) bool {
```
guscarreon, line 1474: As mentioned in the `config/config_test.go` comment, we should expect `false` in the 4th unit test of `TestIsAccountEventEnabled`. For this outcome to be met, this function needs refactoring. 

In a nutshell, want `IsAccountEventEnabled` to return the `events_enabled` value if set. If not, we should check if our second option, `deprecated_events_enabled` is set and return that. If none of the flags was set, we should return `false`.

guscarreon (approved):
LGTM. @shahinrahbariasl thank you for addressing my feedback

shahinrahbariasl:
> LGTM. @shahinrahbariasl thank you for addressing my feedback

No worries, thanks for reviewing. I wasn't fully familiar with the events code but now I am :D

bsardo (commented):
Hi @shahinrahbariasl, 
Thanks for all of the work you’ve been doing here. We’d like to get this merged in very soon. I thought we were close to merging and then another contributor opened up PR #2698 to handle this exact same issue which got me thinking about a few things. If you don’t mind, I’d like us to draw some inspiration from that PR and make a few more updates here. I’ve called out the specific parts that I think we should pull in.

The motivation behind these new requested changes is as follows:
1. We should warn the host when their host config uses the deprecated events_enabled flag
2. We should warn the host that we are using the new flag and ignoring the old flag when their host config specifies both the deprecated and new flags
3. We should warn the host when their host config specifies VASTEvents as this is not supported yet.
4. It would be preferable if we map the fetched account deprecated flag value to the new flag value and only use the new flag value in the application code present in endpoints/events/events.go and exchange/events.go. This way when the time comes to remove support for the deprecated field entirely, the downstream code does not need to be touched.

For item 4, could you please start by lifting the `account/account.go#deprecateEventsEnabledField` function and its corresponding unit test from PR #2698? Note: please refrain from pulling in the VASTEvents warning there as I think that would flood the logs.

Thanks again for all of your work here!

account/account_test.go
```diff
@@ -579,3 +580,40 @@ func TestAccountUpgradeStatusGetAccount(t *testing.T) {
...
+		{
+			name: "events.enabled is non-nil",
+			account: &config.Account{
+				Events: config.Events{
+					Enabled: ptrutil.ToPtr(true),
+				},
+				EventsEnabled: ptrutil.ToPtr(false),
+			},
+			want: ptrutil.ToPtr(true),
+		},
```
guscarreon, line 610: As is, `TestDeprecateEventsEnabledField` covers the following scenarios:

| account | EventsEnabled | Events.Enabled |  Output  |
| ------- | ------------- | -------------- | -------- | 
| not-nil |     true      |     nil        |   true   |
| not-nil |     false     |     true       |   true   |

Can we expand its coverage to the full set of possibilities shown below?

| account | EventsEnabled           | Events.Enabled in       | Events.Enabled out      |
| ------- | ----------------------- | ----------------------- | ----------------------- | 
|  nil    |     N/A                 |     N/A                 |   nil                   |
| not-nil |     nil                 |     nil                 |   nil                   |
| not-nil |     nil                 | not-nil (true or false) | Events.Enabled in |
| not-nil | not-nil (true or false) |     nil                 | EventsEnabled value     |
| not-nil | not-nil (true or false) | not-nil (true or false) | vents.Enabled in  |

bsardo (commented):
This is looking good; I have just a few more quick comments.

config/config_test.go
```diff
@@ -3191,3 +3200,51 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
+			expectedNewFieldValue: nil,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			updatedOldFieldValue, updatedNewFieldValue := migrateConfigEventsEnabled(tc.oldFieldValue, tc.newFieldValue)
+
+			assert.Equal(t, tc.expectedOldFieldValue, updatedOldFieldValue)
+			assert.Nil(t, updatedNewFieldValue)
```
guscarreon, line 3247: `expectedNewFieldValue` seems to not be asserted in this section. I believe @bsardo advocated for an `assert.Nil(t, expectedNewFieldValue)` here.

account/account.go
```diff
@@ -252,3 +256,16 @@ func useGDPRChannelEnabled(account *config.Account) bool {
...
 }
+
+// deprecateEventsEnabledField is responsible for ensuring backwards compatibility of "events_enabled" field.
+// This function favors "events.enabled" field over deprecated "events_enabled" field, if values for both are set.
+// If only deprecated "events_enabled" field is set then it sets the same value to "events.enabled" field.
+func deprecateEventsEnabledField(account *config.Account) {
+	if account != nil && account.Events.Enabled == nil {
+		account.Events.Enabled = account.EventsEnabled
+	}
+	if account != nil {
```
guscarreon, line 267: In order to avoid the second check we can nest the first `if` statement:
```
263   func deprecateEventsEnabledField(account *config.Account) {
264 -     if account != nil && account.Events.Enabled == nil {
265 -         account.Events.Enabled = account.EventsEnabled
266 -     }
267       if account != nil {
    +         if account.Events.Enabled == nil {
    +             account.Events.Enabled = account.EventsEnabled
    +         }
268           // assign the old value to the new value so old and new are always the same even though the new value is what is used in the application code.
269           account.EventsEnabled = account.Events.Enabled
270       }
271   }
account/account.go
```

config/config.go
```diff
@@ -1371,6 +1378,30 @@ func migrateConfigDatabaseConnection(v *viper.Viper) {
 	}
 }
 
+// migrateConfigEventsEnabled is responsible for ensuring backward compatibility of events_enabled field.
+// This function copies the value of newField "events.enabled" and set it to the oldField "events_enabled".
+// This is necessary to achieve the desired order of precedence favoring the account values over the host values given the account fetcher JSON merge mechanics.
```
guscarreon, line 1383: Nitpick: can we split this line in two?

bsardo (dismissed):
LGTM

guscarreon:
Hi @shahinrahbariasl it seems that there is a conflic that must be resolved before we can merge your PR.

bsardo (dismissed):

config/config.go
```diff
@@ -1371,6 +1378,31 @@ func migrateConfigDatabaseConnection(v *viper.Viper) {
...
+		glog.Warningf("%s is deprecated and should be changed to %s", oldField, newField)
+	}
+	if newFieldValue != nil {
+		if oldFieldValue != nil {
+			glog.Warningf("using %s and ignoring deprecated %s", newField, oldField)
+		}
+	}
+
+	updatedOldFieldValue = oldFieldValue
+	if newFieldValue != nil {
```
guscarreon, line 1399: Niptpick: a small optimization would be to avoid the second `nil` check:
```
1385   func migrateConfigEventsEnabled(oldFieldValue *bool, newFieldValue *bool) (updatedOldFieldValue, updatedNewFieldValue *bool) {
1386       newField := "account_defaults.events.enabled"
1387       oldField := "account_defaults.events_enabled"
1388  
     +     updatedOldFieldValue = oldFieldValue
1389       if oldFieldValue != nil {
1390           glog.Warningf("%s is deprecated and should be changed to %s", oldField, newField)
1391       }
1392       if newFieldValue != nil {
1393           if oldFieldValue != nil {
1394               glog.Warningf("using %s and ignoring deprecated %s", newField, oldField)
1395           }
     +         updatedOldFieldValue = ptrutil.ToPtr(*newFieldValue)
1396       }
1397  
1398 -     updatedOldFieldValue = oldFieldValue
1399 -     if newFieldValue != nil {
1400 -         updatedOldFieldValue = ptrutil.ToPtr(*newFieldValue)
1401 -     }
1402  
1403       return updatedOldFieldValue, nil
1404   }
config/config.go
```

guscarreon (commented):
Looks pretty good. I only had a couple of minor comments

bsardo (approved):

guscarreon (approved):
LGTM. @shahinrahbariasl thank you for addressing my feedback

PR #2534, opened 2023-01-19
Openrtb2blocking module part 3: RawBidderResponse Hook implementation

legendko:
Module PRD: https://docs.google.com/document/d/1TpB_I-Oft0LpZuKvVsaLAozOe1u52tpPGkB_jZfnMXo/edit?usp=sharing


VeronikaSolovei9:
I tried to create my own module locally and was able to this! It eventually started to work!

VeronikaSolovei9:
I tried to configure my own module locally and I was able to to this! It eventually started to work!
I had to comment out metrics code, because it threw null pointer exception. 
**It is possible I misconfigured something, but please can you double check?** 
It looks like "moduleCalls" and labels for this module have different format: "vendor_module" and "vendor.module". Because of this CounterVector for my new module cannot be found and it leads to NPE. 
Here are screenshots  with variables in my debugger:
Screen Shot 2023-01-19 at 11 15 13 PM
And this how it is initialized:
Screen Shot 2023-01-19 at 11 03 18 PM

legendko:
> I tried to configure my own module locally and I was able to to this! It eventually started to work! I had to comment out metrics code, because it threw null pointer exception.

Yes, you are correct, there was a bug. We registered the module metrics correctly but the actual module labels that had to be sent were incorrect. Now it is fixed. Thank you.

VeronikaSolovei9:
Thank you! Just to confirm: if vendor.module name has underscores itself, like `my_vendor.my_module` - will it still work?

legendko:
> Thank you! Just to confirm: if vendor.module name has underscores itself, like my_vendor.my_module or my-vendor.my-module - will it still work?

Sure, the both variants would be replaced to `my_vendor_my_module` in the metric labels and would be processed correctly.

SyntaxNode:
> Thank you! Just to confirm: if vendor.module name has underscores itself, like `my_vendor.my_module` or `my-vendor.my-module` - will it still work?

We should ask module contributors to just use alphanumeric characters, like we do for adapters.

VeronikaSolovei9:
> We should ask module contributors to just use alphanumeric characters, like we do for adapters.

Then it should be reflected in documentation

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
```diff
@@ -0,0 +1,381 @@
...
+	bidMediaTypes mediaTypes,
+	bidder, dealID string,
+	failedChecksData map[string]interface{},
+) (map[string]interface{}, error) {
+	badv := cfg.Attributes.Badv
+
+	enforceBlocks, message, err := firstOrDefaultOverride(bidder, bidMediaTypes, getIsActive, badv.ActionOverrides.EnforceBlocks, badv.EnforceBlocks)
+	result.Warnings = mergeStrings(result.Warnings, message)
+	if err != nil {
+		return failedChecksData, fmt.Errorf("failed to get override for badv.enforce_blocks: %s", err)
```
bsardo, line 106: Can we add test coverage for this error case by adding a test case to `TestHandleRawBidderResponseHook` that takes advantage of the validation rule requiring at least `bidder` or `mediaTypes` to be specified as an override condition:
```
{
	description: "Expect error if there is an issue processing enforce blocks overrides. Analytics should have error status tag",
	payload: hookstage.RawBidderResponsePayload{Bidder: bidder, Bids: []*adapters.TypedBid{
		{
			Bid: &openrtb2.Bid{ID: "1", ADomain: []string{"forbidden_domain"}},
		},
	}},
	config: json.RawMessage(`{"attributes": {"badv": {"enforce_blocks": true, "action_overrides": {"enforce_blocks": [{"conditions": {}}]}}}}`),
	expectedBids: []*adapters.TypedBid{
		{
			Bid: &openrtb2.Bid{ID: "1", ADomain: []string{"forbidden_domain"}},
		},
	},
	expectedHookResult: hookstage.HookResult[hookstage.RawBidderResponsePayload]{
		AnalyticsTags: hookanalytics.Analytics{
			Activities: []hookanalytics.Activity{
				{
					Name:   enforceBlockingTag,
					Status: hookanalytics.ActivityStatusError,
				},
			},
		},
	},
	expectedError: hookexecution.NewFailure("failed to process badv block checking: failed to get override for badv.enforce_blocks: bidders and media_types absent from conditions, at least one of the fields must be present"),
},
```

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	if !enforceBlocks {
+		return failedChecksData, nil
+	}
+
+	blockUnknown, message, err := firstOrDefaultOverride(bidder, bidMediaTypes, getIsActive, badv.ActionOverrides.BlockUnknownAdomain, badv.BlockUnknownAdomain)
+	result.Warnings = mergeStrings(result.Warnings, message)
+	if err != nil {
+		return failedChecksData, fmt.Errorf("failed to get override for badv.block_unknown_adomain: %s", err)
+	}
```
bsardo, line 116: Similar to my comment above, can we add test coverage for this error case by adding a test case to `TestHandleRawBidderResponseHook` that takes advantage of the validation rule requiring at least `bidder` or `mediaTypes` to be specified as an override condition:
```
{
	description: "Expect error if there is an issue processing block unknown domains overrides. Analytics should have error status tag",
	payload: hookstage.RawBidderResponsePayload{Bidder: bidder, Bids: []*adapters.TypedBid{
		{
			Bid: &openrtb2.Bid{ID: "1", ADomain: []string{"forbidden_domain"}},
		},
	}},
	config: json.RawMessage(`{"attributes": {"badv": {"enforce_blocks": true, "action_overrides": {"block_unknown_adomain": [{"conditions": {}}]}}}}`),
	expectedBids: []*adapters.TypedBid{
		{
			Bid: &openrtb2.Bid{ID: "1", ADomain: []string{"forbidden_domain"}},
		},
	},
	expectedHookResult: hookstage.HookResult[hookstage.RawBidderResponsePayload]{
		AnalyticsTags: hookanalytics.Analytics{
			Activities: []hookanalytics.Activity{
				{
					Name:   enforceBlockingTag,
					Status: hookanalytics.ActivityStatusError,
				},
			},
		},
	},
	expectedError: hookexecution.NewFailure("failed to process badv block checking: failed to get override for badv.block_unknown_adomain: bidders and media_types absent from conditions, at least one of the fields must be present"),
},
```

modules/prebid/ortb2blocking/module_test.go
```diff
@@ -594,14 +599,698 @@ func TestHandleBidderRequestHook(t *testing.T) {
...
+					Activities: []hookanalytics.Activity{
+						{
+							Name:   enforceBlockingTag,
+							Status: hookanalytics.ActivityStatusError,
+						},
+					},
+				},
+			},
+			expectedError: hookexecution.NewFailure("failed to process badv block checking: failed to get override for badv.allowed_adomain_for_deals: conditions field in account configuration must contain deal_ids"),
+		},
```
bsardo, line 971: Can we increase our code coverage of the error cases associated with `bcat`, `bapp` and `battr` involving bad override conditions for `enforce_blocks`, deal IDs and `block_unknown` similar to this deal ID error test case and the two other test cases I suggested for `badv`?

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
```diff
@@ -0,0 +1,381 @@
...
+	badv := cfg.Attributes.Badv
+
+	enforceBlocks, message, err := firstOrDefaultOverride(bidder, bidMediaTypes, getIsActive, badv.ActionOverrides.EnforceBlocks, badv.EnforceBlocks)
+	result.Warnings = mergeStrings(result.Warnings, message)
+	if err != nil {
+		return failedChecksData, fmt.Errorf("failed to get override for badv.enforce_blocks: %s", err)
+	}
+	if !enforceBlocks {
+		return failedChecksData, nil
+	}
```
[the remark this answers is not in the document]
legendko: I guess the documentation on the parameter `enforceBlocks` is incomplete. In PBS-Java the parameter defines if we should block anything in general. The implementation can be seen here: https://github.com/prebid/prebid-server-java/blob/master/extra/modules/ortb2-blocking/src/main/java/org/prebid/server/hooks/modules/ortb2/blocking/core/BidsBlocker.java#L173 (for a list of attributes), and here https://github.com/prebid/prebid-server-java/blob/master/extra/modules/ortb2-blocking/src/main/java/org/prebid/server/hooks/modules/ortb2/blocking/core/BidsBlocker.java#L202 (for single attribute)
These checkAttribute functions (used for every attribute) check at the very begining if `enforceBlocks` is disabled we consider that this attribute is allowed without any further checking.
So basically yes, as you stated above we skip the checking for adomains and deal exceptions if `enforceBlocks` is `false`

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
```diff
@@ -0,0 +1,381 @@
...
+				overrides = append(overrides, actionOverride...)
+			}
+		}
+	}
+
+	if len(overrides) == 0 {
+		return defaultOverride, nil
+	}
+
+	overrides = append(overrides, defaultOverride...)
```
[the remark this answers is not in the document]
legendko: Yes, you are correct

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
```diff
@@ -0,0 +1,381 @@
...
+	return mediaTypes{string(bid.BidType): struct{}{}}
+}
+
+func shouldBeBlockedDueToBadv(
+	result *hookstage.HookResult[hookstage.RawBidderResponsePayload],
+	cfg config,
+	blockAttr blockingAttributes,
+	bid *openrtb2.Bid,
+	bidMediaTypes mediaTypes,
+	bidder, dealID string,
```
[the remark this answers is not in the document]
legendko: Sure

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
legendko: Yes

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
legendko: Yes

modules/prebid/ortb2blocking/module_test.go
legendko: Sure, I can add those as well. I did not add these tests for the other attributes originally, because basically they would be a full copy of `badv` checking as this `bad override conditions` do not depend on the attributes themselves but the `action_override` structure which is the same for each attribute.

modules/prebid/ortb2blocking/hook_raw_bidder_response.go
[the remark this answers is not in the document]
bretg: I updated the doc to make it clear that app and attr are also in scope for the enforce-blocks flag. Changes highlighted in blue. Believe the examples covered these, but you're right that the text was missing them.

bsardo (approved):

VeronikaSolovei9 (approved):

guscarreon (commented):
.

PR #2536, opened 2023-01-23
Rubicon: Initial ORTB 2.6 support

CTMBNara:


adapters/rubicon/rubicon.go
```diff
@@ -231,6 +233,19 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter, server co
...
+		return nil, []error{err}
+	}
+
+	requestDeepCopy := requestCopy.(*openrtb2.BidRequest)
+	err = openrtb_ext.ConvertUpTo26(&openrtb_ext.RequestWrapper{BidRequest: requestDeepCopy})
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	request = requestDeepCopy
```
CTMBNara, line 247: @SyntaxNode Hi. Is there another best way to do conversion?

bretg:
@CTMBNara - please also downgrade user.consent to user.ext.consent. Thanks!

openrtb_ext/imp_rubicon.go
```diff
@@ -15,6 +15,7 @@ type ExtImpRubicon struct {
 	Visitor          json.RawMessage    `json:"visitor,omitempty"`
 	Video            rubiconVideoParams `json:"video"`
 	Debug            impExtRubiconDebug `json:"debug,omitempty"`
+	PChain           string             `json:"pchain,omitempty"`
```
onkarvhanumante, line 18: should create docs PR to include description for this `pchain` bidder param ([rubicon.md](https://github.com/prebid/prebid.github.io/blob/8f9ef40d8e53204004ff7e327460a0d6a67b2d22/dev-docs/bidders/rubicon.md#L32-L45))

bretg: I'm actually pushing to get rid of this deprecated param entirely, but went ahead and added it to the docs so this PR can be merged. We'll remove it as a separate effort if I get the ok . https://github.com/prebid/prebid.github.io/pull/4397

adapters/rubicon/rubicon.go
```diff
@@ -231,6 +234,19 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter, server co
 }
 
 func (a *RubiconAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	requestCopy, err := copystructure.Copy(request)
+	if err != nil {
+		return nil, []error{err}
+	}
```
onkarvhanumante, line 240: `copy` method used is not preferred here. How about simply assigning request to new var and using it

```
requestCopy := request
```

onkarvhanumante:
@CTMBNara  any updates on this PR

onkarvhanumante:
@CTMBNara,  Rubicon adapter tests are failing since https://github.com/prebid/prebid-server/pull/2536/commits/3c1751e69e971466c0fd8413445fee1d79c6e0c4 changes. Could you fix them.

CTMBNara:
> @CTMBNara, Rubicon adapter tests are failing since [3c1751e](https://github.com/prebid/prebid-server/commit/3c1751e69e971466c0fd8413445fee1d79c6e0c4) changes. Could you fix them.

This problem is caused by `openrtb_ext.ConvertUpTo26`. It's the method that modifies `BidRequest`. Since we are now simply passing it the same `BidRequest` that we are passing in `MakeRequests`, a `Data race ...` error occurs.

To fix this, we need to go back to the previous solution or do a `convert` before `MakeRequests`.

adapters/rubicon/rubicon.go
CTMBNara: @onkarvhanumante, can you answer please? ^^

adapters/rubicon/rubicon.go
```diff
@@ -231,6 +233,15 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter, server co
...
 func (a *RubiconAdapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	requestCopy := request
+
+	err := openrtb_ext.ConvertUpTo26(&openrtb_ext.RequestWrapper{BidRequest: requestCopy})
+	if err != nil {
+		return nil, []error{err}
+	}
+
+	request = requestCopy
+
```
VeronikaSolovei9, line 244: `MakeRequest` should not modify original request. This is a part of JSON test verification. It's easy to accidentally modify a property (or sub-object) in a shallow copy of an object such as any pointer in a bid request: `Site *Site`,  `App *App`, `Device *Device`, `User *User`, etc.

This code is what is causing the race condition error.
 
`ConvertUpTo26` sets `User.Eids` to original `request` and `requestCopy` because `user` is the same object in memory in both of them.

To fix this issue `user` should be copied from the original request and set back to `requestCopy` so `requestCopy` has its own `user`.  

Insert this code after line 236:

```go
if request.User != nil {
		userCopy := *request.User
		requestCopy.User = &userCopy
	}
```
This fixes race condition error

VeronikaSolovei9:
Thank you for the changes. Please keep in mind you will need to make a copy of other sub-objects if you will make changes in them. 
Code looks good to me.
@onkarvhanumante please re-review.

onkarvhanumante (approved):

VeronikaSolovei9 (approved):

PR #2538, opened 2023-01-23
Metrics for deprecated account config fields

AlexBVolcy:
This PR https://github.com/prebid/prebid-server/pull/2330 that was put out a few months ago introduced a change to account related GDPR fields where the value assigned to `gdpr.tcf2.purpose{1-10}.enforce_purpose` is now a boolean instead of a string.

Even though the use of a boolean was now the standard, this config change allowed for backwards compatibility where a string could still be used, and it would just be converted to a boolean.

In this metrics PR, we wanted to introduce a metric that tracks when this backwards compatibility is used, and subsequently let the host know that they're using a deprecated variable type for assigning a value to `gdpr.tcf2.purpose{1-10}.enforce_purpose`

A metric was created for each purpose `accountDeprecationWarningsPurposeXMeter` and is recorded through the function `RecordAccountDeprecationWarnings`

Metrics were also created to track the usage of two other account-related fields that were deprecated: `ChannelEnabled` for both GDPR and CCPA `channelEnabledGDPRMeter ` `channelEnabledCCPAMeter`, which get recorded through the function `RecordAccountGDPRChannelEnabledWarning` and `RecordAccountCCPAChannelEnabledWarning`


metrics/prometheus/prometheus.go
```diff
@@ -483,6 +497,48 @@ func NewMetrics(cfg config.PrometheusMetrics, disabledMetrics config.DisabledMet
...
+
+	metrics.channelEnabledCCPA = newCounterWithoutLabels(cfg, reg,
+		"ccpa_channel_enabled",
+		"Count of requests referencing an account whose config specifies a depreceated ccpa_channel_enabled field")
+	metrics.channelEnabledGDPR = newCounterWithoutLabels(cfg, reg,
+		"gdpr_channel_enabled",
+		"Count of requests referencing an account whose config specifies a depreceated gdpr_channel_enabled field")
+
+	metrics.accountDepreciationSummary = newCounterWithoutLabels(cfg, reg,
+		"account_depreciation_summary",
```
bsardo, line 539: Can we change this to `account_config_summary` for now? I could maybe see us recording errors as well as warnings at some point.
I'm finding the name for this one a bit challenging so it would be good to see what another reviewer thinks.

bsardo (commented):
This is on the right path. I have a few more comments. I'm also thinking that maybe the summary metric should record the number of accounts that require an upgrade rather than the number of fields that are deprecated but let me look into that a bit more first before we make any change there.

account/account_test.go
```diff
@@ -351,3 +355,145 @@ func TestConvertGDPREnforcePurposeFields(t *testing.T) {
...
+			description:         "Account isn't channel enabled, GDPR metric shouldn't be called",
+			givenAccountID:      "valid_acct",
+			givenMetric:         "RecordAccountGDPRChannelEnabledWarning",
+			required:            false,
+			disabled:            false,
+			expectedMetricCount: 0,
+		},
+	}
+
+	for _, test := range testCases {
```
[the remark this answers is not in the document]
hhhjort: You probably want to add a name field with a much smaller name with no spaces instead of `test.description`. Makes the tests easier to reference, and save the test framework from having to modify the name to not have spaces.

account/account_test.go
```diff
@@ -344,3 +355,132 @@ func TestConvertGDPREnforcePurposeFields(t *testing.T) {
...
+		t.Run(test.description, func(t *testing.T) {
+			metrics := &metrics.MetricsEngineMock{}
+			metrics.Mock.On(test.givenMetric, mock.Anything, mock.Anything).Return()
+			metrics.Mock.On("RecordAccountUpgradeStatus", mock.Anything, mock.Anything).Return()
+
+			_, _ = GetAccount(context.Background(), cfg, fetcher, test.givenAccountID, metrics)
+
+			metrics.AssertNumberOfCalls(t, test.givenMetric, test.expectedMetricCount)
+			metrics.AssertNumberOfCalls(t, "RecordAccountUpgradeStatus", test.expectedMetricCount)
```
hhhjort, line 405: Are there no cases where `RecordAccountUpgradeStatus` is called or not called independently of the other given metrics? If not, would it be better to just drop this metrics and replace it with a sum over the other metrics?

account/account_test.go
```diff
@@ -344,3 +355,132 @@ func TestConvertGDPREnforcePurposeFields(t *testing.T) {
...
+	for _, test := range testCases {
+		metrics := &metrics.MetricsEngineMock{}
+		metrics.Mock.On("RecordAccountGDPRPurposeWarning", mock.Anything, mock.Anything).Return()
+		metrics.Mock.On("RecordAccountUpgradeStatus", mock.Anything, mock.Anything).Return()
+
+		_, err := ConvertGDPREnforcePurposeFields(test.givenConfig, metrics, "acct-id")
+		assert.NoError(t, err)
+
+		metrics.AssertNumberOfCalls(t, "RecordAccountGDPRPurposeWarning", test.expectedMetricCount)
+		metrics.AssertNumberOfCalls(t, "RecordAccountUpgradeStatus", test.expectedMetricCount)
```
hhhjort, line 483: Again, are there no cases where `RecordAccountUpgradeStatus` is called or not called independently of the other given metrics? If not, would it be better to just drop this metrics and replace it with a sum over the other metrics?

bsardo: @hhhjort you're right that as of right now `RecordAccountUpgradeStatus` is called every time some other more detailed metric function is called. The plan though is for `RecordAccountUpgradeStatus` to persist long-term but for all of the other field-specific metrics to be replaced by logging once fetchers 2.0 is complete since we'll then be able to perform smarter logging that won't flood the logs. The `RecordAccountUpgradeStatus` metric would then serve as an indicator that you should check your logs for more detailed info about your account config problems.

One question I have at this moment though is whether `RecordAccountUpgradeStatus` should count the number of times an account is referenced that contains at least one account config issue, or if it should count the number of account config issues detected across all of the account references. This PR is currently doing the latter but I'm thinking the former might make more sense.
Thoughts?

For example, assume account A uses one deprecated field and account B uses two deprecated fields.
In the first case, every time account A is referenced we increment by 1 and every time account B is referenced we increment by 1.
In the second case, every time acount A is referenced we increment by 1 and every time B is referenced we increment by 2.

hhhjort: I don't have a strong feeling for this either way. I think the important information is which accounts still need to be updated, which really only cares if the number recorded >=1 for a particular account, not what that number turns out to be. Also there would be uncertainty with a high traffic account being loaded more often than a low traffic one, so even with a high number does that mean it has a lot of fields that need updated, or just that the load was called rather often.

Otherwise this PR is looking good.

bsardo (commented):
Looks good. Just a few cleanup comments.

bsardo (approved):

hhhjort (approved):
LGTM

Issue #2437, opened 2022-10-27
Prebid Server Rules Enforcement


bretg:
We had a discussion during backlog grooming about `rules`. There are several rules that we've been relaxed about enforcing in Prebid Server that Prebid.org publisher members are increasingly wanting to enforce.

The Prebid.js team has gone through a couple of rounds of tightening rules on client-side adapters, and many adapters have been removed. It's time for the Prebid Server community to consider which rules will be strictly enforced and a timeline for enforcement.

Here's an initial proposal based on the [module rules reference](https://docs.prebid.org/dev-docs/module-rules.html#prebidjs-and-prebid-server-module-rules).

### Strictly enforceable rules

1. The company maintaining a bid adapter must be responsive to requests for support and maintenance. 
2. The contact info for bid adapters must be a current group address - not an individual.
3. Endpoint domain names cannot be fully variable. See https://github.com/prebid/prebid-server/issues/2127
4. If the adapter has a client-side adapter, all parameters must be consistent between client- and server-side adapters. This allows publishers to utilize the PBJS [s2sTesting module](https://docs.prebid.org/dev-docs/modules/s2sTesting.html).
5. Bid adapters cannot contact their endpoints directly, but only through the PBS-core interface.
6. Bid adapters cannot include unapproved libraries.
7. Bidder modules must not make requests to endpoints for functionality other than auctions.
8. Bidder modules must not cache bids from previous auctions.

### Rules that will eventually be enforced
1. Once the [iframe-multisync feature](https://github.com/prebid/prebid-server/issues/1986) is available, bid adapters will have a time horizon to ensure that their Iframe usersyncs multi-sync according to the limits imposed.
2. Bidder modules should supply buyer metadata - including advertiser, advertiser domain, network, actual mediatype, and others defined in the Prebid documentation.

### Recommended, but not strictly enforced
1. Bid adapters should support gzip compression on requests from PBS to their auction endpoints.
3. Bid adapters should understand how their endpoints handle currency and floors and transform the PBS request appropriately.
4. If bid adapters are reporting multiple media types in the hb_format as ‘banner’, they should indicate the actual mediatype in the metadata object.

SyntaxNode:
Feedback from a discussion with the Prebid Server Committee:

> Strictly enforceable rules:
> 3. Endpoint domain names cannot be fully variable. See https://github.com/prebid/prebid-server/issues/2127

[...]

PR #2539, opened 2023-01-24
New Adapter: Limelight Digital

imedvedko:
Docs: https://github.com/prebid/prebid.github.io/pull/4297


bretg:
As noted in the docs PR https://github.com/prebid/prebid.github.io/pull/4297, please update this adapter to be able to support the same set of parameters as your client-side adapter.

imedvedko:
ping :-)
As noted in the docs PR https://github.com/prebid/prebid.github.io/pull/4297, we updated this adapter to be able to support the same set of parameters as your client-side adapter. Could you please review it?

static/bidder-info/iionads.yaml
```diff
@@ -0,0 +1,16 @@
+endpoint: "https://{{.Host}}/openrtb/{{.PublisherID}}"
```
VeronikaSolovei9, line 1: This is so weird to see URL like this, but I guess it's possible too. 
Do you want to add validation for the "host" parameter to make sure it's URL, not just a random set of characters?

VeronikaSolovei9: Thank you. We discussed this with the team. Unfortunately this URL format may lead to a potential security issue. We recommend to modify it to something like this (based on unit tests you provided):
"https://www.limelightDigital.{{.MediaType}}.test.com/openrtb/{{.PublisherID}}"

SyntaxNode: Correct @VeronikaSolovei9. We do not allow the domain name to be configurable due to security issues. Additionally, we highly discourage subdomains as they negatively affect the performance of maintaining connects to your bidding server. Please see our bidder development guide for further information: https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html

I am aware there are a few adapters currently in violation. The Prebid Server Committee is working on a remediation plan.

imedvedko: As noted in the docs PR https://github.com/prebid/prebid.github.io/pull/4297, we need to support the same parameters with our Prebid.js adapter. Our Prebid.js has `Host` parameter. And a lot of our client use it on their custom domains which are owned by our clients and which we do not directly control. Are you suggesting we set up cname dns records from our global domain to their custom domains and hope to win performance?

imedvedko: Is it perfect solution if we will create a new alias adapter for each our client with his own domain? No subdomains. Ho `Host` parameters. Will it be easier to maintain? Will you win in performance and connections?

adapters/limelightDigital/limelightDigital.go
```diff
@@ -0,0 +1,172 @@
...
+				Bid:     &seatBid.Bid[i],
+				BidType: getMediaTypeForBid(bid.ImpID, request.Imp),
+			}
+			bidResponse.Bids = append(bidResponse.Bids, b)
+		}
+	}
+	return bidResponse, nil
+}
+
+func (a *adapter) getImpressionExt(imp *openrtb2.Imp) (*openrtb_ext.ImpExtLimelightDigital, error) {
```
Sonali-More-Xandr, line 131: Helper function need not be defined on an adapter. It can simply defined as:-

`func getImpressionExt(imp *openrtb2.Imp) (*openrtb_ext.ImpExtLimelightDigital, error)`

adapters/limelightDigital/limelightDigital.go
```diff
@@ -0,0 +1,172 @@
...
+			if imp.Banner != nil {
+				return openrtb_ext.BidTypeBanner
+			} else if imp.Video != nil {
+				return openrtb_ext.BidTypeVideo
+			} else if imp.Audio != nil {
+				return openrtb_ext.BidTypeAudio
+			} else if imp.Native != nil {
+				return openrtb_ext.BidTypeNative
+			}
+			return openrtb_ext.BidTypeBanner
```
Sonali-More-Xandr, line 168: Could you please help me understand why the BidTypeBanner is returned in this case as well on line 171?

adapters/limelightDigital/limelightDigital.go
```diff
@@ -0,0 +1,172 @@
...
+	if err := json.Unmarshal(bidderExt.Bidder, &limelightDigitalExt); err != nil {
+		return nil, &errortypes.BadInput{
+			Message: "ext.bidder is not provided",
+		}
+	}
+	imp.Ext = nil
+	return &limelightDigitalExt, nil
+}
+
+func (a *adapter) buildEndpointURL(params *openrtb_ext.ImpExtLimelightDigital) (string, error) {
```
Sonali-More-Xandr, line 148: This function need not be defined on adapter. This can be done as follow:-

`func buildEndpointURL(params *openrtb_ext.ImpExtLimelightDigital) (string, error)`

adapters/limelightDigital/LimelightDigitaltest/supplemental/empty_object reponse.json
```diff
@@ -0,0 +1,52 @@
+{
```
Sonali-More-Xandr, line 1: Nitpick - missing underscore ("_") in the filename. Recommend renaming it as :-
empty_object_response.json

imedvedko: Sorry. I fixed it in https://github.com/prebid/prebid-server/pull/2539/commits/a62aea11c7a63b52f404abd5cd74592861825fb3 commit

adapters/limelightDigital/limelightDigital.go
imedvedko: I used `endpointTemplate` adapter param on 153 line. I rewrite it and receive template as argument in https://github.com/prebid/prebid-server/pull/2539/commits/b3d20782a84aaa4c9c184bbfb0dd9c896d56e7b9 commit. Thanks

adapters/limelightDigital/limelightDigital.go
imedvedko: Thank you so much for your review It really looks like an opportunity for improvement. I improve BidType checking in https://github.com/prebid/prebid-server/pull/2539/commits/a8efcf771fdf43622499d870cd6f88a7e92c9bf2 and create a test which checks this case

imedvedko: I renamed test in https://github.com/prebid/prebid-server/pull/2539/commits/fb67dbd204813d43f18ae5a662b523d1e8731fd3 commit

adapters/limelightDigital/limelightDigital.go
VeronikaSolovei9: I'm not sure about this change. For other adapters  `buildEndpointURL` gets an adapter as a receiver. Please refer to usages of this function in other adapters.

adapters/limelightDigital/limelightDigital.go
VeronikaSolovei9: Good point. With new changes L174 can be fully removed or modified to `break`
Also please add test for `native` media type

adapters/limelightDigital/limelightDigital.go
```diff
@@ -0,0 +1,178 @@
...
+			return nil, append(errors, err)
+		}
+
+		url, err := buildEndpointURL(a.endpointTemplate, limelightDigitalExt)
+		if err != nil {
+			return nil, []error{err}
+		}
+
+		// Check if imp comes with bid floor amount defined in a foreign currency
+		if imp.BidFloor > 0 && imp.BidFloorCur != "" && strings.ToUpper(imp.BidFloorCur) != "USD" {
```
VeronikaSolovei9, line 56: please add test for this case

adapters/limelightDigital/limelightDigital.go
imedvedko: @VeronikaSolovei9 Thank you. I revert these changes in https://github.com/prebid/prebid-server/pull/2539/commits/4af1869e6e0d25e15621ec663aa6241e25b43cda commit. Thanks

adapters/limelightDigital/limelightDigital.go
imedvedko: @VeronikaSolovei9 I improved unknown media type filtering in commit  https://github.com/prebid/prebid-server/pull/2539/commits/0564ec79541d9b339727fcfe1f29aceb5e150698 commit and added native media type test in https://github.com/prebid/prebid-server/pull/2539/commits/25733951529ac9f1415931d326d3955a93978729 commit

adapters/limelightDigital/limelightDigital.go
imedvedko: Thanks for your comments. I added the test in https://github.com/prebid/prebid-server/pull/2539/commits/ef1e12ee84b4deedffd9d75323bf497629177837 commit. Could you check it and resolve the conversation?

openrtb_ext/imp_limelightDigital.go
```diff
@@ -0,0 +1,8 @@
+package openrtb_ext
+
+import "encoding/json"
+
+type ImpExtLimelightDigital struct {
```
Sonali-More-Xandr, line 5: The [doc PR](https://github.com/prebid/prebid.github.io/pull/4297/files) has required fields adUnitId and adUnitType which are missing here. If those fields are required, could you please consider adding those here as well, if not, then please update the doc PR to match the parameters.

static/bidder-info/iionads.yaml
```diff
@@ -0,0 +1,16 @@
...
+maintainer:
+  email: "engineering@project-limelight.com"
+capabilities:
+  app:
+    mediaTypes:
```
Sonali-More-Xandr, line 6: The mediaTypes mentioned in the [doc PR](https://github.com/prebid/prebid.github.io/pull/4297/files) are not matching. Could you please update the doc PR to reflect all the supported media types.

openrtb_ext/imp_limelightDigital.go
imedvedko: @Sonali-More-Xandr It is required fields for our Prebid.js dapter. As noted in the docs PR https://github.com/prebid/prebid.github.io/pull/4297, we need to support the same set of parameters with our Prebid.js adapter.

imedvedko: We can not remove these fields from doc because are required fields for our Prebid.js adapter. And we don't need to use these filed in our Prebid Server adapter. Should we add parameters that we don't use to our Prebid Server adapter?

bretg: The server should be able to accept the same parameters, ignoring them if not needed server-side.

This is to allow publishers to use the server-to-server test module, which uses the same set of configuration for your bidder.

static/bidder-info/iionads.yaml
imedvedko: As noted in the docs PR https://github.com/prebid/prebid.github.io/pull/4297, we have to use the same documentation with our Prebid.js adapter. This means that we must specify in the documentation only the set of media types that both our Prebid.js and Prebid Server adapters support.

bretg: It would be fine to add a note to your docs to explain this. e.g. "the client-side adapter doesn't support X, but the server-side adapter does."

openrtb_ext/imp_limelightDigital.go
imedvedko: @bretg Thanks, you are absolutely right. The server is already able to accept the same parameters, it ignore them. I added tests with this case in https://github.com/prebid/prebid-server/pull/2539/commits/0b780e88c6cda75ce2fa382633a7959061224e4b commit

static/bidder-info/iionads.yaml
imedvedko: @bretg Thank you for your comment. I added this note in https://github.com/prebid/prebid.github.io/pull/4297/commits/d7223cf3e374f2e9a4e085283b63fcfe944c34a8 commit

static/bidder-info/iionads.yaml
bsardo: Hi @imedvedko. Thanks for your patience and for working with us on this feedback.

The two main concerns we have for using a macro in the domain are security exposure and connection performance. The security aspect is alleviated by using a fixed top level domain. Due to the potential harm to hosts, we are strict in this requirement. We are working towards fixing  the few adapters currently in violation.

The connection performance advice is for your benefit. Client specific subdomains prevent Prebid Server from reusing connections across your clients which results in more connections needed for your adapter. The issue gets worse the more successful you are in attracting new clients. Hosts may choose to disable your adapter due to this behavior.

The best practice is for your domain to be fixed and any parameters specified in the path or query. How you choose to forward bid requests to your clients is outside of the scope of PBS and for your team to decide. I am hesitant to advise on this but my gut is that cnames are not the answer but perhaps it could be accomplished with vhosts, url rewrites or some other means at the edge.

Please do not create an alias for each client. That is not the intention of this feature.

I hope that helps.

hhhjort: Also note that only the PBS host can see the URL used to connect to your bidder. Well, a debug auction can reveal the http call(s) sent out to the bidder, but beyond that it is completely invisible. So there is little vanity reason for having multiple URLs. A structural reason may exist in the event that each URL goes to a different physical host which each have a different dataset loaded to represent each client. But as Brian points out above, this doesn't seem like a very scalable architecture.

imedvedko: Hi, sorry for the long answer. I added top level domain. For compatibility with our Prebid.js adapter parameters, I left `host` field format in the request the same and just cut out subdomain from it. I also added additional validation of `host` parameter to improve the security of subdomain cutting out https://github.com/prebid/prebid-server/pull/2539/commits/e0b1ab238714f95dea2150f81d3ca3ab4f65b34b#diff-85e257afc296b1b9772d361fe8269082ee7a158207ed84f6fae16e1b42a2426cR12 and added additional params tests https://github.com/prebid/prebid-server/pull/2539/commits/e0b1ab238714f95dea2150f81d3ca3ab4f65b34b#diff-98d261a63ffeb65e201644a1669caaa0d9e6c2fb2a9a9e6de7d429d22814b017R61-R70 . Thanks

imedvedko: I have improved parameters validation through json schema in previous commits. I also added additional domain validation for host parameter in the adapter code in https://github.com/prebid/prebid-server/pull/2539/commits/7b2cc3abd0eb757108d22990267240fa63ca9351 commit. Feel free to request me to rollback this commit if these changes seem redundant to you

hhhjort: This should prevent PBS from being used as a malicious proxy, unless a PBS host or future PR does something silly like define the endpoint as "https://{{.HOST}}.com/openrtb/..." It still has the potential of cluttering up the connection pool depending on how many unique HOST values the adapter might see as it runs on a host. But at least that is just a performance issue rather than a security issue.

bsardo:
Hi @imedvedko, the hold up has been about the macros in your endpoint. With the adjustments you recently made you are within the rules for now, however things will be a bit more strict later this year. I invite you to participate in the discussion in [Prebid Server Rules Enforcement #2437: Dynamic Host Names](https://github.com/prebid/prebid-server/issues/2606) and to help us better understand your use case.

The team will pick this PR up again next week for review. Thanks for your patience.

Sonali-More-Xandr (approved):
Good to go after one more approval.

static/bidder-info/iionads.yaml
```diff
@@ -0,0 +1,16 @@
+endpoint: "http://{{.Host}}.iionads.com/openrtb/{{.PublisherID}}"
```
onkarvhanumante, line 1: should this be `https` ? similar thoughts for `static/bidder-info/limelightDigital.yaml` endpoint

imedvedko: As I can see, it is just not recommended, but it's not prohibited. As I see, a lot of Prebid Server adapters do the same.

imedvedko: Is it blocking the pull request merge?

onkarvhanumante: Although not prohibited now.  But if possible should prefer using `https` over `http`.  Let us know this change is possible. If not then we can merge PR change.

imedvedko: We use `http` for server-to-server communication. Please merge the pull request as it is, if it is possible :-)

bretg: HTTP is tolerated for server-to-server. It's understood that it's more expensive to run load balancers that process SSL.
Client-side adapters must use HTTPS.

imedvedko: @bretg You are absolutely right, so we use `https` in our client side Prebid.js adaper https://github.com/prebid/Prebid.js/blob/master/modules/limelightDigitalBidAdapter.js#L119

onkarvhanumante (approved):

imedvedko:
Thank you so much for the time of each of you and for the deep desire to dive into our needs. :-)

PR #2540, opened 2023-01-24
Initial infrastructure to support bidder openRTB support flags

hhhjort:
This does not make use of the new flags. It also pushes BidderInfos down to cleanOpenRTBRequests() so that the individual "cleaned" requests to the bidders can be modified as per bidder support of openrtb features.


config/bidderinfo_test.go
```diff
@@ -38,6 +39,9 @@ experiment:
   adsCert:
     enabled: true
 endpointCompression: "GZIP"
+openrtb:
+  version: "2.6"
```
SyntaxNode, line 43: Nitpick: The quotes aren't necessary, and we might want to prefer the simple style of `version: 2.6`.

exchange/utils.go
```diff
@@ -49,6 +49,7 @@ func cleanOpenRTBRequests(ctx context.Context,
 	gdprPermsBuilder gdpr.PermissionsBuilder,
 	tcf2ConfigBuilder gdpr.TCF2ConfigBuilder,
 	hostSChainNode *openrtb2.SupplyChainNode,
+	bidderInfos config.BidderInfos,
```
SyntaxNode, line 52: We already have way too many argument. Perhaps we can leave this out for now and address the needs individually. For example, the gpp supported flag might fit nicely in the privacyConfig object.

guscarreon: Should we pass the `bidderInfos` parameter all the way down to the `getAuctionBidderRequests(...)` function as part of this PR? It's likely that we'll need it there.
```
 75 
 76      var allBidderRequests []BidderRequest
 77 -    allBidderRequests, errs = getAuctionBidderRequests(auctionReq, requestExt, bidderToSyncerKey, impsByBidder, aliases, hostSChainNode)
    +    allBidderRequests, errs = getAuctionBidderRequests(auctionReq, requestExt, bidderToSyncerKey, impsByBidder, aliases, hostSChainNode, bidderInfos)
 78 
 79     bidderNameToBidderReq := buildBidResponseRequest(req.BidRequest, bidderImpWithBidResp, aliases, auctionReq.BidderImpReplaceImpID)
 80     //this function should be executed after getAuctionBidderRequests
 81     allBidderRequests = mergeBidderRequests(allBidderRequests, bidderNameToBidderReq)
 82
 83 +-- 98 lines: var gpp gpplib.GppContainer-------------------------------------------------------------------------------------
181     return
182 }
183
184 +-- 33 lines: func ccpaEnabled(account *config.Account, privacyConfig config.Privacy, requestType config.ChannelType) bool {--
217
218   func getAuctionBidderRequests(auctionRequest AuctionRequest,
219       requestExt *openrtb_ext.ExtRequest,
220       bidderToSyncerKey map[string]string,
221       impsByBidder map[string][]openrtb2.Imp,
222       aliases map[string]string,
223 -     hostSChainNode *openrtb2.SupplyChainNode) ([]BidderRequest, []error) {
    +     hostSChainNode *openrtb2.SupplyChainNode,
    +     bidderInfos config.BidderInfos) ([]BidderRequest, []error) {
224  
225       bidderRequests := make([]BidderRequest, 0, len(impsByBidder))
226       req := auctionRequest.BidRequestWrapper
227*-- 14 lines: explicitBuyerUIDs, err := extractBuyerUIDs(req.BidRequest.User)--------------------------------------------------
241  
242       var errs []error
243       for bidder, imps := range impsByBidder {
244           coreBidder := resolveBidder(bidder, aliases)
245  
    +         // We'll probably have to pass a 2.5 or 2.6 request copy depending on the value of the bidder's openrtb flag here
246 -         reqCopy := *req.BidRequest
    +         reqCopy := getVersionCompatibleBidRequest(req.BidRequest, coreBidder, bidderInfos)
247           reqCopy.Imp = imps
exchange/utils.go
```

config/bidderinfo.go
```diff
@@ -87,6 +88,14 @@ type AdapterXAPI struct {
 	Tracker  string `yaml:"tracker" mapstructure:"tracker"`
 }
 
+// OpenRTBInfo specifies the versions/aspects of openRTB that a bidder supports
+// Version is not yet actively supported
+// GPPSupported is not yet actively supported
+type OpenRTBInfo struct {
+	Version      string `yaml:"version" mapstructure:"version"`
+	GPPSupported bool   `yaml:"gpp-supported" mapstructure:"gpp-supported"`
```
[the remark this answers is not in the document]
SyntaxNode: The default is false, so I don't think we need to make it a pointer,

exchange/exchange.go
```diff
@@ -269,7 +269,7 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 	gdprDefaultValue := e.parseGDPRDefaultValue(r.BidRequestWrapper.BidRequest)
 
 	// Slice of BidRequests, each a copy of the original cleaned to only contain bidder data for the named bidder
-	bidderRequests, privacyLabels, errs := cleanOpenRTBRequests(ctx, r, requestExt, e.bidderToSyncerKey, e.me, gdprDefaultValue, e.privacyConfig, e.gdprPermsBuilder, e.tcf2ConfigBuilder, e.hostSChainNode)
+	bidderRequests, privacyLabels, errs := cleanOpenRTBRequests(ctx, r, requestExt, e.bidderToSyncerKey, e.me, gdprDefaultValue, e.privacyConfig, e.gdprPermsBuilder, e.tcf2ConfigBuilder, e.hostSChainNode, e.bidderInfo)
```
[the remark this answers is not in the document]
SyntaxNode: A refactor sounds great and I like this idea, but not on the Exchange object. That's too big. Let's break it out into it's own thing if we go this approach.

config/bidderinfo_test.go
hhhjort: Should I get rid of the quotes on the other values above?

exchange/utils.go
hhhjort: Its not a single flag, but a map of bidder names to their info data. So we would need a map[string]bool or something like that, and a code loop on startup to create this new map. It will also be the only piece in the privacy object that is mapped to bidders. I was also thinking it might future-proof the code a bit, depending on what other openrtb feature flags that come up the we might want to handle in `cleanOpenRTBRequests()`. The 2.5<->2.6 transforms I was thinking either belonged in here, or possibly as a wrapper. And there may be other 2.6 compatibility gymnastics we may have to do in the future on a per bidder basis.

TLDR: Could shove the gpp compatibility flag from BidderInfos into the Privacy config struct, which is not a great fit and already exists elsewhere in the general config struct. But there are other pieces of the BidderInfos we may need in `cleanOpenRTBRequests()` that are not in any way privacy configs that we may need now or in the future.

config/bidderinfo.go
```diff
@@ -87,6 +88,14 @@ type AdapterXAPI struct {
...
+}
+
```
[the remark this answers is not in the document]
hhhjort: Thanks, good catch. I forgot about that piece.

exchange/exchange.go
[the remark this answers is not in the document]
hhhjort: That goes along with Scott's comment above. This could definitely use a refactor. Perhaps even make it a method of the exchange, so it has direct access to the exchange's fields rather than needing to be passed them explicitly. It already had 6 fields from the exchange struct. My addition bumps it up to 7. That may not be too bad of a refactor for this small PR.

config/bidderinfo_test.go
SyntaxNode: Sounds like a good idea to keep things simple.

hhhjort:
On pushing down `BidderInfos` to `getAuctionBidderRequests(...)`, I don't think it is needed for GPP as all the privacy handling is done outside of this function. Do you think we need it there for OpenRTB downgrades?

exchange/utils.go
hhhjort: Additional thought, while responding to Gus, I considered the utility of making this function a method of the exchange.  This will allow us to drop 6 other fields from the argument list, giving us a great savings in fields being passed. I would just have to look through the code to insure there is nothing written that depends on these being a copy of the exchange fields (created on passing into the function) rather than the original fields from the exchange. I would be surprised if there is something subtle going on here like that.

guscarreon:
> On pushing down `BidderInfos` to `getAuctionBidderRequests(...)`, I don't think it is needed for GPP as all the privacy handling is done outside of this function. Do you think we need it there for OpenRTB downgrades?

For the OpenRTB I believe we need that info and it makes sense to me to downgrade to 2.5 in that place. But I agree we can takle this later and it's definitely not in the scope of this PR. I'm ok with ignoring that comment.

config/config.go
```diff
@@ -1420,6 +1420,8 @@ func setBidderDefaults(v *viper.Viper, bidder string) {
 	v.BindEnv(adapterCfgPrefix+".xapi.password", "")
 	v.BindEnv(adapterCfgPrefix+".xapi.tracker", "")
 	v.BindEnv(adapterCfgPrefix+".endpointCompression", "")
+	v.BindEnv(adapterCfgPrefix+".openrtb.version", "")
+	v.BindEnv(adapterCfgPrefix+".openrtb.gpp-supported", "")
```
SyntaxNode, line 1424: Please remove the last `""` argument. That is an incorrect calling convention for `BindEnv`.

guscarreon (commented):
LGTM. We are only missing the following changes to test `TestBidderInfoFromEnv` found inside `config/config_test.go`:
```
diff --git a/config/config_test.go b/config/config_test.go
index 43ab9273..40c4a53a 100644
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -861,6 +861,12 @@ func TestBidderInfoFromEnv(t *testing.T) {
 		defer os.Unsetenv("PBS_ADAPTERS_BIDDER1_USERSYNC_REDIRECT_URL")
 	}
 
+	if oldval, ok := os.LookupEnv("PBS_ADAPTERS_BIDDER1_OPENRTB_VERSION"); ok {
+		defer os.Setenv("PBS_ADAPTERS_BIDDER1_OPENRTB_VERSION", oldval)
+	} else {
+		defer os.Unsetenv("PBS_ADAPTERS_BIDDER1_OPENRTB_VERSION")
+	}
+
 	// set new
 	os.Setenv("PBS_ADAPTERS_BIDDER1_DISABLED", "true")
 	os.Setenv("PBS_ADAPTERS_BIDDER1_ENDPOINT", "http://some.url/override")
@@ -870,6 +876,7 @@ func TestBidderInfoFromEnv(t *testing.T) {
 	os.Setenv("PBS_ADAPTERS_BIDDER1_EXPERIMENT_ADSCERT_ENABLED", "true")
 	os.Setenv("PBS_ADAPTERS_BIDDER1_XAPI_USERNAME", "username_override")
 	os.Setenv("PBS_ADAPTERS_BIDDER1_USERSYNC_REDIRECT_URL", "http://some.url/sync?redirect={{.RedirectURL}}")
+	os.Setenv("PBS_ADAPTERS_BIDDER1_OPENRTB_VERSION", "2.6")
 
 	cfg, _ := newDefaultConfig(t)
 
@@ -882,6 +889,8 @@ func TestBidderInfoFromEnv(t *testing.T) {
 
 	assert.Equal(t, true, cfg.BidderInfos["bidder1"].Experiment.AdsCert.Enabled)
 	assert.Equal(t, "username_override", cfg.BidderInfos["bidder1"].XAPI.Username)
+
+	assert.Equal(t, "2.6", cfg.BidderInfos["bidder1"].OpenRTB.Version)
 }
 
 func TestMigrateConfigPurposeOneTreatment(t *testing.T) {
 ```

hhhjort:
Thanks, good catch Gus. Updating the test.

SyntaxNode (approved):

guscarreon (approved):
LGTM

PR #2545, opened 2023-01-26
Address CVE-2022-41721

VeronikaSolovei9:
updated version of golang.org/x/net to v0.4.0


go.mod
```diff
@@ -30,8 +30,8 @@ require (
 	github.com/vrischmann/go-metrics-influxdb v0.1.1
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/yudai/gojsondiff v1.0.0
-	golang.org/x/net v0.0.0-20220909164309-bea034e7d591
-	golang.org/x/text v0.3.7
+	golang.org/x/net v0.4.0
+	golang.org/x/text v0.5.0
```
mansinahar, line 34: Thank you so much @VeronikaSolovei9 for putting out this fix PR! Any reason for pinning the text package version to `v0.5.0`? I see that the latest version for this package is `0.6.0`. Should we/can we move to the latest?

bsardo (commented):
LGTM though I'm also curious about https://github.com/prebid/prebid-server/pull/2545/files#r1088081338.

go.mod
SyntaxNode: Yes, please move to the latest version available.

mansinahar: Same for the /net package, the latest version is [v0.5.0](https://pkg.go.dev/golang.org/x/net?tab=versions). Can we please pin it to that?

VeronikaSolovei9:
I put v0.4.0 because Trivy suggested to put this as a fixed version. I agree latest would be better. I checked it works with v0.5.0

mansinahar (approved):

PR #2549, opened 2023-01-27
PBS core: add hb_deal targeting to video endpoint response 

shahinrahbariasl:
Greetings, we recently found out that the /video endpoint of PBS doesn't add `hb_deal` targeting to the response. This is a small change to add `hb_deal` to targeting response as well if it exists in bid response from bidders. I also understand that the /video endpoint is going to be deprecated sometime soon. We would like to make this change though since we're planning to use the /video endpoint in the meantime for some testing purposes.

with `dealID` in bidder response:
Screen Shot 2023-01-26 at 1 33 41 PM

without `dealID` in bidder response:
Screenshot 2023-01-27 at 11 08 34 AM

### DIFF
--- a/endpoints/openrtb2/video_auction.go
+++ b/endpoints/openrtb2/video_auction.go
@@ -516,6 +516,7 @@ func buildVideoResponse(bidresponse *openrtb2.BidResponse, podErrors []PodError)
 				HbPb:       tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbpbConstantKey, seatBid.Seat)],
 				HbPbCatDur: tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbCategoryDurationKey, seatBid.Seat)],
 				HbCacheID:  tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbVastCacheKey, seatBid.Seat)],
+				HbDeal:     tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbDealIDConstantKey, seatBid.Seat)],
 			}
 
 			adPod := findAdPod(podId, adPods)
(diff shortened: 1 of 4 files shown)


shahinrahbariasl:
@SyntaxNode @bretg hoping to get your thoughts on this, it's a small change and I don't think affects anything else, just that we can receive the `hb_deal` KV in video response and use it in ad server call if needed.

bretg:
Sounds fine to me. Thanks for jumping in with a PR.

endpoints/openrtb2/video_auction.go
```diff
@@ -516,6 +516,7 @@ func buildVideoResponse(bidresponse *openrtb2.BidResponse, podErrors []PodError)
 				HbPb:       tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbpbConstantKey, seatBid.Seat)],
 				HbPbCatDur: tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbCategoryDurationKey, seatBid.Seat)],
 				HbCacheID:  tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbVastCacheKey, seatBid.Seat)],
+				HbDeal:     tempRespBidExt.Prebid.Targeting[formatTargetingKey(openrtb_ext.HbDealIDConstantKey, seatBid.Seat)],
```
hhhjort, line 519: The other items here should always exist for every bid. Will this create an empty `hb_deal` targeting key with an empty value when there is no deal? I suspect the final JSON will not have an empty `hb_deal` entry, but we should have a test to verify we are not introducing a bunch of empty `hb_deal` keys for bids that have no deals attached.

shahinrahbariasl: Hi @hhhjort , so this won't create a new targeting KV if dealID doesn't exist in bidder response as you can see here: https://github.com/prebid/prebid-server/blob/master/exchange/targeting.go#L69

And then we also have omit empty json tag here so it wouldn't include empty field: https://github.com/prebid/prebid-server/pull/2549/files#diff-ba7cfdcbad8cf8fe91fd813c0487adc8d08e9e63e7eca8eba365dbb222b92128R20

You can see this existing test demonstrate that: https://github.com/prebid/prebid-server/blob/master/exchange/targeting_test.go#L434 but I just added another test case to explicitly check that. You can see the example final responses in the screen shots as well. Let me know if the new test is sufficient. Thanks!

hhhjort (approved):
LGTM

mansinahar (approved):

PR #2560, opened 2023-01-31
New Adapter: VisibleMeasures

VisibleMeasuresHB:
Doc: https://github.com/prebid/prebid.github.io/pull/4312


SyntaxNode:
Please merge in the latest changes from the `master` branch to resolve the Trivy code scanning failure.

VisibleMeasuresHB:
@SyntaxNode last changes from `master` was merged.

adapters/visiblemeasures/visiblemeasures.go
```diff
@@ -0,0 +1,158 @@
...
+
+		temp := reqBodyExt{VisibleMeasuresBidderExt: reqBodyExtBidder{}}
+
+		if visiblemeasuresExt.PlacementID != "" {
+			temp.VisibleMeasuresBidderExt.PlacementID = visiblemeasuresExt.PlacementID
+			temp.VisibleMeasuresBidderExt.Type = "publisher"
+		} else if visiblemeasuresExt.EndpointID != "" {
+			temp.VisibleMeasuresBidderExt.EndpointID = visiblemeasuresExt.EndpointID
+			temp.VisibleMeasuresBidderExt.Type = "network"
+		}
```
onkarvhanumante, line 62: should return error if both `PlacementID` and `EndpointID` are blank

adapters/visiblemeasures/visiblemeasures.go
```diff
@@ -0,0 +1,158 @@
...
+	if responseData.StatusCode == http.StatusNoContent {
+		return nil, nil
+	}
+
+	if responseData.StatusCode != http.StatusOK {
+		err := &errortypes.BadServerResponse{
+			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info.", responseData.StatusCode),
+		}
+		return nil, []error{err}
+	}
```
onkarvhanumante, line 110: should also add check for `http.StatusBadRequest` status

static/bidder-info/visiblemeasures.yaml
```diff
@@ -0,0 +1,19 @@
...
+    
+  app:
+    mediaTypes:
+    - banner
+    - video
+    - native
+userSync:
+  redirect:
+    url: "https://cs.visiblemeasures.com/pserver?gdpr={{.GDPR}}&gdpr_consent={{.GDPRConsent}}&us_privacy={{.USPrivacy}}&redir={{.RedirectURL}}"
+    userMacro: "[UID]"
```
onkarvhanumante, line 19: redirect url returns 404 status response

url: https://cs.visiblemeasures.com/pserver?gdpr=&gdpr_consent=&us_privacy=&redir=https%3A%2F%2Fprebid.adnxs.com%2Fpbs%2Fv1%2Fsetuid%3Fbidder%3Dvisiblemeasures%26gdpr%3D%26gdpr_consent%3D%26f%3Di%26uid%3D%5BUID%5D
https://prebid.adnxs.com/pbs/v1/setuid?bidder=visiblemeasures&gdpr=&gdpr_consent=&f=i&uid=[UID]

Screenshot 2023-02-09 at 2 08 25 PM

VisibleMeasuresHB: It fixed, should works now

adapters/visiblemeasures/visiblemeasures.go
VisibleMeasuresHB: This should not happen, because in schema `static/bidder-params/visiblemeasures.json` described that one of two is required:
`"oneOf": [
      { "required": ["placementId"] },
      { "required": ["endpointId"] }
    ]`

adapters/visiblemeasures/visiblemeasures.go
```diff
@@ -0,0 +1,158 @@
...
+		endpoint: config.Endpoint,
+	}
+	return bidder, nil
+}
+
+func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
+	var err error
+	var adapterRequests []*adapters.RequestData
+
+	reqCopy := *request
```
Sonali-More-Xandr, line 40: Nitpick - This is correct. Although, a suggestion to further optimize it by simply copying `imp` slice reference instead of entire `request`.

```
func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
	var err error

	var originalImpSlice []openrtb2.Imp = request.Imp

	for i := range originalImpSlice {
		request.Imp = []openrtb2.Imp{originalImpSlice[i]}

		var visiblemeasuresExt openrtb_ext.ImpExtVisibleMeasures

		if err = json.Unmarshal(request.Imp[0].Ext, &bidderExt); err != nil {
			return nil, []error{err}
		}
		if err = json.Unmarshal(bidderExt.Bidder, &visiblemeasuresExt); err != nil {
			return nil, []error{err}
		}

		temp := reqBodyExt{VisibleMeasuresBidderExt: reqBodyExtBidder{}}

		if visiblemeasuresExt.PlacementID != "" {
			temp.VisibleMeasuresBidderExt.PlacementID = visiblemeasuresExt.PlacementID
			temp.VisibleMeasuresBidderExt.Type = "publisher"
		} else if visiblemeasuresExt.EndpointID != "" {
			temp.VisibleMeasuresBidderExt.EndpointID = visiblemeasuresExt.EndpointID
			temp.VisibleMeasuresBidderExt.Type = "network"
		}

		finalyImpExt, err := json.Marshal(temp)
		if err != nil {
			return nil, []error{err}
		}

		request.Imp[0].Ext = finalyImpExt

		adapterReq, err := a.makeRequest(request)
		if err != nil {
			return nil, []error{err}
		}

		if adapterReq != nil {
			adapterRequests = append(adapterRequests, adapterReq)
		}
	}
	return adapterRequests, nil
}
```

adapters/visiblemeasures/visiblemeasures.go
```diff
@@ -0,0 +1,158 @@
...
+		if index.Banner != nil {
+			return openrtb_ext.BidTypeBanner, nil
+		}
+		if index.Video != nil {
+			return openrtb_ext.BidTypeVideo, nil
+		}
+		if index.Native != nil {
+			return openrtb_ext.BidTypeNative, nil
+		}
+	}
```
onkarvhanumante, line 153: This is an anti-pattern. Also, this for loop code assumes that impression could either be Banner or Video and not both.  But single impression can have both banner and video in the request. In that case, this code will just assumes that it is banner.
Ideally, you should set response bid type based on bidder server response.

VisibleMeasuresHB: In our case we don't support multi format, and we are also satisfied with the banner in that case.
Сould you please share what it should look like correctly?

Sonali-More-Xandr (approved):
LGTM! GTG after one more approval.

adapters/visiblemeasures/visiblemeasures.go
onkarvhanumante: Since multi format bid is not supported so currently made changes should be enough.  Also as an fyi, `openrtb2.BidResponse` has `MType`  field which can be set in bidder server response and then used in  `MakeBids` to determine impression type

refer https://github.com/prebid/openrtb/blob/main/openrtb2/bid.go#L322-L334 
```
	// Attribute:
	//   mtype
	// Type:
	//   integer
	// Description:
	//   Type of the creative markup so that it can properly be
	//   associated with the right sub-object of the BidRequest.Imp.
	//   Values:
	//   1 = Banner
	//   2 = Video
	//   3 = Audio
	//   4 = Native
	MType MarkupType `json:"mtype,omitempty"`

```

onkarvhanumante (approved):

PR #2567, opened 2023-02-06
Refactor Targeting + Price Granularity JSON Unmarshal

SyntaxNode:
Defaults for targeting and price granularity were embedded into the json unmarshal hooks which caused unpredictable behavior. You couldn't tell if where defaults were set just looking at code since it depended on when the objects are unmarshalled. The unmarshal order is changed by https://github.com/prebid/prebid-server/pull/2423 and tests broken.

Refactored to a deterministic order.

Co-Authored By @guscarreon


endpoints/openrtb2/amp_auction.go
```diff
@@ -690,27 +695,23 @@ func defaultRequestExt(req *openrtb_ext.RequestWrapper) []error {
...
 	}
 
-	// Ensure Targeting and caching is on
+	// create targeting object if missing
 	if prebid.Targeting == nil {
-		prebid.Targeting = &openrtb_ext.ExtRequestTargeting{
-			IncludeWinners:    true,
-			IncludeBidderKeys: true,
-			PriceGranularity:  openrtb_ext.PriceGranularityFromString("med"),
-		}
```
SyntaxNode, line 704: These are just the defaults. They are still set as SetDefaults is called after this. Same approach is used for the video endpoint.

endpoints/openrtb2/amp_auction_test.go
```diff
@@ -1651,7 +1709,7 @@ func TestBuildAmpObject(t *testing.T) {
 					},
 					AT:   1,
 					TMax: 500,
-					Ext:  json.RawMessage(`{"prebid":{"cache":{"bids":{"returnCreative":null},"vastxml":null},"channel":{"name":"amp","version":""},"targeting":{"pricegranularity":{"precision":2,"ranges":[{"min":0,"max":20,"increment":0.1}]},"includewinners":true,"includebidderkeys":true,"includebrandcategory":null,"includeformat":false,"durationrangesec":null,"preferdeals":false}}}`),
+					Ext:  json.RawMessage(`{"prebid":{"cache":{"bids":{}},"channel":{"name":"amp","version":""},"targeting":{"pricegranularity":{"precision":2,"ranges":[{"min":0,"max":20,"increment":0.1}]},"includewinners":true,"includebidderkeys":true}}}`),
```
SyntaxNode, line 1712: Added a lot of omitempty declarations into the targeting objects. Doesn't seem to be designed for marshalling. Fixed.

endpoints/openrtb2/sample-requests/amp/consent-through-query/addtl-consent-through-query.json
```diff
@@ -63,11 +71,44 @@
...
+            "ranges": [
+              {
+                "min": 0,
+                "max": 20,
+                "increment": 0.1
+              }
+            ]
+          },
+          "includewinners": true,
+          "includebidderkeys": true
```
SyntaxNode, line 108: Changes due to auto-formatting and removal of omitempty fields. Same goes for most of the json test files.

endpoints/openrtb2/sample-requests/valid-whole/supplementary/req-ext-bidder-params-backward-compatible-merge.json
```diff
@@ -64,7 +64,18 @@
...
+          "ranges": [
+            {
+              "min": 0,
+              "max": 5,
+              "increment": 0.5
+            }
+          ]
+        },
+        "includewinners": true,
+        "includebidderkeys": true
```
SyntaxNode, line 78: This is the definition for "low", now expanded prior to HoldAuction. Matches behavior now with PBS-Java.

endpoints/openrtb2/video_auction_test.go
```diff
@@ -1368,6 +1354,10 @@ type mockExchangeVideo struct {
 }
 
 func (m *mockExchangeVideo) HoldAuction(ctx context.Context, r exchange.AuctionRequest, debugLog *exchange.DebugLog) (*openrtb2.BidResponse, error) {
+	if err := r.BidRequestWrapper.RebuildRequest(); err != nil {
+		return nil, err
+	}
```
SyntaxNode, line 1359: Simulates activity of real HoldAuction. Require for proper ext content.

exchange/price_granularity.go
```diff
@@ -12,7 +12,7 @@ func GetPriceBucket(cpm float64, config openrtb_ext.PriceGranularity) string {
 	cpmStr := ""
 	bucketMax := 0.0
 	increment := 0.0
-	precision := config.Precision
+	precision := *config.Precision
```
SyntaxNode, line 15: Is this potentially dangerous? Yes. But it's ok due to app design and full test coverage. 🤞 Changing this to return an error caused an annoying domino effect for a case which will never occur due to protections run earlier in the flow.

exchange/targeting_test.go
```diff
@@ -281,15 +281,23 @@ var bid084 *openrtb2.Bid = &openrtb2.Bid{
...
+	truncateTargetAttrValue10       int = 10
+	truncateTargetAttrValue5        int = 5
+	truncateTargetAttrValue25       int = 25
+	truncateTargetAttrValueNegative int = -1
+)
+
+func lookupPriceGranularity(v string) openrtb_ext.PriceGranularity {
+	priceGranularity, _ := openrtb_ext.NewPriceGranularityFromLegacyID(v)
+	return priceGranularity
+}
```
SyntaxNode, line 294: Convenience test method to throw away the second return and use directly for assignments in these test methods.

openrtb_ext/request.go
```diff
@@ -197,141 +171,109 @@ type GranularityRange struct {
...
-			// This way we error as expecting the new custom granularity standard.
+	// price granularity used to be a string referencing a predefined value, try to parse
+	// and map the legacy string before falling back to the modern custom model.
+	legacyID := ""
+	if err := json.Unmarshal(b, &legacyID); err == nil {
+		if legacyValue, ok := NewPriceGranularityFromLegacyID(legacyID); ok {
+			*pg = legacyValue
 			return nil
 		}
 	}
```
SyntaxNode, line 183: Still need to keep custom unmarshal behavior for handling legacy id due to type difference (string vs object).

openrtb_ext/request.go
```diff
@@ -197,141 +171,109 @@ type GranularityRange struct {
...
+					Increment: 0.5,
+				},
+			},
+		}, true
 	}
-	// Return empty if not matched
-	return PriceGranularity{}
-}
-
-var priceGranularityLow = PriceGranularity{
```
SyntaxNode, line 271: Moved inline to factory method. Easier to follow, where never exported anyway.

ortb/default.go
```diff
@@ -0,0 +1,76 @@
...
+		targeting.IncludeWinners = ptrutil.ToPtr(DefaultTargetingIncludeWinners)
+		hasChanges = true
+	}
+
+	if targeting.IncludeBidderKeys == nil {
+		targeting.IncludeBidderKeys = ptrutil.ToPtr(DefaultTargetingIncludeBidderKeys)
+		hasChanges = true
+	}
+
+	return hasChanges
```
SyntaxNode, line 60: The change tracking is annoying, but necessary to avoid unneeded marshal calls. Couldn't find a better way to do this in Go.

endpoints/openrtb2/auction.go
```diff
@@ -1499,8 +1504,56 @@ func validateRequestExt(req *openrtb_ext.RequestWrapper) error {
...
+
+	if t.PriceGranularity != nil {
+		pg := t.PriceGranularity
+
+		if pg.Precision == nil {
+			return errors.New("Price granularity error: precision is required")
+		} else if *pg.Precision < 0 {
+			return errors.New("Price granularity error: precision must be non-negative")
+		} else if *pg.Precision > openrtb_ext.MaxDecimalFigures {
+			return errors.New("Price granularity error: precision of more than 15 significant figures is not supported")
```
hhhjort, line 1543: Do we want to use `openrtb_ext.MaxDecimalFigures` in formatting the error message in case this value should ever change?

SyntaxNode: Good idea.

hhhjort (approved):

AlexBVolcy (approved):

PR #2568, opened 2023-02-07
Floors: host and account config

pm-jaydeep-mohite:


account/account.go
```diff
@@ -15,6 +15,14 @@ import (
 	jsonpatch "gopkg.in/evanphx/json-patch.v4"
 )
 
+type AccountPriceFloors struct {
+	Enabled           bool `mapstructure:"enabled" json:"enabled"`
+	EnforceFloorRate  int  `mapstructure:"enforce_floors_rate" json:"enforce_floors_rate"`
+	BidAdjustment     bool `mapstructure:"adjust_for_bid_adjustment" json:"adjust_for_bid_adjustment"`
+	EnforceDealFloors bool `mapstructure:"enforce_deal_floors" json:"enforce_deal_floors"`
+	UseDynamicData    bool `mapstructure:"use_dynamic_data" json:"use_dynamic_data"`
+}
```
SyntaxNode, line 24: This object is already also in the `config` package, which is the better location IMHO.

config/config.go
```diff
@@ -990,6 +1000,12 @@ func SetupViper(v *viper.Viper, filename string, bidderInfos BidderInfos) {
 	v.SetDefault("account_required", false)
 	v.SetDefault("account_defaults.disabled", false)
 	v.SetDefault("account_defaults.debug_allow", true)
+	v.SetDefault("account_defaults.price_floors.enabled", true)
```
SyntaxNode, line 1003: Should we disable this by default until the feature all parts of the feature are reviewed and merged? Same with `price_floors.enabled`.

SyntaxNode (commented):
Thank you for splitting this up into smaller PRs.

config/config.go
pm-jaydeep-mohite: we can disable

account/account.go
pm-jaydeep-mohite: this shall be removed

config/config.go
```diff
@@ -135,6 +140,11 @@ func (cfg *Configuration) validate(v *viper.Viper) []error {
...
 		glog.Warning(`account_defaults.events will currently not do anything as the feature is still under development. Please follow https://github.com/prebid/prebid-server/issues/1725 for more updates`)
 	}
+
+	if cfg.PriceFloors.Enabled {
+		glog.Warning(`cfg.PriceFloors.Enabled  will currently not do anything as price floors feature is still under development.`)
+	}
+
 	errs = cfg.Experiment.validate(errs)
 	errs = cfg.BidderInfos.validate(errs)
 	return errs
```
bsardo, line 150: Can you add validation for these new config fields in this PR?
```
func (pf *AccountPriceFloors) validate(errs []error) []error {

	if pf.EnforceFloorRate < 0 || pf.EnforceFloorRate > 100 {
		errs = append(errs, fmt.Errorf(`account_defaults.price_floors.enforce_floors_rate should be between 0 and 100`))
	}
}
```

and then call this new function from this top-level `validate` function?

We can then claim that this PR covers all non-fetch logic defined in requirements section [4.1.1 Global and Account Config](https://docs.google.com/document/d/1YAt00RfQaU2Ugciqw1LnirLgjyUz9ibfyjfxWUHR_P0/edit#heading=h.vf5nn3il0n2h).

bsardo, line 150: Can you also add a table-based unit test to `config/config_test.go` that will test this new `validate` function I'm asking for in the above comment? We can then expand it later when we add the fetch config later on. You could either call `func (pf *AccountPriceFloors) validate(errs []error) []error {` directly or call the top-level validate function to achieve coverage.

bsardo (commented):
This looks good; I have few of comments. I think this PR should cover all of requirements section [4.1.1 Global and Account Config](https://docs.google.com/document/d/1YAt00RfQaU2Ugciqw1LnirLgjyUz9ibfyjfxWUHR_P0/edit#heading=h.vf5nn3il0n2h) so I've asked for a couple of things to be added to it.

I also noted a couple of things that don't require any changes so other reviewers are aware.

config/config.go
```diff
@@ -990,6 +1000,12 @@ func SetupViper(v *viper.Viper, filename string, bidderInfos BidderInfos) {
 	v.SetDefault("account_required", false)
 	v.SetDefault("account_defaults.disabled", false)
 	v.SetDefault("account_defaults.debug_allow", true)
+	v.SetDefault("account_defaults.price_floors.enabled", false)
+	v.SetDefault("account_defaults.price_floors.enforce_floors_rate", 100)
+	v.SetDefault("account_defaults.price_floors.adjust_for_bid_adjustment", true)
+	v.SetDefault("account_defaults.price_floors.enforce_deal_floors", false)
+	v.SetDefault("account_defaults.price_floors.use_dynamic_data", true)
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: Set to false by default

config/accounts.go
```diff
@@ -52,6 +53,14 @@ type AccountCCPA struct {
 	ChannelEnabled     AccountChannel `mapstructure:"channel_enabled" json:"channel_enabled"`
 }
 
+type AccountPriceFloors struct {
+	Enabled           bool `mapstructure:"enabled" json:"enabled"`
+	EnforceFloorRate  int  `mapstructure:"enforce_floors_rate" json:"enforce_floors_rate"`
+	BidAdjustment     bool `mapstructure:"adjust_for_bid_adjustment" json:"adjust_for_bid_adjustment"`
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: Renamed variable name to `AdjustForBidAdjustment`

config/config.go
```diff
@@ -135,6 +140,11 @@ func (cfg *Configuration) validate(v *viper.Viper) []error {
 	if cfg.AccountDefaults.Events.Enabled {
 		glog.Warning(`account_defaults.events will currently not do anything as the feature is still under development. Please follow https://github.com/prebid/prebid-server/issues/1725 for more updates`)
 	}
+
+	if cfg.PriceFloors.Enabled {
+		glog.Warning(`cfg.PriceFloors.Enabled  will currently not do anything as price floors feature is still under development.`)
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: removed extra space

config/config.go
```diff
@@ -102,6 +102,11 @@ type Configuration struct {
 	// Hooks provides a way to specify hook execution plan for specific endpoints and stages
 	Hooks       Hooks       `mapstructure:"hooks"`
 	Validations Validations `mapstructure:"validations"`
+	PriceFloors PriceFloors `mapstructure:"price_floors"`
+}
+
+type PriceFloors struct {
+	Enabled bool `mapstructure:"enabled"`
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: This was added to control floors feature at host level

config/config.go
pm-jaydeep-mohite: updated

pm-jaydeep-mohite: added test cases

config/accounts.go
```diff
@@ -52,6 +53,14 @@ type AccountCCPA struct {
...
+	EnforceDealFloors bool `mapstructure:"enforce_deal_floors" json:"enforce_deal_floors"`
+	UseDynamicData    bool `mapstructure:"use_dynamic_data" json:"use_dynamic_data"`
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: Since most of the floors related config parameters are specified with default values in PRD, those existed parameters would be overwritten with JSON patch operation.

config/config.go
```diff
@@ -157,6 +158,14 @@ type AuctionTimeouts struct {
...
 }
 
+func (pf *AccountPriceFloors) validate(errs []error) []error {
+
+	if pf.EnforceFloorRate < 0 || pf.EnforceFloorRate > 100 {
+		errs = append(errs, fmt.Errorf(`account_defaults.price_floors.enforce_floors_rate should be between 0 and 100`))
+	}
+	return errs
+}
+
```
bsardo, line 168: Actually can we move this function into `config/accounts.go` right after the `AccountPriceFloors` struct definition so we keep the receiver methods close their data?

config/config_test.go
```diff
@@ -431,6 +442,15 @@ experiment:
             signing_timeout_ms: 10
 hooks:
     enabled: true
+price_floors:
+    enabled: true
+account_defaults:
+    price_floors:
+        enabled: true
+        enforce_floors_rate: 100
+        adjust_for_bid_adjustment: true
```
bsardo, line 451: Can we set these to something other than the default values, say 0 and false, since the defaults are tested elsewhere?

config/config_test.go
```diff
@@ -3150,3 +3178,44 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
+			},
+			want: []error{errors.New("account_defaults.price_floors.enforce_floors_rate should be between 0 and 100")},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			if got := tt.pf.validate(tt.args.errs); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("AccountPriceFloors.validate() = %v, want %v", got, tt.want)
+			}
```
bsardo, line 3218: You can simplify these lines by making use of the `github.com/stretchr/testify/assert` package which uses `reflect.DeepEqual` under the hood:
```
got := tt.pf.validate(tt.args.errs)
assert.ElementsMatch(t, got, tt.want)
```

config/config_test.go
```diff
@@ -3150,3 +3178,44 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
+		EnforceFloorRate       int
+		AdjustForBidAdjustment bool
+		EnforceDealFloors      bool
+		UseDynamicData         bool
+	}
+	type args struct {
+		errs []error
+	}
+	tests := []struct {
+		name string
```
bsardo, line 3194: Super nitpick: can you call this field `description` instead to match recent tests?

config/config_test.go
```diff
@@ -3150,3 +3178,44 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
 }
+
+func TestAccountPriceFloors_validate(t *testing.T) {
+	type fields struct {
+		Enabled                bool
+		EnforceFloorRate       int
+		AdjustForBidAdjustment bool
+		EnforceDealFloors      bool
+		UseDynamicData         bool
+	}
```
bsardo, line 3189: Can you delete the `fields` struct? It looks like you're planning to just specify the floors config in each test case as `pf` so I don't think you need this.

config/config_test.go
```diff
@@ -3150,3 +3178,44 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
+	type args struct {
+		errs []error
+	}
```
bsardo, line 3192: If we don't envision expanding the args list at this time, I suggest simplifying this by deleting this type and changing `args args` and `want []error` to something like
```
giveErrs []error
wantErrs []error
```

config/config.go
pm-jaydeep-mohite: moved into `config/account.go`

config/config_test.go
pm-jaydeep-mohite: by default floors feature is disabled and in `TestFullConfig`  test feature is enabled.

config/config_test.go
pm-jaydeep-mohite: used `assert` package

config/config_test.go
pm-jaydeep-mohite: updated

config/config_test.go
pm-jaydeep-mohite: removed `fields` struct?

config/config_test.go
pm-jaydeep-mohite: updated

config/config_test.go
```diff
@@ -3150,3 +3178,44 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
...
+				EnforceFloorRate: 100,
+			},
+		},
+		{
+			name: "Invalid configuration",
+			pf: &AccountPriceFloors{
+				EnforceFloorRate: 110,
+			},
+			want: []error{errors.New("account_defaults.price_floors.enforce_floors_rate should be between 0 and 100")},
+		},
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: added

config/config_test.go
```diff
@@ -3150,3 +3178,44 @@ func TestTCF2FeatureOneVendorException(t *testing.T) {
 		assert.Equal(t, tt.wantIsVendorException, value, tt.description)
 	}
 }
+
+func TestAccountPriceFloors_validate(t *testing.T) {
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: updated to `TestAccountPriceFloorsValidate`

config/config_test.go
bsardo: Ah my suggestion here is that we should set fields in `fullConfig` to something other than their default values, which you're doing for all of the floors fields except `enforce_floors_rate` and `adjust_for_bid_adjustment`. The default values for those two fields are 100 and true respectively and we test that the defaults are correctly applied in the config test `TestDefaults`. This test data structure `fullConfig` may be used to verify that values other than the default take precedence over the default values which is the reason for my suggestion.

config/accounts_test.go
```diff
@@ -824,3 +824,40 @@ func TestModulesGetConfig(t *testing.T) {
...
 	}
 }
+
+func TestAccountPriceFloorsValidate(t *testing.T) {
+
+	tests := []struct {
+		description string
+		pf          *AccountPriceFloors
+		want        []error
+		giveErrs    []error
```
AlexBVolcy, line 834: Nitpick: Is this field needed? I see it's not altered by any test case, and you could maybe just hardcode an error array on line 859 when calling `validate()`.

AlexBVolcy (commented):
Looks good, only two minor comments

config/config_test.go
pm-jaydeep-mohite: Updated test case to test  `enforce_floors_rate` and `adjust_for_bid_adjustment` for values other than default values

config/accounts.go
```diff
@@ -52,6 +53,22 @@ type AccountCCPA struct {
 	ChannelEnabled     AccountChannel `mapstructure:"channel_enabled" json:"channel_enabled"`
 }
 
+type AccountPriceFloors struct {
+	Enabled                bool `mapstructure:"enabled" json:"enabled"`
+	EnforceFloorRate       int  `mapstructure:"enforce_floors_rate" json:"enforce_floors_rate"`
```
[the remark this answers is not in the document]
pm-jaydeep-mohite: updated

bsardo (approved):

config/accounts_test.go
pm-jaydeep-mohite: Done

AlexBVolcy (approved):

PR #2572, opened 2023-02-08
[DO NOT MERGE] Trivy test with go mod changes

VeronikaSolovei9:


bsardo:
Closing. Verified Trivy check is now conditionally run.

PR #2577, opened 2023-02-10
OpenRTB 2.6: Remove 2.6-x Fields

SyntaxNode:
Tried to avoid needing to clear out OpenRTB 2.6-x when downgrading to 2.5, but we are now aware of some bidders which fail on the new fields.

### DIFF
--- a/openrtb_ext/convert_down.go
+++ b/openrtb_ext/convert_down.go
@@ -31,8 +31,8 @@ func ConvertDownTo25(r *RequestWrapper) error {
 		}
 	}
 
-	// Remove new OpenRTB 2.6 fields. The spec did not specify that bidders and exchanges
-	// must tolerate new or unexpected fields gracefully until 2.6.
+	// Remove fields introduced in OpenRTB 2.6+. The previous OpenRTB 2.5 spec did not specify that
+	// bidders must tolerate new or unexpected fields.
 	clear26Fields(r)
 	clear202211Fields(r)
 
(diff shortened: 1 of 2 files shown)


openrtb_ext/convert_down.go
```diff
@@ -177,3 +178,122 @@ func moveRewardedFrom26ToPrebidExt(i *ImpWrapper) error {
...
 	return nil
 }
+
+// clear26Fields sets all fields introduced in OpenRTB 2.6 to default values, which
+// will cause them to omitted during json marshal.
+func clear26Fields(r *RequestWrapper) {
+	r.WLangB = nil
+	r.CatTax = 0
+
+	if app := r.App; app != nil {
```
hhhjort, line 188: Is there a reason for this construction (`app := r.App`), rather than just checking if `r.App` is nil, and referencing `r.App` rather than `app`? Is it just to save the two characters `r.`?

openrtb_ext/convert_down.go
```diff
@@ -177,3 +178,122 @@ func moveRewardedFrom26ToPrebidExt(i *ImpWrapper) error {
...
+			}
+		}
+
+		if publisher := r.Site.Publisher; publisher != nil {
+			publisher.CatTax = 0
+		}
+	}
+
+	if device := r.Device; device != nil {
+		device.UA = ""
```
hhhjort, line 231: `device.ua` is a recommended OpenRTB 2.5 field.

openrtb_ext/convert_down.go
```diff
@@ -177,3 +178,122 @@ func moveRewardedFrom26ToPrebidExt(i *ImpWrapper) error {
...
+			audio.PodDur = 0
+			audio.RqdDurs = nil
+			audio.PodID = 0
+			audio.PodSeq = 0
+			audio.SlotInPod = 0
+			audio.MinCPMPerSec = 0
+		}
+
+		if video := imp.Video; video != nil {
+			video.StartDelay = nil
```
hhhjort, line 265: `r.imp[].video.startdelay` is a recommended OpenRTB 2.5 field

openrtb_ext/convert_down.go
SyntaxNode: It's to save on characters, and likely performance. while `r.App` isn't a big deal, `r.App.Content.Producer` starts to become a bit silly.

openrtb_ext/convert_down.go
SyntaxNode: Oops. Thanks for verifying the fields.

openrtb_ext/convert_down.go
SyntaxNode: Oops. Thanks for verifying the fields.

hhhjort (approved):
LGTM

openrtb_ext/convert_down_test.go
```diff
@@ -33,7 +34,30 @@ func TestConvertDownTo25(t *testing.T) {
 			},
 		},
 		{
-			description: "2.6 -> 2.5 + Other Ext Fields",
+			name: "2.6-dropped", // integration with clear26Fields
```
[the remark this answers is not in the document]
SyntaxNode: > It looks like you're just changing it to name to match the name of the Run function parameter right? 

Yes. That's exactly right.

> And is there a specific reason why you're removing spaces from the name?

Spaces are not allowed in the test names and are replaced with an underscore. I'd like to ensure the name here matches what t.Run uses, so you can use tooling to just run this specific test case if you'd like.

bsardo (approved):

PR #2582, opened 2023-02-14
Taboola Bid Adapter: Support page type and position, Always support event trackers

thetalkeren:


adapters/taboola/taboola.go
```diff
@@ -236,12 +228,30 @@ func createTaboolaRequests(request *openrtb2.BidRequest) (taboolaRequests []*ope
...
 
+func makeRequestExt(pageType string, errs []error) json.RawMessage {
+	requestExt := &RequestExt{
+		PageType: pageType,
+	}
+
+	requestExtJson, err := json.Marshal(requestExt)
+	if err != nil {
+		errs = append(errs, err)
+		return make([]byte, 0)
```
gargcreation1992, line 249: Would you like to continue making the taboola request if json marshalling fails here for any reason? If the intention is to not make the taboola request if makeRequestExt errors out then I would suggest to have this function return an error and handle that returned error in the caller(Line 232).

openrtb_ext/imp_taboola.go
```diff
@@ -7,4 +7,6 @@ type ImpExtTaboola struct {
 	TagId           string   `json:"tagid"`
 	BCat            []string `json:"bcat"`
 	BAdv            []string `json:"badv"`
+	PageType        string   `json:"pageType"`
+	Position        *int8    `json:"position"`
```
gargcreation1992, line 11: Is there any particular reason for using 8 bit integer here instead of using an int? If we use 8 bit integer and if the value of position exceeds 8 bits (i.e 127), integer overflow will happen and that would lead to unexpected results.

gargcreation1992 (commented):
left couple of comments but lgtm otherwise.

adapters/taboola/taboola.go
thetalkeren: Thanks for your comment, pushed a fix

openrtb_ext/imp_taboola.go
thetalkeren: Thanks for your comment, pushed a fix

adapters/taboola/taboola.go
```diff
@@ -238,17 +243,17 @@ func createTaboolaRequests(request *openrtb2.BidRequest) (taboolaRequests []*ope
...
 	requestExt := &RequestExt{
 		PageType: pageType,
 	}
 
 	requestExtJson, err := json.Marshal(requestExt)
 	if err != nil {
-		errs = append(errs, err)
-		return make([]byte, 0)
+		fmt.Errorf("could not marshal %s", requestExt)
+		return nil, err
```
gargcreation1992, line 254: Optional suggestion - this is correct but it could be simply be one liner - 
```
return nil, fmt.Errorf("could not marshal %s", requestExt)
```

gargcreation1992 (commented):
let me know if you wanna fix optional suggestion or else I will approve the PR.

adapters/taboola/taboola.go
thetalkeren: Thanks, please approve

gargcreation1992 (approved):

static/bidder-params/taboola.json
```diff
@@ -28,6 +28,12 @@
       "items": {
         "type": "string"
       }
+    },
+    "pageType": {
+      "type": "string"
+    },
+    "position": {
+      "type": "integer"
```
onkarvhanumante, line 36: create docs PR to include description for these new bidder params in [prebid.github.io](https://github.com/prebid/prebid.github.io) repo

onkarvhanumante (commented):
@thetalkeren left one comment for updating [taboola.md](https://github.com/prebid/prebid.github.io/blob/8f9ef40d8e53204004ff7e327460a0d6a67b2d22/dev-docs/bidders/taboola.md)

static/bidder-params/taboola.json
ahmadlob: @onkarvhanumante  we already did it (because at first we added these param in the prebid.js adapter )
https://docs.prebid.org/dev-docs/bidders/taboola.html

was merged from this closed PR:
https://github.com/prebid/prebid.github.io/pull/4235/files

onkarvhanumante: Thanks @ahmadlob  for clarification

onkarvhanumante (approved):

PR #2592, opened 2023-02-16
Openx: Pass through imp.ext wholesale except prebid, bidder

### LINKED ISSUE #2604: Port PR from PBS-Java: Openx: Pass through imp.ext wholesale except prebid, bidder
A PR was merged over on PBS-Java

- [https://github.com/prebid/prebid-server-java/pull/2180](https://github.com/prebid/prebid-server-java/pull/2180)
- timestamp: 2023-02-20T12:25:53Z

laurb9:
OpenX adapter: pass all of incoming `imp.ext` through to endpoint, except for `prebid` and `bidder` fields.

This should not impact existing functionality (`customParams` are still merged into `imp.ext`) but allows reading `gpid` and a host of other `data` of interest server-side while keeping the adapter code simple.

Resolves #2604 

PBS-Java implementation is at https://github.com/prebid/prebid-server-java/pull/2180

### DIFF
--- a/adapters/openx/openx.go
+++ b/adapters/openx/openx.go
@@ -144,8 +144,8 @@ func preprocess(imp *openrtb2.Imp, reqExt *openxReqExt) error {
 			Message: err.Error(),
 		}
 	}
-	delete(impExt, "bidder")
-	delete(impExt, "prebid")
+	delete(impExt, openrtb_ext.PrebidExtKey)
+	delete(impExt, openrtb_ext.PrebidExtBidderKey)
 
 	if openxExt.CustomParams != nil {
 		var err error
--- a/adapters/openx/openx.go
+++ b/adapters/openx/openx.go
@@ -19,10 +19,7 @@ type OpenxAdapter struct {
 	endpoint   string
 }
 
-type openxImpExt struct {
-	CustomParams       map[string]interface{}             `json:"customParams,omitempty"`
-	AuctionEnvironment openrtb_ext.AuctionEnvironmentType `json:"ae,omitempty"`
-}
+type openxImpExt map[string]json.RawMessage
 
 type openxReqExt struct {
 	DelDomain    string `json:"delDomain,omitempty"`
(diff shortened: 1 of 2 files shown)


adapters/openx/openx.go
```diff
@@ -140,18 +137,26 @@ func preprocess(imp *openrtb2.Imp, reqExt *openxReqExt) error {
...
+	// outgoing imp.ext should be same as incoming imp.ext minus prebid and bidder
 	impExt := openxImpExt{}
-	addImpExt := false
+	if err := json.Unmarshal(imp.Ext, &impExt); err != nil {
+		return &errortypes.BadInput{
+			Message: err.Error(),
+		}
+	}
+	delete(impExt, "bidder")
+	delete(impExt, "prebid")
```
gargcreation1992, line 148: This seems to be correct but there is one problem. If golang's built in delete function doesn't find "bidder" in impExt map then it will create a panic which we do not want. IMO, it would be best to first check if key exists in the map before deleting it. 
Also we should use constant already defined `openrtb_ext` package for "prebid" and "bidder", it should be PrebidExtKey and PrebidExtBidderKey respectively. 

Here is how - 
```
if _, ok := impExt[openrtb_ext.PrebidExtBidderKey]; ok {
        delete(impExt, openrtb_ext.PrebidExtBidderKey)
}

if _, ok := impExt[openrtb_ext.PrebidExtKey]; ok {
        delete(impExt, openrtb_ext.PrebidExtKey)
}
```

adapters/openx/openx.go
```diff
@@ -140,18 +137,26 @@ func preprocess(imp *openrtb2.Imp, reqExt *openxReqExt) error {
...
-	}
-	if bidderExt.AuctionEnvironment != openrtb_ext.ServerSideAuction {
-		impExt.AuctionEnvironment = bidderExt.AuctionEnvironment
-		addImpExt = true
+		var err error
+		if impExt["customParams"], err = json.Marshal(openxExt.CustomParams); err != nil {
+			return &errortypes.BadInput{
+				Message: err.Error(),
+			}
+		}
```
gargcreation1992, line 156: This is correct but we do not have to create error var upfront. Here is how you could do this - 
```
if impExt["customParams"], err := json.Marshal(openxExt.CustomParams); err != nil {
			return &errortypes.BadInput{
				Message: err.Error(),
			}
		}
```

gargcreation1992 (changes requested):
requesting few changes.

adapters/openx/openx.go
laurb9: > This seems to be correct but there is one problem. If golang's built in delete function doesn't find "bidder" in impExt map then it will create a panic which we do not want. IMO, it would be best to first check if key exists in the map before deleting it. Also we should use constant already defined `openrtb_ext` package for "prebid" and "bidder", it should be PrebidExtKey and PrebidExtBidderKey respectively.

Thank you for the suggestions, I will use the existing constants. Regarding the [built-in delete](https://pkg.go.dev/builtin#delete), it is now safe - turns out its behavior changed at some point to no longer need a guard and the linter will complain if I add it.

adapters/openx/openx.go
laurb9: The implicit var creation only works with variables; I would get this error:
> non-name impExt["customParams"] on left side of :=

adapters/openx/openx.go
gargcreation1992: thank you for checking that. I read it online that delete is not safe to be used on keys without checking it.

gargcreation1992 (approved):
g2g!

Sonali-More-Xandr (approved):
LGTM!

Issue #1167, opened 2020-01-17
Mirror amp macros in RTC call back in response


patmmccann:
Several values are available as AMP macros that it is not currently possible to use as set targeting values. 

PAGEVIEWID - pageViewId
HREF - equivalent to window.context.location.href
ATTR(height) - Height attribute of the amp-ad element
ATTR(width) - Width attribute of the amp-ad element
ATTR(data-slot) - data-slot attribute of the amp-ad element
ATTR(data-multi-size) - data-multi-size attribute of the amp-ad element
ATTR(data-multi-size-validation) - data-multi-size-validation attribute of the amp-ad element
ATTR(data-override-width) - data-override-width attribute of the amp-ad element
ATTR(data-override-height) - data-override-height attribute of the amp-ad element
ATTR(data-json) - data-json attribute of the amp-ad element
ADCID - adClientId
TGT - Just the targeting piece of data-json
CANONICAL_URL - The canonical URL of the page.
TIMEOUT - The publisher-specified timeout for the RTC callout.

A possible strategy to set targeting to gam with one of these values is to pass one of them to prebid server and have prebid server include it in the response. 

What we would like to do is pass a pageviewid to gam so that we can associate impressions from the same pageview together on our gam log.

bretg:
We discussed this in the Prebid Server meeting today. The suggested approach is:

1) PBS parses the AMP query string variables and puts them all in ext.prebid.amp.data

2) The AMP Stored Request (or anything else that creates PBS openRTB!) could specify which adserver targeting variables to add and their values in a generic way. Proposed:

```
   ext.prebid.adservertargeting:  [{
              "key": "hb_ampparam",
              "source": "bidrequest",    // causes PBS to treat 'value' as a path to pull from the request object
              "value": "ext.prebid.amp.data.ampkey"
            },{
               "key": "hb_static_thing",
               "source": "static",  // causes PBS to just use the 'value' provided
               "value": "my-static-value"
            },{
               "key": "{{BIDDER}}_custom1", // {{BIDDER}} is a macro to be resolved
               "source": "bidresponse",  // causes PBS to treat 'value' as a path to pull from the bidder's response object, specifically seatbid[j].bid[k]
               "value": "seatbid.bid.ext.custom1"
            }
   }]
```
3) While generating the response seatbid[].bid[].ext.prebid.targeting, Prebid Server would parse the request ext.prebid.adservertargeting and follow these rules:

[...]

PR #2593, opened 2023-02-16
Adserver targeting

VeronikaSolovei9:
Implementation of https://github.com/prebid/prebid-server/issues/1167


util/httputil/httputil.go
```diff
@@ -75,3 +76,16 @@ func findRemoteAddr(r *http.Request, v iputil.IPValidator) (net.IP, iputil.IPVer
...
+	rawQuery := url.RawQuery
+	rawQuerySplit := strings.Split(rawQuery, "&")
+	ampData := make(map[string]string, 0)
+	for _, kv := range rawQuerySplit {
+		kvSplit := strings.Split(kv, "=")
+		if len(kvSplit) == 2 {
+			ampData[kvSplit[0]] = kvSplit[1]
+		}
+	}
+	return ampData
```
SyntaxNode, line 90: Please do not reimplement query parsing. It's available in [url.Query()](https://pkg.go.dev/net/url#example-URL.Query)

VeronikaSolovei9: Agree! I didn't know about this function, modified!

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,180 @@
...
+
+func resolveKey(respTargetingData ResponseTargetingData, bidderName string) string {
+	key := respTargetingData.Key
+	if respTargetingData.HasMacro {
+		key = strings.Replace(respTargetingData.Key, bidderMacro, bidderName, -1)
+	}
+	return key
+}
+
+func truncateTargetingKeys(targetingData map[string]string, truncateTargetAttribute *int) map[string]string {
```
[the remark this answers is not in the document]
VeronikaSolovei9: I tried to reuse existing functionality from targeting.go. It doesn't do exactly what I need and also I got imports circular dependencies(exchange package) error. 
I reused `TruncateKey` function only

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,180 @@
...
+			case "ext":
+				getValueFromSeatBidExt(key, path, targetingData, seatExt, warnings)
+			}
+
+		case "ext":
+			getValueFromRespExt(key, path, targetingData, response.Ext, warnings)
+		default:
+			getValueFromResp(key, path, targetingData, bidderRespHolder, response, warnings)
+
+		}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Thank you for your initial comments, I like it so much more after refactoring

adservertargeting/respcache.go
```diff
@@ -0,0 +1,33 @@
...
+
+type respCache struct {
+	// only holds upper level bid response data except of ext and bids
+	bidRespData map[string]string
+}
+
+func (brh *respCache) GetRespData(bidderResp *openrtb2.BidResponse, field string) (string, error) {
+	if len(brh.bidRespData) == 0 {
+		// this code should be modified if there are changes in Op[enRtb format
+		// it's up to date with OpenRTB 2.5 and 2.6
```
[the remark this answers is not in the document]
VeronikaSolovei9: I left a comment for now. Please reach me out to discuss how to create this test, it would be good to have it.

adservertargeting/respcache.go
```diff
@@ -0,0 +1,33 @@
...
+		brh.bidRespData = make(map[string]string)
+		brh.bidRespData["id"] = bidderResp.ID
+		brh.bidRespData["bidid"] = bidderResp.BidID
+		brh.bidRespData["cur"] = bidderResp.Cur
+		brh.bidRespData["customdata"] = bidderResp.CustomData
+		brh.bidRespData["nbr"] = fmt.Sprint(bidderResp.NBR.Val())
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good idea! In this case the whole respcache.go is not needed at all!

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,139 @@
...
+	reqExt, err := reqWrapper.GetRequestExt()
+	if err != nil {
+		warnings = append(warnings, createWarning(err.Error()))
+		return nil, warnings
+	}
+	reqExtPrebid := reqExt.GetPrebid()
+	if reqExtPrebid == nil {
+		return nil, warnings
+	}
+	adServerTargeting := reqExtPrebid.AdServerTargeting
```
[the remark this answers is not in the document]
VeronikaSolovei9: Done!

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,139 @@
...
+	adServerTargeting, warnings := ExtractAdServerTargeting(reqWrapper, resolvedRequest, queryParams)
+	response, warnings = ResolveAdServerTargeting(adServerTargeting, response, warnings, truncateTargetAttribute)
+
+	if len(warnings) > 0 {
+		bidResponseExt.Warnings[openrtb_ext.BidderReservedGeneral] = append(bidResponseExt.Warnings[openrtb_ext.BidderReservedGeneral], warnings...)
+	}
+	return response
+}
+
+func ExtractAdServerTargeting(
```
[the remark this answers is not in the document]
VeronikaSolovei9: This function collects targeting attributes for later processing. In main flow it is executed before final resolving:
func ProcessAdServerTargeting:
```
adServerTargeting, warnings := ExtractAdServerTargeting(reqWrapper, resolvedRequest, queryParams)
response, warnings = ResolveAdServerTargeting(adServerTargeting, response, warnings, truncateTargetAttribute)
```
Is `CollectAdServerTargeting` a good name for it?

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,139 @@
...
+	reqWrapper *openrtb_ext.RequestWrapper, resolvedRequest json.RawMessage,
+	ampData map[string]string) (*adServerTargetingData, []openrtb_ext.ExtBidderMessage) {
+	//this func should receive a finalized request wrapper
+
+	warnings := make([]openrtb_ext.ExtBidderMessage, 0)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good point! Changed to `var warnings []openrtb_ext.ExtBidderMessage`

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,139 @@
...
+
+const (
+	bidderMacro   = "{{BIDDER}}"
+	pathDelimiter = "."
+)
+
+// RequestTargetingData struct to hold pre-processed ad server targeting keys and values
+type RequestTargetingData struct {
+	SingleVal json.RawMessage
+	ImpData   map[string][]byte
```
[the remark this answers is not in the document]
VeronikaSolovei9: Agree, this name is not informative.
Renamed to `TargetingValueByImpId`. This map holds targeting values extracted from different imps, meaning a targeting key like "hb_imp_video_maxduration" may have different value for different imps.
```
 "hb_imp_video_maxduration" ->  "imp1" -> 15 sec
                                "imp2" -> 10 sec
                                "imp3" -> 30 sec
```

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/prebid/prebid-server/openrtb_ext"
+	"strings"
+)
+
+type DataSource string
+
+const (
+	BidRequest  DataSource = "bidrequest"
+	Static      DataSource = "static"
+	BidResponse DataSource = "bidresponse"
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, renamed

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,180 @@
...
+func getValueFromResp(
+	key, path string,
+	targetingData map[string]string,
+	bidderRespHolder respCache,
+	response *openrtb2.BidResponse,
+	warnings []openrtb_ext.ExtBidderMessage) {
+	// path points to value in response, not seat or ext
+	// bidder response has limited number of properties
+	// instead of unmarshal the whole response with seats and bids
+	// try to find data by field name
```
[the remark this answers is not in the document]
VeronikaSolovei9: Agree, moved

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, warnings
+	}
+	adServerTargeting := reqExtPrebid.AdServerTargeting
+
+	if len(adServerTargeting) == 0 {
+		return nil, warnings
+	}
+
+	requestTargetingData := make(map[string]RequestTargetingData, 0)
+	responseTargetingData := make([]ResponseTargetingData, 0)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Replaced

adservertargeting/reqdataprocessor.go
```diff
@@ -0,0 +1,90 @@
...
+		res.ImpData = impData
+		return res, err
+	}
+
+	// get data by key from request
+	requestValue, err := getDataFromRequest(path, dataHolder)
+	if requestValue != nil || err != nil {
+		res.SingleVal = requestValue
+		return res, err
+	}
```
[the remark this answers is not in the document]
VeronikaSolovei9: Right, value can be found in one of: query params, imps or request. 
My idea here is to hide string splits inside of `getValueFromQueryParam`, `getValueFromImp`, and `getDataFromRequest`. If value is found in one of them - func getValueFromBidRequest doesn't look for value anywhere else and returns it.  
`getDataFromRequest` should be executed at the end, after everything else, because it's not clear how it should start. 

My initial version was parsing strings and it was overloaded by if cases. I don't think this approach will make it cleaner. 
Unless I miss something.

adservertargeting/respdataprocessor.go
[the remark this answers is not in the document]
SyntaxNode: Not what I was thinking, but looks good. I meant to suggest something like this:

```
// getValueFromResp optimizes retrieval of paths for a bid response (not a seat or ext)
func getValueFromResp(
```

adservertargeting/respdataprocessor.go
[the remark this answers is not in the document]
SyntaxNode: Not really what I was thinking, but it could work too. I meant to suggest something like this:

```
// getValueFromResp optimizes retrieval of paths for a bid response (not a seat or ext)
func getValueFromResp(
```

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+	adServerTargeting, warnings := CollectAdServerTargeting(reqWrapper, resolvedRequest, queryParams)
+	response, warnings = ResolveAdServerTargeting(adServerTargeting, response, warnings, truncateTargetAttribute)
+
+	if len(warnings) > 0 {
+		bidResponseExt.Warnings[openrtb_ext.BidderReservedGeneral] = append(bidResponseExt.Warnings[openrtb_ext.BidderReservedGeneral], warnings...)
+	}
+	return response
+}
+
+func CollectAdServerTargeting(
```
SyntaxNode, line 58: Please add a comment to help the reader understand what this function is doing. Even if you change it to package scope.

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+	reqWrapper *openrtb_ext.RequestWrapper, resolvedRequest json.RawMessage,
+	queryParams url.Values) (*adServerTargetingData, []openrtb_ext.ExtBidderMessage) {
+
+	var warnings []openrtb_ext.ExtBidderMessage
+
+	adServerTargeting, err := getAdServerTargeting(reqWrapper)
+	if err != nil {
+		warnings = append(warnings, createWarning("Unable to extract adServerTargetingFrom req"))
```
SyntaxNode, line 66: Error strings in Go should not be capitalized. Please start with an lower case character. Consider using "request" instead of "req" for improved clarity.

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, warnings
+	}
+	if len(adServerTargeting) == 0 {
```
SyntaxNode, line 69: Why is this check necessary? If there's an error, it's handled by the above code. If `adServerTargeting` is empty, the for loop with gracefully exit. Is it important for a nil value to be returned instead of an empty `adServerTargetingData` struct? If that doesn't matter, you won't need to return a pointer.

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, warnings
+	}
+
+	requestTargetingData := map[string]RequestTargetingData{}
+	responseTargetingData := []ResponseTargetingData{}
+
+	dataHolder := reqImpCache{resolverReq: resolvedRequest}
```
SyntaxNode, line 76: Consider a different name than `dataHolder`.  What data is it holding? Perhaps this should be `valueCache` or some such?

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+		source := strings.ToLower(targetingObj.Source)
+		switch DataSource(source) {
+		case SourceBidRequest:
+			//causes PBS to treat 'value' as a path to pull from the request object
+			value, err := getValueFromBidRequest(&dataHolder, targetingObj.Value, queryParams)
+			if err != nil {
+				warnings = append(warnings, createWarning(err.Error()))
+				continue
+			}
+			requestTargetingData[targetingObj.Key] = value
```
SyntaxNode, line 88: Consider refactoring to remove the need for the `continue` keyword. Perhaps something like:

```
value, err := getValueFromBidRequest(&dataHolder, targetingObj.Value, queryParams)
if err != nil {
	warnings = append(warnings, createWarning(err.Error()))
} else {
	requestTargetingData[targetingObj.Key] = value
}
```

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	adServerTargetingData := &adServerTargetingData{
+		RequestTargetingData:  requestTargetingData,
+		ResponseTargetingData: responseTargetingData,
+	}
+
+	return adServerTargetingData, warnings
+}
```
SyntaxNode, line 109: I left nitpick comments, but overall I feel this function is well coded and easy to understand.

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// ad server targeting data will go to seatbid[].bid[].ext.prebid.targeting
+	//TODO: truncate keys
+
+	bidsHolder := bidsCache{bids: make(map[string]map[string][]byte)}
+
+	for _, seat := range response.SeatBid {
+		bidderName := seat.Seat
+		for i, bid := range seat.Bid {
+			targetingData := make(map[string]string, 0)
```
SyntaxNode, line 128: Consider replacing with an object initializer or omit the initial size of 0.

adservertargeting/reqcache.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		dh.impsData = impsData
+	}
+	return dh.impsData, nil
+}
+
+type bidsCache struct {
+	//bidder name to bid id to bid data
+	// bidder name is another layer to avoid collisions in case bid ids from different bidders will be the same
+	bids map[string]map[string][]byte
```
SyntaxNode, line 38: I'm a little confused by the keys. Is it map[bid id]map[bidder][byte[]?

adservertargeting/reqdataprocessor.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return res, nil
+}
+
+func getValueFromQueryParam(path string, queryParams url.Values) (json.RawMessage, error) {
+	ampDataSplit := strings.Split(path, "ext.prebid.amp.data.")
```
SyntaxNode, line 53: Using split in this manner is a bit confusing to read. Consider extracting this to a method like:

```
func verifyPrefixAndTrim(path, prefix string) (string, bool) {
...
}
```

then this call site looks like

```
pathTrimmed, hasPrefix := verifyPrefixAndTrim(path, "ext.prebid.amp.data.")
if !hasPrefix {
  return nil, nil
}

if val := queryParams.Get(pathTrimmed); val != "" {
  return json.RawMessage(val), nil
}

return nil, errors.Errorf("value not found for path: %s", path)
```

Do you think that's easier to follow with the details of the split length hidden in the helper func?

adservertargeting/reqdataprocessor.go
```diff
@@ -0,0 +1,104 @@
+package adservertargeting
```
SyntaxNode, line 1: The term "processor" is not descriptive of the functionality IMHO. How about something like "requestlookup"? .. or just a description which better translates to "looks up and returns a value".

adservertargeting/reqdataprocessor.go
```diff
@@ -0,0 +1,104 @@
...
+	reqExt, err := reqWrapper.GetRequestExt()
+	if err != nil {
+		return nil, err
+	}
+	reqExtPrebid := reqExt.GetPrebid()
+	if reqExtPrebid == nil {
+		return nil, err
+	}
+	return reqExtPrebid.AdServerTargeting, nil
+}
```
SyntaxNode, line 22: Consider adding line breaks between the error checking:

```
	reqExt, err := reqWrapper.GetRequestExt()
	if err != nil {
		return nil, err
	}

	reqExtPrebid := reqExt.GetPrebid()
	if reqExtPrebid == nil {
		return nil, err
	}

	return reqExtPrebid.AdServerTargeting, nil
```

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/prebid/openrtb/v17/openrtb2"
+	"github.com/prebid/prebid-server/openrtb_ext"
+	jsonpatch "gopkg.in/evanphx/json-patch.v4"
+	"strings"
+)
+
+const MaxKeyLength = 20
+
+func processRequestTargetingData(adServerTargetingData *adServerTargetingData, targetingData map[string]string, bidImpId string) {
+	if len(adServerTargetingData.RequestTargetingData) == 0 {
```
SyntaxNode, line 16: The length check isn't necessary. Keep in mind that Go's range loop gracefully handles nil or empty lists such that no optimization is necessary.

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,201 @@
...
+	targetingData map[string]string,
+	bidderName string,
+	bid openrtb2.Bid,
+	bidsHolder bidsCache,
+	response *openrtb2.BidResponse,
+	seatExt json.RawMessage,
+	warnings []openrtb_ext.ExtBidderMessage) {
+	if len(adServerTargetingData.ResponseTargetingData) == 0 {
+		return
+	}
```
SyntaxNode, line 39: Same deal here. No need for the length check.

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,201 @@
...
+
+	}
+	bidSplit := strings.Split(path, "seatbid.bid.")
+	value, err := splitAndGet(bidSplit[1], bidBytes, pathDelimiter)
+	if err != nil {
+		warnings = append(warnings, createWarning(err.Error()))
+
+	}
+	targetingData[key] = value
+}
```
SyntaxNode, line 139: IMHO it might be easier to read the method by moving the parameters to a single line. I've also noticed there is no return values which increases the number of arguments. How do you feel about:

```
func getValueFromSeatBidBid(path string, targetingData map[string]string, bidsHolder bidsCache, bidderName string, bid openrtb2.Bid) error {
	bidBytes, err := bidsHolder.GetBid(bidderName, bid.ID, bid)
	if err != nil {
		return createWarning(err.Error())

	}

	path = strings.TrimPrefix(path, "seatbid.bid.")
	value, err := splitAndGet(path, bidBytes, pathDelimiter)
	if err != nil {
		return createWarning(err.Error())

	}

	return value
}
```

where the call site now looks like:

```
if v, err := getValueFromSeatBidBid(path, targetingData, bidsHolder, bidderName, bid); err == nil {
	targetingData[key] = v
} else {
	warnings = append(warnings, err)
}
```

Such that the responsibility of setting the response and collecting errors is moved up a level. Same comment applies to the other getValue methods.

SyntaxNode (commented):
Excellent update! This is a lot easier to follow and appears to match the spec. I would like to see end to end tests to ensure this is hooked into the main auction flow correctly and to provide examples on the feature's usage.

The only feature from the spec I didn't see represented is:
> If the value points to a JSON path that is an object or array, there should be a warning in debug mode

Is that already accounted for? .. and if so, where?

adservertargeting/respdataprocessor.go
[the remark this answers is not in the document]
VeronikaSolovei9: Oh ok, modified

adservertargeting/adservertargeting.go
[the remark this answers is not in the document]
VeronikaSolovei9: Renamed to `collectTargetingValues `

exchange/exchange.go
```diff
@@ -425,6 +427,11 @@ func (e *exchange) HoldAuction(ctx context.Context, r AuctionRequest, debugLog *
 
 	// Build the response
 	bidResponse, err := e.buildBidResponse(ctx, liveAdapters, adapterBids, r.BidRequestWrapper.BidRequest, adapterExtra, auc, bidResponseExt, cacheInstructions.returnCreative, r.ImpExtInfoMap, r.PubID, errs)
+
+	bidResponse = adservertargeting.ProcessAdServerTargeting(r.BidRequestWrapper, r.ResolvedBidRequest, bidResponse, r.QueryParams, bidResponseExt, r.Account.TruncateTargetAttribute)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, renamed!

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,135 @@
...
+func ProcessAdServerTargeting(
+	reqWrapper *openrtb_ext.RequestWrapper,
+	resolvedRequest json.RawMessage,
+	response *openrtb2.BidResponse,
+	queryParams url.Values,
+	bidResponseExt *openrtb_ext.ExtBidResponse,
+	truncateTargetAttribute *int) *openrtb2.BidResponse {
+
+	adServerTargeting, warnings := CollectAdServerTargeting(reqWrapper, resolvedRequest, queryParams)
+	response, warnings = ResolveAdServerTargeting(adServerTargeting, response, warnings, truncateTargetAttribute)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes!

adservertargeting/adservertargeting.go
VeronikaSolovei9: Added

adservertargeting/adservertargeting.go
VeronikaSolovei9: Yes, modified.

adservertargeting/adservertargeting.go
VeronikaSolovei9: Good point! Yes, this check is not needed. And nil value for `adServerTargeting` doesn't matter. I removed the check and left `AdServerTargeting []AdServerTargeting` as array, not as a pointer.

adservertargeting/adservertargeting.go
VeronikaSolovei9: Renamed to `impsCache` and renamed similar variable for bids to `bidCache`

adservertargeting/adservertargeting.go
VeronikaSolovei9: Refactored

adservertargeting/adservertargeting.go
VeronikaSolovei9: Removed 0 size initialization.

adservertargeting/reqcache.go
VeronikaSolovei9: Replaced first comment to `// map[bidder name] map [bid id] bid data`

adservertargeting/reqdataprocessor.go
VeronikaSolovei9: Yes! Definitely looks better.

adservertargeting/reqdataprocessor.go
VeronikaSolovei9: Renamed

adservertargeting/respdataprocessor.go
VeronikaSolovei9: Right, removed.

adservertargeting/respdataprocessor.go
VeronikaSolovei9: This is a great point! How I didn't see it while coding (I wanted to hide all those appends and set to map statements)... It's even better, error check and value set is only required once for all "getValue" functions at the end of `processResponseTargetingData`

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,171 @@
...
+		return bidderResp.BidID, nil
+	case "cur":
+		return bidderResp.Cur, nil
+	case "customdata":
+		return bidderResp.CustomData, nil
+	case "nbr":
+		return fmt.Sprint(bidderResp.NBR.Val()), nil
+
+	default:
+		return "", errors.Errorf("key not found for field in bid response: %s", field)
```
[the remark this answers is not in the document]
VeronikaSolovei9: Added. And found and fixed related bug.

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,171 @@
...
+func getValueFromResp(path string, response *openrtb2.BidResponse) (string, error) {
+	value, err := getRespData(response, path)
+	if err != nil {
+		return "", err
+	}
+	return value, nil
+}
+
+func getRespData(bidderResp *openrtb2.BidResponse, field string) (string, error) {
+	// this code should be modified if there are changes in Op[enRtb format
```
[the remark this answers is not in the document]
VeronikaSolovei9: Fixed

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	return key
+}
+
+func truncateTargetingKeys(targetingData map[string]string, truncateTargetAttribute *int) map[string]string {
+	maxLength := MaxKeyLength
+	if truncateTargetAttribute != nil {
+		maxLength = *truncateTargetAttribute
+		if maxLength < 0 {
+			maxLength = MaxKeyLength
```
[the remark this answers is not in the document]
VeronikaSolovei9: Added

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,171 @@
...
+		path := respTargetingData.Path
+
+		var value string
+		var err error
+		pathSplit := strings.Split(path, pathDelimiter)
+
+		switch pathSplit[0] {
+
+		case "seatbid":
+			switch pathSplit[1] {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good point! Path split may have just 1 element, for instance "cur". It means the value should be taken from bid response itself, like "resp.cur".  However I added targeting keys path validation to "collect" function to make sure it's not empty.

adservertargeting/requestlookup.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	// get data by key from request
+	requestValue, err := getDataFromRequest(path, dataHolder)
+	if requestValue != nil || err != nil {
+		res.SingleVal = requestValue
+		return res, err
+	}
+
+	return res, nil
```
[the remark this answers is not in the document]
VeronikaSolovei9: I tried and it looks like no. This case is logically unreachable, this line exists to satisfy the compiler requirement to have a return statement. 
All "suspicious" cases will be handled and returned with a proper error message in the code statement above: `getDataFromRequest`

adservertargeting/reqcache.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (bdh *bidsCache) GetBid(bidderName, bidId string, bid openrtb2.Bid) ([]byte, error) {
+
+	_, seatBidExists := bdh.bids[bidderName]
+	if !seatBidExists {
+		impToBid := make(map[string][]byte, 0)
+		bdh.bids[bidderName] = impToBid
+	}
+	_, biddExists := bdh.bids[bidderName][bidId]
```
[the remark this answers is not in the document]
VeronikaSolovei9: Fixed

adservertargeting/reqcache.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/buger/jsonparser"
+	"github.com/prebid/openrtb/v17/openrtb2"
+)
+
+type reqImpCache struct {
+	resolverReq json.RawMessage
+	impsData    []json.RawMessage
+}
+
+func (dh *reqImpCache) GetReqJson() []byte {
```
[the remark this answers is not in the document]
VeronikaSolovei9: Renamed

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,138 @@
...
+	adServerTargetingData *adServerTargetingData,
+	response *openrtb2.BidResponse,
+	warnings []openrtb_ext.ExtBidderMessage,
+	truncateTargetAttribute *int) (*openrtb2.BidResponse, []openrtb_ext.ExtBidderMessage) {
+	if adServerTargetingData == nil {
+		return response, nil
+	}
+
+	// ad server targeting data will go to seatbid[].bid[].ext.prebid.targeting
+
```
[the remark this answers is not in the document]
VeronikaSolovei9: I removed the entire comment. I don't think it's needed.

adservertargeting/adservertargeting.go
```diff
@@ -0,0 +1,138 @@
...
+	return adServerTargetingData, warnings
+}
+
+func resolve(
+	adServerTargetingData *adServerTargetingData,
+	response *openrtb2.BidResponse,
+	warnings []openrtb_ext.ExtBidderMessage,
+	truncateTargetAttribute *int) (*openrtb2.BidResponse, []openrtb_ext.ExtBidderMessage) {
+	if adServerTargetingData == nil {
+		return response, nil
```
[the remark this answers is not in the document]
VeronikaSolovei9: Good point. I realized this will never be the case and removed the check.

VeronikaSolovei9:
Good comments, Alex! Please re-review

openrtb_ext/request.go
```diff
@@ -71,6 +71,15 @@ type ExtRequestPrebid struct {
 	// - basic: excludes debugmessages and analytic_tags from output
 	// any other value or an empty string disables trace output at all.
 	Trace string `json:"trace,omitempty"`
+
+	//
+	AdServerTargeting []AdServerTargeting `json:"adservertargeting,omitempty"`
```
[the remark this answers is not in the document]
hhhjort: I would suggest a slight renaming, to avoid thing = list of things. Either `AdServerTargeting []AdServerTarget`, `AdServerTargeting []AdServerTargetRule`, or `AdserverTargets []AdServerTargeting`

adservertargeting/requestlookup.go
```diff
@@ -0,0 +1,137 @@
...
+
+func getAdServerTargeting(reqWrapper *openrtb_ext.RequestWrapper) ([]openrtb_ext.AdServerTargeting, error) {
+	reqExt, err := reqWrapper.GetRequestExt()
+	if err != nil {
+		return nil, err
+	}
+
+	reqExtPrebid := reqExt.GetPrebid()
+	if reqExtPrebid == nil {
+		return nil, err
```
hhhjort, line 21: What error are you returning here? Is it possible to get to this line when `err != nil`?

adservertargeting/respdataprocessor.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return warnings
+}
+
+func buildBidExt(targetingData map[string]string,
+	bid openrtb2.Bid,
+	warnings []openrtb_ext.ExtBidderMessage,
+	truncateTargetAttribute *int) json.RawMessage {
+
+	targetingDataTruncated := truncateTargetingKeys(targetingData, truncateTargetAttribute)
```
hhhjort, line 74: Wouldn't it be better to truncate the keys back when we create them, rather than doing this loop through all the keys again?

endpoints/openrtb2/sample-requests/hooks/auction_bidder_reject.json
```diff
@@ -51,8 +51,7 @@
           {
             "id": "applogy-bid",
             "impid": "some-impression-id",
-            "price": 0,
-            "ext": {"prebid":{"type":"banner"}}
```
hhhjort, line 55: Why did this `ext` need to be removed?

VeronikaSolovei9: It appears bid.ext was not validated in unit tests before.
Actual bid.ext turns out to be 
```
{"origbidcpm":0,"origbidcur":"USD","prebid":{"meta":{"adaptercode":"applogy"},"type":"banner"}}
```
 not 
```
"ext": {"prebid":{"type":"banner"}}
```

After I added bid.ext comparison logic it stoped working. 
File auction_test.go -> func assertBidResponseEqual: 
```
if len(expectedBid.Ext) > 0 {
	assert.JSONEq(t, string(expectedBid.Ext), string(actualBidMap[bidID].Ext), "Incorrect bid extension")
}
```
I removed bid.ext from expected result, because it was not validated before and it was not valid anyways. 
I can add expected bid.ext back and modify them accordingly. So far it's filled with default values only.

adservertargeting/respdataprocessor.go
VeronikaSolovei9: My idea here was to separate "get values" logic from everything else, because "get logic" became complicated. 
Func `truncateTargetingKeys` will be needed to execute for "request" keys and for "response" keys. 
Also I'll need to pass `truncateTargetAttribute` parameter downstream to where we set keys and values to final map, I think it will overload logic and number of parameters.

I think having this logic here makes the code easier to understand. 
I'm open to discuss moving it to another place, if you feel otherwise.

adservertargeting/requestlookup.go
VeronikaSolovei9: Right, no error, refactored

adservertargeting/requestlookup.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// check if key points to imp data
+	impData, err := getValueFromImp(path, dataHolder)
+	if len(impData) > 0 || err != nil {
+		res.TargetingValueByImpId = impData
+		return res, err
+	}
+
+	// get data by key from request
+	requestValue, err := getDataFromRequest(path, dataHolder)
```
AlexBVolcy, line 79: Nitpick: the function names `getValueFromBidRequest` vs. `getDataFromRequest` are pretty similar and I'm wondering if there's a better name for one of them that'll help distinguish the two functions? If you feel fine about them then I'm happy to keep them as they are, but thought it was worth a mention.

AlexBVolcy (commented):
Looks good! Left one small comment.

adservertargeting/requestlookup.go
VeronikaSolovei9: Yes, understand. in this function we try to get a key that was not found in query param or in imps. this is a default case. Renamed to `getDataFromRequestJson`

openrtb_ext/request.go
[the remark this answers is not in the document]
VeronikaSolovei9: renamed to `AdServerTargeting []AdServerTarget`

adservertargeting/requestcache_test.go
```diff
@@ -0,0 +1,141 @@
...
+	 "incorrect":true
+   },
+  "site": {"page": "test.com"}
+}`
+
+	reqNoImps = `{
+  "id": "req_id",
+  "site": {"page": "test.com"}
+}`
+)
```
AlexBVolcy, line 141: Would it be too tedious to extract this consts to their own json files? I'm happy with whatever you think, but it might make tests a little cleaner.

adservertargeting/requestlookup_test.go
```diff
@@ -0,0 +1,404 @@
...
+			expectedError:     false,
+		},
+		{
+			description:       "valid request with correct ext.prebid, no ad server targeting",
+			inputRequestExt:   json.RawMessage(`{"prebid":{}}`),
+			expectedTargeting: nil,
+			expectedError:     false,
+		},
+		{
+			description:       "valid request with correct ext.prebid, no ad server targeting",
```
AlexBVolcy, line 33: I think this description needs to be updated? I think I see ad server targeting present in request ext.

adservertargeting/requestcache_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+		actualReq := reqImpCache.GetReqJson()
+		assert.Len(t, actualReq, test.expectedReqSize, "incorrect request returned")
+		actualImps, err := reqImpCache.GetImpsData()
+		assert.Len(t, actualImps, test.expectedImpsNum, "incorrect number of impressions returned")
+
+		if test.expectedError {
+			assert.Error(t, err, "unexpected error returned")
+		} else {
+			assert.NoError(t, err, "expected error not returned")
```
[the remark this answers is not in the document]
VeronikaSolovei9: Yes, these should be swapped. Will fix in next commit

adservertargeting/requestcache_test.go
VeronikaSolovei9: I like this idea!

adservertargeting/requestlookup_test.go
VeronikaSolovei9: Right, I modified the description.

AlexBVolcy:
This looks good! I'm interested to see what @hhhjort comments before I approve.

hhhjort (approved):
LGTM

AlexBVolcy (approved):

PR #2596, opened 2023-02-16
cross-repo workflow update

bretg:
researched why forked repos weren't happy with this workflow. Have reason to believe this may fix it.

### DIFF
--- a/.github/workflows/cross-repo-issue.yml
+++ b/.github/workflows/cross-repo-issue.yml
@@ -1,7 +1,7 @@
 name: Cross-repo Issue Creation
 
 on:
-  pull_request:
+  pull_request_target:
     types: [closed]
     branches:
       - "master"


bsardo:
Just curious, what were forks unhappy with? I don't see how this workflow affects them since we're just opening up an issue in the java repo.

bretg:
This workflow fails for every forked repo -- the error is that the secret isn't available, and I did find [documentation](https://docs.github.com/en/actions/security-guides/encrypted-secrets) that says "secrets are not passed to the runner when a workflow is triggered from a forked repository".

I was not able to find a confirmed solution to this problem, but did find a lead that suggests this might help. I tested in my private repo that this doesn't hurt the basic functionality.

bsardo (approved):

SyntaxNode (approved):

PR #2608, opened 2023-02-22
Default imp.secure To 1

SyntaxNode:
Implements https://github.com/synta/prebid-server/issues/2478

### DIFF
--- a/ortb/default_test.go
+++ b/ortb/default_test.go
@@ -13,6 +13,9 @@ import (
 )
 
 func TestSetDefaults(t *testing.T) {
+	secure0 := int8(0)
+	secure1 := int8(1)
+
 	testCases := []struct {
 		name            string
 		givenRequest    openrtb2.BidRequest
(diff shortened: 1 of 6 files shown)


ortb/default_test.go
```diff
@@ -329,3 +337,78 @@ func TestSetDefaultsPriceGranularityRange(t *testing.T) {
...
+			expectedImps:     []*openrtb_ext.ImpWrapper{{Imp: &openrtb2.Imp{Secure: &secure0}}, {Imp: &openrtb2.Imp{Secure: &secure1}}},
+			expectedModified: true,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			actualModified := setDefaultsImp(test.givenImps)
+			assert.Equal(t, test.expectedModified, actualModified)
+			assert.ElementsMatch(t, test.expectedImps, test.givenImps)
```
hhhjort, line 411: I think I am a little impressed that this passes. The `givenImps` secure element should be pointing to `DefaultSecure`, and not `secure1` as `expectedImps` does, though the values pointed to do match. Something I will have to try to remember about the behavior of `assert.ElementsMatch`.

hhhjort (approved):
LGTM

ortb/default_test.go
SyntaxNode: Ah. Surprised by the pointers?

Most of the testify assertions use `reflect.DeepEqual` under the covers which compares values instead of pointer addresses.

VeronikaSolovei9 (approved):

PR #2611, opened 2023-02-27
Criteo: add video capability

vraybaud:


adapters/criteo/criteo.go
```diff
@@ -64,15 +64,30 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
...
 }
+
+func getMediaTypeForImps(imps []openrtb2.Imp) map[string]openrtb_ext.BidType {
+	bidTypes := map[string]openrtb_ext.BidType{}
+	for _, imp := range imps {
+		if imp.Video != nil {
+			bidTypes[imp.ID] = openrtb_ext.BidTypeVideo
+		} else {
+			bidTypes[imp.ID] = openrtb_ext.BidTypeBanner
+		}
```
onkarvhanumante, line 89: Please update [bidder info](https://github.com/prebid/prebid-server/blob/23927807bd1d4e3ea3ec270db776ce2de5ba7272/static/bidder-info/criteo.yaml#L5-L11)  to include video as supported type. 

https://github.com/prebid/prebid-server/blob/23927807bd1d4e3ea3ec270db776ce2de5ba7272/static/bidder-info/criteo.yaml#L5-L11

Also create

onkarvhanumante (commented):
@vraybaud  could you also create docs PR to include video as supported type in https://github.com/prebid/prebid.github.io repo

adapters/criteo/criteo.go
vraybaud: Thanks for comment, I've updated the config accordingly.

onkarvhanumante (approved):

adapters/criteo/criteo.go
```diff
@@ -64,15 +64,30 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
...
 
 	return bidResponse, nil
 }
+
+func getMediaTypeForImps(imps []openrtb2.Imp) map[string]openrtb_ext.BidType {
+	bidTypes := map[string]openrtb_ext.BidType{}
+	for _, imp := range imps {
+		if imp.Video != nil {
+			bidTypes[imp.ID] = openrtb_ext.BidTypeVideo
+		} else {
```
gargcreation1992, line 87: Warning - If request has multi format bid, this code will by default set the bid type to Video. Is that the intention here?

adapters/criteo/criteotest/exemplary/simple-video.json
```diff
@@ -0,0 +1,137 @@
+{
```
gargcreation1992, line 1: Would like to request to add a test case in which request is having both mediatype impressions - banner and video.

gargcreation1992 (commented):
left few comments.

vraybaud:
Hi @gargcreation1992 , thanks a lot for your comments. We didn't take into account this case in the pull request as it was. I did some changes in the pull request. We decided to set directly the seatbid.bid.ext.prebid.type in our bidder so we don't add any additional logic regarding this in the adapter. Does it sounds fine for you ? I have updated the tests accordingly.

vraybaud:
Hi @gargcreation1992 , thanks a lot for your comments. We didn't take into account the multi type case in this pull request. I did some changes to handle it. We decided to set directly the seatbid.bid.ext.prebid.type in our bidder so we don't add any additional logic regarding this in the adapter. Does it sound fine for you ? I have updated the tests accordingly.

adapters/criteo/criteo.go
```diff
@@ -64,30 +64,16 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
...
 	bidResponse := adapters.NewBidderResponse()
 	bidResponse.Currency = response.Cur
-	bidTypes := getMediaTypeForImps(request.Imp)
 
 	for _, seatBid := range response.SeatBid {
 		for i := range seatBid.Bid {
+
 			b := &adapters.TypedBid{
-				Bid:     &seatBid.Bid[i],
-				BidType: bidTypes[seatBid.Bid[i].ImpID],
```
gargcreation1992: We cannot un-assign BidType. This has to be set else the response from your adapter will overwrite the `seatbid [i].bid.ext.prebid.type` value in the response. Refer to this comment on [TypedBid struct.](https://github.com/prebid/prebid-server/blob/6aab539f321ca50ffccf1d84f4399ec4567919ba/adapters/bidder.go#L94-L95)

I would suggest to set the BidType based on the `response.seatbid[i].bid.ext.prebid.type`.

gargcreation1992 (changes requested):
left a comment. Also requesting to address [previous comment of adding test cases.](https://github.com/prebid/prebid-server/pull/2611#discussion_r1121221527)

adapters/criteo/criteo.go
vraybaud: Hello, I have done the changes and I am now putting into BidType the value from `response.seatbid[i].bid.ext.prebid.type`

adapters/criteo/criteotest/exemplary/simple-video.json
vraybaud: Thanks, I added test cases with request having both mediatypes.

adapters/criteo/criteo.go
```diff
@@ -67,9 +67,18 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
 
 	for _, seatBid := range response.SeatBid {
 		for i := range seatBid.Bid {
+			var bidExt openrtb_ext.ExtBid
+			unmarshalExtErr := json.Unmarshal(seatBid.Bid[i].Ext, &bidExt)
+			if unmarshalExtErr != nil {
+				err := &errortypes.BadServerResponse{
+					Message: fmt.Sprintf("Missing ad type in response: %d.", responseData.StatusCode),
```
gargcreation1992, line 74: I do not think you should set `responseData.StatusCode` in the error here. The reason is that value will be 200 by the time code reaches here.

gargcreation1992 (changes requested):
left a comment on error log.

adapters/criteo/criteo.go
```diff
@@ -67,9 +67,18 @@ func (a *adapter) MakeBids(request *openrtb2.BidRequest, requestData *adapters.R
...
+				}
+				return nil, []error{err}
+			}
```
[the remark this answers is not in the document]
vraybaud: Thank you for your comment, I've updated the code.

gargcreation1992 (approved):
👍

onkarvhanumante (approved):

Issue #2614, opened 2023-03-01
Stored auction response is returning empty string for bid.ext.prebid.type field


camrice:
When using a request with an `imp.storedauctionresponse.id`, the response will contain an empty `bid.ext.prebid.type` field, even if this field is included in the stored auction response. The response from Prebid Server will instead contain an empty string. Example response below:
```
{
    "bid": {
        ...
        "ext": {
            "appnexus": {
                "auction_id": 4567,
                "bid_ad_type": 3,
                "bidder_id": 2,
                "brand_id": 1
            },
            "origbidcpm": 0,
            "origbidcur": "USD",
            "prebid": {
                "meta": {
                    "adaptercode": "appnexus"
                },
                "type": "",
                "bidid": "af3a5b4b-25d8-4c50-8e72-d7fdf6535382"
            }
        }
    }
}
```

Initial findings below from @VeronikaSolovei9 :

The Prebid type field is present on stored bid responses because these responses will go through `MakeBids()` and will populate the field. In the case of using the Appnexus adapter and a stored auction response, there is a portion that gets missed in `appnexus.go` line 403:

```
bidResponse.Bids = append(bidResponse.Bids, &adapters.TypedBid{
						Bid:          &bid,
						BidType:      bidType,
						BidVideo:     impVideo,
						DealPriority: bidExt.Appnexus.DealPriority,
					})
```

This seems to be needed for the field to work with auction responses.

VeronikaSolovei9:
Thank you Cameron, good finding!

camrice:
Thank you for looking into this Veronika! I'll be in touch for any updates.

VeronikaSolovei9:
Fix PR: https://github.com/prebid/prebid-server/pull/2615

camrice:
Sorry for the late update, but I wanted to confirm that this is now working for our usage. Thank you!

bretg:
Confirmed this is not a problem in PBS-Java.

PR #2615, opened 2023-03-02
Fix for storedauctionresponse bid type

VeronikaSolovei9:
Fix for https://github.com/prebid/prebid-server/issues/2614

### DIFF
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -1317,10 +1317,16 @@ func buildStoredAuctionResponse(storedAuctionResponses map[string]json.RawMessag
 				mType := seat.Bid[i].MType
 				var bidType openrtb_ext.BidType
 				if mType > 0 {
-					if mType <= 4 {
-						//1 = Banner, 2 = Video,  3 = Audio, 4 = Native
-						bidType = openrtb_ext.BidTypes()[mType-1]
-					} else {
+					switch mType {
+					case openrtb2.MarkupBanner:
+						bidType = openrtb_ext.BidTypeBanner
+					case openrtb2.MarkupVideo:
+						bidType = openrtb_ext.BidTypeVideo
+					case openrtb2.MarkupAudio:
+						bidType = openrtb_ext.BidTypeAudio
+					case openrtb2.MarkupNative:
+						bidType = openrtb_ext.BidTypeNative
+					default:
 						return nil, nil, nil, &errortypes.BadServerResponse{
 							Message: fmt.Sprintf("Failed to parse bid mType for impression \"%s\"", seat.Bid[i].ImpID),
 						}
--- a/exchange/exchange.go
+++ b/exchange/exchange.go
@@ -1280,7 +1280,7 @@ func listBiddersWithRequests(bidderRequests []BidderRequest) []openrtb_ext.Bidde
 	return liveAdapters
 }
 
-func getMediaTypeForBid(bid openrtb2.Bid) (openrtb_ext.BidType, error) {
+func getPrebidMediaTypeForBid(bid openrtb2.Bid) (openrtb_ext.BidType, error) {
 	if bid.Ext != nil {
 		var bidExt openrtb_ext.ExtBid
 		err := json.Unmarshal(bid.Ext, &bidExt)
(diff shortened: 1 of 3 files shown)


hhhjort (approved):

exchange/exchange.go
```diff
@@ -1280,6 +1280,20 @@ func listBiddersWithRequests(bidderRequests []BidderRequest) []openrtb_ext.Bidde
 	return liveAdapters
 }
 
+func getMediaTypeForBid(bid openrtb2.Bid) (openrtb_ext.BidType, error) {
```
SyntaxNode, line 1283: Nitpick: Consider changing the name to `getPrebidMediaTypeForBid` to clarify what field this method is extracting.

exchange/exchange.go
```diff
@@ -1300,7 +1314,18 @@ func buildStoredAuctionResponse(storedAuctionResponses map[string]json.RawMessag
 			//set imp id from request
 			for i := range seat.Bid {
 				seat.Bid[i].ImpID = impId
-				bidsToAdd = append(bidsToAdd, &entities.PbsOrtbBid{Bid: &seat.Bid[i]})
+				mType := seat.Bid[i].MType
+				var bidType openrtb_ext.BidType
+				if mType > 0 {
+					bidType = openrtb_ext.BidTypes()[mType-1]
```
SyntaxNode, line 1320: Please validate the mtype is valid (if greater than 0). If it's not valid, consider responding with the same `BadServerResponse` message.

exchange/exchange.go
VeronikaSolovei9: Done!

exchange/exchange.go
VeronikaSolovei9: Added

hhhjort (approved):

exchange/exchange.go
```diff
@@ -1300,7 +1314,25 @@ func buildStoredAuctionResponse(storedAuctionResponses map[string]json.RawMessag
...
+				var bidType openrtb_ext.BidType
+				if mType > 0 {
+					if mType <= 4 {
+						//1 = Banner, 2 = Video,  3 = Audio, 4 = Native
+						bidType = openrtb_ext.BidTypes()[mType-1]
+					} else {
+						return nil, nil, nil, &errortypes.BadServerResponse{
+							Message: fmt.Sprintf("Failed to parse bid mType for impression \"%s\"", seat.Bid[i].ImpID),
+						}
+					}
```
SyntaxNode, line 1327: This code assumes the order of enums in PBS matches that of OpenRTB 3.0. This mapping is intentional, but subject to change in the future on either side. IMHO it would be better to be explicit:

```
if mType > 0 {
	switch mType {
	case openrtb2.MarkupBanner:
		bidType = openrtb_ext.BidTypeBanner
	case openrtb2.MarkupVideo:
		bidType = openrtb_ext.BidTypeVideo
	case openrtb2.MarkupAudio:
		bidType = openrtb_ext.BidTypeAudio
	case openrtb2.MarkupNative:
		bidType = openrtb_ext.BidTypeNative
	default:
		return nil, nil, nil, &errortypes.BadServerResponse{
			Message: fmt.Sprintf("Failed to parse bid mType for impression \"%s\"", seat.Bid[i].ImpID),
		}
	}
```

exchange/exchange.go
```diff
@@ -1300,7 +1314,25 @@ func buildStoredAuctionResponse(storedAuctionResponses map[string]json.RawMessag
...
+				} else {
+					var err error
+					bidType, err = getPrebidMediaTypeForBid(seat.Bid[i])
+					if err != nil {
+						return nil, nil, nil, err
```
SyntaxNode, line 1332: This error case has no test coverage.

exchange/exchange.go
VeronikaSolovei9: Oh I see what you meant. Added!

exchange/exchange.go
VeronikaSolovei9: Added. 
Are we ok to return error as is from `ParseBidType`? Or is it better convert it to `BadServerResponse` for consistency?

SyntaxNode: I don't have a strong opinion on this. I do like consistency and support you on that.

hhhjort (approved):

exchange/exchange.go
VeronikaSolovei9: Discussed offline. No changes needed.

SyntaxNode (approved):

PR #2617, opened 2023-03-02
AppNexus adapter - Accept placementId as int or string

gargcreation1992:
This PR also adds usePaymentRule attribute to ExtImpAppnexus struct to be in consistent with PBJS.


util/typeutil/stringInt.go
```diff
@@ -0,0 +1,32 @@
...
+	case float64:
+		*st = StringInt(int(v))
+	case string:
+		///here convert the string into an int
+		i, err := strconv.Atoi(v)
+		if err != nil {
+			return err
+
+		}
+		*st = StringInt(i)
```
onkarvhanumante, line 29: should add `default` case to return error if non of the above case match

openrtb_ext/imp_appnexus.go
```diff
@@ -1,13 +1,17 @@
...
+	"github.com/prebid/prebid-server/util/typeutil"
+)
 
 // ExtImpAppnexus defines the contract for bidrequest.imp[i].ext.prebid.bidder.appnexus
 type ExtImpAppnexus struct {
-	LegacyPlacementId       int                     `json:"placementId"`
+	LegacyPlacementId       typeutil.StringInt      `json:"placementId"`
 	LegacyInvCode           string                  `json:"invCode"`
 	LegacyTrafficSourceCode string                  `json:"trafficSourceCode"`
-	PlacementId             int                     `json:"placement_id"`
```
onkarvhanumante, line 10: should we use  `json.Number` type here? It handles both int and string values

Sonali-More-Xandr (commented):
We should also make changes in [bidder-params/appnexus.json](https://github.com/prebid/prebid-server/blob/master/static/bidder-params/appnexus.json).

util/typeutil/stringInt.go
```diff
@@ -0,0 +1,32 @@
+package typeutil
```
SyntaxNode, line 1: This might be a better fit for the existing `jsonutil` package as it's a helper for json encoding.

util/typeutil/stringInt.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+type StringInt int
```
SyntaxNode, line 8: There's probably a better name we could use, though I don't have an elegant suggestion. Perhaps other reviewers might have an idea.

util/typeutil/stringInt.go
```diff
@@ -0,0 +1,32 @@
...
+	var item interface{}
+	if err := json.Unmarshal(b, &item); err != nil {
+		return err
+	}
+
+	switch v := item.(type) {
+	case int:
+		*st = StringInt(v)
+	case float64:
+		*st = StringInt(int(v))
```
SyntaxNode, line 21: Why do we need to support a float?

util/typeutil/stringInt.go
SyntaxNode: There's a `return nil` at the end of the method to provide the same effect. I suppose it's just a matter of preference.

util/typeutil/stringInt.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+type StringInt int
+
+func (st *StringInt) UnmarshalJSON(b []byte) error {
+	//convert the bytes into an interface as this will help us check the type of our value
+	var item interface{}
+	if err := json.Unmarshal(b, &item); err != nil {
+		return err
+	}
```
SyntaxNode, line 15: If you're up for running a benchmark test, I am curious if there is a performance cost for unmarshalling to an empty interface (one unmarshal call total), instead of attempting to unmarshal to specific types (one unmarshal call for the target type). Perhaps its even faster to use jsonparser Get methods for int and string?

util/typeutil/stringInt_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnmarshalJSON(t *testing.T) {
+	type Item struct {
+		ItemId StringInt `json:"item_id"`
+	}
+
+	t.Run("happy path - convert string to int", func(t *testing.T) {
```
SyntaxNode, line 15: Test names cannot have spaces, they are converted to underscored under the covers. Therefore, I recommend to use a simple name using hyphens or underscores, ex: `string` and `int` would be descriptive enough names IMHO given the context.

util/typeutil/stringInt_test.go
```diff
@@ -0,0 +1,28 @@
+package typeutil
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnmarshalJSON(t *testing.T) {
```
SyntaxNode, line 10: Since test are package scoped, please include the type name, ex: `TestStringIntUnmarshalJSON` or since there is no other functionality for the StringInt type, `TestStringInt` would also be sufficient.

util/typeutil/stringInt_test.go
```diff
@@ -0,0 +1,28 @@
...
+		assert.Equal(t, 30, int(item.ItemId))
+	})
+
+	t.Run("happy path - make sure integer still works", func(t *testing.T) {
+		jsonData := []byte(`{"item_id":30}`)
+		var item Item
+		assert.NoError(t, json.Unmarshal(jsonData, &item))
+		assert.Equal(t, 30, int(item.ItemId))
+	})
+}
```
SyntaxNode, line 28: Please include error conditions, such as a string which cannot be converted to an integer or a different incompatible type, like a boolean.

SyntaxNode (commented):
I like that you're building a new type instead of using `json.Number`, since that pushes conversion outside of the unmarshal call and IMHO it's better to signal a type error through the unmarshal chain.

util/typeutil/stringInt.go
gargcreation1992: The another name that came up is StringOrInt but I find StringInt more concise.

util/typeutil/stringInt.go
gargcreation1992: good question. To Unmarshal Json in interface, Unmarshal stores JSON numbers in float64. Here is the official doc link - https://pkg.go.dev/encoding/json#Unmarshal

util/typeutil/stringInt.go
gargcreation1992: Using JsonParser seems to be the most efficient to use while doing json.Unmarshal. The only catch is when string is provided as input we have to remove double quotes in Unmarshal function. Here is the how I implemented the benchmarking code. Here are the benchmarking results on my local machine - 

```
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsString-16                	 1761681	       677.5 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsString-16                	 1658487	       710.2 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsString-16                	 1775126	       685.4 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsString-16                	 1744000	       675.1 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsString-16                	 1771329	       680.5 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsInt-16                   	 1715920	       697.6 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsInt-16                   	 1744148	       724.0 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsInt-16                   	 1720200	       704.9 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsInt-16                   	 1626032	       699.2 ns/op
BenchmarkUnmarshalJSONUsingEmptyInterfaceWithInputAsInt-16                   	 1714712	       694.2 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsString-16                    	 2586522	       447.4 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsString-16                    	 2704970	       450.6 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsString-16                    	 2475841	       451.7 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsString-16                    	 2574088	       463.3 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsString-16                    	 2684553	       452.3 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsInt-16                       	 2616318	       447.0 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsInt-16                       	 2645331	       453.4 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsInt-16                       	 2644465	       451.9 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsInt-16                       	 2637205	       461.6 ns/op
BenchmarkUnmarshalJSONUsingJsonParserWithInputAsInt-16                       	 2617648	       450.8 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsString-16    	 2309619	       511.4 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsString-16    	 2301285	       512.1 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsString-16    	 2358621	       509.2 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsString-16    	 2316736	       518.6 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsString-16    	 2313619	       534.0 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsInt-16       	 2541432	       466.0 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsInt-16       	 2620980	       466.5 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsInt-16       	 2603096	       464.6 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsInt-16       	 2617431	       464.7 ns/op
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplaceWithInputAsInt-16       	 2596263	       461.2 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsString-16                      	 2394945	       518.6 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsString-16                      	 2393102	       507.8 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsString-16                      	 2366752	       500.1 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsString-16                      	 2366858	       499.0 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsString-16                      	 2379032	       559.0 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsInt-16                         	 2462937	       487.6 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsInt-16                         	 2549624	       485.0 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsInt-16                         	 2558164	       458.3 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsInt-16                         	 2648469	       447.8 ns/op
BenchmarkUnmarshalJSONUsingDataTypeWithInputAsInt-16                         	 2622322	       464.6 ns/op
```

SyntaxNode: Excellent! Thank you for running the benchmarks so we can decide based on actual performance data. I summarized the different approaches:

```
BenchmarkUnmarshalJSONUsingEmptyInterface              694.86 ns/op average
BenchmarkUnmarshalJSONUsingJsonParser                  453.00 ns/op average
BenchmarkUnmarshalJSONUsingJsonParserAndStringReplace  490.83 ns/op average
BenchmarkUnmarshalJSONUsingDataType                    492.78 ns/op average
```

The `EmptyInterface` clearly performs the worst, so I think we should rule that out. I'm not surprised by this finding since reflection is still expensive in Go. 

`JsonParser` is the fastest and might be a good choice. You may even be able to further improve performance by replacing `bytes.HasPrefix` with a hardcoded check of the first character (if there is a first character). It seems `JsonParserAndStringReplace` is the lower performance variant and should be excluded as an option.

The `DataType` approach has good performance if we want to keep with the stock json package, but I see no compelling reason for doing so. That package has improved validation checks, but by the time we've reached this part of code the json has already been validated.

adapters/appnexus/appnexustest/supplemental/placement-id-as-string-test.json
```diff
@@ -0,0 +1,151 @@
...
...FEIi6jeuTm_LoTRib7t2u2tLMlnMqNgkAAAECCPA_EQEHEAAA8D8ZCQkIAAAhCQkI8D8pEQkAMQkJqAAAMKqI2wQ4vgdAvgdIAlC95ZchWPj8Q2AAaJFAeJLRBIABAYoBA1VTRJIFBvBQmAEBoAEBqAEBsAEAuAECwAEEyAEC0AEJ2AEA4AEB8AEAigI7dWYoJ2EnLCAxMzc2ODYwLCAxNTE5MzA5NDAxKTt1ZigncicsIDY5NTk1ODM3Nh4A8IqSAvUBIXRETkdfUWlHbW9vSEVMM2xseUVZQUNENF9FTXdBRGdBUUFSSXZnZFFxb2piQkZnQVlMTURhQUJ3QUhnQWdBRUFpQUVBa0FFQm1BRUJvQUVCcUFFRHNBRUF1UUVwaTRpREFBRHdQOEVCS1l1SWd3QUE4RF9KQVhfelYzek1zXzBfMlFFQUFBAQMkRHdQLUFCQVBVQgEOLEFKZ0NBS0FDQUxVQwUQBEwwCQjwTE1BQ0FNZ0NBT0FDQU9nQ0FQZ0NBSUFEQVpBREFKZ0RBYWdEaHBxS0I3b0RFV1JsWm1GMWJIUWpUbGxOTWpvME1ESXqaAjkhT3d3R0FRNvgA8E4tUHhESUFRb0FEb1JaR1ZtWVhWc2RDTk9XVTB5T2pRd01qTS7YAugH4ALH0wHqAgpwcmViaWQub3Jn8gIRCgZBRFZfSUQSBzEzNzY4NjDyARQMQ1BHXwEUNDM1MDMwOTjyAhEKBUNQARPwmQgxNDg0NzIzOIADAYgDAZADAJgDFKADAaoDAMADkBzIAwDYAwDgAwDoAwD4AwOABACSBAkvb3BlbnJ0YjKYBACiBAwxNTIuMTkzLjYuNzSoBJrMI7IEDAgAEAAYACAAMAA4ALgEAMAEAMgEANIEEWRlZmF1bHQjTllNMjo0MDIz2gQCCADgBADwBL3llyGIBQGYBQCgBf____8FA1ABqgULc29tZS1yZXEtaWTABQDJBQAFARTwP9IFCQkFC2QAAADYBQHgBQHwBd4C-gUECAAQAJAGAZgGAA..&s=08b1535744639c904684afe46e3c6c0e4786089f&test=1&referrer=prebid.org&pp=${AUCTION_PRICE}\"],\"jstracker\":\"<script src=\\\"http://www.dummyurl.js\\\"></script>\"}",
```
VeronikaSolovei9, line 83: Is such a big "adm" needed here? This test is about placement Id, not adm. Please consider to use simpler string, look for examples in other json tests

VeronikaSolovei9:
There are some adapters that support multitype parameters. I hope you can apply it for Appnexus too. Please consider this example: 
Declare ext parameter like this: 
https://github.com/prebid/prebid-server/blob/6b4b3f16f3682173cfc458de523761c2d983681a/static/bidder-params/adf.json#L8

Then in imp_{bidder}.go declare type like this: 
https://github.com/prebid/prebid-server/blob/6b4b3f16f3682173cfc458de523761c2d983681a/openrtb_ext/imp_adf.go#L8

And then in code covert it to the type you need, look for usages of this `MasterTagID` field: 
https://github.com/prebid/prebid-server/blob/6b4b3f16f3682173cfc458de523761c2d983681a/adapters/adf/adf.go#L54

Will it solve your case?

openrtb_ext/imp_appnexus.go
```diff
@@ -1,20 +1,25 @@
...
+	PlacementId             typeutil.StringInt      `json:"placement_id"`
 	InvCode                 string                  `json:"inv_code"`
 	Member                  string                  `json:"member"`
 	Keywords                []*ExtImpAppnexusKeyVal `json:"keywords"`
 	TrafficSourceCode       string                  `json:"traffic_source_code"`
 	Reserve                 float64                 `json:"reserve"`
 	Position                string                  `json:"position"`
 	UsePmtRule              *bool                   `json:"use_pmt_rule"`
+	UsePaymentRule          *bool                   `json:"usePaymentRule"`
```
VeronikaSolovei9, line 22: This parameter has no usages, do we really need to add it?

util/typeutil/stringInt.go
gargcreation1992: thanks @SyntaxNode! Based on your suggestion above, I was able to achieve some more performance improvement by directly checking the first character. Here is the revised version. Please take a look - https://github.com/prebid/prebid-server/pull/2617/commits/9e4089964012932b327478fc68b37b411a51bdcf

util/typeutil/stringInt.go
gargcreation1992: this is done - https://github.com/prebid/prebid-server/pull/2617/commits/cad0d8555f0ae53f362d09c6f837cebb5c9ab707

adapters/appnexus/appnexustest/supplemental/placement-id-as-string-test.json
gargcreation1992: this is done - https://github.com/prebid/prebid-server/pull/2617/commits/311e9f3f084e49142987c68479fe959b1141c7e6

util/typeutil/stringInt_test.go
gargcreation1992: this is done - https://github.com/prebid/prebid-server/pull/2617/commits/0141e91006f95b9a243dc679fd28b050750951bf

util/typeutil/stringInt_test.go
gargcreation1992: this is done - https://github.com/prebid/prebid-server/pull/2617/commits/9e4089964012932b327478fc68b37b411a51bdcf

openrtb_ext/imp_appnexus.go
gargcreation1992: > 991604b

@Sonali-More-Xandr this is done here - https://github.com/prebid/prebid-server/pull/2617/commits/991604bcf26e15a5a80445470a37033b22b40beb

gargcreation1992:
> We should also make changes in [bidder-params/appnexus.json](https://github.com/prebid/prebid-server/blob/master/static/bidder-params/appnexus.json).

@Sonali-More-Xandr this is done here - https://github.com/prebid/prebid-server/pull/2617/commits/991604bcf26e15a5a80445470a37033b22b40beb

openrtb_ext/imp_appnexus.go
gargcreation1992: @VeronikaSolovei9 thank you for your review. We need usePaymentRule here to be in consistent with PBJS input params.

gargcreation1992:
> Will it solve your case?

This should solve the case. We evaluated json.Number as an option data type to accept both integer and string. However, we did not take that option because json.Number implementation pushes the conversion outside of json.Unmarshal call. This is also documented in this [PR comment as well.](https://github.com/prebid/prebid-server/pull/2617#pullrequestreview-1322054803)

util/typeutil/stringInt_test.go
gargcreation1992: this is done - https://github.com/prebid/prebid-server/commit/9e4089964012932b327478fc68b37b411a51bdcf

openrtb_ext/imp_appnexus.go
gargcreation1992: @onkarvhanumante We are not taking this option. Please see comment here - https://github.com/prebid/prebid-server/pull/2617#pullrequestreview-1322054803

util/typeutil/stringInt.go
gargcreation1992: This is not applicable now since we have changed the implementation of the Unmarshal function based on benchmarking results.

onkarvhanumante (approved):
nicely done. good work on commit breakdowns 👍🏼 💯

adapters/appnexus/params_test.go
```diff
@@ -41,6 +41,7 @@ func TestInvalidParams(t *testing.T) {
 
 var validParams = []string{
 	`{"placement_id":123}`,
+	`{"placement_id":"123"}`,
 	`{"placementId":123,"position":"above"}`,
```
Sonali-More-Xandr, line 45: Nitpick - Can we also add `{"placementId":"123"}` in validParams

gargcreation1992: thanks @Sonali-More-Xandr, this helped in fixing a small bug as well in bidderparams - https://github.com/prebid/prebid-server/pull/2617/commits/ad93e5a76cfcb96ac14724710eb99c1b36f97230

Sonali-More-Xandr (approved):
Good work, Ashish!

openrtb_ext/imp_appnexus.go
VeronikaSolovei9: Ok, please don't forget to add it to appnexus.json for consistency

VeronikaSolovei9:
We discussed `json.Number` vs `StringInt` usages with the team. Probably in future `StringInt` will replace json.Number across all adapters. But not now :)

VeronikaSolovei9:
> @VeronikaSolovei9 thank you for your review. We need usePaymentRule here to be in consistent with PBJS input params.

Ok, please don't forget to add it to appnexus.json for consistency

gargcreation1992:
> > @VeronikaSolovei9 thank you for your review. We need usePaymentRule here to be in consistent with PBJS input params.
> 
> Ok, please don't forget to add it to appnexus.json for consistency

added here - https://github.com/prebid/prebid-server/pull/2617/commits/04177c5f2003010b263d61d7256f2fb0bcbdfd27

> UPD: please make sure it's also updated here: https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/appnexus.md

will raise the PR to change the docs.

onkarvhanumante (dismissed):

openrtb_ext/imp_appnexus.go
```diff
@@ -1,20 +1,25 @@
...
+	LegacyInvCode            string                  `json:"invCode"`
+	LegacyTrafficSourceCode  string                  `json:"trafficSourceCode"`
+	PlacementId              jsonutil.StringInt      `json:"placement_id"`
+	InvCode                  string                  `json:"inv_code"`
+	Member                   string                  `json:"member"`
+	Keywords                 []*ExtImpAppnexusKeyVal `json:"keywords"`
+	TrafficSourceCode        string                  `json:"traffic_source_code"`
+	Reserve                  float64                 `json:"reserve"`
+	Position                 string                  `json:"position"`
+	UsePmtRule               *bool                   `json:"use_pmt_rule"`
```
SyntaxNode, line 21: Nitpick: Consider naming this variable `UsePaymentRule` to align with `DeprecatedUsePaymentRule `

adapters/appnexus/appnexus.go
```diff
@@ -284,15 +284,18 @@ func preprocess(imp *openrtb2.Imp, defaultDisplayManagerVer string) (string, boo
...
 	}
 	if appnexusExt.InvCode == "" && appnexusExt.LegacyInvCode != "" {
 		appnexusExt.InvCode = appnexusExt.LegacyInvCode
 	}
 	if appnexusExt.TrafficSourceCode == "" && appnexusExt.LegacyTrafficSourceCode != "" {
 		appnexusExt.TrafficSourceCode = appnexusExt.LegacyTrafficSourceCode
 	}
+	if appnexusExt.DeprecatedUsePaymentRule != nil && *appnexusExt.DeprecatedUsePaymentRule {
+		appnexusExt.UsePmtRule = appnexusExt.DeprecatedUsePaymentRule
+	}
```
SyntaxNode, line 298: If both the new and deprecated payment rule are provided, this code will use the deprecated one. This is the opposite behavior as the PlacementID.

gargcreation1992: After discussion with the team, I have updated to precede `use_pmt_rule` over `usePaymentRule`. Also changed appnexus.json to reflect this change. Here is the change - https://github.com/prebid/prebid-server/pull/2617/commits/d4fff68d92d3d2dfd1042ea36c45c70e18c7f01b

openrtb_ext/imp_appnexus.go
gargcreation1992: thank you! changed here - https://github.com/prebid/prebid-server/pull/2617/commits/310a5f8baee8655070551e78f473f70b7dad36ee

onkarvhanumante (dismissed):

Sonali-More-Xandr (approved):
Good work, Ashish!

onkarvhanumante (approved):

PR #2619, opened 2023-03-03
Appnexus adapter - iab categories initialisation changes

onkarvhanumante:
- Builder function initialises iabCategoryMap. This function makes using of opts.json file.  
- PR makes changes to initialise  IAB categories using in memory map. Thereby reduces IO operation to read opts.json. Also remove opts.json and related code.
- Note: Scope of iabCategoryMap initialisation is limited to Builder function and we don't want to update iabCategoryMap after initialisation. Therefore iabCategoryMap is not declared as global variable.


adapters/appnexus/appnexus.go
```diff
@@ -471,17 +473,3 @@ func resolvePlatformID(platformID string) int {
 
 	return defaultPlatformID
 }
-
-func loadCategoryMapFromFileSystem() map[string]string {
-	// Load custom options for our adapter (currently just a lookup table to convert appnexus => iab categories)
-	opts, err := os.ReadFile("./static/adapter/appnexus/opts.json")
-	if err == nil {
-		var adapterOptions appnexusAdapterOptions
```
Sonali-More-Xandr, line 479: Now that we have in memory map for IAB categories, can we also remove `appnexusAdapterOptions` struct?

onkarvhanumante: good catch.  https://github.com/prebid/prebid-server/pull/2619/commits/03459f1350f2644165a2f878c0aaddb9373899a7 removes `appnexusAdapterOptions`

Sonali-More-Xandr (approved):

adapters/appnexus/appnexus.go
```diff
@@ -455,9 +450,12 @@ func appendMemberId(uri string, memberId string) string {
...
 	bidder := &adapter{
-		URI:            config.Endpoint,
-		iabCategoryMap: loadCategoryMapFromFileSystem(),
-		hbSource:       resolvePlatformID(config.PlatformID),
+		URI: config.Endpoint,
+		iabCategoryMap: map[string]string{
+			"1": "IAB20-3",
+			"9": "IAB5-3",
+		},
+		hbSource: resolvePlatformID(config.PlatformID),
```
gargcreation1992, line 458: the code formatting looks off to me maybe because of map literal. Should we create a constant instead since this map value will not change during the execution of the program?

onkarvhanumante: > the code formatting looks off to me maybe because of map literal

This might be related to way diff is viewed in browser.

> Should we create a constant instead since this map value will not change during the execution of the program?
Map cannot be declared as constant.

Refer https://go.dev/doc/effective_go#constants

`Constants in Go are just that—constant. They are created at compile time, even when defined as locals in functions, and can only be numbers, characters (runes), strings or booleans`

gargcreation1992 (approved):
👍

PR #2621, opened 2023-03-03
Support GDPR and USPrivacy from GPP downgrade for adapters that do not support GPP

hhhjort:


exchange/utils.go
```diff
@@ -847,6 +859,49 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
 	return allBidderRequests
 }
 
+func gdprFromGPP(r *openrtb2.BidRequest, gpp gpplib.GppContainer) {
+	if r.Regs != nil && r.Regs.GDPR == nil {
+		if len(r.Regs.GPPSID) > 0 {
+			// Set to 0 unless SID exists
+			r.Regs.GDPR = ptrutil.ToPtr[int8](0)
+			for _, id := range r.Regs.GPPSID {
+				if id == int8(gppConstants.SectionTCFEU2) {
```
SyntaxNode, line 868: Seeing this cast... Do you think we should change the gpp constants to int8 for easier comparison with the openrtb library?

exchange/utils.go
```diff
@@ -847,6 +859,49 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+					r.Regs.GDPR = ptrutil.ToPtr[int8](1)
```
SyntaxNode, line 869: I believe r.Regs and r.User is shared memory at this point in code, right? We will need to make a copy of these objects before applying a bidder specific setting. Are there tests we can add a higher level that would catch this scenario?

exchange/utils.go
```diff
@@ -847,6 +859,49 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+						if id == gppConstants.SectionUSPV1 {
+							r.Regs.USPrivacy = gpp.Sections[i].GetValue()
+						}
+					}
+				}
+			}
+		}
+	}
+
+}
```
SyntaxNode, line 903: The many indents make this hard to read. Consider refactoring to something like:

```
if r.Regs == nil {
	return
}

if len(r.Regs.USPrivacy) > 0 {
	return
}

for _, id := range r.Regs.GPPSID {
	if id == int8(gppConstants.SectionUSPV1) {
		for i, id := range gpp.SectionTypes {
			if id == gppConstants.SectionUSPV1 {
				r.Regs.USPrivacy = gpp.Sections[i].GetValue()
			}
		}
	}
}
```

exchange/utils_test.go
```diff
@@ -3476,3 +3478,306 @@ func TestCleanOpenRTBRequestsFilterBidderRequestExt(t *testing.T) {
...
+	testCases := []struct {
+		name            string
+		desc            string
+		initialRequest  *openrtb2.BidRequest
+		gpp             gpplib.GppContainer
+		expectedRequest *openrtb2.BidRequest
+	}{
+		{
+			name:            "Empty",
+			desc:            "Empty Request",
```
SyntaxNode, line 3505: Consider removing the desc field and use a comment beside the name instead:

```
name: "empty" // empty request
```

exchange/utils.go
```diff
@@ -847,6 +859,49 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
 	return allBidderRequests
 }
 
+func gdprFromGPP(r *openrtb2.BidRequest, gpp gpplib.GppContainer) {
+	if r.Regs != nil && r.Regs.GDPR == nil {
+		if len(r.Regs.GPPSID) > 0 {
```
SyntaxNode, line 864: Should be `r.Regs.GPPSID != nil` to detect if SID exists. We can set GDPR=0 if `GPPSID` exists but is empty.

exchange/utils.go
```diff
@@ -847,6 +859,49 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+				r.User.Consent = gpp.Sections[i].GetValue()
+				return
+			}
+		}
+	}
+
+}
+func privacyFromGPP(r *openrtb2.BidRequest, gpp gpplib.GppContainer) {
+	if r.Regs != nil && len(r.Regs.USPrivacy) == 0 {
+		if len(r.Regs.GPPSID) > 0 {
```
SyntaxNode, line 890: The length check is not necessary. The range iterator gracefully handles nil or empty arrays.

exchange/utils.go
```diff
@@ -179,6 +181,16 @@ func (rs *requestSplitter) cleanOpenRTBRequests(ctx context.Context,
...
+		// GPP downgrade: always downgrade unless we can confirm GPP is supported
+		if binfo, ok := rs.bidderInfo[string(bidderRequest.BidderCoreName)]; ok {
+			if binfo.OpenRTB == nil || !binfo.OpenRTB.GPPSupported {
+				gdprFromGPP(bidderRequest.BidRequest, gpp)
+				privacyFromGPP(bidderRequest.BidRequest, gpp)
+			}
+		} else {
+			gdprFromGPP(bidderRequest.BidRequest, gpp)
+			privacyFromGPP(bidderRequest.BidRequest, gpp)
+		}
```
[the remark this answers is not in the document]
hhhjort: fixed

exchange/utils.go
```diff
@@ -847,6 +859,49 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
 	return allBidderRequests
 }
 
+func gdprFromGPP(r *openrtb2.BidRequest, gpp gpplib.GppContainer) {
```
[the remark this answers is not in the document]
hhhjort: Finding good names is always a challenge.

exchange/utils.go
hhhjort: That is an idea. But it cannot be done in this PR.

exchange/utils.go
hhhjort: OH, yes, that is a good catch. I will have to think about how it may be possible to test for. Can certainly rig a test to catch this one specific instance. But a test to generally catch that some conditional code in PBS Core might overwrite shared memory would be difficult, as it would require a test that covers all possible conditions and then checks for changes in shared memory. Hard to cover for conditions that haven't been written yet.

exchange/utils.go
hhhjort: Fixing, and this reminds me we don't have a way of distinguishing between an empty GPPSID and an unset GPPSID.

exchange/utils_test.go
hhhjort: That is an idea. But is it worth the work to move all the descriptions to comments in a test? It doesn't change the performance of the final program.

exchange/utils.go
hhhjort: Thanks, I forgot slices could be nil

exchange/utils.go
```diff
@@ -847,6 +854,60 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
 }
 
+func setLegacyGDPRFromGPP(r *openrtb2.BidRequest, gpp gpplib.GppContainer) {
+	if r.Regs != nil && r.Regs.GDPR == nil {
+		if r.Regs.GPPSID != nil {
+			// Set to 0 unless SID exists
+			r.Regs.GDPR = ptrutil.ToPtr[int8](0)
+			for _, id := range r.Regs.GPPSID {
+				if id == int8(gppConstants.SectionTCFEU2) {
+					regs := *r.Regs
```
SyntaxNode, line 864: Please move the copy up a level so that it also applies to `r.Regs.GDPR = ptrutil.ToPtr[int8](0)`.

exchange/utils.go
```diff
@@ -847,6 +854,60 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+		for i, id := range gpp.SectionTypes {
+			if id == gppConstants.SectionTCFEU2 {
+				if r.User == nil {
+					r.User = &openrtb2.User{
+						Consent: gpp.Sections[i].GetValue(),
+					}
+				}
+				user := *r.User
+				user.Consent = gpp.Sections[i].GetValue()
+				r.User = &user
```
SyntaxNode, line 882: Consider using an else clause to avoid setting the consent twice and making an unnecessary copy:

```
if r.User == nil {
	r.User = &openrtb2.User{
		Consent: gpp.Sections[i].GetValue(),
	}
} else {
	user := *r.User
	user.Consent = gpp.Sections[i].GetValue()
	r.User = &user
}
```

exchange/utils.go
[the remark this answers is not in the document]
SyntaxNode: Indeed :)

exchange/utils.go
SyntaxNode: No worries if it's too hard to instrument. Thank you for investigating.

exchange/utils_test.go
SyntaxNode: It doesn't impact performance. It's a little odd IMHO to see a `desc` field with no usage. We used to use it in generating assert messages, but that's no longer necessary with using `t.Run`.

exchange/utils.go
hhhjort: Oh, yes. This is where I need the facepalm emoji. Thanks.

exchange/utils.go
SyntaxNode: LOL. No worries.

exchange/utils.go
```diff
@@ -847,6 +854,60 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+
+	if r.User == nil || len(r.User.Consent) == 0 {
+		for i, id := range gpp.SectionTypes {
+			if id == gppConstants.SectionTCFEU2 {
+				if r.User == nil {
+					r.User = &openrtb2.User{
+						Consent: gpp.Sections[i].GetValue(),
+					}
+				} else {
+					user := *r.User
```
SyntaxNode, line 880: Please add a test for the else clause.

exchange/utils.go
```diff
@@ -847,6 +854,60 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+					user.Consent = gpp.Sections[i].GetValue()
+					r.User = &user
+				}
+			}
+		}
+	}
+
+}
+func setLegacyprivacyFromGPP(r *openrtb2.BidRequest, gpp gpplib.GppContainer) {
```
SyntaxNode, line 889: Nitpick: Consider `setLegacyUSPrivacyFromGPP` or `setLegacyUSPFromGPP` or `setLegacyCCPAFromGPP`.

exchange/utils_test.go
```diff
@@ -3476,3 +3478,293 @@ func TestCleanOpenRTBRequestsFilterBidderRequestExt(t *testing.T) {
...
+				Regs: &openrtb2.Regs{
+					GPPSID: []int8{6},
+					GDPR:   ptrutil.ToPtr[int8](0),
+				},
+				User: &openrtb2.User{
+					Consent: "GDPRConsent",
+				},
+			},
+		},
+	}
```
SyntaxNode, line 3644: I think we need to add tests to clearly define the behavior of a nil GPPSID. The only current case covering this scenario is "Empty". Perhaps consider a case where GPPSID is nil and there is a GPP string and a case where GPPSID is nil and there is not a GPP string.

exchange/utils.go
```diff
@@ -179,6 +181,11 @@ func (rs *requestSplitter) cleanOpenRTBRequests(ctx context.Context,
 			privacyEnforcement.Apply(bidderRequest.BidRequest)
 			allowedBidderRequests = append(allowedBidderRequests, bidderRequest)
 		}
+		// GPP downgrade: always downgrade unless we can confirm GPP is supported
+		if binfo, ok := rs.bidderInfo[string(bidderRequest.BidderCoreName)]; ok && (binfo.OpenRTB == nil || !binfo.OpenRTB.GPPSupported) {
```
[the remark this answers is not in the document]
hhhjort: Seems reasonable.

exchange/utils.go
hhhjort: Done

exchange/utils.go
hhhjort: I will go with `setLegacyUSPFromGPP`

SyntaxNode (approved):

exchange/utils.go
```diff
@@ -902,6 +919,60 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+			r.Regs = &regs
+		}
+	}
+
+	if r.User == nil || len(r.User.Consent) == 0 {
+		for i, id := range gpp.SectionTypes {
+			if id == gppConstants.SectionTCFEU2 {
+				if r.User == nil {
+					r.User = &openrtb2.User{
+						Consent: gpp.Sections[i].GetValue(),
```
VeronikaSolovei9, line 942: Is it always guaranteed that `gpp.SectionTypes` has same number of elements as `gpp.Sections`. Using index from `range gpp.SectionTypes` in `gpp.Sections[i]` seems risky. Do you think length check is needed here?

exchange/utils.go
```diff
@@ -902,6 +919,60 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+			for _, id := range r.Regs.GPPSID {
+				if id == int8(gppConstants.SectionTCFEU2) {
+					regs.GDPR = ptrutil.ToPtr[int8](1)
+				}
+			}
+			r.Regs = &regs
+		}
+	}
+
+	if r.User == nil || len(r.User.Consent) == 0 {
```
VeronikaSolovei9, line 937: Nitpick: does it look cleaner:
```
if r.User == nil || len(r.User.Consent) == 0 {
		for i, id := range gpp.SectionTypes {
			if id == gppConstants.SectionTCFEU2 {
				user := &openrtb2.User{}
				if r.User != nil {
					user = r.User
				}
				user.Consent = gpp.Sections[i].GetValue()
				r.User = user
			}
		}
	}
```

exchange/utils.go
```diff
@@ -902,6 +919,60 @@ func mergeBidderRequests(allBidderRequests []BidderRequest, bidderNameToBidderRe
...
+		return
+	}
+
+	if len(r.Regs.USPrivacy) > 0 || r.Regs.GPPSID == nil {
+		return
+	}
+	for _, id := range r.Regs.GPPSID {
+		if id == int8(gppConstants.SectionUSPV1) {
+			for i, id := range gpp.SectionTypes {
+				if id == gppConstants.SectionUSPV1 {
```
VeronikaSolovei9, line 965: This check duplicates L963: `if id == int8(gppConstants.SectionUSPV1)`

exchange/utils.go
hhhjort: Cleaned up

exchange/utils.go
hhhjort: It should not be a duplicate, but it is not clear that the second `id` is a new one in the inner scope. I will rename to ensure Go treats it correctly and avoids this confusion for readers.

hhhjort: This also has the assumption that the two arrays line up from before, so clearing that up as well.

VeronikaSolovei9: Thank you, looks good

VeronikaSolovei9 (approved):
Thank you for the changes, looks good

SyntaxNode (approved):

VeronikaSolovei9 (approved):

PR #2630, opened 2023-03-08
IX Bid Adapter: Fix overwriting of bids while building bid response

ccorbo:
This fixes a bug in the ix.go adapter.

Currently the code  adds the same pointer for every bid. And this temporary variable will be set to last bid in the last iteration (last element of the array), so now the bidderResponse.bids array will contain the same bid for every element which will be the last bid processed.

Fix: do not add the address of the loop variable, but the address of the array element.

Added unit test to show fix.


adapters/ix/ix_test.go
```diff
@@ -276,3 +276,82 @@ func TestMakeRequestsErrIxDiag(t *testing.T) {
...
+		Body:       body,
+	}
+
+	expectedBidCount := 2
+
+	bidResponse, _ := bidder.MakeBids(mockedReq, mockedExtReq, mockedRes)
+	assert.Equal(t, expectedBidCount, len(bidResponse.Bids))
+	assert.Equal(t, "1", bidResponse.Bids[0].Bid.ID)
+	assert.Equal(t, "2", bidResponse.Bids[1].Bid.ID)
+}
```
SyntaxNode, line 357: Please add a json test for multiple bids. That is our primary test framework.

ccorbo: Thanks, updated to add multibid.json which tests this scenario

adapters/ix/ix.go
```diff
@@ -186,7 +186,8 @@ func (a *IxAdapter) MakeBids(internalRequest *openrtb2.BidRequest, externalReque
 	var errs []error
 
 	for _, seatBid := range bidResponse.SeatBid {
-		for _, bid := range seatBid.Bid {
+		for i := 0; i < len(seatBid.Bid); i++ {
```
Sonali-More-Xandr, line 189: Nitpick - We can use index value returned by range of slice directly. The first value returned in case of array or slice is index. This can written simply as:-
`for i, bid := range seatBid.Bid`

ccorbo: Thanks updated

gargcreation1992 (approved):
lgtm!

Sonali-More-Xandr (approved):

PR #2632, opened 2023-03-09
Appnexus adapter - refactor makeRequest 

onkarvhanumante:
PR refactors makeRequest method


adapters/appnexus/appnexus.go
```diff
@@ -85,15 +85,10 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
...
-	if request.App != nil {
-		source, err1 := jsonparser.GetString(request.App.Ext, openrtb_ext.PrebidExtKey, "source")
-		version, err2 := jsonparser.GetString(request.App.Ext, openrtb_ext.PrebidExtKey, "version")
-		if (err1 == nil) && (err2 == nil) {
-			defaultDisplayManagerVer = fmt.Sprintf("%s-%s", source, version)
-		}
-	}
+	// appnexus adapter expects imp.displaymanagerver to be populated in openrtb2 endpoint
+	// but some SDKs will put it in imp.ext.prebid instead
+	defaultDisplayManagerVer := buildDefaultDisplayManageVer(request)
```
gargcreation1992, line 90: nitpick - I think we should not call this defaultDisplayManagerVer since we are building it using `req.App.Ext`. Should we just call it - `displayManagerVer`. This would also be inline with openrtb2 spec - `imp.displaymanagerver`

adapters/appnexus/appnexus.go
```diff
@@ -82,7 +82,7 @@ type appnexusReqExt struct {
 var maxImpsPerReq = 10
 
 func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.ExtraRequestInfo) ([]*adapters.RequestData, []error) {
-	memberIds := make(map[string]bool)
+	memberIds := make(map[string]struct{})
```
gargcreation1992, line 85: nitpick - I think this should be declared where it is being used. Before for loop, perhaps?

adapters/appnexus/appnexus.go
```diff
@@ -89,12 +89,18 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
...
 
-	var adPodId *bool
+	var (
+		adPodId         *bool
+		uniqueMemberIds []string
+	)
 
 	for i := 0; i < len(request.Imp); i++ {
 		memberId, impAdPodId, err := preprocess(&request.Imp[i], defaultDisplayManagerVer)
 		if memberId != "" {
```
gargcreation1992, line 107: We should check for the err here before checking any returned value from preprocess function.

gargcreation1992: nvm. this is addressed in subsequent commit.

adapters/appnexus/appnexus.go
```diff
@@ -95,7 +95,15 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 	)
 
 	for i := 0; i < len(request.Imp); i++ {
+		// If the preprocessing failed, the server won't be able to bid on this Imp. Delete it, and note the error.
 		memberId, impAdPodId, err := preprocess(&request.Imp[i], defaultDisplayManagerVer)
+		if err != nil {
+			errs = append(errs, err)
+			request.Imp = append(request.Imp[:i], request.Imp[i+1:]...)
```
gargcreation1992, line 102: I think we should refactor this more to make it simpler rather than modifying input request.Imp slice based on error. How about creating a new Imp slice and change the request.Imp at the end of the for loop.

adapters/appnexus/appnexus.go
```diff
@@ -223,7 +215,11 @@ func splitRequests(imps []openrtb2.Imp, request *openrtb2.BidRequest, requestExt
 	headers.Add("Content-Type", "application/json;charset=utf-8")
 	headers.Add("Accept", "application/json")
 
-	marshalAndSetRequestExt(request, requestExtension, errs)
+	var err error
+	request.Ext, err = json.Marshal(requestExtension)
+	if err != nil {
+		errs = append(errs, err)
+	}
```
gargcreation1992, line 218: nitpick - I think this could also be written as - 
```
if request.Ext, err := json.Marshal(requestExtension); err != nil {
         errs = append(errs, err)
}
```

adapters/appnexus/appnexus.go
```diff
@@ -82,52 +82,50 @@ type appnexusReqExt struct {
...
-			defaultDisplayManagerVer = fmt.Sprintf("%s-%s", source, version)
-		}
-	}
-	var adPodId *bool
+	// appnexus adapter expects imp.displaymanagerver to be populated in openrtb2 endpoint
+	// but some SDKs will put it in imp.ext.prebid instead
+	defaultDisplayManagerVer := buildDefaultDisplayManageVer(request)
+
+	var (
+		adPodId         *bool
```
gargcreation1992, line 93: adPodId represents an Id not a bool. Should we consider renaming it something that specifies boolean. For example - `hasImpAdPodId` or `IsAdPodIdDefined` or something else.

gargcreation1992 (changes requested):
left few suggestions.

adapters/appnexus/appnexus.go
onkarvhanumante: renamed as part of https://github.com/prebid/prebid-server/pull/2632/commits/1750cdf3934b374d82917e4326dc94f3dc39514e

adapters/appnexus/appnexus.go
onkarvhanumante: good call 👍🏼 

https://github.com/prebid/prebid-server/pull/2632/commits/5592ee5355b26355ac5565fd018328d4fba79db4 refactor append logic

adapters/appnexus/appnexus.go
onkarvhanumante: https://github.com/prebid/prebid-server/pull/2632/commits/0a16005e7f948517fd599b62c4f882d0334f3b6b makes the changes

adapters/appnexus/appnexus.go
onkarvhanumante: https://github.com/prebid/prebid-server/commit/0a16005e7f948517fd599b62c4f882d0334f3b6b makes the changes

adapters/appnexus/appnexus.go
```diff
@@ -31,52 +31,52 @@ type adapter struct {
...
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
 		if memberId != "" {
-			memberIds[memberId] = true
+			if _, ok := memberIds[memberId]; !ok {
+				memberIds[memberId] = struct{}{}
+				uniqueMemberIds = append(uniqueMemberIds, memberId)
```
Sonali-More-Xandr, line 58: Good change!

adapters/appnexus/appnexus.go
```diff
@@ -119,26 +119,22 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 	// For this all impressions in  request should belong to the same pod
 	// If impressions number per pod is more than maxImpsPerReq - divide those imps to several requests but keep pod id the same
 	// If  adpodId feature disabled and impressions number per pod is more than maxImpsPerReq  - divide those imps to several requests but do not include ad pod id
-	if isVIDEO == 1 && *adPodId {
+	if isVIDEO == 1 && *shouldGenerateAdPodId {
```
Sonali-More-Xandr, line 122: Can we move this code to a different function?

onkarvhanumante: https://github.com/prebid/prebid-server/pull/2632/commits/c1ec4ac2d47ee9f35d2e88ef17d6a0e7389cf396 extracts logic into separate function

Sonali-More-Xandr (approved):
LGTM!

gargcreation1992 (approved):

PR #2635, opened 2023-03-10
AppNexus Adapter - Add Request Copy

Sonali-More-Xandr:
In the current adapter code, we are directly updating the request.Imp object. We are deleting the impression, if the preprocessing fails. Instead of deleting impression from the request object, we now create a shallow copy of the request object and perform operations on this copy.

### DIFF
--- a/adapters/appnexus/appnexus.go
+++ b/adapters/appnexus/appnexus.go
@@ -163,8 +163,6 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 	reqExt.Appnexus.IsAMP = isAMP
 	reqExt.Appnexus.HeaderBiddingSource = a.hbSource + isVIDEO
 
-	imps := requestCopy.Imp
-
 	// For long form requests if adpodId feature enabled, adpod_id must be sent downstream.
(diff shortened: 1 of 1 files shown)


adapters/appnexus/appnexus.go
```diff
@@ -162,7 +163,7 @@ func (a *adapter) MakeRequests(request *openrtb2.BidRequest, reqInfo *adapters.E
 	reqExt.Appnexus.IsAMP = isAMP
 	reqExt.Appnexus.HeaderBiddingSource = a.hbSource + isVIDEO
 
-	imps := request.Imp
+	imps := requestCopy.Imp
```
onkarvhanumante, line 166: `requestCopy` is already a copy. So no need to again create a new copy by assigning value to `imps`

Sonali-More-Xandr: addressed in [eae473c](https://github.com/prebid/prebid-server/pull/2635/commits/eae473c30e193c5f1fd16e65eb37dba9ae962818)

Sonali-More-Xandr:
Closing this PR as the shallow copy is not required here since imps is already copied and it does not affect the actual object.

PR #2637, opened 2023-03-13
adtrgtme: add bidder param site_id

grakholsky:
* add param tests and update current tests


static/bidder-params/adtrgtme.json
```diff
@@ -4,5 +4,12 @@
...
 
-  "properties": {}
+  "properties": {
+    "site_id": {
+      "type": "integer",
+      "description": "Site ID"
+    }
+  },
+
+  "required": ["site_id"]
```
Sonali-More-Xandr, line 14: Adding a required parameter may cause the request tot fail. As per the design,  any request that has any bidder that fails the JSON schema will invalidate the entire request. The idea was that this would only happen when a publisher creates/updates their prebid config. When they do this, they should be looking to see if prebid is returning any bids, but probably aren't sensitive to if all bidders are bidding. They would otherwise not notice that a bidder is misconfigured until at some point they look at their analytics and notice the bidder is not bidding on a particular page or configuration setup. We think the current design choice is the better option to ensure everyone has a chance to bid. 
The recommended way here is to add `site_id` as optional and then have the adapter check and return an error if `site_id` is missing. If the parameters fail the schema check, we fail the entire request. If an adapter returns an error, we allow the other adapters to keep going and don't fail the whole request. At a later point in time, when the publishers have all had a chance to update to the new schema, we can make it required here. We just don't want all the publisher who are currently using your adapter to suddenly have all their prebid configs invalidated by surprise.

grakholsky: Hello!
A very long description, but it still needs to be clarified why we can't use required fields as opposed to other adapters.

Sonali-More-Xandr: If there is no traffic to this adapter then it is safe to add a required field.

gargcreation1992: @grakholsky Thank you for being patient on this one!
The suggestion was if adapter suddenly requires site_id to work then the integration with the current publishers might fail as they might need some time to add this required field in their request. Now since you have clarified that there is no traffic to this adapter at the moment then this PR is good to go in that perspective. Hope this helps!

adapters/adtrgtme/adtrgtme.go
```diff
@@ -65,25 +74,24 @@ func (v *adapter) MakeRequests(
...
-		return "", &errortypes.BadInput{
-			Message: "request.Site.ID is not provided",
-		}
-	} else if openRTBRequest.App != nil {
-		if openRTBRequest.App.ID != "" {
-			return fmt.Sprintf("%s?s=%s&prebid", v.endpoint, openRTBRequest.App.ID), nil
-		}
-		return "", &errortypes.BadInput{
-			Message: "request.App.ID is not provided",
+func (v *adapter) getImpressionExt(imp *openrtb2.Imp) (openrtb_ext.ExtImpAdtrgtme, error) {
```
gargcreation1992, line 77: this function doesn't have to be defined as pointer receiver on adapter as this function is not using any of params defined on adapter. So simply - 
``` 
func getImpressionExt(imp *openrtb2.Imp) (openrtb_ext.ExtImpAdtrgtme, error) {
\\
}
```

adapters/adtrgtme/adtrgtme.go
```diff
@@ -32,9 +32,24 @@ func (v *adapter) MakeRequests(
...
+	var ext openrtb_ext.ExtImpAdtrgtme
 	var requests []*adapters.RequestData
 	var errors []error
 
+	for i, imp := range openRTBRequest.Imp {
+		var err error
+		ext, err = v.getImpressionExt(&imp)
+		if err != nil {
+			errors = append(errors, err)
+			break
```
gargcreation1992, line 44: Do you intend to break out of the imp loop if `getImpressionExt` fails or would you like to continue for remaining imps?

adapters/adtrgtme/adtrgtme.go
grakholsky: If it is critical for the adapter and it will somehow affect the work. We do it as a separate function.

adapters/adtrgtme/adtrgtme.go
grakholsky: I don't understand what you mean.

adapters/adtrgtme/adtrgtme.go
```diff
@@ -32,9 +32,24 @@ func (v *adapter) MakeRequests(
 	[]*adapters.RequestData,
 	[]error,
 ) {
+	var ext openrtb_ext.ExtImpAdtrgtme
 	var requests []*adapters.RequestData
 	var errors []error
 
+	for i, imp := range openRTBRequest.Imp {
```
gargcreation1992, line 39: When ranging over a slice, two values are returned for each iteration. The first is the index, and the second is a copy of the element at that index. 
I think you intend to pass the openRTBRequest.Imp directly in the for loop and not the copy of the `openRTBRequest.Imp[i]`. So, I would suggest to access imp using index instead. Here is the sample - 
```
for i, _ := range openRTBRequest.Imp {
\\ access imp using openRTBRequest.Imp[i] <-- this would be the actual openRTBRequest.Imp[i] and not copy of openRTBRequest.Imp[i]

}

File: adapters/adtrgtme/adtrgtme.go (hunk 3 above)
Comment (gargcreation1992) [reply adapters/adtrgtme/adtrgtme.go:44]: So the `break` statement here breaks out of the for loop. This would mean that if there are remaining `openRTBRequest.Imp[i]` impression then it will not be processed. Do you intend to keep this behaviour or would you like to continue to process for remaining impressions?

Comment (gargcreation1992) [reply adapters/adtrgtme/adtrgtme.go:44]:
```
if err != nil {
			errors = append(errors, err)
			continue
}
```
The above code will continue to process for the remaining impressions in the for loop.

File: adapters/adtrgtme/adtrgtme.go (hunk 2 above)
Comment (gargcreation1992) [reply adapters/adtrgtme/adtrgtme.go]: I am not suggesting you to remove this function. 
In `getImpressionExt` function, you are not accessing any value of the adapter struct. I am suggesting to not define it as pointer receiver. Here is the code that you can use -
```
func getImpressionExt(imp *openrtb2.Imp) (openrtb_ext.ExtImpAdtrgtme, error) {
		  return openrtb_ext.ExtImpAdtrgtme{}, &errortypes.BadInput{
			  Message: "ext.bidder not provided",
		  }
	  }
	  
  var ext openrtb_ext.ExtImpAdtrgtme
	  if err := json.Unmarshal(bidderExt.Bidder, &ext); err != nil {
		  return openrtb_ext.ExtImpAdtrgtme{}, &errortypes.BadInput{
			  Message: "ext.bidder not provided",
		  }
	  }
	  return ext, nil
}
```

File: adapters/adtrgtme/adtrgtme.go (hunk 4 above)
Comment (grakholsky) [reply adapters/adtrgtme/adtrgtme.go:39]: It doesn't matter to us. We will receive impressions with the same extensions and need to get any of them. And remove extensions from the outgoing HTTP requests to our server.

Comment (grakholsky) [reply adapters/adtrgtme/adtrgtme.go:39]: For this reason, we overwrite it in the same variable `ext`
```
var ext openrtb_ext.ExtImpAdtrgtme
....
for i, imp := range openRTBRequest.Imp {
var err error
ext, err = getImpressionExt(&imp)
```

File: adapters/adtrgtme/adtrgtme.go (hunk 2 above)
Comment (grakholsky) [reply adapters/adtrgtme/adtrgtme.go]: I do not write that we will remove it. We made it function as you requested.

File: adapters/adtrgtme/adtrgtme.go (hunk 3 above)
Comment (grakholsky) [reply adapters/adtrgtme/adtrgtme.go:44]: I answered below https://github.com/prebid/prebid-server/pull/2637#discussion_r1145785244

Comment (gargcreation1992) [reply adapters/adtrgtme/adtrgtme.go:44]: This comment is different from https://github.com/prebid/prebid-server/pull/2637#discussion_r1145785244
Rephrasing - Suppose your adapter has request for multiple impressions in single request - imp1, imp2, imp3. 
In the for loop, getImpressionExt returns an error then line 44 `break` statement will not process imp2 and imp3 impressions. Would that be ok for your adapter since it may be the case where imp2 and imp3 were the valid Imps and your adapter could have bid on it.
Also could you please add a test case for multiple impressions?

File: adapters/adtrgtme/adtrgtme.go [hunk 5]
```diff
@@ -18,7 +18,7 @@ type adapter struct {
 }
 
 // Builder builds a new instance of the Adtrgtme adapter for the given bidder with the given config.
-func Builder(bidderName openrtb_ext.BidderName, config config.Adapter, server config.Server) (adapters.Bidder, error) {
+func Builder(_ openrtb_ext.BidderName, config config.Adapter, _ config.Server) (adapters.Bidder, error) {
```
gargcreation1992, line 21: I guess it is ok to pass name instead of `_`. You should not have to change this and let's be consistent.

adapters/adtrgtme/adtrgtme.go
```diff
@@ -27,7 +27,7 @@ func Builder(bidderName openrtb_ext.BidderName, config config.Adapter, server co
 
 func (v *adapter) MakeRequests(
 	openRTBRequest *openrtb2.BidRequest,
-	requestInfo *adapters.ExtraRequestInfo,
+	_ *adapters.ExtraRequestInfo,
```
gargcreation1992, line 30: Same comment here. Let's pass in the name instead of `_` to be consistent with other adapters.

adapters/adtrgtme/adtrgtme.go
```diff
@@ -132,7 +130,7 @@ func (v *adapter) checkResponseStatusCodes(response *adapters.ResponseData) erro
 
 func (v *adapter) MakeBids(
 	openRTBRequest *openrtb2.BidRequest,
-	requestToBidder *adapters.RequestData,
+	_ *adapters.RequestData,
```
gargcreation1992, line 133: let's be consistent and add name instead of passsing `_`.

adapters/adtrgtme/adtrgtme.go
```diff
@@ -36,54 +36,52 @@ func (v *adapter) MakeRequests(
...
-		return "", &errortypes.BadInput{
-			Message: "request.Site.ID is not provided",
-		}
-	} else if openRTBRequest.App != nil {
-		if openRTBRequest.App.ID != "" {
-			return fmt.Sprintf("%s?s=%s&prebid", v.endpoint, openRTBRequest.App.ID), nil
-		}
-		return "", &errortypes.BadInput{
-			Message: "request.App.ID is not provided",
+func siteIDFromImp(imp *openrtb2.Imp) (uint64, error) {
```
gargcreation1992, line 67: I think a better function name should be `getSitedIdFromImp`.

adapters/adtrgtme/adtrgtme.go
```diff
@@ -36,54 +36,52 @@ func (v *adapter) MakeRequests(
...
-		requestCopy.Imp = []openrtb2.Imp{imp}
-
-		requestJSON, err := json.Marshal(openRTBRequest)
+	for i := range openRTBRequest.Imp {
+		siteID, err := siteIDFromImp(&openRTBRequest.Imp[i])
 		if err != nil {
 			errors = append(errors, err)
 			continue
 		}
+		openRTBRequest.Imp[i].Ext = nil
```
gargcreation1992, line 45: Warning that you are modifying the original openRTBRequest.Imp[i].Ext and we should not ideally be modifying the input impression. I think this for loop could be re-written as - 
```
for _, imp := range openRTBRequest.Imp {
                 siteID, err := getSiteIDFromImp(&imp)
		if err != nil {
			errors = append(errors, err)
			continue
		}
                 
                 imp.Ext = nil
                 requestCopy.Imp = []openrtb2.Imp{imp}

                 requestBody, err := json.Marshal(requestCopy)
		if err != nil {
			errors = append(errors, err)
			continue
		}
		
		requestData := &adapters.RequestData{
			Method:  http.MethodPost,
			Uri:     v.buildRequestURI(siteID),
			Body:    requestBody,
			Headers: makeRequestHeaders(openRTBRequest),
		}
		
		requests = append(requests, requestData)
}
```

gargcreation1992 (approved):
lgtm now! thank you for being patient on this one.

Sonali-More-Xandr (approved):

PR #2638, opened 2023-03-14
Appnexus Adapter - refactor makeBids function

gargcreation1992:


util/bidderutil/bidderutil.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/prebid/prebid-server/adapters"
+	"github.com/prebid/prebid-server/errortypes"
+)
+
+func CheckResponseStatusCodeForErrors(response *adapters.ResponseData) error {
```
Sonali-More-Xandr, line 11: Good change!

adapters/appnexus/appnexus.go
```diff
@@ -379,13 +378,10 @@ func getMediaTypeForBid(bid *appnexusBidExt) (openrtb_ext.BidType, error) {
 }
 
 // getIabCategoryForBid maps an appnexus brand id to an IAB category.
-func (a *adapter) getIabCategoryForBid(bid *appnexusBidExt) (string, error) {
+func (a *adapter) findIabCategoryForBid(bid *appnexusBidExt) (string, bool) {
 	brandIDString := strconv.Itoa(bid.Appnexus.BrandCategory)
-	if iabCategory, ok := a.iabCategoryMap[brandIDString]; ok {
-		return iabCategory, nil
-	} else {
-		return "", fmt.Errorf("category not in map: %s", brandIDString)
```
Sonali-More-Xandr, line 387: with new changes, we are losing this error statement. Can we handle the !ok as well?

gargcreation1992: We are actually not using the error statement in the code except at this place. Neither there was a test case written for this. In the previous version of code, this error was not appended anywhere in the final response. So we should not keep it unless we use it. This particular function gets the iabCategory if present else should return empty string.

adapters/appnexus/appnexus.go
```diff
@@ -303,63 +304,61 @@ func makeKeywordStr(keywords []*openrtb_ext.ExtImpAppnexusKeyVal) string {
...
+				errs = append(errs, err)
+				continue
 			}
+
+			iabCategory, found := a.findIabCategoryForBid(&bidExt)
+			if found {
+				bid.Cat = []string{iabCategory}
+			}
+
+			if len(bid.Cat) > 1 {
```
Sonali-More-Xandr, line 343: can we keep this if statement as the else part of `if found` statement as was done in the previous code

util/bidderutil/bidderutil.go
```diff
@@ -0,0 +1,27 @@
+package bidderutil
```
onkarvhanumante, line 1: few thoughts here,

- File name should be in snake case i.e `bidder_util.go`
- May be no need for creating separate package for this util. This can be added `adapters` package similar to other utils we have today.
  ```
      adpapters/openrtb_util.go
       adpapters/bidder_util.go
  ```
- also should we rename `bidder_util.go` to `response_util.go`

util/bidderutil/bidderutil.go
```diff
@@ -0,0 +1,27 @@
...
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode)
+	}
+
+	return nil
+}
+
+func IsResponseCodeStatusNoContent(response *adapters.ResponseData) bool {
```
onkarvhanumante, line 25: Generally we say `http status code`. From this perspective, how about renaming `IsResponseCodeStatusNoContent` to `IsResponseStatusCodeNoContent`  or `hasNoContentStatusCode` ?

util/bidderutil/bidderutil.go
```diff
@@ -0,0 +1,27 @@
...
+
+func CheckResponseStatusCodeForErrors(response *adapters.ResponseData) error {
+	if response.StatusCode == http.StatusBadRequest {
+		return &errortypes.BadInput{
+			Message: fmt.Sprintf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode),
+		}
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("Unexpected status code: %d. Run with request.debug = 1 for more info", response.StatusCode)
```
onkarvhanumante, line 19: should add test case to assert type of errors `errortypes.BadInput, error` returned

adapters/appnexus/appnexus.go
gargcreation1992: https://github.com/prebid/prebid-server/pull/2638/commits/9f17bab35ba64c9c810c4040794e7c6c6c6a67d5

util/bidderutil/bidderutil.go
gargcreation1992: [87d05e0](https://github.com/prebid/prebid-server/pull/2638/commits/87d05e0af3b5db96bcbc7e269be6d79a3c99a522)

util/bidderutil/bidderutil.go
gargcreation1992: [87d05e0](https://github.com/prebid/prebid-server/pull/2638/commits/87d05e0af3b5db96bcbc7e269be6d79a3c99a522)

util/bidderutil/bidderutil.go
gargcreation1992: > * File name should be in snake case i.e `bidder_util.go`
I am being consistent with what we have in utils package.
> * May be no need to create separate package for this util. This can be added  in `adapters` package similar to other utils we have today.
There is already a utils folder under adapters. So this should be fine. I have changed the location to httputil package instead of creating a new one.
>   ```
>       adpapters/openrtb_util.go
>       adpapters/bidder_util.go
>   ```
> * also should we rename `bidder_util.go` to `response_util.go`
renamed to responseutil.go in consistent with existing file names.
[87d05e0](https://github.com/prebid/prebid-server/pull/2638/commits/87d05e0af3b5db96bcbc7e269be6d79a3c99a522)

Sonali-More-Xandr (approved):

onkarvhanumante (approved):

PR #2653, opened 2023-03-17
New Adapter: Evolution Technologies

imedvedko:
Docs: https://github.com/prebid/prebid.github.io/pull/4421


bsardo:
Hi @imedvedko. I just wanted to check if there is any relationship between these two because the naming might be a bit confusing to some:
```
	BidderEVolution             BidderName = "e_volution"
	BidderEvolutionTechnologies BidderName = "evolutionTechnologies"
```

apykhteyev:
Hi @bsardo . This company is a customer who's using our solution and wishes to white label it under the name `evolutionTechnologies` because already have some in-house implementation `e_volution`.  Unfortunately, I don't know the current status of their in-house solution, but we were asked to use such a naming. Hope it clarifies

static/bidder-info/iionads.yaml
```diff
@@ -1,4 +1,4 @@
-endpoint: "http://{{.Host}}.iionads.com/openrtb/{{.PublisherID}}"
+endpoint: "http://ads-pbs.iionads.com/openrtb/{{.PublisherID}}?host={{.Host}}"
```
onkarvhanumante, line 1: This change doesn't seems specific to adding `evolution technologies` adapter

static/bidder-info/limelightDigital.yaml
```diff
@@ -1,4 +1,4 @@
-endpoint: "http://{{.Host}}.ortb.net/openrtb/{{.PublisherID}}"
+endpoint: "http://ads-pbs.ortb.net/openrtb/{{.PublisherID}}?host={{.Host}}"
```
onkarvhanumante, line 1: similar to https://github.com/prebid/prebid-server/pull/2653/files#r1143367385, change doesn't seems specific to adding evolution technologies adapter

onkarvhanumante (approved):

SyntaxNode:
> Hi @bsardo . This company is a customer who's using our solution and wishes to white label it under the name `evolutionTechnologies` because already have some in-house implementation `e_volution`. Unfortunately, I don't know the current status of their in-house solution, but we were asked to use such a naming. Hope it clarifies

Could you please reach out to your customer to explain why they need two separate bidders? Per https://github.com/prebid/prebid-server/issues/2654 we're trying to be more consistent across Prebid.org Projects. Do they plan to deprecate `e_volution` in favor of `evolutionTechnologies`?

@bretg This will cause a targeting conflict with the `evolution` bidder in PBS-Java.

bretg:
@imedvedko - please consider a different biddercode, since this name conflicts with an existing code for in the first 6 letters. See https://docs.prebid.org/prebid-server/developers/add-new-bidder-go.html#choose-a-name

 e.g. How about "evtech" as the bidder code? Shorter is better for ad server targeting.

SyntaxNode:
FWIW: "evtech" sounds great to me.

apykhteyev:
> Hi @bsardo . This company is a customer who's using our solution and wishes to white label it under the name `evolutionTechnologies` because already have some in-house implementation `e_volution`. Unfortunately, I don't know the current status of their in-house solution, but we were asked to use such a naming. Hope it clarifies

Just got info - that the customer wants to keep both adapters for now. I hope there will not be massive issues to keep both.

onkarvhanumante (approved):

